PR #155, opened 2015-01-03
Ikvm video

coolo:


backend/VNC.pm
```diff
@@ -486,6 +518,11 @@ sub receive_message {
       : $message_type == 1     ? $self->_receive_colour_map()
       : $message_type == 2     ? $self->_receive_bell()
       : $message_type == 3     ? $self->_receive_cut_text()
+      : $message_type == 0x39  ? $self->_receive_ikvm_session()
+      : $message_type == 0x04  ? $self->_discard_ikvm_message(20)
+      : $message_type == 0x16  ? $self->_discard_ikvm_message(1)
+      : $message_type == 0x37  ? $self->_discard_ikvm_message(2)
+      : $message_type == 0x3c  ? $self->_discard_ikvm_message(8)
```
aaannz: I noticed message type 0x33 ("video-get-info" as noted in _discard_ikvm_message comment) is missing in the list above. Is this intentional?

aaannz:
Only one question, otherwise LGTM, but note that I'm not familiar with iKVM at all. ;)

backend/VNC.pm
coolo: actually in my testing I didn't notice a single of these messages beside 0x39, which is part of the login. All the others I only added for completness out of the ObjectiveC code. We can add 0x33 for further completness, but I don't think it will matter :)

coolo:
pushed it now with an additional commit to ignore 0x33

PR #156, opened 2015-01-04
support keyboard on ikvm

coolo:
ikvm has different key codes which makes the whole mapping a bit more
complex.


coolo:
I tried to be careful not to break regular VNC, but as I had to touch the code I can't say for sure - but I did a basic test

PR #158, opened 2015-01-08
Major refactoring of the qemu backend

coolo:
This basically splits the qemu backend into 4 parts:

- the backend::driver class, which lives in the main thread and implements $bmwqemu::backend
  it passes all commands to a 2nd thread
- backend::baseclass implements the basics of this 2nd thread, but is abstract
- backend::vnc_backend implements a generic VNC backend
- backend::qemu implements the start of qemu and the connection of VNC and derives from the above

backend::ipmi is there, but not tested in this scenario. I concentrated on pulling qemu apart to be reusable for ipmi


coolo:
KVM on ipmi works now: http://stephan.kulow.org/ipmi.ogv

PR #159, opened 2015-01-09
Set default resolution for -vga std

k0da:
Override default screen resolution for distributons where bochs_drm
driver is not avaialble.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -77,7 +77,7 @@ sub start_qemu($) {
         $vars->{QEMUVGA} ||= "cirrus";
     }
     else {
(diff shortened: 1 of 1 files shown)


coolo:
"Option g not supported for this target" because man qemu says: ppc and sparc only

k0da:
-vga std is isolated to cases where OFW is spcified.

coolo:
I see

coolo:
you didn't test that ... https://github.com/os-autoinst/os-autoinst/commit/a3e9c86c7484d7d5319566a939aa129aa29f977f

k0da:
sorry for that

PR #161, opened 2015-01-15
[needs testing] Add serial support for IPMI

asdil12:
This patch is not yet tested.

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -161,7 +161,7 @@ sub start_serial_grab() {
 sub stop_serial_grab($) {
     my $self = shift;
     return unless $self->{'serialpid'};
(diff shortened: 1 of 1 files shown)


coolo:
doesn't work

asdil12:
This seems to work (as far as I can test - that means that I can at least see the BIOS messages in the serial0 file).

asdil12:
@coolo Could you give it a try and merge if it is ok?

coolo:
I gave it a try and the serial0 file indeed contains useful stuff. but it also contains so many 0 bytes that it's basically useless ;(

I guess we need some kind of post processing

coolo:
rebased and pushed to master

PR #162, opened 2015-01-16
enable for perl 5.20 (Tumbleweed)

froh:


coolo:
perl 5.20 is on 13.2 too and about every openQA developer uses TW, so I wonder what's special about your perl :)

PR #163, opened 2015-01-18
Be generally more careful in qemu shutdown

coolo:
- define a die handler in the backend thread
  to makes sure the VM is stopped
- don't send qemu a TERM 3 times, the steps are TERM+sleep+KILL


coolo:
need to deploy this - verified it fixes the current problems

PR #166, opened 2015-01-20
fix typing capital letters in VNC

coolo:
The charmap in testapi was correct, but unused

### DIFF
--- a/backend/VNC.pm
+++ b/backend/VNC.pm
@@ -501,7 +501,6 @@ sub init_ikvm_keymap {
     for my $key ("a".."z") {
         my $code = 0x4 + ord($key) - ord('a');
         $self->keymap->{$key} = $code;
-        $self->keymap->{uc($key)} = [$self->keymap->{'shift'}, $code];
     }
     for my $key ("1".."9") {
         $self->keymap->{$key} = 0x1e + ord($key) - ord('1');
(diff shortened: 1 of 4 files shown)


lnussel:
on my workstation this commit breaks typing 'ttyS0' with captial S in the bootloader

coolo:
but types ttys0?

coolo:
confirmed ;(

coolo:
I don't get why though. https://tools.ietf.org/html/rfc6143#section-7.5.4 is the section in doubt and this is what we're sending:

type_string map Y shift-y
_send_key_event 1 ffe1
_send_key_event 1 79
_send_key_event 0 79
_send_key_event 0 ffe1
type_string map 2 2
_send_key_event 1 32
_send_key_event 0 32
type_string map D shift-d
_send_key_event 1 ffe1
_send_key_event 1 64
_send_key_event 0 64
_send_key_event 0 ffe1
type_string map E shift-e
_send_key_event 1 ffe1
_send_key_event 1 65
_send_key_event 0 65
_send_key_event 0 ffe1

but we get y2de - shift is gently ignored. But not on power, where this was necessary ;(

coolo:
in case you need a quick workaround, look in backend/baseclass.pm for this line:
 $self->{charmap}->{$c} = "shift-\L$c";
and uncomment it. Then you get back the old behaviour of pressing Y instead of shift-y

k0da:
I'm fixing the original issue. So shift-letter will not be nescessary anymore

coolo:
ok, I pushed c956545 to do the shift-letter only for ipmi again

PR #168, opened 2015-01-22
Fix boot order for PowerPC

k0da:
On qemu-system-ppc64 spapr machine, once=d boot option doesn't work.
Use order=cd for now.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -285,8 +285,11 @@ sub start_qemu() {
         if ( $vars->{PXEBOOT} ) {
             push( @params, "-boot", "n");
         }
-        else {
+        elsif ( !$vars->{OFW} ) {
             push( @params, "-boot", "once=d,menu=on,splash-time=5000" );
+        }
+        else {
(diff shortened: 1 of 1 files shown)


coolo:
how does this boot ever the cd then?

k0da:
When we start the test, disk doesn't have a bootloader, so it tries "c" (disk first) can't find bootloader then tries "d" (cdrom) -> succeeds. On first reboot we have bootloader on the disk. So we boot from it.

coolo:
and how would you boot the cd in update test cases? IMO we need to get rid of this bizare logic and eject the CD if we don't want it anymore.

k0da:
I hope to get once option get fixed soon.

coolo:
we need the explicit eject for IPMI anyway (where we don't support CDs yet, but will hopefully at some point)

coolo:
Can we close this now?

k0da:
Yes, this fix is found. There are even two versions of it submitted upstream. Mean while I'll update qemu package and send to Rudi

PR #169, opened 2015-01-22
add API to record a soft failure (flag formerly known as dent)

coolo:


coolo:
I didn't yet test that, but I wanted to record it

coolo:
Now tested (and fixed)

PR #173, opened 2015-01-28
Enable s390 vnc

froh:


coolo:
that sounds like you need a rebase

froh:
yah.  I *love* rebasing.

and I liked your pointer to AUTOLOAD.  I use that to dynamically proxy the s3270 console in the test run.  all calls to the s3270 are wrapped, and the backend unwraps them, calls the backend function, wraps the result and sends it back to the caller.

I wrap all and any arguments, btw, so the test case just calls `$s3270->do_fancy_stuff(arg1, k1=>v1, k2=>v2)` as it did before.

For now, I do not create a replacement function dynamically, I just do the AUTOLOAD call every time.

coolo:
.... and tidy :)

backend/VNC.pm
```diff
@@ -9,10 +9,16 @@ use Time::HiRes qw( usleep );
 use Carp;
 use tinycv;
 
+use Crypt::DES;
+
+use Carp qw(confess cluck carp croak);
+use Data::Dumper qw(Dumper);
+use feature qw/say/;
+
```
coolo: if you use new modules, add them DEPENDENCIES.txt

backend/baseclass.pm
```diff
@@ -22,7 +22,6 @@ our $backend;
 sub new {
     my $class = shift;
     my $self = bless( { class => $class }, $class );
-    $self->init();
     $self->{'started'} = 0;
```
coolo: hmm, why doesn't crash that all the time? :)

coolo: I mean in master

ppmclibs/tinycv_impl.cc
```diff
@@ -438,12 +438,19 @@ void image_map_raw_data(Image *a, const unsigned char *data)
 }
 
 
-void image_map_raw_data_rgb555(Image *a, const unsigned char *data)
+void image_map_raw_data_rgb555(Image *a, const unsigned char *data, bool big_endian)
 {
```
coolo: as you didn't change the call into this, I assume you broke it. But I suggest the real thing to do is reverting the rgb555 changes.

I will remove the whole function and replace it with a proper call to the _full version once I can test IPMI again.

backend/baseclass.pm
froh: because I had the change locally and it was an oversight not pushing it yet.
I now prefer to git stash my local changes and test before I push...

ppmclibs/tinycv_impl.cc
froh: ok so I don't do anything there?

backend/VNC.pm
froh: done.
tidy, too.

backend/driver.pm
```diff
@@ -223,6 +227,14 @@ sub _read_json($) {
     return $hash;
 }
 
+###################################################################
+# enable _send_json to send regular expressions
+package Regexp;
```
coolo: add a line feed after the package to mark an overwrite

backend/s390x.pm
```diff
@@ -13,66 +14,125 @@ use feature qw/say/;
...
+    return $self;
+}
 
-sub init() {
+sub setup_3270_console() {
     my $self = shift;
 
-    my $vars = \%bmwqemu::vars;
+    confess "ZVMHOST unset in vars.json" unless get_var("ZVM_HOST", undef);
+    confess "ZVM_GUEST unset in vars.json" unless get_var("ZVM_GUEST", undef);
```
coolo: why the undef? it's default

backend/driver.pm
froh: [ ] done.

ppmclibs/tinycv_impl.cc
froh: fixed.

backend/s390x.pm
froh: removed.

ppmclibs/tinycv_impl.cc
```diff
@@ -442,8 +442,7 @@ void image_map_raw_data_rgb555(Image *a, const unsigned char *data)
...
   for (int y = 0; y < a->img.rows; y++) {
     for (int x = 0; x < a->img.cols; x++) {
-      long pixel = *data++;
-      pixel += *data++ * 256;
+      long pixel;
       unsigned char blue = pixel % 32 * 8;
```
coolo: quite uninitialized now

froh: duh.

we should add -Wall to Makefile.PL, somehow.

coolo:
will add then when I push

PR #176, opened 2015-01-29
increase dents when found the workaround property

nilxam:


basetest.pm
```diff
@@ -93,6 +93,16 @@ sub record_screenmatch($$;$) {
         properties => [@$properties],
     };
 
+    # When found the needle had workaround property
+    # mark the result as dent and increase the dents
+    for my $property (@$properties) {
+        if ($property eq 'workaround') {
+            $result->{dent} = 1;
```
coolo: this dent in results is unused - we better clean this. but unrelated to your commit

PR #179, opened 2015-01-29
use virtio-blk-device on aarch64

lnussel:
aarch64 doesn't have pci support yet and virtio-blk is an alias for
virtio-blk-pci


coolo:
tidy

lnussel:
hmm, we could achieve the same with job templates that set HDDMODEL=virtio-blk-device

PR #181, opened 2015-02-06
advance s390 backend

froh:
- turn 'gnah' hack into  console proxy
- clean up snapshot timer handling
- cleanly die if the remote vnc server is not reachable, even if it doesn't disconnect cleanly


froh:
will clean up as instructed and split into individual pull requests

PR #182, opened 2015-02-10
Implement multipath scenario

k0da:
This patch implements multipath scenario. Basically we export the same
divice twice.

The mandatory here is we have to use virtio-scsi which implements serial
attribute of a device.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -165,6 +165,10 @@ sub start_qemu() {
     $vars->{NUMDISKS}  ||= 1;
     $vars->{HDDSIZEGB} ||= 10;
     $vars->{HDDMODEL}  ||= "virtio-blk";
+    if ($vars->{MULTIPATH}) {
+        $vars->{HDDMODEL} = "virtio-scsi-pci";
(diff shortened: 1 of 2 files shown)


backend/qemu.pm
```diff
@@ -264,13 +268,27 @@ sub start_qemu() {
...
+            }
             $vars->{HDDMODEL} = "scsi-hd";
         }
         for my $i ( 1 .. $vars->{NUMDISKS} ) {
             my $boot = "";    #$i==1?",boot=on":""; # workaround bnc#696890
-            push( @params, "-drive", "file=$basedir/l$i,cache=unsafe,if=none$boot,id=hd$i" );
-            push( @params, "-device", "$vars->{HDDMODEL},drive=hd$i" . ( $vars->{HDDMODEL} =~ /ide-hd/ ? ",bus=ide.@{[$i-1]}" : '' ) );
+            if ($vars->{MULTIPATH}) {
+                for my $c ( 1 .. $vars->{PATHCNT} )
+                    # pathname is a .. d
```
coolo: you're missing a  { here - that's the real problem

PR #183, opened 2015-02-12
Support new variable, QEMURAM, so you can request a VM with more RAM tha...

sysrich:
...n 1024 - Defaults to 1024 if not set.


aaannz:
I understand the historical value of individual commits, but still I would prefer to rebase all of them to one and then force push to this branch.

coolo:
I usually just take his commits and squash them before pushing

coolo:
pushed as 80db71d

PR #185, opened 2015-02-16
allow connection to 10.0.2.2

nadvornik:


coolo:
the question how useful this is going on - with IPMI and s390x backends the SUT runs on a different IP than os-autoinst.

Possibly we just drop that code all together?

nadvornik:
Now the SUT expects that os-autoinst is on 10.0.2.2. 
With qemu user mode networking it is translated to 127.0.0.1, with other network configuration it is not translated.
So IMHO the code is OK for now.
Later we want to allow more flexible configurations with Open vSwitch.

coolo:
that's true - we'd need to change a lot more

PR #187, opened 2015-02-17
testapi for mutex locks

aaannz:
I'm using simple curl to ask for status responses, checking only 200 for success, else wait forever. I kind of don't like it but I didn't want to take heavy Mojo::UA or something similar to testapi.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1,3 +1,18 @@
+# Copyright (C) 2015 SUSE Linux Products GmbH
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -1,3 +1,18 @@
+# Copyright (C) 2015 SUSE Linux Products GmbH
```
coolo: there is no such company in 2015 :)

testapi.pm
```diff
@@ -475,6 +491,43 @@ sub wait_screen_change(&@) {
...
+}
+
+sub _mutex_call($$) {
+    my ($action, $name) = @_;
+    # blocking call to worker, worker asks Scheduler for lock
+    bmwqemu::fctlog("Trying mutex action $action on lock $name");
+    my $port = $bmwqemu::vars{'WORKER_PORT'};
+    my $path = "$action/$name";
+    while (1) {
+        my $res = qx(curl -s -o /dev/null -w \"\%{http_code}\" -I http://localhost:$port/$path);
```
coolo: and there is really no way you can make use of the HTTP interface already available in os-autoinst? like a long poll from the worker to it?

I find this whole interaction model very complex and hard to explain ;(

aaannz: What do you mean by 'from worker to it'? To os-autoinst http? I can't imagine yet how that is supposed to work for locking. This part is for direction os-autoinst -> worker (worker then acts as a proxy to scheduler).

PR #190, opened 2015-02-19
introduce CDDMODEL

lnussel:
aarch64 needs a way to use virtio cd


coolo:
this code is getting seriously out of control ;(

coolo:
and there is a D too much in your commit log

aplanas:
Clearly we need a XML definition for qemu configuration.  This XML can also have different backends (real hardware, kvm, vbx, qemu, vmware, ...)

... oh wait http://libvirt.org/

coolo:
do you know someone in the cloud team by chance? :)

PR #194, opened 2015-02-25
misc

lnussel:


coolo:
I think we're in urgent need to document the variables that os-autoinst and its backends act upon. Our tests are our thing, but os-autoinst has meanwhile many variables to control a lot ;(

PR #195, opened 2015-02-26
interactive mode fixes

lnussel:


coolo:
the question really is if the webui really should wait with the interactive features for an os-autoinst round trip. The webui knows when it enabled it after all

PR #198, opened 2015-02-27
Split results.json

coolo:
This is not to be merged before openQA worker is adapted to this, but for review and discussion I open the pull request early

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -324,18 +324,9 @@ sub enqueue_screenshot() {
     else {    # new
         $image->write($filename) || die "write $filename";
         # copy new one to shared directory, remove old one and change symlink
-        cp($filename, $bmwqemu::liveresultpath);
-        unlink($bmwqemu::liveresultpath .'/'. basename($lastscreenshotName)) if $lastscreenshot;
         $bmwqemu::screenshotQueue->enqueue($filename);
         $lastscreenshot          = $image;
         $lastscreenshotName      = $filename;
-        unless(symlink(basename($filename), $bmwqemu::liveresultpath.'/tmp.png')) {
-            # try to unlink file and try again
-            unlink($bmwqemu::liveresultpath.'/tmp.png');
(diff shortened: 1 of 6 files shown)


backend/baseclass.pm
```diff
@@ -324,18 +324,9 @@ sub enqueue_screenshot() {
...
         $bmwqemu::screenshotQueue->enqueue($filename);
         $lastscreenshot          = $image;
         $lastscreenshotName      = $filename;
-        unless(symlink(basename($filename), $bmwqemu::liveresultpath.'/tmp.png')) {
-            # try to unlink file and try again
-            unlink($bmwqemu::liveresultpath.'/tmp.png');
-            symlink(basename($filename), $bmwqemu::liveresultpath.'/tmp.png');
-        }
-        rename($bmwqemu::liveresultpath.'/tmp.png', $bmwqemu::liveresultpath.'/last.png');
-
```
aaannz: Is this WIP or how the livestream will work?

coolo: the worker will push $last to the webui together with the usual status updates. I'll may add the filename of the last screenshot to the status.json

coolo: even the last changed screenshot and the worker then can avoid uploading unchanged images.

coolo: but yes, this is WIP - as long as I didn't finish the openQA side I expect that I have to touch this.

coolo:
F**** - I played with git push options and managed to push to master ;(

coolo:
I wasn't able to reopen, so this is now #199 - after force pushing the old master to master

aaannz:
Rewrite the history, that's how winners rule :D

backend/baseclass.pm
coolo: So this won't work as the status is not written every time - so I went back to a symlink. But the symlink is in qemuscreenshots (where I actually missed it for a while when debugging ipmi :)

PR #203, opened 2015-03-09
issue 5640: only allow scripts inside of CASEDIR

coolo:
Expect relative paths and warn if absolute path is given


coolo:
btw: I stopped the fetchneedles cron job to avoid deployment problems

coolo:
this really shouldn't change behaviour, so no change to tests *required* but strongy recommended

PR #207, opened 2015-03-10
don't require Data::Dump at all and Data::Dumper only where used for die...

coolo:
...s


bmwqemu.pm
```diff
@@ -16,7 +15,6 @@ use threads::shared;
 use Thread::Queue;
 use POSIX;
 use Term::ANSIColor;
-use Data::Dump qw(dump dd);
```
aaannz: dump is still used at line 279 and 658

coolo: thanks for spotting - c39389e

PR #208, opened 2015-03-10
don't require tests to be nested at least two levels

AdamWill:
All of openSUSE's tests are in tests/foo/bar/sometest.pm, but
Fedora's are not, they are in tests/bar/sometest.pm. This regex
requires the former layout, for no apparent reason because it
doesn't actually capture anything other than 'bar' and
'sometest'.

I tested this with 'foo/bar/test.pm', 'bar/test.pm', and
'foo/monkeys/bar/test.pm' it captures 'bar' and 'test' in all
cases, so I think it's OK.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -18,7 +18,7 @@ sub loadtest($) {
         warn "loadtest needs a script below $casedir\n";
         $script = File::Spec->abs2rel( $script, $bmwqemu::vars{CASEDIR} );
     }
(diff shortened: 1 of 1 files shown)


AdamWill:
Hum, so actually we have `tests/sometest.pm` while SUSE has `tests/category/sometest.pm`. So while the Fedora layout worked before, `$category` would always be 'tests'. That's how it works with the Fedora layout after my change too.

So...arguably Fedora's layout is bad and we should add more nesting, but if it kinda-worked before, probably it still should kinda-work.

PR #209, opened 2015-03-11
fix "do_run" loop for steady screenshot stream

froh:
- refactored vnc update logic

- use just one screen capture loop, which both updates the framebuffer
  and produces screenshots at a steady pace.  use this loop also from
  key press events, to keep up with screen changes.

can't get it working with ipmi at this time, assume pebcac in setting up ipmi.

### DIFF
--- a/backend/vnc_backend.pm
+++ b/backend/vnc_backend.pm
@@ -8,36 +8,37 @@ use feature qw/say/;
 use Data::Dumper qw(Dumper);
 use Carp qw(confess cluck carp croak);
 
-sub fetch_all_pending_screenshots() {
-    my ($self, $timeout, $min_time) = @_;
-
+sub request_screen_update($ ) {
+    my ($self) = @_;
     return unless $self->{'vnc'};
+    $self->{vnc}->send_update_request();
+}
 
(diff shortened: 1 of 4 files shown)


backend/vnc_backend.pm
```diff
@@ -8,36 +8,37 @@ use feature qw/say/;
...
-    if ($@) {
-        bmwqemu::diag "VNC failed $@";
-        $self->close_pipes();
-    }
-
+sub capture_screenshot($ ) {
+    my ($self) = @_;
+    return unless $self->{'vnc'};
+    $self->{vnc}->fetch_pending_updates();
     return unless $self->{'vnc'}->_framebuffer;
```
coolo: you said you tested, but this function doesn't seem to exist:

DIE Can't locate object method "fetch_pending_updates" via package "backend::VNC" at /usr/lib/os-autoinst/backend/vnc_backend.pm line 20.

(found in production stupid me)

froh: Must be a git fetch rebase push error on my side.  will check first thing
I'm at my desk.

froh: PR #210 should fix it.  I'm sorry.

PR #212, opened 2015-03-13
vnc type_string: give max_interval a meaning

froh:
- bootloader code needs _slow_ key presses when scrolling the text
  input field.  This is in theory accomplished with the max_interval
  parameter in the test cases.  this patch brings the paramter to life
  and thus allows for speedy keyboard input in normal cases and
  throttling in the bootloader special case.
- fix the logic for the very first screenshot after connecting,  this
  takes out superflouous and unwanted screen update requests.


froh:
had to tidy this...

coolo:
merged manually (after tidy)

PR #214, opened 2015-03-13
allow download of data in subdirs

nadvornik:
https://progress.opensuse.org/issues/6468

downloading of directories as cpio archive is preserved


coolo:
is it guaranteed that this doesn't contain .. and stuff?

nadvornik:
.. in path is checked by regexp, .. in cpio archive does not pass the -f check.

PR #215, opened 2015-03-18
add a parameter to type_string to suppress logging

AdamWill:
Sometimes maybe we type stuff we don't want to log for the
world to see, so add a param which lets us not log the actual
string typed if we don't want to.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -25,7 +25,7 @@ our $serialdev;
 
 sub send_key($;$);
 sub check_screen($;$);
-sub type_string($;$);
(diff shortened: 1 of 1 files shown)


lnussel:
hmm, could we use type_password for that? Ie extend it to accept a string parameter which would be used insted of the built in password when specified?

coolo:
I actually have an alternative patch on my laptop that I discussed with adam, but I forgot to commit.

coolo:
adam, please check #219

PR #217, opened 2015-03-19
various small fixes

froh:
- main fix: drain VNC socket before requesting a new screen update
- improvements:
     - better error messages
     - cleaner log
     - run with all vncviewer clients
- various s390 specific fixes

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -267,6 +267,11 @@ sub update_line_number() {
     }
 }
 
+# pretty print like Data::Dumper but without the "VAR1 = " prefix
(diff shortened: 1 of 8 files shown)


bmwqemu.pm
```diff
@@ -267,6 +267,11 @@ sub update_line_number() {
     }
 }
 
+# pretty print like Data::Dumper but without the "VAR1 = " prefix
+sub pp {
+    Data::Dumper->new(\@_)->Terse(1)->Dump();
+}
+
```
coolo: neat!

PR #219, opened 2015-03-21
add a parameter to type_string to suppress logging

coolo:
Sometimes maybe we type stuff we don't want to log for the
world to see, so add a param which lets us not log the actual
string typed if we don't want to.

Superseding #215

### DIFF
--- a/backend/driver.pm
+++ b/backend/driver.pm
@@ -122,12 +122,6 @@ sub send_key($) {
     return $self->_send_json({ 'cmd' => "send_key", 'arguments' => { 'key' => $key } });
 }
 
-sub type_string($$) {
-    my ( $self, $text, $max_interval ) = @_;
-    return unless ($text);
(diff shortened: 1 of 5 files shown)


lnussel:
what about keeping the api of type_string and introduce an internal _type_string with the ne api that is used by both type_string and type_password?

coolo:
type_string should be able to avoid logging even for !passwords - and beside I don't like these magic numbers in the type_string calls.

coolo:
to document it: I discussed with ludwig that the extra {} are too perlish for our testapi, but perl allows

type_string 'string', secret => 1 too - but that requires some more magic in the call - but in this case it's worth it IMO

coolo:
done now:

Either one will work:
<code>type_string 'hallo', secret => 1;</code>
or
<code>type_password 'hallo';</code>

PR #220, opened 2015-03-23
remove hardcoded time limit

lnussel:
the worker will kill os-autoinst if it takes too long


lnussel:
looks like a failure on travis side

PR #229, opened 2015-04-07
Assets

nadvornik:
https://progress.opensuse.org/issues/6972

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -157,12 +157,11 @@ upload log file to openqa host
 
 sub upload_logs($) {
     my $file = shift;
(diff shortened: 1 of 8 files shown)


testapi.pm
```diff
@@ -170,6 +170,38 @@ sub ensure_installed {
...
 
+=head2 upload_asset
+
+upload log file to openqa host
+
+=cut
+
+sub upload_asset($;$) {
+    my ($file, $public) = @_;
+    my $host = "10.0.2.2:" . (get_var('QEMUPORT') + 1);
```
coolo: as we have this construct now in 3 functions - can't you use autoinst_url ?

PR #230, opened 2015-04-08
add missing dependencies and compile check test

aaannz:


aaannz:
hmm.. still not running compile checks, hold on

t/Makefile.am
```diff
@@ -1,3 +1,3 @@
-TESTS = 01-test_needle.pl 02-test_ocr.pl 03-testapi.pl
+TESTS = 00-compile-check-all.t 01-test_needle.pl 02-test_ocr.pl 03-testapi.pl
```
coolo: it's .pl in here

aaannz: Yeah, I renamed it just now and forgot about that.

aaannz:
Ok, this is taking a lot more time I expected.
The problem is that tinycv is compiled using system perl libraries instead of perlbrew ones. The -I from configure correctly points to perlbrew libs, but perl binary is used system one. That may be one culprit. Or MakeMaker creates Makefile using system sources, dunno.

@coolo any ideas?

coolo:
I fixed and merged it

PR #231, opened 2015-04-08
Local xvnc

froh:
many changes, best reviewed as a sequence using gitk...

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -31,6 +31,7 @@ sub loadtest($) {
     }
     else {
         # perl code generating perl code is overcool
+        # FIXME turn this into a proper eval instead of a generated string
         my $code = "package $name;";
         $code .= "use lib '$casedir/lib';";
         my $basename = dirname($script);
--- a/backend/VNC.pm
+++ b/backend/VNC.pm
@@ -94,7 +94,7 @@ sub login {
         PeerAddr => $hostname || 'localhost',
         PeerPort => $port     || '5900',
         Proto    => 'tcp',
-    ) || Carp::confess "Error connecting to $hostname\n". Dumper($self) ."\n$@";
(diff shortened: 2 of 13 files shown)


froh:
the first seven patches are straightforward refactorings which simplify the code.  these are the pacthes that go 'all over the place'

the last two patches primarily affect the s390 backend, with very littel and local effect on qemu and ipmi:  the effect to qemu and ipmi is to consolidate connect_vnc/relogin_vnc into one function that is then used from all three backends.

the changes to testapi.pm and baseclass.pm actually are extensions that at this point are only implemented by the s390 backend.  they should imnsho be used and implemented by the other backends as well, however at this point in time they do not affect the other backends in any way.

As they are needed for s390, I'd appreciate if you'd review an possibly integrate.

coolo:
merged into WC

coolo:
I kind of expected this to happen, so here you have the facts:

<pre>
DIE encountered object 'backend::VNC=HASH(0x7ffac000bc40)', but neither allow_blessed nor convert_blessed settings are enabled at /usr/lib/os-autoinst/backend/baseclass.pm line 458 thread 1.

 at /usr/lib/os-autoinst/backend/baseclass.pm line 46 thread 1.
	backend::baseclass::die_handler('encountered object \'backend::VNC=HASH(0x7ffac000bc40)\', but...') called at /usr/lib/perl5/5.18.2/Carp.pm line 100 thread 1
	Carp::croak('encountered object \'backend::VNC=HASH(0x7ffac000bc40)\', but...') called at /usr/lib/perl5/vendor_perl/5.18.2/JSON/backportPP.pm line 506 thread 1
</pre>

more in https://openqa.suse.de/tests/29251/file/autoinst-log.txt

coolo:
https://github.com/os-autoinst/os-autoinst/pull/235

PR #238, opened 2015-04-17
S390 worker

froh:
enable switching between multiple vnc servers:  the local Xvnc 'worker' console, on which the remote access terminal emulators are started.  potentially the remote vnc server on the system under test, once that is up and running.

This also enables the locally running terminal emulators proper, types 'ssh' and 'ssh-X' which connect to the SUT once a linux is running there, and 's3270', which connects to zVM.

I've tried this with various flavours of remote graphics:
   ssh -X, true remote X, bot use the local worker Xvnc as $DISPLAY
   vnc will access the remote Xvnc server on the SUT once that is available

it continues to work with qemu, it should also just work with ipmi.

### DIFF
--- a/backend/s390x.pm
+++ b/backend/s390x.pm
@@ -16,7 +16,7 @@ use backend::s390x::s3270;
 
 use backend::VNC;
 
-use testapi qw(get_var check_var);
+use testapi qw(get_var check_var set_var);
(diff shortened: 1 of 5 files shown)


froh:
yes.

PR #242, opened 2015-04-24
Do not force "scsi" interface for cdrom, add "ide" support

aginies:
Do not force "scsi" interface for cdrom, "ide" can be used if CDMODEL=ide-cd var is used. By default "scsi" will be used.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -345,7 +345,9 @@ sub start_qemu() {
         for my $i (1 .. 6) {    # check for up to 6 ADDON ISOs
             if ($vars->{"ISO_$i"} && $vars->{"ADDONS"}) {
                 my $addoniso = $vars->{"ISO_$i"};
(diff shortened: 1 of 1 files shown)


backend/qemu.pm
```diff
@@ -345,7 +345,9 @@ sub start_qemu() {
         for my $i (1 .. 6) {    # check for up to 6 ADDON ISOs
             if ($vars->{"ISO_$i"} && $vars->{"ADDONS"}) {
                 my $addoniso = $vars->{"ISO_$i"};
-                push(@params, "-drive", "if=scsi,id=addon_$i,file=$addoniso,media=cdrom");
+		my $cdinterface ="if=scsi";
+		if ($vars->{CDMODEL} eq "ide-cd") { $cdinterface = "if=ide"; }
```
coolo: you indented incorrectly (use ./tools/tidy pre-commit)

PR #244, opened 2015-04-24
don't use external system("mv")

coolo:


froh:
...like so:
https://github.com/froh/os-autoinst/commit/78c0ba6a16ec6f1373aa1af31b1c6309282bf0f6

coolo:
I don't think that's wise. This external python script is just temporary

PR #245, opened 2015-04-24
s390: wait_serial and linux-vt-xterm

froh:
- add wait_serial console support for s390
- add linux vt impersonation to xterm, so an ssh session from this xterm _looks_ like a physical Linux console.
  eliminates need for reneedling ssh installs.
- minor doc fixes
- enforce 16bit Xvnc on worker


coolo:
this might be fun, but I don't really want fonts and font converters as part of os-autoinst. Create an external package for this

coolo:
package that python xterm as a different package with the fonts required and then change os-autoinst to call it. I don't think we want yet another python script called from perl

PR #247, opened 2015-04-27
S390 serial console and xterm-console

froh:
now uses xterm-console package, sorry couldn't clean up the log any more.  tested.
  * add wait_serial console support for s390
  * use xterm-console, a linux vt impersonation of xterm, so an ssh session looks like a physical Linux console.  eliminates need for reneedling ssh installs.
  * minor doc fixes
  * enforce 16bit Xvnc on worker


coolo:
I rebased that a bit

PR #250, opened 2015-05-11
add precompiled regex support to wait_serial

aaannz:
- support individual precompiled regexes (qr//)
- support multiple regexes in ARRAYref


aaannz:
Hmm.. It looks like that making life easier with smartmatch is unsupported in SLES, so I'll need to rewrite ~~ to actual loop.

PR #255, opened 2015-05-18
Cleanup orphaned script processes

asdil12:
Cleanup orphaned ipmitool and script processes


coolo:
this is possibly the worst possible approach. there *must* be another way

lnussel:
we need to understand why those processes are left over in the first place. To kill them all easily it may make sense to use a separate process group for os-autoinst manged by the worker. That way the worker can kill all children of os-autoinst.

coolo:
see #258 for my suggestion. all that extra killing are just workarounds.

PR #257, opened 2015-05-18
allow commands through curl

coolo:


coolo:
this is proof of concept - to be discussed

coolo:
@ancorgs - the openQA side is actually not that hard.

okurz:
Is this still relevant?

aaannz:
I doubt that this implementation is still relevant, but the idea is. I think we should move this to progress however.

okurz:
"we should" = ?

coolo:
I'd actually do this completely different today

PR #258, opened 2015-05-19
IPMI: set the process group of the script caller so we can kill all c…

coolo:
…hilds in a go


lnussel:
I'd prefer this process group handling somewhere higher up but it might be ok as workaround for now.

PR #259, opened 2015-05-20
allow downloading assets via os-autoinst

nadvornik:
The openQA server is accessible with user mode networking. 
With TAP devices it would need additional NAT configuration. I think that assets should work without such extra requirements on the network - that means serve them via os-autoinst.

This depends on https://github.com/os-autoinst/openQA/pull/371

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -100,16 +100,6 @@ sub _test_data_file {
 
     $self->app->log->debug("Request for file $file.");
 
-    my $fd;
-    unless (open($fd, '<:raw', $file)) {
-        # ERROR HANDLING
-        $self->render(text => "Can't open $file", status => 404);
-        return;
-    }
-    local $/ = undef;    # slurp mode
-    my $data = <$fd>;
-    close($fd);
-
     my $filetype;
 
     if ($file =~ m/\.([^\.]+)$/) {
@@ -120,7 +110,7 @@ sub _test_data_file {
     $filetype ||= "application/octet-stream";
     $self->res->headers->content_type($filetype);
 
-    return $self->render(data => $data);
+    return $self->reply->asset(Mojo::Asset::File->new(path => $file));
 }
 
 sub test_data {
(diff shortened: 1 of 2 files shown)


lnussel:
hmm, the long term plan would be to get rid of nfs, that means os-autoinst would have to be a proxy to openqa

commands.pm
```diff
@@ -206,6 +216,8 @@ sub run_daemon {
     my $secret = $bmwqemu::vars{'CONNECT_PASSWORD'} || 'notsosecret';
     app->secrets([$secret]);
 
+    push @{app->static->paths}, '/';    # we check the path in each method
```
lnussel: even if that is the case now that still looks evil

nadvornik: Setting 2 paths (data, assets) could cause conflicts for files with the same name because the files are always searched in all paths.
So IHMO it is better to set common parent directory and because the paths can be arbitrary (they are set by worker), the only known common parent directory is /

It is as secure as before when the files were just read and sent.

testapi.pm
```diff
@@ -433,6 +433,29 @@ sub autoinst_url() {
...
+Special values REPO_\d and ASSET_\d points to the asset configured
+in the corresponding variable
+
+=cut
+
+sub data_url($) {
+    my ($name) = @_;
+    if ($name =~ /^REPO_\d$/) {
+        return autoinst_url . "/assets/repo/" . get_var($name);
+    }
```
coolo: I'm not so sure this is what we want in the long run. http installations are nice, but far from all installation ways we support. So where would you call that in a test?

commands.pm
coolo: but you can use serve_asset with a checked asset path, no?

testapi.pm
nadvornik: for now I want to use it for autoyast variable in bootloader.pm:
" autoyast=" . data_url(get_var("AUTOYAST"))

coolo: I get the /assets/other part but not the REPO

nadvornik: OK, I added REPO because it seemed similar, but I have no real world use case right now. I can remove it if you want.

lnussel: can we separate functions for assets and data? Ie a separate assset_url where the first parameter must be the type. this avoids guessing.

nadvornik: That's exaxtly te opposite of what I want. I need one generic function and then choose between asset and data by setting variables (without touching the code).

nadvornik:
I rewrote it without Static

coolo:
I think it's fine now

commands.pm
```diff
@@ -140,6 +130,22 @@ sub test_data {
     return $self->render_not_found;
 }
 
+sub get_asset {
+    my ($self) = @_;
+
+    my $path = join '/', $bmwqemu::vars{'ASSETDIR'}, $self->param('assettype'), $self->param('assetname');
```
lnussel: can we guarantee that assettype and assetname is not .. here? Only relpath is checked below

coolo: otherwise the get won't match. That's basic mojo

PR #260, opened 2015-05-21
generate subtitles in WebVTT format

bmwiedemann:
this happens on-the-fly during os-autoinst
but can be done later when using a log created with DEVEL_SUBTITLES=1

### DIFF
--- a/subtitle.pm
+++ b/subtitle.pm
@@ -1,0 +1,71 @@
+# © 2015 SUSE Linux GmbH
+# by Bernhard M. Wiedemann <bwiedemann suse de>
+# Licensed under GPL v2 or later
+
+package subtitle;
+use threads;
+use threads::shared;
+
+our $fps         = 24;
+our $granularity = int($fps * 0.5);
+our $ssnoffset   = 0;
+our $subopts     = " line:0 align:start";
+
+sub ssn2videotime($) {
+    my $ssn = shift;
+    my $sec = ($ssn + $ssnoffset) / $fps;
+    return sprintf("%02i:%02i.%03i", $sec / 60, $sec % 60, 1000 * ($sec - int($sec)));
+}
+
+our $subnum;
+our $subtext : shared;
+our $starttime;
+our $subfd;
+
+sub init_subtitle($) {
+    my $filename = shift;
+    open($subfd, ">", $filename) or die $!;
+    print $subfd "WEBVTT FILE\n\n";
+    $subtext   = '';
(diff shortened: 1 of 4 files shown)


subtitle.pm
```diff
@@ -0,0 +1,70 @@
...
+                print $subfd "$subnum\n$starttime --> $endtime$subopts\n$subtext\n";
+                $subnum++;
+                $subtext = '';
+            }
+            $starttime = $endtime;
+        }
+    }
+    else {
+        # filter for interesting parts
+        return unless m/send_key|type_string|assert_screen|check_screen/;
```
aaannz: jfyi, in current zdup test, I'm matching wait_serial results instead of assert_screen. I'm for wait_serial to be also interesting.

subtitle.pm
```diff
@@ -0,0 +1,70 @@
...
+sub finish_subtitle() {
+    close $subfd;
+    $subfd = undef;
+}
+
+sub add_subtitle_line($) {
+    return unless $subfd;
+    local $_ = shift;
+    if (m/wrote screenshot #(\d+)/) {
+        my $ssn = $1;
```
coolo: this is really bizzare code. Passing strings to functions who then parse integers out of it.

subtitle.pm
```diff
@@ -0,0 +1,70 @@
...
+                $subnum++;
+                $subtext = '';
+            }
+            $starttime = $endtime;
+        }
+    }
+    else {
+        # filter for interesting parts
+        return unless m/send_key|type_string|assert_screen|check_screen|wait_serial/;
+        # nicify
```
coolo: this really belongs in log_call and not parsing random strings

subtitle.pm
```diff
@@ -0,0 +1,70 @@
+# 2015 by Bernhard M. Wiedemann
+# Licensed under GPL v2 or later
```
lnussel: work time => (c) SUSE Linux GmbH or SUSE LLC. Reminds me that os-autoinst needs proper license headers in all files.

subtitle.pm
bmwiedemann: The reason for this is, that it makes development much easier if you can run the subtitle-generator on the log file without having to run iso2video and wait for KVM.

coolo: then please write an extra file with this info, but parsing the generic log file is so backward

coolo:
I can't merge as it is - and you don't seem to be interested in improving it.

PR #263, opened 2015-06-05
multimachine networking with vde or openvswitch

nadvornik:
depends on https://github.com/os-autoinst/openQA/pull/378

https://progress.opensuse.org/issues/6444
https://etherpad.nue.suse.com/p/openqa-multimachine-network

VDE support has certain limitations:
- all workers must be on single host
- there seems to be a bug slirpvde - with 4 machines in parallel it sends some packets to wrong MAC

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -279,6 +279,20 @@ sub start_qemu() {
         symlink($i, "$basedir/l$i") or die "$!\n";
     }
 
+    if ($vars->{NICTYPE} eq "vde") {
+        if (system('vde_switch', '-d', '-s', "/tmp/openqa_vde$vars->{NIC_VLAN}.ctl") == 0) {
+            if (system('slirpvde', '-d', '-s', "/tmp/openqa_vde$vars->{NIC_VLAN}.ctl") != 0) {
+                die "Can't start slirpvde -d -s /tmp/openqa_vde$vars->{NIC_VLAN}.ctl";
+            }
+        }
+        else {
+            if (!-d "/tmp/openqa_vde$vars->{NIC_VLAN}.ctl") {
+                die "Can't start vde_switch -d -s /tmp/openqa_vde$vars->{NIC_VLAN}.ctl";
+            }
+            # else vde_switch is already running
+        }
+    }
+
     pipe(my $reader, my $writer);
     my $pid = fork();
     die "fork failed" unless defined($pid);
@@ -301,6 +315,10 @@ sub start_qemu() {
         }
         elsif ($vars->{NICTYPE} eq "tap") {
             push(@params, '-netdev', "tap,id=qanet0,ifname=$vars->{TAPDEV},script=no,downscript=no");
+        }
+        elsif ($vars->{NICTYPE} eq "vde") {
+            # use different bridge for each NIC_VLAN
+            push(@params, '-netdev', "vde,id=qanet0,sock=/tmp/openqa_vde$vars->{NIC_VLAN}.ctl");
         }
         else {
             die "uknown NICTYPE $vars->{NICTYPE}\n";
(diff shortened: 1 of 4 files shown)


backend/qemu.pm
```diff
@@ -225,6 +225,21 @@ sub start_qemu() {
     if (defined($vars->{RAIDLEVEL})) {
         $vars->{NUMDISKS} = 4;
     }
+
+    if ($vars->{NICTYPE} ne "user") {
+        # ensure MAC addresses differ globally
+        # and allow MAC addresses for more than 256 workers (up to 65535)
+        my $workerid = $vars->{WORKER_ID};
+        $vars->{NICMAC} = sprintf('52:54:00:12:%02x:%02x', int($workerid / 256), $workerid % 256);
```
lnussel: I'm not sure that's a good idea. What if we want to upload the resulting hard disk image to use it for further tests? The nic mac would change all the time then. I think for single machine jobs it's fine to use one specific mac.

lnussel: ah forget about it. 'ne' vs 'eq'. I read it he wrong way around :)

backend/qemu.pm
```diff
@@ -264,6 +279,20 @@ sub start_qemu() {
         symlink($i, "$basedir/l$i") or die "$!\n";
     }
 
+    if ($vars->{NICTYPE} eq "vde") {
+        if (system('vde_switch', '-d', '-s', "/tmp/openqa_vde$vars->{NIC_VLAN}.ctl") == 0) {
```
lnussel: what if NIC_VLAN is undef?

/tmp is evil :-) Never use that.

nadvornik: Scheduler should make sure that it is not undef. If it is undef anyway, then connecting to unnumbered vde swith is IMHO a reasonable fallback. OK, I could fix the warning.

"/tmp/openqa_vde$vars->{NIC_VLAN}.ctl" is a directory. AFAIK creating directories under /tmp is safe.

lnussel: it's only safe if the program creating the directory does some checks. otherwise an attacker may pre-create it as symlink. Just don't bother. Avoid /tmp, problem solved. We could use /var/lib/openqa/pool/net or something like that.

etc/os-autoinst/set_tap_vlan.openvswitch.example
```diff
@@ -0,0 +1,14 @@
...
+# /etc/os-autoinst/set_tap_vlan <tap device> <vlan number>
+#
+# installation:
+# copy this script to /etc/os-autoinst/set_tap_vlan
+# make it executable (chmod 755)
+# configure sudo for the command below
+#
+# _openqa-worker ALL = NOPASSWD: /usr/bin/ovs-vsctl set port tap* tag=*
+
+sudo ovs-vsctl set port $1 tag=$2
```
lnussel: didn't notice that before. that is ugly, can't this be done via systemd service that is started before the openqa worker?

lnussel:
security hat still objects use of /tmp here :)

etc/os-autoinst/set_tap_vlan.openvswitch.example
nadvornik: This part depends on host network configuration, so it should be easily adjustable. Therefore I think that a script in /etc fits better.

nadvornik:
vde documentation encourages using /tmp so I suppose that vde_switch command handles it correctly. If not, then the problem is not limited to os-autoinst.

lnussel:
wouldnt be the first time upstream does stupid things

coolo:
then let's fix that in the openvswitch package

etc/os-autoinst/set_tap_vlan.openvswitch.example
coolo: but configuring things by means of editing scripts is pretty much sle11. we moved on.

lnussel:
I'll merge this to move this forward. I'll redo the vde support though. I also still don't like the way openvswitch is handled either. callling sudo from the worker is just wrong.

PR #264, opened 2015-06-12
Shorten FTP_SERVER so it works with SLE12

sysrich:


coolo:
having this hardcoded is wrong anyway

sysrich:
Yeah, it's on my list of 'things I want to refactor for s390x'

https://databasedweller.files.wordpress.com/2014/02/spongeboblist.jpg

PR #265, opened 2015-06-16
move upload assets in the engine

lnussel:


isotovideo
```diff
@@ -145,6 +145,20 @@ diag "FAIL" if $r;
...
+    for my $i (1 .. $nd) {
+        my $dir = 'assets_private';
+        my $name = $bmwqemu::vars{"STORE_HDD_$i"} || undef;
+        unless ($name) {
+            $name = $bmwqemu::vars{"PUBLISH_HDD_$i"} || undef;
+            $dir = 'assets_public';
+        }
+        next unless $name;
+        $bmwqemu::backend->do_upload_image({'hdd_num' => $i, 'name' => $name, 'dir' => $dir });
+    }
```
coolo: note that the extra '' is not needed

testapi.pm
```diff
@@ -681,6 +681,22 @@ sub console($) {
...
+        if ($status eq 'shutdown') {
+            $autotest::current_test->take_screenshot('ok');
+            return;
+        }
+        --$timeout;
+        sleep 1 if $timeout >= 0;
+    }
+    $autotest::current_test->take_screenshot('fail');
+    croak "Machine didn't shut down!";
+}
```
coolo: why croak? we already log the call and a die will do. I don't really like using die in some calls and croak in others.

testapi.pm
```diff
@@ -5,6 +5,7 @@ use Exporter;
 use strict;
 use File::Basename qw(basename);
 use Time::HiRes qw(sleep gettimeofday);
+use Carp;
```
coolo: tsss

backend/qemu.pm
```diff
@@ -141,11 +141,26 @@ sub do_upload_image() {
...
         bmwqemu::diag "preparing hdd $hdd_num for upload as $name\n";
         mkpath($img_dir);
-        unlink("$img_dir/$name");
-        symlink("../raid/l$hdd_num", "$img_dir/$name");
+        if ($format eq 'raw') {
+            die "$!\n" unless system('qemu-img', 'convert', '-O', $format, "raid/l$hdd_num", "$img_dir/$name") == 0;
+        }
+        elsif ($format eq 'qcow2') {
+            if ($bmwqemu::vars{'MAKETESTSNAPSHOTS'}) {
+                # including all snapshots is prohibitively big
```
coolo: extra ''

PR #267, opened 2015-06-17
parse junit file

nadvornik:
depends on https://github.com/os-autoinst/openQA/pull/384 and https://github.com/os-autoinst/openQA/pull/385

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -5,6 +5,7 @@ use Exporter;
 use strict;
 use File::Basename qw(basename);
 use Time::HiRes qw(sleep gettimeofday);
+use Mojo::DOM;
 
 require bmwqemu;
 
@@ -14,7 +15,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars send_key ty
   assert_and_click mouse_hide mouse_set mouse_click mouse_dclick
   type_password get_var check_var set_var become_root x11_start_program ensure_installed
   autoinst_url script_output validate_script_output eject_cd power upload_asset upload_image
-  activate_console select_console console deactivate_console data_url assert_shutdown
+  activate_console select_console console deactivate_console data_url assert_shutdown parse_junit_log
 );
 
 our %cmd;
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -704,6 +705,85 @@ sub console($) {
     }
 }
 
+=head2 parse_junit_log
+
+parse_junit_log("report.xml")
+
+Parse jUnit log file and creates extra test result from it. The file must
+be first uploaded by upload_logs().
+
```
coolo: what's the point here? Why upload and parse in one go?

coolo: it's not like we're going to parse unuploaded files

testapi.pm
```diff
@@ -704,6 +705,85 @@ sub console($) {
...
+        my $ts_category = $ts->{package};
+        $ts_category =~ s/[^A-Za-z0-9._-]/_/g;    # the name is used as part of url so we must strip special characters
+        my $ts_name = $ts_category;
+        $ts_category =~ s/\..*$//;
+        $ts_name =~ s/^[^.]*\.//;
+        $ts_name =~ s/\./_/;
+
+        push @tests,
+          {
+            "flags"    => {"important" => 1},
```
coolo: you guys have some weird relationship to quotes :)

But as a matter of fact none of the quotes is required here.

testapi.pm
```diff
@@ -704,6 +705,85 @@ sub console($) {
...
+        my $ts_result = 'ok';
+        $ts_result = 'fail' if $ts->{failures} || $ts->{errors};
+
+        my $result = {
+            "result"  => $ts_result,
+            "details" => [],
+            "dents"   => 0,
+        };
+
+        my $i = 1;
```
coolo: It's not obvious what $i is - please pick a better name. It only becomes clear that it's a counter 28 lines later

testapi.pm
```diff
@@ -704,6 +705,85 @@ sub console($) {
...
+                }
+                close $fd;
+                $details->{text} = $text_fn;
+            }
+            push @{$result->{details}}, $details;
+            $i++;
+        }
+
+        my $fn = bmwqemu::result_dir() . sprintf("/result-%s.json", $ts_name);
+        bmwqemu::save_json_file($result, $fn);
```
coolo: why use sprintf here? "/result-$ts_name.json" is quite readable IMO. And it's not as if we wouldn't use this form everywhere else.

nadvornik:
upload_logs is the easiest way to transfer the file from VM to os-autoinst where it is parsed.
Keeping the original jUnit file is IMHO a good idea anyway.

Other issues are fixed.

coolo:
Sorry, I mistyped. What I wanted to ask: why *not* upload and parse in one call. I don't see a use case of parsing *without* uploading too

coolo:
please rebase - and answer :)

nadvornik:
ok, uploading is now part of the function too.

testapi.pm
```diff
@@ -720,6 +721,84 @@ sub assert_shutdown(;$) {
...
+file is then parsed as jUnit format and extra test results are created from it.
+
+=cut
+
+sub parse_junit_log {
+    my ($file) = @_;
+
+    upload_logs($file);
+
+    $file =~ s/.*\///;
```
coolo: we use basename in this file already - and you should too because it makes quite more obvious what's going on.

PR #268, opened 2015-06-18
Move key_round to testapi and renamed it to send_key_until_needlematch

nilxam:
Move key_round() to testapi from opensusebasetest, and renamed it to send_key_if_tag_unmatch(), action#7944

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -8,7 +8,7 @@ use Time::HiRes qw(sleep gettimeofday);
 
 require bmwqemu;
 
-our @EXPORT = qw($realname $username $password $serialdev %cmd %vars send_key type_string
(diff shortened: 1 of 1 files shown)


coolo:
I don't think unmatch is a word. @sysrich ?

coolo:
and the if is wrong too - as it does it not just once but repeatedly.

nilxam:
How about send_key_until_tag_matched() ?

sysrich:
unmatch can be a word.. like undo, so it implies an 'action' note a 'state'.

send_key_until_needlematch?
send_key_until_match?
send_key_until_needle?
send_key_until?

repeat_key_until?
repeat_key?

nilxam:
updated with named send_key_until_needlematch

PR #269, opened 2015-06-22
update vde support

lnussel:


nadvornik:
Connecting slirpvde with one particular job looks strange.

So far the recommended setup for 2 workers and 1 control job is this:

A <------ C
B <------/

Control job C is started last, so the VDE_USE_SLIRP must be added to job A or B and the dependencies will have to be more complicated.

lnussel:
in your example A and B would have PARALLEL_WITH=C, right? so C would be first. so one would schedule C on a machine that has VDE_USE_SLIRP set

nadvornik:
no, it is just opposite, C has PARALLEL_WITH=A,B and is started after A and B. 

Also, after thinking about it, I don't like setting MAC from the variable in multi-machine tests. With testing multiple product in parallel there is a high risk of getting duplicate MAC addresses on the same network (on different vlans but it is ugly anyway).

PR #274, opened 2015-07-08
handle openvswitch via dbus service

nadvornik:


os-autoinst-openvswitch
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/perl
+package OVS;
+use strict;
+
+use base qw(Net::DBus::Object);
+use Net::DBus::Exporter qw(org.opensuse.os_autoinst.switch);
```
aaannz: I would prefer `org.opensuse.os_autoinst.openvswitch` service instead of plain switch.

systemd/os-autoinst-openvswitch.service
```diff
@@ -0,0 +1,17 @@
+# unit description file for os-autoinst openvswitch helper
+# start using e.g.
+# systemctl start os-autoinst-openvswitch.service
+[Unit]
+Description=os-autoinst openvswitch helper
+Wants=openvswitch.service
+After=openvswitch.service
+Before=openqa-worker.target
```
aaannz: Shouldn't you modify openqa-worker@ to Wants=os-autoinst-openvswitch.service?

aaannz:
Missing Net::DBus in dependencies.txt

How utilized you expect this service to be? I'm thinking, instead of modifying wants of openqa-worker@, what about making this service as one shot kind? Making it autostarted by systemd and when the request is handled just exit. @lnussel what do you think?

That leads me to other question, how this will work when the service is restarted? At startup you set some openflow rules, now when this service is restarted, it tries to do it once more. Will that work ok?

nadvornik:
I used generic name 'switch' because in theory there can be different implementation with another switch.

Repeated setting of openflow rules works OK, but I am not sure about autostarting.

I want to put all this to optional subpackage so modifying penqa-worker@ did not look right.

lnussel:
would it make sense to move this to openQA instead? openQA already uses dbus and we could keep os-autoinst simpler.

lnussel:
.. and it could use openqa.ini for the configuration then

nadvornik:
It belongs to os-autoinst, as long as we claim that os-autoinst is independent on the rest of openQA.

coolo:
that's a rather poor argument taking that mutexes are in openQA too

coolo:
but I'm not sure what would be "simpler" this way either - so I'm fine with the PR. Just saying that os-autoinst's independent days are long gone at least for complex cases (as this one)

nadvornik:
Here are the changes for rpm package:
https://build.opensuse.org/package/show/home:nadvornik:branches:devel:openQA/os-autoinst

It should start building as soon as this is merged.

os-autoinst-openvswitch
bmwiedemann: Could also be org.os_autoinst.openvswitch because os-autoinst can be used for other OSes

lnussel:
we need to rework this. I'll merge it for now to unblock you.
One idea I have is to outsource stuff like this to scripts like /usr/lib/os-autoinst/netsetup-$NICTYPE <up|down>. That way os-autoinst would only need to know about tap devices but not about openvswitch.

PR #276, opened 2015-07-10
misc

lnussel:


testapi.pm
```diff
@@ -59,13 +61,29 @@ sub record_soft_failure {
...
+    return undef;
+}
+
+sub match_is_workaround($) {
+    my ($tag) = @_;
+    if ($last_matched_needle) {
+        return $last_matched_needle->{needle}->has_property($tag);
+    }
+    return undef;
 }
```
lnussel: note: untested :)

testapi.pm
```diff
@@ -59,13 +61,29 @@ sub record_soft_failure {
...
+sub match_has_tag($) {
+    my ($tag) = @_;
+    if ($last_matched_needle) {
+        return $last_matched_needle->{needle}->has_tag($tag);
+    }
+    return undef;
+}
+
+sub match_is_workaround($) {
+    my ($tag) = @_;
```
nilxam: why not just hardcoded $tag = 'workaround'; in case the function named match_is_workaround?

PR #278, opened 2015-07-23
moved autoinst_url() to the backend to get different behaviour between s390x and other archs

Soulofdestiny:
please don't accept until sr#318154 was deployed

### DIFF
--- a/backend/s390x.pm
+++ b/backend/s390x.pm
@@ -18,6 +18,8 @@ use backend::VNC;
 
 use testapi qw(get_var check_var set_var);
 
+use Net::Address::IP::Local;
+
 sub new {
     my $class = shift;
     my $self = bless({class => $class}, $class);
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -484,10 +484,10 @@ returns the base URL to contact the local os-autoinst service
...
 =cut
 
-sub autoinst_url() {
-    # move to backend?
-    return "http://10.0.2.2:" . (get_var("QEMUPORT") + 1);
-}
+#This was moved to backend
+#sub autoinst_url() {
+#    return "http://10.0.2.2:" . (get_var("QEMUPORT") + 1);
+#}
```
aaannz: I don't think keeping comments is necessary

backend/s390x.pm
```diff
@@ -18,6 +18,8 @@ use backend::VNC;
 
 use testapi qw(get_var check_var set_var);
 
+use Net::Address::IP::Local;
```
aaannz: You might want to update DEPENDENCIES.txt

coolo:
did you test this actually?

backend/s390x.pm
lnussel: not properly packaged anyways. Note that this module tries to connect to a server in the internet to find out it's own address.

PR #279, opened 2015-07-24
explicitly start the bridge interface

nadvornik:
there is a race during startup:
1. wicked is started, creates TAP devices
2. openvswitch is started, connects the TAP devices to a bridge,
   creates bridge local interface br0
3. wicked assigns an IP to br0, os-autoinst-openvswitch service
   is starting at the same time

calling ifup ensures that os-autoinst-openvswitch is started
on already configured interface

### DIFF
--- a/systemd/os-autoinst-openvswitch.service
+++ b/systemd/os-autoinst-openvswitch.service
@@ -10,7 +10,9 @@ Before=openqa-worker.target
 [Service]
 Type=dbus
 BusName=org.opensuse.os_autoinst.switch
+Environment=OS_AUTOINST_USE_BRIDGE=br0
 EnvironmentFile=-/etc/sysconfig/os-autoinst-openvswitch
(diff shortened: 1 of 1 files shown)


systemd/os-autoinst-openvswitch.service
```diff
@@ -10,7 +10,9 @@ Before=openqa-worker.target
 [Service]
 Type=dbus
 BusName=org.opensuse.os_autoinst.switch
+Environment=OS_AUTOINST_USE_BRIDGE=br0
```
aaannz: Why do you add this envvar? Isn't it in sysconfig below?

nadvornik: it is a default in case the sysconfig file does not exist

aaannz: Ok, I wasn't sure about the ordering. But systemd.exec(5) states EnvironmentFile has priority so I rest my case :)

lnussel:
doesn't look like the right fix to me but well...

PR #280, opened 2015-07-24
do not die on unconfigured openvswitch

nadvornik:


backend/qemu.pm
```diff
@@ -595,6 +598,7 @@ sub start_qemu() {
         };
         if ($@) {
             print "$@\n";
+            print "WARNING: Can't switch NICVLAN number, independent tests may be running on the same network.\n\n";
```
aaannz: Is it safe to continue with warning? I thought dying here was ok, just error message was misleading.

nadvornik: So far we used TAP devices without openvswitch and it worked (with some limitation, as described in https://progress.opensuse.org/issues/6438 ). I don't want to completely break this option, so I think a warning fits better here.

PR #281, opened 2015-07-29
add super keycode for regression tests

mitiao:


coolo:
mind adding the IPMI key code too? I guess this translates to this:

0xE3	Keyboard Left GUI

mitiao:
I got the point from http://www.freebsddiary.org/APC/usb_hid_usages.php
and "Left GUI" (windows, command (apple), meta)
will add it.

PR #282, opened 2015-08-03
Type passwords a little slower to avoid issues with GNOME

sysrich:


testapi.pm
```diff
@@ -434,7 +434,7 @@ if no string is given
 sub type_password {
     my ($string) = @_;
     $string //= $password;
-    type_string $string, secret => 1;
+    type_string $string, 3, secret => 1;
```
aaannz: Are you sure this works? Mixing new and old syntax.
I think this should be like
```perl
type_string $string, max_interval => 3, secret => 1;
```

sysrich:
It works, but I've now set it to use only the new syntax, what do you think?

sysrich:
Well darn..Travis doesn't like it..I'm too tired, will look at tomorrow, unless someone wants to tell me what's gone wrong @aaannz ? :)

aaannz:
Well, you changed the default interval for type_password, so you need to update t/03-testapi.pl too.

sysrich:
We have tests?! Woaaaaaaah..

(will do tomorrow, leave this pull request in this messy state so I dont forget please)

sysrich:
Now has a saner number, and should be fine once #284 is merged

PR #288, opened 2015-08-28
add API assert_upload_logs

dzedro:
Created this for https://progress.opensuse.org/issues/7454
I'm not sure if it makes sense to add another `*upload_logs`
Maybe `upload_logs` could do this by default

http://10.100.98.90/tests/1703/modules/textinfo/steps/1/src

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -16,7 +16,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars send_key se
   type_password get_var check_var set_var become_root x11_start_program ensure_installed
   autoinst_url script_output validate_script_output eject_cd power upload_asset upload_image
   activate_console select_console console deactivate_console data_url assert_shutdown parse_junit_log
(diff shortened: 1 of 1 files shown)


nilxam:
but according to https://progress.opensuse.org/issues/7454 this change still rely on needle since this can not know whether the log upload success or fail :)

dzedro:
This change will rely on return value of curl not needle.
I assume if curl exits with zero, upload is successful. :)

coolo:
I don't like this duplicated code at all. split the common code in a 3rd function

nilxam:
without --fail argument then I'm not so sure it will return the correct error code, maybe I'm wrong.

dzedro:
Added curl -f -v argument and used assert_script_run

sle12
http://10.100.98.90/tests/1869/modules/textinfo/steps/3
opensuse
http://10.100.98.90/tests/1868/modules/textinfo/steps/3

PR #290, opened 2015-09-02
use Ubuntu Trusty Tahr as base for test tests

aaannz:
- move Net::DBus installation up to avoid conflits with other packages


aaannz:
ok, bug is in MakeMaker ( https://rt.cpan.org/Public/Bug/Display.html?id=106808 ) so I guess we just wait for new version?

aaannz:
Nothing to do here then

PR #294, opened 2015-09-11
functions for parsing ENV variables as an array

psladek:
usage e.g: 
MY_ADV_OPTIONS=opt1,opt2,opt2

.. 
if (check_var_array('MY_ADV_OPTIONS',$opt)) {
 ...do setup option $opt
}

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -16,7 +16,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars send_key se
   type_password get_var check_var set_var become_root x11_start_program ensure_installed
   autoinst_url script_output validate_script_output eject_cd power upload_asset upload_image
   activate_console select_console console deactivate_console data_url assert_shutdown parse_junit_log
-  assert_script_run assert_script_sudo match_has_tag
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -267,6 +267,21 @@ sub check_var($$) {
...
+    return $default if !@vars;
+    return \@vars;
+}
+
+sub check_var_array($$) {
+    my ($var, $val) = @_;
+    my $vars_r = get_var_array($var);
+    my %vars_h = map { $_ => 1 } @{$vars_r};
+    return 1 if (defined $vars_h{$val});
+    return 0;
```
coolo: return defined $vars_h{$val}); ?

coolo: and actually the map is useless too - you can just return grep { $_ eq $val } @$vars_r

coolo:
please squash the commits and prepare a PR to openQA - where we unfortunately keep the API documentation

testapi.pm
psladek: hm, was writting it probably too much in inspiration to the 'check_var' function, no problem to simplify it like this :)

psladek:
ok, more simplification and squash

PR #296, opened 2015-09-15
Added additional NIC devices to qemu backend

dzyuzin:
Added new variables NICTYPE_1, NICTYPE_2 (only TAP devices), TAPDEV_1, TAPDEV_2, NICMAC_1 and NICMAC_2

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -76,6 +76,9 @@ SKIPTO;name of snapshot;;Restore VM from given snapshot. Needs HDD image with sn
 TAPDEV;device name;undef;TAP device name to which virtual NIC should be connected. Usually undef so automatic matching is used
 TAPDEV_1;device name;undef;TAP device name to which the second virtual NIC should be connected. Must be set if NICTYPE_1 is set
 TAPDEV_2;device name;undef;TAP device name to which the third virtual NIC should be connected. Must be set if NICTYPE_2 is set
+TAPSCRIPT;script path;undef;script to configure TAPDEV network interface after VM is started, see qemu -net tap,script option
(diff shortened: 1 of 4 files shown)


doc/backend_vars.asciidoc
```diff
@@ -50,8 +50,12 @@ MAKETESTSNAPSHOTS;integer;0;save snapshot for each test in qcow image
 MULTINET;boolean;0;VM will have two virtual NIC. Can't be used with NICTYPE tap
 MULTIPATH;boolean;0;Add HDD drives as multipath devices. Override HDDMODEL to virtio-scsi-pci
 NICMAC;any MAC address;52:54:00:12:34:56;MAC address to be assigned to virtual network card
+NICMAC_1;any MAC address;52:54:01:12:34:56;MAC address to be assigned to virtual network card
+NICMAC_2;any MAC address;52:54:02:12:34:56;MAC address to be assigned to virtual network card
```
coolo: it's the same default MAC address?

coolo: ah, tiny difference - I spotted it :)

coolo:
run ./tools/tidy before commit

aaannz:
I must say I don't like this approach very much. I would like to see something more general like iterating through NICTYPE_%n etc. And also why limit them to TAP. We can eventually remove MULTINET (or simulate by adding multiple NICTYPEs for backwards compatibility).
I don't have particular use case in mind, just doing this seems strange to me.

Also see `git rebase -p` from upstream might help to remove the merge commit from PR.

nadvornik:
Here is an alternative implementation:
https://github.com/os-autoinst/os-autoinst/pull/319

PR #301, opened 2015-09-22
Revert "avoid calling status in the backend if not needed"

lnussel:
To be able to upload images the status must be queried before the VM
is stopped.

This reverts commit 0907920c15c2b5ec6fd51c272003747db58b0ba9.


coolo:
please make status a stub then instead of a crashing call

PR #304, opened 2015-10-06
proper shift handling with qemu

coolo:
reading the qemu vnc code, it behaves differently if the client
supports LED State pseudo encoding. So do so - and hopefully avoid
suprising caps lock (which qemu otherwise guesses at some wild heuristic)


sysrich:
LGTM

coolo:
it passed all tests with flying flags (is that english? :)

PR #305, opened 2015-10-06
fix power api

lnussel:


coolo:
be careful not to scare people with using the word "power" :)

sysrich:
Ahhhhhhhhhhhhhhhhhhhhhhhhhh powerrrrrrrrrrnooooooooioiii

PR #306, opened 2015-10-08
fix booting from cdrom

lnussel:


sysrich:
Thanks for this Ludwig

backend/qemu.pm
```diff
@@ -418,29 +420,33 @@ sub start_qemu() {
...
-        for my $i (1 .. 6) {    # check for up to 6 ADDON ISOs
-            if ($vars->{"ISO_$i"}) {
-                my $addoniso = $vars->{"ISO_$i"};
-                push(@params, "-drive", "if=scsi,id=addon_$i,file=$addoniso,media=cdrom");
-            }
+        for my $k (sort grep { /^ISO_\d+$/ } keys %$vars) {
+            my $addoniso = $vars->{$k};
+            my $i = $k;
+            $i =~ s/^ISO_//;
+            push(@params, '-drive',  "media=cdrom,if=none,id=cd$i,format=raw,file=$addoniso");
```
dzedro: what is if=none, why not e.g. if=virtio ?

lnussel: https://github.com/qemu/qemu/blob/master/docs/qdev-device-use.txt

dzedro: thank you

backend/qemu.pm
```diff
@@ -418,29 +420,33 @@ sub start_qemu() {
...
+
         if ($iso) {
             if ($vars->{USBBOOT}) {
                 push(@params, "-drive",  "if=none,id=usbstick,file=$iso,snapshot=on");
                 push(@params, "-device", "usb-ehci,id=ehci");
                 push(@params, "-device", "usb-storage,bus=ehci.0,drive=usbstick,id=devusb");
             }
-            elsif ($vars->{CDMODEL}) {
+            else {
                 push(@params, '-drive',  "media=cdrom,if=none,id=cd0,format=raw,file=$iso");
```
dzedro: I think snapshot=on is missing for UEFI

lnussel: indeed. meanwhile I found that adding snapshot=on here breaks loading cd drives again which is needed for zdup on uefi o_O.

lnussel:
ok, I gave up on zdup with UEFI. we can only test that with NET iso then

PR #307, opened 2015-10-12
make sure only one scsi controller is created

lnussel:


backend/qemu.pm
```diff
@@ -240,6 +240,18 @@ sub start_qemu() {
...
+
+    # TODO: we shouldn't use HDDMODEL resp CDMODEL to specify the
+    # controller.
+    # XXX: it is undefined if HDDMODEL and CDMODEL would use
+    # different kinds of virtio scsi
+    my $virtio_scsi_controller;
+    for my $var (qw/HDDMODEL CDMODEL/) {
+        if ($vars->{$var} =~ /virtio-scsi.*/) {
+            $virtio_scsi_controller = $vars->{$var};
+            $vars->{$var} = "scsi-cd";
```
aaannz: wouldn't this create 'scsi-cd' model even for HDD? Isn't there 'scsi-hd' for that?

lnussel: ah, you are right of course. I've been reshuffling this code for too long :)

PR #313, opened 2015-10-22
missing l when creating qemu image

dzedro:
This should maybe fix multipath on ppc64le

https://openqa.suse.de/tests/148782/file/autoinst-log.txt

log:
> running /usr/bin/qemu-img create raid/1 -f raw 10G
> qemu-system-ppc64 -drive file=raid/l1
> Could not open 'raid/l1': Invalid argument


dzedro:
ah ...

PR #314, opened 2015-10-22
USBinstall fix

dzedro:
for ISO_* can be hard coded scsi, when for some reason 'scsi-cd' could not be initialized.
Jut have to increase send_key in bootloader http://10.100.98.90/tests/2756/modules/bootloader/steps/2

USBinstall:
http://10.100.98.90/tests/2756
USBinstall+UEFI:
http://10.100.98.90/tests/2757
SDK:
http://10.100.98.90/tests/2750


lnussel:
I think the better fix is #315

PR #317, opened 2015-10-23
show in log default timeout instead of undef for assert_screen

dzedro:
It's a bit confusing if instead of default timeout 30 is in log undef

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -33,6 +33,7 @@ sub mydie;
 # shared vars
 
 our $screenshotQueue = Thread::Queue->new();
+our $default_timeout //= 30;    # assert timeout, 0 is a valid timeout
 my $prestandstillwarning : shared = 0;
 
 my @ocrrect;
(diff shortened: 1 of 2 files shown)


bmwqemu.pm
```diff
@@ -663,7 +663,7 @@ sub save_needle_template($$$) {
 sub assert_screen {
     my %args         = @_;
     my $mustmatch    = $args{'mustmatch'};
-    my $timeout      = $args{'timeout'} // 30;    # 0 is a valid timeout
+    my $timeout      = $args{'timeout'};    # 0 is a valid timeout
```
coolo: if you remove that here, make sure check_screen has a default too

dzedro: true, updated

aaannz:
There is still `assert_and_click` to add default timeout. And pray we won't want to change default timeout once more. (Or you can come up with some other elegant way to solve 'undef' issue :) )

dzedro:
This should be good now :tada

testapi.pm
```diff
@@ -59,13 +59,15 @@ sub record_soft_failure {
 }
 
 sub assert_screen($;$) {
-    my ($mustmatch, $timeout) = @_;
+    my $mustmatch = shift;
+    my $timeout = shift || $bmwqemu::default_timeout;
```
aaannz: Well, the deleted '0 is a valid timeout' still holds, so you should use `//` instead of `||`. Here and everywhere else where 0 is a valid value.

dzedro: true, I changed assert timeout to `//` and wait_idle timeout to `||`

PR #318, opened 2015-10-24
don't hardcode libexecdir location in cv.pm

AdamWill:
cv.pm assumes that if it's installed it will be in
'/usr/lib/os-autoinst', but in fact it's installed to
pkglibexecdir, and that's not always /usr/lib - on Fedora it's
/usr/libexec. So on installation, substitute this path to
$(pkglibexecdir). This is done the same way as an existing
substitution in isotovideo; I don't really like this approach
(using a post-install hook) but I went with it for consistency
and minimal change.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,7 +57,8 @@ pkglibexec_SCRIPTS = \
 bin_SCRIPTS = isotovideo
 
 install-exec-hook:
-	sed -i -e "s,\$$installprefix = undef;,\$$installprefix = '$(pkglibexecdir)';," $(DESTDIR)$(bindir)/isotovideo
+	sed -i -e "s,\$$installprefix = undef;,\$$installprefix = '$(pkglibexecdir)';," $(DESTDIR)$(bindir)/isotovideo && \
+	sed -i -e "s,\$$sysdir = undef;,\$$sysdir = '$(pkglibexecdir)';," $(DESTDIR)$(pkglibexecdir)/cv.pm
 
 pkglibexec_FOLDERS = \
 	distri \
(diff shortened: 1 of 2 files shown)


aaannz:
You're right. Since we are already doing it, LGTM. Any other opinion @lnussel ?

coolo:
the only thing I'd like to see changed is that we seding on a variable assignment (technically perl can have constants too, but that's somewhat artificial) and not on some random if.

coolo:
[in case to sed is a verb]

AdamWill:
sure, I can twiddle that in a bit, we're in release crunch mode ATM, so might be next week :)

coolo:
you will delay anyway :)

AdamWill:
this giant bag of crack says otherwise!

coolo:
We don't call Richard that anymore :)

coolo:
but my goldmaster is in final validation too ;)

https://openqa.opensuse.org/tests/overview?distri=opensuse&version=42.1&build=0265&groupid=7

AdamWill:
OK, how's that?

coolo:
but the idea is good - perhaps use a magic string instead

AdamWill:
well, again, I was following the example of the existing case, where the 'default' value is also 'undef'. I don't mind changing it to anything else at all, though.

coolo:
the default is not the problem - the eq call is causing the warning. So if you want to stick with that default, you need to if ($sysdir && $sysdir eq $libdir)

AdamWill:
Let's go with that, then, as I kinda like how it looks.

PR #319, opened 2015-11-04
support for multiple networks in one job

nadvornik:
- allow comma-separated values for NICMAC, NICVLAN and TAPDEV variables

open issues:
- the autogenerated numbers for TAP devices could be improved
- vde support is not yet done, with N workers x M interfaces it is easy to exceed the number of ports
- do we still need the old variable MULTINET?


coolo:
we really need to start looking into alternatives to this variable mess ;(

PR #321, opened 2015-11-10
also record matched needles that are not the best match

lnussel:


coolo:
this way you still miss needles - the first time they are best and then the 3rd needle will replace $best, but $secondbest is lost between the chairs

lnussel:
I think L138 handles that

coolo:
Yeah. But I wonder if we can't get some sorting into the needles with the same level of matching - or we might keep 2 needles created in parallel forever

coolo:
see https://github.com/os-autoinst/os-autoinst/pull/322

PR #324, opened 2015-11-16
make sure we pass perlcritic --gentle

coolo:
This is mainly about overly using subroutine prototypes - and seeing just
how inconsistent we use them, it's perfectly fine to get rid of them


aaannz:
cool! LGTM

PR #326, opened 2015-11-17
(re)add explicit return statements to functions

coolo:
I wasn't actually aware that perl does implicit returns - you never stop
learning ;)


coolo:
the key here is the return in assert_screen - which would return 1 on failure otherwise

coolo:
(and that 1 is a random value from the previous debug calls!)

PR #327, opened 2015-11-18
Added CONNECT_PASSWORD to autoinst_url, script_output and upload_logs…

dzyuzin:
… for HA tests

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -198,7 +198,6 @@ sub upload_logs {
     else {
         type_string(autoinst_url() . "/uploadlog/$basename\n");
     }
-    type_string(autoinst_url() . "/uploadlog/$basename\n");
     wait_idle();
     return;
 }
--- a/testapi.pm
+++ b/testapi.pm
@@ -543,6 +543,7 @@ returns the base URL to contact the local os-autoinst service
 sub autoinst_url() {
     # move to backend?
     if (get_var("CONNECT_PASSWORD")) {    #CONNECT_PASSWORD is used in HACLUSTER
+        use Net::Domain;
         my $fqdn = Net::Domain::hostfqdn();
         return "http://$fqdn:" . (get_var("QEMUPORT") + 1);
     }
--- a/testapi.pm
+++ b/testapi.pm
@@ -192,6 +192,12 @@ sub upload_logs {
     bmwqemu::log_call('upload_logs', file => $file);
     type_string("curl --form upload=\@$file ");
     my $basename = basename($file);
+    if (get_var("CONNECT_PASSWORD")) {
+        type_string("'" . autoinst_url() . "/uploadlog/$basename?connect_password=" . get_var("CONNECT_PASSWORD") . "'\n");
+    }
+    else {
+        type_string(autoinst_url() . "/uploadlog/$basename\n");
+    }
     type_string(autoinst_url() . "/uploadlog/$basename\n");
     wait_idle();
     return;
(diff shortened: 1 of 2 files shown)


dzyuzin:
http://ix64sap1012.qa.suse.de/tests/overview?build=00045&groupid=1&distri=sle&version=12-SP1

testapi.pm
```diff
@@ -536,7 +541,14 @@ returns the base URL to contact the local os-autoinst service
 
 sub autoinst_url() {
     # move to backend?
-    return "http://10.0.2.2:" . (get_var("QEMUPORT") + 1);
+    if (get_var("CONNECT_PASSWORD")) {    #CONNECT_PASSWORD is used in HACLUSTER
+        use Net::Domain;
+        my $fqdn = Net::Domain::hostfqdn();
+        return "http://$fqdn:" . (get_var("QEMUPORT") + 1);
```
coolo: this might work sometimes, but it's not good enough

coolo: well, scratch that.

testapi.pm
```diff
@@ -588,7 +600,12 @@ sub script_output($;$) {
     $wait ||= 10;
 
     my $suffix = _random_string;
-    type_string "curl -f -v " . autoinst_url . "/current_script > /tmp/script$suffix.sh && echo \"curl-\$?\" > /dev/$serialdev\n";
+    if (get_var("CONNECT_PASSWORD")) {
+        type_string "curl -f -v '" . autoinst_url . "/current_script?connect_password=" . get_var("CONNECT_PASSWORD") . "' > /tmp/script$suffix.sh && echo \"curl-$?\" > /dev/$serialdev\n";
```
coolo: I don't like this code duplication and I think autoinst_url can be extended to do that for you to begin with. You will need to pass a path then

testapi.pm
```diff
@@ -536,7 +541,14 @@ returns the base URL to contact the local os-autoinst service
 
 sub autoinst_url() {
     # move to backend?
-    return "http://10.0.2.2:" . (get_var("QEMUPORT") + 1);
+    if (get_var("CONNECT_PASSWORD")) {    #CONNECT_PASSWORD is used in HACLUSTER
+        use Net::Domain;
+        my $fqdn = Net::Domain::hostfqdn();
```
coolo: Net::Domain was first released with perl v5.7.3 - that's good enough not to put in DEPENDENCIES :)

I wonder if it shouldn't be possible to always use fqdn instead of the rather unknown 10.0.2.2

dzyuzin: It will not work if names resolution is broken (in HA tests I have a support server with DNS, so it always works).

coolo: @lnussel - do the KVM instances have a way to resolve openqa.opensuse.org properly?

lnussel: I think this code doesn't talk to openqa but to os-autoinst. 10.0.2.2 is correct there to reach the host from within kvm with user mode networking. It should just be moved to the backend as it's very qemu user mode specific. when using tap devices os-autoinst should probably use the ip address(es) it has on the tap interface where qemu gets connected.
Use of any fqdn guesses is most likely broken by design. don't do that please.

coolo: I meant openqa.opensuse.org the worker host not the webui. I wonder if it's really necessary to require 10.0.2.2

lnussel: there is not guarantee that a worker host can be resolved or that the address resolved from the host name is the right one. enumerating all ip addresses of the system would be an alternative.

coolo: enumerating as in typing a while loop ? Perhaps we can make once a reachable test - and remember the proper IP.

coolo:
#334 supersedes this one

PR #328, opened 2015-11-19
fix broken tab settings

lnussel:


coolo:
is there no bashtidy? :)

PR #312, opened 2015-10-22
improved lock and mm api

nadvornik:
depends on https://github.com/os-autoinst/openQA/pull/449

PR #329, opened 2015-11-20
improved lock and mm api

aaannz:
this is rebased #312 since I can't write to Vladimir's repo.


coolo:
We need to have updated documentation - I created https://progress.opensuse.org/issues/9628 for that

PR #331, opened 2015-11-21
do not add needle's name to tags

nilxam:


nilxam:
found in https://openqa.opensuse.org/tests/101287/modules/first_boot/steps/6/edit

nilxam:
and I'd suggest we should merge and deploy this fix asap.

coolo:
we live with it for quite some time, so no hurry to deploy :)

PR #332, opened 2015-11-21
Splitting the s390 backend into top level consoles

coolo:
This is still not finished as the consoles use things out of "PARMFILE", which shouldn't even exist.

But for that we need to change the tests to create the parm file itself - and pass host name to the console instead of relying on the backend to know where to connect ssh to.

The IPMI backend is changed too to use a local xterm too. I would like to see this merged and deployed/debugged in production.


coolo:
And while I hacked the ipmi and s390 backend, I found (and fixed) several bugs in the VNC implementation ;)

okurz:
When removing the vbox backend, the documentation should be adjusted, too. See https://github.com/os-autoinst/os-autoinst/blob/master/INSTALL.asciidoc

coolo:
I would remove the complete documentation - our documentation lives in openQA repo

coolo:
thanks for your reviews - I will cleanup the documentation in a followup PR. there is more to remove ;(

Issue #200, opened 2015-03-02
Is http://r.opensu.se/devel:openQA/r useful now?


choldrim:
As the title: Is http://r.opensu.se/devel:openQA/r useful now in openSUSE setup?
https://github.com/os-autoinst/os-autoinst/blob/master/INSTALL.asciidoc

coolo:
I guess it depends on your distro. on 13.1 it works, on 13.2 and factory it doesn't - most likely because l0ki no longer maintains the site ;(

http://r.opensu.se/devel:openQA/openSUSE_Factory/r works though

PR #333, opened 2015-11-23
remove outdated documentation

coolo:
A lot of this is either outdated or duplicated with openQA documentation (or both :)


doc/needles.txt
```diff
@@ -1,90 +0,0 @@
-Some packages are required to run crop.py. So for the first time do:
```
okurz: IMHO this file should not be removed as is. Certainly, "openqa" references should be removed but it also describes assert_screen and such which are actually defined in os-autoinst (not openqa). Other files are fine to cleanup.

okurz:
this PR would resolve issue https://github.com/os-autoinst/os-autoinst/issues/200

doc/needles.txt
coolo: The documentation *is* in the openqa repo - which part is in which repository is of no relevance to me. Perhaps we should add a pointer to the README though

PR #334, opened 2015-11-24
Connect with token

coolo:
superseding #327

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -177,6 +177,8 @@ sub init {
     $vars{VNC}           ||= 90;
     $vars{INSTLANG}      ||= "en_US";
     $vars{IDLETHRESHOLD} ||= 18;
+    # openQA already sets a random string we can reuse
+    $vars{JOBTOKEN} ||= random_string(10);
 
     # FIXME: does not belong here
     if (defined($vars{DISTRI}) && $vars{DISTRI} eq 'archlinux') {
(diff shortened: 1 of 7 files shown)


coolo:
I wonder if I should allow GETs without token. tons of tests fetch random data - and it's in public git anyway :)

coolo:
Ludwig had a great idea to avoid changing tests - we now require the token to be part of the path instead of a parameter. E.g. http://127.0.0.1:15223/CmUXXFhCJJ/data/ instead of http://127.0.0.1:15223/data?connect_token=CmUXXFhCJJ

This way autoinst_url() is still the proper URL to append paths to

bmwqemu.pm
```diff
@@ -975,6 +977,16 @@ sub _scale_timeout {
     return $timeout * ($vars{TIMEOUT_SCALE} // 1);
 }
 
+# just a random string useful for pseudo security or temporary files
+sub random_string {
+    my ($count) = @_;
+    $count //= 4;
```
aaannz: just nitpick, but if $count is 0, this will leave it untouched. Do you really want random_string to produce empty string (still random though :) )?
I would prefer ||=

aaannz:
Clever. One note, else LGTM

bmwqemu.pm
coolo: if someone really calls random_string(0) - then yes, he will get an empty string

aaannz: yeah, I was thinking more in line random_string($value_from_outer_space). to ensure we will not compromise something by reusing some other var. But I don't oppose strongly. :)

coolo: don't use untrusted variables to pass to random variables

PR #338, opened 2015-11-25
minor enhancements, cleanup, "autodie" to guard "system" calls

okurz:


ocr.pm
```diff
@@ -47,13 +49,12 @@ sub tesseract {
...
-    if (system('tesseract', $imgfn, $txtfn) == 0) {
-        $txtfn .= '.txt';
-        if (open(my $fh, '<:encoding(UTF-8)', $txtfn)) {
-            local $/;
-            $txt = <$fh>;
-            close $fh;
-        }
+    system('tesseract', $imgfn, $txtfn);
+    $txtfn .= '.txt';
+    if (open(my $fh, '<:encoding(UTF-8)', $txtfn)) {
```
aaannz: AFAIK if using autodie(:all) you don't need to check for open too. Unless you wanted only autodie(:system) which I think you wanted.

coolo:
Now that the tests pass - does it still work?

ocr.pm
okurz: Yes, that should be improved. My intention was to really use autodie(:all) and hide all "defensive programming" error checks so I will update the handling of open, too. Thanks for catching this.

okurz:
@coolo please define "still works". I expect some tests could fail in weird cases when previously an erroneous call to 'system' would have silently be skipped causing problems later on. *if* this happens, we should handle this properly.
Still, don't merge, will handle the "open" calls properly, too (see comment by @aaannz )

coolo:
did you run a test with this change or not?

coolo:
you forgot to put the new dependency in cpanfile

coolo:
merged manually and fixed the cpanfile

okurz:
you were too fast. see "Still, don't merge, will handle the "open" calls properly, too". Well, next time I will delete the pull-request earlier. For this change here, I will propose an upcoming change, then

coolo:
you already created a merge conflict, so I didn't want to keep it hanging longer ;)

okurz:
ok, follow up PR is out. In general, I don't have problems to rebase my changes and re-apply a PR which is a workflow I actually prefer so you can keep me dangling anytime in the future ;-)

PR #342, opened 2015-11-27
Also guard exec, open, ... calls with autodie

okurz:
This should have been part of
https://github.com/os-autoinst/os-autoinst/pull/338
before "too fast" merging :-)

Tested locally with a x86_64 test on a qcow image as well as a s390 worker at least until the
first needle matching.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -10,6 +10,8 @@ use File::Basename;
 use Time::HiRes qw(gettimeofday);
 use bmwqemu;
 use IO::Select;
+require IPC::System::Simple;
+use autodie qw(:all);
 
 use Data::Dumper;
 use feature qw(say);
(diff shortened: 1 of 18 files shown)


backend/baseclass.pm
```diff
@@ -274,13 +277,9 @@ sub unlink_crash_file {
...
 
 sub write_crash_file {
-    if (open(my $fh, ">", $iscrashedfile)) {
-        print $fh "crashed\n";
-        close $fh;
-    }
-    else {
-        warn "cannot write '$iscrashedfile'";
-    }
+    (open(my $fh, ">", $iscrashedfile));
```
coolo: what is the magic of these extra () around the call?

okurz: It's not magic but stupid leftover of deleting the surrounding 'if'. I will correct this. anything else?

coolo: no other stupid leftovers I can spot

PR #345, opened 2015-11-27
Remove redundant pair of parentheses

okurz:


sysrich:
:+1:

coolo:
sorry for the Mergatio praecox

coolo:
What's interesting: perlcritic has no warning about *that* problem ;)

```
Subroutine "write_crash_file" does not end with "return" at line 279, column 1.  See page 197 of PBP.  (Severity: 4)
Builtin function called with parentheses at line 280, column 6.  See page 13 of PBP.  (Severity: 1)
Useless interpolation of literal string at line 280, column 19.  See page 51 of PBP.  (Severity: 1)
File handle for "print" or "printf" is not braced at line 281, column 5.  See page 217 of PBP.  (Severity: 1)
Return value of flagged function ignored - print at line 281, column 5.  See pages 208,278 of PBP.  (Severity: 1)
```

okurz:
@coolo you wanna go "perlcritic --brutal"? Let's see how far we can get :-)

coolo:
I don't, I just wondered if perlcritic would have found it

PR #346, opened 2015-11-29
new console API

coolo:
the matching branch for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/843

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,7 +6,7 @@ doc_DATA = \
 	INSTALL.asciidoc \
 	COPYING \
 	doc/basetest.html \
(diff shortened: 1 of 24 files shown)


coolo:
the functionality is there, but it needs some cleanup - and it's only tested (and sure to break anywhere else) on qemu

coolo:
it doesn't have to be syntactical correct unless I have it available when I need to debug something

coolo:
BTW: I have nothing against proper logging - with some switch to enable debugging in the code. Usually I only enable the debug lines of the place that I consider breaking. With VNC you have *too much* debug overwise.

PR #349, opened 2015-12-01
Another refactoring of the consoles

coolo:
This is providing a very basic backend, to be filled


consoles/console.pm
```diff
@@ -34,6 +34,13 @@ sub screen {
     return;
 }
 
+# convenince function
```
okurz: s/convenince/convenience/, let's call it "helper" :-)

consoles/localXvnc.pm
```diff
@@ -7,42 +7,50 @@ use IPC::Run ();
...
+    my $display = ":$port";
+    my $pid     = fork();
+    die unless defined $pid;
+    if (!$pid) {
+        listen($s, 1);
+        my $peer;
+        accept($peer, $s);
+        close($s);
+        open(STDIN,  "<&", $peer) || die "can't dup client to stdin";
+        open(STDOUT, ">&", $peer) || die "can't dup client to stdout";
```
okurz: the "open" are handled by autodie, same as exec in line 36

Soulofdestiny:
Travis checks failed because ssh.pm was deleted - delete it also from Makefile.am

coolo:
cool that we have test cases, nah? :)

consoles/localXvnc.pm
coolo: I copied tutorial code ;)

consoles/localXvnc.pm
```diff
@@ -7,42 +7,50 @@ use IPC::Run ();
...
-    my $display_id = get_var("VNC") || die "VNC unset in vars.json.";
-    my $display = ":" . $display_id;
-    # FIXME: do the full monty xauth authentication, with a local
-    # XAUTHORITY=./XAuthority file
+    # start Xvnc on a random high port and use that port also as $DISPLAY
+
+    my $tcpproto = getprotobyname('tcp');
+    my $s;
+    socket($s, PF_INET, SOCK_STREAM, $tcpproto) || die "socket: $!\n";
+    bind($s, sockaddr_in(0, INADDR_ANY)) || die "bind: $!\n";
```
aaannz: bind is already covered by autodie too

backend/svirt.pm
```diff
@@ -0,0 +1,49 @@
+package backend::svirt;
+use strict;
+use base ('backend::baseclass');
+
+# this is a fake backend to some extend. We do don't start VMs, but provide ssh access
```
aaannz: typo in comment. Is the meaning 'we do not start ...' right?

consoles/localXvnc.pm
coolo: and socket is not?

backend/svirt.pm
```diff
@@ -0,0 +1,49 @@
...
+}
+
+sub do_savevm {
+    my ($self, $args) = @_;
+    return {};
+}
+
+sub do_loadvm {
+    my ($self, $args) = @_;
+    die "if you need loadvm, you're screwed with IPMI";
```
aaannz: Is this copy&paste or is IPMI really related to this?

consoles/localXvnc.pm
aaannz: Doesn't seem so ( http://perldoc.perl.org/autodie.html )

backend/svirt.pm
coolo: it's still true, no? :)

coolo:
I merged the test code to avoid @Soulofdestiny and @okurz having to base their changes on a moving branch. But I expect fine tunings to be necessary

okurz:
very sweet of you but you don't need to look after us regarding git, we can handle moving targets as long as you keep us informed what to do and where, it's fine :-)

coolo:
I removed the WIP from the summary - just to avoid any further confusions ;)

coolo:
the svirt backend is still not feature complete, but the wait_serial refactoring is important for s390 too

PR #350, opened 2015-12-02
connect select_user_console from distribution

nilxam:


coolo:
I don't like this approach. I will revert the change in the tests for now

Issue #339, opened 2015-11-26
"doc/backend_vars.asciidoc" is changed by tests


okurz:
**problem**: files tracked by the VCS should not be overwritten or changed by tests or tools if not explicitly called by the user. This is confusing users and means that the change needs to be ignored when commiting changes all the time.

**suggestion**: If the changes are needed for testing, changes should be a on non-tracked temporary copy of the file, at best in a "build" directory or at least "t".

aaannz:
Idea behind modifying this was to help (and later possibly force) contributors to properly document changes. So when user calls check_vars_docu, it will add missing entries and user then just needs to fill description, default value, etc. Travis then checks for missing entries/details and warns about.

okurz:
@aaannz ok, I get it. shouldn't "make check" then fail locally?
Also, for me "backend_vars.asciidoc" **is** changed even though I don't remove/add any variables. What am I doing wrong? Does this also happen to you/others?

aaannz:
It's probably poorly written, so it will overwrite the file even when nothing was changed. But the actual content should be the same, unless there is some indentation changes. I'll take a look.

okurz:
this is the change I get if I call just make check locally:

```
diff --git a/doc/backend_vars.asciidoc b/doc/backend_vars.asciidoc
index cce1e92..aebdc53 100644
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -1,14 +1,6 @@
 Supported variables per backend
 -------------------------------
 
-.Common
-[grid="rows",format="csv"]
-[options="header",cols="^m,^m,^m,v",separator=";"]
-|====================
-Variable;Values allowed;Default value;Explanation
-NOVIDEO;boolean;;don't encode video if set
-|====================
-
 .IPMI backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
@@ -19,31 +11,23 @@ IPMI_PASSWORD;;;
 IPMI_USER;;;
 |====================
 
-.KVM2USB backend
-[grid="rows",format="csv"]
-[options="header",cols="^m,^m,^m,v",separator=";"]
-|====================
-Variable;Values allowed;Default value;Explanation
-HWSLOT;;;
-ISO;;;
-|====================
-
 .QEMU backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
[...]

PR #351, opened 2015-12-03
Add 'get_required_var'

okurz:


doc/backend_vars.asciidoc
```diff
@@ -19,31 +11,23 @@ IPMI_PASSWORD;;;
...
-|====================
-
 .QEMU backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
 Variable;Values allowed;Default value;Explanation
-ADDONS;boolean;0;Enables use of ISO_$i variables
+$k;;;
+$var;;;
```
aaannz: Yeah, this is not really useful. I'm taking notes, will update the script later. ( note to #339 )

testapi.pm
```diff
@@ -298,6 +298,19 @@ sub get_var {
...
+
+  get_required_var($variable)
+
+Like get_var but without default. Fails when variable can not be retrieved.
+
+=cut
+
+sub get_required_var {
+    my ($var) = @_;
+    return $bmwqemu::vars{$var} // die "Could not retrieve required var $var";
```
coolo: how about using real words in such statements instead of var?

doc/backend_vars.asciidoc
okurz: sorry, shouldn't be part of this commit but thanks for noticing :-)

testapi.pm
okurz: I just assumed you'd prefer "variable" so I changed it

coolo:
please fix the merge requests

okurz:
@coolo huh?

coolo:
"This branch has conflicts that must be resolved"

PR #354, opened 2015-12-03
Remove the time warp code

sysrich:


coolo:
I checked and the penguins don't really get in the way. so we need an extra CHRISTMAS needle set, but that's it :)

PR #356, opened 2015-12-03
fix 2nd VNC connection typing

coolo:
by not overwriting the same hash with different data.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -595,46 +595,50 @@ sub init_x11_keymap {
     my ($self) = @_;
 
     return if $self->keymap;
-    $self->keymap($keymap_x11);
+    # create a deep copy - we want to reuse it in other instances
+    my %keymap = %$keymap_x11;
+
     for my $key (30 .. 255) {
-        $self->keymap->{chr($key)} ||= $key;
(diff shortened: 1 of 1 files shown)


consoles/VNC.pm
```diff
@@ -595,46 +595,50 @@ sub init_x11_keymap {
     my ($self) = @_;
 
     return if $self->keymap;
-    $self->keymap($keymap_x11);
+    # create a deep copy - we want to reuse it in other instances
+    my %keymap = %$keymap_x11;
```
aaannz: Isn't this only regular `HASH` copy? By deep copy I usually understand copying `HASH`es, `ARRAY`s, etc.. nested in other `HASH` and so on which usually needs some more perl magic. I know that in this case it is the same as %keymap is flat. Just for future reference and if there is some neat trick I was ignoring.

coolo:
you can copy a hash by values or by reference. What it did before was copying by reference - and now it's creating a new hash with the values of the old one -> deep copy.

If the HASH is nested, its values will contain references - and you have the same problem again (as you say). There you indeed need tricks in perl unfortunately. My favorite one is decoding encoded json ;)

But the keymap is just a flat hash, so we're fine.

PR #357, opened 2015-12-03
better UEFI firmware handling: check existence, Fedora support

AdamWill:
There's a couple of problems with how UEFI firmware is handled
at present. If you pass the firmware location as UEFI_BIOS, the
check that the file actually exists gets skipped, because of
how things are ordered. Also, it will try and find a UEFI
firmware for you on x86_64 if BIOS and UEFI_BIOS aren't set,
but it only knows one possible location, the one in the package
for openSUSE. The packages from https://kraxel.org, commonly
used on Fedora, place a UEFI firmware somewhere else; this adds
support for that location (and makes it easy to support other
known locations as we find them). Finally, I made it possible
to pass an absolute path rather than one relative to
/usr/share/qemu, if you like. This also puts almost all the
handling in one place, leaving only the actual addition of the
qemu param to happen later (after the @params array is defined).

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -216,8 +216,34 @@ sub start_qemu {
         die "no Qemu/KVM found\n" unless $qemubin;
     }
 
-    if ($vars->{BIOS} && !-e '/usr/share/qemu/' . $vars->{BIOS}) {
+    if ($vars->{UEFI}) {
+        # XXX: compat with old deployment
+        $vars->{BIOS} //= $vars->{UEFI_BIOS};
+    }
+
+    if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
+        # Non-absolute paths are assumed relative to /usr/share/qemu
+        $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS};
+    }
+
+    if ($vars->{BIOS} && !-e $vars->{BIOS}) {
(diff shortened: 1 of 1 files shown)


backend/qemu.pm
```diff
@@ -216,10 +216,36 @@ sub start_qemu {
...
-    if ($vars->{BIOS} && !-e '/usr/share/qemu/' . $vars->{BIOS}) {
+    if ($vars->{UEFI}) {
+        # XXX: compat with old deployment
+        $vars->{BIOS} //= $vars->{UEFI_BIOS};
+    }
+
+    if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
+        # Non-absolute paths are assumed relative to /usr/share/qemu
+        $vars->{BIOS} == '/usr/share/qemu/' . $vars->{BIOS};
+    }
```
coolo: == ? unless this is yet another perl trick this must be =

coolo:
otherwise fine with me, but I'm far from an UEFI expert

nilxam:
Looks fine to me. It's default OVMF firmware(without specific key, if pure_uefi does this meaning) generated from EDK2 source, we're renamed them in OVMF firmwares.

backend/qemu.pm
AdamWill: yes, sorry, I had some kinda virus yesterday - did this in another place too, but caught that one as I hit it in testing...I didn't test this bit.

PR #361, opened 2015-12-04
make sure that junit testsuite name is unique

nadvornik:
some slenkins tests use the same name for all testsuites, only numeric id is changing


coolo:
I bet at some point you will have to bite the bullet and write unit tests for your testsuite parser :)

PR #364, opened 2015-12-07
Allow to implement openQA test modules in python

bmwiedemann:
Allow to implement openQA test modules in python
    
    because some people prefer python over perl.
    The implementation uses the Inline::Python perl module

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -18,8 +18,8 @@ sub loadtest {
         warn "loadtest needs a script below $casedir - $script is not\n";
         $script = File::Spec->abs2rel($script, $bmwqemu::vars{CASEDIR});
     }
-    unless ($script =~ m,(\w+)/([^/]+)\.pm$,) {
-        die "loadtest needs a script to match \\w+/[^/]+.pm\n";
+    unless ($script =~ m,(\w+)/([^/]+)\.p[my]$,) {
+        die "loadtest needs a script to match \\w+/[^/]+.p[my]\n";
     }
     my $category = $1;
     my $name     = $2;
@@ -36,7 +36,19 @@ sub loadtest {
         $code .= "use lib '$casedir/lib';";
         my $basename = dirname($script);
         $code .= "use lib '$casedir/$basename';";
-        $code .= "require '$casedir/$script';";
+        if ($script =~ m/\.pm$/) {
+            $code .= "require '$casedir/$script';";
+        }
+        elsif ($script =~ m/\.py$/) {
+            $code .= "
+                use base 'basetest';
+                use Inline Python => <<'END_OF_PYTHON_CODE';\n";
+            $code .= `cat $casedir/$script`;
+            $code .= "\nEND_OF_PYTHON_CODE\n";
+        }
+        else {
+            die "impossible codepath";
+        }
         eval $code;    ## no critic
         if ($@) {
             my $msg = "error on $script: $@";
--- a/doc/example-bootloader.py
+++ b/doc/example-bootloader.py
@@ -1,4 +1,5 @@
 import perl
+perl.use("registration")
 
 def run(self):
     print "This python code ran"
@@ -10,6 +11,7 @@ def run(self):
         perl.send_key(2 + perl.get_var("NUMDISKS"))
     perl.assert_screen("inst-bootmenu", 15)
     perl.send_key_until_needlematch('inst-oninstallation', 'down', 10, 5)
+    perl.registration.registration_bootloader_params()
     perl.type_string("qwertyuiopasdfghjkl")
     print "TODO: more code"
     perl.send_key('ret')
--- a/doc/example-bootloader.py
+++ b/doc/example-bootloader.py
@@ -1,0 +1,19 @@
+import perl
+
+def run(self):
+    print "This python code ran"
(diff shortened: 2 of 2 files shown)


okurz:
+1

IMHO you should write future proof python, i.e. "print()" instead of "print …". The other parts look right.

I especially like the "perl calls python which uses perl"-part :-)

aaannz:
I think you'll need to update cpanfile to get all the dependencies.
But, contrary to @okurz I personally don't like this approach. :) And I would be highly annoyed to write `perl.` prefixes.

What would be interesting is to export testapi as a service (either HTTP as @coolo explored in #257 (although this is on different topic) or some other IPC (hey, we have DBUS there already ;) )). Which is IMO quite a task and even once testapi is converted, how about test libs and basetests?
But if we eventually would like to support, or provide ability to write tests in other languages, we would eventually need do that anyway.

coolo:
interesting hack, but I agree 100% with @aaannz

sysrich:
Me too, I've thought about this repeatedly, and end up at the same libs and basetests concerns as @aaannz 

Without a solution for that, this feels like a 'hack' to me. Anyone using it would have to implement their entire test suite, main.pm and all, in Python.

That prevents it from being useful in the openSUSE context entirely - where we want individual teams to be able to contribute tests in a language of their own choice

bmwiedemann:
I think the one missing thing is that you cannot easily create a python class that is a sub-class of a perl class.
But if users can live with non-pure python test modules, they can have the use base 'footest'; and use Inline Python  on top of the python test code (that would not even need any code changes in os-autoinst core)

sysrich:
As he's a bit of a python fan (understatement?) and would be a likely user of this feature, I've thrown this PR under the nose of @AdamWill who's opinion is that it would be of limited use.. quoting

"it's only really useful if you need to do a whole bunch of clever logic or flow control or something and you don't want to learn to do it in perl, which seems a fairly minor case"

As even he mostly agrees with @aaannz , @coolo and myself I'm going to close this PR

Nice idea, but not the way we want to address the desire to support different test languages in openQA

okurz:
we should really use labels to distinguish PRs and issues, e.g. like the bugzilla resolved states and such. this pr could have "discussion" and "wontfix" or similar.

jcrben:
Did anyone ever mess around with alternative approaches to different languages? I'm interested in this project but my brain already has quite a few programming languages and given Perl's continued decline (https://insights.stackoverflow.com/trends?tags=perl) it's tough for me to muster the energy to work with it.

aaannz:
I did and I still have some WIP branch with this. I never finished it though, there are numerous races and problems when parsing main.pm test loader in my approach. As I mentioned before above I did refactor testapi into testapi server and client. Idea was, that when you want to add another language, you just need to write testapi wrapper in given language dispatching commands though socket. 

In my perl implementation server is accessible through query_isotovideo sub handling socket comms. You can take a look into my branch https://github.com/aaannz/os-autoinst/commits/decoupled_testapi however it is not working 100%, it can start tests under some circumstances, but usually hit race condition during main.pm parsing. And is in dire need of rebasing.

coolo:
That there are less perl questions on stackoverflow over time is not necessarly a bad sign ;)

And you don't need to be a master perl hacker to write tests.

jcrben:
True about being a master hacker, altho I think less perl questions might be more due to a decline of novices and a lack of major web / data science libraries rather than a lack of bugs and perfect docs. 😛 

One of the bigger barriers is just dealing with the mess of setting up a new toolchain. 
From https://github.com/os-autoinst/os-autoinst#how-to-contribute it's not clear that you have an automated dev environment using something like docker or vagrant. I know from experience that things like pyenv, phpbrew, nvm, delve (go) are fragile to set up and keep unbroken - I prefer to wrap them in dev environments. The fact that perl's community is a fraction of the size suggests its toolchain will be fragile.

coolo:
that won't change if we add more dependencies to allow different languages. to contribute means writing perl - even on master level possibly. But the tests itself are a different beast.

BTW: checkout https://github.com/os-autoinst/openQA/blob/master/docker/testing/README.md if you want to look at openqa in a container

mudler:
Less Perl questions, since most of answers to your questions have already been answered before stackoverflow even existed 😃  for an example of good scientific libraries - have a look at PDL (i prefer it over Matlab/Octave, e.g.) or BioPerl; This [xkcd](https://xkcd.com/519/) nicely highlights how Perl can be also a learning experience!

I don't think the "language decline" is a good point to support a switch, Perl community never disappeared, it's just though to keep newcomers in the loop - On the other side i agree that we should likely support writing testsuites in other languages as well.


My 2c: to really achieve this in a clean way, the testapi should be delivered in a free language context (e.g. RPC, RMI, REST Services ) and tests would be a DSL that refer to basic API calls.

coolo:
actually we tried - and it's horrible :)

But according to the graph given, most newbies ask questions about javascript - so if just http://search.cpan.org/~claesjac/JavaScript-1.16/ was uptodate ...

AdamWill:
@jcrben FWIW, I came into using openQA for Fedora knowing zero perl (only Python and some PHP) and picking it up as I go along hasn't really been a problem. Especially for writing tests you really just don't need to know a whole lot. I'm not sure the return on investment for doing this would be worth it.

PR #365, opened 2015-12-07
add small program to convert a sound file to a png

coolo:
and needle it using our standard API. This avoids the dependency on this
rather fragile and intrasparent DTMF+multimon combination


coolo:
the /suse/coolo in the PR gives it away that I'm torn on what to do with the program. package snd2png as extra package? (to avoid .travis installing its dependencies also) or install it from os-autoinst

coolo:
decided that installing it will do for now - makes it easier to sync changes

coolo:
http://gertrude.suse.de/tests/42/modules/aplay/steps/12

coolo:
I guess we can also remove the "DTMF" display from the webui ;)

PR #366, opened 2015-12-09
readd overlap in the sampling

coolo:
This just makes the image smoother - and less victim of bad time aligning

### DIFF
--- a/snd2png/snd2png.cpp
+++ b/snd2png/snd2png.cpp
@@ -1,3 +1,4 @@
+
 /*
   snd2png.cpp - idea is from http://snd2fftw.sourceforge.net/, but the code doesn't
   have more similiarity to its grandfather than to a random fftw example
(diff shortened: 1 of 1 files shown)


snd2png/snd2png.cpp
```diff
@@ -66,6 +67,11 @@ max_row_value (double *points, int count, int index)
 int
 main (int argc, char *argv[])
 {
+  if (argc < 3) {
+    fprintf (stderr, "Usage: snd2png [soundfile] [imagefile]\n");
```
aaannz: Doesn't `[]` usually mark optional argument?

coolo: that has true in it ;)

okurz:
-1 for typo in commit message, otherwise considered +1

coolo:
fixed the argument description

PR #367, opened 2015-12-09
reimplement the wait_(screen_still|changed_screen) functions

coolo:
not to check the image ourselves, but to rely on the backend thread to check

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -19,7 +19,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   send_key send_key_until_needlematch type_string type_password
 
   assert_screen check_screen assert_and_dclick save_screenshot
-  wait_screen_change assert_and_click mouse_hide mouse_set mouse_click
+  assert_and_click mouse_hide mouse_set mouse_click
   mouse_dclick mouse_tclick match_has_tag
 
   script_run script_sudo script_output validate_script_output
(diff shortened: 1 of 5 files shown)


okurz:
+1

I have good experiences stating the reason or benefit of the change in the subject line of the commit and pull requests, i.e. turn around commit subject and description.

testapi.pm
```diff
@@ -279,9 +279,50 @@ sub upload_asset {
...
+    my ($callback) = @_;
+
+    bmwqemu::log_call('wait_screen_change');
+
+    # get the initial screen
+    $bmwqemu::backend->set_reference_screenshot;
+    $callback->() if $callback;
+
+    my $starttime        = time;
+    my $timeout          = 10;
```
aaannz: hard coded timeout? I'm for to have 'timeout => [0-9]+' as option.

aaannz: Or as another positional argument as in `wait_still_screen` lower.

aaannz:
Travis complaining on using prototypes in testapi. When did we remove prototypes from testapi? I thought this was the only place we actually wanted them.

testapi.pm
coolo: I have no idea how to express this in the parameter line ;)

coolo:
actually last time I checked you require the prototype for perl to accept code as parameter

aaannz:
Indeed, but I have been thinking, albeit not expressed enough apparently, differently about testapi and I'm in favour of having testapi simulate kind-of-DSL. I surely can live with that, I'm just surprised.

testapi.pm
aaannz: `wait_screen_change(sub {...}, $timeout)` wouldn't work?
and
`my ($cb, $timout) = @_;` on other side?

coolo:
it still works kind-of-DSL - all perl requires is that it knows it's a function.

coolo:
what you're lacking without prototypes is the "too few arguments" error

coolo:
added a timeout paramter and shut up perlcritic (hopefully)

aaannz:
Aha, all those years and didn't know that! I needed to try it right now to actually believe you and you were right!

PR #368, opened 2015-12-11
Move assert screen

coolo:


backend/baseclass.pm
```diff
@@ -25,10 +25,12 @@ our $backend;
 use parent qw(Class::Accessor::Fast);
 __PACKAGE__->mk_accessors(
     qw(
-      update_request_interval last_update_request
-      screenshot_interval last_screenshot last_image
-      reference_screenshot)
-);
+      update_request_interval last_update_request screenshot_interval
+      last_screenshot last_screenshot_name_ last_image
```
aaannz: Why is there underscore at the end of `last_screenshot_name_`?

testapi.pm
```diff
@@ -307,7 +426,7 @@ within the block to avoid races between the action and the screen change
 
 =cut
 
-sub wait_screen_change(&@;$) {
+sub wait_screen_change(&@) {
```
aaannz: I think, since we cleared my confusion, this prototype can be removed.

coolo: no, as I expressed - without the prototype you can't write 

```
wait_screen_change { send_key 'esc'; };
```

Only 
```
wait_screen_change sub { send_key 'esc' }; 
```
would work, but that's a bit too much perl

backend/baseclass.pm
coolo: because I need a property to store the attribute. And then I need a function so the testapi can fetch it. But as the testapi<->backend communication are hashes, so I can't call the attribute function directly.

coolo:
the convert_blessed converts needle objects calling TO_JSON in the needle.

backend/baseclass.pm
coolo: now I understand - all other privates have the underscore at the start

coolo:
0239a41

PR #371, opened 2015-12-16
Fix zkvm serial

coolo:


okurz:
-1. I think the behaviour is ok to skip saving a vm but I would much more prefer to have a refactoring on the names. It should ignore "saving a VM" if asked. Failing is the correct behaviour. Maybe "save_snapshot" would be better which can then (silently) fail. "load_snapshot" would then tell in a log message that there is no snapshot and starts from scratch instead.

coolo:
how is save_snapshot and save_vm different beside the name?

okurz:
not at all. It's all about good names

coolo:
http://refactoring.com/catalog/renameMethod.html is even in the catalogue ;)

okurz:
exactly. http://clean-code-developer.de/die-grade/roter-grad/#Einfache_Refaktorisierungsmuster_anwenden
I can bring you the red wristband tomorrow

PR #372, opened 2015-12-17
do not even try loadvm or savevm unless running on qemu

coolo:
we have to abort the tests - instead of dieing in notimplemented


coolo:
@okurz ^

okurz:
LGTM

okurz:
You reached *red grade* :-) : geeko-red-grade

PR #374, opened 2016-01-06
test creating support fixes

aaannz:
- first commit allows one to provide KEEPHDDs variable even when VM pool is empty, thus disk images has to be created yet.
- the rest changes behavior of MAKETESTSNAPHOSTS:
Instead of creating snapshot before each test it only acts as an override to 'milestone' test flag. This way snapshot is always only one, storing the latest successful test module. User than can use SKIPTO to restart failed test module.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2015 SUSE LLC
+# Copyright © 2012-2016 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -116,10 +116,18 @@ sub runalltests {
     write_test_order();
 
     for my $t (@testorder) {
-        my $flags = $t->test_flags();
+        my $flags    = $t->test_flags();
+        my $fullname = $t->{fullname};
 
-        if (!$vmloaded && $t->{fullname} eq $firsttest) {
-            load_snapshot($firsttest) if $bmwqemu::vars{SKIPTO};
+        if (!$vmloaded && $fullname eq $firsttest) {
+            if ($bmwqemu::vars{SKIPTO}) {
+                if ($bmwqemu::vars{TESTDEBUG}) {
+                    load_snapshot('lastgood');
+                }
+                else {
+                    load_snapshot($firsttest);
+                }
+            }
             $vmloaded = 1;
         }
         if ($vmloaded) {
@@ -128,7 +136,7 @@ sub runalltests {
             $t->start();
 
             # avoid erasing the good vm snapshot
-            if ($snapshots_supported && (($bmwqemu::vars{SKIPTO} || '') ne $t->{fullname}) && $bmwqemu::vars{MAKETESTSNAPSHOTS}) {
+            if ($snapshots_supported && (($bmwqemu::vars{SKIPTO} || '') ne $fullname) && $bmwqemu::vars{MAKETESTSNAPSHOTS}) {
                 make_snapshot($t->{fullname});
             }
 
@@ -136,9 +144,8 @@ sub runalltests {
             $t->save_test_result();
 
             if ($@) {
-
                 bmwqemu::diag $@;
-                if ($flags->{fatal} || !$snapshots_supported) {
+                if ($flags->{fatal} || !$snapshots_supported || $bmwqemu::vars{TESTDEBUG}) {
                     bmwqemu::stop_vm();
                     return 0;
                 }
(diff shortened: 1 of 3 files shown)


okurz:
sounds awesome (but honestly I don't understand it yet ;-) ) don't you need to update doc/backend_vars.asciidoc , too? Or maybe you can also provide your hints to https://github.com/os-autoinst/openQA/blob/master/docs/WritingTests.asciidoc#using-snapshots-to-speed-up-development-of-tests ?

autotest.pm
```diff
@@ -147,13 +142,13 @@ sub runalltests {
                 }
             }
             else {
-                if ($snapshots_supported && $flags->{milestone}) {
+                if ($snapshots_supported && ($flags->{milestone} || $bmwqemu::vars{MAKETESTSNAPSHOTS})) {
                     make_snapshot('lastgood');
```
nilxam: hmm...that means if MAKETESTSNAPSHOTS=1 all snapshot tag name(qemu-img snapshot -l) will be lastgood, am I right? I thought the goal of MAKESNAPSHOTS is user can get a snapshot with clear snapshot point(name) what they can jump to the place they want, this change sounds not a good idea to me...

aaannz: Right now I'm merely questioning the usefulness of the original implementation. What is the use case of having snapshots for each test module?
When I develop new tests I only needed to quickly get to the latest failed test module to fix it. (This brings another question whether the rollback for non fatal tests at L:149 should be allowed)

Ultimately I see that I will introduce another variable to preserve the original functionality and introduce mine.

nilxam:
I'm not sure I like this change as 1) it will not really SKIP to where I would like to stop 2) only one lastgood snapshot tag inside of snapshot if MAKETESTSNAPSHOTS enabled...

autotest.pm
nilxam: eg. sometimes I'd like to jump to a step before system settings changed then doing debugging

coolo: How do you jump to the latest failed test if the test before that didn't create a snapshot?

backend/qemu.pm
```diff
@@ -405,33 +405,34 @@ sub start_qemu {
...
+
+    # fresh HDDs
+    for my $i (1 .. $vars->{NUMDISKS}) {
+        # skip HDD refresh when HDD exists and KEEPHDDS or SKIPTO is set
+        next if ($keephdds && (-e "$basedir/$i.lvm" || -e "$basedir/$i"));
+        no autodie qw(unlink);
+        unlink("$basedir/l$i");
+        if (-e "$basedir/$i.lvm") {
+            symlink("$i.lvm", "$basedir/l$i") or die "$!\n";
+            runcmd("/bin/dd", "if=/dev/zero", "count=1", "of=$basedir/l1");    # for LVM
```
coolo: why would you hardcode paths - especially to /bin?

coolo: ah, the code just moved - it's still bogus.

aaannz:
I'm not sure I get the 1) point. Sure if you want to skip to arbitrary test module, then this won't work. But if you want to skip to the first failed module, then it will skip to where SKIPTO points.

backend/qemu.pm
aaannz: I'm not sure we even use LVM anywhere, so I just copied it without much thinking.

coolo:
SKIPTO can only skip to created snapshots - and so it needs to be created in the previous run

aaannz:
Well in this change SKIPTO only skips test modules and once on correct module it loads 'lastgood' snapshot as it is the only one present.

aaannz:
It decouples test module names from snapshot names.

coolo:
in installation there is no lastgood

aaannz:
If you have MAKETESTSNAPSHOT then there is. It overrides missing 'milestone' flag and create lastgood anyway.

nilxam:
for my 1) is about, eg, I want to skip to a fresh installed system before consoletest_setup change the system settings

aaannz:
I don't see the use case unless consoletest_setup failed. Maybe when one makes mistake and module passes even when not supposed to. But in this case I would rather wait for another 10mins then have 20GB+ qcow file.

I experimented with MAKETESTSNAPSHOTS=1|all to create all snapshots and when e.g. MAKETESTSNAPSHOTS='console-consotetest_setup' would create snapshot just before consoletest_setup module. But I found it annoying to use and even harder to explain.

coolo:
you misunderstand the use case. We want to debug the system under test, not the test.

aaannz:
Yes, indeed that's the case. I'll add this idea using different variable for test debug. Now back to the naming problem. :)

aaannz:
I restored original meaning of MAKETESTSNAPSHOTS and instead introduced TESTDEBUG variable which overrides 'milestone' and 'fatal' test flags: lastgood snapshot is updated after each successful test module and whole job is aborted after first failing test module. SKIPTO && TESTDEBUG skips to specified test module and loads lastgood snapshot. Doc update included, except openQA test writing documentation which is still to be updated.

okurz:
LGTM

PR #375, opened 2016-01-07
do not add perl variables as backend variables

aaannz:
- update doc/backend_vars.asciidoc
- parse backend/baseclass.pm as common backend
- skip perl variables (i.e. bmwqemu::vars{$k} -> skip $k)

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -1,12 +1,13 @@
 Supported variables per backend
 -------------------------------
 
-.Common
+.COMMON backend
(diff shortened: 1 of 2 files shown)


okurz:
is this related to https://github.com/os-autoinst/os-autoinst/issues/339 ?

aaannz:
Yes, but it's not perfect. E.g. it always tries to remove ISO_$i because of changes in code I have quiet a difficulty to match it (without some nasty, case specific, hacks)

aaannz:
rebased

PR #376, opened 2016-01-08
Allow folder separated test distribution collections

okurz:
If a main.pm can not be found in the CASEDIR itself, it reverts to a new
approach trying to load the main.pm from a subfolder specified by DISTRI
within CASEDIR. This allows e.g. *opensuse* and *sle* flavors of SUSE
products to be maintained within one branch of test repository.

Prerequisite for https://progress.opensuse.org/issues/10162

DISTRI was actually unused but still checked for existance since
9df5fb6.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -65,6 +65,7 @@ $bmwqemu::scriptdir = $installprefix;
 bmwqemu::init();
 
 # Sanity checks
+die "CASEDIR environment variable not set, unknown test case directory" if !defined $bmwqemu::vars{CASEDIR};
 die "No scripts in $bmwqemu::vars{CASEDIR}" if !-e "$bmwqemu::vars{CASEDIR}";
 
 bmwqemu::clean_control_files();
@@ -112,8 +113,14 @@ use commands;
 
 $bmwqemu::vars{BACKEND} ||= "qemu";
 
-# Load the main.pm from the casedir checked by the sanity checks above
-require $bmwqemu::vars{CASEDIR}."/main.pm";
+# Try to load the main.pm from one of the following in this order:
+#  - product dir
+#  - casedir
+#
+# This allows further structuring the test distribution collections with
+# multiple distributions or flavors in one repository.
+$bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
+require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
 
 # set a default distribution if the tests don't have one
 $testapi::distri ||= distribution->new;
@@ -121,7 +128,7 @@ $testapi::distri ||= distribution->new;
 # init part
 bmwqemu::save_vars();
 # make sure the needles are initialized before the backend thread is started
-needle::init();
+needle::init($bmwqemu::vars{PRODUCTDIR} . "/needles");
 bmwqemu::init_backend( $bmwqemu::vars{BACKEND} );
 
 if ($init) {


okurz:
marking as [WIP] for now, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/920#issuecomment-170524899

okurz:
updated:
* also load needles directory depending on DISTRI if not found in CASEDIR

isotovideo
```diff
@@ -116,8 +116,16 @@ use commands;
...
+# Try to load the main.pm from either casedir checked by the sanity checks
+# above or fall back to load from a "distri" subdirectory to allow test
+# distribution collections with multiple distributions or flavors in one
+# repository.
+if (-f "$bmwqemu::vars{CASEDIR}/main.pm") {
+    require $bmwqemu::vars{CASEDIR} . "/main.pm";
+}
+else {
+    require $bmwqemu::vars{CASEDIR} . "/products/$bmwqemu::vars{DISTRI}/main.pm";
+}
```
okurz: As discussed with coolo: Should be
```
IF $TEST_VARIANT
ELSEIF $DISTRI-$VERSION
ELSEIF $DISTRI
ELSE $CASEDIR
```
and use path as resolved internally to load needles from same dir

aaannz: Isn't `$CASEDIR eq $DISTRI-$VERSION`?

okurz: It can be but this does not matter as it is a different meaning, "$CASEDIR" is the test repo which you *can* resolve by $DISTRI-$VERSION, which is used in the current approach by manual symlinks and such. In our case we use it within the $CASEDIR to distinguish different main.pm and needles but keeping the same tests

okurz:
TODO to self: openQA also needs to know the needle dir so maybe openQA resolves what the TEST_VARIANT is and tells this to os-autoinst when calling isotovideo:

> <coolo> okurz: openQA needs to know where the needles are, so we can't have the intelligence in os-autoinst ;(
<okurz> coolo: that's bad. it needs to know for the needle display in time stream or only for the pull down menue of candidates?
<coolo> okurz: needle show and needle edit
<okurz> coolo: is it both resolved in lib/OpenQA/Schema/Result/Jobs.pm:1009 ?
<coolo> okurz: lib/OpenQA/Utils.pm has a needledir too - this should be harmonized no matter what :)

isotovideo
aaannz: I don't follow, but I'll wait for the code.

coolo: you have to read the discussion in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/920 too ;)

okurz:
updated:
* load "main.pm" *and* needles from "PRODUCTDIR", if set, fallback to "CASEDIR"

--> e.g. please merge, it still works with non-updated *openQA* and/or *os-autoinst-distri-opensuse*. I can and will update *openQA* afterwards, then tests. Change in os-autoinst is backwards compatible.

isotovideo
```diff
@@ -112,8 +113,15 @@ use commands;
...
 
-# Load the main.pm from the casedir checked by the sanity checks above
-require $bmwqemu::vars{CASEDIR}."/main.pm";
+# Try to load the main.pm from one of the following in this order:
+#  - product dir
+#  - casedir
+#
+# This allows further structuring the test distribution collections with
+# multiple distributions or flavors in one repository.
+diag "product dir : $bmwqemu::vars{PRODUCTDIR}";
```
aaannz: This will produce undefined variable warning when not set. Why not switch it with L123?

okurz: > This will produce undefined variable warning when not set. Why not switch it
> with L123?

That shouldn't have be there, thanks for catching.

PR #377, opened 2016-01-11
add a new variable UEFI_PFLASH

lcp:
The OVMF upstream recommends to specify the firmware file with "pflash"
rather than "-bios" so the non-volatile variables can be stored in the
firmware file instead of NVvars in the ESP.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -244,6 +244,10 @@ sub start_qemu {
         $vars->{BIOS} //= $vars->{UEFI_BIOS};
     }
 
+    if ($vars->{UEFI_PFLASH}) {
+        $vars->{UEFI} = 1;
+    }
+
     if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
         # Non-absolute paths are assumed relative to /usr/share/qemu
         $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS};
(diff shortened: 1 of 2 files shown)


aaannz:
Please tidy up your code ( call `./tools/tidy` and amend/update PR).

backend/qemu.pm
```diff
@@ -540,7 +544,10 @@ sub start_qemu {
             }
         }
 
-        if ($vars->{BIOS}) {
+	if ($vars->{UEFI_PFLASH}) {
+            runcmd('cp', $vars->{BIOS}, 'ovmf.bin');
```
coolo: I'm not a huge fan of calling external tools for this - File::Copy offers a fine copy function

lcp:
Updated the commit per comments.

lcp:
I just found that the firmware file has to be converted into qcow2 format or savevm would fail, and it's remove the need to copy the file. I would update the commit later.

doc/backend_vars.asciidoc
```diff
@@ -65,6 +65,7 @@ TAPDEV;device name;undef;TAP device name to which virtual NIC should be connecte
 TAPSCRIPT;;;
 TESTDEBUG;boolean;0;Enable test debugging: override 'milestone' and 'fatal' test flags to 1. Snapshot are created after each successful test module and each fail aborts test run
 UEFI;;;
+UEFI_PFLASH;;;
```
coolo: the point of this file is to document the meaning of the variables. So it's your chance to add a description here.

lcp:
Updated doc/backend_vars.asciidoc

PR #378, opened 2016-01-11
Add more contribution documentation

okurz:


README.asciidoc
```diff
@@ -47,5 +50,42 @@ and we can use VNC to connect to the QEMU instance:
...
+once to setup your workspace and before every commit
+----
+make check
+----
+
+* You *may* also run local tests on your machine or in your own development
+environment to verify everything works as expected.
+
+* The git commit messages *should* follow the rules stated on
+http://chris.beams.io/posts/git-commit/[How to Write a Git Commit Message]
```
okurz: Feel free to give your feedback on this, e.g. the rule to capitalize each commits subject line. My personal opinion is that we should stick to the more common way of doing it, i.e. just stick to the described rules.

coolo: You can write your commit messages as you like, but we have *way* too many non-natives in the team to require poems in commit logs.

okurz: That's why it says *should*. Single line commit messages like "Fix typo in foobar test" is just fine and following the rules. I don't mind if someone expresses his non-nativeness (like me being german) in his commit messages but it should be understandable. And I am pretty sure everybody understands what "Start a line with a capital letter" if it is just about this simple part of the ruleset.

coolo: let me phrase it differently: I think the difference between may and should is not clear to all git contributors.

okurz: Should I
* EITHER include a reference to https://www.ietf.org/rfc/rfc2119.txt
* OR replace *should* by *must*
* OR rephrase it to say: "Some reviewers will repeatedly annoy you mentioning that your commit message does not meet these rules if you don't follow them"

okurz:
As an outlook. After finding a consensus on this one I want to go further downstream, e.g. for openQA basically adopt the same rules, i.e. mention there that we stick to the rules mentioned here. And for the test distributions I want to have the same rules (maybe less strict) but with the addendum of mentioning either bugzilla or progress items as well as "nice-to-have" local confirmation of test runs and such

README.asciidoc
coolo: the 3rd :)

coolo: the english I'd understand is:

+* Don't underestimate the importance of the git commit messages - take the guide
+http://chris.beams.io/posts/git-commit/[How to Write a Git Commit Message] as reference.

okurz: Exactly, that's why it says *should*.
* option 4: Replace *should* by *may*
* option 5: Rewrite to
```
For git commit messages use the rules stated on
http://chris.beams.io/posts/git-commit/[How to Write a Git Commit Message]
as a reference
```

coolo: :+1: for 5

okurz:
updated:

* git commit rules "as reference", no *should*

PR #379, opened 2016-01-12
JSON fixes

nadvornik:
- use incr_parse() from JSON module
- utf-8 is disabled in both _send_json and _read_json
  (the originaly used JSON::decode_json is utf-8 enabled which
  caused inconsistencies)
- write json without trailing "\n"
  parsing ends at last '}' and if the following "\n" appeared at
  the begginig of buffer, it was considered to be part of next
  command and caused _read_json to block without screen updates etc.
  until the next command really arrived
- changed MAGIC_PIPE_CLOSE_STRING to JSON format to simplify
  parsing

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -33,7 +33,6 @@ use Data::Dumper;
 use feature qw(say);
 
 my $framecounter = 0;    # screenshot counter
-our $MAGIC_PIPE_CLOSE_STRING = "xxxQUITxxx\n";
 
 # should be a singleton - and only useful in backend thread
 our $backend;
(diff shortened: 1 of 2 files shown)


coolo:
lgtm

okurz:
Well it has less code than before so that is good :-)

There is a typo in the commit message "begginig" which could be fixed. And maybe you can reference what this change actually fixes? E.g. a test or a new use case it allows.

nadvornik:
The typo is fixed.

The original bug can be easily reproduced for example by something like
wait_serial("\xFF", 200)
which causes infinite loop in _read_json.

okurz:
> wait_serial("\xFF", 200)
> which causes infinite loop in _read_json.

that's a good one :-)

coolo:
I have the gut feeling that this won't be the last time we touch this code though :)

PR #380, opened 2016-01-12
Add -kernel, -initrd and -append QEMU arguments and also use AArch64 setup for ARM too

garretraziel:
This contains two commits - first adds possibility to use direct kernel boot with `-kernel`, `-initrd` and `-append` QEMU arguments (I tried to merge this with code for specifying BIOS variable to avoid code duplication). Second uses AArch64 (ARM64) options (`-device VGA` and not supporting boot order and also connecting kbd keyboard) for 32bit ARM also.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -321,7 +321,7 @@ sub start_qemu {
     my $arch_supports_boot_order = 1;
     my $use_usb_kbd;
     my @vgaoptions;
-    if ($vars->{ARCH} eq 'aarch64') {
+    if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm') {
         push @vgaoptions, '-device', 'VGA';
         $arch_supports_boot_order = 0;
         $use_usb_kbd              = 1;
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -248,13 +248,14 @@ sub start_qemu {
         $vars->{UEFI} = 1;
     }
 
-    if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
-        # Non-absolute paths are assumed relative to /usr/share/qemu
-        $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS};
-    }
-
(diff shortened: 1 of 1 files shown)


okurz:
+1

coolo:
fyi: this touches the same location as #377

backend/qemu.pm
```diff
@@ -244,13 +244,14 @@ sub start_qemu {
...
 
-    if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
-        # Non-absolute paths are assumed relative to /usr/share/qemu
-        $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS};
-    }
-
-    if ($vars->{BIOS} && !-e $vars->{BIOS}) {
-        die "'$vars->{BIOS}' missing, check BIOS\n";
+    foreach my $attribute (qw/BIOS KERNEL INITRD/) {
+        if ($vars->{$attribute} && $vars->{$attribute} !~ /^\//) {
```
aaannz: I think check for `..` wouldn't hurt.

coolo:
please rebase

backend/qemu.pm
garretraziel: I don't understand what do you mean by that. Do you mean normalizing path like Python's [normpath](https://docs.python.org/2/library/os.path.html#os.path.normpath)?

garretraziel:
Rebased to master.

backend/qemu.pm
aaannz: I mean to prevent escape from `/usr/share/qemu/` directory.

garretraziel: Escaping `/usr/share/qemu` directory is still possible by using absolute path, but sure, I will do it.

aaannz: Ah right. Scratch all that. I set my mind that only relative path is allowed. And I completely ignored that use parse `KERNEL` and `INITRD` in the same loop.

coolo:
aehm, that's not rebase that's merged. Please use git rebase - https://git-scm.com/docs/git-rebase

garretraziel:
Sorry, muscle memory.

PR #383, opened 2016-01-14
Fix test suites without distribution

coolo:
smaller test suites don't require a distribution, but my previous attempt
to fix the problem was too late as it only hit the main thread not the
backend thread. So we need to do the setup of the default distribution earlier

(tested this time with os-autoinst-distri-example)


shyukri:
Confirm that fix works fine. Thanks!

PR #384, opened 2016-01-14
Easier detectable errors for non-needle failures

coolo:
Both wait_serial and die '' were pretty hard to detect on the webui
as they only reveal themselves to the knowing as red screens with a
hint in the log file. That's very hard for beginners to grasp

So change that: In case of wait_serial and die we record a 'fail' text result
together with a screenshot.

See https://progress.opensuse.org/issues/6526

taking this test

```
sub run {
   wait_serial 'hallo', 3;
   die "No Hallo";
}

```

openQA result in master:
before

openQA result after my PR:
before1

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -94,16 +94,13 @@ sub record_screenmatch {
     $tags           ||= [];
     $failed_needles ||= [];
 
-    my $count    = ++$self->{test_count};
-    my $testname = ref($self);
-
     my $h          = $self->_serialize_match($match);
     my $properties = $match->{needle}->{properties} || [];
     my $result     = {
         needle     => $h->{name},
         area       => $h->{area},
-        tags       => [@$tags],                                  # make a copy
-        screenshot => sprintf("%s-%d.png", $testname, $count),
+        tags       => [@$tags],                        # make a copy
+        screenshot => $self->next_resultname('png'),
         result     => 'ok',
         properties => [@$properties],
     };
@@ -164,11 +161,6 @@ sub _serialize_match {
             $na->{$i} = $a->{$i};
         }
         $na->{similarity} = int($a->{similarity} * 100);
-        if ($a->{diff}) {
-            my $imgname = sprintf("%s-%d-%s-diff%d.png", $testname, $count, $name, $diffcount++);
-            $a->{diff}->write(join('/', bmwqemu::result_dir(), $imgname));
-            $na->{diff} = $imgname;
-        }
         push @{$h->{area}}, $na;
     }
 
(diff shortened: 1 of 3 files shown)


basetest.pm
```diff
@@ -94,16 +94,13 @@ sub record_screenmatch {
...
-
     my $h          = $self->_serialize_match($match);
     my $properties = $match->{needle}->{properties} || [];
     my $result     = {
         needle     => $h->{name},
         area       => $h->{area},
-        tags       => [@$tags],                                  # make a copy
-        screenshot => sprintf("%s-%d.png", $testname, $count),
+        tags       => [@$tags],                               # make a copy
+        screenshot => $self->$self->next_resultname('png'),
```
okurz: Is this related? Why?

basetest.pm
```diff
@@ -164,11 +161,6 @@ sub _serialize_match {
             $na->{$i} = $a->{$i};
         }
         $na->{similarity} = int($a->{similarity} * 100);
-        if ($a->{diff}) {
-            my $imgname = sprintf("%s-%d-%s-diff%d.png", $testname, $count, $name, $diffcount++);
-            $a->{diff}->write(join('/', bmwqemu::result_dir(), $imgname));
-            $na->{diff} = $imgname;
-        }
```
okurz: why delete?

basetest.pm
```diff
@@ -291,13 +280,29 @@ sub runtest {
         $self->post_run_hook();
     };
     if ($@ || ($self->{result} || '') eq 'fail') {
-        my $msg = "test $name " . ($@ ? 'died: ' . $@ : 'failed');
-        warn $msg;
```
okurz: Can you also give an example who the log output changes?

okurz:
-1: really nice, minor issues. Probably doesn't need code change to resolve my questions

basetest.pm
coolo: good thing you spotted the $self->$self :)

basetest.pm
coolo: it's a very old idea - and these diffs aren't even uploaded to openQA.

basetest.pm
coolo: who?

okurz: s/who/how

coolo: ah, it doesn't change at all. I just moved the trigraph into the if

basetest.pm
coolo: actually the diff elements are gone since 99ff3f5d9b572e779b7ddbeb303cb29862925d39 - Jul 2014

basetest.pm
```diff
@@ -291,13 +280,29 @@ sub runtest {
...
+                $self->_result_add_screenshot($result);
+            }
+            my $details = {result => 'fail'};
+
+            my $text_fn = $self->next_resultname('txt');
+            open my $fd, ">", bmwqemu::result_dir() . "/$text_fn";
+            print $fd "# Test died:\n$@\n";
+            close $fd;
+            $details->{text}  = $text_fn;
+            $details->{title} = 'DIE';
```
aaannz: Looking at the screen shot, I'm not sure we want openQA to scream 'DIE' on unsuspecting user. Too bad `$text_fn` won't fit, but what about putting there first few chars that fit and `...` to signalize hint text?

sysrich: Discussed with @coolo , my vote is for replacing 'DIE' with a red "Failed" thumbnail

coolo: a4ec8a2

PR #386, opened 2016-01-15
use fputs rather than fprintf to avoid 'format not a literal'

AdamWill:
These three identical lines cause the infamous 'format not a
string literal and no format arguments' warning in GCC. On
many distros (including Fedora) the default compile flags for
packages turn this warning into an error.

I don't think there's a real problem here, because ERR_NOT_
ENOUGH_MEMORY is a constant defined in the code. But we should
avoid triggering the warning/error in any case.

https://fedoraproject.org/wiki/Format-Security-FAQ

### DIFF
--- a/snd2png/snd2png.cpp
+++ b/snd2png/snd2png.cpp
@@ -89,7 +89,7 @@ main (int argc, char *argv[])
 			   info_in.channels);
   if (!infile_data)
     {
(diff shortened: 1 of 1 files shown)


coolo:
oops - that just shows that I did too much perl lately :)

PR #388, opened 2016-01-15
drop some stale requires from cpanfile

AdamWill:
Neither of these is used any more. The same change should be
made to the spec file.


coolo:
yeah, those came from the kvm2usb backend

PR #389, opened 2016-01-16
testapi: Prevent misleading 'assert_screen' name tag in 'check_screen'

okurz:


coolo:
Hmm, isn't it rather confusing to have fcres for functions never called?

coolo:
I.e. don't use a constant but calculate $function_name depending on $check.

okurz:
I don't know why the caller of "check/assert" would help as we know this from the 'Debug: …' line stating the line of the test script already. What I am trying to correct here *first* is that it shows "assert_screen" when it could be actually "check_screen". If you want to not show the internal helper function "_check_or_assert" it would make sense to continue on #390 and set the name explicitly to `(caller(2))[3]` instead of `(caller(1))[3]`in this case.
WDYT?

coolo:
aehm, now that looks completely wrong. $check is a boolean not a name

okurz:
yes, sorry. was looking at wrong debug message.

updated:
* as per your comment (for real now?)


this is how it should look like
```
<<< assert_screen(mustmatch='password-prompt', timeout=30)
MATCH(password-prompt-20141126:0.00)
no match 30
MATCH(password-prompt-20141126:0.84)
>>> assert_screen: found password-prompt-20141207, similarity 1.00 @ 0/16
…
Debug: /var/lib/openqa/share/tests/opensuse/tests/console/zypper_ref.pm:20 called testapi::check_screen
<<< check_screen(mustmatch='new-repo-need-key', timeout=20)
MATCH(MY_NEEDLE-20160116:1.00)
>>> check_screen: found MY_NEEDLE-20160116, similarity 0.99 @ 57/0
```

Still I recommend to go on with #390 then and call 'caller[2]' explitly for this method. WDYT?

coolo:
LGTM. I still prefer this minimal change

PR #392, opened 2016-01-18
Support 8 bit VNC

coolo:
Some (otherwise headless) Power blades can show their serial console through
VNC, but it's 256 colours in a colour map only. So far we only supported true
colour (e.g. rgb555 in 16 bits or rgb888 in 32 bits). To support indexed colours
I had to rework the VNC logic quite a bit, but I had it already prepared

I removed RRE support because it's basically unused now that we have ZLRE (which
has RRE as subtype, but adds zlib compression on top)

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -21,6 +21,7 @@ __PACKAGE__->mk_accessors(
     qw(hostname port username password socket name width height depth save_bandwidth
       no_endian_conversion  _pixinfo _colourmap _framebuffer _rfb_version screen_on
       _bpp _true_colour _do_endian_conversion absolute ikvm keymap _last_update_request
+      vncinfo
       ));
 our $VERSION = '0.40';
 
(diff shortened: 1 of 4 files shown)


k0da:
Those are not Blades, can be any tty exported with linuxvnc.

consoles/VNC.pm
```diff
@@ -1025,7 +1008,15 @@ sub _receive_ikvm_encoding {
...
-    die 'we do not support color maps';
+    $self->socket->read(my $map_infos, 5);
+    my ($padding, $first_colour, $number_of_colours) = unpack('Cnn', $map_infos);
+
+    for (my $i = 0; $i < $number_of_colours; $i++) {
+        $self->socket->read(my $colour, 6);
+        my ($red, $green, $blue) = unpack('nnn', $colour);
+        tinycv::set_colour($self->vncinfo, $first_colour + $i, $red / 256, $green / 256, $blue / 256);
+    }
+    #die "we do not support color maps $first_colour $number_of_colours";
```
okurz: why do you disable the die?

ppmclibs/tinycv.xs
```diff
@@ -28,17 +28,22 @@ tinycv::Image read(const char *file)
...
+			    green_mask, green_shift,
+			    blue_mask, blue_shift);
 
    OUTPUT:
      RETVAL
 
+void set_colour(tinycv::VNCInfo info, unsigned int index, unsigned red, unsigned green, unsigned blue)
+   CODE:
+     image_set_vnc_color(info, index, red, green, blue);
+
```
okurz: you are introducing w/s at EOL here

PR #393, opened 2016-01-18
Save git hash of executed tests

okurz:
If the test directory (CASEDIR) is a git repository the current commit hash is
saved just before the tests are loaded and stored inside "status.json" at the
end of execution. The git commit hash can be used for reference in later
analysis, e.g. in openQA to display the file not in the "most recent" state
but the actual state the test source was in at the time of execution.

If CASEDIR is not a git repo "UNKNOWN" is stored instead. A "dirty" state of
the repository is *not* detected.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -120,6 +120,11 @@ $bmwqemu::vars{BACKEND} ||= "qemu";
 # This allows further structuring the test distribution collections with
 # multiple distributions or flavors in one repository.
 $bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
+
+# as we are about to load the test modules store the git hash that has been
+# used. If it is not a git repo fail silently, i.e. store an empty variable
+chomp($bmwqemu::test_git_hash = qx{git rev-parse HEAD});
+
 require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
 
 # set a default distribution if the tests don't have one
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -62,6 +62,7 @@ share(@ocrrect);
 our $interactive_mode;
 our $waiting_for_new_needle;
 our $screenshotpath = "qemuscreenshot";
+our $test_git_hash;
 
 # shared vars end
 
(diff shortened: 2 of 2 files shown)


isotovideo
```diff
@@ -120,6 +120,11 @@ $bmwqemu::vars{BACKEND} ||= "qemu";
 # This allows further structuring the test distribution collections with
 # multiple distributions or flavors in one repository.
 $bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
+
+# as we are about to load the test modules store the git hash that has been
+# used. If it is not a git repo fail silently, i.e. store an empty variable
+chomp($bmwqemu::test_git_hash = qx{git rev-parse HEAD});
```
coolo: and git magically knows what directory you're talking about?

okurz: hm, it "magically" found the wrong repo of os-autoinst not the tests, obviously.

updated:
 * fix to go to actual test directory
 * writing to `vars.json` instead which is stored as result file within openQA jobs and as `vars.json` is changed by the test anyway

coolo: vars.json is the wrong file IMO

okurz: that does not help

okurz: would you prefer to just print it in a debug output and parse that from openQA reading the logfile?

coolo: No, status.json is the right file

okurz: > No, status.json is the right file

but it's not saved by openQA so …?

coolo: change openQA and don't abuse unrelated files.

okurz: I am not sure what the intention regarding `status.json` is but maybe the right solution would be to not change vars.json at all and either
 * store everything in `status.json`
 * store all changes done to vars.json instead inside `results.json` or similar

coolo: there is no global results.json. results are per test - while the status is for the whole test run. And these files are read by the worker and transferred back. vars.json is just kept for documentation - and removed after 14 days.

coolo:
I'm tempted to close this

okurz:
it's ok to close it as inactive. I failed to provide an update in time so I will close it for now. I will keep track of this in my local feature branch.

PR #394, opened 2016-01-19
Coolos exceptions

coolo:


basetest.pm
```diff
@@ -298,17 +310,15 @@ sub runtest {
...
         }
-        warn $msg;
-        $self->{post_fail_hook_running} = 1;
-        eval { $self->post_fail_hook; };
-        bmwqemu::diag "post_fail_hook failed: $@\n" if $@;
-        $self->{post_fail_hook_running} = 0;
-        $self->fail_if_running();
-        die $msg . "\n";
+    }
+    if (($self->{result} || '') eq 'fail') {
```
aaannz: shouldn't this be `elsif`?

coolo: yes, but run_post_fail is fatal - I only put the comment in once

OpenQA/Exceptions.pm
```diff
@@ -0,0 +1,30 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package OpenQA::Exceptions;
```
aaannz: This somehow bothers me. I don't want to introduce OpenQA namespace to os-autoinst unless we are going to make os-autoinst part of the openQA (I don't mean to move the repo, just I consider os-autoinst as independent tool (pending some exceptions like mmapi)).

coolo: Actually I spent most of my afternoon thinking about the classname and what namespace it should be in.

coolo: To me os-autoinst is a standalone tool within openQA. Most of what it does, it does because of openQA. I.e. the results-*.json files are meaningless without a displaying tool.

OpenQA/Exceptions.pm
```diff
@@ -0,0 +1,33 @@
...
+
+use Exception::Class (
+    'OpenQA::Exception::InternalException' => {
+        description => 'internal errors not for the user'
+    },
+
+    'OpenQA::Exception::FailedNeedle' => {
+        description => 'assert_screen failed',
+        isa         => 'OpenQA::Exception::InternalException',
```
okurz: that is confusing. I would define a BaseException and let both derive from there

basetest.pm
```diff
@@ -298,17 +310,15 @@ sub runtest {
             $details->{text}  = $text_fn;
             $details->{title} = 'Failed';
             push @{$self->{details}}, $details;
+            $self->run_post_fail("test $name died");
         }
         else {
```
aaannz: Whole `else` block can be removed since `$self->{result}` is set to 'fail' and the same code is then executed in next `if`. So there is reason to keep it `if` :)

aaannz:
Ok, you convinced me (after like 20minutes staring into diff and finally noticing that little 'isa' key in FailedNeedle to make it all sense). One suggestion to not make it too easy for you, else LGTM.

basetest.pm
coolo: the log message is different

aaannz: I see `$self->run_post_fail("test $name failed");` in both cases.

coolo: ah, I misunderstood what else you're talking about

coolo:
rewrote history

PR #395, opened 2016-01-20
Remember activated consoles and rollback too

coolo:
In case of tty consoles requiring console, we loose the login and need
to relogin. In case of local xterms, this is a bit more tricky - but
we cross that bridge when we come there (as local xterms aren't used
on platforms supporting rollback, that's a bit in the future)

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -71,13 +71,14 @@ sub loadtest {
 }
 
 our $current_test;
+our $last_milestone;
 
 sub set_current_test {
     ($current_test) = @_;
     bmwqemu::save_status();
 }
 
-sub write_test_order() {
+sub write_test_order {
 
     my @result;
     for my $t (@testorder) {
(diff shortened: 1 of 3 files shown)


coolo:
It worked fine in my example distribution test

aaannz:
LGTM

okurz:
+1

nilxam:
Maybe I am wrong, doesn't it will break console tests? I mean in console tests, after rollback and reset all consoles, but console is/may activated in system actually. Eg. A console test failedand rollback to end of consoletest_setup, then user-console still activated.

coolo:
You're right. We need to rollback to the last milestone test, not to the previous test. I feel stupid ;(

coolo:
updated - please check afd200c60098fc1b83122832b91e3d1f2a6c010e

nilxam:
:+1:

PR #397, opened 2016-01-21
Fix Makefile to install the perl modules properly

coolo:
Install OpenQA/Exceptions.pm and fix the permissions of the installed modules


okurz:
LGTM but how did this slip through if travis is actually using the Makefile? What is missing?

coolo:
travis doesn't install?

PR #398, opened 2016-01-21
do not store JSON objects in global hash

nadvornik:
This fixes https://progress.opensuse.org/issues/10330

Putting JSON object to global hash causes the leak. I guess it is related to threads and shared variables.
Using it in local variable (the same method calls with the same data) seems to work fine.

This patch stores simple strings instead of whole JSON objects.


coolo:
closer to what we did before actually.

PR #401, opened 2016-01-22
Avoid wait_idle calls in script_run

coolo:
Replaced it with wait_serial for a random string

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -78,11 +78,11 @@ sub ensure_installed {
     wait_still_screen(7, 90);    # wait for install
 }
 
-sub become_root() {
+sub become_root {
     my ($self) = @_;
 
     testapi::script_sudo("bash", 0);    # become root
-    testapi::script_run("test $(id -u) -eq 0 && echo 'imroot' > /dev/$testapi::serialdev");
+    testapi::script_run("test $(id -u) -eq 0 && echo 'imroot' > /dev/$testapi::serialdev", 0);
     testapi::wait_serial("imroot", 5) || die "Root prompt not there";
     testapi::script_run("cd /tmp");
 }
@@ -91,8 +91,9 @@ sub become_root() {
 
 script_run($program, [$wait_seconds])
 
-Run $program (by assuming the console prompt and typing it).
-Wait for idle before  and after.
+Run $program (by assuming the console prompt and typing it). After it, echo
+something to serial line and wait for it. To avoid that, use wait_seconds 0
+and wait yourself, e.g. by wait_idle
 
 =cut
 
@@ -101,8 +102,15 @@ sub script_run {
     # start console application
     my ($self, $name, $wait) = @_;
 
-    testapi::type_string "$name\n";
-    testapi::wait_idle($wait);
+    testapi::type_string "$name";
+    if ($wait > 0) {
+        my $str = bmwqemu::random_string(5);
+        testapi::type_string " ; echo $str > /dev/$testapi::serialdev\n";
+        testapi::wait_serial($str);
+    }
+    else {
+        testapi::send_key 'ret';
+    }
 }
 
 =head2 script_sudo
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -294,6 +294,10 @@ sub wait_idle {
     my $timeout = shift || $bmwqemu::idle_timeout;
     $timeout = bmwqemu::scale_timeout($timeout);
 
+    # report wait_idle calls more offensiv to help with https://progress.opensuse.org/issues/5830
```
okurz: s/offensiv/offensively/

okurz:
-1, fix typo and put a bit more details in the commit message for reasoning and benefit of this change. You might want to reference the progress issue. After these I would be ok if you merge it directly, i.e. --> auto-LGTM

distribution.pm
```diff
@@ -101,8 +102,15 @@ sub script_run {
     # start console application
     my ($self, $name, $wait) = @_;
 
-    testapi::type_string "$name\n";
-    testapi::wait_idle($wait);
+    testapi::type_string "$name";
+    if ($wait > 0) {
```
aaannz: Please add check if `$wait` is defined. Else we will have undef warnings all over the place.

aaannz: Ah, it's added on higher level. Ok, nevermind.

testapi.pm
```diff
@@ -294,6 +294,10 @@ sub wait_idle {
...
+    use Carp qw(cluck);
+    cluck "Wait_idle called";
```
okurz: I would add "deprecated: …" here

aaannz:
My one concern is use case like in `installation/zdup.pm`. `script_run "zypper dup"`, then `wait_serial` to check output. It's called without explicit `$timeout` so it will get default 30s. Would new `script_run` "steal" serial output? I think so.

testapi.pm
coolo: dude!

testapi.pm
coolo: it's not deprecated. I just want to reduce it's usage

coolo:
yes, it will - that's why I added , 0 to script_runs waiting itself.

coolo:
the API change is a big concern to me too - but I disagree with adding another function. The overhead in documentation and testing is just too big and we need to do this cleanup to be portable.

Fedora uses stable openQA - and we'll need an upgrade guide.

coolo:
Of course if you can think of a gentle way to do the switch, I consider it. But remember we have assert_script_sudo, script_sudo, script_run and assert_script_run all sharing the same problem.

testapi.pm
okurz: IMHO it's deprecated but not obsolete. "deprecated" means "please don't use" as in "reduce it's usage". That's what you said.

testapi.pm
okurz: well, if you change other stuff then you can also change this :-)

okurz:
> Fedora uses stable openQA - and we'll need an upgrade guide.

Then a first easy step is to mark the "wait_idle" as deprecated with the corresponding annoying log warning in "stable openQA"

coolo:
that won't help a dime as long as script_run and send_key use it. It will be just *very* annoying

sysrich:
I agree with @coolo

The first step to either deprecation or obsolescence is removing wait_idle
from script_run and send_key

I don't care which term or approach we take, as long as the outcome is that
we can move away from wait_idle usage in our test suite forthwith

And that means removing its use from those functions now,despite it being a
major change which will necessitate some work for us and other openQA users

okurz:
I am fine if we go that road but I don't agree that "the first step to deprecation" should be changing existing functionality. That is how deprecation differs from plain changing behaviour. I see marking "wait_idle" as deprecated as a hint to other users of openQA. For os-autoinst-distri-opensuse developers we can of course try to get rid of wait_idle but that is not directly related.

coolo:
I never said that wait_idle is obsolete or deprecated. You said that - I said, I want to get rid of its usage in our tests. Direct or indirect

coolo:
extended the API documentationa and the commit log -  7e44b20

okurz:
LGTM

coolo:
testing the changes in staging atm: https://tortuga.suse.de/tests/16

coolo:
(ran out of worker disk space)

coolo:
merging this now so we can deploy it tomorrow morning to find more issues in the tests

PR #402, opened 2016-01-27
Always store last

coolo:


okurz:
did you check the interactive mode still works with this change? If yes, auto-LGTM

nilxam:
LGTM. it should worked on interactive mode, interactive more is not really matter with assert_screen, and update_status is keep updating screenshot per md5 :-)

coolo:
interactive_mode is actually very easy to break in this code, but I don't think I did :)

PR #403, opened 2016-01-27
Improve error message on needle reading problems

okurz:
There is no use in trying to handle problems reading needle data in a custom
way. autodie provides much better details in this case.

Verified locally with a needle with wrong permissions. In before it would just
state that the needle is broken without a reason.

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -41,10 +41,8 @@ sub new {
     }
     else {
         local $/;
-        # TODO preserving old behaviour, why should failing read be allowed?
-        no autodie qw(open);
         open(my $fh, '<', $jsonfile);
-        eval { $json = decode_json(<$fh>) };
+        $json = decode_json(<$fh>);
         close($fh);
         if (!$json || $@) {
             warn "broken json $jsonfile: $@";


needle.pm
```diff
@@ -41,10 +41,8 @@ sub new {
     }
     else {
         local $/;
-        # TODO preserving old behaviour, why should failing read be allowed?
-        no autodie qw(open);
         open(my $fh, '<', $jsonfile);
-        eval { $json = decode_json(<$fh>) };
+        $json = decode_json(<$fh>);
         close($fh);
         if (!$json || $@) {
```
aaannz: So I guess `$@` can be removed as well.

aaannz:
The question is if we want to abort test due to wrong/inaccessible needle or just ignore that needle and continue with the rest?

needle.pm
okurz: no because the decode json can still report on internal errors or so.

okurz:
hiding needle access problems only makes debugging test failures harder. What would be a valid use of skipping individual needles? The test logic just won't work properly. I encountered this when my local needle repo was having needles from another user that were not readable by "_openqa-worker" causing problems. So I see three different scenarios

* 1: individual needles can not be read
* 2: all needles can not be read
* 3: all needles can be read

In case of 2 probably a lot of tests will fail even though the error might not be obvious since the "warn" is only in the logfile. In case of 1 it might be even more hideous why something is not working as only individual needles are skipped or "not matched". In both cases this is an issue that needs to be fixed by the openqa worker administrator and should be obvious.

needle.pm
aaannz: `decode_json` modifies `$@`? Is it documented somewhere?
What I found is that If there is an error it will `croak` which if not catched by `eval` you removed will terminate the script.

aaannz:
I'm just asking. Someone did that that way.
My reasoning is, when needle is unable to be read it is wrong needle and should be (re)created correctly. Sure, you will not notice it in webui, so you will just create new needle. And this old one will later be removed as unused.

okurz:
Hm, In my consideration a needle that is unreadable is not a "missing one". It got the filename from somewhere so the file is there.

coolo:
I can think of a case: tests that run while the admin cleans up old needles. The needles are there, but they were considered pointless anyway - why would they break the running test if we have a chance to just skip it?

coolo:
at least the test will die - and hopefully gets rescheduled

PR #404, opened 2016-01-27
First PowerVM backend implementation

k0da:
This is first implementation of a PowerVM backend.

### DIFF
--- a/backend/pvm.pm
+++ b/backend/pvm.pm
@@ -1,0 +1,158 @@
+package backend::pvm;
+use strict;
+use base ('backend::baseclass');
+use bmwqemu qw(fileContent diag save_vars);
+require IPC::System::Simple;
+use autodie qw(:all);
+use File::Basename;
+use Digest::MD5 qw(md5_hex);
+
+sub new {
+    my $class = shift;
+    my $self  = $class->SUPER::new;
+
+    $self->{pid}         = undef;
+    $self->{children}    = [];
+    $self->{pidfilename} = 'pvm.pid';
+    $self->{pvmctl}      = '/usr/bin/pvmctl';
+    die "pvmctl not found" unless -x $self->{pvmctl};
+
+    return $self;
+}
+
+sub do_start_vm {
+    my $self = shift;
+    $self->unlink_crash_file;
+    $self->start_lpar();
+    return {};
+}
+
+sub runcmd {
+    diag "running " . join(' ', @_);
+    return system(@_);
+}
+
+sub pvmctl {
+    my $self = shift @_;
+    my $vars = \%bmwqemu::vars;
+    my ($type, $action) = @_;
+    push(my @cmd, $self->{pvmctl}, $type, $action);
+    my $lpar = $vars->{LPAR};
+
+    if ($type =~ /lpar/) {
+        push(@cmd, "-i", "name=" . $lpar) if ($action =~ /power|restart|delete/);
+        if ($action =~ /create/) {
+            my ($type, $action, $cpu, $memory) = @_;
+            push(@cmd, qw{--proc-type shared --sharing-mode uncapped --type AIX/Linux});
+            push(@cmd, "--proc", $cpu, "--mem", $memory, "--name", $lpar);
+            push(@cmd, "--proc-unit", $cpu * 0.05);
+        }
+    }
+    elsif ($type =~ /scsi/) {
+        my ($type, $action, $kind, $file) = @_;
+        push(@cmd, "--type", $kind, "--stor-id", $file);
+        push(@cmd, "--lpar", "name=" . $lpar);
+        push(@cmd, "--vg", "name=rootvg") if ($type =~ /lv/);
+    }
+    elsif ($type =~ /lv/) {
+        my ($type, $action, $name, $size) = @_;
+        push(@cmd, "--name", $name, "--size", $size);
+    }
+    elsif ($type =~ /eth/) {
+        my ($type, $action, $vlan, $vswitch) = @_;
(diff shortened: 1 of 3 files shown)


aaannz:
Tidy is failing the travis.

k0da:
FIxed

backend/pvm.pm
```diff
@@ -0,0 +1,151 @@
...
+    powercontrol('off');
+    runcmd("rmvterm", "--id", $vars->{LPARID});
+    for my $i (1 .. $vars->{NUMDISKS}) {
+        my $disk = $vars->{LPAR} . "_" . $i;
+        $cmd = "$pvmctl scsi delete --type lv --lpar name=$vars->{LPAR} --stor-id=$disk";
+        runcmd($cmd);
+        $cmd = "sudo viosvrcmd --id 1 -c \"rmbdsp -sp rootvg -bd $disk\"";
+        runcmd($cmd);
+    }
+    $cmd = "$pvmctl scsi delete --type vopt --lpar name=$vars->{LPAR} --stor-id=$vars->{VIOISO}";
```
coolo: what bothers me here is that you duplicate this code - and even inconsistently. Here you use a string and in the create case you use an array. Both share a lot of paramters, so possibly they both should inherit a common array.

backend/pvm.pm
```diff
@@ -0,0 +1,151 @@
...
+    }
+    if ($action eq 'reset') {
+        $cmd = "restart";
+    }
+    runcmd("/usr/bin/pvmctl", "lpar", $cmd, "-i", "name=" . $vars->{LPAR});
+}
+sub attach_console {
+    my ($self, $args) = @_;
+    my $vars    = \%bmwqemu::vars;
+    my $vncport = qx/sudo \/usr\/sbin\/mkvtermutil --id $vars->{LPARID} --vnc --local --log serial0 2>\/dev\/null/;
```
coolo: this evil quoting looks horrible. You can use qx{} too - using / as seperator with a string with tons of / is kind of crazy

k0da:
Comments are addressed.

doc/backend_vars.asciidoc
```diff
@@ -84,3 +84,17 @@ VIRSH_HOSTNAME;string;SSH Host with virtsh;
...
+[options="header",cols="^m,^m,^m,v",separator=";"]
+|====================
+Variable;Values allowed;Default value;Explanation
+MEM;integer;2048;amount of RAM;
+LPAR;string;osauto;LPAR name to be created;
+NUMDISKS;integer;1;Number of disks;
+HDDSIZEGB;integer;15;Disk size in GB;
+NICVLAN;integer;1;VLAN to attach to;
+VSWITCH;string;VSWITCH0;A virtula switch to connect to;
+CPUS;integer;1;Number of CPUS for LPAR;
```
okurz: hm, the other backends use a prefix for their variables. Please check again. Do you really also need "MEM" and such?
@aaannz how does this relate to https://github.com/os-autoinst/os-autoinst/issues/339 ?

aaannz: Well, qemu backend doesn't use any prefix. Also its variables are not universal across backends. i.e. IPMI certainly doesn't have MEM option, so I am for listing it (and other similar) here.

k0da:
mem is fixed. 

Are vars needs to be prefixed? They are listed in documentation.

backend/pvm.pm
```diff
@@ -0,0 +1,164 @@
...
+    $vars->{MEM}  ||= "2048";
+    #disk settings
+    $vars->{NUMDISKS}  ||= 1;
+    $vars->{HDDSIZEGB} ||= 15;
+    #network settings
+    $vars->{NIC}     ||= "sea";
+    $vars->{NICVLAN} ||= 1;
+    $vars->{VSWITCH} ||= "ETHERNET0";
+    # generate the ISO name for VIO
+    my $os      = $vars->{DISTRI};
```
coolo: I was about to merge this, but I notice this and I have my doubts this belongs here. Such distribution specific things don't really belong in os-autoinst, left alone in a backend. IMHO this belongs in openQA - set in the medium

k0da: Yes. But this limitation is really stupid and I'm afraid I can't make it fixed. (Those parts are closed source). And I want to keep similarity across architectures, The iso name just changed temporarily in background.

coolo: you can still set the short ISO in openQA - we support variable expansion in the scheduler

k0da: Right, but isotovideo will stop working then.

coolo: isotovideo will require a VIOISO variable in the test - that's not "stop working"

k0da:
Ok, got rid of it. By generating a md5sum of a filename.

backend/pvm.pm
```diff
@@ -0,0 +1,159 @@
...
+
+    my $id = substr(qx/pvmctl lpar list -d LogicalPartition.id --where LogicalPartition.name=$vars->{LPAR}/, 3);
+    chomp($id);
+    $vars->{LPARID} = $id;
+    bmwqemu::save_vars();
+
+    #unfortunately NovaLink can't take files longer than 38 chars. So we need to shorten ISO name here
+    #we copy isos from nfs mount on VIO side to VMLibrary
+    my $source_iso = $vars->{ISO};
+    $source_iso =~ s/var\/lib\/openqa\/share\/factory\///g;
```
coolo: I would feel better if you used 'iso/' . basename($source_iso) - in the light of https://progress.opensuse.org/issues/10470 it sounds bad to hardcode openQA paths in os-autoinst.

k0da: Makes sense, yes. Thanks

PR #405, opened 2016-01-28
Add watchdog to sanity check timing

coolo:
If it takes longer than 2 seconds to write a screenshot, our
assumptions about timing in tests is screwed, so make sure
the worker is able to handle that

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -101,7 +101,7 @@ sub run {
     $self->{select}->add($self->{cmdpipe});
 
     $self->last_update_request("-Inf" + 0);
-    $self->last_screenshot("-Inf" + 0);
+    $self->last_screenshot(undef);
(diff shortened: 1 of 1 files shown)


coolo:
This PR is mainly to document the code I'm deploying on our workers - we need to define a sane limit. I'm unable to proceed a test with this limit on my workstation, so it's probably too strict.

coolo:
so as I expected, 90% of the tests fail now. The highscore for a single capture_screenshot is 33.4s

coolo:
I now added warnings for slow needle matching and slow capture_screenshot - and upped the limit to die for to 10 * screenshot_interval (which is 5s)

coolo:
I think it's a good thing to have - so I removed the tag

aaannz:
:+1:

PR #406, opened 2016-01-29
Add module name to uploaded files

nilxam:
Main goal is avoding the uploaded files be overwritten. Now the uploaded files are named $modulename-$filename

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -170,11 +170,18 @@ sub upload_file {
         mkdir($target) or die "$!";
     }
 
-    my $upname = basename($self->param('filename'));
+    my $upname   = $self->param('upname');
+    my $filename = basename($self->param('filename'));
+    # Only renaming the file if upname parameter has posted ie. from upload_logs()
+    # With this it won't renamed the file in case upload_assert and autoyast profile
+    # as those are not called from upload_logs.
+    if ($upname) {
+        $filename = basename($upname);
+    }
 
-    $upload->move_to("$target/$upname");
+    $upload->move_to("$target/$filename");
 
-    return $self->render(text => "OK: $upname\n");
+    return $self->render(text => "OK: $filename\n");
 }
 
 
--- a/testapi.pm
+++ b/testapi.pm
@@ -390,8 +390,9 @@ sub upload_logs {
     my ($file) = @_;
 
     bmwqemu::log_call('upload_logs', file => $file);
-    my $cmd      = "curl --form upload=\@$file ";
     my $basename = basename($file);
+    my $upname   = ref($autotest::current_test) . '-' . $basename;
+    my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
     $cmd .= autoinst_url("/uploadlog/$basename");
     return assert_script_run($cmd);
 }


aaannz:
I would rather prefer different approach.
* update `testapi.pm::upload_forms` like
```perl
#...
my $name = $autotest::current_test->{fullname} . '-' . basename($self->param('filename'));
my $cmd  = "curl --form upload=\@$file --form name=$name";
# ...
```
* update `commands.pm` upload forms handler to take name from `name` parameter

We already had discussion about getting rid of threads and go multi-process and `:shared` variables will not work there.

coolo:
that's a valid point

commands.pm
```diff
@@ -170,11 +170,18 @@ sub upload_file {
...
     }
 
-    my $upname = basename($self->param('filename'));
+    my $upname   = $self->param('upname');
+    my $filename = basename($self->param('filename'));
+    # Only renaming the file if upname parameter has posted ie. from upload_logs()
+    # With this it won't renamed the file in case upload_assert and autoyast profile
+    # as those are not called from upload_logs.
+    if ($upname) {
+        $filename = $upname;
```
coolo: you still need to call basename on the parameter - or you will give your filesystem away.

nilxam: it was basenamed in upload_logs() https://github.com/os-autoinst/os-autoinst/pull/406/files#diff-bd61f4009ce05bc0647daf1170d6facaL394 , we just basename it twice here like filename param. but updated.

testapi.pm
```diff
@@ -390,8 +390,9 @@ sub upload_logs {
     my ($file) = @_;
 
     bmwqemu::log_call('upload_logs', file => $file);
-    my $cmd      = "curl --form upload=\@$file ";
     my $basename = basename($file);
+    my $upname   = ref($autotest::current_test) . '-' . $basename;
```
aaannz: Can you use `$autotest::current_test->{fullname}`? It will then use `$category-$name` prefix. I think it will better address from where the logs are and I'm also afraid of tests with the same name possibly overwriting their logs. Or am I overthinking it too much?

nilxam: To me it just a too long filename to me, and I can't see the use case that the log will be overwrited  from anything else, do we have duplicate module name?

I'm fine with the long name if it's preferred. What other people think here? :)

okurz: :+1: for long name, better be unambiguous. We should try not to have double module name but we can not rely on it.

aaannz:
See https://github.com/os-autoinst/os-autoinst/pull/409#issuecomment-177817938 i.e. 'shutdown' we have `shutdown/shutdown.pm` and `x11/shutdown.pm`.

testapi.pm
aaannz: See https://github.com/os-autoinst/os-autoinst/pull/409#issuecomment-177817938 i.e. 'shutdown' we have `shutdown/shutdown.pm` and `x11/shutdown.pm`.

coolo: We should fail if module names are duplicated in one test - the job module index in the DB is the name not the fullname. And we only show the name in the overview page. So let's not make this more complicated *here*

okurz: Agreed, even better. Where can we fail? i.e. where are all modules referenced by shortname?

coolo: IMO loadtest should fail - right now it stores the tests in %tests by $fullname. I would make that fatal by $name

okurz: good, tracked as https://progress.opensuse.org/issues/10514

PR #408, opened 2016-01-29
Make the string used in script_* predictable

coolo:
Random strings are cool, but they make needling the screen content
more dangerous. So use the first bits of md5 as base64 instead

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -45,6 +45,7 @@ require IPC::System::Simple;
 use autodie qw(:all);
 
 use distribution;
+use Digest::MD5 qw(md5_base64);
 
 sub mydie;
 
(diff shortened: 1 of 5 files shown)


bmwqemu.pm
```diff
@@ -399,6 +400,18 @@ sub random_string {
...
 
+# return a short string representing the given string by passing it through
+# the MD5 algorithm and taking the first characters
+sub hashed_string {
+    my ($string, $count) = @_;
+    $count //= 5;
+
+    my $hash = md5_base64($string);
+    # plus sign is problematic in regexps
+    $hash =~ s,\+,_,g;
```
aaannz: If you would use `md5_hex` then there is no need for substitute. Is there a reason to use base64?

coolo: you get more bits in 5 base64 characters than in 5 hex characters.

coolo: 4 times as many actually - reducing the chance of a hash clash dramatically.

distribution.pm
```diff
@@ -104,7 +104,7 @@ sub script_run {
 
     testapi::type_string "$name";
     if ($wait > 0) {
-        my $str = bmwqemu::random_string(5);
+        my $str = bmwqemu::hashed_string("SR$name$wait");
```
okurz: You could generate the prefix with
```
(caller(0))[3] =~ s/([^:]*::\b|(?<=_))(\w)([^_]*_?)/\u$2/rg
```
… but this would be the road to madness, right?

coolo: KISS

okurz: I agree. it's ugly but I could not find a better way. However I would have suggested to put it into a well documented helper function anyway :-)
Don't the different "script" functions have too much duplicate code anyway?

PR #409, opened 2016-01-30
Enable strictures and warnings for all tests

aaannz:
Few times I noticed people doesn't import strict and warnings modules.
Especialy strictures are important for safe development so import them in
our basetest class.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -24,6 +24,12 @@ use JSON;
 use POSIX;
 use testapi ();
 
+# enable strictures and warnings in all tests globaly
+sub import {
+    strict->import;
+    warnings->import;
+}
+
 sub new {
     my ($class, $category) = @_;
     $category ||= 'unknown';


okurz:
good idea. Does this mean it will import these if not imported but no problem still if explicitly imported in each test?

aaannz:
Yes, if module is already imported perl will ignore another imports, so no problem if test has `use strict` itself.

Yes, perlcritic has a rule to demand strictures to be enabled. But I don't remember if it can be enabled selectively or only when using stricter ruleset.

OpenQA itself already has strictures enabled since Mojo use the same approach as in this PR. So everything inheriting any Mojo module already has strictures (unless import method is overloaded). I think we have explicit `use strict;` in all os-autoinst modules. About tests, I didn't test it, bud I'm for just to see it once we deploy.

okurz:
> About tests, I didn't test it, bud I'm for just to see it once we deploy.

yeah, I have seen to much stuff break lately which IMHO could be catched before deployment, i.e. earlier …

coolo:
so what *did* you test?

coolo:
Some of these are let's say false positives as e.g. we have shutdown/shutdown.pm and x11/shutdown.pm and os-autoinst loads both as "package shutdown". Unfortunately these warnings are very easy to ignore as they rush by in the log. Can we make them fatal for the test? otherwise we will still only see them after deployment.

```
Prototype mismatch: sub autoyast_reboot::run () vs none at /home/coolo/prod/os-autoinst-distri-opensuse//tests/autoyast/autoyast_reboot.pm line 38.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/autoyast/autoyast_reboot.pm line 20.
Subroutine create_qaset_config redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 15.
Subroutine test_run_list redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 22.
Subroutine junit_type redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 26.
Subroutine test_suite redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 30.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 35.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/qa_automation/qa_run.pm line 111.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/slenkins/login.pm line 20.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/slenkins/login.pm line 26.
Prototype mismatch: sub boot::run () vs none at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/boot.pm line 29.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/boot.pm line 20.
Prototype mismatch: sub boot::test_flags () vs none at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/boot.pm line 37.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/boot.pm line 31.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/login.pm line 20.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/login.pm line 44.
Useless use of string in void context at /home/coolo/prod/os-autoinst-distri-opensuse//tests/support_server/setup.pm line 89.
Prototype mismatch: sub shutdown::run: none vs () at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 82.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 23.
Prototype mismatch: sub shutdown::test_flags: none vs () at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 86.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 84.
Subroutine trigger_shutdown_gnome_button redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 18.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 23.
Subroutine test_flags redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/shutdown.pm line 84.
Subroutine start_firefox redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/firefox.pm line 15.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11/firefox.pm line 28.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11regressions/firefox/sle12/firefox_java.pm line 17.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/x11regressions/firefox/sle12/firefox_mhtml.pm line 17.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_bootloader.pm line 14.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_users.pm line 14.
Subroutine y2snapper_create_snapshot redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_snapper.pm line 19.
Subroutine y2snapper_select_snapshot redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_snapper.pm line 46.
Subroutine clean_and_quit redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_snapper.pm line 66.
Subroutine run redefined at /home/coolo/prod/os-autoinst-distri-opensuse//tests/yast2_ui/yast2_snapper.pm line 79.
```

okurz:
:+1: to what @coolo said

coolo:
So the only real warning triggered here is

```
Useless use of string in void context at tests/support_server/setup.pm line 89.
```

aaannz:
> so what did you test? 

Except os-autoinst itself I didn't test anything.

> Can we make them fatal for the test?

To clarify, you mean fatal for tests or fatal only when running `make check`? There is `FATAL` key which can trigger all or some warnings to be errors.

> So the only real warning triggered here is

That's good, at least we have tests in great shape already. :)

okurz:
what do you think about https://progress.opensuse.org/issues/10506, i.e. directly using http://search.cpan.org/~chromatic/Modern-Perl-1.20150127/lib/Modern/Perl.pm

coolo:
FATAL in make check on tests would do. It's anyway questionable if this should be enforced by os-autoinst or isn't rather a os-autoinst-tests-opensuse policy

aaannz:
I would enforce it the same way Mojolicious enforce it. Once you inherit from `basetest` you gotta follow the rules.

@okurz we would need to inherit from Modern::Perl to achieve the same effect. I'm not 100% against but nor for.

coolo:
I don't see the value of yet another perl module dependency.

okurz:
fine by me, thx for the explanation

aaannz:
I looked briefly into making warnings fatal during tests, but I can't find how to do it in pretty way. From perl command line I have only `-W` and `-w`, both only enabling warnings, but not making them errors. So if I want to make them fatal, I would need to i.e. rely on envvar in basetest::import function, which seem pretty ugly to me.

coolo:
I was thinking along this:

```
--- a/Makefile
+++ b/Makefile
@@ -21,3 +21,4 @@ testing" && exit 2)
        test -e tools || ln -s os-autoinst/tools .
        tools/tidy --check
        export PERL5LIB="../..:os-autoinst:lib:tests/installation:tests/x11:tests/qa_automation:$$PERL5LIB" ; for f in `git ls-files "*.pm" || find . -name \*.pm|grep -v /os-autoinst/` ; d
o perl -c $$f 2>&1 | grep -v " OK$$" && exit 2; done ; true
+       for f in `find tests -name \*.pm`; do perl -w -I lib -I os-autoinst -e "package livecdreboot; use strict; use warnings; require '$$f';"; done
```

okurz:
I recommend to put the call on each file into the for loop on the previous line, don't call find twice and also better use "git ls-files" for speedup.

aaannz:
Didn't thought about it this way. What about moving the find into the perl code so that we can catch even some potential inter-test conflicts? Like the same name etc.

okurz:
I don't know about the last one. That might be a bit too much for a first step. Maybe in another PR :-)

coolo:
Having two modules with the same name is fine as long as they are not in the same test. So you can't test this here unless you make a flow analysis of main.pm

aaannz:
Aha, that sounds confusing to me. `You can have same name, unless you call them in the same test` vs. `You can have same test name` vs. `You can not have same name`. The latter two we can check during compile phase. The first one easily only during runtime.

okurz:
@aaannz don't try to go to far with this PR, everything else can be tracked in https://progress.opensuse.org/issues/10514 and should not be lost :-)
I also suggest to make all test module names unambiguous or come up with a clear definition of "full module name" which is composed of component and module name itself

PR #410, opened 2016-01-30
Testapi documentation update

aaannz:
Some new and some updated documentation, I also moved things a little to be
in related parts to each other. It makes docu generation easier.

Because I moved some parts all around the place careful with merging no to cause some missing subroutines implementations. I need to run it on one worker to check all is fine. So for now just check the docu and language.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -82,15 +82,40 @@ sub init {
 
 ## no critic (ProhibitSubroutinePrototypes)
 
+=head2 set_distribution
+
+    set_distribution($distri);
+
+Set distribution object.
+
+You can use distribution object to implement distribution specific helpers.
+
+=cut
 sub set_distribution {
     ($distri) = @_;
     return $distri->init();
 }
 
+=head1 video output handling
+
+=head2 save_screenshot
+
+  save_screenshot;
+
+Saves screenshot of current SUT screen.
+
+=cut
 sub save_screenshot {
     return $autotest::current_test->take_screenshot;
 }
 
+=head2 record_soft_failure
+
+  record_soft_failure
+
+Current test module result is success, but workarounds were used.
+
+=cut
 sub record_soft_failure {
     bmwqemu::log_call('record_soft_failure');
     $autotest::current_test->{dents}++;
@@ -217,6 +242,16 @@ sub _check_or_assert {
     return;    # never reached
 }
 
+=head2 assert_screen
+
+  assert_screen($mustmatch [,$timeout]);
+
+Wait for needle with tag C<$mustmatch> to appear on SUT screen.
+C<$mustmatch> can be string or C<ARRAYREF> of string (C<['tag1', 'tag2']>).
+
+Returns matched needle or throws C<NeedleFailed> exception if $timeout timeout is hit. Default timeout is 30s.
+
+=cut
 sub assert_screen {
     my ($mustmatch, $timeout) = @_;
     $timeout //= $bmwqemu::default_timeout;
@@ -224,6 +259,16 @@ sub assert_screen {
     return _check_or_assert($mustmatch, $timeout, 0);
 }
 
+=head2 check_screen
+
+  check_screen($mustmatch [,$timeout]);
+
+Similar to C<assert_screen> but does not throw exceptions. Use this for optional matches.
+Check C<assert_screen> for parameters.
+
+Returns matched needle or C<undef> if timeout is hit. Default timeout is 30s.
+
+=cut
 sub check_screen {
     my ($mustmatch, $timeout) = @_;
     $timeout //= $bmwqemu::default_timeout;
@@ -231,6 +276,13 @@ sub check_screen {
     return _check_or_assert($mustmatch, $timeout, 1);
 }
 
+=head2 match_has_tag
+
+  match_has_tag($tag);
+
+Returns true if last matched needle has C<$tag> else return C<undef>.
+
+=cut
 sub match_has_tag {
     my ($tag) = @_;
     if ($last_matched_needle) {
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -224,6 +259,16 @@ sub assert_screen {
...
 }
 
+=head2 check_screen
+
+  check_screen($mustmatch [,$timeout]);
+
+Wait for needle with tag C<$mustmatch> to appear on SUT screen.
+C<$mustmatch> can be string or C<ARRAYREF> of strings (C<['tag1', 'tag2']>).
+
+Returns matched needle or C<undef> if timeout is hit. Default timeout is 30s.
```
okurz: I think also for a reader it is easier if the documentation is not "copy&paste and adapt" from assert_screen but referring to assert_screen, e.g. what do you think about:
```
Like C<assert_screen> but returns C<undef> if timeout is hit instead of throwing
an exception. Use this for optional matches.
```

testapi.pm
```diff
@@ -277,177 +333,33 @@ sub assert_and_click {
...
-  wait_screen_change { $code };
+  wait_screen_change { CODEREF [,$timeout] };
 
-wrapper around code that is supposed to change the screen. This is basically the
-opposite to wait_still_screen. Make sure to put the commands to change the screen
+Wrapper around code that is supposed to change the screen.
+This is the opposite to C<wait_still_screen>. Make sure to put the commands to change the screen
 within the block to avoid races between the action and the screen change
 
   wait_screen_change {
```
okurz: Prepend by "example:"?

testapi.pm
```diff
@@ -535,10 +448,9 @@ sub get_var {
 
   get_required_var($variable)
 
-Like get_var but without default. Fails when variable can not be retrieved.
+Returns content of test variable C<$variable> or throws exception if variable can not be retrieved.
```
okurz: I like the former, see my reasoning above regarding "copy vs. referencing"

testapi.pm
```diff
@@ -604,8 +510,62 @@ sub check_var_array {
     return grep { $_ eq $val } @$vars_r;
 }
 
-## helpers
+=head1 script execution helpers
+
+=head2 wait_serial
+
+  wait_serial($regex or ARRAYREF of $regexes [[, $timeout_sec], $expect_not_found]);
```
okurz: I prefer to name variables here same as the actual name, i.e. either say "$timeout" or rename variable within code to "$timeout_sec"

testapi.pm
```diff
@@ -604,8 +510,62 @@ sub check_var_array {
...
+    bmwqemu::fctres('wait_serial', "$regexp: $matched");
+    return $ret->{string} if ($matched eq "ok");
+    return;    # false
+}
+
+=head2 x11_start_program
 
+    x11_start_program($program[, $timeout, $options]);
+
+Start C<$program> when in graphical desktop environment.
```
okurz: I don't understand the "when" here, maybe just delete.

testapi.pm
```diff
@@ -614,16 +574,17 @@ sub x11_start_program {
 
 =head2 script_run
 
-  script_run($program, [$wait_seconds]);
+  script_run($program, [$wait]);
```
okurz: :+1:

testapi.pm
```diff
@@ -636,12 +597,12 @@ sub script_run {
 
   assert_script_run($command);
 
-run $command via script_run and die if it's exit status is not zero.
-The exit status is checked by via magic string on the serial port.
-So make sure not to tamper the serial output.
+Run C<$command> via C<script_run> and C<die> if it's exit status is not zero.
```
okurz: s/it's/its/

testapi.pm
```diff
@@ -674,14 +635,14 @@ sub script_sudo {
 
   assert_script_sudo($command[, $wait]);
 
-run $command via script_sudo and die if it's exit status is not zero.
-The exit status is checked by via magic string on the serial port.
-So make sure not to tamper the serial output.
+Run C<$command> via C<script_sudo> and C<die> if it's exit status is not zero.
```
okurz: s/it's/its/

okurz: Also, as the text looks confusingly similar to "assert_script_run" I would suggest to reference "assert_script_run" and "script_sudo" instead of repeating the content with only slight differences. This is to prevent a reader to fall into the trap of not finding the differences. Of course, the name should say it all already.

testapi.pm
```diff
@@ -1005,51 +1229,36 @@ tests to select
...
-console (select_console), no matter which console you send commands to
-here.
+Select the named console for further testapi interaction (send_text,
+send_key, wait_screen_change, ...)
 
-=back
+If this the first time, a test selects this console, the distribution
+will get a call into activate_console('root-console', $console_obj) to
+make sure to actually log in root. For the backend it's just a tty
+object (in this example) - so it will sure the console is active,
```
okurz: s/sure/ensure/

okurz:
-0.5

nice work, see some inline comments for improvements

aaannz:
Comments addressed in 071c0be. Once it is ok, I'll squash the commits and push.

okurz:
I didn't think of this approach to put "fixes" into its own commit for easier review of changes within one PR. It's only "dangerous" if coolo immediately merges it before you squash where we could end up with "ugly commits" in master :-/

Just one trivial typo encountered. Fix and auto-LGTM, i.e. if not waiting for comments by others, fix the typo, squash the fixes, push and merge right away :-)

aaannz:
Rebased & squashed. Waiting for travis.

aaannz:
Done a GNOME testrun,  checked for regressions and it appears it's ok.

PR #411, opened 2016-02-01
Set iso read-only

k0da:
We need to change access mode for ISO, in order to be able to
mount it multiple times.

### DIFF
--- a/backend/pvm.pm
+++ b/backend/pvm.pm
@@ -112,10 +112,12 @@ sub start_lpar {
     my $iso_present = qx/pvmctl media list -d VirtualOpticalMedia.media_name --where VirtualOpticalMedia.name=$iso/;
     if ($iso_present !~ /$iso/) {
         #copy over from nfs to VMLibrary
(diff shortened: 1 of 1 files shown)


okurz:
LGTM

backend/pvm.pm
```diff
@@ -112,10 +112,12 @@ sub start_lpar {
     my $iso_present = qx/pvmctl media list -d VirtualOpticalMedia.media_name --where VirtualOpticalMedia.name=$iso/;
     if ($iso_present !~ /$iso/) {
         #copy over from nfs to VMLibrary
-        my $cmd = "sudo viosvrcmd --id 1 -r -c\" cp $source_iso /var/vio/VMLibrary/$iso\"";
-        runcmd($cmd);
-        $cmd = "sudo viosvrcmd --id 1 -r -c\" chown padmin:staff /var/vio/VMLibrary/$iso\"";
-        runcmd($cmd);
+        my @viocmd = ("sudo viosvrcmd --id 1 -r -c\" cp $source_iso /var/vio/VMLibrary/$iso\"");
```
coolo: I remember again why I didn't merge it right away. I wondered why you put a space before the command here but not in the new call.

k0da:
Better now?

PR #413, opened 2016-02-03
Fix testapi parse_junit_log

jtzhao:
parse_junit_log uses upload_logs to upload junit files, but forgets to add current test name as a prefix to the file name, causing it fails to open the junit file.
issue: https://progress.opensuse.org/issues/10552

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -814,7 +814,7 @@ sub parse_junit_log {
 
     upload_logs($file);
 
-    $file = basename($file);
+    $file = ref($autotest::current_test) . '-' . basename($file);
 
     open my $fd, "<", "ulogs/$file";
     my $xml = join("", <$fd>);


nilxam:
hmm...you already knew the junit file path, is it possible you can just use it instead of open the junit file from /ulogs, and uploading the junit file in the end of run()?

aaannz:
Well until it is uploaded, the file is stored at SUT. So we need to upload it to get it to os-autoinst. But since `upload_logs` use `assert_script_run` we don't have to check it's return value. So I propose to further change `upload_logs` subroutine to return the new name. Or to return local path to the uploaded file.
So we will then use:
```perl
$file = upload_logs($file);
```

Upside is if we choose to change naming again we don't need to do that anywhere else.

nilxam:
oh, yeah, you're right, it did parsing out of SUT.

nilxam:
and I like the idea :+1:

jtzhao:
I agree. I can submit another to make upload_logs to return the uploaded file path.

coolo:
yeah, do that please

PR #414, opened 2016-02-03
Fail early on empty script_run

okurz:
If "script_run" is executed without content bash will fail on an invalid
command starting with "; ...". Better die early with explicit error message
about the test being written in a wrong way.

openqa_empty_script_run_fails

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -102,6 +102,7 @@ sub script_run {
     # start console application
     my ($self, $name, $wait) = @_;
 
+    die "script must be non-zero" unless $name;
     testapi::type_string "$name";
     if ($wait > 0) {
         my $str = bmwqemu::hashed_string("SR$name$wait");


coolo:
I respect your good intentions, but this is just Aktivisimus. script_run's argument *has* to be a valid bash script and you only test one possible misbehaviour. *any* syntax error is mortal to script_run

okurz:
"" IMHO can be considered a valid bash script but the unexpecting test writer might be stumbled that some stuff is added after this string within the testapi. That is why the testapi is not trying to check that the script is valid but only that it's assumption that something can be appended with ";" is valid as in "programming by contract".

sysrich:
I agree with @coolo - I don't want to see the backend full of cludges to catch bad test writing that can be immediately spotted in bad behaviour by those badly written tests.

okurz:
I think the "testapi" should help the test writer with explicit help even more so in this case where it is the backend adding the "random string" causing the "bad behaviour". Yes, we can not check for every stupid thing a test writer might be doing but that is not the point here. Every function should check for its assertions, that is what I meant with [programming by contract](https://en.wikipedia.org/wiki/Design_by_contract).

coolo:
again: if you can catch *every* invalid argument, I'm fine. As long as you only add random checks for random errors, it's noise.

PR #415, opened 2016-02-03
Pass format to do_extract_asstest

k0da:
In order to avoid code duplication in different backends, compute
format in isotovideo and send it to the backend

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -204,7 +204,9 @@ if (!$r && (my $nd = $bmwqemu::vars{NUMDISKS})) {
             $dir = 'assets_public';
         }
         next unless $name;
-        push @toextract, { hdd_num => $i, name => $name, dir => $dir };
+        $name =~ /\.([[:alnum:]]+)$/;
+        my $format = $1;
+        push @toextract, { hdd_num => $i, name => $name, dir => $dir, format => $format };
     }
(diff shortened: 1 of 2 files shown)


okurz:
looks ok to me but so far the benefit is not visible. I would be interested where you use it.

k0da:
This is a preparation for https://github.com/k0da/os-autoinst/commit/1dd527cfe0938b6fb805d71f03bbb79f8bde54b9

okurz:
ah, ok. thx. maybe you should correct the commit message and PR message to "do_extract_**assets**" :-)

k0da:
Woah.. Sure I will

isotovideo
```diff
@@ -204,7 +204,8 @@ if (!$r && (my $nd = $bmwqemu::vars{NUMDISKS})) {
             $dir = 'assets_public';
         }
         next unless $name;
-        push @toextract, { hdd_num => $i, name => $name, dir => $dir };
+        my ($name,$format) = split /\./,$name;
```
coolo: this is asking for trouble. the next one to put a . in its asset name will be F***Ked. the current regexp is much stricter

k0da: Ok, lets use the old regexp then

PR #417, opened 2016-02-05
enable virtio rng by default

lnussel:
virtio rng is a feature introduced in qemu 1.3 resp kernel 2.6.26
already. Let's use it to avoid entropy starvation.

https://progress.opensuse.org/issues/9584

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -461,6 +461,10 @@ sub start_qemu {
             push(@params, "-cpu", $vars->{QEMUCPU});
         }
 
+        if ($vars->{QEMU_VIRTIO_RNG}) {
+            push(@params, '-device', 'virtio-rng-pci');
+        }
+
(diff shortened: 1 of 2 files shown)


lnussel:
FWIW after some consideration I decided not to use a variable. presence of the device shouldn't hurt. If the guest kernel supports it it will just make use of it.

coolo:
I'm sure worried that we loose the capability to test stock qemu - and as such many issues our users will find, we will not.

coolo:
s,stock,default,

lnussel:
So always test without virtio rng like now and only let the fips test use the virtio rng?

coolo:
what's libvirt's default - do you know? I would like us to have test scenarios that test the defaults

lnussel:
there is an option to add an RNG. Depends on how one created the profile I guess. The ones from setupgrubfornfsinstall don't create a RNG at least :)

PR #418, opened 2016-02-05
Fix hashed_string not to return /

coolo:
I didn't expect / to create a problem, but that was short sightened. As we
the return value for script names, it requires subdirs - which we don't want

Referene: https://openqa.suse.de/tests/197862/modules/autoyast_verify/steps/3


okurz:
Extend the comment within the file (only mentions plus sign so far), then auto-LGTM

coolo:
changed the comment now

PR #419, opened 2016-02-05
support  for slepos images

nadvornik:
These changes are needed for the following scenario:
- job 1 builds a hdd image with kiwi and uploads it with upload_asset
- job 2 boots with the image on HDD_1, the image is resized during boot to fit the hdd size.


coolo:
I wonder why we hardcode qcow in one case and depend on HDDFORMAT in the other case - and how long it will take before someone requires HDDFORMAT_$i :(

nadvornik:
I suppose qcow is the only format that supports backing files.

PR #420, opened 2016-02-08
Pvm keephdd

k0da:
Extract HDD assets


coolo:
As this includes https://github.com/os-autoinst/os-autoinst/pull/415, my comment there applies

okurz:
+1

PR #421, opened 2016-02-09
Disable too strict stall detection again

okurz:
7653967c704b0812ee5f08103a06601405a9da05 introduced a timeout limit with
intentional crash on hit. As it looks like we are hitting this limit more
often than expected, still. Many tests would continue fine even though a stall
is detection somewhere in between.

If it is desired to put the intentional crash back on stall it should be made
optional so that we can enable it in "development" runs but just record and
continue in "productive" runs.

### DIFF
--- a/INSTALL.asciidoc
+++ b/INSTALL.asciidoc
@@ -16,9 +16,6 @@ cd os-autoinst
 3.  install the BuildRequires, as per the http://build.opensuse.org[spec
 file]
 +
------------------------------------------------------------------------------------------------------
-sudo zypper in libtheora-devel opencv-devel pkg-config libtool autoconf automake gcc-c++
------------------------------------------------------------------------------------------------------
 4.  build the beastie locally
 +
 ----------------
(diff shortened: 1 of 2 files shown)


coolo:
no, the die is there for a reason - so we are not bothered by failing tests, but that the test get rescheduled as incomplete

coolo:
And no, this is not for development - this *is* a production feature. If you want passing tests to continue, then implement what you suggested in the progress item: a stall flag that makes failures to crashes.

okurz:
Please read https://progress.opensuse.org/issues/10460#note-10
tests get retriggered but consistently fail at this point. This issue together with other recent improvements seriously hinders me to run tests locally and do my work. I am sure others might also struggle with the same issue but maybe don't see the reason.

I will try to help with "remember stall flag and crash if test fails later" but it takes more time.

coolo:
take your time

okurz:
:tired_face:

okurz:
obsoleted by #427

PR #423, opened 2016-02-11
Pass optional argument to record_soft_failure for reason

okurz:
If a reason is provided this is shown in the logfile and later on can also be
passed to basetest::record_serialresult.


coolo:
at least update the inline docu

okurz:
updated:
* Infile doc for record_soft_failure extended
* Use Test::Output in unit tests to also check output

aaannz:
:+1:

PR #425, opened 2016-02-12
make script_output work as documented

nadvornik:
check real exit code, not the fact that the script reached last line
- exit 0 now works
- failures are detected before timeout


aaannz:
:+1:

okurz:
good idea in general, just small issues:

* Please start the git commit message with a capital letter
* Can you tell which tests are affected or where tests failed where this should help?

PR #427, opened 2016-02-16
Make stall detection a little less annoying

coolo:
Instead of dying right away, only die if a check_screen would fail due to it

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -44,7 +44,7 @@ __PACKAGE__->mk_accessors(
       last_screenshot _last_screenshot_name last_image
       reference_screenshot interactive_mode
       assert_screen_tags assert_screen_needles assert_screen_deadline
-      assert_screen_fails assert_screen_last_check
+      assert_screen_fails assert_screen_last_check stall_detected
       ));
 
 sub new {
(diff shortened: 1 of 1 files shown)


coolo:
Doing https://github.com/os-autoinst/os-autoinst/pull/421 properly

backend/baseclass.pm
```diff
@@ -806,6 +807,10 @@ sub check_asserted_screen {
             $self->freeze_vm();
             return {waiting_for_needle => 1, filename => $self->write_img($img, $img_filename), candidates => $failed_candidates};
         }
+        if ($self->stall_detected) {
+            backend::baseclass::write_crash_file();
+            bmwqemu::mydie "assert_screen fails, but we detected a timeout in the process, so we abort";
```
aaannz: This code is called by check_screen too. I imagine it can be misleading for some to report always as assert_screen.
How about using `check_asserted_screen fails and we ...` as in warning above;

okurz:
fix the message as proposed by aaannz then auto-LGTM

backend/baseclass.pm
coolo: I can leave out the function name completely too

PR #428, opened 2016-02-16
Exclude particular tests by variable

nadvornik:
This will be useful for example on this test:
http://openqa.suse.de/tests/203975/modules/grub_test/steps/3
This is a really minimal installation, so clone does not work because of missing packages and grub is text-only.
So I will set EXCLUDE_MODULES=clone,grub_test

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -6,6 +6,7 @@ Supported variables per backend
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
 Variable;Values allowed;Default value;Explanation
+EXCLUDE_MODULES;string;;comma separated names or fullnames of excluded test modules
 NOVIDEO;boolean;0;Do not encode video if set
 SCREENSHOTINTERVAL;;;
 |====================
--- a/basetest.pm
+++ b/basetest.pm
@@ -66,6 +66,12 @@ Can eg. check vars{BIGTEST}, vars{LIVETEST}
 =cut
 
 sub is_applicable {
+    my ($self) = @_;
+    my %excluded = map { $_ => 1 } split(/\s*,\s*/, $bmwqemu::vars{EXCLUDE_MODULES});
+
+    return 0 if $excluded{$self->{class}};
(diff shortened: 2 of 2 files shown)


coolo:
I don't like this at all. There is already main.pm doing logic and there is is_applicable in the tests itself - and now you add a third way - that only you know about. It might be a quick fix for your very problem, but I think this is wrong.

okurz:
would it make sense to do something like "is_not_applicable" as the opposite of "is_applicable"?

nadvornik:
The behavior of the system is given by AUTOYAST variable, so the fine-tuning of the tests should be also in a variable. It IMHO does not belong to main.pm nor the modules.

nadvornik:
is_applicable / is_not_applicable does not help here. I don't think that we want to put a blacklist of jobs to each test module.

sysrich:
I agree with coolo, I don't like this idea at all. What is wrong with putting the logic in the main.pm?

nadvornik:
I moved it to basetest.pm

nadvornik:
In general, I don't think that all the exceptions belong to main.pm. Main.pm is for describing the general workflow and order of test.
The particular test parameters are defined via variables, main.pm should not know about all possible combinations. So if one particular module does not work with certain combination of variables, it should be disabled via variable.

Also, it separates admin role (setting variables) and test developer role (editing main.pm or modules)

okurz:
approach looks ok to me as I can not propose something better for now. If others agree this is the way to go please extend the in-file doc on "is_applicable" accordingly

coolo:
And add to doc/backend_* something meaningful

doc/backend_vars.asciidoc
```diff
@@ -6,6 +6,7 @@ Supported variables per backend
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
 Variable;Values allowed;Default value;Explanation
+EXCLUDE_MODULES;comma separated names;;names or fullnames of excluded test modules
```
coolo: I think the 2nd column should read "string" or "list"

PR #429, opened 2016-02-17
Use agent auth instead there is no password given

coolo:
We need to support password login to get into the installed system, but
to get into the libvirt host, we want to avoid passwords in the settings


coolo:
there is one catch I didn't expect when trying to use pubkey: openqa_worker's home is /var/lib/empty

coolo:
so we can't really use this to avoid passwords in config

PR #430, opened 2016-02-17
assert_script_run: Return optional custom message on die

okurz:
See
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/967#discussion_r50756596

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -52,5 +52,6 @@ requires 'warnings';
 on 'test' => sub {
   requires 'Perl::Critic';
   requires 'Test::Output';
+  requires 'Test::Fatal';
 }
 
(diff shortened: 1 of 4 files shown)


okurz:
@aaannz as suggested by you

aaannz:
Did you want to include snippets of testapi mocking in this commit? Seems out of topic to me.

okurz:
> Seems out of topic to me.

This is what I came up with to actually cover "assert_script_run" from the tests. I does look like a lot of changes. The reason is that the "test driver" was not doing much so far and the "die" exception throwing was not tested so far. I decided to put the mocking just where it is needed but we can also move it into the test driver itself later on if we share it in more tests.

PR #431, opened 2016-02-22
Modify the PR 415 back, since it create hdd as *.qcow2.qcow2 name

SeroSun:
By modify from
https://github.com/os-autoinst/os-autoinst/pull/415/files
the new created hdd will named as *.qcow2.qcow2, it will make the testscript keep use the old hdd file as the new build's hdd named as *.qcow2.qcow2 which not configured in testcase. So I suggest to modify the code back.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -186,14 +186,15 @@ sub do_extract_assets {
     my $hdd_num = $args->{hdd_num};
     my $name    = $args->{name};
     my $img_dir = $args->{dir};
-    my $format  = $args->{format};
+    $name =~ /\.([[:alnum:]]+)$/;
+    my $format  = $1;
     if (!$format || $format !~ /^(raw|qcow2)$/) {
         bmwqemu::diag "do_extract_assets: only raw and qcow2 formats supported $name $format\n";
     }
     elsif (-f "raid/l$hdd_num") {
-        bmwqemu::diag "preparing hdd $hdd_num for upload as $name in $format\n";
+        bmwqemu::diag "preparing hdd $hdd_num for upload as $name\n";
         mkpath($img_dir);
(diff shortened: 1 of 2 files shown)


okurz:
+1

nilxam:
I think you don't need to change isotovideo(ie. $format still passing to do_extract_assets) but do not append the $format following with the $name, I don't know whether without $format sets in pvm backend  will break the use case from @k0da since https://github.com/os-autoinst/os-autoinst/commit/43a16a07b4644bf42695c497f30bff39566a1472 , but at least passing format to backend should not hurt.

aaannz:
-1 as @nilxam said. Just modify `qemu.pm` no to append `$format` suffix.

SeroSun:
@nilxam @aaannz I agree :) 
I'll update the isotovideo code changes back.

backend/qemu.pm
```diff
@@ -186,14 +186,15 @@ sub do_extract_assets {
     my $hdd_num = $args->{hdd_num};
     my $name    = $args->{name};
     my $img_dir = $args->{dir};
-    my $format  = $args->{format};
+    $name =~ /\.([[:alnum:]]+)$/;
+    my $format = $1;
```
coolo: this is now very confusing. You strip the format off $name - and then also remove the code that appends it

coolo:
https://github.com/os-autoinst/os-autoinst/pull/432 is my take on this

coolo:
merged that one - but we need to redeploy soonish

backend/qemu.pm
SeroSun: Hi @coolo, this binding operation is to let $1 get format in $name, and $name still have format. The code just look like what I modified before PR 415. So I got the $format in that way is because I'm not sure if getting format from $args have problem or not (since it didn't use this way before..)

PR #432, opened 2016-02-22
Do not append $format to $name - it already contains it

coolo:


aaannz:
:+1:

SeroSun:
Hi coolo, when will this merge deploy in openqa.suse.de?

coolo:
I would prefer soonish - but it's up to richard

PR #433, opened 2016-02-22
Coolos gadget backend

coolo:
This PR is to document my process - and most importantly to find a name for this :)

'gadget' is the nick name of this thingy, but we need a better one.

Everything with KVM in the name is very misleading and I don't think we should hard code the name of the company or the current device into it - hopefully there are more than that.

The next best thing I can think of is 'remotevnc', but possibly you have other ideas - so I left the nick name for now.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,6 +13,7 @@ backendexec_DATA = \
 	backend/baseclass.pm \
 	backend/console_proxy.pm \
 	backend/driver.pm \
+	backend/generalhw.pm \
 	backend/ipmi.pm \
 	backend/qemu.pm \
 	backend/s390x.pm \
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -251,7 +251,7 @@ sub post_start_hook {
     my ($self) = @_;
 
     # ignored by default
-    return 0;
+    return {};
 }
 
 sub start_vm {
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -351,7 +351,9 @@ sub cpu_stat {
 sub write_img {
     my ($self, $image, $filename) = @_;
 
-    if (!-f $filename) {
+    return if (!$image);
+
+    if ($filename && !-f $filename) {
         $image->write($filename) || return;
     }
     return $filename;
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -1,0 +1,155 @@
+# Copyright © 2016 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+# this backend uses a KVM connector speaking VNC and external tools
+# for serial line and power cycling
+
+package backend::generalhw;
+use strict;
+use base ('backend::baseclass');
+use bmwqemu qw(diag);
+use testapi qw(get_required_var get_var);
+use IPC::Run ();
+require IPC::System::Simple;
+use autodie qw(:all);
+use File::Basename qw(basename);
+
+sub new {
+    my $class = shift;
+    # required for the tests to access our HTTP port
+    get_required_var('WORKER_HOSTNAME');
+    return $class->SUPER::new;
+}
+
+sub get_cmd {
+    my ($self, $cmd) = @_;
+
+    my $dir = get_required_var('GENERAL_HW_CMD_DIR');
+    if (!-d $dir) {
+        die "GENERAL_HW_CMD_DIR is not pointing to a directory";
+    }
+    $cmd = get_required_var($cmd);
+    $cmd = "$dir/" . basename($cmd);
+    if (!-x $cmd) {
+        die "CMD $cmd is not an executable";
+    }
+    return $cmd;
+}
+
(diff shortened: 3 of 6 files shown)


aaannz:
I guess the VNC IP address and SNMP OIDs will be later moved to variables (which brings the side question if we shouldn't keep our internal layout to ourselves and keep the pleasure of discovery to our potential intruders). I also noticed one IPMI copy-paste diag message ;) But about that try::tiny usage, any plans to exceptionalize this?

okurz:
Isn't "gadget" something like a "backend proxy"? Someone has to update the picture in https://github.com/os-autoinst/openQA/blob/master/docs/GettingStarted.asciidoc#architecture btw

coolo:
a "backend proxy"? I don't see where it's a proxy. And yes, the architecture picture is not only old, it's also simplified - which is ok for architecture overviews IMO

coolo:
http://stephan.kulow.org/gadget.ogv - we miss the grub due to the slowness of the gadget ;(

coolo:
So the code is ready - if we can live with the name we could merge :)

backend/gadget.pm
```diff
@@ -0,0 +1,139 @@
...
+use base ('backend::baseclass');
+use bmwqemu qw(diag);
+use testapi qw(get_required_var get_var);
+use IPC::Run ();
+require IPC::System::Simple;
+use autodie qw(:all);
+
+sub new {
+    my $class = shift;
+    get_required_var('WORKER_HOSTNAME');
```
aaannz: Why is this here? `WORKER_HOSTNAME` is not used here so maybe in base class, but is this a leftover or actual check that this var is present before usage?

backend/gadget.pm
```diff
@@ -0,0 +1,139 @@
...
+    chomp $stderr;
+
+    die join(' ', @cmd) . ": $stderr" unless ($ret);
+    bmwqemu::diag("IPMI: $stdout");
+    return $stdout;
+}
+
+sub poweroff_host {
+    my ($self) = @_;
+    $self->run_cmd(get_required_var('GADGET_POWEROFF_CMD'));
```
aaannz: Ok, this rises a red flag for me. Do we really want to execute arbitrary command someone put into those variables? Since this is designed for specific device I would keep the snmp command there and supply only domain and possibly OID.
Thinking about that, OID will be probably registered so universally known so that may be hardcoded I guess.

backend/gadget.pm
```diff
@@ -0,0 +1,139 @@
...
+sub start_serial_grab {
+    my ($self) = @_;
+
+    $self->{serialpid} = fork();
+    if ($self->{serialpid} == 0) {
+        setpgrp 0, 0;
+        open(my $serial, '>',  $self->{serialfile});
+        open(STDOUT,     ">&", $serial);
+        open(STDERR,     ">&", $serial);
+        exec(get_required_var('GADGET_SOL_CMD'));
```
aaannz: Same here, some sanitation or command white-listing, pretty please.

backend/gadget.pm
coolo: no, this is not only snmp. this can be a ipmitool command or a snmp or a vendor specific command or ... what are we protecting against?

backend/gadget.pm
coolo: I have to leave that to @sysrich , as I don't know what possible commands we possibly have to take care of.

backend/gadget.pm
aaannz: We are protecting against scenario, where someone's compromised credentials are used by adversary to access openQA webui, set custom commands to these variables and through worker attack internal network.
If there is any weakness in webui, one doesn't even have to get anyone operators credentials.

aaannz: So I'm either for creating whitelist of allowed commands, or more annoying, I guess, but IMO secure, put these commands in worker.ini and load them from there.

backend/gadget.pm
coolo: upload_logs needs it for !qemu backends - and to avoid very late suprises, we check the existance of the variable early in these backends. you will see the same in svirt and ipmi

backend/gadget.pm
coolo: os-autoinst reading workers.ini? os-autoinst only gets the vars.json - it would be up to the worker to discard everything it got from the webui for *_CMD. Is that what you want?

aaannz: Yes, something like that. Worker discard specific variables from webui and replace them by commands read from workers.ini.

aaannz: In that case of course, this PR is fine. :)

coolo: Fine with me, but we need another solution to have one worker slot with multiple variable sets. Imagine one SLED maintenance patch we need to test on multiple graphic cards. It would be a pitty to reserve 20 worker slots just in case - still you only want one gadget to be accessed by one worker at a time.

So possibly support @MACHINE slots in workers.ini? 

```
[1]
WORKER_CLASS=sled_gadgets

[1@nvidia]
GADGET_VNC_IP...

[1@amd]
GADGET_VNC_IP...

coolo:
As I'm not fit enough to fix it now, I should at least write down the summary of the discussion:

instead of putting free command strings in the config file, we will require the worker admin to put a script in a directory and reference that script's name. E.g. put the snmp command in /etc/openqa/worker_scripts/poweroff_gadgetA.sh and reference POWEROFF_CMD=poweroff_gadgetA.sh in the workers.ini. os-autoinst will only filenames in a given directory. 

So far the discussion. Now I wonder if it's legit to call scripts in /etc and/or if we should hardcode /etc/openqa in os-autoinst. Both feel wrong, possibly we can pass that directory from worker to os-autoinst and only have a @prefix@/scripts default in os-autoinst ?

coolo:
Leaves the backend name. I'm kind of attached to the name 'gadget' by now - 'remotevnc' sounds politically correct in the negative sense ,)

coolo:
I have some more suggestions:
- phriliria
- tokturork
- throndrson
- hardeknud

coolo:
I'm done here - please review

okurz:
LGTM

PR #434, opened 2016-02-23
Block test execution on mutex_unlock

aaannz:
Unlocking mutex that does not exists or is locked now results in
waiting until the lock is available, parent's job finish or test timeout.

Also introduce `mutex_try_unlock` subroutine to try unlocking without
blocking. Returns true on successful unlock, false otherwise.

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -20,7 +20,7 @@ use strict;
 use warnings;
 
 use base qw/Exporter/;
-our @EXPORT = qw/mutex_create mutex_lock mutex_unlock mutex_try_lock/;
+our @EXPORT = qw/mutex_create mutex_lock mutex_unlock mutex_try_lock mutex_try_unlock/;
 
 require bmwqemu;
 use mmapi qw/api_call/;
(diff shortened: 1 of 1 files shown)


aaannz:
I want to add some tests to this, but it grew little complex and started to take some time so I'll add them in another PR. I don't want to make @sysrich and @dzyuzin waiting. :)

okurz:
LGTM Because you know what you are doing and others will complain if I ask for tests right now :-/ But you kept the coverage limit :-)

dzyuzin:
Why the behaviour of lockAPI is different from usual semaphores?
E.g. pthreads: pthread_mutex_unlock never locks, it either returns success, or fail (if the mutex not locked/locked by other thread).
That's really confusing.

aaannz:
You are right.  I misread the POSIX pages, I thought I was reading about pthread_mutex_unlock. when I was reading about pthread_mutex_lock. O.o . Without this PR is does follow the pthread_mutex_unlock. So this is PR is bunch of bollocks, disregard.

PR #435, opened 2016-02-23
Change size of default qcow image for z/KVM

Soulofdestiny:
Increase size of created qcow image for z/KVM tests because
4G is too small for e.g. allpatterns test


okurz:
shouldn't the allpattern test set this then? I don't think it makes sense to increase the default in the backend as 4G should be way enough for simple minimal images without explicit requirements specified.

coolo:
... what Oliver said

PR #437, opened 2016-02-25
Use ipxe kernel to support iBFT

dzedro:


okurz:
LGTM

aaannz:
Please update `doc/backend_vars.asciidoc`.

dzedro:
Sorry, didn't know
Updated

aaannz:
I was about to merge this, but noticed [WIP] flag. Still valid?

dzedro:
It is valid, but I found one problem, have to fix it and maybe it will change also this PR

coolo:
I don't see the relation between IBFT and PXE. To me this is like if $vars{FOO} do_bar.

dzedro:
Why relation to PXE ? I mentioned ipxe kernel http://ipxe.org/

coolo:
you load kernel for PXE if IBFT is set - and this connection is only in your head and is very specifc to the scenario you're working on. os-autoinst should be driven to load PXE kernel if e.g. IPXE is set or alike

aaannz:
I guess iPXE kernel is used here to support iSCSI boot. But I agree that as iPXE is quite versatile specific `IPXE` variable makes sense to me.

dzedro:
true, will change that variable name, but IPXE is already used, no idea why is PXE named as IPXE
[https://github.com/dzedro/os-autoinst-distri-opensuse/blob/master/tests/installation/bootloader.pm#L23](url)

dzedro:
I need help, I need to push _-kernel /usr/share/qemu/ipxe.lkrn -append 'dhcp && sanhook --drive 0x80 iscsi:10.100.98.90::3260:1:iqn.2016-01.openqa.de:for.openqa'_  to make IPXE/IBFT work.
for example, if I write it like this:
`        if ($vars->{IIPXE}) {`
 `           push(@params, '-kernel', '/usr/share/qemu/ipxe.lkrn');`
`            push(@params, '-append', "\'config", '&&', 'sanhook', '--drive', '0x80', "$vars->{ISCSI}" . "\'");`
`        }`
Job ends with error:
`12:20:26.4359 QEMU: qemu-system-x86_64: -append 'config: could not open disk image &&: Could not open '&&': No such file or directory`
[http://10.100.98.90/tests/1475/file/autoinst-log.txt](url)

The problem are the quotes I would say, if I try to push just `-append dhcp` it works
, but pushing multiple commands with `&&` or `;` is problem.

Any idea ?

coolo:
the argment to append needs to be in one string.

```
push(@params, '-append', "'config && sanhook --drive 0x80 $vars->{ISCSI}'");

dzedro:
@coolo true :+1: , it's working

dzedro:
I would leave the IBFT variable, I don't see any other test than iBFT test which would use this IPXE kernel and IPXE variable is already used
Any other ideas for variable names are welcome

dzedro:
Or use `NBF;boolean;0;open source network boot firmware e.g. to attach iscsi target on boot http://ipxe.org/` ?

coolo:
you can do much more with the ipxe kernel than just doing iscsi - http://ipxe.org/start

So +1 for using NBF and document iscsi as example

dzedro:
updated with NBF variable

PR #438, opened 2016-02-26
Increase default RAM to 2048

lnussel:
1G RAM is rather low nowadays and already fails when installing with
online updates:
https://openqa.opensuse.org/tests/124068/modules/setup_online_repos/steps/2

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -458,7 +458,7 @@ sub start_qemu {
         $SIG{__DIE__} = undef;    # overwrite the default - just exit
         my @params = ("-serial", "file:serial0", "-soundhw", "ac97", "-global", "isa-fdc.driveA=", @vgaoptions);
 
-        push(@params, '-m', $vars->{QEMURAM} || '1024');
+        push(@params, '-m', $vars->{QEMURAM} || '2048');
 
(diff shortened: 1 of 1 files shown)


lnussel:
The other options would be to reconfigure all machine settings to a new default or add extra machines with more ram obviously.

sysrich:
Given the currently published minimum requirements of openSUSE (https://en.opensuse.org/Hardware_requirements) SLE (https://www.suse.com/products/server/technical-information/) and Fedora (https://docs.fedoraproject.org/en-US/Fedora/21/html/Release_Notes/sect-Release_Notes-Welcome_to_Fedora_.html) are all 1GB or less I'm against changing the default in os-autoinst, which will affect every operating system

okurz:
-1

I think this should be adjusted the product requirements as stated, i.e. what does SLE tell what it works on as "bare minimum" because this is what we check.

"os-autoinst" is very much low the chain and appliance images and such should also work with 1GB, right?

coolo:
It's ok that we do not support installation with online updates on a default machine - but schedule it on a 64bit-largemem. I'm adding another veto here

lnussel:
ok, will add more machines to openqa.o.o then

PR #439, opened 2016-02-26
Fix continue waitforneedle does not work

nilxam:
It's a regression causes worker stay in waiting_for_needle block, ie.
freezing VM and checking the same tag, won't quit  _check_or_assert()


nilxam:
require https://github.com/os-autoinst/openQA/pull/579

nilxam:
I hope I have a better solution what do not need to create another control file, but I failed, any suggestions are welcome.

nilxam:
Updated PR, this one more simple and do not need to require another control file.

okurz:
+1

nilxam:
Updated again with more fixes...

okurz:
+1, still

Issue #341, opened 2015-11-26
[doc] add note that vars.json can be overwritten by test runs -> should use VCS or local copy


okurz:

coolo:
we can have a HACKING.md wich such tips

okurz:
Normally when someone tells "we can" it means "you should do it", but coming from you that normally means "I have already done it it and you will see the PR merged before you even finished writing *this* comment" ;-)

coolo:
your set of rules lacks "coolo + documentation => /dev/null" ;)

okurz:
Fixed with #440

PR #440, opened 2016-03-01
Add notice to README that vars.json can be overwritten

okurz:
Fixes github #341.


aaannz:
Can you move the text few lines up to L45? Before instructing user to start isotovideo?

PR #445, opened 2016-03-05
increase script_output to 30 s

MalloZup:
instead of 10 seconds, make default 30s.


sysrich:
:+1: 

While the problematic tests COULD be fixed by just using the timeout variable we have in script_output, the fact is script_outputs timeout of 10 seconds is an anomaly when compared to assert_script_run, script_run, and assert_screen which all use a timeout of 30s

MalloZup:
ok . we can merge this, and see how the snapper_undochange behaviour is with 30 seconds. by the way yes, 30 s should as default timeout.

coolo:
agreed to all what was said :)

okurz:
if all should have the same timeout then I propose to refactor to use one constant or configuration variable where all derive from

PR #447, opened 2016-03-09
Don't crash on empty VNC updates

coolo:


coolo:
I see you guys staring blank, so I'll self merge :)

PR #448, opened 2016-03-14
Fix spelling (catched -> caught)

DimStar77:


coolo:
it's just debug - better remove the line

PR #451, opened 2016-03-15
Run icewm-lite so xterm -fullscreen can work

coolo:
The window manager not only makes sure xterm gets the full screen, it also
makes sure that mouse_hide is not stealing focus from the xterm window, so
we can continue typing


Soulofdestiny:
fullscreen is working, i'm right now adjusting the needles for s390

Issue #426, opened 2016-02-15
Reporting screen match with needle that is not directly in `needles/` directory fails


garretraziel:
Although openQA can (and will) load needles from subdirectories of `needles/` recursively and os-autoinst does match against that needles, writing information about matched needle fails.

At the start of openQA log, there is:

    init needles from /var/lib/openqa/share/tests/fedora/needles
    loaded 272 needles

also, test then starts without problem and shows:

    MATCH(bootloader_uefi:0.00)
    no match 19
    MATCH(bootloader_uefi:0.00)
    no match 18

but after match is found, it fails with:

    MATCH(bootloader_uefi:1.00)
    Can't find /var/lib/openqa/share/tests/fedora/needles/bootloader_uefi.png at /usr/libexec/os-autoinst/needle.pm line 89.
    test _boot_to_anaconda died: Can't call method "has_property" on unblessed reference at /usr/libexec/os-autoinst/basetest.pm line 113.

It can't find `/var/lib/openqa/share/tests/fedora/needles/bootloader_uefi.png` because said needle is in `/var/lib/openqa/share/tests/fedora/needles/console/bootloader_uefi.png`.

It happens in openQA 4.3 with os-autoinst 4.3. Full log is [here](http://fpaste.org/321804/28886814/).

okurz:
So I recommend to change the code as follows:
1. Make openQA read non-recursively
2. Optional: Extend openQA to read from recursive folder structure, which would include the at least following features
 - Load(init) needles from needles directory recursively
 - Extend needle editor to also write in subdirectories
 - … and create dirs if not existing
 - how to handle old test results when the needle paths change?
 - adapt documentation

IHMO this is a feature request which is not an easy task to do right now, better create a ticket here: https://progress.opensuse.org/projects/openqav3/issues/new

okurz:
@garretraziel after #452 is done can you comment the impact on the current issue?

garretraziel:
#452 indeed fixes this problem, although it doesn't implement all necessary functionality you've talked about (it doesn't create directories nor it enables user to create needles in directories, it merely enables you to load from them), but it's good enough for this bug.

PR #452, opened 2016-03-16
Foundations for directory nesting for needles

rajcze:
This is a part of implementation (together with a patch for OpenQA <https://github.com/os-autoinst/openQA/pull/613>) that adds the foundations for directory nesting of needles. The problem is initially described in this issue: https://github.com/os-autoinst/os-autoinst/issues/426

Overall, the implementation strives to make backward-compatible changes, so although new testruns will store additional information, there should not be errors in viewing older results.

This patch adds a `file` item to the result's matchfile, which contains path to the respective needle's json file definition. This item is then used in the OpenQA's needle editor to provide the desired functionality.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -121,6 +121,7 @@ sub record_screenmatch {
         screenshot => $self->next_resultname('png'),
         result     => 'ok',
         properties => [@$properties],
+        file       => $h->{file},
     };
 
     # When the needle has the workaround property,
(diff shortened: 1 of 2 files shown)


basetest.pm
```diff
@@ -171,8 +172,9 @@ sub _serialize_match {
     my $diffcount = 0;
 
     my $name = $cand->{needle}->{name};
+    my $file = $cand->{needle}->{file};
 
-    my $h = {name => $name, error => $cand->{error}, area => []};
+    my $h = {name => $name, error => $cand->{error}, area => []}, file => $file;
```
aaannz: This is wrong. Move the `}` to the end.

aaannz:
LGTM

PR #453, opened 2016-03-17
New s390 serial

coolo:
This is trying to make wait_serial more reliable on zVM


sysrich:
@Soulofdestiny - what do you think?

coolo:
works now - together with the tests changes

PR #455, opened 2016-03-22
Implements ssh connections to SUT

Soulofdestiny:
For now, we need two ssh connections to the SUT
one for starting a getty and a second for starting iucvconn

removed stop_serial_grab, we don't have that implemented in s390 backend

### DIFF
--- a/backend/s390x.pm
+++ b/backend/s390x.pm
@@ -56,8 +56,6 @@ sub do_start_vm {
 sub do_stop_vm {
     my ($self) = @_;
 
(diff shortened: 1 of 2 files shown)


coolo:
you can't run that getty in the background on the same ssh?

Soulofdestiny:
it works manually... but somehow  exec->chan can't handle the '&', i will refactor this sometime

coolo:
yeah, exec only takes one command

PR #456, opened 2016-03-22
Add VNC to a libvirt-based VM

mnowaksuse:
Port 5901 is hardcoded as is it in `lib/susedistribution.pm`. VNC is set to listen on All interfaces (i.e. 0.0.0.0) otherwise we are prevented from connecting the VNC session. "force-shared" is used with qemu backend as well.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -173,6 +173,30 @@ sub add_pty {
     return;
 }
 
+# this is an equivalent of QEMU's '-vnc' option for tests where we watch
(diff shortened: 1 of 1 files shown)


coolo:
Hardcoding the port will not work - you limit yourself to one instance on the libvirt host. So better be prepared

coolo:
And call ./tools/tidy before commit

coolo:
we're hardcoding it in susedistribution atm because there we talk with a SUT and yast hardcodes this port. But that's not your problem, you have free choice.

PR #457, opened 2016-03-23
Some fixes for the slenkins tests

coolo:
- ifup the tap device as we set it up
- make sure the openvswitch service starts before firewall
- also block the minion to mess with us before this is done

### DIFF
--- a/os-autoinst-openvswitch
+++ b/os-autoinst-openvswitch
@@ -22,6 +22,10 @@ sub init_switch {
     $self->{BRIDGE} = $ENV{OS_AUTOINST_USE_BRIDGE};
     $self->{BRIDGE} //= 'br0';
 
+    ## coolo and richard terrible hack to make bridge
+    system("/usr/bin/ovs-vsctl add-br $self->{BRIDGE}");
+    system("/usr/sbin/ifup $self->{BRIDGE}");
+
     #the bridge must be already created and configured
     system('ovs-vsctl', 'br-exists', $self->{BRIDGE});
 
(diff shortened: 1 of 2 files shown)


coolo:
th whole bootprocess is horrible between wicked and openvswitch and our stuff, but we came up with a rather crude fix to setup the bridge and the tap devices all in our init script and do it after wicked but before firewall. This seems to work rather reliable actually

os-autoinst-openvswitch
```diff
@@ -22,6 +22,10 @@ sub init_switch {
     $self->{BRIDGE} = $ENV{OS_AUTOINST_USE_BRIDGE};
     $self->{BRIDGE} //= 'br0';
 
+    ## coolo and richard terrible hack to make bridge
+    system("/usr/bin/ovs-vsctl add-br $self->{BRIDGE}");
+    system("/usr/sbin/ifup $self->{BRIDGE}");
+
```
nadvornik: The bridge configuration is persistent, so I'd expect that this fails on second reboot. Or is the bridge deleted somewhere?

os-autoinst-openvswitch
```diff
@@ -33,6 +37,10 @@ sub init_switch {
     die "can't parse bridge local port MAC" unless $self->{MAC};
     die "can't parse bridge local port IP"  unless $self->{IP};
 
+    ## coolo and richard terrible hack to make tap devices
+    for my $i (0..10) {
+      system("ovs-vsctl add-port $self->{BRIDGE} tap$i tag=999");
+    }
```
nadvornik: We will need more tap devices to support multiple networks: tap64, tap65, ... and tap128, tap129...
See https://github.com/os-autoinst/os-autoinst/blob/master/backend/qemu.pm#L366
This is probably a good opportunity to add them.

os-autoinst-openvswitch
coolo: yes, as we learned - wicked resets the bridge config ;(

os-autoinst-openvswitch
coolo: indeed - teamd suite fails right now as it expects 133.

This PR is mainly to document what we applied - it feels rather dirty

sysrich:
DO NOT MERGE, @nadvornik is on the case, proposed proper solution inbound to the workers

PR #458, opened 2016-03-23
Remove hardcoded worker IP from NBF variable

dzedro:
http://10/100.98.90/tests/1568
@sysrich salt all workers with iSCSI target for iBFT ^^
btw right now it's broken, openqaworker3 has worker class qemu_x86_64_ipxe_firmware instead of openqaworker2 and openqaworker2 has /usr/share/qemu/ipxe.lkrn instead of openqaworker3


coolo:
hmm, they are supposed to have it all

sysrich:
all workers now have the iscsi config, including the lkrn, so we don't NEED this..though it's nice to have

PR #459, opened 2016-03-24
Add emulator as Xen HVM won't add it automatically

mnowaksuse:
KVM gets emulator added automatically, Xen HVM does not.


coolo:
merged manually as you didn't rebase

PR #460, opened 2016-03-31
openvswitch service fixes

nadvornik:
ip link set $tap up is one remaining change from https://github.com/os-autoinst/os-autoinst/pull/457 which might still be useful.

Other changes fixes the boot order with the new openvswitch which starts before wicked.


aaannz:
:+1:

PR #461, opened 2016-03-31
Add TAPDOWNSCRIPT for tap networks

dzyuzin:
Sometimes (in case of using tapscript to connect tap device to OVS) network settings cannot be restored automatically, TAPDOWNSCRIPT is needed (see netdev tap, downscript in man qemu).


aaannz:
LGTM

PR #463, opened 2016-04-06
Add support for executing tests via Xen HVM and PV

mnowaksuse:
This change adds two new variables VIRSH_VMM_FAMILY and VIRSH_VMM_TYPE
which determine hypervisor (VMM) used.

```
VMM             VIRSH_VMM_FAMILY    VIRSH_VMM_TYPE
kvm             kvm                 hvm
xen (fullvirt)  xen                 hvm
xen (paravirt)  xen                 linux
```

For s390 tests via zKVM VIRSH_VMM_FAMILY and VIRSH_VMM_TYPE ought to be
set to "kvm", and "hvm", respectively in order the VM is created
properly.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -36,8 +36,8 @@ sub new {
     $self->instance(get_var('VIRSH_INSTANCE', 1));
     # default name
     $self->name("openQA-SUT-" . $self->instance);
-    $self->vmm_family(get_required_var('VIRSH_VMM_FAMILY'));
-    $self->vmm_type(get_required_var('VIRSH_VMM_TYPE'));
+    $self->vmm_family(get_var('VIRSH_VMM_FAMILY', 'kvm'));
+    $self->vmm_type(get_var('VIRSH_VMM_TYPE', 'hvm'));
     $self->_init_xml();
 
     return $self;
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -36,8 +36,8 @@ sub new {
     $self->instance(get_var('VIRSH_INSTANCE', 1));
     # default name
     $self->name("openQA-SUT-" . $self->instance);
-    $self->vmm_family(get_var('VIRSH_VMM_FAMILY'));
-    $self->vmm_type(get_var('VIRSH_VMM_TYPE'));
+    $self->vmm_family(get_required_var('VIRSH_VMM_FAMILY'));
+    $self->vmm_type(get_required_var('VIRSH_VMM_TYPE'));
     $self->_init_xml();
 
     return $self;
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -34,7 +36,9 @@ sub new {
     $self->instance(get_var('VIRSH_INSTANCE', 1));
     # default name
     $self->name("openQA-SUT-" . $self->instance);
-    $self->_init_xml;
+    $self->vmm_family(get_var('VIRSH_VMM_FAMILY'));
+    $self->vmm_type(get_var('VIRSH_VMM_TYPE'));
```
aaannz: I assume kvm should be the default, right? Please use
```perl
$self->vmm_family(get_var('VIRSH_VMM_FAMILY', 'kvm'));
$self->vmm_type(get_var('VIRSH_VMM_TYPE', 'hvm'));
```
Or are those two vars supposed to be mandatory? Then I'm missing check for presence.

Soulofdestiny:
I will recheck this tomorrow morning to make sure that it does not break zKVM testing

consoles/sshVirtsh.pm
mnowaksuse: `VIRSH_VMM_FAMILY` and `VIRSH_VMM_TYPE` are meant to be mandatory. I don't think of KVM as a default one. Fixed.

coolo:
please don't require new variables if we can have a default - and add the variables to docs/backend_vars

mnowaksuse:
What is the default going to be, and why?

coolo:
just as the BACKEND defaults to qemu, svirt should default to kvm IMO. BTW: why a variable and not an argument within the bootloader test? you need to have xen specific bootloader anyway.

coolo:
that test could then require a variable

consoles/sshVirtsh.pm
```diff
@@ -183,6 +206,17 @@ sub add_pty {
...
     $console->appendChild($elem);
 
+    if (!($self->vmm_family eq 'xen' && $self->vmm_type eq 'linux')) {
+        my $serial = $doc->createElement('serial');
+        $serial->setAttribute(type => 'pty');
+        $devices->appendChild($serial);
+
+        $elem = $doc->createElement('target');
+        $elem->setAttribute(type => 'isa-serial');
+        $elem->setAttribute(port => $args->{port});
```
coolo: this broke zkvm as it doesn't have a isa-serial

coolo: I guess it's really safer to do this intellegence from a bootloader_* test instead of here through magic variables.

mnowaksuse: This should be addressed in PR #471 .

PR #465, opened 2016-04-06
Support newer Supermicro boards (X10)

coolo:
They no longer send 15bit RGB, but a jpeg like format. The deocder
is based on research done by the novnc project, but the implementation
is ours


XGWang0:
Hi Coolo:
We used the latest package to do ipmi test, got following error : 
"09:54:20.6085 capture loop failed Can't close(GLOB(0x5c39fc8)) filehandle: 'No child processes' at /usr/lib/os-autoinst/backend/baseclass.pm line 268",  then it seems like hanging. I try to close it via close BUTTON at the left side of page,  it doest not work. Additionally, the relevant worker are down.

coolo:
I see the same on James' server and I'm debugging it atm

coolo:
that is apparmor blocking you - if you feel like it, you can fine tune the profile. We disabled apparmor on the workers running ipmi tests, so the profile is a bit outdated.

PR #471, opened 2016-04-08
Let user define the PTY type (console, serial)

mnowaksuse:
The <serial ...> is now to be created by:

  add_pty({pty_dev => 'serial', ...)

Similarly with <console ...>.


coolo:
wouldn't this require a change in our bootloader_zkvm?

mnowaksuse:
I believe that changes necessary for `installation/bootloader_zkvm.pm` are in place thanks to yours e8cd036.

coolo:
I did that so I could deploy

PR #472, opened 2016-04-12
Add hold_key and release_key to support alt+tab switcher

mitiao:
follow send_key to implement the hold_key and release_key to support alt+tab 
this request came from https://progress.opensuse.org/issues/11326
and the test result:
http://147.2.207.208/tests/1038

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -686,7 +686,7 @@ sub init_ikvm_keymap {
 
 
 sub map_and_send_key {
-    my ($self, $keys) = @_;
+    my ($self, $keys, $down_flag) = @_;
 
     if ($self->ikvm) {
         $self->init_ikvm_keymap;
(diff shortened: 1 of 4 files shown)


backend/driver.pm
```diff
@@ -144,6 +144,16 @@ sub send_key {
     return $self->_send_json({cmd => 'send_key', arguments => {key => $key}});
 }
 
+sub hold_key {
```
coolo: you don't need these if you call the backend function as hold_key({key => $key}). The send_key function is historic

consoles/VNC.pm
```diff
@@ -686,7 +686,8 @@ sub init_ikvm_keymap {
 
 
 sub map_and_send_key {
-    my ($self, $keys) = @_;
+    my ($self, $keys, $down_flag) = @_;
+    $down_flag //= '';
```
coolo: Please don't. Use defined to check for undefined values

backend/driver.pm
mitiao: Ah, I see.
Thanks.

consoles/VNC.pm
mitiao: Fixed

PR #473, opened 2016-04-12
Allow send_key to process multiple keys

kravciak:


coolo:
aehm, what's the point?

kravciak:
just improvement, if you think it is useless don`t merge, but there are a lot of tests doing
  send_key "down";
  send_key "left";
  send_key "up";
  ....
which could look cleaner:
  send_key "down left up";  # comment where did i jump

coolo:
I don't think this is cleaner, no. And it's not really the norm either, but makes the API more complex

PR #474, opened 2016-04-12
Export autotest::loadtests to avoid helpers in main.pm

aaannz:
- without this one needs to use either
    autotest::loadtest('tests/<testname>')
  or custom helper (as we do in opensuse tests)
- export loadtest only on demand for compatibility reasons

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -18,6 +18,8 @@ package autotest;
 use strict;
 use bmwqemu;
 use basetest;
+use Exporter qw/import/;
+our @EXPORT_OK = qw/loadtest $current_test/;
 
 use File::Basename;
 use File::Spec;


coolo:
what implies that tests have to be in a tests/ sub directory?

aaannz:
That's right. Nothing technical, just our convention.
Without that however this export is useless to me since by this I wanted to avoid explaining difference between `autotest::loadtest` and our `loadtest` helper during workshops/hands on sessions.

coolo:
how about checking if the file actually exists? I would like to avoid a loadtest "notexistant" to cause an error saying "no such file tests/notexistant" and users saying "yeah?" :)

aaannz:
Do you mean check if `tests/something` exists and if so load that one? And if not try to load the one without `tests` path prefix? I don't like this.

Ok, how about I remove this `tests` check and prefix adding and add it to distri-opensuse helper in same way. That way I can teach people to just use this version writing `loadtest 'tests/test.pm'` and if they some day use this format in opensuse main.pm we correct them in review but even if not, the opensuse helper will not add another `tests` prefix to the path.

coolo:
now your commit message doesn't fit the commit anymore.

PR #477, opened 2016-04-14
[do not merge] Add correct shutdown capability on s390

Soulofdestiny:


Soulofdestiny:
@coolo I'm not sure about the part with the x3270 and the local Xvnc

backend/s390x.pm
```diff
@@ -69,15 +66,15 @@ sub do_stop_vm {
 
     # now cleanly disconnect from the guest and then kill the local
     # Xvnc
-    $self->deactivate_console({testapi_console => 'sut'});
-    $self->deactivate_console({testapi_console => 'worker'});
+    # FIXME: "console worker does not exist at /home/os-autoinst/consoles/s3270.pm line 491 thread 1."
```
coolo: that's very old code from froh. Just erase it. we can revive it if we found a use

backend/s390x.pm
```diff
@@ -69,15 +66,15 @@ sub do_stop_vm {
...
+    #$self->deactivate_console({testapi_console => 'sut'});
+    #$self->deactivate_console({testapi_console => 'worker'});
     return;
 }
 
 sub status {
     my ($self) = @_;
-    # FIXME: do something useful here.
-    carp "status not implemented";
+    return 'shutdown';
```
coolo: now that I call a fake

Soulofdestiny: yeah... it's kinda the same as in svirt and ipmi backend, isn't it?

Soulofdestiny:
updated

backend/s390x.pm
coolo: well, they return nothing - you return something pretty concrete. Perhaps you can track the status in the backend and set it to shutdown in stop_vm

coolo:
I think what you really want to do is moving assert_shutdown into the backend - and get rid of this status function that we actually don't use meaningfully

coolo:
I finally managed to implement it properly - I couldn't test it though as the current product doesn't reach shutdown test ;(

Soulofdestiny:
I will test this in my local instance

PR #478, opened 2016-04-14
Add Retry for ssh-connections on s390

Soulofdestiny:
We see sporadic fails when trying to connect via ssh on s390

Just retry it up to five times if no connection can be established

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -971,15 +971,29 @@ sub new_ssh_connection {
     my ($self, %args) = @_;
 
     my $ssh = Net::SSH2->new;
-    $ssh->connect($args{hostname});
-    $args{username} ||= 'root';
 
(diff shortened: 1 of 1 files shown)


backend/baseclass.pm
```diff
@@ -971,15 +971,29 @@ sub new_ssh_connection {
...
+            $args{username} ||= 'root';
+
+            if ($args{password}) {
+                $ssh->auth(username => $args{username}, password => $args{password});
+            }
+            else {
+                # this relies on agent to be set up correctly
+                $ssh->auth_agent($args{username});
+            }
+            print "Connection to $args{username}\@$args{hostname} established\n" if $ssh->auth_ok;
```
coolo: use bmwqemu::diag instead (without \n)

Soulofdestiny:
updated

PR #479, opened 2016-04-15
Set serial to /dev/hvc0 for Xen PV

mnowaksuse:


coolo:
we need to get rid of this function and have the backends set it.

But nonetheless - please reformat the code so I don't have to scroll sideways

PR #480, opened 2016-04-18
Fix get_last_mouse_set to work with consoles

coolo:
This fixes assert_and_click moving the mouse randomly


coolo:
ACKed by xjin

PR #481, opened 2016-04-18
Use xhci usb on POWER

k0da:
xhci should have less CPU consumption compared to ohci and ehci

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -327,7 +327,9 @@ sub start_qemu {
     }
     elsif ($vars->{OFW}) {
         $vars->{QEMUVGA} ||= "std";
+        $vars->{QEMUMACHINE} = "usb=off";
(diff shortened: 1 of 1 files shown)


backend/qemu.pm
```diff
@@ -327,7 +327,9 @@ sub start_qemu {
     }
     elsif ($vars->{OFW}) {
         $vars->{QEMUVGA} ||= "std";
+        $vars->{QEMUMACHINE} = "usb=off";
```
dirkmueller: I think you want to do ".= ',usb=off' " here.. ?

k0da: No, I have expected -machine usb=off

nilxam:
LGTM

nilxam:
already running on openqa.o.o a while, considers it was TESTED.

PR #482, opened 2016-04-18
Make zKVM capable of variable CPU and RAM

Soulofdestiny:
For some tests we need more RAM and vCPUs on z/KVM

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -95,12 +95,12 @@ sub _init_xml {
     $root->appendChild($elem);
 
     $elem = $doc->createElement('memory');
-    $elem->appendTextNode('512');
+    $elem->appendTextNode(get_var('QEMURAM', '512'));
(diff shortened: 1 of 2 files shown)


coolo:
please document in docs/backend_vars that QEMU variables affect the svirt backend

Soulofdestiny:
done

coolo:
@mnowaksuse WDYT?

mnowaksuse:
Looks fine with me as we may need to craft RAM size for VMware anyway, because it's "memory size has to be a multiple of 4096", once we get there.

Soulofdestiny:
this is apparently not working correctly, e.g. the xml-parameters are not set
https://openqa.suse.de/tests/358287

when i check the xml file after the system was defined, it hast still default values in it

Soulofdestiny:
mhm... works on local instance - it's not deployed yet in productive?

coolo:
we have 3492b1 deployed

PR #483, opened 2016-04-20
Use /dev/hvc on Xen with pv_ops only (12SP2)

mnowaksuse:
Kernels with pv_ops default to /dev/hvc, rest of Xen PV kernels sticks
to /dev/xvc. Better way to test that feature is welcome.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -78,11 +78,16 @@ Used for internal initialization, do not call from tests.
 =cut
 sub init {
     $serialdev = get_var('SERIALDEV', "ttyS0");
-    if (   get_var('OFW')
-        || check_var('BACKEND', 's390x')
-        || (check_var('VIRSH_VMM_FAMILY', 'xen') && check_var('VIRSH_VMM_TYPE', 'linux')))
-    {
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -78,12 +78,17 @@ Used for internal initialization, do not call from tests.
...
     $serialdev = get_var('SERIALDEV', "ttyS0");
-    if (   get_var('OFW')
-        || check_var('BACKEND', 's390x')
-        || (check_var('VIRSH_VMM_FAMILY', 'xen') && check_var('VIRSH_VMM_TYPE', 'linux')))
-    {
+    if (get_var('OFW') || check_var('BACKEND', 's390x')) {
         $serialdev = "hvc0";
     }
+    elsif (check_var('VIRSH_VMM_FAMILY', 'xen') && check_var('VIRSH_VMM_TYPE', 'linux')) {
+        if (get_var('VERSION', '') =~ /12SP2/) {
```
coolo: the VERSION is never 12SP2 on our instance. And the version is if anything exactly '12-SP2', so check_var will do

coolo:
once the alpha3 rush is over I will kill $serialdev from init and move this into main.pm

PR #487, opened 2016-04-26
don't hardcode libexecdir in openvswitch service file

AdamWill:
once again, a hardcoding of libexecdir that breaks Fedora, so
we need to have the Makefile handle it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -94,6 +94,7 @@ CLEANFILES = \
 	doc/testapi.html 
 
 PERL_MODULE = ppmclibs/blib/arch/auto/tinycv/tinycv.so
+OPENVSWITCH_SERVICE = systemd/os-autoinst-openvswitch.service
 
 ppmclibs/Makefile: ppmclibs/Makefile.PL
 	cd ppmclibs && \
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -102,6 +103,11 @@ ppmclibs/Makefile: ppmclibs/Makefile.PL
 $(PERL_MODULE): ppmclibs/Makefile
 	$(MAKE) -C ppmclibs
 
+$(OPENVSWITCH_SERVICE):
+	rm -f $@ $@.tmp
+	sed -e 's,@pkglibexecdir[@],$(pkglibexecdir),g' '$(top_srcdir)/$@.in' > '$(top_srcdir)/$@.tmp'
```
coolo: you don't believe in sed -i ?

AdamWill: It doesn't seem appropriate here. The file is not usable as-is, so it should be marked as such by calling it `.in` and not editing it in-place. It's not like the other cases, where the file is usable as-is when os-autoinst is run out of the git checkout. If anything we could drop the use of the `.tmp` file I guess, it's not doing a whole lot, I just copied it out of some autotools docs I think.

coolo: right, I overlooked. But this replacement of @pkglibexecdir@ is done by autoconf if you list it in configure.ac

AdamWill: that's not really the right way to do it, per [this page](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Installation-Directory-Variables.html) - at least as I read it, see the bit from "Most of these variables have values that rely on prefix or exec_prefix." onwards. Note particularly "A corollary is that you should not use these variables except in makefiles".

coolo: I used AC_CONFIG_FILES for such purpose for years, but ok :)

PR #488, opened 2016-04-27
Support JeOS on VMware

mnowaksuse:
This adds support for running JeOS on VMware (net install won't work as
VMware does not support direct kernel boot).

Serial port is exposed on ESX server as `virsh console' is not supported
by esx libvirt driver, netcat is used instead.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -22,6 +22,7 @@ use testapi qw/get_var/;
 require IPC::System::Simple;
 use autodie qw(:all);
 use XML::LibXML;
+use File::Temp qw/tempfile/;
 
 use Class::Accessor "antlers";
 has instance   => (is => "rw", isa => "Num");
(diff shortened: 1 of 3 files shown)


mnowaksuse:
Requires https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1300.

consoles/sshVirtsh.pm
```diff
@@ -315,6 +332,26 @@ sub add_disk {
 sub define_and_start {
     my ($self) = @_;
 
+    my $remote_vmm = "";
+    if ($self->vmm_family eq 'vmware') {
+        my $libvirtauthfilename = '/var/lib/libvirt/images/libvirt.auth';
```
coolo: does it have to live in /images ?

consoles/sshVirtsh.pm
```diff
@@ -327,18 +364,18 @@ sub define_and_start {
     $chan->close();
 
     # shut down possibly running previous test (just to be sure) - ignore errors
-    $self->{ssh}->channel()->exec("virsh destroy " . $self->name);
-    $self->{ssh}->channel()->exec("virsh undefine " . $self->name);
+    $self->{ssh}->channel()->exec("virsh " . "$remote_vmm" . " destroy " . $self->name);
```
coolo: just put the variable in the string - no need to concat if it's a string (while $self->name is a function call) and I find "$virsh $remote_vmm destroy " easier to read

consoles/sshVirtsh.pm
mnowaksuse: Nope. `/var/lib/libvirt/images/` is just our favorite dumpster folder, see `$xmlfilename`. Suggestions?

mnowaksuse:
z/KVM test got updated in os-autoinst/os-autoinst-distri-opensuse#1300.

consoles/sshVirtsh.pm
coolo: I would have expected a temporary file actually

PR #489, opened 2016-04-27
Do not boot from dvd if BOOT_HDD_IMAGE is set

nadvornik:


coolo:
please document the variable

PR #490, opened 2016-04-29
add an option to upload_logs to avoid dying on failure

AdamWill:
in Fedora we use upload_logs a lot in post-fail hooks, and in
some cases the file may not exist. For instance, anaconda
creates lots of log files, some of which may not exist in some
circumstances. We would like to just be able to try and upload
them all but not have the post-fail hook die if any of them
does not exist, instead of having to test for each file's
existence before uploading it, or have a bunch of conditionals
that magically know which log files will exist in which state.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1004,21 +1004,32 @@ sub data_url($) {
 
 =head2 upload_logs
 
-  upload_logs $file;
+  upload_logs($file [, failok => 0 ]);
 
 Upload C<$file> to OpenQA WebUI as a log file and
-return the uploaded file name.
+return the uploaded file name. If failok is not set, a failed upload
+will cause the test to die. Failed uploads happen if the file does not
+exist or is over 20GiB in size, so failok is useful when you just want
+to upload the file if it exists but not mind if it doesn't.
 
 =cut
 sub upload_logs {
-    my ($file) = @_;
+    my $file   = shift;
+    my %args   = @_;
+    my $failok = $args{failok} || 0;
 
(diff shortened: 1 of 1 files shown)


AdamWill:
I'm testing this right now. What's that you say? Test first *then* submit a PR? Nahhh, that doesn't sound like fun...

AdamWill:
welp, looks like it works: compare https://openqa.fedoraproject.org/tests/14834 (sad! missing log files! post fail hook died half way through when `/tmp/dnf.log` didn't exist!) to https://openqa.stg.fedoraproject.org/tests/15371 (happiness! all the logs are there! we don't care about stupid dnf!)

the weird behaviour on the very last log file is just because I made a stupid typo when editing that file's `upload_logs` command and put the `, 1` in the filename, heh.

coolo:
yeah, but I don't like the ,1 - wherever we have this in a API the test writers tend to just copy&paste this magic number without knowing what it does (see type_string and send_key ). Better have a named parameter as we added for type_string and as discussed in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1069

AdamWill:
hehe, I can believe it =) OK, will update that soon. thanks.

AdamWill:
How does that look?

AdamWill:
oh perltidy, why must you.

coolo:
do your friends know what kind of perl hacker you became? :)

AdamWill:
@coolo you must NEVER TELL

actually i just ripped that off from `type_string`, sooo....

PR #492, opened 2016-05-04
Make Backend capable of SSH-X and clean up

Soulofdestiny:


consoles/console.pm
```diff
@@ -19,6 +19,7 @@ use strict;
 use warnings;
 require IPC::System::Simple;
 use autodie qw(:all);
+use testapi;
```
coolo: why? that sounds actually quite dangerous

Soulofdestiny: yeah that was leftover from try-and-error -> fixed

PR #494, opened 2016-05-11
Add Hyper-V svirt support

mnowaksuse:
We need ssh -X connection so we can run Xvnc via vncserver.

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -17,7 +17,7 @@
 package backend::svirt;
 use strict;
 use base ('backend::baseclass');
-use testapi qw(get_required_var);
+use testapi qw(get_required_var check_var);
 
 use IO::Select;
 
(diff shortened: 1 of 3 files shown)


mnowaksuse:
Required: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1348

coolo:
call sshCommand with , 1 instead of hardcoding -X

okurz:
+1

mnowaksuse:
It seems that I don't need `-X` at all.

cpanfile
```diff
@@ -33,6 +33,7 @@ requires 'Mojolicious::Lite';
 requires 'Net::DBus';
 requires 'Net::SNMP';
 requires 'Net::SSH2';
+requires 'Net::Telnet';
```
coolo: this seems to be a bit undermotivated

coolo: Cherry pick https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/3880f711b9dc1e73ecf77e171d727d3713246d49 and revert that line here

okurz:
LGTM

PR #498, opened 2016-05-18
Add run_cmd to svirt BACKEND

mimi1vx:


coolo:
please add some comment on how and where to use this

consoles/sshVirtsh.pm
```diff
@@ -355,4 +355,18 @@ sub attach_to_running {
     $self->backend->start_serial_grab($self->name);
 }
 
+# Allow send command to libvirt host and get exist status of command
+# for example:
+# my $ret = $svirt->run_cmd("virsh snapshot-create-as snap1");
+# die "snapshot creation failed" unless $ret == 0;
+sub run_cmd {
+    my ($self, $args) = @_;
```
coolo: it's just a detail, but as the string contains the command including the arguments, I think calling it $args is misleading.

okurz:
LGTM

PR #499, opened 2016-05-20
Add posibility to set custom SUT name in svirt backend

mimi1vx:
It uses VIRSH_GUESTNAME and VIRSH_INSTANCE, defaults to
"openQA-SUT-1"

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -34,8 +34,8 @@ sub new {
     my $self = $class->SUPER::new($testapi_console, $args);
 
     $self->instance(get_var('VIRSH_INSTANCE', 1));
-    # default name
-    $self->name("openQA-SUT-" . $self->instance);
+    # default name is openQA-SUT-1
+    $self->name(get_var('VIRSH_GUESTNAME', "openQA-SUT-") . $self->instance);
     $self->vmm_family(get_var('VIRSH_VMM_FAMILY', 'kvm'));
     $self->vmm_type(get_var('VIRSH_VMM_TYPE', 'hvm'));
     $self->_init_xml();


consoles/sshVirtsh.pm
```diff
@@ -34,8 +34,8 @@ sub new {
     my $self = $class->SUPER::new($testapi_console, $args);
 
     $self->instance(get_var('VIRSH_INSTANCE', 1));
-    # default name
-    $self->name("openQA-SUT-" . $self->instance);
+    # default name is openQA-SUT-1
+    $self->name(get_var('VIRSH_GUESTNAME', "openQA-SUT-") . $self->instance);
```
aaannz: I have a mixed feelings about the separator (-) to be a part of VIRSH_GUESTNAME.

@mimi1vx What separator are those test machines using? Can the dash be hardcoded?
```perl
$self->name(get_var('VIRSH_GUESTNAME', 'openQA-SUT') . '-' . $self->instance);
```
or
```perl
$self->name(sprintf('%s-%u', get_var('VIRSH_GUESTNAME', 'openQA-SUT'), $self->instance));
```

mimi1vx: why hardcode any character ? waht about 'underscore' or 'dot' ?

aaannz: I just don't like VIRSH_GUESTNAME to have trailing separator.
But true, we don't have to use any separator and it's more future-proof not to hardcode anything. Disregard my comments.

coolo:
I added attach_to_running for this functionality.

coolo:
and just as @aaannz I don't like that GUESTNAME and INSTANCE are mangled. If you already know the  domain name, then why split it artificually

mimi1vx:
@coolo oh, I'm blind ... , please forget about this PR

PR #501, opened 2016-05-23
Reduce self test execution time

okurz:
By making the sleep time within wait_serial overridable the execution of the
tests in t/03-testapi.pl are reduced to near 0 instead of multiple seconds as
in before.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -555,7 +555,7 @@ sub wait_serial {
     if ($expect_not_found) {
         $matched = !$matched;
     }
-    sleep 1;                               # wait for one more screenshot
+    bmwqemu::wait_for_one_more_screenshot();
 
     # to string, we need to feed string of result to
     # record_serialresult(), either 'ok' or 'fail'
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -419,6 +419,11 @@ sub hashed_string {
     return substr($hash, 0, $count);
 }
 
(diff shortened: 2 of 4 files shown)


testapi.pm
```diff
@@ -66,6 +66,7 @@ our $password;
 our $serialdev;
 
 our $last_matched_needle;
+our $screenshot_wait_time = 0;
```
dzedro: if this will be hardcoded 0 then why is sleep used at all ?

okurz: don't look, embarassing :blush:

fixed in update

coolo:
hmm, I see the requirement but the solution feels overly hacky. To everyone reading testapi.pm on its own this variable looks really suspicious :)

okurz:
I am open for suggestions. But if you and me don't come up with a better solution, what should we do?

coolo:
just brain storming:

having this sleep actually in save_screenshot called via some grace_period function that we stub from the tests like http://stackoverflow.com/questions/16860377/how-can-i-mock-a-sub-in-another-module

okurz:
> just brain storming: …

sure, that's an alternative. See the new implementation if you like it more.

Changed:
* variable for wait time replaced by 'grace_period function' within bmwqemu which is mocked in tests. Actually, the "Test::MockModule" looks nice and easy and can be helpful for more use cases now.

PR #502, opened 2016-05-23
Accept positional and named arguments for assert_script_run

okurz:
Using no hash ref as a more modern approach, compare to type_string.

Related discussion:
os-autoinst/os-autoinst-distri-opensuse#1069

Verification test run: http://lord.arch/tests/760

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -610,9 +610,13 @@ sub script_run {
 
 =head2 assert_script_run
 
-  assert_script_run($command [, $wait, $fail_message]);
+  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message]);
 
-Run C<$command> via C<script_run> and C<die> if its exit status is not zero.
+Deprecated mode
+
+  assert_script_run($cmd [, $timeout [, $fail_message]]);
+
+Run C<$cmd> via C<script_run> and C<die> if its exit status is not zero.
 The exit status is checked by magic string on the serial port.
 See C<script_run> for default timeout.
 C<$fail_message> is returned in the die message if specified.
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -620,13 +624,19 @@ I<Make sure the command does not write to the serial output.>
 
 =cut
 sub assert_script_run {
-    my ($cmd, $timeout, $fail_message) = @_;
+    my ($cmd, $args, $fail_message) = @_;
+    unless (ref $args eq ref {}) {
```
aaannz: Where did you find such construct?
```perl
unless (ref $args eq 'HASH') {
```
is enough.

okurz: found it on https://rosettacode.org/wiki/Named_parameters#Perl
any other remarks?

aaannz: Ok, but can you change it to comparing to HASH? We already use it that way. See [Schema::Result::Jobs#L262](https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Schema/Result/Jobs.pm#L262) or [Scheduler::Scheduler#L625](https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Scheduler/Scheduler.pm#L625) or [WebSockets::Server#L165](https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebSockets/Server.pm#L165)

okurz: On Tuesday 24 May 2016 05:55:13 Ondřej Holeček wrote:
> > @@ -620,13 +624,19 @@ I<Make sure the command does not write to the serial
> > output.>> 
> >  =cut
> >  sub assert_script_run {
> > 
> > -    my ($cmd, $timeout, $fail_message) = @_;
> > +    my ($cmd, $args, $fail_message) = @_;
> > +    unless (ref $args eq ref {}) {
> 
> Ok, but can you change it to comparing to HASH? We already use it that way.

done

coolo:
that might have been state of the art in 2006, but in type_string we do the same thing without a hash ref

coolo:
and you'll find perl using this kind of named parameters all over in CPAN, while mojolicious does it the HASH way. But we need to have _one_ way to pass named parameters

okurz:
> But we need to have *one* way to pass named parameters

Would you be ok with me changing type_string to also use a hash as a param then? I did not find any occurence of type_string being used with named arguments in os-autoinst-distri-opensuse.

coolo:
Well, http://modernperlbooks.com/books/modern_perl/chapter_10.html is making a long explanation on how list casting works to then add a side node that perl best practises prefers hash ref (for reasons we don't really care about IMO).

So it all boils down to what exactly is the error on syntax errors - on the costs of ugly code. I stand with the list way :)

coolo:
To make this more clear let's spell it out:

```
assert_script_run('false', {timeout => 0, fail_message => 'my custom fail message'});
vs.
assert_script_run('false', timeout => 0, fail_message => 'my custom fail message');
```

It's clearly a nuiance, so no I don't mind if we went with hashref for type_string too. But I expect most of our developers not getting the meaning of the requirement of {}

aaannz:
At the end I am with @coolo here. I understand that there is the hash ref benefit of proper error place, but using lists feels more natural when not using blessed objects like in case of Mojo.

okurz:
hmpf, pulled @aaannz in with a different expectation on his view ;-)

coolo:
Fair usage though :)

okurz:
Verification test run: http://lord.arch/tests/760

PR #504, opened 2016-05-24
Prevent warning about uninitialized variable in script_run

okurz:
As observed in https://openqa.suse.de/tests/395751/file/autoinst-log.txt

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -103,6 +103,7 @@ sub script_run {
     # start console application
     my ($self, $name, $wait) = @_;
 
+    $wait //= $bmwqemu::default_timeout;
     testapi::type_string "$name";
     if ($wait > 0) {
         my $str = bmwqemu::hashed_string("SR$name$wait");


aaannz:
If script run produces uninit var warning then it is not called via testapi. Is that like, hmm, supported?

aaannz:
Found it in `lib/susedistribution.pm` in tests. Funny that we sometimes use testapi, sometimes directly distribution object.

coolo:
as testapi uses $distri using testapi doesn't make sense in some cases. actually testapi is just too fat, but splitting it now is a hard sell

coolo:
I would prefer if you spell out the default for $wait instead of having a guard before every if

aaannz:
> actually testapi is just too fat

True to that, esp. this `distribution.pm` and its derivatives backed stuff makes it harder to explain to new comers, esp. on conferences whit people outside SUSE background. Different topic however.

> I would prefer if you spell out the default for $wait instead of having a guard before every if

That is the preferred solution for me too. Just take it from testapi script_run:
```perl
$wait //= $bmwqemu::default_timeout;
```

okurz:
updated:
* set default as necessary instead of if

PR #506, opened 2016-05-27
Put temp files to /tmp instead of default temp dir

mnowaksuse:
This puts temporary files to /tmp dir, before it put files to
/var/lib/openqa/pool/#/tmp/ which does not exist on VIRSH_HOST (does on
worker but that's not accessible).

(Well the branch should have been named as a _VMware_ fix...)

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -335,7 +335,7 @@ sub define_and_start {
 
     my $remote_vmm = "";
     if ($self->vmm_family eq 'vmware') {
-        my ($fh, $libvirtauthfilename) = tempfile();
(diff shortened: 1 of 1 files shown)


sysrich:
I think it's important that openQA workers 'keep to themselves' in /var/lib/openqa/pool as much as possible.

On production systems pool is on a seperate filesystem from the OS root, so this commit could lead to unexpected consequences, such as the exhaustion of the root filesystem or increased IO to the slower OS disk (compared to the fast SSDs we use on our production pools)

can we have os-autoinst check whether it can write to the default temp dir, and only use /tmp as a fallback?

coolo:
this is just a file name created on the vmware server. There is no point in creating it in pool, no matter if worker can write there or not - the worker is not running on that machine it needs the file

PR #507, opened 2016-05-28
Move bmwqemu functions (ab-)used in openSUSE tests to test API

okurz:
The functions 'save_vars', 'hashed_string' and 'diag' are currently used in
os-autoinst-distri-opensuse but only the testapi functions should be used. If
we need them we may as well just include them in the test API and subsequently
clean up tests (and then bmwqemu). 'random_string' sounds similar to
'hashed_string' but is only used within bmwqemu so only the doc has been
adapted but the function not moved.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -105,7 +105,7 @@ sub script_run {
 
     testapi::type_string "$name";
     if ($wait > 0) {
-        my $str = bmwqemu::hashed_string("SR$name$wait");
+        my $str = hashed_string("SR$name$wait");
         testapi::type_string " ; echo $str > /dev/$testapi::serialdev\n";
         testapi::wait_serial($str);
     }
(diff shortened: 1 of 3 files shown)


coolo:
that main.pm calls save_var you can consider a bug - as every isotovideo already saves the vars after loading main.pm

okurz:
So first remove save_vars call, then move stuff to testapi, then rename rest in tests?

okurz:
updated:
* only `hashed_string` and `diag` added to testapi

distribution.pm
```diff
@@ -105,7 +105,7 @@ sub script_run {
 
     testapi::type_string "$name";
     if ($wait > 0) {
-        my $str = bmwqemu::hashed_string("SR$name$wait");
+        my $str = hashed_string("SR$name$wait");
```
Soulofdestiny: should be testapi::hashed_string
please don't deploy :)

PR #508, opened 2016-05-30
Fix use of hashed_string after ce6d089e

okurz:


okurz:
@coolo, @Soulofdestiny please review

coolo:
I don't think we were in danger of deploying this week :)

PR #509, opened 2016-06-02
Get output from SSH commands

mnowaksuse:
And remove Xen HVM emulator setup which is not needed.


okurz:
LGTM

consoles/sshVirtsh.pm
```diff
@@ -330,6 +317,27 @@ sub add_disk {
     return;
 }
 
+# In list context returns pair ($stdout, $stderr). In void (and scalar)
+# context just logs stdout and stderr, returns nothing.
+sub get_ssh_outputs {
+    my $chan = shift;
```
coolo: please don't use shift for parameters. Look around in this file

coolo:
I'm a bit puzzled by the plural. I consider stderr and stdout part of the one output, it's just two channels of it.

mnowaksuse:
Sub is now in singular.

PR #510, opened 2016-06-07
Try to reconnect s390 terminal more often

okurz:
Related progress issue: https://progress.opensuse.org/issues/10584

### DIFF
--- a/consoles/s3270.pm
+++ b/consoles/s3270.pm
@@ -245,8 +245,7 @@ sub expect_3270() {
                 while (my $line = $self->{raw_expect_queue}->dequeue_nb()) {
                     push @$result, $line;
                 }
-
-                confess "status line matches neither buffer_ready nor buffer_full:\n" . Dumper($result) . $status_line;
+                warn "status line matches neither buffer_ready nor buffer_full:\n" . Dumper($result) . $status_line;
             }
 
         }
--- a/consoles/s3270.pm
+++ b/consoles/s3270.pm
@@ -471,7 +471,7 @@ sub connect_and_login() {
     $reconnect_ok //= 0;
     my $r;
     ###################################################################
-    # try to connect exactly twice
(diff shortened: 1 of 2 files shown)


coolo:
tidy wants that comment off the screen :)

okurz:
pair-programmed with mgriessmeier, regarded as auto-LGTM :-)

Soulofdestiny:
LGTM :+1:

consoles/s3270.pm
```diff
@@ -246,7 +246,10 @@ sub expect_3270() {
                     push @$result, $line;
                 }
 
-                confess "status line matches neither buffer_ready nor buffer_full:\n" . Dumper($result) . $status_line;
+                # TODO the following should be "confess" but since
+                # 7a2e1b41 confess itself dies inside the die handler w/o
+                # helpful output
+                warn "status line matches neither buffer_ready nor buffer_full:\n" . Dumper($result) . $status_line;
```
coolo: this is no longer failing, just warning. Is this intended?

okurz: yes as it looks like we can handle this better from a higher layer. The confess was visible some time ago but since 7a2e1b41 it seems to be swallowed by a die handler somewhere else so the backtrace was never visible and the test would just die and we don't have a chance to handle it. I think https://openqa.suse.de/tests/437837/file/autoinst-log.txt is one of these cases, couldn't find a more obvious example right now.

coolo: I don't care too much about the backtrace - I wonder if expect_3270 should be fatal or not

coolo: It's headed with ;)
## FIXME FIXME FIXME redesign with three goals

okurz: removed the TODO comment. just `warn` should be ok (discussed with coolo), fatal is not necessary here

PR #513, opened 2016-06-10
Increase verbosity in os-autoinst logfile

Soulofdestiny:
It was confusing that the output which is shown in a detail step of a module
is not displayed in the autoinst log, so we want to make sure to paste this also
in the uploaded logfile


okurz:
LGTM

@coolo as discussed personally yesterday. IIRC this merely gives back the old behaviour before the failed test output was forwarded to (only) the WebUI

PR #514, opened 2016-06-13
Feature/improve s3270

Soulofdestiny:
* consoles: Add explicit 'expected_status' to expect_s3270
* consoles: Die on timeout when searching for custom status line entry
* consoles: Correct name for s3270 parameter to 'delete_lines'
* consoles: Slight refactor of s3270 to improve readability


okurz:
LGTM - pair programmed with @Soulofdestiny 

@coolo please take a look

PR #515, opened 2016-06-13
Reduce thumbnail size - openQA shows them much smaller anyway

coolo:


okurz:
I would recommend to keep this open a bit longer, maybe someone else has a comment on this. I am ok with the change.

coolo:
The thumbnails when done in openQA were once 80x60 - and I guess scaling these to 60x45 looked even uglier, so that's why (I assume) we went with the larger the size. But I can't find any history of it. the size within os-autoinst has always been 160x120 when I introduced it

coolo:
seems noone else cares, I'm ready to merge

aaannz:
As you noted, it may get uglier. But I guess just merge it, we can change it back anytime.

PR #517, opened 2016-06-15
Adapt zKVM QEMURAM default to match other archs

Soulofdestiny:


okurz:
as this is only setting the default value if not already set, what about not setting a default at all and using `get_required_var`?

Soulofdestiny:
Well, I like the approach to have a default value more, because then you don't need to care about setting the variable in every single testcase, but only in the testcases you need specific values

okurz:
discussed with mgriessmeier:
Currently the value of QEMURAM is only set as override in some test_suites or machines, by default this value here is the value being used. As "1024MB" is a sane default not only for anything openSUSE or SUSE specific I +1 this commit. Anyway, to DRY I recommend to extract setting this value to a common baseclass for qemu and other virtual machine backends and set the default from there, probably the same for HDD size. Setting this value can also be done as "product specific" from the tests distribution, e.g. based on what has been documented as the default for the product we test.

aaannz:
:+1:

okurz:
See #518 for an alternative approach.

sysrich:
:+1:

coolo:
this is obsolete now, right?

Soulofdestiny:
should be -> #518

PR #518, opened 2016-06-15
Extract baseclass for qemu+svirt to set common values

okurz:
DRY when setting - already differing - values for common settings.

Related to https://github.com/os-autoinst/os-autoinst/pull/517

Verified on qemu+svirt

Verification run on svirt: http://lord.arch/tests/1287

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,7 +18,8 @@ backendexec_DATA = \
 	backend/qemu.pm \
 	backend/s390x.pm \
 	backend/svirt.pm \
-	backend/pvm.pm
+	backend/pvm.pm \
+	backend/virt.pm
 
 consolesexecdir = $(pkglibexecdir)/consoles
 consolesexec_DATA = \
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -16,7 +16,7 @@
 
 package backend::qemu;
 use strict;
-use base ('backend::baseclass');
+use base ('backend::virt');
 use File::Path qw/mkpath/;
 use File::Temp ();
 use Time::HiRes qw(sleep gettimeofday);
@@ -340,7 +340,6 @@ sub start_qemu {
     }
     push(@vgaoptions, "-vga", $vars->{QEMUVGA}) if $vars->{QEMUVGA};
 
-    $vars->{QEMUCPUS} ||= 1;
     if (defined($vars->{RAIDLEVEL})) {
         $vars->{NUMDISKS} = 4;
     }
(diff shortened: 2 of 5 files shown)


okurz:
testing on svirt(zkvm) first

okurz:
@Soulofdestiny

doc/backend_vars.asciidoc
```diff
@@ -66,69 +88,26 @@ QEMU_VIRTIO_RNG;boolean;0;Enable virtio random number generator
 RAIDLEVEL;;;
 SKIPTO;full name of test module;;Restore VM from snapshot and continue by running specified test module. Needs HDD image with snapshots present
 TAPDEV;device name;undef;TAP device name to which virtual NIC should be connected. Usually undef so automatic matching is used
+TAPDOWNSCRIPT;;;
 TAPSCRIPT;;;
-TESTDEBUG;boolean;0;Enable test debugging: override 'milestone' and 'fatal' test flags to 1. Snapshot are created after each successful test module and each fail aborts test run
```
aaannz: Why do you remove TESTDEBUG?

doc/backend_vars.asciidoc
```diff
@@ -66,69 +88,26 @@ QEMU_VIRTIO_RNG;boolean;0;Enable virtio random number generator
...
 USBBOOT;boolean;0;Mount ISO as USB disk and boot VM from it
 VDE_PORT;integer;worker instance + 10;number of vde switch port to connect
 VDE_SOCKETDIR;string;.;directory where vde_switch control socket is to be found
 VDE_USE_SLIRP;integer;1;whether to start slirpvde
 VNC;integer;worker instance + 90;Display on which VNC server is running. Actual port is 5900 + VNC
 VNCKB;;;
+WORKER_HOSTNAME;;;
 |====================
 
-.SVIRT backend
```
aaannz: What happened to SVIRT backend? Did I miss something?

coolo:
You didn't touch the Makefile -> this won't fly

Soulofdestiny:
I think I don't get this - And I'm still not a fan of having no default values...

sysrich:
**@mnowaksuse** - what is your opinion also, this PR touches a lot of your stuff.

I also share @Soulofdestiny's concern, different backends have different parameters, different defaults, s390x has 'whatever the mainframe admin set up', IPMI/generalhw backends has 'whatever the crap the real hardware has', many svirt implementations (hyperv, vmware) might not have full control over what they can do with a VM. I do not think we can abstract away every problem..

coolo:
aside the obviously broken docu commit I see no problem. The default values just moved to a common place.

mnowaksuse:
I am fine with this change.

okurz:
updated:
* not updating doc, sorry for the mess, doc/backend_vars.asciidoc auto-update is not nice
* included in Makefile.am

I am not doing anything different than extracting the default value assignment for the same variable to a common place. I am not "abstracting away any problem" but just preventing to introduce differing values for the default again. Still unsolved is that "QEMURAM" is not a good name for a variable used by the svirt problem. I don't see an easy change as we would need to do this step by step to keep consistent use of the variable in tests, backend, database settings, maybe even openQA.

PR #520, opened 2016-06-16
Cast the size type to match perl's to avoid warnings

coolo:
from 5.24's perldelta:
 The C<EXTEND> and C<MEXTEND> macros have been improved to avoid various
 issues with integer truncation and wrapping. In particular, some casts
 formerly used within the macros have been removed. This means for example
 that passing an unsigned nitems arg is likely to raise a compiler warning
 now (it's always been documented to require a signed value; formerly int,
 lately SSize_t).


aaannz:
:+1:

PR #521, opened 2016-06-16
Adapt script_run to return exit code if waiting for finish

aaannz:
needed for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1374

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -92,26 +92,30 @@ sub become_root {
 
 script_run($program, [$wait_seconds])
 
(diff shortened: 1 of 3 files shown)


coolo:
24 days? Not bad :)

But :+1:

aaannz:
Drafted during openQA hackaton :)

aaannz:
I'll make one small modification to this as I don't handle situation when we are waiting for a finish, but timeout. In that case I'll return undef. I'll also change return value when not waiting from 0 to undef.

okurz:
nice :-)

PR #522, opened 2016-06-17
Handle waiting timeouts for script_run

aaannz:
- return undef if wait timeouts
- return undef when not waiting


distribution.pm
```diff
@@ -111,11 +111,12 @@ sub script_run {
         my $str = testapi::hashed_string("SR$cmd$wait");
         testapi::type_string " ; echo $str-\$-? > /dev/$testapi::serialdev\n";
```
mimi1vx: $-? returns what ?

PR #523, opened 2016-06-21
Fix signal handler to have stable shutdown

coolo:
Do not call stop_vm through JSON but just kill the backend process
and leave the shutdown to itself

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -79,17 +79,17 @@ my $backend;
 # all so ugly ...
 sub signalhandler {
 
-    my $sig = shift;
-    print STDERR "$$: signalhandler got $sig\n";
+    my ($sig) = @_;
+    diag("$$: signalhandler got $sig");
     if ($autotest::running) {
         $autotest::running->fail_if_running();
         $autotest::running = undef;
     }
     if (defined $backend && $backend->{backend_pid}) {
-        print STDERR "killing backend process\n";
+        diag("$$: killing backend process $backend->{backend_pid}");
         kill('SIGTERM', $backend->{backend_pid});
         waitpid($backend->{backend_pid}, 0);
-        print STDERR "done with backend process\n";
+        diag("$$: done with backend process");
         $backend->{backend_pid} = 0;
     }
     # mark it as no longer working
@@ -101,10 +101,10 @@ sub signalhandler {
         $test->save_test_result();
     }
     if (defined $cpid) {
-        print STDERR "killing commands process\n";
+        diag("$$: killing commands process $cpid");
         kill('SIGTERM', $cpid);
         waitpid($cpid, 0);
-        print STDERR "done joining commands process\n";
+        diag("$$: done joining commands process");
         $cpid = 0;
     }
     _exit(1);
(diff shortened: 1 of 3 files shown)


isotovideo
```diff
@@ -73,19 +74,24 @@ bmwqemu::clean_control_files();
...
+my $backend;
 
 # all so ugly ...
 sub signalhandler {
 
-    # do not start a race about the results between the threads
-
     my $sig = shift;
-    diag("signalhandler $$: got $sig");
+    print STDERR "$$: signalhandler got $sig\n";
```
okurz: why not diag? maybe duplicate these messages to both log and system journal?

isotovideo
```diff
@@ -73,19 +74,24 @@ bmwqemu::clean_control_files();
...
-        $autotest::running->fail_if_running();
-        $autotest::running = undef;
+	$autotest::running->fail_if_running();
+	$autotest::running = undef;
+    }
+    if (defined $backend && $backend->{backend_pid}) {
+	print STDERR "killing backend process\n";
+        kill('SIGTERM', $backend->{backend_pid});
+        waitpid($backend->{backend_pid}, 0);
+        print STDERR "done with backend process\n";
```
okurz: and tidy does not complain about broken indention?

isotovideo
```diff
@@ -95,14 +101,32 @@ sub signalhandler {
...
+        print STDERR "done joining commands process\n";
+	$cpid = 0;
     }
-    exit(1);
+    _exit(1);
 }
 
+sub signalhandler_chld {
+
+    my $sig = shift;
```
okurz: `my ($sig) = @_;`?

isotovideo
coolo: possibly it's not running on isotovideo?

isotovideo
coolo: this was just my debugging - diag is fine. And no, diag can't write in the system journal - the worker catches all output of isotovideo into autoinst-log.txt

coolo:
Added tidy and diag

isotovideo
okurz: that's what I am saying, diag is writing to autoinst-log.txt but "print STDERR" might print to syslog so could do
```perl
sub super_important_log {
    my (@args) = @_;
    diag(@args);
    print STDERR "@args\n";
}
…
super_important_log("signalhandler $$: got $sig");
```

okurz:
As I could just observe my s390x worker to fool me, it can't reconnect, i investigated and found a stale worker process to be the culprit. So I will try now your change. Additionally, what do you think about:
* If SIGINT is called a second time or more, try harder to kill ourselves and children
* Spawn an internal "watchdog" on termination and kill children and parent process after a grace period if not shut down already, i.e. KILL after 30s

I did not observe problems when ending the process with SIGQUIT

okurz:
As I could just observe my s390x worker to fool me, it can't reconnect, i investigated and found a stale worker process to be the culprit. So I will try now your change. Additionally, what do you think about:
* If SIGINT is called a second time or more, try harder to kill ourselves and children
* Spawn an internal "watchdog" on termination and kill children and parent process after a grace period if not shut down already, i.e. KILL after 30s
* If a connection can not be established to e.g. vnc or something, die hard. I see error messages like "Connection refused" but nobody cares

I did not observe problems when ending the process with SIGQUIT

PR #524, opened 2016-06-27
Split into even more processes

coolo:
Make the tests a process too - and make isotovideo an event loop.

This is heavily broken and unfinished - it's merely to give you an early glimpse :)


okurz:
+1

coolo:
I officially add code now ;) +559 −553

coolo:
this works now - but this is such a heavy change that more people need to test it

isotovideo
```diff
@@ -148,41 +147,28 @@ require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
 # set a default distribution if the tests don't have one
 $testapi::distri ||= distribution->new;
 
-# start the command fork before we get into the backend, the command child
-# is not supposed to talk to the backend directly
-$cpid = commands::start_server($bmwqemu::vars{QEMUPORT} + 1);
+testapi::init();
+
+use Data::Dumper;
+print Dumper($testapi::distri);
```
okurz: remove or use proper logging call?

isotovideo
```diff
@@ -148,41 +147,28 @@ require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
...
 
-    if (!bmwqemu::alive) {
-        bmwqemu::start_vm or die $@;
-    }
+my ($testpid, $testfd) = autotest::start_process();
+
+init_backend();
+
+open(my $fd, ">", "os-autoinst.pid");
+print $fd "$$\n";
```
okurz: here, too

myjsonrpc.pm
```diff
@@ -0,0 +1,109 @@
...
+    my ($to_fd, $cmd) = @_;
+
+    # TODO: make this a class object
+    # allow regular expressions to be automatically converted into
+    # strings, using the Regex::TO_JSON function as defined at the end
+    # of this file.
+    my $JSON = JSON->new()->convert_blessed();
+    my $json = $JSON->encode($cmd);
+
+    #cluck "$$: send_json $json";
```
okurz: delete

myjsonrpc.pm
```diff
@@ -0,0 +1,109 @@
...
+        $hash = $JSON->incr_parse();
+        if ($hash) {
+            # remember the trailing text
+            $sockets->{$fd} = $JSON->incr_text();
+            if ($hash->{QUIT}) {
+                print "received magic close\n";
+                return;
+            }
+            use Data::Dumper;
+            #bmwqemu::diag("read_json " . Dumper($hash));
```
okurz: delete two lines

myjsonrpc.pm
```diff
@@ -0,0 +1,109 @@
...
+        $JSON->incr_parse($sockets->{$fd});
+        delete $sockets->{$fd};
+    }
+
+    my $s = IO::Select->new();
+    $s->add($socket);
+
+    my $hash;
+
+    # starting a IPMI host can take a while, so we need to be patient
```
okurz: I think this line doesn't apply here anymore

okurz:
Testing locally I did not manage to enable interactive mode, see

openqa_rework_interactive_does_not_stop

what I did:
start test run (gnome default scenario on SLES), clicked "interactive mode" as soon as grub showed up, wheel keeps on spinning.

See http://lord.arch/tests/1822 for logs.

coolo:
I will have a look as time permits

okurz:
On Monday 04 July 2016 22:08:34 Stephan Kulow wrote:
> I will have a look as time permits

As long as you don't merge "prematurely", there is time ;-)

coolo:
the 1822 test is not running with the rework_interactive_mode openQA worker. Otherwise you see the annoying /isotovideo/status in the log (and the screenshot)

coolo:
the /status lines disppear when I set the commands.pm webserver to warning instead of debug. But it's really handy at this point to know what commands were issued. And the annoying /status lines only appear in the log when you look at the live view

okurz:
* I would also like to see the commands that have been issued as they appear because of manual interaction. But the continous automatic update is not helping. Maybe use loglevel "info" by default and output only the important events, i.e. not the one that happen all the item?

* Did you also test isotovideo call without being triggered by openQA?
* Does continuation from snapshots still work?
* how can I exit the needle editor w/o saving during live?
* clicking on "stop wait for needle" during run jumps to "#details" (in firefox?)
* http://gertrude/tests/213063#live clicked on "stop" but but stopped only after like 20 more needle matches. I pressed "stop" during grub menue but it only stopped after it typed user name and such in installer
* in interactive mode waiting for needle, when clicking on "disable interactive mode" exits interactive mode but VM is still stopped vs. expected: Should continue

coolo:
14b0311 for your last point

coolo:
https://github.com/os-autoinst/openQA/pull/767/commits/3a5f6a703ad1ccf96a78364215fbdb3f44da07db for the jumping buttons

coolo:
b87da75 for annoying /status

All other points you mention are not regressions and exceed the scope of this PR

okurz:
After we have been talking yesterday I want to answer one question by myself as reference for others:

> * how can I exit the needle editor w/o saving during live?

current approach is to either rely on browser "back" button or manually edit the adress in the adress bar to go back to the test details page. coolo wants to add a "Return without save" button to do this from the UI.

> * Did you also test isotovideo call without being triggered by openQA?

It seems I found at least one problem there. If isotovideo crashes at some point it might leave processes left behind, e.g. the command server.

> * Does continuation from snapshots still work?

@coolo ?

coolo:
why would it not?

okurz:
because we can't keep all eventualities in our head

aaannz:
So far LGTM, the only thing is that this is once again update breaking openQA compatibility so needs to be updated in tandem. Can you add something like `isotovideo/version` route?

coolo:
that means isotovideo has to know its version - so far we only track that in rpm :)

aaannz:
Or just invent some not related to rpm version but API version?

coolo:
hmm, we have an API towards the worker and we have an API towards the tests. What about a plain counter incremented with every breaking change and offer this through /isotovideo and through testapi?

aaannz:
Something like that.

For different PR is however how worker should react. /isotovideo route is only available only when job is running, at that time worker is already known to openQA. Fail and unregister itself? Or send some notfication to openQA so admin can see outdated workers in worker page?

coolo:
duplicate_job and stop itself.

coolo:
now that I think about it, it only solves future problems as the current master worker won't check that route. So I assume this is actually a feature for after the merge ;)

okurz:
better now than never?

yes, a plain counter. Maybe look [here](http://zguide.zeromq.org/page:all#UDP-Beacon-Framing) for an example how others are recommending the same. Might not be the best ref but the complete is actually a recommended read.

coolo:
I have my own code to steal from ;)
https://github.com/icecc/icecream/blob/master/services/comm.h#L39

coolo:
ok, I will merge these now - and then work on the version route in os-autoinst and we can add a check to the worker afterwards.

PR #525, opened 2016-06-27
Documentation improvements

okurz:
- Get rid of subroutine prototype
- Ensure correct spelling in testapi POD
- Correct parameter doc for send_key

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -75,13 +75,17 @@ sub check_screen;
 sub type_string;
 sub type_password;
 
+
+=head1 internal
+
 =head2 init
 
 Used for internal initialization, do not call from tests.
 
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -375,7 +384,7 @@ Example:
 Returns true if screen changed or C<undef> on timeout. Default timeout is 10s.
 
 =cut
-sub wait_screen_change(&@) {
+sub wait_screen_change {
```
coolo: I think passing blocks is special - i.e. test this for real

okurz: removed this, the "no critic" line did not work because I erroneously put it into the POD, fixed.

PR #526, opened 2016-06-28
[WIP] Add floppy disk drive

dzedro:
Needed for windows 10 installtion to install scsi drivers otherwise windows can't see scsi disk

http://10.100.98.90/tests/2093/modules/win10/steps/2

https://hackweek.suse.com/14/projects/1468

Yes I tried to use usb, maybe I did something wrong but more different tries didn't windows see

PR #527, opened 2016-06-30
Revert "Add floppy disk drive"

dzedro:
Reverts os-autoinst/os-autoinst#526


dzedro:
It's OK, working, but we really don't need FDD in 2016 I guess :)

PR #528, opened 2016-07-04
Output git hash of executed tests

okurz:
If the test directory (CASEDIR) is a git repository the current commit hash is
saved just before the tests are loaded and output into the logs. The git
commit hash can be used for reference in later analysis, e.g. in openQA to
display the file not in the "most recent" state but the actual state the test
source was in at the time of execution.

If CASEDIR is not a git repo "UNKNOWN" is stored instead. A "dirty" state of
the repository is *not* detected.

Example output

```
12:29:34.6697 git hash of test distribution: c9421dc270a53d5f408e22607e437be8ae385349
12:29:34.6803 usingenv DESKTOP=textmode
12:29:34.6804 usingenv TEXTMODE=1
```

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -40,6 +40,7 @@ use testapi qw(diag);
 use Getopt::Std;
 require IPC::System::Simple;
 use autodie qw(:all);
+use Cwd;
 use POSIX qw(:sys_wait_h _exit);
 
 # avoid paranoia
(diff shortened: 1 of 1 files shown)


coolo:
just as with the sha1sum change - you need to white liste this in the appamor profile

okurz:
see https://github.com/os-autoinst/openQA/pull/769

aaannz:
:+1: , I'm fine storing this within vars.json

okurz:
@aaannz please take a look into the discussion in https://github.com/os-autoinst/os-autoinst/pull/393 where coolo disagrees with storing the value in `vars.json` and I can understand this decision. Adding and uploading one other file with just the results should be easy and can be done if we decide which one it will be.

aaannz:
I read that before commenting.

I had the similar thoughts, but I saw an use case where this can be made an actual openQA test variable. isotovideo then could checkout and test different branch/detached heads and so on. Not needed for us, true, but once we dominate the world, people will ask for this.

coolo:
Yeah. Please note that the main reason for not merging #393 was not the file, but the fact that it didn't work for shit :)

okurz:
just tell me the file name, please. I added the `chdir`. that's not the problem

coolo:
If you want the webui to display it, it should be in the jobs table. This means it has to be put into update_status - at least once. This means the worker needs an extra route - as we discussed in in #524 to offer a /isotovideo/version route, this could also output the tests git and then the worker needs to read this.

For now storing it in vars.json is good enough though.

PR #532, opened 2016-07-08
when match is equal quality, prefer workaround needle

AdamWill:
resolves openQA issue #779. When two needles both match with
the same match quality, we currently just take whichever sorts
first by name alphabetically. We would like to predictably
prefer workaround needles to non-workaround in this case, so
we rename `cmp_by_error__` to `cmp_by_error_type_` and have it
also consider the needle type and prefer workaround needles, if
the match quality is equal.


AdamWill:
BTW, I was kinda surprised at this - I thought the return values should be exactly the other way around. But doing it this way passes the test and doing it the other way doesn't. I guess I just mis-grokked something in how `sort` works.

coolo:
As you want workarounds first in the list, the compare has to return less than 0 for workarounds. a workaround needle is the smaller value so to say.

AdamWill:
eh, seems obvious now. Dunno why I was hung up on it yesterday...

coolo:
because your brain considers better values to be larger - you're part of american culture for too long ;)

AdamWill:
GO MURICA!

PR #533, opened 2016-07-08
Add a version route to commands

coolo:
curl -s
http://localhost:20113/98Xoquw0SPu7So7v/isotovideo/version | json_pp
{
   "version" : 6,
   "test_git_hash" : "5c4b7d1bc8be0b7c4a3a5bb6bc128dd6d34ee1b7"
}

The 6 is a random number, but it will increase and the worker will
require a minium version of it.

And it can transmit the test_git_hash to the webui so it can render the
correct sources even before the final vars.json is uploaded

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -23,6 +23,7 @@ use Try::Tiny;
 use Socket;
 use POSIX qw(_exit);
 use autodie qw(:all);
+use myjsonrpc;
 
 BEGIN {
     # https://github.com/os-autoinst/openQA/issues/450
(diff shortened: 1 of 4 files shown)


okurz:
* typo in commit message: minium
* Add the test to the Makefile?

coolo:
I wasn't even aware we have a fixed list in the Makefile. fixed both now

okurz:
fails now because can't find commands.pm. Add

```perl
BEGIN {
    unshift @INC, '..';
}
```

to 07-commands.pl

Ugly, yes. should be prove-ived

okurz:
After that, LGTM

coolo:
the problem we saw while deploying was btw not a minimum version problem, but a maximum. I.e. we the worker will have to restart itself when the os-autoinst version changed

PR #534, opened 2016-07-09
Rework how exit status of isotovideo is caluculated

coolo:


okurz:
don't understand it enough to give ok

AdamWill:
I don't know this code in enough detail to code review this, but it does at least appear to fix https://github.com/os-autoinst/openQA/issues/781 so far.

coolo:
I know of one problem still - which is also in master, but I want to fix this in all. If you cancel a job and isotovideo happens to have talked to the backend, it gets stuck. Which is the opposite of the behaviour where it died leaving the backend running, but it's still nasty.

coolo:
it's really scary in how strange situations you can catch isotovideo if you send it TERM at random points of time ;(

okurz:
+1

coolo:
I still have a hanging worker in 1 out of 20 job restarts, but this fixes more than it breaks and the remaining hang is hopefully caught with the new debug.

PR #536, opened 2016-07-11
Kill all children at the end

coolo:


coolo:
self merge as this is kind of pressing and everyone else is at the workshop

okurz:
LGTM (after the fact ;-) )

PR #537, opened 2016-07-11
The backend injected variables are for the tests

coolo:


coolo:
I had to extend this - and pushed it directly

aaannz:
As much as I trust you and that you know what you are doing, I find this as a bad practice. Skipping PR workflow makes it harder for me to keep at least some overview what's going on and also I used PRs as base points for release notes.

coolo:
I know - and that's why annotated the small fix. The "what's going on" story is still the same - I just did it wrong in the first run and broke 200 tests, so I went the easy way.

PR #538, opened 2016-07-12
Fix can handle

coolo:


coolo:
self merging to get zkvm back online for RC1

PR #539, opened 2016-07-12
Fix CDMODEL ide-cd

dzedro:
ide-cd have broken since 4df959b8e0f118e47a358a03777c259e417ba245 and fails with QEMU: qemu-system-x86_64: -device ide-cd,drive=cd0,bus=scsi0.0: Bus 'scsi0.0' not found

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -533,7 +533,7 @@ sub start_qemu {
             }
         }
 
-
+        my $cdbus = $vars->{CDMODEL} ne 'ide-cd' ? ',bus=scsi0.0' : '';
         if ($iso) {
             if ($vars->{USBBOOT}) {
                 push(@params, "-drive",  "if=none,id=usbstick,file=$iso,snapshot=on");
@@ -545,7 +545,7 @@ sub start_qemu {
                 # XXX: workaround for OVMF wanting to write NVvars into first FAT partition
                 # we need to replace -bios with proper pflash drive specification
                 $params[-1] .= ',snapshot=on' if $vars->{UEFI};
-                push(@params, '-device', "$vars->{CDMODEL},drive=cd0,bus=scsi0.0");
+                push(@params, '-device', "$vars->{CDMODEL},drive=cd0" . $cdbus);
             }
         }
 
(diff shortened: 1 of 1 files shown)


dzedro:
tested e.g. http://10.100.98.90/tests/2329

backend/qemu.pm
```diff
@@ -545,16 +545,20 @@ sub start_qemu {
                 # XXX: workaround for OVMF wanting to write NVvars into first FAT partition
                 # we need to replace -bios with proper pflash drive specification
                 $params[-1] .= ',snapshot=on' if $vars->{UEFI};
-                push(@params, '-device', "$vars->{CDMODEL},drive=cd0,bus=scsi0.0");
+                my $bus //= '';
```
aaannz: I don't get why are you using `//` operator. The `$bus` will be always undef this time so you can just do plain initialization `my $bus = '';`

Or fancy `my $bus = $vars->{CDMODEL} ne 'ide-cd' ? ',bus=scsi0.0' : '';`

dzedro: true,  tried to avoid but just = is enough :) I tried also similar fancy way but it didn't work at the and
```
Use of uninitialized value $bus in concatenation (.) or string at /usr/lib/os-autoinst/backend/qemu.pm line 549.
Use of uninitialized value $bus in concatenation (.) or string at /usr/lib/os-autoinst/backend/qemu.pm line 559.
```

dzedro: Updated :) have to work on my fanciness

okurz:
LGTM

backend/qemu.pm
```diff
@@ -545,16 +545,18 @@ sub start_qemu {
                 # XXX: workaround for OVMF wanting to write NVvars into first FAT partition
                 # we need to replace -bios with proper pflash drive specification
                 $params[-1] .= ',snapshot=on' if $vars->{UEFI};
-                push(@params, '-device', "$vars->{CDMODEL},drive=cd0,bus=scsi0.0");
+                my $bus = $vars->{CDMODEL} ne 'ide-cd' ? ',bus=scsi0.0' : '';
```
coolo: seriously? It has to be moaning about this line being duplicated? :)

I recommend adding a add_cdmodel_device(\@params,  $i) function.

okurz: you are right :-)

dzedro:
@coolo what about creating $bus outside of if conditions ?

coolo:
that looks even more wrong - as $bus will be wrong for many other if/else paths.

backend/qemu.pm
dzedro: I could imagine how to do it if the push(@params, *) would be same in both cases
Now I have no idea what you mean

dzedro:
This is the best option

aaannz:
:+1:

PR #541, opened 2016-07-18
Stop the VM more reliable

coolo:
If isotovideo is TERMinated, we need to make sure to exit the loop
no matter when it happened - and stop the VM at the end of it

https://progress.opensuse.org/issues/12566

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -104,6 +104,17 @@ sub kill_autotest {
     $testpid = 0;
 }
 
+sub kill_backend {
+    bmwqemu::stop_vm();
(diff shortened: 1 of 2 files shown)


okurz:
LGTM, wouldn't mind tests as part of os-autoinst tests, though

coolo:
I wouldn't mind them either. Perhaps you know someone in QA...

AdamWill:
I'm on vacation for the next week and a half, FWIW - I may find a couple of hours to update our staging deployment to whatever @coolo has cooked up lately some day, but no promises. Our production is still running the Before Coolo code at this point.

sysrich:
BC - Before Coolo..

PR #542, opened 2016-07-19
Comment myjsonrpc debug for now

coolo:
Hopefully we can remove it when we survived this week without loss
of workers :)


okurz:
don't keep disabled code, you have git for that. But who am I telling anyway…

coolo:
no lo sé

PR #543, opened 2016-07-20
Expect fail with wait_serial

dzedro:


aaannz:
I'm confused. Isn't this something expect_not_found should handle? Do you have an example?

aaannz:
Ah, didn't notice link to the test. However I think this is indeed covered by expect_not_found. And if it isn't, you really should add this as the last parameter otherwise you are breaking api for existing tests.

okurz:
or better convert to using hash for optional parameters, e.g. see https://github.com/os-autoinst/os-autoinst/pull/502

dzedro:
expect_not_found is doing the same

PR #546, opened 2016-07-22
testapi: Add check_act_and_assert helper method

okurz:
'check_act_and_assert' can be used wherever check_screen is used to conduct an
optional action before an assert_screen, e.g. to check for an optional close
dialog. The approach to check for multiple tags in one assert_screen and check
if an optional needle has been matched to conduct an optional action can help
save time as the check_screen call does not run into a timeout if the
optional action is not necessary to be executed.

---

This change is [Reviewable](https://reviewable.io/reviews/os-autoinst/os-autoinst/546)

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -40,7 +40,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
 
   assert_screen check_screen assert_and_dclick save_screenshot
   assert_and_click mouse_hide mouse_set mouse_click
-  mouse_dclick mouse_tclick match_has_tag
+  mouse_dclick mouse_tclick match_has_tag check_act_and_assert_screen
 
   script_run script_sudo script_output validate_script_output
   assert_script_run assert_script_sudo
@@ -263,6 +263,32 @@ sub match_has_tag {
         return $last_matched_needle->{needle}->has_tag($tag);
     }
     return;
+}
+
+=head2 check_act_and_assert_screen
+
+  check_act_and_assert_screen($mustmatch, $check_tag, $opt_action[, $timeout]);
+
+Helper method to check on multiple needles specified in C<$mustmatch> and
+C<$check_tag> and executes C<$opt_action> if C<$check_tag> found. C<$timeout>
+is forwarded to C<assert_screen>.
+
+This method can be used whenever an optional action might be necessary to
+execute before a subsequent C<assert_screen>.
+
+Example:
+
+  check_act_and_assert_screen('clean-desktop', myapp-unsaved-changes => send_key('esc'));
+
+=cut
+sub check_act_and_assert_screen {
+    my ($mustmatch, $check_tag, $opt_action, $timeout) = @_;
+    my $ret = assert_screen([@{$mustmatch}, $check_tag], $timeout);
+    if (match_has_tag($check_tag)) {
+        eval { &$opt_action };
+        $ret = assert_screen($mustmatch, $timeout);
+    }
+    return $ret;
 }
 
 =head2 assert_and_click


okurz:
See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1554 for an example how it could be used

coolo:
I don't find the API intuitive. To start with the name. check_act_and_assert sounds to me like 'go to the theater and assert the costumes are proper'. If the function can die (which it can as it is a wrapped assert_screen) it should start with assert_ to be in line with our other assertions. E.g. assert_screen_conditional ?

And I wonder if the API shouldn't offer multiple options as a proper hash.

okurz:
Yes, I am also not happy with the name. That is also the reason why this change kept lingering around for so long without being applied. But `assert_screen_optional` sounds like a variant to assert_screen which it is not. I understand the reason why you want the `assert_`-prefix but `check_act_and_assert` is a totally new function using internally assert_screen, match_has_tag and takes an action, i.e. a `sub` for the optional action. Maybe change assert_screen options alltogether? See below

> And I wonder if the API shouldn't offer multiple options as a proper hash.

Do you mean

```perl
check_act_and_assert_screen('clean-desktop', {'myapp-unsaved-changes' => send_key('esc')});
```
? Wouldn't that be the opposite of what you told in https://github.com/os-autoinst/os-autoinst/pull/502 ?

Maybe what you had in mind or what comes to my mind right now is to offer all this as additional functionality to `assert_screen`? E.g. support all of these

```perl
assert_screen('clean-desktop', 90);
assert_screen('clean-desktop', {'myapp-unsaved-changes' => send_key('esc')});
assert_screen('clean-desktop', timeout => 90);
assert_screen('clean-desktop', {'myapp-unsaved-changes' => send_key('esc')}, timeout => 90);
```

coolo:
```
assert_screen(['clean-desktop',
               { 'myapp-unsaved-changes' => sub { send_key('esc') } },
               { 'total-desaster' => sub { die 'oops'; } }
              ], 90);
```

Please note that 'foo' => send_key('esc') will create a hash pair with 'foo' and the return value of send_key.

okurz:
so you want to keep just two arguments to assert_screen and check if any `@tags` entry is a hash?

coolo:
I was just playing with your proposal. This whole function is getting very perl-expertish and I'd think this is the wrong trend. Most test writers won't have an idea what they're doing by putting sub {} in a function call.

coolo:
To summarize: I think this makes it *harder* to write proper tests, not easier (I would take your changes to distri-opensuse mostly being invalid as proof) - so I object.

okurz:
yes, looks perl-expertish - which is weird mentioning this in combination with a change *I* propose - but the problem I want to solve is that a manual tester would not do

* look at the screen if a warning message appears and ignore all others for 30 seconds (`check_screen`)
* then look for what you really expect (`assert_screen`)

so I want to make this construct of checking for multiple intermediate alternative ways more obvious and easier to use. Of course, defining subs within the function call can look weird at first but this is why the function is there: To put everything in a way that works if you get it to run. And my changes to `o-a-d-o` were wrong because the `assert_screen` extension is still wrong and I am not a perl expert but when `assert_screen` can do more the right way it will help the user.

Compare the cleanliness of

```perl
assert_screen(['clean-desktop',
               { 'myapp-unsaved-changes' => sub { send_key('esc') }}]);
```

to

```perl
assert_screen([qw/clean-desktop myapp-unsaved-changes/]);
if (match_has_tag('my-app-unsaved-changes')) {
    send_key('esc');
    assert_screen('clean-desktop');
}
```

For the *first one* the developer has to understand why there is a `sub` within the function call, granted. For the *second one* the developer has to learn what `match_has_tag` does, he has to understand how assert_screen handles multiple needle tags, he has to learn that there must be an assert_screen on 'clean-desktop' within the if again. All this for a simulation how a manual tester would really work which the testapi should offer by itself.

okurz:
Or to rephrase: The testapi should IMHO actually not have a `check_screen` at all, only `assert_screen`. Are there any circumstances where you want to check for a needle that is defined by check_screen where you can't find a needle that you actually expect?

coolo:
your two code snippets are not doing the same thing - and this is also what makes the whole approach so complicated. In the explict example you only match clean-desktop, in the wrapped variant you also go back to assert_screen on myapp-unsaved-changes. But you don't do anything if it matches. And depending on the circumstances you might want to remove the trigger from the tags or not.

Deprecation of check_screen sounds very dangerous to me as it will lead to more use of stupid needles. You know how an error popup looks like in a yast module - it's easy to needle. But how a yast dialog looks like that does not show a popup is usually a bad needle as you need to match the areas usually covered by the error popup but these will contain data you don't want to pollute your needles with.

aaannz:
Commenting on the two examples I like the current way better. It is, at least for me, much more clearer what and when I want to achieve
I can already see how hard it will be to explain this function properly to new comers.
Explaining them that `assert_screen` accepts multiple needles is IMO easy.

coolo:
We still should think if we can improve the API in a less obscure way - because I agree with okurz's sentiment that doing the right thing should be easier than it is currently (I just disagree that the proposal *is* easier :)

coolo:
If I had a DSL and not perl, I guess I'd want to write

```
assert_screen {
    final => 'clean-desktop',
    onmatch => {
       'myapp-unsaved-changes' => {
           run => sub {
               send_key('esc');
            },
           await_final => 1
        }
    }
}

**Comment by coolo** [comment]
ok, this *is* perl :)

**Comment by okurz** [comment]
*sigh* so you want me to do *what* now?

**Comment by okurz** [comment]
> Deprecation of check_screen sounds very dangerous to me as it will lead to more use of stupid needles. You know how an error popup looks like in a yast module - it's easy to needle. But how a yast dialog looks like that does not show a popup is usually a bad needle as you need to match the areas usually covered by the error popup but these will contain data you don't want to pollute your needles with.

Then prevent stupid needles by not allowing so many people to needle in live instance without review. It shouldn't be too hard to create automatic pull requests when saving needles from the live instance instead of directly pushing to the master branch. I want users to try `assert_screen`+`@tags`+`match_has_tag` *first* before they resort to `check_screen` and waste time by looking at something that is not there or even worse break their head about how long the timeout should be not to mess up a test looking at a SUT which changes *within* the timeout, see https://progress.opensuse.org/issues/12628 and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1559#discussion_r72047639 for examples.

**Comment by aaannz** [comment]
What about variants:

`assert_screen 'needleX'`
`assert_screen 'needleX', onmatch 'needleY' => sub {..}, onmatch 'needleZ' => sub {..}, ...`

where `onmatch` is actually another sub which just fill some internal hash structure used by, and emptied at the end, of assert_screen?

**Comment by coolo** [comment]
Just for me: that is not perl, right? You never know with this language

**Comment by coolo** [comment]
@okurz what I want you to do is continue thinking about the problem instead of jumping at the first solution crossing your mind

**Comment by aaannz** [comment]
> Just for me: that is not perl, right? You never know with this language 

Actually no, this is how it should looks like in perl.

Example implementation:

```perl
use strict;
use warnings;

my %needle_table = ();
my $match;

sub match_has_tag {
  $match = 'Z';
  return 0;
}

sub onmatch {
  my $needle = shift;
  my $action = shift;
  $needle_table{$needle} = $action;
}

sub assert_screen
{
  my $needle = shift;
  my @needles = ($needle, keys %needle_table);
  #assert stub
  if (match_has_tag($needle)) {
    return;
  }
  else {
    $needle_table{$match}->();
  }
  %needle_table = ();
}

assert_screen 'X', onmatch 'Y' => sub { print "matched Y\n"; }, onmatch 'Z' => sub {print "matched Z\n"};
```

Just manually change `onmatch` return and $match to trigger X or Y or Z match.

okurz:
I will probably almost always jump to the first solution I have in my head and adapt it as soon as I hit problems or you guys ;-) It's all good I just wanted to make sure you understand my motivation :-)
That solution by @aaannz or the DSL proposal by @coolo is fancy at least but so far I don't see where this is clearer than what I have hingestümpert

okurz:
@coolo you like the solution proposal by @aaannz ?

coolo:
the api looks okayish (leaves to be decided where to put the timeout) - but the implementation detail seems really dangerous to me

aaannz:
It is only example to prove this syntax can work. To solve the timeout option `onmatch` function must explicitly do `return @_;` that way you can put `timeout => XXX` anywhere there.

Inspiration for this was the way how Mojo templates functions are written. Don't know if the implementation is the same, just appearance.

okurz:
so you just wanted to discuss it to death? Maybe we can take my original proposal and make it a bit safer, e.g. check that the arguments are a real sub, not a function already evaluated to 0/1

aaannz:
I would really love to have one assert_screen with advanced options which fit seamlessly. Which is  something you probably not wanted to hear. :D
I'm not sure what @coolo meant by dangerous implementation details, probably the use of external data structure making the nesting of assert_screen really tricky, but that may be possible to overcome.

okurz:
I am fine with adding more options to `assert_screen` instead of adding another sub. So I will try to follow your proposal at some time and come back with code changes. What I don't understand is why I can not just rename `assert_screen` to `_assert_screen` and create a new `assert_screen` calling the old `assert_screen` plus more.

okurz:
open for too long. Closing to cleanup open PRs.

PR #547, opened 2016-07-25
Support arbitrary options for '-netdev user'

mnowaksuse:
For FATE#319639 we need to test that YaST2 uses in installation hostname
from network. One way to test it is by -netdev user,hostname=MyHostName
Qemu option.

Test case:

    start qemu with -netdev user,hostname=
    start installation ifcfg=*=dhcp
    check whether hostname is set to 'myhostname1'

Verification:

    # ps aux | grep hostnameXXX
    _openqa+ 13293 13.1  0.1 5074336 45956 ?       Sl   16:35   0:00
    /usr/bin/qemu-system-x86_64 -serial file:serial0 -soundhw ac97
    -global
    isa-fdc.driveA= -vga cirrus -m 4096 -cpu qemu64 -netdev
    user,id=qanet0,hostname=myhostnameXXX

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -482,7 +482,11 @@ sub start_qemu {
 
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
-                push(@params, '-netdev', "user,id=qanet$i");
+                my $nictype_user_options;
+                if ($vars->{NICTYPE_USER_OPTIONS}) {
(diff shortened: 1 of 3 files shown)


okurz:
LGTM

coolo:
fine for me too.

okurz:
@coolo can you then please merge because I can't?

coolo:
I wanted to give some more time to review - it's not like we're going to deploy any minute

PR #548, opened 2016-07-26
Check if $is_shutdown is actually a number in order to avoid warnings…

DimStar77:
… when

comparing strings ('') to numeric values.

This resolves warnings on shutdown like:
Argument "" isn't numeric in numeric lt (<) at /usr/lib/os-autoinst/testapi.pm line 775.


okurz:
LGTM

coolo:
I prefer https://github.com/os-autoinst/os-autoinst/pull/544 though

DimStar77:
oh, hadn't seen that variant - looks also good of course... let's go with #544

PR #549, opened 2016-07-26
Update install guide to include more information

richiejp:
Add more information about dependencies and configuration.
It also changes the minimal configuration.

### DIFF
--- a/INSTALL.asciidoc
+++ b/INSTALL.asciidoc
@@ -13,8 +13,7 @@ installation guide]
 git clone YOUR-WRITEABLE-COPY-OF/os-autoinst.git
 cd os-autoinst
 ------------------------------------------------
-3.  install the BuildRequires, as per the http://build.opensuse.org[spec
-file]
+3.  install the BuildRequires, as per the https://build.opensuse.org/package/show/devel:openQA/os-autoinst[spec file] 
 +
 -----------------------------------------------------------------------------------------------------------------------------------
 sudo zypper in git-core make libtheora-devel opencv-devel fftw3-devel libsndfile-devel pkg-config libtool autoconf automake gcc-c++
--- a/INSTALL.asciidoc
+++ b/INSTALL.asciidoc
@@ -44,15 +44,27 @@ cd /tmp/os-autoinst-run
 --------------------------
 7.  create a minimal "vars.json" config file
 +
+.vars.json
+[source,json]
 ---------------------------------------------------------
-cat <<EOF >vars.json
 {
-    "DISTRI":  "sle",
-    "CASEDIR": "/space/SVN/os-autoinst-distri-opensuse/",
-    "BACKEND": "s390x"
+   "ARCH" : "x86_64",
+   "BACKEND" : "qemu",
+   "CASEDIR" : "/path/to/os-autoinst-distri-opensuse",
+   "DESKTOP" : "gnome",
+   "DISTRI" : "opensuse",
+   "ISO" : "/path/to/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20160715-Media.iso",
+   "PRODUCTDIR" : "/path/to/os-autoinst-distri-opensuse/products/opensuse",
+   "VNC" : 90,
 }
-EOF
 ---------------------------------------------------------
++
+You will need to correct the file paths to point to real locations. Some of
+the variables you can use are listed
+https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc[here].
+Test case specific variables are listed in the distri directories
+e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/variables[os-autoinst-distri-opensuse/variables].
++
 8.  give it a shot
 +
 ------------------------------


INSTALL.asciidoc
```diff
@@ -2,19 +2,30 @@
...
++
+-----------------------------------------------------------------------
+zypper qemu tesseract optipng opencv tesseract-ocr sqlite3 mumble     \
+       libtheora-devel opencv-devel fftw3-devel libsndfile-devel      \
+       pkg-config libtool autoconf automake gcc-c++ make perl-JSON    \
+       perl-Mojolicious perl-Mojolicious-Plugin-AssetPack             \
+       perl-IPC-System-Simple perl-Exception-Class perl-Test-ANSIColor\
+       perl-Test-Compile perl-Test-More perl-Test-Simple perl-XML     \
+       perl-carp perl-Class-Accessor perl-Net-SSH2 perl-Crypt-DES
+-----------------------------------------------------------------------
```
okurz: While I value your approach to update the documentation by adding this explicit list a new update will be necessary pretty soon and I would not recommend to duplicate the requirements in a hand-written document. A better approach would be to rely on the tools to tell the user what's missing, in this case it's autotools. Of course, it would be nice to rewrite the automake approach to something more modern if that works, e.g. either more perlish Makefile stuff or CMake :-)

INSTALL.asciidoc
```diff
@@ -25,16 +36,20 @@ sudo zypper in libtheora-devel opencv-devel fftw3-devel libsndfile-devel pkg-con
 autoreconf -f -i
 ./configure
 make
+make install
```
okurz: I would mark that as optional. As having `os-autoinst` usable is basically having isotovideo available it is good enough to just build it and use the isotovideo as is after `make`.

INSTALL.asciidoc
```diff
@@ -25,16 +36,20 @@ sudo zypper in libtheora-devel opencv-devel fftw3-devel libsndfile-devel pkg-con
...
 ----------------
++
+If you do not run `make install` then you need to ensure Perl can find
+`ppmclibs/blib/arch/auto/tinycv.so` and `ppmclibs/tinycv.pm` some other way.
```
okurz: should work out of the box if you call isotovideo where it is, at least does for me and I did not add anything special to an LD_LIBRARY_PATH or so

INSTALL.asciidoc
```diff
@@ -25,16 +36,20 @@ sudo zypper in libtheora-devel opencv-devel fftw3-devel libsndfile-devel pkg-con
...
 5.  get the test cases and the needles.
 +
 The needles have to be a subdir of the test cases!
 +
 -----------------------------------------------------------------------------
 mkdir distri
 cd distri
-git clone git@github.com:os-autoinst/os-autoinst-distri-opensuse.git opensuse
-cd opensuse
+git clone git@github.com:os-autoinst/os-autoinst-distri-opensuse.git
```
okurz: I would really recommend to keep the alias "opensuse" intact as this is commonly used when referring to the test distribution with the `DISTRI` variable

INSTALL.asciidoc
```diff
@@ -45,15 +60,27 @@ cd /tmp/os-autoinst-run
 --------------------------
 7.  create a minimal "vars.json" config file
 +
+.vars.json
+[source,json]
 ---------------------------------------------------------
-cat <<EOF >vars.json
```
okurz: why delete this? If you keep it as an executable command it's easier to use for shell users

INSTALL.asciidoc
```diff
@@ -45,15 +60,27 @@ cd /tmp/os-autoinst-run
...
 {
-    "DISTRI":  "sle",
-    "CASEDIR": "/space/SVN/os-autoinst-distri-opensuse/",
-    "BACKEND": "s390x"
+   "ARCH" : "x86_64",   
+   "BACKEND" : "qemu",
+   "CASEDIR" : "/path/to/os-autoinst-distri-opensuse",
+   "CDMODEL" : "virtio-scsi-pci",
```
okurz: That might be a bit too special for a beginners document, don't you think?

INSTALL.asciidoc
```diff
@@ -45,15 +60,27 @@ cd /tmp/os-autoinst-run
...
+   "VNC" : 90,
 }
-EOF
 ---------------------------------------------------------
++
+You will need to correct the file paths to point to real locations. Some of
+the variables you can use are listed
+https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc[here].
+Test case specific variables are listed in the distri directories
+e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/variables[os-autoinst-distri-opensuse/variables].
```
okurz: I am pretty sure this is already mentioned somewhere else and I don't think repeating it here is fitting

INSTALL.asciidoc
```diff
@@ -67,7 +94,8 @@ use of git repo with with openQA
 
 to use os-autoinst with openQA it needs to be "installed". Placing a
 symlink to /usr/lib/os-autoinst that points to the git checkout should
-work. To make tests and needles writable don't touch distri/ and use
+work, as well as symlinking isotovideo, debugviewer and snd2png to /usr/bin.
```
okurz: No, you shouldn't do this. Either use from local paths or install it with the proper distribution utilities, e.g. `zypper`

okurz:
Thank you for your updates. Please see the inline comments.

INSTALL.asciidoc
richiejp: What is the correct way of doing this? Running `autogen.sh` I guess, which I may not have done. Instead I probably just ran `autoreconf` and then forgot about it. I have used SCons most recently for a build system and it works OK, but I don't fancy being the one to try porting OpenQA to it. There is also the predecessor to SCons: Cons, which is written in Perl.

INSTALL.asciidoc
richiejp: Perhaps you ran the autools script correctly and I did not.

INSTALL.asciidoc
richiejp: It is does not work in Fish and the user will need to change it anyway to correct the file paths, which they will probably do in an editor. So I think that it mainly serves as a distraction, but I suppose most OpenQA users will not be bothered by it either way.

INSTALL.asciidoc
richiejp: I am a beginner, this is the config which I found worked for me. A simpler starting point would be welcome :-)

INSTALL.asciidoc
richiejp: I don't know where that is documented. If it is documented somewhere else then this guide should link to it either at the bottom in further reading or inline. Perhaps you are referring to this: https://github.com/os-autoinst/openQA/blob/master/docs/WritingTests.asciidoc#variables

I think that there needs to be an explicit statement connecting vars.json and the variables passed from within OpenQA because I am still not certain about this.

INSTALL.asciidoc
aaannz: I agree with @okurz to not explicitly add this list. We add new deps all the time.
There are `zypper if --requires os-autoinst` and `zypper if -t srcpackage --requires os-autoinst` calls you can use to get required capabilities for both building and running.

INSTALL.asciidoc
aaannz: Given your commit 5e5d098d this line should be omitted.

INSTALL.asciidoc
aaannz: I'm fine having it here. The reasoning by @richiejp is valid IMO.

INSTALL.asciidoc
richiejp: I think most of it can be omitted now that I know a bit more about os-autoinst. Also I think it would be more appropriate to use something like: https://github.com/os-autoinst/os-autoinst-distri-example/ here because this guide should stay focused on the backend to limit overlap with the OpenQA documentation.

okurz: sure, would make sense. I only see the problem that os-autoinst-distri-example 
is less tested and used. But feel free to provide a suggestion how to update 
the documentation.

okurz:
*bump*

richiejp:
Sorry, I am having trouble with this, so will just close it. It looks like you have already made some changes which make the uncontested parts of this redundant anyway.

okurz:
well, you should be able to rebase.

richiejp:
It is the subject matter and document format which is the problem. I will give it one last shot though.

PR #550, opened 2016-07-26
Move the verification of iamges part to the backend

nilxam:
Move verify the images to the backend and also use needle:tags in the  backend. verify_sound_image in basetest.pm just to do post-work.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -900,6 +900,22 @@ sub last_screenshot_name {
     return {filename => $self->write_img($self->last_image, $self->_last_screenshot_name)};
 }
 
+sub verify_image {
+    my ($self, $args) = @_;
+    my $imgpath   = $args->{imgpath};
+    my $mustmatch = $args->{mustmatch};
+
+    my $img = tinycv::read($imgpath);
+    my $needles = needle::tags($mustmatch) || [];
+
+    my ($foundneedle, $failed_candidates) = $img->search($needles, 0, 1);
+    if ($foundneedle) {
+        return {found => $foundneedle, candidates => $failed_candidates};
+    }
+    return {candidates => $failed_candidates};
+}
+
+
 sub retry_assert_screen {
     my ($self, $args) = @_;
 
--- a/basetest.pm
+++ b/basetest.pm
@@ -125,9 +125,12 @@ sub record_screenmatch {
         json       => $h->{json},
     };
 
+    # make sure needle is blessed
+    my $foundneedle = bless $match->{needle}, "needle";
+
     # When the needle has the workaround property,
     # mark the result as dent and increase the dents
-    if ($match->{needle}->has_property('workaround')) {
+    if ($foundneedle->has_property('workaround')) {
         $result->{dent} = 1;
         $self->{dents}++;
         bmwqemu::diag("needle '$h->{name}' is a workaround");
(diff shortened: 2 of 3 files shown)


coolo:
But you do more - you also create the $backend variable and leak it into the autotest process.

nilxam:
indeed. updated.

coolo:
you didn't really motivate this change though - why are you touching it?

nilxam:
a lazy way to fix firefox_audio missing needle issue, and I wondering needle::tag() call really works currently, looks like needle::all gives 0.

note. my instance has patched a lot of different place though...

coolo:
wait a second. this means the firefox needle is compared in the autotest process? We need to fix that. Only the backend reloads the needles. So instead of working around it, make sure verify_sound_image is actually working on filenames passed to the backend.

nilxam:
knock! knock!

okurz:
can you fix the typo s/iamges/images/ in the commit message and please describe a bit more what this is intended to do? You are describing what the change is doing but not why and how this fixes the firefox sound test I guess? Can you provide a link to a local verification run?

backend/baseclass.pm
```diff
@@ -900,6 +900,22 @@ sub last_screenshot_name {
...
+    my $mustmatch = $args->{mustmatch};
+
+    my $img = tinycv::read($imgpath);
+    my $needles = needle::tags($mustmatch) || [];
+
+    my ($foundneedle, $failed_candidates) = $img->search($needles, 0, 1);
+    if ($foundneedle) {
+        return {imgpath => $imgpath, found => $foundneedle, candidates => $failed_candidates};
+    }
+    return {imgpath => $imgpath, candidates => $failed_candidates};
```
aaannz: I don't get why are you returning the same `$imgpath`? I'm probably missing the whole context, but looks weird.

basetest.pm
```diff
@@ -127,10 +127,12 @@ sub record_screenmatch {
...
-    if ($match->{needle}->has_property('workaround')) {
-        $result->{dent} = 1;
-        $self->{dents}++;
-        bmwqemu::diag("needle '$h->{name}' is a workaround");
+    for my $t (@${properties}) {
+        if ($t eq 'workaround') {
+            $result->{dent} = 1;
+            $self->{dents}++;
+            bmwqemu::diag("needle '$h->{name}' is a workaround");
+        }
```
aaannz: What is the reason for this change? `$match->{needle}` is no longer an object?

backend/baseclass.pm
nilxam: ah, yes, it's superfluous.

basetest.pm
nilxam: bless/unbless problem, I can bless it if preferred.

nilxam:
http://147.2.211.60/tests/382 local verification run

isotovideo
```diff
@@ -349,6 +349,19 @@ while ($loop) {
                     }})->{tags};
             next;
         }
+        if ($rsp->{cmd} eq 'verify_sound_image') {
```
coolo: please note. Just prefix the query with backend_ and let isotovideo pass it as is

coolo: aehm, please not

nilxam:
updated. and http://147.2.211.60/tests/386 for local verification run

coolo:
perfecto!

okurz:
awesome, thank you

PR #552, opened 2016-07-27
Extend loadtest to schedule same module multiple times

okurz:
Previously `loadtest` would trigger a test module only once as a latter call
to `loadtest` with same argument overwrites a previous call. This prevents
reuse of test modules which should appear multiple times in the same schedule.
An example where one would like to apply this is to call a test module in a
system, then upgrade the system and call the same previous test module again.

Example output when called from isotovideo context:

scheduling consoletest_setup tests/console/consoletest_setup.pm
scheduling check_console_font tests/console/check_console_font.pm
scheduling textinfo tests/console/textinfo.pm
scheduling hostname tests/console/hostname.pm
console-check_console_font already scheduled
scheduling check_console_font1 tests/console/check_console_font.pm
scheduling zypper_lr tests/console/zypper_lr.pm

If this approach works depends on how the test modules are written of course.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -46,33 +46,33 @@ sub loadtest {
     my $name     = $2;
     my $test;
     my $fullname = "$category-$name";
-    if (exists $tests{$fullname}) {
-        $test = $tests{$fullname};
-        return unless $test->is_applicable;
+    # perl code generating perl code is overcool
+    # FIXME turn this into a proper eval instead of a generated string
+    my $code = "package $name;";
+    $code .= "use lib '$casedir/lib';";
+    my $basename = dirname($script);
+    $code .= "use lib '$casedir/$basename';";
+    $code .= "require '$casedir/$script';";
+    eval $code;    ## no critic
+    if ($@) {
+        my $msg = "error on $script: $@";
+        bmwqemu::diag($msg);
+        die $msg;
     }
-    else {
-        # perl code generating perl code is overcool
-        # FIXME turn this into a proper eval instead of a generated string
-        my $code = "package $name;";
-        $code .= "use lib '$casedir/lib';";
-        my $basename = dirname($script);
-        $code .= "use lib '$casedir/$basename';";
-        $code .= "require '$casedir/$script';";
-        eval $code;    ## no critic
-        if ($@) {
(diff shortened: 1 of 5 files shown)


aaannz:
I completely forgot about this, I did similar thing during openQA hackaton for @nadvornik Only I opted to handle naming on openQA side and os-autoinst patch was then trivial
```diff
diff --git a/autotest.pm b/autotest.pm
index 7ed3025..8ea5270 100644
--- a/autotest.pm
+++ b/autotest.pm
@@ -67,8 +67,8 @@ sub loadtest {
         $tests{$fullname} = $test;

         return unless $test->is_applicable;
-        push @testorder, $test;
     }
+    push @testorder, $test;
     bmwqemu::diag "scheduling $name $script";
 }
```

This looks good however.

coolo:
why didn't travis run for this? any theory?

autotest.pm
```diff
@@ -46,33 +46,32 @@ sub loadtest {
...
-        $tests{$fullname} = $test;
-
-        return unless $test->is_applicable;
-        push @testorder, $test;
+    $test             = $name->new($category);
+    $test->{script}   = $script;
+    $test->{fullname} = $fullname;
+    my $nr = '';
+    while (exists $tests{$fullname . $nr}) {
+        ++$nr;
```
coolo: can you please not use ++ on strings. This is heavy abuse of perl's insanity.

okurz: *ab*use of insanity? I thought that's the reason we *use* perl? It's doing exactly what I want here. How else would I do "<name>, <name>1, <name>2"?

coolo: just change this line with

```
# to all perl hardcore hackers: fuck off!
$nr = $nr eq '' ? 1 : $nr + 1;

coolo: and the reason we use perl is because bernhard started openqa in it - that's about the complete list of reasons :)

okurz:
updated as you suggested, travis CI triggered now, too. Don't know why it didn't in before. My theory is that a very unprobable network routing issue paired with race condition in some github internal retry handler of git hook event handling caused the event to be lost and this will never happen again as it is so  unlikely.

PR #556, opened 2016-08-03
 Specify SCSI controller independantly from CDMODEL and HDDMODEL

richiejp:
Introduces the SCSICONTROLLER backend var which allows the user to
explicitly set the SCSI controller, HDD and CD models/devices
independantly of each other. The old behaviour of setting the controller
type through the CDMODEL or HDDMODEL variables is still supported, but
deprecated. Now these variables should always be set to their actual
storage device type.

When the user sets CDMODEL to scsi-cd and/or HDDMODEL to scsi-hd the
SCSICONTROLLER var will default to virtio-scsi-pci if it has not already
been specified. For more exotic SCSI devices the user will need to ensure
SCSICONTROLLER has been set correctly.

This update should resolve https://progress.opensuse.org/issues/12848

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -288,26 +288,30 @@ sub start_qemu {
     # disk settings
     $vars->{NUMDISKS}  ||= 1;
     $vars->{HDDSIZEGB} ||= 10;
-    $vars->{CDMODEL}   ||= "virtio-scsi-pci";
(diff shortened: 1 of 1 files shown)


sysrich:
I really wanted to hate this pull request... but it LGTM :+1:

PR #558, opened 2016-08-08
consoles: vnc_base: Make sure 'mouse_set' can also be called to hide in corner

okurz:


okurz:
I didn't think more deeply about this if it makes sense but could help with problems like https://tortuga.suse.de/tests/390#step/xfce_lightdm_logout_login/8

coolo:
I think providing defaults here is wrong. but changing if ($x) to if (defined $x) and providing a more specific warning is more accurate

okurz:
"a more specific warning" should be a `die` because otherwise we will ignore it, right?

coolo:
if you ask me to choose: a die would be better than a silent return

okurz:
upated with explicit die

aaannz:
LGTM

aaannz:
Since  this can break some tests let's discuss when to merge this on next call.

PR #559, opened 2016-08-09
assert_and_click: handle undefined previous mouse position

AdamWill:
assert_and_click tries to record the previous mouse cursor
position and return the cursor there after clicking. However if
the mouse has never been explicitly positioned anywhere before,
`backend_get_last_mouse_set` returns undefined for each
co-ordinate, which results in a couple of 'use of uninitialized
value in int' warnings and the cursor being set to 0,0, which
may trigger the GNOME overview under Wayland. So catch when the
values are undefined and in this case, call mouse_hide() - which
puts the cursor at bottom-right - instead. This avoids both the
warnings and the overview trigger.

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -188,9 +188,11 @@ sub mouse_hide {
     $self->{mouse}->{x} = $self->{vnc}->width - 1;
     $self->{mouse}->{y} = $self->{vnc}->height - 1;
 
-    my $border_offset = int($args->{border_offset});
-    $self->{mouse}->{x} -= $border_offset;
-    $self->{mouse}->{y} -= $border_offset;
+    if (defined $args->{border_offset}) {
+        my $border_offset = int($args->{border_offset});
+        $self->{mouse}->{x} -= $border_offset;
+        $self->{mouse}->{y} -= $border_offset;
+    }
 
     bmwqemu::diag "mouse_move $self->{mouse}->{x}, $self->{mouse}->{y}";
     $self->{vnc}->mouse_move_to($self->{mouse}->{x}, $self->{mouse}->{y});
--- a/testapi.pm
+++ b/testapi.pm
@@ -311,8 +311,14 @@ sub assert_and_click {
     }
     # We can't just move the mouse, or we end up in a click-and-drag situation
     sleep 1;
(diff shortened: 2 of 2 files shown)


consoles/vnc_base.pm
```diff
@@ -188,9 +188,11 @@ sub mouse_hide {
     $self->{mouse}->{x} = $self->{vnc}->width - 1;
     $self->{mouse}->{y} = $self->{vnc}->height - 1;
 
-    my $border_offset = int($args->{border_offset});
```
coolo: you could have fixed this by just passing a || 0 within the int cast - the -= 0 wouldn't have hurt. 

But of course your version is spelling it out with more whitespace :)

AdamWill: yeah, I dunno, it just feels better to me this way. either works.

AdamWill:
I suppose there's one possible risk here which is that moving the mouse to bottom-right may trigger a tooltip on some desktops with a traditional bottom panel, which *may* interfere with some needle match...I'll see if that happens on any Fedora tests, but you (SUSE) probably have a lot more than us which may be affected.

sysrich:
I think this is a 'safe' change (or to put it another way, I think @AdamWill is worrying too much ;))

we use mouse_hide all the time and I can never think of a time it causes a tooltip or hotcorner activation, makes sense to me to use it if assert_and_click has nowhere else to go.

AdamWill:
just for amusement's sake - I **just** realized this exact patch caused a problem I've been banging my head against for weeks:

https://openqa.fedoraproject.org/tests/32777

In that test (and a few others) we run Firefox in a bare X session (no window manager at all, just `startx /usr/bin/firefox`) to test a webapp. When you do that, at least on Fedora, you get about an 800x600 Firefox window in a 1024x768 display, and some dead black space to the right and bottom. I **just** figured out that if the cursor is resting in the dead black space, keypresses don't reach the browser. So our tests which type into the browser now break after the first `assert_and_click`, which moves the mouse to bottom-right...

Now I know what the bug is I can work around it easily, of course. Just thought it was funny that it drove me nuts for weeks and I spent all last night fiddling around to try and figure it out and it turned out to be my own damn fault all along...

okurz:
glad you like your own changes ;-)

PR #560, opened 2016-08-13
Fix assert_script_run with just one named argument

coolo:
The error has been in travis output since like forever:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/152011606/log.txt
-> Argument "fail_message" isn't numeric in multiplication (*) at
../bmwqemu.pm line 329.


okurz:
LGTM

PR #561, opened 2016-08-15
Initialize $nictype_user_options to avoid undefined warning

aaannz:


backend/qemu.pm
```diff
@@ -486,7 +486,7 @@ sub start_qemu {
 
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
-                my $nictype_user_options;
+                my $nictype_user_options = '';
                 if ($vars->{NICTYPE_USER_OPTIONS}) {
                     $nictype_user_options = ',' . $vars->{NICTYPE_USER_OPTIONS};
                 }
```
okurz: would it make sense to use conditional assignment then to have it even cleaner?:

```perl
my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
```

aaannz: Done

okurz:
LGTM

PR #564, opened 2016-08-24
virtualization: workaround for unstable type_string on ipmi machine by typing more slowly

alice-suse:
Background: type_string on ipmi is unstable which blocks virtualization test. Originally I thought the maximum interval means typing with the slowest speed, however today I found that it is the opposite. So we will need to change the type speed for type_string behavior for ipmi.
 
Solution: Why I change in the type_string function in testapi.pm is that several files called this function, and inside each file , there may be serveral calls for this function. So code change is the least if change it here. 

Verification: We can not verify it locally here because the issue is only on the ipmi worker in openqa.suse.de. So we can only see whether it solves the problem after merging the code. If it does not work, we will revert the code later, and find other solutions. I think it is worth a merge.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1165,7 +1165,9 @@ sub type_string {
         %args = @_;
     }
     my $log = $args{secret} ? 'SECRET STRING' : $string;
-    my $max_interval = $args{max_interval} // 250;
+    #workaround for unstable type_string behavior on ipmi machine, so type more slowly as VERY_SLOW_TYPING_SPEED setting in lib/utils
+    my $default_max_interval = check_var('BACKEND', 'ipmi') ? 4 : 250;
+    my $max_interval = $args{max_interval} // $default_max_interval;
     bmwqemu::log_call(string => $log, max_interval => $max_interval);
     query_isotovideo('backend_type_string', {text => $string, max_interval => $max_interval});
 }


alice-suse:
@okurz  please help to review. Thanks.

okurz:
good idea. I suggest to ask in irc://irc.suse.de/testing for help to test this

aaannz:
Can not test is, but I don't like loosing option to specify custom interval for ipmi. Can you redo this to set max interval only when not set by parameter?

i.e.
```perl
my $default_max_int = check_var('BACKEND', 'ipmi') ? 4 : 250;
my $max_interval = $args{max_interval} // $default_max_int;
```

alice-suse:
@okurz  There is only one ipmi worker in openqa.suse.de from kulow's words. If someone use this machine manually, it may break the daily build test. What do you think?

@aaannz  ok, i will do it

okurz:
On Wednesday 24 August 2016 20:21:19 xiao li ai wrote:
> @okurz  There is only one ipmi worker in openqa.suse.de from kulow's words.
> If someone use this machine manually, it may break the daily build test.
> What do you think?

yes, I recommend to disable the worker and then test manually. After test is 
done, the worker can be switched on again.

alice-suse:
@okurz  Since now you are most familiar with this change,  would you please help to make a test? This result is very important and will impact the next step plan.

okurz:
I am sorry I don't have access to machines easily right now as I am in vacation until mid of next week.
@Soulofdestiny @dzedro can you help?

alice-suse:
oh, have a nice vacation!

coolo:
Sorry, but this is the wrong place to tweak that.

alice-suse:
Where is the right place?

coolo:
in my head :)

coolo:
I.e. the idea is good, the implementation needs change

alice-suse:
ok, wait for your implementation then .

coolo:
This workaround is actually not needed - and would really slow down *everything* alot

PR #568, opened 2016-08-25
[NEEDHELP] Fix usb-kbd 'key-event-full'

Soulofdestiny:
See last comment in https://progress.opensuse.org/issues/12250

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -634,7 +634,7 @@ sub start_qemu {
             if ($vars->{VNC} !~ /:/) {
                 $vars->{VNC} = ":$vars->{VNC}";
             }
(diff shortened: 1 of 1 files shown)


Soulofdestiny:
I'm definitely not sure about this, so I'm asking you to help me with that :)
Unfortunately I can't set labels

okurz:
seems like a good idea but needs recent qemu version. I guess we should check on overdrive2 and malbec if their qemu supports this first. I would suggest to build a check into the backend to check version of qemu first before trying this.

Soulofdestiny:
sounds like a good idea, yes
i checked overdrive and malbec - the versions differ, but on overdrive the version is new enough to support this

coolo:
this is definitely wrong. 200 ms is an eternity. And we already send the keys delayed.

PR #570, opened 2016-08-29
Make 07-commands.t callable by prove

okurz:


okurz:
anyone has an idea how `t/07-commands.t` can be changed to be callable by prove?

okurz:
@coolo can you help me with that? I tried do make the tests compatible with `prove` but failed with t/07-commands.t. It hangs when executing with `prove t/07-commands.t`.

coolo:
http://paste.opensuse.org/41614718

coolo:
execute summary: kill($pid) does not what it does in C - and prove waits for subprocesses to finish

okurz:
thx, fixed and updated.

coolo:
So now it fails due to https://github.com/os-autoinst/os-autoinst/issues/602

okurz:
but works locally. As long as we don't introduce a call to "prove" in make we should be good and @kraih will fix #602 asap, right?

coolo:
merged into #618

PR #571, opened 2016-08-29
Write the complete check result to the result file

nilxam:
In the old code, write a dummy needle to the result file in interactive mode seems wrong, it should write the complete check result from check_asserted_screen to the result file. Also moved $rsp->{saveresult}
part to the timeout block but just do not set overall state, $rsp->{timeout} must be 1 when $rsp->{saversult} was set to 1.

https://progress.opensuse.org/issues/13456

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -155,24 +155,7 @@ sub _check_backend_response {
 
     my $tags = $rsp->{tags};
 
-    if ($rsp->{saveresult}) {
-        my $img = tinycv::read($rsp->{filename});
-        $autotest::current_test->record_screenfail(
-            img     => $img,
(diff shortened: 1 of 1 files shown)


nilxam:
Quick fix, the second test run is still running.

@AdamWill can you try it on your machine too?

AdamWill:
OK, I've applied it to our staging instance. Seems to work in a quick test. Will update if any issues appear in more extended use.

aaannz:
:+1:

nilxam:
any problem not to merge it here?

AdamWill:
still seems to be working fine here, I've done several needles in interactive mode fine, and not noticed any other negative effects.

PR #573, opened 2016-08-31
Make sure IPMI backend can slow down the typing

coolo:


coolo:
This is my version of https://github.com/os-autoinst/os-autoinst/pull/564 - but while I was looking for a way to test it, I figured that https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1780 is hopefully enough and we actually don't have to slow down the typing (which would make all tests much, much slower to execute)

coolo:
not needed and makes everything more complex.

PR #574, opened 2016-09-01
don't sha1sum qcow assets (poo#13526)

AdamWill:
it takes a long time and blocks the worker process, preventing
it from sending status updates to the server and making it
vulnerable to the 'dead worker' check. Since we're *also*
checksumming in the worker itself, this really isn't needed.

Combined with several other changes I'm submitting for openQA,
this should address poo#13526.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -203,8 +203,6 @@ sub do_extract_assets {
                 symlink("../raid/l$hdd_num", "$img_dir/$name");
             }
         }
-        chomp(my $sha1 = qx{sha1sum -b $img_dir/$name});
-        bmwqemu::diag "sha1sum $sha1";
     }
     else {
         bmwqemu::diag "do_extract_assets: hdd $hdd_num does not exist";


aaannz:
:+1:

okurz:
+1

Calculating checksum at multiple occassions is not necessarily harmful but good to find out problems with the image upload itself. We started calculating the checksum when we saw problems during upload. I agree with this change but to me it looks like the right solution is still pending, this is merely a workaround to reduce the probability of the error you mentioned appearing.

AdamWill:
yeah, it's basically a trade-off: it's not worth the cost of having jobs killed by the 'dead job checker' just for the benefit of getting a checksum from os-autoinst. If we could somehow have the checksum calculated in os-autoinst without blocking the openQA worker (so it continued to send status updates to the server) that'd be fine, but unless someone can figure that out, I think it's best to drop the checksum from os-autoinst for now.

coolo:
I also don't have feelings towards the extra sha1sum, but IMO the timeouts need to be tweaked to also work with larger files. And if this extra sha1sum call makes the real problem we have easier to reproduce, I'm for keeping it around

AdamWill:
Mmmmf. I don't really agree with that. The code on master should be as close to 'working reliably' as we can make it. If you want to reproduce a bug, well, it's easy enough to set up a testing branch and throw as many slow blocking calls in as many places as you like. But it's silly (IMHO) to intentionally leave code that we know causes problems in master, breaking real-world deployments, purely to try and debug the problems.

coolo:
but the bug is in openQA not os-autoinst - so what yo do here is a workaround. os-autoinst should not be held responsible for what the worker does.

AdamWill:
well, that's kinda arguable. It's currently the general design of openQA that it expects the worker process to stay responsive. Adding this checksum to os-autoinst caused the openQA worker process to be unresponsive for the whole time it takes the checksum to finish. Is that really a bug in openQA, or is it os-autoinst not respecting openQA's expectations?

IOW, I'd argue that anyone who wants to make os-autoinst do long, blocking operations where it didn't before has the responsibility to somehow make openQA OK with that, rather than just doing it and telling openQA to suck it up...

PR #576, opened 2016-09-02
Make cancelling more reliable

coolo:
This should fix 95% of https://progress.opensuse.org/issues/12178

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -106,7 +106,6 @@ sub kill_autotest {
 }
 
 sub kill_backend {
-    bmwqemu::stop_vm();
     if (defined $bmwqemu::backend && $bmwqemu::backend->{backend_pid}) {
         diag("killing backend process $bmwqemu::backend->{backend_pid}");
         kill('TERM', $bmwqemu::backend->{backend_pid});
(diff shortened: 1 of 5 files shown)


isotovideo
```diff
@@ -410,16 +409,17 @@ if ($testfd) {
     kill_autotest;
 }
 
-diag "isotovideo done" unless $r;
-diag "FAIL" if $r;
+diag "isotovideo " . ($r ? 'failed' : 'done');
```
okurz: 'failed' : 'succeeded' ?

isotovideo
```diff
@@ -410,16 +409,17 @@ if ($testfd) {
...
-    $clean_shutdown = $bmwqemu::backend->_send_json({cmd => 'is_shutdown'});
-    diag "BACKEND SHUTDOWN $clean_shutdown";
-};
-
-bmwqemu::stop_vm();
+if (!$r) {
+    eval {
+        $clean_shutdown = $bmwqemu::backend->_send_json({cmd => 'is_shutdown'});
+        diag "BACKEND SHUTDOWN $clean_shutdown";
+    };
```
okurz: The exception is never handled but should be, at least in a log output

isotovideo
coolo: we don't know the job state at this point (and it doesn't matter) - succeeded expresses more than we know

isotovideo
coolo: I disagree

okurz:
LGTM

PR #577, opened 2016-09-02
Fix the exit code of isotovideo

coolo:
The exit code of a perl script is not the one you give to exit, but is the
value of $? at the end of all END blocks (see perldoc -f exit). Interesting
caveat


okurz:
LGTM

Issue #580, opened 2016-09-05
cannot work with INSTALL.asciidoc's instructions


henrich:
It says "sudo zypper in openQA" but failed.

should add
$ sudo zypper ar -f obs://devel:openQA/openSUSE_Leap_42.1 openQA
$ sudo zypper ar -f obs://devel:openQA:Leap:42.1/openSUSE_Leap_42.1 openQA-perl-modules
$ sudo zypper in openQA

then, we need git for cloning code from GitHub
$ sudo zypper in git

before configure & make, we should install make
$ sudo zypper in make

and there's some Perl modules are lucking
$ sudo zypper in perl-IPC-System-Simple
$ sudo zypper in perl-class
$ sudo zypper in Perl-exception-class

coolo:
we should simply erase this.

okurz:
See #582 for corrections. The perl dependencies are properly listed in the cpanfile. You would need to either call the installation commands on all dependencies or call e.g. `cpanm --installdeps .`

henrich:
Thanks!

And it's better to note about perl modules installation since some users are not familiar with Perl and don't know about package manager.

henrich:
Could you tell me more, please?

AS instruction, create vars.json and adjust "CASEDIR" : "/home/henrich/distri/opensuse/"

> henrich@linux-zaqb:/tmp/os-autoinst-run> ~/os-autoinst/isotovideo 
> 13:40:39.3994 2179 got sigchld
> 13:40:39.3995 2179 git hash of test distribution: 923b5afeacb2fdfae969557635693818705b07f0
> Can't locate /home/henrich/distri/opensuse//main.pm in @INC (you may need to install the ::home::henrich::distri::opensuse::::main module) (@INC contains: /tmp/os-autoinst-run/blib/arch /tmp/os-autoinst-run/blib/lib /home/henrich/os-autoinst /usr/lib/perl5/site_perl/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/site_perl/5.18.2 /usr/lib/perl5/vendor_perl/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/vendor_perl/5.18.2 /usr/lib/perl5/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/5.18.2 /usr/lib/perl5/site_perl .) at /home/henrich/os-autoinst/isotovideo line 209.
> 13:40:39.4006 2179 awaiting death of commands process
[...]

PR #583, opened 2016-09-06
INSTALL.asciidoc: Mention installing perl dependencies based on cpanfile

okurz:
Related issue https://github.com/os-autoinst/os-autoinst/issues/580

### DIFF
--- a/INSTALL.asciidoc
+++ b/INSTALL.asciidoc
@@ -7,26 +7,32 @@ the
 https://github.com/os-autoinst/openQA/blob/master/docs/Installing.asciidoc[openQA
 installation guide]
 +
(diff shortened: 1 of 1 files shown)


aaannz:
:+1: 
I'm kind of torn on cpan usage instead of pkg manager, but I guess this is universal approach.

coolo:
We're not doing anyone a favor if they follow this. You link the openQA guide, which lets you call zypper - to then mess with it. Actually if you install openQA-worker too (i.e. followed the instructions till the end) you will also have os-autoinst and its perl dependencies.

coolo:
:-1:

PR #584, opened 2016-09-07
testapi: add `check_type_string`, with screen change checking

AdamWill:
this adds a new `check_type_string` function, which basically
just does `type_string`, but waits for the screen to change
every few characters. You can configure how many characters
to type per `wait_still_screen` with the `size` parameter. I
wrote this to try and mitigate an issue we are having where
Fedora tests occasionally seem to exhaust the VM's input buffer
and thus miss characters (or repeat a character a bunch of
times, if the VM happens to get a key_down event but not the
key_up event), but it seems generally useful and may as well go
in testapi, I think.

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -21,6 +21,7 @@ sub fake_send_json {
     my ($to_fd, $cmd) = @_;
     push(@$cmds, $cmd);
 }
+
 sub fake_read_json {
     my ($fd) = @_;
     my $lcmd = $cmds->[-1];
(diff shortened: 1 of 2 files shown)


coolo:
I would prefer this implemented as option to type_string and implemented directly in the backend

coolo:
type_string 'hallo', max_interval => 10, wait_screen_changes => 3

AdamWill:
well, the problem with doing it that way is you can't use `wait_still_screen`, since that's in `testapi`. You'd have to re-implement it at the backend level. IIRC anyway...

coolo:
That's a bit too much refactoring - but IMO at least make it part of the type_string api instead of having yet another function

AdamWill:
ok, will do tomorrow.

PR #585, opened 2016-09-07
Commands server: Listen to ipv6 as well

coolo:
to support systems where localhost resolves to IPv6


aaannz:
I thought `*` will make mojo listen on all interfaces including ipv6. Or is this more about limiting listening only on localhosts?

coolo:
http://mojolicious.org/perldoc/Mojo/Server/Daemon#listen documents it pretty well

PR #586, opened 2016-09-07
Improve testapi test

Martchus:
context: https://progress.opensuse.org/issues/10932

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -411,7 +411,7 @@ sub record_soft_failure_result {
     open my $fd, ">", bmwqemu::result_dir() . "/$text_fn";
     print $fd "# Soft Failure:\n$reason\n";
     close $fd;
-    $details->{text} = $text_fn;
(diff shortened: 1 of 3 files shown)


okurz:
I am not sure. My intention with the mockup code in the subtest was to only provide the bare minimum what is necessary to make the test pass even if it is making objects incomplete. I would be interested in what you intend to do based on this but I am not against this change.

Martchus:
This is what I want to do: https://github.com/Martchus/os-autoinst/commits/record_softfailure_icon
In order to fix: https://progress.opensuse.org/issues/10932

In order to test this feature, I need the first 4 lines of this commit. Otherwise required methods are not available and it will fail in `_result_add_screenshot`.

I just thought this is also an easy way to clean the rest of the test code and to increase test coverage by the way.

Martchus:
And, what do you think now?

PR #588, opened 2016-09-07
Have the backend do clean shutdown even if not told through pipe

coolo:


okurz:
how did you test?
* I assume cancelling a qemu process and checking no qemu-system-x86_64 process is left over?
* does s390x cancelling still work?
* can you test with an IPMI backend, too?
* does plain shutdown on success still work?
* does shutdown on die from test still work?
* does it work when isotovideo is called manually as well as from openQA worker?
* as above but aborting isotovideo with ^C?
* do the "incompletes" still work?

coolo:
I tested:
  - successful shutdown is clean
  - qemu shuts down if stopping with INT (^c)
  - qemu shuts down with TERM (as the worker would do) in various states of the test
  - exit status is correct in all terms

Soulofdestiny:
LGTM

PR #591, opened 2016-09-08
Don't try to do snapshots on raw images

dzedro:
It was silently failing also before
https://openqa.suse.de/tests/538852#downloads
Now it is failing more obvious
https://openqa.suse.de/tests/543216#downloads


coolo:
we were just discussing how this can have worked before - as it's now broken in TW tests

PR #592, opened 2016-09-08
Don't kill the first ssh-connection

Soulofdestiny:
The ssh connection with getty running in it, was killed when the iucvconn was established.
This results in a race condition and the sporadic fails of wait_serial
Now we ensure that agetty is running the whole time

### DIFF
--- a/consoles/sshIucvconn.pm
+++ b/consoles/sshIucvconn.pm
@@ -32,16 +32,21 @@ sub new {
 }
 
 sub activate {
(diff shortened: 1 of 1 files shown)


okurz:
yeah, I don't get it but if it fixes things … :-)

coolo:
at least you could have said "some more tests would be nice"

okurz:
ah, missed my chance

PR #593, opened 2016-09-09
Provide lib/ of the test distribution to main.pm

coolo:
This way we avoid every product's main.pm having a BEGIN clause with
hacky includes - and using functions in main.pm is handy

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/1452#discussion_r66057803 and https://progress.opensuse.org/issues/12258


okurz:
LGTM

PR #596, opened 2016-09-12
Don't leave the commands server when it can't react

coolo:
This lead to the openqa worker hanging while trying to get status
updates. This mainly was a problem for jobs publishing hdds

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -263,7 +263,7 @@ sub run_daemon {
         $daemon->run;
     }
     catch {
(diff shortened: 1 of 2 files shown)


coolo:
Fixing https://progress.opensuse.org/issues/13682

Before: https://tortuga.suse.de/tests/507 and previous
After: 509 and following

okurz:
LGTM although ... Ah, nevermind ;-)

PR #600, opened 2016-09-16
Wait a second before and after typing

dzedro:
This would solve lot of issues, depends how slow/safe/human we want to be to avoid race condition e.g. https://progress.opensuse.org/issues/13766

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1163,7 +1163,8 @@ sub type_string {
     # special argument handling for backward compat
     my $string = shift;
     my %args;
-    if (@_ == 1) {    # backward compat
+    wait_still_screen 1, 5;    # wait a second before typing
+    if (@_ == 1) {             # backward compat
         %args = (max_interval => $_[0]);
     }
     else {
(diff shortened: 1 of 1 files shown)


okurz:
just in before the wait_screen_change option was added. I would suggest to test if this helps already or if we should add your proposal, but as optional, not mandatory in all cases.

AdamWill:
nack. please don't do this. The right way to do this is to put it in your distro-specific high-level functions - `type_slowly` or `type_very_slowly` or whatever they're called.

I already wrote `type_safely` and `type_very_safely` functions for Fedora which use the 'wait for screen change between character sets' feature we recently added to `type_string` *and* `wait_still_screen` - but for longer - after typing; if you do this, those will just be stacking `wait_still_screen` calls on top of each other with *no way* to avoid it.

IMHO `type_string`'s basic behaviour is too longstanding to make drastic changes like this unconditionally. At the very least any change like this should be optional and not the default.

AdamWill:
Also consider the case where we're typing into a screen which changes constantly (in Fedora, for e.g., we test the cockpit webapp, which handily shows a constantly-scrolling graph all the damn time). The `type_string` will now waste 10 seconds.

coolo:
we need to come up with a good solution to make this more reliable, but slowing down everything is not the solution, no

PR #604, opened 2016-09-22
Don't overwrite 'reset_consoles' in svirt backend

Soulofdestiny:


coolo:
this not about the usage of it - the import makes it part of the backend, which is just plain *wrong* - if called or not is of no importance.

PR #607, opened 2016-09-28
scale timeout for assert_shutdown

AdamWill:
assert_shutdown wasn't scaling its timeout, it probably should.


coolo:
Please make your commit message start with a capital letter - as suggested by the link in README.

PR #609, opened 2016-10-06
Skip needles with missing PNG file

asdil12:
See https://progress.opensuse.org/issues/14066

### DIFF
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -3,6 +3,7 @@
 use strict;
 use warnings;
 use Test::More;
+use Try::Tiny;
 use File::Basename;
 
 # optional but very useful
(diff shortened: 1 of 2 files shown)


okurz:
missing test. We have better test structure now in os-autoinst. You may try to use it :-)

asdil12:
Added the test.

t/01-test_needle.t
```diff
@@ -63,6 +64,15 @@ $needle = needle->new($data_dir . "kde.ref.json");
...
 ok(!defined $res, "no match with different art");
 
+try {
+    $img1 = tinycv::read($data_dir . "kde.ref.png");
+    $needle = needle->new($data_dir . "kde.ref.json");
+    my $needle_nopng = needle->new($data_dir . "console.ref.json");
+    $needle_nopng->{png} .= ".missing.png";
+    $res = $img1->search([$needle_nopng, $needle]);
+};
+ok(defined $res, "skip needles without png");
```
coolo: I suggest to initialize the res - otherwise the next test added above relying on $res will make your test invalid.

PR #610, opened 2016-10-08
Wiggle the mouse in case of duplicated mouse_hides

coolo:
this is basically a qemu workaround, but qemu VNC will ignore two
requests to move the mouse to the same absolute position. Human users
don't notice that as they *move* the mouse, they don't jump. So we
need to simulate mouse moves at least to some degree.

The duplicated mouse_hides being ignored are basically in situations
like:

mouse_hide;
restart_X; # will set the mouse cursor into the middle
mouse_hide;

(the mouse will be still at the center of the screen)

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -181,23 +181,44 @@ sub release_key {
     return {};
 }
 
+sub _mouse_move {
+    my ($self, $x, $y) = @_;
+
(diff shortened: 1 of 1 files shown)


aaannz:
:+1:

coolo:
in case you wonder: I would like to test this more before merging

coolo:
I tested the basic functionality - if the workaround works in real life we will have to monitor

PR #611, opened 2016-10-09
Fix memory corruption in snd2png

coolo:
There were invalid reads into the audio data on both reading the input
and the mapping to the image. We actually saw crashes very seldomly,
but a white image resulting out of invalid data causing invalid scaling
of the resulting image to be all zeros

Fixing https://progress.opensuse.org/issues/10640

Before: https://tortuga.suse.de/tests/604 (and many other examples in production)
After: 606 and up :)


coolo:
I also increased the debug output slightly to avoid other symptoms

.travis.yml
```diff
@@ -14,4 +14,4 @@ script:
   - ./configure
     # 'coverage' executes all tests and checks code coverage against threshold
     # in Makefile.am
-  - make && make coverage-coveralls VERBOSE=1
+  - make && make check coverage-coveralls VERBOSE=1
```
okurz: Doesn't coverage-coveralls include this implicitly?

okurz: But then every test is executed twice. Can you add an explicit target for doing all the other steps you want to cover? Or exclude actual test execution from the `t` directory from `make check`?

PR #612, opened 2016-10-10
Fix call to mouse_move in mouse_set

coolo:
Fixing https://tortuga.suse.de/tests/630/file/autoinst-log.txt


coolo:
(this was fortunately nowhere else deployed)

PR #615, opened 2016-10-11
Fix stale VNC screens during s390 reboot

coolo:
The problem is a bit tricky as the VNC server goes away because of the
reboot and the client has no way built into the RFB protocol to detect that.
So all we have are timestamps and forced updates to verify the existance
of the VNC server. We give the system 2 seconds before we ask for a live
sign, and another 2 to turn off the screen (blank image not matching any
check_screens at that point)


coolo:
You can see the problem in https://openqa.suse.de/tests/611292#step/desktop_mainmenu/2

PR #616, opened 2016-10-11
Decouple video encoder from backend

coolo:
Instead of waiting for the slow (because CPU intensive and reniced)
video encoder, just write PNG frames to a file and only at the end
wait for it to finish up the video.

The video encoder will monitor the file given and work on it until
it gets a SIGTERM. This makes sure it can't stall the backend

### DIFF
--- a/videoencoder.cpp
+++ b/videoencoder.cpp
@@ -1,5 +1,8 @@
 /* Modified by Stephan Kulow <coolo@suse.de> 2014 to take the PNGs
-  from stdin and use opencv directly */
+   from stdin and use opencv directly */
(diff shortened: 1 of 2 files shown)


videoencoder.cpp
```diff
@@ -328,9 +368,9 @@ main (int argc, char *argv[])
 	  fflush (ogg_fp);
 	  fsls = 0;
 	}
-      //fprintf(stderr, "%s\n", input_png);
     }
 
+  printf("last frame\n");
```
okurz: where is this sent to? Is it just a debugging output?

coolo: just debugging to verify the frame is received. It will end up in autoinst-log.txt

PR #617, opened 2016-10-11
Fix full update in VNC

coolo:
the logic was about to get a full update after 3 seconds of silence,
not the other way around

To be tested by @Soulofdestiny


coolo:
(still to be tested, but as master is broken anyway, we need to do this anyway :)

PR #619, opened 2016-10-12
Improve the VNC stall detection

coolo:
We need two variables to differ between a VNC stall and a VNC screen
not moving. So we track the time of the last forced update. Only if
2 seconds since *that* update request passed, we know the server is gone.
To normal update requests it's actually *supposed* not to answer if the
screen didn't move

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -21,7 +21,7 @@ __PACKAGE__->mk_accessors(
     qw(hostname port username password socket name width height depth save_bandwidth
       no_endian_conversion  _pixinfo _colourmap _framebuffer _rfb_version screen_on
       _bpp _true_colour _do_endian_conversion absolute ikvm keymap _last_update_received
-      vncinfo old_ikvm
+      _last_update_requested vncinfo old_ikvm
       ));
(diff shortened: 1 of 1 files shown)


consoles/VNC.pm
```diff
@@ -782,25 +783,26 @@ sub send_update_request {
...
-    # after 3 seconds: force incremental update
     # after 4 seconds: turn off screen
     my $time_since_last_update = time - $self->_last_update_received;
 
     # if there were no updates, send a forced update request
     # to get a defined live sign. If that doesn't help, consider
     # the framebuffer outdated
     if ($self->_framebuffer) {
-        if ($time_since_last_update > 4) {
+        if ($self->_last_update_requested - $self->_last_update_received > 2) {
```
richiejp: Perhaps the constant 2 could be given a meaningful name, like `$grace_period` or similar.

coolo: perhaps. But it doesn't matter to me as I consider it obvious

coolo:
has been tested in 3 tests now without flaw. So let's assume this works

PR #620, opened 2016-10-13
Fix remote consoles on zkvm

coolo:
The C locale is affecting ncurses font more than what I intended with
this change. it should be UTF-8 still


okurz:
LGTM - bug reported?

coolo:
about non-utf-8 locales being non-utf-8? no.

PR #621, opened 2016-10-13
Save machine memory dumps

foursixnine:
Added the necessary code to perform an online migration of the machine. 

After the vmstate.gz file has been generated, the machine can be restored or migrated by appending: **-incoming "exec: gzip -c -d vmstate.gz"** at the end of the qemu line. By grepping autoinst.txt looking for _starting qemu service:_ the line to spawn qemu can be found.

---

This change is [Reviewable](https://reviewable.io/reviews/os-autoinst/os-autoinst/621)

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -155,32 +155,31 @@ sub can_handle {
 sub save_state {
 
     my ($self, $args) = @_;
-    my $rsp = 0;
+    my $rsp    = 0;
     my $vmname = $args->{name};
 
     bmwqemu::diag("Trying to save a QEMU image");
 
-    $rsp = $self->_send_hmp("stop");
+    $self->freeze_vm;
 
-    diag "Sucessfully stopped our vm: $rsp";
-    die unless ($rsp eq "stop");
+    $rsp = $self->handle_qmp_command(JSON::from_json('{ "execute": "migrate_set_speed", "arguments": { "value": "4095m" } }'));
+    $rsp = $self->handle_qmp_command(JSON::from_json('{ "execute": "migrate", "arguments": { "uri": "exec:gzip -c > vmstate.gz" } }'));
 
-    $rsp    = $self->_send_hmp("migrate_set_speed 4095m");
-    die unless ($rsp eq "migrate_set_speed 4095m");
-    diag "Set up speed to: $rsp" 
+    do {
 
-    #For compression pixz was suggested, but it might add a new dependency
-    # also it would use multiple CPU's, which might be a problem in production.
-    $rsp    = $self->_send_hmp("migrate \"exec:gzip -c > vmstate.gz\"");
-    die unless ($rsp eq "migrate \"exec:gzip -c > vmstate.gz\"");
-    diag "Saved the machine state $rsp" 
+        sleep 0.5;    #We want to wait a decent amount of time, a file of 10GB will be
+                      # migrated in about 40secs with an ssd drive. and no heavy load.
 
-    $rsp    = $self->_send_hmp("cont");
-    die unless ($rsp eq "cont");
-    diag "Allowed the VM to keep running $rsp";
-    
+        $rsp = $self->handle_qmp_command({execute => "query-migrate"});
 
-    return;
+        diag "QEMU MIGRATING BYTES TOTAL:     \t" . $rsp->{return}->{ram}->{total};
+        diag "QEMU MIGRATING BYTES REMAINING:   \t" . $rsp->{return}->{ram}->{remaining};
+
+    } until ($rsp->{return}->{status} eq "completed");
+
+    diag "Migration completed.";
+    $self->cont_vm;
+
 }
 
 sub save_snapshot {
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -152,34 +152,54 @@ sub can_handle {
     return;
 }
 
-sub save_state {
-
+sub save_memory_dump {
     my ($self, $args) = @_;
-    my $rsp    = 0;
-    my $vmname = $args->{name};
+    my $rsp = 0;
 
-    bmwqemu::diag("Trying to save a QEMU image");
+    bmwqemu::diag("Migrating the machine.");
 
-    $self->freeze_vm;
+    mkpath("ulogs");
 
-    $rsp = $self->handle_qmp_command(JSON::from_json('{ "execute": "migrate_set_speed", "arguments": { "value": "4095m" } }'));
-    $rsp = $self->handle_qmp_command(JSON::from_json('{ "execute": "migrate", "arguments": { "uri": "exec:gzip -c > vmstate.gz" } }'));
+    $rsp = $self->handle_qmp_command(
+        {
+            execute   => "migrate",
+            arguments => {
+                uri => sprintf("exec:gzip -c > ulogs/%s-vm-memory-dump.gz", $args->{filename}),
+            }});
+
+    die(sprintf("Migration failed: desc: %s, class: %s, stopped", $rsp->{error}->{desc}, $rsp->{error}->{class})) if ($rsp->{error});
+
 
     do {
 
-        sleep 0.5;    #We want to wait a decent amount of time, a file of 10GB will be
+        sleep 0.5;    #We want to wait a decent amount of time, a file of 1GB will be
                       # migrated in about 40secs with an ssd drive. and no heavy load.
-
         $rsp = $self->handle_qmp_command({execute => "query-migrate"});
 
-        diag "QEMU MIGRATING BYTES TOTAL:     \t" . $rsp->{return}->{ram}->{total};
-        diag "QEMU MIGRATING BYTES REMAINING:   \t" . $rsp->{return}->{ram}->{remaining};
+        diag "Migrating total bytes:     \t" . $rsp->{return}->{ram}->{total};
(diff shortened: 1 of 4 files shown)


foursixnine:
@agraf followed your recomendations... both sleeps are fixed.

okurz:
Please write your commit messages in present imperative tense. Also see http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages. You can reference the progress issue which describes the feature request.

backend/qemu.pm
```diff
@@ -152,6 +152,60 @@ sub can_handle {
...
+            hdd_num => $args->{disk},
+            name    => sprintf("%s-%d-vm_disk_file.qcow2", $args->{filename}, $args->{disk}),
+            dir     => "ulogs",
+            format  => "qcow2"
+        });
+
+    diag "Sucessfully saved disk #1.";
+
+}
+
```
okurz: too many blank lines

backend/qemu.pm
```diff
@@ -152,6 +152,60 @@ sub can_handle {
...
+    $self->do_extract_assets(
+        {
+            hdd_num => $args->{disk},
+            name    => sprintf("%s-%d-vm_disk_file.qcow2", $args->{filename}, $args->{disk}),
+            dir     => "ulogs",
+            format  => "qcow2"
+        });
+
+    diag "Sucessfully saved disk #1.";
```
okurz: shouldn't you include the argument `$args->{disk}` here, too?

backend/qemu.pm
```diff
@@ -152,6 +152,60 @@ sub can_handle {
...
+
+    do {
+
+        sleep 0.5;    #We want to wait a decent amount of time, a file of 1GB will be
+                      # migrated in about 40secs with an ssd drive. and no heavy load.
+
+        $rsp = $self->handle_qmp_command({execute => "query-migrate"});
+
+        diag "QEMU MIGRATING BYTES TOTAL:     \t" . $rsp->{return}->{ram}->{total};
+        diag "QEMU MIGRATING BYTES REMAINING:   \t" . $rsp->{return}->{ram}->{remaining};
```
okurz: why capital letters?

backend/qemu.pm
```diff
@@ -152,6 +152,60 @@ sub can_handle {
...
+sub save_memory_dump {
+
+    my ($self, $args) = @_;
+    my $rsp    = 0;
+    my $vmname = $args->{name};
+
+    $rsp = $self->handle_qmp_command(
+        {
+            execute   => "migrate_set_speed",
+            arguments => {value => "4095m"}});
```
okurz: why?

testapi.pm
```diff
@@ -822,6 +825,77 @@ sub eject_cd {
...
+Saves the SUT memory state using C<$filename> as base for the memory dump filename, 
+the default will be the current test's name. 
+
+I<Currently only qemu backend is supported.>
+
+=cut
+
+sub save_memory_dump {
+
+    my $filename = shift || ref($autotest::current_test);
```
okurz: more commonly we try to use

```perl
my ($filename) = @_;
$filename //= ref($autotest::current_test);
```

What do you think about it?

testapi.pm
```diff
@@ -822,6 +825,77 @@ sub eject_cd {
...
+sub save_storage_drives {
+
+    my $filename = shift || ref($autotest::current_test);
+    die "Method should be called within a post_fail_hook" unless ((caller(1))[3]) !~ /post_fail_hook/;
+
+    bmwqemu::log_call();
+    bmwqemu::diag("Trying to save machine drives");
+    bmwqemu::load_vars();
+
+    #TODO: right now, we're saving all the disks
```
okurz: just delete the `TODO:`. If it's an issue we should track it in progress tickets. If it's not an issue it will never be done and it does not harm :-)

testapi.pm
```diff
@@ -822,6 +825,77 @@ sub eject_cd {
...
+
+=head2 freeze_vm
+
+  freeze_vm;
+
+If the backend supports it, freeze the vm. This will allow the vm to be paused/frozen
+within the test, but only from the post_fail_hook. So that memory and disk dumps
+can be extracted without any risk of data changing. 
+
+It is assumed that the test writer will call freeze_vm before getting the memory or disk dumps.
```
okurz: what happens if we don't?

backend/qemu.pm
foursixnine: Yes, it is fixed now.

testapi.pm
```diff
@@ -822,6 +822,19 @@ sub eject_cd {
     query_isotovideo('backend_eject_cd');
 }
 
+=head2 save_state
+
+  save_state;
+
+if backend supports it, save machine state
```
foursixnine: The documentation has been updated and extended :)

backend/qemu.pm
foursixnine: How about now?... Changed wording a bit

testapi.pm
foursixnine: Doc was changed to explain why freeze_vm should be called when doing both, disk and memory dump.

testapi.pm
```diff
@@ -822,6 +825,83 @@ sub eject_cd {
...
+paused/frozen within the test, but only from the post_fail_hook. So that memory
+and disk dumps can be extracted without any risk of data changing.
+
+Call this method to ensure memory and disk dump refer to the same machine state.
+
+I<Currently only qemu backend is supported.>
+
+=cut
+
+sub freeze_vm {
```
agraf: Doesn't this need an unfreeze too to be able to interact with the VM after we saved it off?

backend/qemu.pm
```diff
@@ -152,6 +152,57 @@ sub can_handle {
...
+            execute   => "migrate_set_speed",
+            arguments => {value => $migration_speed}});
+
+    bmwqemu::diag("Trying to save a QEMU image");
+
+    $rsp = $self->handle_qmp_command(
+        {
+            execute   => "migrate",
+            arguments => {
+                uri => sprintf("exec:gzip -c > ulogs/%s-vm-memory-dump.gz", $args->{filename}),
```
agraf: No pigz/pixz with nice level?

foursixnine: Not yet :(

coolo: on top of that: we have enough problems with CPU stalls - and we have no time pressure at this point. The test is failed anyway.

testapi.pm
```diff
@@ -822,6 +825,83 @@ sub eject_cd {
...
+within a post fail hook. Different filenames should be provided if the dump is
+being used within the test itself.
+
+I<Currently only qemu backend is supported.>
+
+=cut
+
+sub save_memory_dump {
+    my ($args) = @_;
+    $args->{filename} = $args->{filename} || ref($autotest::current_test);
```
coolo: you can stlil use ||= :)

backend/qemu.pm
agraf: That's why putting it on a friendly nice level is critical. And once you've done that, you can as well use as much cpu power as you have spare.

coolo: No. this feature is implemented because we don't expect the developer debugging the failure available. If she was, we would have preferred to just keep the VM running. 

And if the developer is not available, we have all time on earth to dump and upload. The CPU of the qemu process is spare (as the VM is frozen), so use it fully and free up the worker slot.

testapi.pm
foursixnine: Done :) Looks beautiful now.

foursixnine:
There's currently a bug, that triggers when there's nothing on the pool directory, and leaves the machine migration stalled. I still need to hunt that bug down. It is  related to how the handle_qmp_command works. More on that later...

okurz:
Please delete the disabled code about the migration speed (keep it in a local branch). Besides this, I think we are good to go. IMHO we can worry about the empty pool dir issue later

foursixnine:
@okurz I just removed the comments. Somehow the travis build failed... but while installing stuff via apt...

okurz:
occassionally travis checks fail, in this case because apt-get couldn't install stuff. Any project admin of `os-autoinst` could retrigger the build or you just force push again after trivial changes.
another point: I guess you can rename the PR subject and remove the "PoC:"

foursixnine:
Finally i found the root of the problem with the stalled migrations, [bsc#1008148](https://bugzilla.suse.com/show_bug.cgi?id=1008148). 

Also added a check to verify that the migration is able to run, otherwise the method will die. Now we're able to merge and use this feature as soon as the qemu gets patched.

okurz:
> Now we're able to merge and use this feature as soon as the qemu gets patched.

you probably mean "as soon as qemu is patched, the factory package of qemu got updated, maintenance updates for an update of qemu will be created and accepted into supported released products, updates applied and rolled out on all our productive instances". So: expect the fix to be available in some months from now and better live with a workaround for the time being :-)

sysrich:
> So: expect the fix to be available in some ~~months~~ **_WEEKS_** from now and better live with a workaround for the time being :-)

Fixed it for you..shouldn't be THAT long to get qemu maintenance updates out with @agraf and @foursixnine working together on it

foursixnine:
Bruce already got the patch provided by @agraf so it's all matter of waiting now.

coolo:
we don't have to wait with deploying this though - only with using it :)

aaannz:
This can be merged, right?

foursixnine:
@aaannz From my side yes,

PR #623, opened 2016-10-17
zKVM: implement proper shutdown check

Soulofdestiny:


Soulofdestiny:
local verification
http://opeth.suse.de/tests/3797

backend/svirt.pm
```diff
@@ -56,6 +56,13 @@ sub do_start_vm {
     return {};
 }
 
+sub is_shutdown {
+    my ($self) = @_;
+
+    my $instance = get_var('VIRSH_INSTANCE');
```
coolo: there are like 4 places now that check the instance - and 3 of them have a default to 1. So you better extract that somewhere.

Soulofdestiny:
closing - because https://github.com/os-autoinst/os-autoinst/pull/626 is in a much better shape

PR #624, opened 2016-10-17
WIP: virtio_console

richiejp:
This allows the user to select a 'virtio-terminal' which is a text only (serial) console. This restricts the type of input and output which can be achieved and introduces a moderate amount of extra code, but is orders of magnitude quicker than sending key presses though VNC. The primary motivation for adding this is the LTP, which is made of thousands of executables which are best started individually by the OpenQA test script.

Currently the implementation is not quite transparent and adds a couple of functions to testapi, however these can probably be removed. Systems which do not support virtio can achieve a similar speed up by using an old fasioned serial port (or SSH) piped over whatever transport, everything will probably be the same except the transport/sockets used.

A significant proportion of the additions below are made up of the unit tests and documentation.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -939,17 +939,6 @@ sub cont_vm {
     return $self->handle_qmp_command({execute => 'cont'});
 }
 
-sub wait_terminal {
-    my $self  = shift;
-    my %nargs = %{(shift)};
-
-    $nargs{do_while_idle} = sub {
-        $self->run_capture_loop(undef, 0.1);
-    };
-
-    return $self->{current_screen}->read_until($nargs{pattern}, $nargs{timeout}, %nargs);
-}
-
 1;
 
 # vim: set sw=4 et:
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -637,6 +637,12 @@ sub start_qemu {
             push(@params, "-k", $vars->{VNCKB}) if ($vars->{VNCKB});
         }
 
+        if ($vars->{VIRTIOCONSOLE}) {
+            push(@params, '-device', 'virtio-serial');
+            push(@params, '-chardev', "socket,path=/tmp/$vars->{VIRTIOCONSOLE},server,nowait,id=$vars->{VIRTIOCONSOLE}");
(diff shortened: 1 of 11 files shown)


testapi.pm
```diff
@@ -553,6 +555,72 @@ sub wait_serial {
     return;    # false
 }
 
+=head2 wait_terminal
+
+  wait_terminal($pattern, timeout => 30, expect_not_found => 0, record_output => 0,
+                exclude_match => 0, no_regex => 0);
+
+This function is only valid when the console is set to a text terminal. It waits
```
coolo: so why not implement the function in the console then?

richiejp: Because it probably won't be specific to one console, but many text based consoles. On the other hand it might fit into wait_serial and not be necessary at all. I will try to (re)move it.

backend/qemu.pm
```diff
@@ -637,6 +636,13 @@ sub start_qemu {
             push(@params, "-k", $vars->{VNCKB}) if ($vars->{VNCKB});
         }
 
+        if ($vars->{VIRTIO_CONSOLE}) {
```
coolo: and is there a downside of always enabling this?

richiejp: Not to my knowledge, I can leave it as always on after testing.

testapi.pm
```diff
@@ -687,7 +621,7 @@ sub assert_script_run {
     }
     my $str = hashed_string("ASR$cmd");
     # call script_run with idle_timeout 0 so we don't wait twice
-    script_run("$cmd; echo $str-\$?- > /dev/$serialdev", 0);
+    script_run("$cmd; echo $str-\$?- | tee /dev/$serialdev", 0);
```
richiejp: Perhaps it is not available on some systems, but I don't like using it regardless, it will probably at least break some needles. I would prefer to simply remove the redirection when using a serial terminal, but I need to know what the current console is. That is not really such a problem though, thinking about it.

richiejp:
It seems to be working pretty well now, I am currently testing and working on the LTP test runner with it. I am going to rebase and squash the commits to clean it up and get it ready for a pull request. I will probably have to close this request and open a new one.

PR #625, opened 2016-10-18
update known UEFI firmware locations for Fedora distro package

AdamWill:
Fedora now has a distro edk2-ovmf package, so add the pure EFI
firmware from that to the two lists of known UEFI firmware file
locations (and update the comment).

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -59,6 +59,10 @@ our $istty;
 our $direct_output;
 our $standstillthreshold = scale_timeout(600);
 
+# Known locations of OVMF (UEFI) firmware: first is openSUSE, second is
+# the kraxel.org nightly packages, third is Fedora's edk2-ovmf package.
+our @ovmf_locations = ('/usr/share/qemu/ovmf-x86_64-ms.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd', '/usr/share/edk2/ovmf/OVMF_CODE.fd');
+
 our %vars;
 
 sub load_vars() {
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -136,9 +136,7 @@ sub _init_xml {
     }
 
     if (get_var('UEFI') and check_var('ARCH', 'x86_64') and !get_var('BIOS')) {
-        # These are known locations for openSUSE and Fedora (respectively).
-        my @known = ('/usr/share/qemu/ovmf-x86_64-ms.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd');
-        foreach my $firmware (@known) {
+        foreach my $firmware (@bmwqemu::ovmf_locations) {
             if (!$self->run_cmd("test -e $firmware")) {
                 set_var('BIOS', $firmware);
                 $elem = $doc->createElement('loader');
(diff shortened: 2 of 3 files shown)


consoles/sshVirtsh.pm
```diff
@@ -136,8 +136,14 @@ sub _init_xml {
...
     if (get_var('UEFI') and check_var('ARCH', 'x86_64') and !get_var('BIOS')) {
-        # These are known locations for openSUSE and Fedora (respectively).
-        my @known = ('/usr/share/qemu/ovmf-x86_64-ms.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd');
+        # These are known locations: first is openSUSE, second is for
+        # the kraxel.org nightly packages, third is Fedora's edk2-ovmf
+        # package.
+        my @known = (
+            '/usr/share/qemu/ovmf-x86_64-ms.bin',
+            '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd',
+            '/usr/share/edk2/ovmf/OVMF_CODE.fd',
```
okurz: Probably github is broken and just shows this incorrectly. I wouldn't expect you from simply duplicating this even more ;-)

[DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself). You are calling the same code multiple times, please [extract method](http://refactoring.com/catalog/extractMethod.html).

AdamWill: Sigh, I didn't break it, why do I have to fix it? Why not yell at whoever wrote that file? :)

okurz: http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule

AdamWill: pfah, I never was in the scouts. I was one of the bad kids.

aaannz: Actually I think bmwqemu is the good place to put it. We have some defaults value already in there and some common handy subs too.

foursixnine: Shouldn't we move those defaults to a different place like a real common ground, or rename the file so it makes more sense to have it as common ground for this kind of generic and common items?... e.g OpenQA::Utils (AutoInst::Utils :) )?

okurz: that's something I planned for a long time (and I guess also others did) but did never manage to complete so far, see https://github.com/os-autoinst/os-autoinst/pull/628 for my start to extract stuff from `bmwqemu` with a sensible name. Feel free to continue there and do it properly as your perl skills probably exceed mine :-)

bmwqemu.pm
```diff
@@ -59,6 +59,10 @@ our $istty;
 our $direct_output;
 our $standstillthreshold = scale_timeout(600);
 
+# Known locations of OVMF (UEFI) firmware: first is openSUSE, second is
+# the kraxel.org nightly packages, third is Fedora's edk2-ovmf package.
+our @ovmf_locations = ('/usr/share/qemu/ovmf-x86_64-ms.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd', '/usr/share/edk2/ovmf/OVMF_CODE.fd');
```
AdamWill: I prefer this split across lines, one line per file location - as in my original PR - but tidy complained about that, so we get a giganto-line.

coolo: you can snip out tidy for a portion of code.

PR #626, opened 2016-10-19
svirt: support asset upload

mnowaksuse:
Bare bone verification run: http://assam.suse.cz/tests/4048.

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -98,7 +98,7 @@ sub can_handle {
 sub is_shutdown {
     my ($self) = @_;
     my $vmname = $self->console('svirt')->name;
-    my $rsp    = $self->run_cmd("! virsh list --state-shutoff | grep -w $vmname");
+    my $rsp    = $self->run_cmd("! virsh dominfo $vmname | grep -w 'shut off'");
     return $rsp;
 }
 
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -49,7 +49,6 @@ sub do_start_vm {
         });
 
     $ssh->backend($self);
-    $self->select_console({testapi_console => 'svirt'});
 
     # remove backend.crashed
     $self->unlink_crash_file;
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -92,6 +92,13 @@ sub can_handle {
     return;
 }
 
+sub is_shutdown {
+    my ($self) = @_;
+    my $vmname = "openQA-SUT-" . get_var('VIRSH_INSTANCE', 1);
+    my $rsp = $self->run_cmd("! virsh list --state-shutoff | grep -w $vmname");
+    return $rsp;
+}
+
 sub save_snapshot {
     my ($self, $args) = @_;
     my $snapname = $args->{name};
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -74,6 +74,7 @@ sub run_cmd {
     my $chan = $self->{ssh}->channel();
     $chan->exec($cmd);
     bmwqemu::diag "Command executed: $cmd";
+    $chan->send_eof;
     $chan->close();
     return $chan->exit_status();
 }
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -97,15 +97,15 @@ sub can_handle {
 
 sub is_shutdown {
     my ($self) = @_;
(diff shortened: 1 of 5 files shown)


backend/svirt.pm
```diff
@@ -49,10 +53,12 @@ sub do_start_vm {
         });
 
     $ssh->backend($self);
-    $self->select_console({testapi_console => 'svirt'});
 
     # remove backend.crashed
     $self->unlink_crash_file;
+
+    bmwqemu::save_vars();    # update variables
+
```
okurz: You are separating by a lot of blank lines, also in the above paragraph. IMHO doesn't make it more readable

backend/svirt.pm
```diff
@@ -91,6 +98,13 @@ sub can_handle {
     return;
 }
 
+sub is_shutdown {
+    my ($self) = @_;
+    my $vmname = "openQA-SUT-" . get_var('VIRSH_INSTANCE', 1);
+    my $rsp = $self->run_cmd("! virsh list --state-shutoff | grep -w $vmname");
```
okurz: Didn't @Soulofdestiny Add this just recently to the back end?

testapi.pm
```diff
@@ -1060,16 +1060,26 @@ Or you can upload public assets that will have a fixed filename
 replacing previous assets - useful for external users:
 
     upload_asset '/tmp/suse.ps', 1;
+
+If you just want to upload the file and verify that it was uploaded
+correctly on your own, add optional 'failok' parameter:
+
+    upload_asset '/tmp/suse.ps', 1, 1;
```
okurz: Why would you want this?

backend/svirt.pm
coolo: his PR is stll open

okurz: And I guess your comment there also applies here: https://github.com/os-autoinst/os-autoinst/pull/623/files#r83916516
I'd say whoever fixes this first, wins ;-)

testapi.pm
```diff
@@ -1048,7 +1048,7 @@ sub upload_logs {
 
 =head2 upload_asset
 
-  upload_asset $file [,$public];
+  upload_asset $file [,$public[,$failok]];
```
aaannz: Why is it called `$failok`? What about `$nowait`, `$noassert` or `$nocheck`?

backend/svirt.pm
```diff
@@ -36,6 +42,9 @@ sub new {
 sub do_start_vm {
     my ($self) = @_;
 
+    my $vars = \%bmwqemu::vars;
+    $vars->{NUMDISKS} ||= 1;    # this is what's supported now
```
aaannz: I don't get why the comment. Do you really mean just to filter out diskless systems?

Soulofdestiny:
@mnowaksuse - regarding https://github.com/os-autoinst/os-autoinst/pull/623 I think I will let this to you then, since it should be easier to handle this way ;)

But, can you check "is_shutdown" again? because I implemented it with grepping for "running" since the logic seems to be somehow inverted in backend/baseclass.pm

Soulofdestiny:
edit: nevermind - I missed your negation  -> :+1:

backend/svirt.pm
mnowaksuse: Nope, it was just meaningless remnant.

backend/svirt.pm
```diff
@@ -21,13 +21,19 @@ use testapi qw(get_var get_required_var check_var);
 
 use IO::Select;
 
+use Class::Accessor "antlers";
+has instance => (is => "rw", isa => "Num");
+has name     => (is => "rw", isa => "Str");
+
```
coolo: I would really prefer if the backend didn't duplicate this and just get the name from the console. Because you can change the name in the console (attach_to_running) and the backend wouldn't follow

mnowaksuse: Sure. But (1) I have no idea how do I, while in backend, get those properties from console, and (2) I am not sure if it's available at that point as console gets initialized later, I believe.

coolo: you don't need it in the constructor. you need it when you call the functions. So I would query the console at that time - and you get the console by its name.

mnowaksuse: Thanks. Now, hopefully addressed.

mnowaksuse:
Updated to address @coolo's review. New verification run: http://assam.suse.cz/tests/4127.

PR #627, opened 2016-10-20
Don't be too eager to shut off VNC screen

coolo:
The VNC stall detection didn't actually wait another 2 seconds, but would
disable the screen if there was no update in between 2 loop iterations, which
happens every 0.5 seconds. We actually *meant* to wait 2 more seconds before
disabling the screen, so add these 2 to the original 2


okurz:
does it work?

coolo:
https://tortuga.suse.de/tests/753/file/autoinst-log.txt -> 8 'stalled'
https://tortuga.suse.de/tests/754/file/autoinst-log.txt -> 0 'stalled'

PR #628, opened 2016-10-20
WIP -- check tests and test locally with isotovideo on vars.json -- Extract log functions from bmwqemu

okurz:


autotest.pm
```diff
@@ -56,7 +57,7 @@ sub loadtest {
     eval $code;    ## no critic
     if ($@) {
         my $msg = "error on $script: $@";
-        bmwqemu::diag($msg);
+        log::diag($msg);
```
okurz: "function of its own"? You mean package/file?

coolo: right.

foursixnine: Actually, the longer plan with #691 is to move it to a single package, and then use that one as base for bmwqemu... or simply add the logger everywhere (As bmwqemu is already present, i think inheritance can be a good trick too)

Ending up with:

```perl
package OpenQA::Log
use Log::Log4perl

package bmwqemu
use base OpenQA::Log
```

okurz: yes, that sounds good

okurz:
#691 should be better

PR #630, opened 2016-10-24
Add a flock together with the rsync in a shared needle directory for needle caching

foursixnine:
The needles are being copied form SOURCENEEDLES or PRODUCTDIR/needles
to a shared directory just one level above the current working
directory, allowing all of the workers share needles and using the
blocking flock to ensure that only one worker at a time is tyring to
rsync through isotovideo

Error code 999 will be triggered when the
lock can't be obtained.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -50,7 +50,7 @@ no autodie qw(kill);
 use Cwd;
 use POSIX qw(:sys_wait_h _exit);
 use Carp qw(cluck);
-use Time::HiRes qw(gettimeofday tv_interval sleep);
+use Time::HiRes qw(gettimeofday tv_interval sleep time);
 
 # avoid paranoia
 $Getopt::Std::STANDARD_HELP_VERSION = 1;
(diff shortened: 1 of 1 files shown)


isotovideo
```diff
@@ -172,12 +172,24 @@ sub signalhandler_chld {
...
-    # make sure the needles are initialized before the backend process is started
     no autodie qw(system);
-    my $res = system("rsync -avP --delete --exclude=.git $bmwqemu::vars{PRODUCTDIR}/needles/ $dir/needles/");
-    die "Failed to rsync needles: $@" if $res;
-    needle::init("$dir/needles");
+
+    #We want to use the pool directory instead of the
+    # cwd (which would be the worker's directory)
+    # to be able to share the needles between workers
+    my $shared_needle_dir = getcwd() . "/..";
```
coolo: the worker can't write in /var/lib/openqa/pool. Possibly in your installation, but not on production. You will have to rely on the admin to configure such a directory in workers.ini

okurz: can we then please also make the whole syncing behaviour optional configured over the workers ini? At least we could make the old behaviour optional. Maybe even make the sync call itself configurable.

aaannz: 1) Configurable cache dir, 2) don't forget that worker can run tests from different distri - I suggest to add distri to the cache dir and 3) there is the plan worker can be shared between more webuis - add openqa host to the path.

So at the end path should IMO be something like `File::Spec->catdir($cachedir_from_config, $openqa_host, $distri)`.

okurz:
:+1: great job!

PR #631, opened 2016-10-25
Query vars.json of another job (POO#14306)

nadvornik:
Usage:

use mmapi;
my $parents = get_parents();
my $vars = get_job_autoinst_vars($parents->[0]);
print $vars->{WORKER_ID}

The function returns current content of vars.json. It can be updated by calling bmwqemu::save_vars() if needed.

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -188,6 +188,13 @@ sub upload_file {
     return $self->render(text => "OK: $filename\n");
 }
 
+sub get_vars {
+    my ($self) = @_;
+
+    bmwqemu::load_vars();
+    return $self->render(json => {vars => \%bmwqemu::vars});
+}
+
 sub current_script {
     my ($self) = @_;
     return $self->reply->asset(Mojo::Asset::File->new(path => 'current_script'));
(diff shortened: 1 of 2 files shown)


coolo:
As you're adding it to the API you need to provide some kind of API documentation. How to use the function, in what contexts. What to expect, ... 

Similiar to what you put in the commit log

nadvornik:
It will be separate pull request on openQA repo

coolo:
I don't like this discrepancy between testapi and mmapi. Defining the scope of multi machine testing is most likely not relevant for POD, but the expected arguments can very well live in the code directly.

jerrytt:
job_id = 315 
url = http://linux-j7uw:20033/dwgYAut1bZujCa1Q

Mojo::Message::Response can not get response .

Use of uninitialized value in numeric eq (==) at /usr/lib/os-autoinst/mmapi.pm line 212.
Use of uninitialized value in concatenation (.) or string at /usr/lib/os-autoinst/mmapi.pm line 216.

The reason is the local dns does not work correctly . 

after replace http://linux-j7uw:20033/dwgYAut1bZujCa1Q ==> http://localhost:20033/dwgYAut1bZujCa1Q

I can get the expect result :)

nadvornik:
IMHO this is a problem of host network configuration.

PR #634, opened 2016-10-25
vnc_base: Prevent 'uninitialized' warning in mouse_move

okurz:


coolo:
I would rather die than to move to 0,0

okurz:
I wasn't sure if this isn't a valid use case after all.

okurz:
updated to die instead although I am not sure where this will fail.

coolo:
the warning is still there - and I didn't realize what you tried to fix (taking that your originally pushed a null backend with it) ;(

The uninitialized variable is in {mouse} not in the parameter.

okurz:
Sorry for the confusion. I am pretty sure the warning observed was within the parameters of course that could also be true for the mouse object

PR #636, opened 2016-10-27
testapi: Make type_password arguments configurable

okurz:


testapi.pm
```diff
@@ -1199,9 +1199,11 @@ Uses C<$testapi::password> if no string is given.
 
 =cut
 sub type_password {
-    my ($string) = @_;
+    my ($string, %args) = @_;
     $string //= $password;
-    type_string $string, max_interval => 100, secret => 1;
+    $args{max_interval} //= 100;
+    $args{secret} //= 1;
```
okurz: I know. So? I am just setting sane defaults in case any args are given as argument. Would you recommend I silently overwrite $args{secret} with "1" in case the user sets something different?

coolo: I'm saying you should hardcode secret and only accept max_interval as argument

coolo:
yours will work, but the way I would have done it:

```
type_string $password, secret => 1, max_interval => ($args{max_interval} // 100);

okurz:
yep, looks good, except the "$password" is only default if no string given. updated.

coolo:
it was pseudo code :)

PR #637, opened 2016-10-27
Add virtio serial console backend and API

richiejp:
This allows the test writer to log into and interact with a serial
terminal directly. Initially this is just for the virtio_console under
QEMU, but can be extended to any serial console.

Presently the only way of sending text to a tty running on a SUT is to
send the keystrokes via VNC. Output from the SUT can be redirected to
and read from, a serial port, however the test writer can not send text
directly to the serial port without circumventing the test API.

This patch introduces a new console backend which can be selected in the
same manner that the existing console backends are, but is limited to
text input and output. This means that there is no video feed, but
entering commands is orders of magnitude faster because the commands are
sent and interpreted as text rather than simulated keystrokes.

This adds the is_serial_terminal subroutine to the testapi which is
exported on request. Otherwise the API should be unchanged and fully
backwards compatible.


t/09-terminal.t
```diff
@@ -0,0 +1,340 @@
...
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+use 5.018;
+use warnings;
+use Carp qw( confess );
+use English qw( -no_match_vars );
+use POSIX qw( :sys_wait_h pause );
+use Socket qw( PF_UNIX SOCK_STREAM sockaddr_un );
+use Time::HiRes qw( usleep );
+
+use Test::More tests => 12;
```
aaannz: We stopped using counted tests and use `done_testing` instead at the end of tests.

t/09-terminal.t
```diff
@@ -0,0 +1,340 @@
...
+
+# Either write $msg to the socket or die
+sub try_write {
+    my ($fd, $msg) = @_;
+
+    open(my $logfd, '>>', $log_path);
+    print $logfd $msg;
+    close($logfd);
+
+  WRITE: while (1) {
```
aaannz: Why are you using labels instead of plain while - next - last? Everytime I read labels somewhere I start to expect some weird code flow.

backend/baseclass.pm
```diff
@@ -653,9 +658,16 @@ sub wait_serial {
     my $matched = 0;
     my $str;
 
+    if ($self->{current_console}->is_serial_terminal) {
+        #$args->{do_while_idle} = sub { $self->run_capture_loop(undef, 0.1, 1, 10); };
```
richiejp: I will remove it.

t/09-terminal.t
richiejp: OK, but this is not recommended by the unit test library author.

t/09-terminal.t
richiejp: According to the Perl documentation I should be able to remove the while(1) and just use a plain block with a label, which I think is preferable because the intention is to only loop if EAGAIN happens, but it didn't work for some reason and left the loop when it shouldn't. I left the label there for readability, but I can remove it if it just makes people worried.

distribution.pm
```diff
@@ -109,7 +110,12 @@ sub script_run {
     testapi::type_string "$cmd";
     if ($wait > 0) {
         my $str = testapi::hashed_string("SR$cmd$wait");
-        testapi::type_string " ; echo $str-\$?- > /dev/$testapi::serialdev\n";
+        if (testapi::is_serial_terminal) {
+            testapi::type_string " ; echo $str-\$?-\n";
+        }
+        else {
+            testapi::type_string " ; echo $str-\$?- > /dev/$testapi::serialdev\n";
+        }
```
okurz: not so clean. If the implementations differ then this should be abstracted in the individual backends or consoles. So better provide a proposal of how to abstract that functionality in the testapi itself. The current tests use some redirection to the serial devices a lot. I assume if we can't abstract this away properly into os-autoinst this might be a hint that direct usage of a serial port is not a good idea to be combined with what os-autoinst is currently used for.

t/03-testapi.t
```diff
@@ -108,13 +114,16 @@ my $details = $autotest::current_test->{details}[-1];
 is($details->{title}, 'Soft Failed', 'title for soft failure added');
 like($details->{text}, qr/basetest-[0-9]+.*txt/, 'file for soft failure added');
 
+require distribution;
+testapi::set_distribution(distribution->new());
+select_console('a-console');
+is(is_serial_terminal, 0, 'Not a serial terminal');
```
okurz: :+1:

t/09-terminal.t
okurz: yes, and we still decided to not count manually :-)

t/09-terminal.t
```diff
@@ -0,0 +1,340 @@
...
+BEGIN {
+    unshift @INC, '..';
+}
+
+use consoles::virtio_terminal;
+use testapi ();
+
+our $VERSION;
+
+my $log_path = './09-terminal-io.log';
```
okurz: I suggest from the tests to treat the workspace as read-only and use proper temp files or directories if necessary.

t/09-terminal.t
```diff
@@ -0,0 +1,340 @@
...
+# so wait for fake terminal to create socket and emit SIGCONT. Pause only
+# returns if a signal is received which has a handler set.
+$SIG{CONT} = sub { };
+pause;
+
+test_terminal_directly;
+
+$SIG{CHLD} = sub { };
+waitpid($pid, 0);
+report_on_fake_terminal;
```
okurz: wow, that's a lot of testing :-)

testapi.pm
```diff
@@ -18,7 +18,7 @@ package testapi;
 
 use base Exporter;
 use Exporter;
-use strict;
+use 5.018;
```
okurz: why?

testapi.pm
```diff
@@ -525,19 +556,20 @@ sub wait_serial {
 
     # wait for a message to appear on serial output
     my $regexp           = shift;
-    my $timeout          = shift || 90;    # seconds
-    my $expect_not_found = shift || 0;     # expected can not found the term in serial output
+    my $timeout          = shift || 90;                                      # seconds
+    my $expect_not_found = shift || 0;                                       # expected can not found the term in serial output
```
okurz: that comment moved a bit far away from the code. I suggest to either add a blank line after this line and fix indendation or put the comment above each code line.

testapi.pm
```diff
@@ -619,7 +651,12 @@ sub assert_script_run {
     }
     my $str = hashed_string("ASR$cmd");
     # call script_run with idle_timeout 0 so we don't wait twice
-    script_run("$cmd; echo $str-\$?- > /dev/$serialdev", 0);
+    if (is_serial_terminal) {
+        script_run("$cmd; echo $str-\$?-", 0);
+    }
+    else {
+        script_run("$cmd; echo $str-\$?- > /dev/$serialdev", 0);
```
okurz: see my comment about abstraction of differing behaviour in testapi and backend

testapi.pm
```diff
@@ -636,6 +673,8 @@ C<$wait_seconds> defaults to 2 seconds.
 
 I<The implementation is distribution specific and not always available.>
 
+TODO: Make this compatible with serial terminals
```
okurz: better not add 'TODOs' to the code which will stay there forever unless someone wants to really fix it. I suggest to keep "open points" in tickets and incomplete implementation including TODOs in code within your local repo. You can keep branches with experimental content.

testapi.pm
```diff
@@ -681,18 +722,33 @@ The default timeout for the script is 30 seconds. If you need more, pass a secon
...
+        type_string($current_test_script, terminate_with => 'EOT');
+        wait_serial("$suffix-0-");
+    }
+    else {
+        open my $fh, ">", 'current_script' or die("Could not open file. $!");
+        print $fh $current_test_script;
+        close $fh;
+        assert_script_run "curl -f -v " . autoinst_url("/current_script") . " > /tmp/script$suffix.sh";
+        script_run "clear";
+    }
```
okurz: see above about abstraction

testapi.pm
```diff
@@ -1387,7 +1452,7 @@ here.
 
 =cut
 sub console {
-    my ($testapi_console) = @_;
+    my $testapi_console = shift || $selected_console;
```
okurz: we are commonly using "modern style", that is:

```perl
my ($testapi_console) = @_;
$testapi_console //= $selected_console;
```

consoles/virtio_screen.pm
```diff
@@ -0,0 +1,251 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+package consoles::virtio_screen;
+use 5.018;
+use warnings;
+use English qw( -no_match_vars );
+use Time::HiRes qw( gettimeofday usleep );
+use integer;
```
aaannz: `use 5.018` implies `use strict;` ( it does so since `5.012` )

t/09-terminal.t
aaannz: For the consistency of codebase I'm for removal of labels and just hide it under plain while-for cycles. Unless really necessary.

testapi.pm
okurz: yeah, sorry. In this case I meant

```perl
my ($testapi_console) = @_;
$testapi_console ||= $selected_console;
```

t/09-terminal.t
richiejp: OK, in one other place I have used it, it is actually necessary, but otherwise I will remove them.

distribution.pm
richiejp: This is a good point, I will think about it, there is no way of the distribution or backend overriding some platform specific code in the testapi in the first place so it should be refactored anyway. I don't see a major conceptual problem with doing both input and output to a serial terminal in os-autoinst, it is just another IO method that isn't too different from the current VNC-serial-port hybrid method. In this particular case we just need to provide an abstraction for a "command delimiter".

Having said that I think it is OK to have an if statement with alternatives for graphical and text terminals because it is very unlikely there will be a third or forth type for the foreseeable future, but the testapi should not have any shell specific code in it.

t/09-terminal.t
richiejp: OK, I will put it in /tmp or disable it unless some command line arg is specified, Travis does not need to write it out.

testapi.pm
richiejp: ok

t/09-terminal.t
okurz: I guess [File::Temp](http://perldoc.perl.org/File/Temp.html) is the way to go

distribution.pm
aaannz: Let's focus on this refactoring in next PR, I would like to merge this PR this week.

aaannz:
Btw. how much WIP is this now? Apart some of the mentioned issues it looks good to me. Do you plan some more big changes?

richiejp:
Based on what Oliver said I am exploring some options to increase decoupling: https://progress.opensuse.org/issues/14626

I also have a big change in my local branch which cuts down on the occurrences of is_serial_terminal and moves a lot of OS specific code out of testapi and into distribution.pm, but doesn't go as far as the progress feature. I haven't decided whether I think that is too much for now, perhaps it should be done in a different pull request.

If all of that is deferred until a later time or refused outright, then I could remove the WIP.

okurz:
I don't think you need to restructure everything in the first take but I would appreciate if you can attend the already commented minor issues so that we don't forget these. If it is feasible I suggest to add less in one go and start simple but that is also for consideration.

richiejp:
I forgot to do one or three things before pushing, so commits I would usually squash are present.

@aaannz the lock api tests you wrote were not enabled and name clashed with the terminal tests. So I renamed my test and enabled the lock api tests. Did you deliberately leave them disabled?

richiejp:
I forgot to do one or three things before pushing, so commits I would usually squash are present.

@aaannz the lock api tests you wrote were not enabled and name clashed with the terminal tests. So I renamed my test and enabled the lock api tests. Did you deliberately leave them disabled?

aaannz:
> the lock api tests you wrote were not enabled and name clashed with the terminal tests. So I renamed my test and enabled the lock api tests. Did you deliberately leave them disabled?

Not deliberately, I missed makefile edit. Thanks for enabling them.

aaannz:
My :+1: stands. Any further comments @okurz ?

okurz:
can you start the git subject lines with capital letter after each colon, too? It's a trivial task using `git rebase --interactive`, I guess you know how to do it. Otherwise, disregard, not a big issue.

richiejp:
I have squashed a couple of commits and changed the grammar. I may have to delete my remote github branch and create a new pull request though.

okurz:
you should have been able to just force push into a branch with the same name. Anyway, I will take a look into #644

richiejp:
Ah, yes, that is useful to know.

PR #638, opened 2016-11-04
Fix needle loading for remote workers when cache is enabled

foursixnine:
The initialization of needles will set the shared_cache variable,
in this case .json files will be loaded from the original location
and the .png files will be loaded from the cache, avoiding workers
to run jobs forever.

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -30,6 +30,7 @@ use Time::HiRes qw(time);
 our %needles;
 our %tags;
 our $needledir;
+our $shared_cache;
 our $cleanuphandler;
 
 sub new {
@@ -81,7 +82,13 @@ sub new {
     $self->{file} = $jsonfile;
     $self->{name} = basename($jsonfile, '.json');
     my $png = $self->{png} || $self->{name} . ".png";
-    $self->{png} = File::Spec->catpath('', dirname($jsonfile), $png);
+
+    my $original_needles_dir = dirname($jsonfile);
+    $self->{png} = File::Spec->catpath('', $original_needles_dir, $png);
+
+    # we want to load needles from the cache if it is present
+    $self->{png} =~ s/$original_needles_dir/$shared_cache/ if ($shared_cache);
+
     if (!-s $self->{png}) {
         warn "Can't find $self->{png}";
         return;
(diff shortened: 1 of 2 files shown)


needle.pm
```diff
@@ -195,7 +202,9 @@ sub wanted_ {
 }
 
 sub init {
-    $needledir = shift if @_;
+    $needledir    = shift if @_;
+    $shared_cache = shift if @_;
```
aaannz: I'm for `($needledir, $shared_cache) = @_;`

aaannz:
Actually, what is the problem fixing here? Race during rsync with other workers? Why should job run forever with reading json files from cache?

needle.pm
coolo: this is changing the behaviour though. In one cache you leave a previously defined value and in your case it will set it to undef.

I wonder though if we care.

aaannz: What do you mean by 'in one cache'? I expect init sub to change everything needed. IOW I don't expect init subs to be idempotent.

needle.pm
```diff
@@ -81,7 +82,13 @@ sub new {
     $self->{file} = $jsonfile;
     $self->{name} = basename($jsonfile, '.json');
     my $png = $self->{png} || $self->{name} . ".png";
-    $self->{png} = File::Spec->catpath('', dirname($jsonfile), $png);
+
+    my $original_needles_dir = dirname($jsonfile);
+    $self->{png} = File::Spec->catpath('', $original_needles_dir, $png);
+
+    # we want to load needles from the cache if it is present
+    $self->{png} =~ s#$original_needles_dir#$shared_cache#g if ($shared_cache);
```
foursixnine: @aaannz Mine too, i just panic when there are slashes in the string i want to replace... anywho, changed it, looks beautiful now :+1:

needle.pm
foursixnine: i changed it to be  `($needledir, $shared_cache) = @_; ` $shared_cache will be undefined anyways, and no harm will come if the cache is not enabled.

foursixnine:
@aaannz, when the needle caching is activated... remote workers are unable to find png files.

[This ](https://openqa.opensuse.org/tests/299065 )is one of the failed jobs reported by coolo
[This](http://deimos.suse.de/tests/726) is a run on a remote worker (lanner.arch)

PR #639, opened 2016-11-05
Fix cache usage for needles (again)

coolo:
Using regular expressions to replace paths is problematic - and using / is
even more problematic

But that's what happening if you trust reviewer's advise :)

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -84,10 +84,8 @@ sub new {
     my $png = $self->{png} || $self->{name} . ".png";
 
     my $original_needles_dir = dirname($jsonfile);
-    $self->{png} = File::Spec->catpath('', $original_needles_dir, $png);
-
     # we want to load needles from the cache if it is present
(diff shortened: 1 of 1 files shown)


coolo:
This is now deployed on openqa.opensuse.org and works

aaannz:
Whats wrong with `/`? This isn't bash, variable expansion is done after parsing substitute.

coolo:
It confuses me :)

needle.pm
```diff
@@ -202,7 +200,7 @@ sub wanted_ {
 }
 
 sub init {
-    my ($needledir, $shared_cache) = @_;
+    ($needledir, $shared_cache) = @_;
```
okurz: why this?

coolo: There was a misunderstanding: https://github.com/os-autoinst/os-autoinst/pull/638#discussion_r86559861

Issue #640, opened 2016-11-07
qemu backend: specifying boot order with -boot order=X doesn't work with UEFI/OVMF


garretraziel:
This applies to Fedora with os-autoinst-4.4, qemu-2.6.2 and both edk2-ovmf-20160418gita8c39ba (from Fedora stable repos) and newest edk2-ovmf from github, but I suppose that this is also true for openSUSE.

tl;dr: `-boot order=X` that os-autoinst uses for specifying what to boot doesn't really work on UEFI (at least on Fedora). We should replace it with `bootindex=0` on the device we want to boot from. We should also consider adding support for handling `OVMF_VARS.fd`.

Long story:
os-autoinst is using qemu parameter `-boot order=c` to boot from disk, `-boot order=d` to boot from cdrom and `-boot once=d` to boot from cdrom on the first boot, but from disk after that. os-autoinst is also using the same parameters for handling UEFI boot order and it seemingly works, except that it doesn't. UEFI support is provided by OVMF and if we look into qemu+ovmf whitepaper [here](http://www.linux-kvm.org/downloads/lersek/ovmf-whitepaper-c770f8c.txt), we can see that `-boot order=X` is legacy command option for UEFI and also, after trying it on Fedora, it doesn't work for me at all. The same documentation also mentions that we should add `bootindex=X` to devices to specify boot order instead.

OVMF boots from cdrom by default and during installation, when bootloader is installed, installator adds new boot option for OS and sets it as default option, so it fortunately behaves like specifying `-boot once=d` out of the box. Problem is when you want to boot from disk - when user wants to boot from disk, we should add `bootindex=0` to first connected disk.

[...]

PR #641, opened 2016-11-07
use bootindex for boot order on UEFI

garretraziel:
This should partly solve #640. It worked with OVMF from git, but there was some problem with `bootindex=0` with OVMF from stable Fedora repo. I would appreciate if you'd test in on your machine with your version of OVMF.

This code appends `-boot order=X` when on BIOS as before, but it uses better way how to specify boot order on UEFI with `bootindex=X`.

It also adds possibility to specify boot device verbosely as `BOOTFROM=cdrom` and `BOOTFROM=disk`, because `c` and `d` don't really make sense in context of UEFI, although it leaves old behaviour as it is.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -275,9 +275,18 @@ sub start_qemu {
         die "no dmi data for '$vars->{LAPTOP}'\n" unless -d "$bmwqemu::scriptdir/dmidata/$vars->{LAPTOP}";
     }
 
+    my $bootfrom = "";    # branch by "disk" or "cdrom", not "c" or "d"
     if ($vars->{BOOT_HDD_IMAGE}) {
         # skip dvd boot menu and boot directly from hdd
         $vars->{BOOTFROM} //= 'c';
+    }
+    if ($vars->{BOOTFROM} eq "d" || $vars->{BOOTFROM} eq "cdrom") {
+        $bootfrom = "cdrom";
+        $vars->{BOOTFROM} = "d";
+    }
+    if ($vars->{BOOTFROM} eq "c" || $vars->{BOOTFROM} eq "disk") {
+        $bootfrom = "disk";
+        $vars->{BOOTFROM} = "c";
     }
 
     my $iso = $vars->{ISO};
@@ -323,6 +332,9 @@ sub start_qemu {
     my $arch_supports_boot_order = 1;
     my $use_usb_kbd;
     my @vgaoptions;
+    if ($vars->{UEFI}) {    # UEFI/OVMF supports ",bootindex=N", but not "-boot order=X"
+        $arch_supports_boot_order = 0;
+    }
     if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm') {
         push @vgaoptions, '-device', 'VGA';
         $arch_supports_boot_order = 0;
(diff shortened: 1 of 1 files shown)


garretraziel:
Info about why is this necessary is in #640.

richiejp:
I haven't tested it (yet), but your reasoning seems sound.

I am wondering if the code could be simplified slightly by always setting bootindex=0 for the device we want to boot from? I think that if no firmware is present then it will just be ignored.

garretraziel:
That's true and works. We can add both `-boot order=X` to UEFI tests and `bootindex=0` to BIOS test and it's ignored without warning. I haven't done that not to mystify someone who will be reading `autoinst.log` who might think that it actually works :-).

richiejp:
Ideally yes, only necessary options should be present, but I think the QEMU backend needs an overhaul, to switch to a more declarative style to make that practical. So I would remove any unnecessary logic for now, to keep it as simple as possible.

More importantly though, I tested UEFI boot with the 'generic' OVMF rom which comes in the qemu-ovmf-x86_64 package without your patch and it booted from the wrong device. With your patch it booted from the correct device. OpenSUSE Tumbleweed was used on guest and host.

If a maintainer would like to decide on the style issue, then I think it is good to go.

dzedro:
Why not drop `-boot order` and use only `bootindex` ?

garretraziel:
@dzedro from what I tried, BIOS doesn't support `bootindex=0`.

dzedro:
@garretraziel I tried it now and it does work well also with non-UEFI
I'm not sure if bootindex is starting with 0, have to check some documentation

```
/usr/bin/qemu-system-x86_64 -global isa-fdc.driveA= -vga cirrus -m 1024 -cpu qemu64 -netdev user,id=qanet0 -device virtio-net,netdev=qanet0,mac=52:54:00:12:34:57,bootindex=1 -device virtio-scsi-pci,id=scsi0 -device virtio-blk,drive=hd1,bootindex=3 -drive file=3,cache=unsafe,if=none,id=hd1,format=qcow2 -drive media=cdrom,if=none,id=cd0,format=raw,file=/var/lib/openqa/factory/iso/SLE-12-SP2-Server-DVD-x86_64-GM-DVD1.iso -device scsi-cd,drive=cd0,bus=scsi0.0,bootindex=2 -device usb-ehci -device usb-tablet -smp 1 -enable-kvm -drive media=cdrom,if=none,id=cd5,format=raw,file=/var/lib/openqa/factory/iso/SLE-12-SP2-RT-DVD-x86_64-Build0032-Media1.iso -device scsi-cd,drive=cd5,bus=scsi0.0
```

dzedro:
works with 0 same way, I think it doesn't matter what is the value, on if it's less or greater, tried  bootindex=10,11,13 with same result

richiejp:
It might depend on the version of (Sea)BIOS as well, perhaps only newer versions support bootindex.

okurz:
if you need it to run I would be ok for me to have support for both `bootindex` and `boot`.

dzedro:
I tested it and UEFI doesn't work with our ovmf, but it's problem of the ovmf, this PR is OK
Of course non-UEFI works well also with bootindex :)

dzedro:
sorry it was my bad, it's working well with UEFI

backend/qemu.pm
```diff
@@ -275,10 +275,19 @@ sub start_qemu {
         die "no dmi data for '$vars->{LAPTOP}'\n" unless -d "$bmwqemu::scriptdir/dmidata/$vars->{LAPTOP}";
     }
 
+    my $bootfrom = "";    # branch by "disk" or "cdrom", not "c" or "d"
     if ($vars->{BOOT_HDD_IMAGE}) {
         # skip dvd boot menu and boot directly from hdd
         $vars->{BOOTFROM} //= 'c';
     }
+    if ($vars->{BOOTFROM} eq "d" || $vars->{BOOTFROM} eq "cdrom") {
```
okurz: This causes warnings when `BOOTFROM` variable is not defined at all. Please fix.

okurz: Fixed by https://github.com/os-autoinst/os-autoinst/pull/647

okurz:
We are having more problems with this, e.g. see https://progress.opensuse.org/issues/15178 . Problem is that "-boot=once" does not work anymore or at least is not specified. Has anyone looked into the bootoptions more and has an idea or should we handle it by explicitly go through the boot menues all the time?

dzedro:
bootindex doesn't have any special [documentation](http://wiki.qemu.org/download/qemu-doc.html) or is missing, I guess there are no such options as with -boot like "once", so with bootindex we should go explicitly through the boot menus.

garretraziel:
I'm not sure where is the problem (link to openQA in https://progress.opensuse.org/issues/15178 doesn't work for me), but `-boot=once` for non-UEFI architectures should work as before (and if it doesn't, then there is a bug) and, for UEFI, it sadly looks like there is nothing which would mimic this behaviour.

garretraziel:
BTW, documentation for OVMF (UEFI) under QEMU is here: http://www.linux-kvm.org/downloads/lersek/ovmf-whitepaper-c770f8c.txt

PR #642, opened 2016-11-09
Fix crash on caching needles

coolo:
The issue is a bit complicated to explain, but perl and opencv don't go
*too* well together because opencv uses threads and perl and threads in
general don't go together.

So we have to be extra careful - so call rsync early in the bootup

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -171,33 +171,32 @@ sub signalhandler_chld {
     }
 }
 
+sub sync_needles_to_cache {
+
+    no autodie qw(system);
+
+    my $needles_source = $bmwqemu::vars{PRODUCTDIR} . "/needles";
+    my $shared_cache = File::Spec->catdir($bmwqemu::vars{CACHEDIRECTORY}, $bmwqemu::vars{OPENQA_HOSTNAME}, $bmwqemu::vars{DISTRI});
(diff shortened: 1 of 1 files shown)


coolo:
Read about the reasoning on https://progress.opensuse.org/issues/14694

okurz:
@foursixnine

isotovideo
```diff
@@ -246,6 +250,11 @@ $bmwqemu::vars{TEST_GIT_HASH} = $test_git_hash;
 # is not supposed to talk to the backend directly
 ($cpid, $cfd) = commands::start_server($bmwqemu::vars{QEMUPORT} + 1);
 
+# rsync before we read main.pm - it might trigger opencv and we need
+# to move this as late as possible due to the evil udev thread hooks
+# opencv adds
+$shared_cache = sync_needles_to_cache;
```
foursixnine: i'd change this to
` sync_needles_to_cache if $bmwqemu::vars{CACHEDIRECTORY}; `
just to remove that `if` inside the   `sync_needles_to_cache` and that extra variable. But other than that looks good to me

coolo:
oops, this isn't merged yet? then I just deployed a broken worker :(

okurz:
LGTM

PR #644, opened 2016-11-14
Add virtio serial console backend and API

richiejp:
https://github.com/os-autoinst/os-autoinst/pull/637


okurz:
test fail and seem to show a lot of warnings about uninitialized $serialdev, too

richiejp:
It is a race condition in the unit test, I still need to fix it.

richiejp:
The $serialdev variable has always been unset, I shouldn't have done anything to effect how it is initialized.

t/10-terminal.t
```diff
@@ -63,7 +63,7 @@ my $next_test             = "GOTO NEXT\n";
 
 # If test keeps timing out, this can be increased or you can add more calls to
 # alarm in fake terminal
-my $timeout = 30;
+my $timeout = 60;
```
okurz: Wait, how long does the test take?

richiejp:
It only takes about 1-2 seconds on my machine, but I set a much higher timeout so that it never fails due to being ran in a much slower environment. I reduced it back to 30 seconds though.

It turns out there was a race condition which I reproduced using sleep and have now fixed.

okurz:
yes, I can confirm it only takes 2 seconds on my machine, too. *phew*, thought it would take much longer. I wouldn't have like that ;-)

The tests some to output a lot of stuff. That is not your fault as most of the tests do this. Care to take a look into [Test::Output](https://metacpan.org/pod/Test::Output) and make the output nicer? I suggest you compare to what we do in os-autoinst/openQA. By catching the expected output we could prevent overlooked warnings.

okurz:
so you don't want to use `Test::Output`?

richiejp:
I don't think it is worth writing or maintaining unit tests to check the logging output in this particular case. If it proves to be more important then I will add some, it is me that will use them the most, so it is in my interest to make sure they work. However I have redirected the output to another temp file to clear up the noise and also check STDERR is clear of any unexpected messages to catch Perl warnings.

I am not actually sure this is better though, a better solution might be implemented in the Makefile/Test harness to redirect STDERR to a script and parse it. When I am running the unit tests in development it is quite useful to see all the output interleaved, although I can comment out the redirection in the script so it is not much of an issue.

richiejp:
I could perhaps use Test::Output if it lets me nest tests, but I thought just redirecting all output would be as easy unless I start testing the output of individual functions.

okurz:
good enough, great :-)

richiejp:
:+1:

coolo:
I tested this and it looks ok. So please rebase (replacing die with croak)

PR #645, opened 2016-11-16
Destroy VM on do_stop_vm()

mnowaksuse:
Virtual machine managed by svirt backend is now being destroyed on
do_stop_vm() call.

Verification run: http://assam.suse.cz/tests/4128

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -63,6 +63,13 @@ sub do_stop_vm {
     my ($self) = @_;
 
     $self->stop_serial_grab;
+
+    unless (get_var('SVIRT_KEEP_VM_RUNNING')) {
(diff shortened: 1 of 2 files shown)


coolo:
I'm worried that this also destroys the kgraft use case. Please check with Ondrej

mimi1vx:
@coolo  you are right.

mnowaksuse:
OK, thanks @mimi1vx.
Updated the code to keep the VM running if `SVIRT_KEEP_VM_RUNNING` variable is defined. The VM is now also being undefined.

coolo:
I still don't want to danger the kgraft VMs. Once they are destroyed, we can't just rollback

coolo:
@mimi1vx ok with you?

PR #647, opened 2016-11-19
qemu: Only evaluate BOOTFROM if set to prevent warning

okurz:


okurz:
Problem was introduced [here](https://github.com/os-autoinst/os-autoinst/pull/641/files#diff-ce37f278fec25235d9f905442ecd8e5aR283)

backend/qemu.pm
```diff
@@ -325,18 +325,20 @@ sub start_qemu {
...
-        $vars->{BOOTFROM} = "c";
+    if (my $bootfrom_var = $vars->{BOOTFROM}) {
+        if ($bootfrom_var eq 'd' || $bootfrom_var eq 'cdrom') {
+            $bootfrom = 'cdrom';
+            $vars->{BOOTFROM} = 'd';
+        }
+        if ($bootfrom_var eq 'c' || $bootfrom_var eq 'disk') {
+            $bootfrom = 'disk';
+            $vars->{BOOTFROM} = 'c';
+        }
```
aaannz: As you are already touching this, can you change second `if` to `elsif` and add `else` block throwing exception on unknown/unsupported boot order?

okurz: done, please see my update

PR #648, opened 2016-11-21
Change the PNG compression

foursixnine:
The compression (Which was basically none), is basically disabled
with this change, altough the changes are not that big, it might improve
peformance on test and reduce "Slow IO" messages, when the system is
under heavy load.

The attached file shows the time each file took to be encoded/written. 

Note that times are expressed in ms, and write time > 3ms is a lot considering that i have a m2 drive, pretty fast. 

Filename | Encoding Time | Write Time | Total time
--- | --- | --- | ---
testresults/.thumbs/consoletest_setup-2.png	|0	|97	|97
testresults/install_and_reboot-15.png	|32	|9	|42
testresults/install_and_reboot-27.png	|22	|8	|31
qemuscreenshot/shot-0000001663.png	|52	|5	|57
qemuscreenshot/shot-0000001509.png	|32	|5	|38
testresults/install_and_reboot-26.png	|21	|5	|26

[benchmarktest.txt](https://github.com/os-autoinst/os-autoinst/files/603048/benchmarktest.txt)

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -271,7 +271,7 @@ bool image_write(Image *s, const char *filename)
   compression_params.push_back(CV_IMWRITE_PNG_COMPRESSION);
   // compression level from 0-9, default is 3 but we go lower/faster
   // and let openQA run optipng on those we want to save
-  compression_params.push_back(1);
+  compression_params.push_back(0);
 
   if (!imencode(".png", s->img, buf, compression_params)) {
     std::cerr << "Could not encode image " << filename << std::endl;


aaannz:
In my understanding, disabling compression actually produce bigger files -> more time writing -> more time in IO ops. How is this going to solve 'Slow IO' messages?

aaannz:
By the way, the table shows times before or after the change? Do you have similar benchmark for the other case for comparison?

coolo:
what @foursixnine forgot to mention is we debugged https://progress.opensuse.org/issues/14804 - and the video encoder is not the problem, but the time spent there is mostly in 
```
$img->write($filename)
```

And while it says 'slow IO' it is actually heavily CPU bound.

coolo:
Of course a stall of 11 seconds is not to be healed by changing O1 to O0, but we thought that its a good idea nonetheless to save CPU.

foursixnine:
@aaannz That's one of my fears, but actually the difference is not so big.

The table is after the change, i will upload later a comparison... including filesize... But writing is not the bigger bottleneck i fear.

coolo:
we continued with benchmarks and png is just a horrible idea. It's not yet clear how opencv manages to do that, but as a matter of fact level 0 or 7 doesn't make a difference. *But* if you do not give any level, it will use BEST_SPEED and save around 50%.

But it's still slow at that level. @foursixnine did much more advanced benchmarks, but mine shows that for 300 times writing a file that is about 500K as PNG it takes

* 8.5 seconds with compression 0
* 9.0 seconds with compression 7
* 5.4 seconds without compression level (default)
* 1.1 seconds to write the jpegs
* 0.3 seconds to write ppm

As we don't need this png most of the time, this is a lot of wasted CPU cyles.

We use the written pngs for 3 things:
 - as input for the video encoder -> there the file format doesn't matter
 - the worker may transfer last.png to the live view -> there we need png or jpeg
 - test results need to be small and lossless

So a possibly risky idea is:
  - use uncompressed ppm as input to the video encoder
  - the video encoder will then transform last.jpg async (and the video encoder might skip either this or the video creation based on files touched to avoid wasting CPU cycles not used because we're not in live mode, but only the worker knows, so the default should be conversion)
  - write out testresults uncompressed too and have the worker convert it before calling optipng. We might want to abuse the videoencoder for that too.

coolo:
compression level considered harmful -> moved to progress

PR #649, opened 2016-11-24
serial terminal: Make virtio chardev optional

richiejp:
Stop the error "Invalid parameter 'logfile'" by making the virtio console optional.

Possibly the log file parameter does not exist on older versions of QEMU. I need to determine if this is the case and if so how long it will take before we can use this feature. However in the mean time this patch will stop the reported error.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -732,6 +732,7 @@ sub start_qemu {
         if ($vars->{AUTO_INST}) {
             push(@params, "-drive", "file=$basedir/autoinst.img,index=0,if=floppy");
         }
+        bmwqemu::diag(`$qemubin -version`);
         bmwqemu::diag("starting: " . join(" ", @params));
 
         # don't try to talk to the host's PA
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -715,7 +715,7 @@ sub start_qemu {
             push(@params, "-k", $vars->{VNCKB}) if ($vars->{VNCKB});
         }
 
-        {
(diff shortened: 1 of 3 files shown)


foursixnine:
It's related to qemu versions... This will only work on versions > 2.6 (Patch for logfile parameter was introduced this year).

Maybe it's time to start thinking about having version support? (Which for me is kind of a bad idea, but can help to solve this kind of issues)

coolo:
Then we need to ensure select_console dies heavily and loudly if VIRTIO_CONSOLE is not set

richiejp:
select_console will die saying it can't open the virtio socket or that the socket doesn't exist. I can probably come up with a better solution, but not tonight.

coolo:
yes, it will die heavily - but not loud enough :)

richiejp:
It now checks for the var and croaks with a descriptive error message if it is not set.

We could run QEMU with `$qemubin -version` and parse the version number at the begining of `start_qemu` and then remove the logfile parameter if it is pre 2.6.1. I think that is better left to another pull request though.

For some reason lockapi unit test is failing, I don't think that is anything to do with my changes, is it? @aaannz

foursixnine:
@richiejp Adding the version to the logfile for now would be just fine, while we come with a solution for the qemu version problem.

richiejp:
@foursixnine do you mean document which version is required?

foursixnine:
@richiejp No, i mean that somewhere, add the output of $qemubin --version so that it can be actually seen on the logs.

PR #650, opened 2016-11-28
Track explicitly the state of the VNC stall detection

coolo:
The idea to check this based on timestamp is just too fragile. So set it
to 1 if we tried to do a force update and only afterwards set it to 2
and output a "stall detected" and set the screen to black once

This way we avoid the impression of a "loop" as described in
https://progress.opensuse.org/issues/14618


coolo:
method to test:

1. start isotovideo with qemu backend
2. wait a bit
3. kill -STOP <qemu>
4. isotovideo should after ~4 seconds output "VNC stall detected" *once*
5. isotovideo should afterwards continue not to find anything
6. kill -CONT <qemu>
7. isotovideo should continue to find screens again and continue testing

okurz:
thank you for the description of a test case

PR #653, opened 2016-12-02
Do not take screenshot when VNC connection is dead

mnowaksuse:
POO#14618

With qemu backend we use '-no-shutdown' Qemu option which prevents qemu
process to exit after it's guest OS quits. In svirt backend we don't
have such option so after OS is shutdown, the VNC connection is
terminated with it's domain - we should not take a screenshot in such a
case.

Should fix https://openqa.suse.de/tests/654678 but there may be other occurrences.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -858,8 +858,12 @@ sub assert_shutdown {
             bmwqemu::diag("Backend does not implement is_shutdown - just sleeping");
             sleep($timeout);
         }
-        if ($is_shutdown) {    # -1 counts too
-            $autotest::current_test->take_screenshot('ok');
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -859,7 +859,10 @@ sub assert_shutdown {
             sleep($timeout);
         }
         if ($is_shutdown) {    # -1 counts too
-            $autotest::current_test->take_screenshot('ok');
+                               # With svirt backend the VNC connection is dead, can't take a screenshot
```
coolo: please move the comments a bit so that perltidy does not align them. this looks silly :)

PR #654, opened 2016-12-02
Use more common 'qw()' style

okurz:


consoles/sshVirtsh.pm
```diff
@@ -18,11 +18,11 @@ package consoles::sshVirtsh;
 use base 'consoles::sshXtermVt';
```
coolo: we really should do something about this as well. Because in here we use base 'SINGLE' and in other cases use base qw(SINGLE).

base 'SINGLE' sounds like the more reasonable thing to do - perhaps there is a perlcritic for that

coolo:
I would merge it, but I would prefer if we squashed it with the fix for the single value cases - see https://github.com/os-autoinst/os-autoinst/pull/655

PR #655, opened 2016-12-03
Consistent qw usage

coolo:
This is going one step further than @okurz's origina PR

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -159,7 +159,7 @@ distclean-local: clean-local
 
 check-local:
 	$(srcdir)/tools/tidy --check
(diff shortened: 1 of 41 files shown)


coolo:
I would like to squash the 2 fixes so we only have one policy commit to destroy git blame

okurz:
+1

okurz:
LGTM

coolo:
So I squashed the commits now - it now looks like okurz also did the single quote thingy. If you prefer I reset the author, just tell - but we have to chose one :)

okurz:
reset the author of which commit? It looks like there are two commits, one by you and one by me, but I don't actually care

coolo:
there were 3 originally

PR #656, opened 2016-12-05
Correct needle directory

dzedro:
Fix interactive mode https://openqa.suse.de/tests/655726/file/autoinst-log.txt


okurz:
awesome! I am wondering, how did it hit us not in before?

coolo:
the cache feature isn't that old - and interactive mode not that well tested

PR #657, opened 2016-12-07
Change displayed match error from MSE to RMSE

coolo:
This is the a bit easier to understand value - see
https://progress.opensuse.org/issues/14886

I decided not to change the stored error in .json so I
can do the same change in openQA without having to worry about
migrations

### DIFF
--- a/ppmclibs/tinycv.pm
+++ b/ppmclibs/tinycv.pm
@@ -22,6 +22,7 @@ use warnings;
 use bmwqemu 'diag';
 
 use File::Basename;
+use Math::Complex 'sqrt';
 
 require Exporter;
 require DynaLoader;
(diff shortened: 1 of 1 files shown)


coolo:
I just realize openQA doesn't care about the error anyway

okurz:
LGTM

ppmclibs/tinycv.pm
```diff
@@ -22,6 +22,7 @@ use warnings;
 use bmwqemu 'diag';
 
 use File::Basename;
+use Math::Complex 'sqrt';
```
rwmanos: I think you could have avoided this import if you had saved the result of the square root calculated in map_error_to_avg. It's the same, isn't it? (just represented as percentage in the webui)

coolo: map_error_to_avg is in openQA - this is os-autoinst

PR #658, opened 2016-12-07
WIP: Change the backend to use ppm and videoencoder to create the pngs

foursixnine:
Change the backend to write ppm instead of png's,
also the `last.png` file would be written on request by a file named
`live_log` in the testresults directory on the pool.

Changed videoencoder to make it write the png when it has finished
encoding a new frame.

This PR is related to [Change the isotovideo backend to write ppm files](https://progress.opensuse.org/issues/14976)

For now png and ppm will coexist (but the ppm will be a png internally
too). The only problem so far i've encountered is that the last.png is
not updated as often as we would like to.

http://deimos.suse.de/tests/1232#live had a live execution and it
shown the problem described above, delay in the WebUI... which took at
one point almost a minute, but somehow the test results were not properly uploaded.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -37,7 +37,6 @@ my $framecounter = 0;    # screenshot counter
 
 # should be a singleton - and only useful in backend process
 our $backend;
-our $want_png = 0;
 
 use parent 'Class::Accessor::Fast';
 __PACKAGE__->mk_accessors(
@@ -182,8 +181,6 @@ sub run_capture_loop {
 
                 last if $time_to_timeout <= 0;
             }
-
-            $want_png = (-e bmwqemu::result_dir() . "/live_log") ? 1 : 0;
 
             my $time_to_update_request = ($update_request_interval // $self->update_request_interval) - ($now - $self->last_update_request);
             if ($time_to_update_request <= 0) {
@@ -388,8 +385,7 @@ sub enqueue_screenshot {
     $framecounter++;
 
     my $filename = $bmwqemu::screenshotpath . sprintf("/shot-%010d.ppm", $framecounter);
-    my $lastlink = $bmwqemu::screenshotpath . "/last.png";
-
+    
     my $lastscreenshot = $self->last_image;
 
     # link identical files to save space
@@ -404,13 +400,6 @@ sub enqueue_screenshot {
         $self->write_img($image, $filename) || die "write $filename";
         $self->last_image($image);
         $self->_last_screenshot_name($filename);
-
-        if ($want_png) {
-            no autodie 'unlink';
-            unlink($lastlink);
-            symlink(basename($self->_last_screenshot_name), $lastlink);
-        }
-
     }
 
     if ($sim > 50) {    # we ignore smaller differences
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -406,7 +406,6 @@ sub enqueue_screenshot {
         $self->_last_screenshot_name($filename);
 
         if ($want_png) {
-            print "We want the png";
             no autodie 'unlink';
             unlink($lastlink);
             symlink(basename($self->_last_screenshot_name), $lastlink);
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -37,6 +37,7 @@ my $framecounter = 0;    # screenshot counter
 
 # should be a singleton - and only useful in backend process
 our $backend;
+our $want_png = 0;
 
 use parent 'Class::Accessor::Fast';
 __PACKAGE__->mk_accessors(
@@ -181,6 +182,8 @@ sub run_capture_loop {
 
                 last if $time_to_timeout <= 0;
             }
+
+            $want_png = (-e bmwqemu::result_dir() . "/live_log") ? 1 : 0;
 
             my $time_to_update_request = ($update_request_interval // $self->update_request_interval) - ($now - $self->last_update_request);
             if ($time_to_update_request <= 0) {
@@ -384,7 +387,7 @@ sub enqueue_screenshot {
 
     $framecounter++;
 
-    my $filename = $bmwqemu::screenshotpath . sprintf("/shot-%010d.png", $framecounter);
+    my $filename = $bmwqemu::screenshotpath . sprintf("/shot-%010d.ppm", $framecounter);
     my $lastlink = $bmwqemu::screenshotpath . "/last.png";
 
     my $lastscreenshot = $self->last_image;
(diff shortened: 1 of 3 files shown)


videoencoder.cpp
```diff
@@ -347,6 +338,14 @@ main (int argc, char *argv[])
 
 	  rgb_to_yuv (&image, ycbcr);
 
+	std::string new_filename = filename;
+
+	new_filename = new_filename.replace(new_filename.end() - 4, new_filename.end(), ".png");
+
+	imwrite(new_filename, image);
+	unlink(filename);
+	link(new_filename.c_str(), filename); #tried with symlink but somehow the link is broken
```
foursixnine: this should be a symlink to avoid writing twice... In a quick test, tried it but the link was broken... will take a look into this later.

okurz:
It would be nice if your commit messages also mention the reason of the change and the benefit, not only the caveats and disadvantages ;-)

ppmclibs/tinycv_impl.cc
```diff
@@ -267,13 +267,8 @@ Image *image_read(const char *filename)
...
 {
   vector<uchar> buf;
-  vector<int> compression_params;
-  compression_params.push_back(CV_IMWRITE_PNG_COMPRESSION);
-  // compression level from 0-9, default is 3 but we go lower/faster
-  // and let openQA run optipng on those we want to save
-  compression_params.push_back(1);
 
-  if (!imencode(".png", s->img, buf, compression_params)) {
+  if (!imencode(".ppm", s->img, buf)) {
```
coolo: I think we should simplify this and plainly call imwrite. This avoids passing memory around

backend/baseclass.pm
```diff
@@ -401,9 +404,13 @@ sub enqueue_screenshot {
         $self->write_img($image, $filename) || die "write $filename";
         $self->last_image($image);
         $self->_last_screenshot_name($filename);
-        no autodie 'unlink';
-        unlink($lastlink);
-        symlink(basename($self->_last_screenshot_name), $lastlink);
+
+        if ($want_png) {
```
coolo: I would move this functionality into the videnencoder and only write out a png if live_log exists.

coolo: otherwise you read stale symlinks most of the time

videoencoder.cpp
```diff
@@ -347,6 +338,14 @@ main (int argc, char *argv[])
 
 	  rgb_to_yuv (&image, ycbcr);
 
+	std::string new_filename = filename;
+
+	new_filename = new_filename.replace(new_filename.end() - 4, new_filename.end(), ".png");
```
coolo: you can just append .png - so if we switch to .jpeg you don't have to hunt why you end up with ..png :)

okurz:
Seriously, please provide a benefit, not only the caveats. And please reference the ticket you are working on for this one.

backend/baseclass.pm
```diff
@@ -384,7 +384,7 @@ sub enqueue_screenshot {
 
     $framecounter++;
 
-    my $filename = $bmwqemu::screenshotpath . sprintf("/shot-%010d.png", $framecounter);
+    my $filename = $bmwqemu::screenshotpath . sprintf("/shot-%010d.ppm", $framecounter);
     my $lastlink = $bmwqemu::screenshotpath . "/last.png";
```
coolo: you can remove this variable then

ppmclibs/tinycv_impl.cc
coolo: how about we rename this function to write_ppm to make it clear we're doing something ugly if we do write_ppm('result.png')

videoencoder.cpp
```diff
@@ -1,5 +1,3 @@
-/* Modified by Stephan Kulow <coolo@suse.de> 2014 to take the PNGs
-   from stdin and use opencv directly */
```
coolo: you're changing history here

videoencoder.cpp
```diff
@@ -136,18 +128,18 @@ rgb_to_yuv (Mat * image, th_ycbcr_buffer ycbcr)
...
-
-	  yuv_y[x + y * yuv_w] =
-	    clamp ((65481 * r + 128553 * g + 24966 * b + 4207500) / 255000);
-	  yuv_u[x + y * yuv_w] =
-	    clamp ((-33488 * r - 65744 * g + 99232 * b + 29032005) / 225930);
-	  yuv_v[x + y * yuv_w] =
-	    clamp ((157024 * r - 131488 * g - 25536 * b + 45940035) / 357510);
-	}
+  {
+    unsigned char b = image->data[image->channels () * (image->cols * y + x) + 0];
```
coolo: are that tabs? Please a) do not replace spaces with tabs and b) do not use tabs at all.

foursixnine: Let me take a look, i don't remember doing that replace (Ghost sublime editor issues lol)

foursixnine:
@okurz Fixed the PR description, before doing the final update... i'll add the motivation to the last commit message

ppmclibs/tinycv_impl.cc
foursixnine: Well the idea is to have this function to be able to write either ppms, png or whatever... As we talked, i was thinking about changing it to imwrite which will itself call the encode.

coolo: yeah, but I love the fact that we're optimizing pngs that aren't :)

backend/baseclass.pm
foursixnine: You mean lastlink?

ppmclibs/tinycv_impl.cc
foursixnine: so, you're up for "write_ppm" ?... i still want to see if write_img can be overloaded... but i guess i'll do that as a side thing. (also passing the extension to write_img is a good idea)

videoencoder.cpp
foursixnine: @coolo did a re-check against the version in master, the file has a mix between tabs and spaces (and it needs some beautify too)... These are spaces actually

videoencoder.cpp
```diff
@@ -320,37 +318,45 @@ main (int argc, char *argv[])
...
+
+                if (!image.data) { //
+                    std::cout << "Could not open or find the image" << std::endl;
+                    return -1;
+                  }
+
+                rgb_to_yuv (&image, ycbcr);
+
+                cwd = get_current_dir_name();
+                livelog_file = cwd + "/live_log";
```
coolo: this is the complex form of stating 'live_log'

videoencoder.cpp
```diff
@@ -320,37 +318,45 @@ main (int argc, char *argv[])
...
+                cmd_offset = ftell(cmd);
+                fclose(cmd);
+        } else {
+                fclose(cmd);
+                sleep(1);
+                continue;
+        }
+
+        if (line[0] == 'E') {
+                const char *filename = line + 2;
```
coolo: The indenting style is now totally messed up. I guess I will do an emacs session with it when you're done :)

coolo:
I merged this into my branch: https://github.com/os-autoinst/os-autoinst/pull/664

PR #660, opened 2016-12-08
Do not use eatmydata anymore in qemu.pm backend

michelmno:
This is related to problem identified by bug
https://bugzilla.suse.com/show_bug.cgi?id=1009472#c16

Signed-off-by: Michel Normand <normand@linux.vnet.ibm.com>


coolo:
I'm actually fine with just dropping it unconditionally

coolo:
I remove the recommends from the package too

PR #661, opened 2016-12-08
svirt: Add functionality of copying images to boot from

Soulofdestiny:


coolo:
CC @mnowaksuse

consoles/sshVirtsh.pm
```diff
@@ -322,6 +322,10 @@ sub add_disk {
                 $self->run_cmd("qemu-img create $file $size -f qcow2") && die "qemu-img create failed";
             }
         }
+        elsif ($args->{copy}) {
+            $file = "/var/lib/libvirt/images/$file";
+            $self->run_cmd("rsync -avL $args->{file} $file") && die "rsync failed";
```
richiejp: Only a minor nit, but you could use single quotes for `rsync failed` because no substitution is intended.

coolo: the -L is rather suprising when copying files.

Soulofdestiny: yeah - this was a leftover from experimenting with symlinks

okurz:
LGTM

okurz:
oops, I merged the test, but can not merge this one … so coolo should work during the weekend to fix it

coolo:
TGIF

PR #663, opened 2016-12-08
Add SATA controller when SATA disk is present

garretraziel:
This adds Intel's ich9-ahci bus when `ide-drive` is specified as HDDMODEL (this looks like it is preferred way [when testing SATA](https://rubenerd.com/sata-on-qemu/)).

It has unfortunate side effect - it's not possible to connect CD via PATA and HDD via SATA at the same time, because when ich9-ahci is present, `CDMODEL=ide-cd` makes QEMU connect it via SATA, not PATA.

We need this when we want to test whether Fedora works on SATA.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -584,6 +584,11 @@ sub start_qemu {
             }
         }
 
+        if ($vars->{ATACONTROLLER}) {
+            # SATA devices need SATA controller
+            push(@params, "-device", "$vars->{ATACONTROLLER},id=ahci0");
+        }
+
         for my $i (1 .. $vars->{NUMDISKS}) {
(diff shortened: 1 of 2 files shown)


richiejp:
It might be better for you to set the ATACONTROLLER var on tests that need it, rather than make ich9-ahci the default. There may be existing instances where people are using ide-drive expecting the default QEMU behaviour.

garretraziel:
They can use `ide-hd` instead, but you are right, dropping automatic setting of `ATACONTROLLER` is backward-compatible.

coolo:
please add the variable to doc/backend_vars.asciidoc

garretraziel:
Done.

PR #664, opened 2016-12-09
Several changes to the image handling

coolo:


coolo:
this is still not ready. For once we need the worker to create the live_log file

coolo:
I tried to benchmark these changes and it is hard to judge if this is an improvement. I put 8 workers on a quadcore Xeon with HT and 16G (this is really not recommended :)

And I started these workers at the same time to grab 8 jobs - once with master and once with this branch and plot the first 10 minutes.

Before:

cpu

After:

cpu

So both times the machine gets really, really busy and IO bound, but with PPMs the situtation actually gets worse. There is a minute where the machine gets so busy that it gets almost nothing done:

procs

The problem though is that this can't be mapped to the PNG/PPM change directly - there is noise involved when you stress a system so badly. But it doesn't shout "PPM IS GOOD!" either :(

backend/baseclass.pm
```diff
@@ -395,23 +396,23 @@ sub enqueue_screenshot {
 
     my $mt1 = gettimeofday;
 
-    # 54 is based on t/data/user-settings-*
+    # we have two different similarity levels - one (slightly higher value, based
+    # t/data/user-settings-*) to determine if it's worth it to recheck needles
+    # and one (slightly lower as less signifant) determining if we write the frame
```
okurz: s/signifant/significant/

snd2png/snd2png.cpp
```diff
@@ -1,234 +1,205 @@
...
+    int times = info_in.frames / window_size - 1;
+    if (times * window_size + overlap > info_in.frames)
+        times--;
+
+    fprintf(stderr, "times %d\n", times);
+
+    double** points = (double**)malloc(sizeof(double*) * times);
+    double max_value = 0;
+
+    for (int TimePos = 0; TimePos < times; TimePos++) {
```
okurz: maybe you want to use *unsigned* here?

coolo: that's quite a bit more to type and we're talking about values around 1000

coolo:
So I extended the benchmark a bit. Now simulating 2 staging projects starting at the same time. 2 times one of rescue_system, RAID1, cryptlvm and minimalx starting at the same time on cold caches.

Before:
cpu

After:
cpu

The purple colour is the problem: these are the seconds when vmstat didn't out anything (stalled). One note though: the rescue_system tests are done after 150 seconds, so most of the time there are only 6 jobs running.

To summarize: this is not the improvement we hoped for.

okurz:
convincing figures

coolo:
I replaced all plots with a IMO clearer representation

coolo:
Next thing I tried was playing with the qemu IO - mainly from cache= (all based on master code otherwise untouched).

This cache=unsafe (what we use in master):
cpu

cache=none
cpu

cache=writeback (qemu default)
cpu

cache=writethrough
cpu

cache=directsync
cpu

The catch is: the VMs become so slow, that the RAID tests actually timeout (at around 800) - and the cryptlvm and minimalx tests fail too.

So slowing down qemu is definitely a way to get more reproducible results, but the options we have make them just too slow.

okurz:
Interesting findings. So would you say you found a way to prevent the stalls and at least we could "offer this option"?

coolo:
No, but I found a way to get an understanding what is going on :)

coolo:
http://git.qemu.org/?p=qemu.git;a=blob_plain;f=docs/throttle.txt;hb=HEAD is a really good attack vector I'd think

foursixnine:
@coolo but throttling the processes at a point might slow everything down so bad that tests could start failing even more (Similar to what I found while playing with nice levels)... 

Or you mean that the situation gets so bad that throttling the IO on qemu actually helps?. I haven't seem qemu lagging so bad other than when I used `stress-ng` or played with different nice leves...

coolo:
my preference is slowing down qemu linearly to keep predictable results - what we're seeing at the moment is the illusion of fast machines (mainly fast IO in this benchmark situation) up to a point where everything stalls for 12 seconds. Throttling needs to be done carefully of course, but I like to have options to test

foursixnine:
Makes sense then... +1

coolo:
iops-total=100 didn't make the tests fail, but it's not very convincing either

cpu

coolo:
this is with throttling.bps-write=1000000,throttling.bps-write-max=10000000 - and only the RAID tests fail. All the others are just very, very slow :)

cpu

coolo:
the RAID tests create a lot of IO during initializing - and the installation-started needle "only" waits 2 minutes

aaannz:
Did you try what impact has different backing storage? qcow2 vs. qed vs. raw for filebased storage? What about using lvm directly or ceph?

coolo:
I don't know what qed is - and raw is out of question as it does not support snapshots. And no idea what benefit we would gain from making the storage setup more complex. This is plain xfs on sda5

coolo:
So this is with throttling.bps-write=5000000,throttling.bps-write-max=100000000  (no tests failing)

cpu

coolo:
with throttling.bps-write=2500000,throttling.bps-write-max=100000000 we're moving into the right direction. No tests fail

cpu

coolo:
with throttling.bps-write=1800000 no tests fail, but are still very slow

cpu

coolo:
So this appears a fine option, but it's a lot of fine tuning. So @sysrich wanted to see the benchmark under different IO schedulers.

cfq is the default, so it's already listed - repeating it for easier comparision:
cpu

With nope no tests fail, but they are also very slow:
cpu

The deadline scheduler IO scheduler doesn't look much different (also no test fails, even though the tests are a bit quicker) - but it makes the whole webui feel *much* less responsive
cpu

CFQ with throttling still sounds like the better option.

coolo:
throttling.bps-write=2100000:
cpu

This is tricky - 2.1MB/s is just too slow.

aaannz:
To be clear, you changed IO schedulers for SUT or host? Don't we use NVMes now? There should be deadline or noop as default IMO.

coolo:
We use NVMEs on some of our production servers, yes - which only support nope.

But the benchmarked system is not a production server. It's a dell workstation system - and perfect to debug the problems we see in https://progress.opensuse.org/issues/13276. The production systems used for these do not have NVMEs either.

coolo:
and I didn't change anything about the SUTs in either tests, I only changed host.

coolo:
(and to be clear: almost nothing of my tests have anything to do with the changes in this pull request :)

mwilck:
Have your tried tuning  `vm.dirty_bytes` and `vm.dirty_background_bytes` sysctls on the host?

coolo:
tuned to? The default ratios are 20% and 10% here, that sounds good to begin with. Do you expect higher or lower to perform?

foursixnine:
@coolo Let's try with vm.dirty_background_bytes since we can "spare" some CPU, a high number here would help...

coolo:
sysctl vm.dirty_background_ratio=50:

cpu

mwilck:
Sorry, I didn't express myself clearly. I'd suggest to turn these **down** rather than up. On modern systems with lots of RAM, the `dirty_X_bytes` are preferrable against the `dirty_X_ratio`, because even 1% of a large physical RAM may be "too much".
My personal rule of thumb is: 'dirty_background_bytes=(# of bytes IO subsystem can typically write out in a second)`. `dirty_bytes=~10x that value`.

coolo:
setting dirty_background_ratio=dirty_ratio=5 indeed looked pretty much the same. Continueing to play with the settings

coolo:
So setting dirty_background_bytes looks really promising.

vm.dirty_background_bytes = 50000
vm.dirty_bytes = 5000000

cpu

And the tests are reasonably quick - unfortunately 2 out of 6 fail:

```
15:27:30.3871 6797 ||| finished first_boot installation at 2016-12-12 15:27:30 (67 s)
15:27:30.3874 6797 Creating a VM snapshot lastgood
DIE ERROR: timeout reading hmp socket
15:32:30.5330 6810 sending TERM to qemu pid: 6842
```

```
15:35:26.7928 6798 ||| finished zypper_up update at 2016-12-12 15:35:26 (39 s)
15:35:26.7931 6798 Creating a VM snapshot lastgood
DIE ERROR: timeout reading hmp socket

 at /usr/lib/os-autoinst/backend/baseclass.pm line 73.
	backend::baseclass::die_handler('ERROR: timeout reading hmp socket\x{a}') called at /usr/lib/os-autoinst/backend/qemu.pm line 883
	backend::qemu::_read_hmp('backend::qemu=HASH(0x676b898)') called at /usr/lib/os-autoinst/backend/qemu.pm line 950
	backend::qemu::_send_hmp('backend::qemu=HASH(0x676b898)', 'savevm lastgood') called at /usr/lib/os-autoinst/backend/qemu.pm line 207
	backend::qemu::save_snapshot('backend::qemu=HASH(0x676b898)', 'HASH(0x6ca10b8)') called at /usr/lib/os-autoinst/backend/baseclass.pm line 68
	backend::baseclass::handle_command('backend::qemu=HASH(0x676b898)', 'HASH(0x6ca0268)') called at /usr/lib/os-autoinst/backend/baseclass.pm line 448
	backend::baseclass::check_socket('backend::qemu=HASH(0x676b898)', 'IO::Handle=GLOB(0x66fef58)') called at /usr/lib/os-autoinst/backend/qemu.pm line 977
	backend::qemu::check_socket('backend::qemu=HASH(0x676b898)', 'IO::Handle=GLOB(0x66fef58)', 0) called at /usr/lib/os-autoinst/backend/baseclass.pm line 209
	eval {...} called at /usr/lib/os-autoinst/backend/baseclass.pm line 171
	backend::baseclass::run_capture_loop('backend::qemu=HASH(0x676b898)', 'IO::Select=ARRAY(0x66ea828)') called at /usr/lib/os-autoinst/backend/baseclass.pm line 120
	backend::baseclass::run('backend::qemu=HASH(0x676b898)', 6, 9) called at /usr/lib/os-autoinst/backend/driver.pm line 85
	backend::driver::start('backend::driver=HASH(0x54b9678)') called at /usr/lib/os-autoinst/backend/driver.pm line 48
	backend::driver::new('backend::driver', 'qemu') called at /usr/bin/isotovideo line 201
	main::init_backend() called at /usr/bin/isotovideo line 272
15:40:27.0571 6808 waitpid for 6862 returned 0
15:40:27.0571 6808 sending TERM to qemu pid: 6862
```

Not sure how big a qemu snapshot is, but 5 minutes is a lot of time too. So I guess I need to leave dirty_bytes high

mwilck:
> vm.dirty_background_bytes = 50000

That's **really low**, just 10 pages ... You could tune it up a little higher I think.

coolo:
This is quite frustrating. It's just so easy to screw this up.

vm.dirty_background_bytes = 500000
vm.dirty_ratio = 50

cpu

coolo:
So after this small detour to IO, let's come back to this PR. These are before and after on small dirty buffers.

PNG:
cpu

PPM:
cpu

coolo:
The CPU peaks are definitely smaller in the PPM version - but the IO load seems to be rather increased. Both on a very tiny scale though.

coolo:
As you're unlinking the images now in an indepdent process, make sure the backend is passing the image data instead of image filenames in results of check_screen (as discussed).

foursixnine:
@coolo Ok

coolo:
the DEBUG_IO is obviously quite noisy right now.

So is it possible to only output if we cross let's say 0.1s? I think everything below is just random :)

```
07:30:54.3019 17744 WARNING: enqueue_screenshot took 1.85 seconds - slow IO? (opencv: 0.01 - encoder: 1.84)
DEBUG_IO: for qemuscreenshot/shot-0000000633.ppm
NAME                        TIME        CUMULATIVE      PERCENTAGE
 scaling                     0.001       0.001           0.081%
 similarity                  0.004       0.006           0.240%
 write_encoder_frame E       1.841       1.847           99.674%
 _stop_                      0.000       1.847           0.005%
```

And now that we have a stopwatch, I think I also want to to see the load png functions stopped and only output if we the whole check_screen took longer :)

coolo:
And as expected, I get now crashes: Test died: Can't call method "write_with_thumbnail" on an undefined value at /usr/lib/os-autoinst/basetest.pm line 155.

This happens if the backend returns a filename that the videoencoder already deleted.

coolo:
So forget most of my previous benchmarks - I made a very bad mistake in believing the file systems were indepdendent. So here is the final before-after benchmark - and it shows the CPU usage didn't really change. 

cpu

cpu

But it's still quite strange - we have 

```
21:22:58.7089 6046 WARNING: enqueue_screenshot took 3.75 seconds
21:22:58.7090 6046 DEBUG_IO: 
NAME                        TIME        CUMULATIVE      PERCENTAGE
 scaling                     0.001       0.001           0.038%
 similarity                  3.746       3.747           99.897%
 convert ppm data            0.002       3.749           0.053%
 _stop_                      0.000       3.750           0.012%
```

3.7 seconds for a similarity call sounds a lot

coolo:
but it's quite rare now

coolo:
but we don't touch the file system there

foursixnine:
@coolo Is the warning message showing less often now right?... Anywho... maybe it's related more to a cpu problem rather than IO.

Maybe enabling the debug flags in tinycv_impl can show a hint?.

coolo:
the warning is still frequent enough to be worried. And if it's really similarity eating the time, it can't even be memory/swap

coolo:
The good news though: if the warning appears, it's because of similarity.
```
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.530       0.531           99.390%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.527       0.529           99.322%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.529       0.531           99.295%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.945       0.947           99.276%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.357       1.360           99.518%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.843       0.847           99.274%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.620       1.622           99.900%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.053       1.055           99.576%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.601       0.603           99.466%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.233       1.235           99.673%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  2.724       2.726           99.910%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.256       1.258           99.704%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.729       0.730           99.565%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.625       0.627           99.366%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.627       0.629           99.486%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.531       0.532           99.725%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  0.638       0.640           99.084%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.672       1.674           99.915%
/var/lib/openqa/pool/1/autoinst-log.txt: similarity                  1.166       1.168           99.843%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.576       0.578           99.205%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.925       0.927           99.578%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.634       0.636           99.082%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.158       1.159           99.567%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.530       0.533           99.171%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.573       1.575           99.909%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.993       1.995           99.808%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.469       1.471           99.687%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.305       1.307           99.751%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.715       0.716           99.499%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  1.245       1.247           99.882%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.625       0.626           99.495%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.833       0.834           99.624%
/var/lib/openqa/pool/7/autoinst-log.txt: similarity                  0.836       0.838           99.614%
```

So we only have to find out what's causing this - then we're done :)

coolo:
ok, getPSNR is easy to optimize: https://github.com/os-autoinst/os-autoinst/pull/674

Which changes the message to
```
NAME                        TIME        CUMULATIVE      PERCENTAGE
 scaling                     0.001       0.001           0.136%
 similarity                  0.006       0.007           0.557%
 convert ppm data            1.040       1.047           99.264%
 _stop_                      0.000       1.048           0.043%

coolo:
so we just turned from IO bound to memory bound  :)

coolo:
So I'm still seeing the warnings - but I'm clearly overbooking this machine. Next step is overbooking a production machine :)

coolo:
so let's get over this and merge

PR #665, opened 2016-12-13
QEMU: Add QEMU_NUMA var to enable NUMA

richiejp:
Setting QEMU_NUMA to 1 activates NUMA *simulation* so long as QEMUCPUS > 1.

This allows LTP tests to be ran which require NUMA, although it may not have the same effect as running it on actual NUMA hardware.


okurz:
LGTM

PR #666, opened 2016-12-13
Enable pipefail option for script_output

asmorodskyi:
By default bash return latest command result as result of pipe.
Changing to fail whole pipeline if at least one command fail.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -744,11 +744,11 @@ sub script_output($;$) {
 
     my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -e $run_script\n");
+        type_string("/bin/bash -oe pipefail $run_script\n");
         wait_serial($run_script, undef, 0, no_regex => 1);
     }
     else {
-        type_string "(/bin/bash -ex $run_script)| tee /dev/$serialdev\n";
+        type_string "(/bin/bash -eox pipefail $run_script)| tee /dev/$serialdev\n";
     }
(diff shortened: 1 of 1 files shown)


asmorodskyi:
@richiejp , @okurz please review

testapi.pm
```diff
@@ -744,11 +744,11 @@ sub script_output($;$) {
 
     my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -e $run_script\n");
+        type_string("/bin/bash -oe pipefail $run_script\n");
         wait_serial($run_script, undef, 0, no_regex => 1);
     }
     else {
-        type_string "(/bin/bash -ex $run_script)| tee /dev/$serialdev\n";
+        type_string "(/bin/bash -eox pipeline $run_script)| tee /dev/$serialdev\n";
```
okurz: s/pipeline/pipefail/

okurz:
based on the number of the pull request I suggest @Soulofdestiny can review, his style of music ;-)

richiejp:
May cause a load of tests to fail, but in a good way :-). Looks fine as far as the serial terminal is concerned.

asmorodskyi:
verification run http://kimball.arch.suse.de/tests/31#

coolo:
https://openqa.suse.de/tests/678902#step/fips_setup/12 looks suspicious like a fallout of this

okurz:
yes, because it worked in unusual ways in before.

Isn't it expected that a number #666 breaks something? ;-) See https://progress.opensuse.org/issues/15572 for next steps. I pinged qingming to take a look.

PR #667, opened 2016-12-14
Add Benchmark::Stopwatch and timming marks (and unlink ppm)

foursixnine:
This is still related to the task [poo#14976 #1](https://progress.opensuse.org/issues/14976) regarding timing marks and also to the task 3rd (unlinking the ppm)

Whith the last two commits we get [Benchmark::Stopwatch](https://metacpan.org/pod/Benchmark::Stopwatch).

A live run can be seen at: http://deimos.suse.de/tests/1285

### DIFF
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -1,0 +1,273 @@
+use strict;
+use warnings;
+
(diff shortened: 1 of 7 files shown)


coolo:
please squash the stopwatch commits - and make sure to explain why you copied the old class.

PR #670, opened 2016-12-16
Use the stopwatch to detect a slow operations

foursixnine:
Change the baseclass.pm and needle.pm to use Benchmark::Stopwatch
instead of using by calculation by hand.

Fix tabs in basetest.pm

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -380,8 +380,7 @@ sub enqueue_screenshot {
 
     return unless $image;
 
-    my $starttime = gettimeofday;
-    my $watch     = OpenQA::Benchmark::Stopwatch->new();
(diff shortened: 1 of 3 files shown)


backend/baseclass.pm
```diff
@@ -396,11 +395,9 @@ sub enqueue_screenshot {
     $sim = $lastscreenshot->similarity($image) if $lastscreenshot;
     $watch->lap("similarity");
 
-    my $mt1 = gettimeofday;
-
     # we have two different similarity levels - one (slightly higher value, based
     # t/data/user-settings-*) to determine if it's worth it to recheck needles
-    # and one (slightly lower as less signifant) determining if we write the frame
+    # and one (slightly lower as less significant) determining if we write the frame
```
okurz: finally fixed :-)

PR #671, opened 2016-12-16
crop.py: Correct style according to flake8

okurz:
Also make print python3 compatible.


crop.py
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/python
-# Copyright (c) 2013 SUSE Linux Products GmbH
-# Author: Ludwig Nussel
```
coolo: what's the problem with author?

okurz: we don't track authors in other files. We have git for that. I don't mind keeping the author.

PR #673, opened 2016-12-16
Use json ppm

coolo:


backend/baseclass.pm
```diff
@@ -406,16 +407,15 @@ sub enqueue_screenshot {
...
 
     if ($sim > 50) {    # we ignore smaller differences
         $self->write_encoder_frame('R');
         $watch->lap("write_encoder_frame R");
     }
     else {
-        $self->write_img($image, $filename) || die "write $filename";
-        $self->write_encoder_frame("E $filename");
+        #$self->write_img($image, $filename) || die "write $filename";
+        #$self->write_encoder_frame("E $filename");
```
okurz: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

t/11-image-ppm.t
```diff
@@ -0,0 +1,42 @@
...
+BEGIN {
+    unshift @INC, '..';
+}
+
+use OpenQA::Benchmark::Stopwatch;
+
+
+# optional but very useful
+eval 'use Test::More::Color';                 ## no critic
+eval 'use Test::More::Color "foreground"';    ## no critic
```
okurz: sounds nice but I can't see a difference when trying it out locally on another test.

backend/baseclass.pm
coolo: I guess I should have said it in the description: but this *is* my local git repository. I just happened to create a PR from it.

t/11-image-ppm.t
coolo: you mean colours? It's no reason to be ashamed. Happens to many

coolo: it also renders passed as green

PR #677, opened 2016-12-19
Required changes for CASP VMX images

kravciak:
Enable second HDD device in sshVirsh.
Use SERIALDEV variable if set

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -94,8 +94,10 @@ Used for internal initialization, do not call from tests.
 =for stopwords xen hvc0 xvc0 ipmi ttyS
 =cut
 sub init {
-    $serialdev = get_var('SERIALDEV', 'ttyS0');
-    if (get_var('OFW') || check_var('BACKEND', 's390x')) {
+    if (get_var('SERIALDEV')) {
+        $serialdev = get_var('SERIALDEV');
+    }
+    elsif (get_var('OFW') || check_var('BACKEND', 's390x')) {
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -99,11 +99,11 @@ sub init {
...
     }
     elsif (check_var('VIRSH_VMM_FAMILY', 'xen') && check_var('VIRSH_VMM_TYPE', 'linux')) {
-        if (check_var('VERSION', '12-SP2')) {
-            $serialdev = "hvc0";
+        if (check_var('VERSION', '12-SP1')) {
+            $serialdev = "xvc0";
         }
         else {
-            $serialdev = "xvc0";
+            $serialdev = "hvc0";
```
okurz: why is it necessary to turn this around now? Was it never tested in before?

kravciak: I can use "check_var('VERSION', '12-SP2') or is_casp" or turn it around - this way it will work also for SP3 + ...seems cleaner

okurz: Actually I don't quite get why we need to distinguish between SLE versions in the testapi. That is pretty wrong. Maybe you want to consult with @mnowaksuse and come up with an approach where the backend does not need to know anything about SLE. Maybe some `$distri` method can help here, e.g. as is done for `script_run` and such.

coolo: this code really needs to move to main.pm and set SERIALDEV. If that doesn't work, we need to make it work.

PR #678, opened 2016-12-19
Even safer shutdown to get published disks

coolo:
While we try to stop the backend process, we will receive a SIGCHLD.
During handling this we should make sure not to remove sockets, we're
possibly writing to. This happens all the time, but for some reason
more likely on aarch64. This wouldn't be such a big problem if this
wouldn't stop (silently) to stop isotovideo from publishing disks.

See https://progress.opensuse.org/issues/14000

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -144,8 +144,6 @@ sub signalhandler {
 
 sub signalhandler_chld {
 
-    diag("got sigchld");
-
     while ((my $child = waitpid(-1, WNOHANG)) > 0) {
         if ($child == $cpid) {
             diag("commands webserver died");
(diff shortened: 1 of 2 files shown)


isotovideo
```diff
@@ -489,7 +485,9 @@ if (!$r) {
         diag "BACKEND SHUTDOWN $clean_shutdown";
     };
     # don't rely on the backend in a sane state if we failed - just kill it later
-    bmwqemu::stop_vm();
+    eval { bmwqemu::stop_vm(); };
+    # unsafe shutdown means we should not publish
+    $r = 1 if $@;
```
okurz: wouldn't it make sense to also output `$@`?

coolo:
https://openqa.suse.de/tests/679864 is on a hot patched aarch64 worker. As the problem is sporadic, it doesn't prove my fix works - but at least it proves my fix doesn't make it worse :)

isotovideo
coolo: that would only scare you

coolo: I changed that now - but you've been warned

okurz: omg, now I'm *really* scared

PR #681, opened 2016-12-20
More timing marks

foursixnine:
Add more timing marks for check_assert_screen, as this is quite verbose, i decided to add a debug variable (enable by default for now), this screenshot is for the job with debug referenced below

Two sucessfull runs can be seen at [Without debug #1333](http://deimos.suse.de/tests/1333) and [With debug #1323](http://deimos.suse.de/tests/1323)

screenshot from 2016-12-20 19-01-23


coolo:
I guess we can eliminate laps that proved to be < 1% in a thousand runs :)

PR #682, opened 2016-12-21
Include the subprocesses in the coverage report

coolo:


coolo:
no idea why the testapi coverage does not count - the autotest process correctly calls ::report

PR #687, opened 2017-01-05
Allow to start qemu later

nadvornik:
This will be used as a replacement for waiting in main.pm.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -60,7 +60,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   wait_idle wait_screen_change wait_still_screen wait_serial record_soft_failure
   become_root x11_start_program ensure_installed eject_cd power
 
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -1413,6 +1413,22 @@ sub freeze_vm {
     query_isotovideo('backend_freeze_vm');
 }
 
+=head2 cont_vm
```
coolo: as we're dragging this into public API, can we please fix the function name to be english?

nadvornik: continue_vm ? Or do you have any better suggestion?

coolo: what is the opposite of freeze? continue is at least part of http://www.wordhippo.com/what-is/the-opposite-of/freeze.html

But I like melt better :)

But resume_vm has its appeal too

nadvornik:
OK, I renamed it to resume_vm

okurz:
LGTM

PR #689, opened 2017-01-05
Add assert_screen_change

okurz:
Variant of wait_screen_change but dying on no screen change.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -57,7 +57,8 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   upload_asset upload_image data_url assert_shutdown parse_junit_log
   upload_logs
 
-  wait_idle wait_screen_change wait_still_screen wait_serial record_soft_failure
+  wait_idle wait_screen_change assert_screen_change wait_still_screen wait_serial
+  record_soft_failure
   become_root x11_start_program ensure_installed eject_cd power
 
   save_memory_dump save_storage_drives freeze_vm resume_vm
@@ -353,7 +354,7 @@ sub assert_and_dclick {
 
 =head2 wait_screen_change
 
-  wait_screen_change { CODEREF [,$timeout] };
+  wait_screen_change(CODEREF [,$timeout]);
 
 Wrapper around code that is supposed to change the screen.
 This is the opposite to C<wait_still_screen>. Make sure to put the commands to change the screen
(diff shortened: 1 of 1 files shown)


coolo:
I think this CODEREF is a bit too perl specific. I suggest to add an example for perl beginners

okurz:
this comes from `wait_screen_change`. I can reference the documentation block from there, though, or just repeat the example in a short way without duplicating everything

coolo:
Linking will do, yes.

okurz:
updated

testapi.pm
```diff
@@ -396,6 +397,24 @@ sub wait_screen_change(&@) {
     return 0;
 }
 
+=head2 assert_screen_change
+
+  assert_screen_change { CODEREF [,$timeout] };
```
coolo: This is actual incorrect syntax because the CODEREF is the {} - and the optinal timeout is after it. The same error is in the wait_screen_change reference

okurz: Is this better?

```perl
diff --git a/testapi.pm b/testapi.pm
index 25a4157..165a8aa 100755
--- a/testapi.pm
+++ b/testapi.pm
@@ -354,7 +354,7 @@ sub assert_and_dclick {
 
 =head2 wait_screen_change
 
-  wait_screen_change { CODEREF [,$timeout] };
+  wait_screen_change(CODEREF [,$timeout]);
 
 Wrapper around code that is supposed to change the screen.
 This is the opposite to C<wait_still_screen>. Make sure to put the commands to change the screen
@@ -399,7 +399,7 @@ sub wait_screen_change(&@) {
 
 =head2 assert_screen_change
 
-  assert_screen_change { CODEREF [,$timeout] };
+  assert_screen_change(CODEREF [,$timeout]);
 
 Run C<CODEREF> with C<wait_screen_change> but C<die> if screen did not change
 within timeout. Look into C<wait_screen_change> for details
```

coolo: I think so.

okurz: updated

testapi.pm
```diff
@@ -396,6 +397,24 @@ sub wait_screen_change(&@) {
...
+within timeout. Look into C<wait_screen_change> for details.
+
+Example:
+
+  assert_screen_change { send_key 'alt-f4' };
+
+=cut
+
+sub assert_screen_change(&@) {
+    ::wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
```
okurz: I don't know perl, what does `::` do here? I just want to call the other method in the same file and without that it didn't work. Also, deleting the parameter part `(&@)` on both `wait_screen_change` and `assert_screen_change` did not work, ideas?

foursixnine: Perl black magic, `::wait_screen_change` is short for `main::wait_screen_change`
```perl

use strict;
use warnings;

sub foo {
    print "They called me!";
}

::foo;
foo;
```

PR #690, opened 2017-01-11
testapi: Extend documentation for `assert|check_screen`

okurz:
Based on our experiences writing tests and the best practices there are some
recommendations regarding the use of timeout value, when to use
`assert_screen` and when to use `check_screen` and such.

Related progress issue: https://progress.opensuse.org/issues/10380

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -45,8 +45,8 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   assert_and_click mouse_hide mouse_set mouse_click
   mouse_dclick mouse_tclick match_has_tag
 
-  script_run script_sudo script_output validate_script_output
-  assert_script_run assert_script_sudo
+  assert_script_run script_run assert_script_sudo script_sudo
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -690,23 +682,24 @@ sub assert_script_run {
...
+  script_run($program [, $wait]);
 
-Run C<$program> using sudo. Handle the sudo timeout and send password when appropriate.
-C<$wait_seconds> defaults to 2 seconds.
+Run C<$program> (by assuming the console prompt and typing it).
 
-I<The implementation is distribution specific and not always available.>
+The C<$wait> parameter will (unless 0) wait for the script to finish
+by following the script with an echo to serial line and waiting
+for it. Default timeout is 30s.
```
aaannz: `script_run` calls `$distri->script_run` thus I'm for to include
`I<The implementation is distribution specific and not always available.>` here as well.

okurz: That's not really related to that PR but ok, updated

PR #694, opened 2017-01-12
Remove duplication in script_run and assert_script_run

AdamWill:
`assert_script_run` and `script_run` currently duplicate the
work to echo the exit code of the command to the serial port
and act on it. Instead, let's have `assert_script_run` just
take the exit code from `script_run` and die unless it's 0.

Note that this leaves `script_sudo` and `assert_script_sudo`
alone. Ideally I'd like to harmonize those too, but that's more
complex and debatable. I did add a note that `assert_script_sudo`
is unlikely to actually work with the default `script_sudo`.

I cleaned up the docs a bit, and renamed `script_runs` args
to match `assert_script_run`'s. And I made the fail message for
the assert functions a bit more accurate (it can die because the
command timed out as well as because it failed).

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -681,12 +681,16 @@ Deprecated mode
 
   assert_script_run($cmd [, $timeout [, $fail_message]]);
 
-Run C<$cmd> via C<script_run> and C<die> if its exit status is not zero.  The
-exit status is checked by magic string on the serial port with a waiting time
-up do C<$timeout>. Default timeout is 30s.
-Use C<script_run> if C<$cmd> can be expected to fail.
+Run C<$cmd> via C<$distri->script_run> and C<die> unless it returns zero (indicating
+successful completion of C<$cmd>). See C<distri->script_run> for default timeout.
+Use C<script_run> instead if C<$cmd> may fail.
 
 C<$fail_message> is returned in the die message if specified.
+
+I<The C<script_run> implementation is distribution specific and not always available.
+For this to work correctly, it must return 0 if and only if C<$command> completes
+successfully. It must NOT return 0 if C<$command> times out. The default implementation
+should work on *nix operating systems with a configured serial device.>
 
 =cut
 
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -712,7 +714,9 @@ Run C<$command> via C<script_sudo> and then check by C<wait_serial> if its exit
 status is not zero.
 See C<wait_serial> for default timeout.
 
-I<The implementation is distribution specific and not always available.>
+I<The implementation is distribution specific and not always available. Note this
+is quite likely to fail with the default C<script_sudo> implementation, as it attempts
+to echo to the serial device without root permissions.>
```
okurz: IMHO not true in general.

If it would be true, the notice should be on `script_run` as well.

That is depending on how the tests are implemented and how the distribution handles permissions of the serial device, whatever has been provided as a serial device. fedora has the same default permission for /dev/ttyS0 as *suse but that does not mean it can not work. Either the test non-root user is added to the group, permissions are adjusted, owner is changed (how *suse does it: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/consoletest_setup.pm#L73 - yes, not necessarily nice), a special serial device is given reserved for that use.

I suggest:

> Make sure the non-root user has permissions to write to the supplied serial device C<$serialdev>

AdamWill: sure, sounds reasonable. I guess I was assuming you were just only ever using `script_run` as root, like us. :) I like your text, I'll adjust to use it and maybe add the same note in `script_run`.

AdamWill:
there, tweaked the docs a bit.

okurz:
We have a regression -> https://progress.opensuse.org/issues/16372 E.g. https://openqa.opensuse.org/tests/344258 fails because of this. I don't know if it has been foreseen but this PR changed the default timeout of assert_script_run from 90s to 30s

PR #696, opened 2017-01-13
Support IPMI_HW=dell

coolo:
This is not yet used for anything, because iDrac8 talks standard VNC (if configured),
but to differ it in the ipmi backend, I explicitly set it also in the VNC console

Validation: http://147.2.212.248/tests/664#step/login_console/14


coolo:
tests/665 validates it still works with supermicro

PR #697, opened 2017-01-16
Provide `stop_vm` argument for attach_to_running()

mnowaksuse:
Set 'stop_vm' argument should the VM be destroyed at the end of testing.

Verification: http://assam.suse.cz/tests/4689

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -517,14 +517,17 @@ __END"
 }
 
 sub attach_to_running {
-    my ($self, $name) = @_;
+    my ($self, $args) = @_;
 
+    my $name = ref($args) ? $args->{name} : $args;
     $self->name($name) if $name;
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -517,14 +517,17 @@ __END"
 }
 
 sub attach_to_running {
-    my ($self, $name) = @_;
+    my ($self, $args) = @_;
 
+    my $name = $args->{name};
```
coolo: that sounds like you changing API. You can be backward compatible in checking ref($args). If it's undefined $args is $name

mnowaksuse:
Both API notations are now supported.

okurz:
LGTM

okurz:
@mnowaksuse can this be turned into "keep virtual machines around for investigation on demand" based on https://progress.opensuse.org/issues/12836 ?

PR #699, opened 2017-01-18
Re-scale screen on Xen PV to fix assert_and_click

mnowaksuse:
When testing Xen PV SUT under svirt backend openQA believes that SUT has
800x600 px resolution, but it's actually 1024x768. Be it a bug in Xen's
Qemu VNC (e.g. not sending re-scale desktop information) or openQA's VNC
client, scaling it by a factor of 1/(1024/800) makes assert_and_click()
work on Xen PV.

E.g.: `assert_screen` matches needle, computes it's geometrical centre
and clicks to that point. If the computed point is 215x156, openQA
actually clicks to 273x198 (approximately). With this workaround, openQA
places pointer to the correct area (though not exactly 100 %).

Verification run: http://assam.suse.cz/tests/4714

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -345,10 +345,14 @@ sub assert_and_click {
 
     # last_matched_needle has to be set, or the assert is buggy :)
     my $lastarea = $last_matched_needle->{area}->[-1];
-    my $rx       = 1;                                                  # $origx / $img->xres();
-    my $ry       = 1;                                                  # $origy / $img->yres();
-    my $x        = int(($lastarea->{x} + $lastarea->{w} / 2) * $rx);
-    my $y        = int(($lastarea->{y} + $lastarea->{h} / 2) * $ry);
+    my $rx       = 1;                                    # $origx / $img->xres();
+    my $ry       = 1;                                    # $origy / $img->yres();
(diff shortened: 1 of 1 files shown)


coolo:
sorry no. If there is a bug in the VNC communication, it should be fixed - and not adding crude workarounds in the api

okurz:
It might be worthwhile to work around issues in dependencies as a *workaround* while investing some time in fixing the real problem. So both should work

mnowaksuse:
Charles is investigating it: https://bugzilla.suse.com/show_bug.cgi?id=1020616.

sysrich:
I would be happier with bsc#1020616 being fixed than accepting this PR

👎

okurz:
On Tuesday, 24 January 2017 07:52:46 CET Richard Brown wrote:
> I would be happier with bsc#1020616 being fixed than accepting this PR

I guess we all would but the question is more like:
How long are you willing to wait before you agree that this workaround would 
make sense *for the time being*

coolo:
If you have to wait every morning for two school kids to get dressed, you become patient....

mnowaksuse:
The BSC seems to be on track.

mnowaksuse:
The process of upstream kernel fix got derailed. Developer might try to fix it in Xen. But that won't be soon.

mnowaksuse:
Superseded by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/2861.

PR #700, opened 2017-01-18
Remove duplicated @vgaoptions

foursixnine:
This was causing the following message from the qemu backend on xgene2.arch:

RAMBlock "vga.vram" already registered, abort!


dzedro:
Are you sure ? This is only place where are `@vgaoptions` pushed into `@params`
No `@vgaoptions` at all ?

dzedro:
taking back, I was comparing it with my old qemu.pm without https://github.com/foursixnine/os-autoinst/blob/1dd8c6ed929cafe8debbface8fdaada405088537/backend/qemu.pm#L536

PR #702, opened 2017-01-19
More fixes for the DELL VNC implementation on idrac

coolo:
- sleep 4 instead 3 seconds between ipmi commands
- throw out ZLRE encoding as it is not RFB compliant (I also
  threw it out in general due to problems with some qemu implementations,
  reenable it with save_bandwith = 1 in the vnc console setup)
- do not catch generic dies in the vnc code, only wait for the connection.
  Every crash afterwards is a real one. Should make it easier to debug
  also the zkvm update case

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -67,7 +67,8 @@ sub ipmitool {
 # DELL BMCs are touchy
 sub dell_sleep {
     my ($self) = @_;
-    return unless ($bmwqemu::vars{IPMI_HW} || '') eq 'dell', sleep 3;
+    return unless ($bmwqemu::vars{IPMI_HW} || '') eq 'dell';
+    sleep 4;
 }
 
 sub restart_host {
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -7,7 +7,6 @@ use bytes;
 use bmwqemu 'diag';
 use Time::HiRes qw( usleep gettimeofday time );
 use Carp;
-use tinycv;
 use List::Util 'min';
 
 use Crypt::DES;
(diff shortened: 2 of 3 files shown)


consoles/VNC.pm
```diff
@@ -21,7 +20,7 @@ __PACKAGE__->mk_accessors(
     qw(hostname port username password socket name width height depth save_bandwidth
       no_endian_conversion  _pixinfo _colourmap _framebuffer _rfb_version screen_on
       _bpp _true_colour _do_endian_conversion absolute ikvm keymap _last_update_received
-      _last_update_requested _vnc_stalled vncinfo old_ikvm
+      _last_update_requested _vnc_stalled vncinfo old_ikvm dell
```
aaannz: Quickly going though the changes and I don't see anywhere this variable is set. Found only where is checked. Is it intentional?

backend/ipmi.pm
```diff
@@ -67,7 +67,8 @@ sub ipmitool {
 # DELL BMCs are touchy
 sub dell_sleep {
     my ($self) = @_;
-    return unless ($bmwqemu::vars{IPMI_HW} || '') eq 'dell', sleep 3;
+    return unless ($bmwqemu::vars{IPMI_HW} || '') eq 'dell';
+    sleep 4;
```
okurz: why? (the code should tell, not you)

consoles/VNC.pm
```diff
@@ -159,7 +164,7 @@ sub login {
 }
 
 sub _handshake_protocol_version {
-    my $self = shift;
+    my ($self) = @_;
```
okurz: aha! so *you* prefer to also replace the single argument assignment shift's with `@_`…

consoles/VNC.pm
coolo: the joy of the base class - this line in backend/ipmi sets it:
```
$vncopts->{dell} = 1;

backend/ipmi.pm
coolo: you see the comment above the 2 line function?

consoles/VNC.pm
coolo: I never said anything else. I just once told you to the reason why others prefer the shift

okurz:
You are getting me wrong. I just was not sure if you are on the side of @kraih ;-)

okurz:
>you see the comment above the 2 line function?

No, too far away

coolo:
We agree to disagree on the argument unpacking. Just as mojolicious uses invalid english as commit logs :)

PR #703, opened 2017-01-20
Do not use absolute path in needle JSON

aaannz:
- different project directories can be used on worker and webui
leading to errors during upload. Instead of absolute path (with
symlinks expanded) use path relative to project directory and keep
the links unexpanded

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -51,7 +51,13 @@ sub new {
             return;
         }
     }
-    my $self = {tags => ($json->{tags} || [])};
+    my $self = {};
+
+    die "Needle $jsonfile is not under project directory $bmwqemu::vars{PRJDIR}" if (index($jsonfile, $bmwqemu::vars{PRJDIR}));
+    # store json file path relative to $prjdir
+    $self->{file} = substr($jsonfile, length($bmwqemu::vars{PRJDIR}) + 1);
+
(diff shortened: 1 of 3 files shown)


okurz:
I think when changing this part it makes sense to add tests, too.

needle.pm
```diff
@@ -79,7 +81,7 @@ sub new {
         return;
     }
 
-    $self->{file} = $jsonfile;
+    $self->{file} = substr($jsonfile, length($bmwqemu::vars{PRJDIR}) + 1);
```
coolo: this looks very dangerous. I would prefer if you split prefix directory and relative filename somewhere above and die then if the prefix directory doesn't match - and take the relative filename here. The 2 statements are 30 lines apart - which makes them dangerous. One cache refactoring later and we're screwed

aaannz:
Updated, hope I got what you meant. Still need to update cpanfile I think

coolo:
cpanfile? these File: stuff you use are core

aaannz:
But we list them anyway. And I checked, they are already there.

But another thing .. this changes an API, since current openQA expects full path. So I need to change isotovideo version or what is called that variable. Why is it coming to me only now?

coolo:
because we still don't care for that variable - the complaints to have a check only come once its too late :)

PR #704, opened 2017-01-22
[WIP][DISCUSSION] -- wait_screen_change option on send_key; deprecating do_wait

okurz:


testapi.pm
```diff
@@ -1137,17 +1153,11 @@ sub mouse_tclick(;$$) {
...
-    sleep $time;
-    query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
-    sleep $time;
-    query_isotovideo('backend_mouse_button', {button => $button, bstate => 0});
-    sleep $time;
-    query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
-    sleep $time;
-    query_isotovideo('backend_mouse_button', {button => $button, bstate => 0});
+    for (1..3) {
+        query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
```
coolo: this seems unrelated - and is wrong as it removes one sleep between up and down

testapi.pm
```diff
@@ -939,11 +945,25 @@ Special characters naming:
...
+        %args = (do_wait => $_[0]);
+    }
+    else {
+        %args = @_;
+    }
+    $args{do_wait} //= 0;
+    $args{wait_screen_change} //= 0;
     bmwqemu::log_call(key => $key);
+    if ($args{wait_screen_change}) {
+        wait_screen_change { query_isotovideo('backend_send_key', {key => $key}) };
```
coolo: why not pass the argument value as 2nd argument?

testapi.pm
```diff
@@ -939,11 +945,25 @@ Special characters naming:
 =cut
 
 sub send_key {
-    my ($key, $do_wait) = @_;
-    $do_wait //= 0;
+    my ($key) = shift;
```
aaannz: You don't need `()` when using shift.

okurz:
when I added '[DISCUSSION]' I did not meant: nit pick my perl code but tell me if you like the idea in general or not before I cleanup the perl code based on your suggestions

kravciak:
There are multiple events that can trigger wait_screen_change:
- window animation while still loading changes
- blinking cursor
- system time change

I think checking for "any" screen change is not good enough after send_key.
Using send_key & wait_screen_change combination will also introduce new issues. (current example poo#16216)

okurz:
@kravciak while you are right I don't really see this as related. It is not the testapi's fault that someone is misusing it. In the event examples you mentioned I see two options:
* implicit `wait_still_screen`
* explicit wait for expected result, e.g. `assert_screen` or `wait_serial`

okurz:
so I did not get any comments that tell me it's a bad idea. I hope that means I can invest more time, streamline it and a proper PR will be accepted then.

okurz:
@asmorodskyi

backend/baseclass.pm
```diff
@@ -990,7 +990,7 @@ sub new_ssh_connection {
             last;
         }
         else {
-            bmwqemu::diag "Could not connect to $args{username}\@$args{hostname}, Retry";
+            bmwqemu::diag "Could not connect to $args{username}\@$args{hostname}, Retrying after 10 seconds...";
```
asmorodskyi: Magic number , need to move to variable and put same variable in line below

coolo:
please reopen if you found the 'more time'

testapi.pm
okurz: uhm, can you be more specific?

PR #706, opened 2017-01-25
Refactor and use abs path for png lookup

aaannz:
- move json parsing only after proper path is evaluated
- use absolute path for png lookup

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -39,9 +39,26 @@ sub new {
     my $json;
     if (ref $jsonfile eq 'HASH') {
         $json = $jsonfile;
-        $jsonfile = $json->{file} || join('/', $needledir, $json->{name} . '.json');
+        $jsonfile = $json->{file} || File::Spec->catfile($needledir, $json->{name} . '.json');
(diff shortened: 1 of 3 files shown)


needle.pm
```diff
@@ -39,31 +39,36 @@ sub new {
...
     elsif (-f File::Spec->catfile($bmwqemu::vars{PRJDIR}, $jsonfile)) {
         # json file path already relative
         $self->{file} = $jsonfile;
+        $jsonfile = File::Spec->catfile($bmwqemu::vars{PRJDIR}, $jsonfile);
     }
     else {
         die "Needle $jsonfile is not under project directory $bmwqemu::vars{PRJDIR}";
     }
 
+    # $json->{file} contains parh relative to $bmwqemu::vars{PRJDIR}
```
coolo: typo

dzedro:
tested, [works](http://10.100.12.155/tests/4829)
but there is still issue with webui, I don't see any needles in the Candidate needle: roll-down menu

PR #707, opened 2017-01-26
Add optional 'no_wait' parameter for check/assert_screen

okurz:
The default interval of checking screens every 1 second is an unsuitable
choices in multiple occasions. When screens change without external
interaction within a certain time, for example a boot menue, it is safer to
check more than once per second. In other occasions the screen repeatedly
alternates in an even interval and the screen check is conducted at the same
interval but shifted in time so that always the "wrong" screen is catched. In
this case a lower timeout which can also be an odd value should be selected.

In the cases of os-autoinst-distri-opensuse the openSUSE Krypton live CD is
built with SUSE kiwi v>=8 which yields a boot screen timeout of 1-2 seconds
Checking only once per second is just not often enough to be able to check the
screen and conduct actions immediately afterwards, e.g. press a key to abort
the timeout.

The second example is checking a text entry field in oomath where the needle
covers the blinking text cursor to be sure the focus is on that field. The
test fails here repeatedly but obviously not every time because the needle
checking interval matches the blinking cursor interval but is of by a fraction
of a second so that only the instances of the cursor not being shown is
matched.

Related progress issue: https://progress.opensuse.org/issues/16038

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -301,6 +301,9 @@ my $current_test_name;
 # timeout for the select (only set for check_screens)
 my $timeout = undef;
 
+# do not wait for timeout if set
+my $no_wait = undef;
+
 # marks a running check_screen
 our $tags = undef;
 
@@ -326,11 +329,17 @@ sub _calc_check_delta {
 }
 
 sub check_asserted_screen {
-    my ($force_timeout) = @_;
+    my ($force_timeout, $no_wait) = @_;
 
-    my $delta = _calc_check_delta;
-    # come back later, avoid too often called function
-    return if $timeout > 0.05;
+    if ($no_wait) {
+        # prevent CPU overload by waiting at least a little bit
+        $timeout = 0.1;
+    }
+    else {
+        _calc_check_delta;
+        # come back later, avoid too often called function
+        return if $timeout > 0.05;
+    }
     ($last_check_seconds, $last_check_microseconds) = gettimeofday;
     my $rsp = $bmwqemu::backend->_send_json({cmd => 'check_asserted_screen'}) || {};
     # the test needs that information
@@ -354,7 +363,7 @@ sub check_asserted_screen {
         $timeout = undef;
     }
     else {
-        _calc_check_delta;
+        _calc_check_delta unless $no_wait;
     }
 }
 
@@ -399,16 +408,16 @@ while ($loop) {
             next;
         }
         if ($rsp->{cmd} eq 'check_screen') {
-            my $mustmatch = $rsp->{mustmatch};
-            my $timeout   = $rsp->{timeout};
-            my $check     = $rsp->{check};
+            my $mustmatch     = $rsp->{mustmatch};
+            my $check_timeout = $rsp->{timeout};
+            $no_wait = $rsp->{no_wait} // 0;
 
             $tags = $bmwqemu::backend->_send_json(
                 {
                     cmd       => 'set_tags_to_assert',
                     arguments => {
                         mustmatch => $mustmatch,
-                        timeout   => $timeout
+                        timeout   => $check_timeout
                     }})->{tags};
             next;
         }
@@ -436,7 +445,7 @@ while ($loop) {
                 $bmwqemu::backend->_send_json({cmd => 'retry_assert_screen'});
                 $needinput = 0;
                 $timeout   = .1;
-                check_asserted_screen(1);
+                check_asserted_screen(1, $no_wait);
             }
             myjsonrpc::send_json($r, {interactive => $interactive});
             next;
@@ -455,14 +464,14 @@ while ($loop) {
             $bmwqemu::backend->_send_json({cmd => 'retry_assert_screen', arguments => {reload_needles => $reload}});
             # that's enough for the webui to know
             myjsonrpc::send_json($r, {ret => 0});
-            check_asserted_screen(1);
+            check_asserted_screen(1, $no_wait);
             next;
         }
         die "Unknown command $rsp->{cmd}";
     }
 
     if (defined $tags && !$needinput) {
-        check_asserted_screen;
+        check_asserted_screen(0, $no_wait);
     }
 }
 
--- a/testapi.pm
+++ b/testapi.pm
@@ -246,19 +246,19 @@ sub _check_backend_response {
 }
 
 sub _check_or_assert {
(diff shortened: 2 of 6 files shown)


coolo:
I wonder if an explicit interval makes sense. In the cases you describe we basically want to check as often as we can instead of waiting for a sleep.

testapi.pm
```diff
@@ -244,19 +244,19 @@ sub _check_backend_response {
 }
 
 sub _check_or_assert {
-    my ($mustmatch, $timeout, $check) = @_;
+    my ($mustmatch, $timeout, $check, %args) = @_;
```
coolo: having 3 arguments and then %args doesn't appear to be a good choice

isotovideo
```diff
@@ -296,6 +296,9 @@ $testfd->write("GO\n");
 my $interactive = 0;
 my $needinput   = 0;
 
+# screen checking interval in s (float)
+my $interval = 1.;
```
coolo: I don't like this global variable. This is internal to one function and shouldn't be global

testapi.pm
```diff
@@ -244,19 +244,19 @@ sub _check_backend_response {
...
+    my $rsp = query_isotovideo('check_screen', {mustmatch => $mustmatch, timeout => $timeout, check => $check, interval => $args{interval}});
     # seperate function because it needs to call itself
-    return _check_backend_response($rsp, $check, $timeout, $mustmatch);
+    return _check_backend_response($rsp, $check, $timeout, $mustmatch, $args{interval});
 }
 
 =head2 assert_screen
 
-  assert_screen($mustmatch [,$timeout]);
+  assert_screen($mustmatch [,$timeout] [, interval => $interval]);
```
coolo: if you extend the API, then please also support 'mustmatch', timeout => 30, interval => 1

I don't think we need to deprecate the simple form - as it's just a very frequent call

coolo:
So I opt to go for a boolean setting the interval to a minimum. 0.1 second to avoid complete CPU overload. But in practise assert_screen will take longer most of the time anyway

testapi.pm
okurz: why?

isotovideo
okurz: can't get it to work. somehow if I try to forward variables where used it's not picked up for the actual check or something. Is the global variable visible by some call that goes out and comes back over RPC which does not work when forwarding the variables?

okurz:
> I wonder if an explicit interval makes sense. In the cases you describe we basically want to check as often as we can instead of waiting for a sleep.

as I didn't see an answer by you on IRC: Do you rather prefer a boolean switch and just skip the sleep after all?

testapi.pm
okurz: done

testapi.pm
okurz: changed

okurz:
updated:
* more minor cleanup changes
* use boolean switch "no_wait" instead of interval value
* show float numbers on "no match <nr>/no change <nr>"
* extend check/assert_screen with both named as well as positional arguments

okurz:
I would work on increasing the coverage again but as @aaannz wants to replace pitux I think we should do that later not to make more conflicts on the full stack test code

okurz:
can someone help me understand https://travis-ci.org/os-autoinst/os-autoinst/builds/195589428#L448 ?

okurz:
updated with some minor documentation fixes to force retest

okurz:
Log output from t/99-full-stack.t:

```
10:40:04.2164 26083 <<< testapi::assert_screen(mustmatch='core', timeout=15, no_wait=1)
10:40:04.2460 26088 MATCH(boot-core:0.00)
10:40:04.2461 26088 no match: 15.0s
10:40:04.3469 26088 no change: 14.9s
10:40:04.4474 26088 no change: 14.8s
10:40:04.5480 26088 no change: 14.7s
10:40:04.6487 26088 no change: 14.6s
10:40:04.7689 26088 MATCH(boot-core:0.00)
10:40:04.7691 26088 no match: 14.5s
…
10:40:09.8179 26088 MATCH(boot-core:0.97)
10:40:09.9117 26083 >>> testapi::_check_backend_response: found boot-core, similarity 0.97 @ 1/74
…
10:40:11.1998 26083 <<< testapi::assert_screen(mustmatch='on_prompt', timeout=30)
10:40:11.9278 26088 MATCH(boot-on_prompt:0.00)
10:40:11.9279 26088 no match: 29.3s
10:40:12.9176 26088 MATCH(boot-on_prompt:0.00)
10:40:12.9178 26088 no match: 28.3s
…
10:40:26.9296 26088 MATCH(boot-on_prompt:1.00)
10:40:26.9986 26083 >>> testapi::_check_backend_response: found boot-on_prompt, similarity 1.00 @ 2/64
```

testapi.pm
```diff
@@ -244,43 +246,50 @@ sub _check_backend_response {
...
+    $timeout = shift if (@_ % 2);
+    my %args = (@_, timeout => $timeout // $bmwqemu::default_timeout);
+    bmwqemu::log_call(mustmatch => $mustmatch, %args);
+    return _check_or_assert($mustmatch, 0, %args);
 }
 
 =head2 check_screen
 
-  check_screen($mustmatch [,$timeout]);
+  check_screen($mustmatch [, [$timeout] | [timeout => $timeout]] [, no_wait => $no_wait]);
```
coolo: I don't think this code supports check_screen 'hallo', timeout => 1

My reading of the code implies it will use default_timeout in these cases - and you didn't cover this code either

okurz:
* changed order of precedence to "default < $timeout < timeout => $timeout"

```perl
-    my %args = (@_, timeout => $timeout // $bmwqemu::default_timeout);
+    my %args = (timeout => $timeout // $bmwqemu::default_timeout, @_);
```

I tested manually with `t/99-full-stack.t` and failed to update `t/03-testapi.t` with proper tests for `check_screen` and `assert_screen` which I also tried some times in before but always failed to complete.

okurz:
*bump*

coolo:
whatever 'bump' on a sunday afternoon means - bumped on your keyboard?

okurz:
Means whenever a reviewer sees this in his inbox during his normal working hours. If I wanted you desperately to merge on Sunday I should have called you, right? ;-)

coolo:
Nah, a note in IRC will do :)

PR #708, opened 2017-01-26
Poweroff the VM after tests

aaannz:
- apart from covering assert_shutdown, this is needed for image upload tests within openQA


coolo:
we need to revert it - it just doesn't shutdown and fails the test, but waits a minute. pitux just doesn't seem to support poweroff

PR #709, opened 2017-01-26
Replace pitux with tinycore as test image

aaannz:
- pitux image is unable to pass shutdown test and
was rather limited. TinyCore is bigger (11MB vs 2MB)
but can be used for far more testing and passes
shutdown test


okurz:
Please state in your commit and PR *why* not *what*

The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

okurz:
@asdil12

asdil12:
@aaannz 
Is there a special reason for that?
The tinycore image is ~7 times as large as the existing one.
That is why I used pitux in the first place: To avoid large files in git.

aaannz:
updated and there is also discussion on irc

okurz:
👎 

… updating my PR with conflict handling

aaannz:
I don't follow, what is the problem?

okurz:
* image is bigger
* test takes longer
* `git grep pitux`
* I need to update my test (that I can accept)

aaannz:
True, it tests shutdown, which also takes some time. But noted.

okurz:
ok, didn't consider this. Then that is ok for a full stack test. I take that point back, sorry :-)

PR #710, opened 2017-01-31
serial terminal: Use newline to clear buffer instead of EOT

richiejp:
Send a newline character which causes the terminal to flush its input buffer
and cat to read the incoming data. Once the terminal's input buffer is empty,
we send EOT (End Of Transmission) which causes the EOF condition in
cat (i.e. read() returns zero).

Previously EOT was used twice, once to flush the buffer and then a second time
to exit. If the buffer was already empty this would result in cat exiting
after the first EOT then the second would cause the shell to log out. Unlike
EOT newline can not cause the EOF condition because it is written to the buffer.

https://en.wikipedia.org/wiki/End-of-Transmission_character


okurz:
LGTM

PR #711, opened 2017-01-31
Give back default timeout of 90s to assert_script_run

okurz:
Fixes a regression introduced by b674942.

As Adam said: "it was historically the default for long enough we think we
should keep it that way" :-)

Related progress issue: https://progress.opensuse.org/issues/16372


AdamWill:
👍 LGTM. sorry for breaking this!

AdamWill:
in fact, I think you can now safely remove the three lines following the `log_call` too (they're no longer necessary now we go ahead and set the timeout to 90 if it's 0 or undefined).

okurz:
updated: deleted the "undef if zero" part as the default timeout in the same level serves the same purpose.

PR #712, opened 2017-02-02
svirt: Do not use is_shutdown()

mnowaksuse:
If we gather information about VM's state under svirt via
`query_isotovideo`, it keeps executing `read_json` and no further
command is executed.

The root cause may be that something touches closed VNC connection
(which is closed on gues OS poweroff on libvirt), which ends up in
stall.

Fails on: https://openqa.suse.de/tests/754073
Verification run: http://assam.suse.cz/tests/5014

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1359,22 +1359,30 @@ sub assert_shutdown {
     $timeout //= 60;
     bmwqemu::log_call(timeout => $timeout);
     $timeout = bmwqemu::scale_timeout($timeout);
-    while ($timeout >= 0) {
-        my $is_shutdown = query_isotovideo('backend_is_shutdown') || 0;
-        if ($is_shutdown < 0) {
-            bmwqemu::diag("Backend does not implement is_shutdown - just sleeping");
-            sleep($timeout);
-        }
-        # -1 counts too
-        if ($is_shutdown) {
-            $autotest::current_test->take_screenshot('ok');
-            return;
-        }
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -1359,22 +1359,30 @@ sub assert_shutdown {
...
-        my $is_shutdown = query_isotovideo('backend_is_shutdown') || 0;
-        if ($is_shutdown < 0) {
-            bmwqemu::diag("Backend does not implement is_shutdown - just sleeping");
-            sleep($timeout);
-        }
-        # -1 counts too
-        if ($is_shutdown) {
-            $autotest::current_test->take_screenshot('ok');
-            return;
+    if (check_var('BACKEND', 'svirt')) {
```
coolo: Sorry, I see your need to find a solution - but I won't accept these crude hacks.

okurz: @mnowaksuse looks like you are destroying the object oriented approach here. Try to avoid if/else by using polymorphy

coolo: @okurz - your generic bla blah doesn't help here. The polymorph backends only exist in the backend while the testapi only has a driver.

okurz:
> The polymorph backends only exist in the backend while the testapi only has 
a driver.

so change that?

testapi.pm
sysrich: Please find a solution that doesn't put such a crude hack into the testapi

PR #713, opened 2017-02-03
Only abort on system stalls detected when explicitly asked for

okurz:
This changes the previous behaviour of always aborting the test in case of a
system stall to only abort when explicitly selected.

A system stall is bad but we can not fix it in the scope of os-autoinst
because it is an infrastructure issue. Also, not every check_screen or
assert_screen with a stall within is necessary for us to die on. Same as a
manual tester that goes for a cup of coffee. If he misses a dialog on the
screen with a timeout, this would be bad, but in most cases when he returns to
check the screen either the SUT is still busy doing something or already
waiting at the state where the user would need to give new inputs and the
tester actually checks the screen against his reference.

Explicit checking for stalls can be selected on the calls of
`check|assert_screen` as well as for the whole test execution with the
`vars.json` variable `DETECT_STALL`.

Manually verified with the help of `t/99-full-stack.t` checking for correct
settings of the flag based on the testapi calls.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2016 SUSE LLC
+# Copyright © 2012-2017 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -852,7 +852,7 @@ sub check_asserted_screen {
         # make sure we recheck later
         $self->assert_screen_last_check(undef);
 
-        if ($self->stall_detected) {
+        if (($args->{abort_on_stall} || $bmwqemu::vars{ABORT_ON_STALL}) && $self->stall_detected) {
             backend::baseclass::write_crash_file();
             bmwqemu::mydie "assert_screen fails, but we detected a timeout in the process, so we abort";
         }
--- a/isotovideo
+++ b/isotovideo
@@ -304,6 +304,9 @@ my $timeout = undef;
 # do not wait for timeout if set
 my $no_wait = undef;
 
+# detect stalls of the system where we are blind during the screen checks
+my $abort_on_stall = undef;
+
 # marks a running check_screen
 our $tags = undef;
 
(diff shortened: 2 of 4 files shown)


okurz:
Includes #707.

@coolo, @foursixnine Something for consideration and discussion regarding the handling of stalls

coolo:
please rebase - so I can actually see the forrest. But I guess I dislike this.

coolo:
we already fail only on failed asserts. And just because the developer didn't think about adding yet another parameter doesn't mean the stall is less severe

okurz:
updated:
* rebased
* renamed `detect_stall` to `abort_on_stall`

> And just because the developer didn't think about adding yet another parameter doesn't mean the stall is less severe

it *is* less severe to the test reviewer. the I/O stalls are severe in general and therefore reported with a WARNING but they are in the domain of the infrastructure admins.

okurz:
> we already fail only on failed asserts

After you repeated this to me verbally I understood, my feature does not make sense. We already failed the assert, the stall detection is not changing that, it is only explaining the reason.

AdamWill:
Well. I don't think that's the whole story.

This can cause a test to fail when it otherwise would not, which in my book is a problem. The problem is that the `bmwqemu::mydie "assert_screen fails, but we detected a timeout in the process, so we abort";` line is in `check_asserted_screen`, which isn't only used for `assert_screen`. It's also used for `check_screen`.

In the Fedora tests - and I'm guessing in the SUSE tests too - we have various cases where we run `check_screen` and actually quite often expect it to fail; all we're doing is checking whether we're in a particular situation or not, and if we aren't, that's totally fine and the test expects to continue.

But because this `mydie` statement is there, if there happens to be a stall during one of these `check_screen` calls, the test will die - even though there isn't actually any failure yet. If that line wasn't there, the test would happily continue and work.

So I don't think it's actually true to say that 'we already fail only on failed asserts`, really, and to me, that's a problem. I would say that the test should only die if the stall occurred during an `assert_screen`, not during a `check_screen`, but that might be a bit hard to implement - I'm looking at it now.

AdamWill:
Filed https://github.com/os-autoinst/os-autoinst/pull/735 as kind of a suggestion for this. WDYT?

PR #715, opened 2017-02-07
Check unregistered needles when failing assert_screen

coolo:
https://progress.opensuse.org/issue/820

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -28,6 +28,7 @@ use OpenQA::Benchmark::Stopwatch;
 
 our %needles;
 our %tags;
+our %unregistered;
 our $needledir;
 our $cleanuphandler;
 
@@ -133,10 +134,11 @@ sub save {
 }
 
 sub unregister {
-    my ($self) = @_;
+    my ($self, $reason) = @_;
     for my $g (@{$self->{tags}}) {
         @{$tags{$g}} = grep { $_ != $self } @{$tags{$g}};
         delete $tags{$g} unless (@{$tags{$g}});
+        $unregistered{$g}->{$self} = $reason;
     }
 }
 
@@ -200,14 +202,8 @@ sub has_property {
 sub TO_JSON {
     my ($self) = @_;
 
-    my $hash = {
-        tags       => $self->{tags},
-        properties => $self->{properties},
-        area       => $self->{area},
-        file       => $self->{file},
-        png        => $self->{png},
-        name       => $self->{name}};
-    return $hash;
+    my %hash = map { $_ => $self->{$_} } qw(tags properties area file png name);
+    return \%hash;
 }
 
 sub wanted_ {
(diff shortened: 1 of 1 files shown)


needle.pm
```diff
@@ -203,13 +200,10 @@ sub has_property {
...
-        tags       => $self->{tags},
-        properties => $self->{properties},
-        area       => $self->{area},
-        file       => $self->{file},
-        png        => $self->{png},
-        name       => $self->{name}};
+    my $hash;
+    for my $key (qw(tags properties area file png unregistered name)) {
+        $hash->{$key} = $self->{$key};
+    }
```
okurz: @kraih would probably recommend

```perl
my $hash = map { $_ => $self->{$_} } qw(tags properties area file png unregistered name));
```
or something along these lines

needle.pm
```diff
@@ -239,19 +233,20 @@ sub init {
...
-    if (!$goods || !@tags) {
+    if (!$goods || !@wanted) {
         return $goods || [];
     }
     my @results;
 
     # now check that it contains all the other tags too
   NEEDLE: for my $n (@$goods) {
-        for my $t (@tags) {
+        for my $t (@wanted) {
```
okurz: thank you for replacing the ambiguous `@tags` within the method `tags`. I got confused by it already :-)

backend/baseclass.pm
```diff
@@ -856,6 +856,13 @@ sub check_asserted_screen {
             backend::baseclass::write_crash_file();
             bmwqemu::mydie "assert_screen fails, but we detected a timeout in the process, so we abort";
         }
+        elsif (!$self->assert_screen_check) {
+            my @unregistered_needles = grep { $_->{unregistered} } @{$self->assert_screen_needles};
+            my ($foundneedle, $candidates) = $img->search(\@unregistered_needles, 0, 1, undef);
+            # the best here is still a failure, as unregistered
+            push(@$failed_candidates, $foundneedle) if $foundneedle;
+            push(@$failed_candidates, @$candidates);
```
okurz: well, that's basically what I wanted to say today in the call. Just add some `diag` and we already have something useful :-)

needle.pm
kraih: Yes, i would. 😃 (`my %hash = map {...} qw(...)`)

coolo: actually it burned in my fingers :)

coolo:
It's poo#820 - there is just a s missing in the expanded url

needle.pm
AdamWill: Yes, now there's only the `tags()` method and the `%tags` hash ;)

coolo:
this seems to miss a commit or two. At least unregistered is unused

coolo:
but what is there is harmless so merging this

PR #716, opened 2017-02-07
testapi: Add generic 'record_info' function similar to 'record_soft_failure'

okurz:
This generic 'record_info' function can be used to record a step result
without influencing the actual result and having more flexibility
regarding the content and what should be displayed.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -368,6 +368,27 @@ sub next_resultname {
     }
 }
 
+=head2 record_resultfile
+
+    $self->record_resultfile($title, $output [, result => $result] [, resultname => $name]);
+
+Record result file to be parsed when evaluating test results, for example
+within the openQA web interface.
+=cut
+sub record_resultfile {
+    my ($self, $title, $output, %nargs) = @_;
+    my $filename = $self->next_resultname('txt', $nargs{resultname});
+    my $detail = {
+        title  => $title,
+        result => $nargs{result},
+        text   => $filename,
+    };
+    push @{$self->{details}}, $detail;
+    open my $fh, '>', bmwqemu::result_dir() . "/$filename";
+    print $fh $output;
+    close $fh;
+}
+
 sub record_serialresult {
     my ($self, $ref, $res, $string) = @_;
 
@@ -378,19 +399,10 @@ sub record_serialresult {
         # the screenshot is not the fail, it's just for documentation
         $self->_result_add_screenshot($result);
     }
-
-    my $details = {result => $res};
-
-    my $text_fn = $self->next_resultname('txt');
-    open my $fd, ">", bmwqemu::result_dir() . "/$text_fn";
-    print $fd "# wait_serial expected: $ref\n\n";
-    print $fd "# Result:\n";
-    print $fd "$string\n";
-    close $fd;
-    $details->{text}  = $text_fn;
-    $details->{title} = 'wait_serial';
-    push @{$self->{details}}, $details;
-
+    my $output = "# wait_serial expected: $ref\n\n";
+    $output .= "# Result:\n";
+    $output .= "$string\n";
+    $self->record_resultfile('wait_serial', $output, result => $res);
     return $result;
 }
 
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -177,7 +177,31 @@ sub record_soft_failure {
...
+openQA can show a info box as part of the job results details. Use this instead
+of C<record_soft_failure> for example when you do not want to mark the job as a
+softfail. The optional value C<$result> can be 'ok' (default), 'fail',
+'softfail'.
+
+=cut
+
+sub record_step {
+    my ($title, $output, $result) = @_;
+    $result //= 'ok';
```
richiejp: I would prefer this to be named argument in case we want to extend the API in the future which is quite likely in this case (i.e. for meta data).

So do something like:
```
my ($title, $output) = @_[0 .. 1];
my %nargs = @_[2 .. $#_];
$nargs{result} ||= 'ok'
```

basetest.pm
```diff
@@ -375,6 +375,20 @@ sub next_resultname {
     }
 }
 
+sub record_resultfile {
+    my ($self, $title, $output, $result) = @_;
+    my $filename = $self->next_resultname('txt');
```
richiejp: We use both of `next_resultname`'s arguments, so you probably should add an optional named argument for that.

richiejp:
This could remove quite a few lines from the LTP runner so I fully approve. Could also lead to displaying the results in a better way for tests not focused around needles.

testapi.pm
okurz: why `||= 'ok'` and not `//= 'ok'? to also check correct content?

richiejp: You are right, `//=` is better. Otherwise `0` and other falsy values are coerced to 'ok' which could lead to unexpected behaviour.

testapi.pm
```diff
@@ -177,7 +177,31 @@ sub record_soft_failure {
...
+    die "unsupported \$result $result" unless $result =~ /^(ok|fail|softfail)$/;
```
aaannz: Given the recent discussion about possible new result state, I'm for creating helper like is_valid_result.

aaannz:
I like this. One thing is not clear to me. How do you tell webui to ignore results from record_step?

sysrich:
I think we need to have a generic solution for hiding/folding away results that people don't care

I do not like the name `record_step` and think `record_info` is more accurate, as discussed in person

testapi.pm
okurz: done

basetest.pm
okurz: done

testapi.pm
okurz: done

okurz:
> How do you tell webui to ignore results from record_step?

For now you don't. As stated by @sysrich and also discussed with him. It is a generally valid wish but not really applicable here.

People already implemented their own way for this "feature" within os-autoinst-distri-opensuse. I just wanted to have it more streamlined, generic and "upstream". `record_info` is to `wait_serial`, `script_run`,`assert_script_run` what `save_screenshot` is to `check_screen`,`assert_screen` :-)

basetest.pm
```diff
@@ -367,6 +368,27 @@ sub next_resultname {
...
+=head2 record_resultfile
+
+    $self->record_resultfile($title, $output [, result => $result] [, resultname => $name]);
+
+Record result file to be parsed when evaluating test results, for example
+within the openQA web interface.
+=cut
+sub record_resultfile {
+    my ($self, $title, $output, %nargs) = @_;
+    my $filename = $self->next_resultname('txt', $nargs{resultname});
```
richiejp: This doesn't produce a warning when `$nargs{resultname}` is undefined?

okurz: AFAIK just forwarding an undef value does not produce a warning as the value is not actually "used" here. see https://github.com/os-autoinst/os-autoinst/pull/720 for my upcoming changes about preventing warnings to slip through within the execution of tests. btw, I think you can help with https://github.com/os-autoinst/os-autoinst/pull/720/commits/278c46cad82692bbee45fc0396d5cba1206a3367#diff-200ab83e6b455d66c0ea0a727f5f6dd1 :-)

okurz:
updated:
* rebased: much simpler diff now

basetest.pm
richiejp: Yes, I suppose evaluating to undef is not an error, but using it with an operator other than `defined` is. I am not sure if my unit tests will work 100% correctly with `Test::Warnings`, so you can assign me a progress ticket to look into it, assuming that is what you are talking about?

okurz: yes, it's https://progress.opensuse.org/issues/10704

I am convinced we should be able to prevent *all* warnings in the tests we have.

PR #717, opened 2017-02-07
Delete obsolete automatic denting on needle name

okurz:
bnc# is obsolete and we have workaround property which should be used for
this.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -137,14 +137,6 @@ sub record_screenmatch {
         bmwqemu::diag("needle '$h->{name}' is a workaround");
     }
 
-    # Hack to make it obvious that some test passed by applying a hack
-    # (such as clicking away some error popup). Those hacks are indicated by a
-    # needle containing "bnc" in its name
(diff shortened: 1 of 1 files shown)


coolo:
did you check all bnc* have a workaround property?

okurz:
> did you check all bnc* have a workaround property?

no

sysrich:
> no

then I think this PR needs to be reconsidered

-1

okurz:
IMHO you both wasted time to write your comments when it would be faster to check the existance of any needle in openSUSE / SLE repo with name "*bnc*". I checked, there are *none*, so the PR is safe for use. And I am sure fedora does not care about needles with "bnc" in name.

coolo:
see? now you checked what I asked you for. So no time wasted.

okurz:
On Thursday, 9 February 2017 07:07:56 CET Stephan Kulow wrote:
> see? now you checked what I asked you for. So no time wasted.

The initial answer "no" was less work then checking. The statement hidden 
inside there was "… and we should not care for various reasons, bnc# being 
outdated is just one"

PR #718, opened 2017-02-08
testapi: Simplify save_memory_dump call

richiejp:
Use idiomatic Perl named arguments. That is, interpret the argument list as a
map, so that named arguments can be passed without brackets. Also record the
arguments in the log and correct the documentation.

AFAIK the only user of this function with named arguments is the LTP runner.


foursixnine:
LGTM

okurz:
LGTM

richiejp:
Thanks for review

PR #720, opened 2017-02-09
Use Test::Warnings where possible to prevent warnings in tests

okurz:


okurz:
@aaannz one warning that is catched by the above tests is
https://github.com/os-autoinst/os-autoinst/blob/master/needle.pm#L45
when `PRJDIR` is not defined. How should that code behave?

aaannz:
Ah right! Didn't look into the https://progress.opensuse.org/issues/16362 yet. But will do in a way that PRJDIR is mandatory and CASEDIR, PRODUCTDIR etc., if not set, should be derived from PRJDIR. Can be?

aaannz:
Hmm.. hold that thought. PRJDIR doesn't contain distri etc. so no.. better to infer PRJDIR from CASEDIR.

okurz:
good. This PR is not urgent to be merged but I think it is helpful to use to fix the other issues.

cpanfile
```diff
@@ -52,10 +52,12 @@ on 'test' => sub {
   requires 'Perl::Critic';
   requires 'Test::Output';
   requires 'Test::Fatal';
+  requires 'Test::Warnings';
   requires 'Test::Pod';
   requires 'Test::MockModule';
   requires 'Pod::Coverage';
   requires 'Devel::Cover';
+  requires 'String::Random';
```
coolo: this looks like a merge victim

okurz:
updated

PR #723, opened 2017-02-14
Allow to set qemu -smbios via variable

nadvornik:


okurz:
LGTM

doc/backend_vars.asciidoc
```diff
@@ -71,6 +71,7 @@ QEMU_VIRTIO_RNG;boolean;0;Enable virtio random number generator
 QEMU_NUMA;boolean;0;Enable NUMA simulation, requires QEMUCPUS to be greater than one
 RAIDLEVEL;;;
 SKIPTO;full name of test module;;Restore VM from snapshot and continue by running specified test module. Needs HDD image with snapshots present
+SMBIOS;see qemu -smbios ?;undef;pass this value to qemu -smbios
```
coolo: How likely is it that we will use this option for anything else than drafting qemu command lines? If slim -> please prefix with QEMU_

nadvornik:
added QEMU_ prefix

PR #724, opened 2017-02-16
virtio terminal: Reduce CPU time using select

richiejp:
Block by calling select on the socket instead of looping over non-blocking
read. Unlike read, select takes a timeout. Also remove the idle callback,
which is not used and would complicate setting the timeout for select.

This reduces the user CPU time and the number of context switches.

https://progress.opensuse.org/issues/17092


nadvornik:
works fine for me

PR #725, opened 2017-02-24
Improve isotovideo command line handling

okurz:
Related progress issue: https://progress.opensuse.org/issues/16538

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -16,10 +16,27 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 #
 
-use strict;
+=head1 SYNOPSIS
 
-local $Devel::Trace::TRACE;
-$Devel::Trace::TRACE = 0;
+isotovideo [OPTIONS]
+
+Parses vars.json and tests the given assets/ISOs.
+
+=head1 OPTIONS
+
+=over 4
+
+=item B<-d, --debug>
+
+Enable direct output to STDERR instead of autoinst-log.txt
+
+=item B<-h, -?, --help>
+
+Show this help.
+
+=cut
+
+use strict;
 
 my $installprefix;    # $bmwqemu::scriptdir
 
@@ -43,7 +60,7 @@ use autotest;
 use commands;
 use distribution;
 use testapi 'diag';
-use Getopt::Std;
+use Getopt::Long;
 require IPC::System::Simple;
 use autodie ':all';
 no autodie 'kill';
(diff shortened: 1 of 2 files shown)


isotovideo
```diff
@@ -53,21 +70,24 @@ use Carp 'cluck';
...
-$Getopt::Std::STANDARD_HELP_VERSION = 1;
+my %options;
 
-sub HELP_MESSAGE {
-    print "$0 [-d]\n";
-    print "Parses vars.json and tests the given assets/ISOS\n\n";
-    print " -d enables direct output to STDERR instead of autoinst-log.txt\n";
+sub usage($) {
+    my $r = shift;
+    eval "use Pod::Usage; pod2usage(1);";
```
aaannz: Again perl-critic
`./isotovideo: Expression form of "eval" at line 79, column 5.  See page 161 of PBP.  (Severity: 5)`

Use
```perl
eval { use Pod::Usage; pod2usage(1); };
```
instead

isotovideo
```diff
@@ -53,21 +70,24 @@ use Carp 'cluck';
...
 
-# avoid paranoia
-$Getopt::Std::STANDARD_HELP_VERSION = 1;
+my %options;
 
-sub HELP_MESSAGE {
-    print "$0 [-d]\n";
-    print "Parses vars.json and tests the given assets/ISOS\n\n";
-    print " -d enables direct output to STDERR instead of autoinst-log.txt\n";
+sub usage($) {
```
aaannz: This doesn't fly with perl-critic
`./isotovideo: Subroutine prototypes used at line 77, column 1.  See page 194 of PBP.  (Severity: 5)`

okurz:
perlcritic complaints fixed, thx

isotovideo
```diff
@@ -53,21 +70,23 @@ use Carp 'cluck';
...
-    print " -d enables direct output to STDERR instead of autoinst-log.txt\n";
+sub usage {
+    eval { use Pod::Usage; pod2usage(1); };
+    if ($@) {
+        die "cannot display help, install perl(Pod::Usage)\n";
+    }
 }
 
+GetOptions(\%options, 'debug|d', 'help|h|?') or usage(1);
+usage(0) if $options{help};
```
aaannz: I have just one question, why calling `usage(1)` and `usage(0)` when the parameter doesn't seem to be used in `usage`. Was that meant to be passed to `pod2usage`?

okurz: I dunno, maybe copy-pasted from wrong example. I guess I got it from openQA/script/client

okurz: how would you like it to be?

aaannz: I think passing it to the `pod2usage` is what we want. In case of wrong options -> exit code 1, in case of explicit help -> exit code 0.

okurz: done

PR #726, opened 2017-03-01
[WIP] Give more feedback for unable to connect

Soulofdestiny:


okurz:
can we be more specific about which network and which VNC and such? As it took me a year to understand what components are involved for the zkvm/zVM backends :-)

backend/baseclass.pm
```diff
@@ -991,6 +991,7 @@ sub new_ssh_connection {
         }
         else {
             bmwqemu::diag "Could not connect to $args{username}\@$args{hostname}, Retry";
+            bmwqemu::diag "Possible solutions: check network-settings e.g. VIRSH_CMDLINE or S390_NETWORK_PARAMS and VNC/SSH settings in there";
```
foursixnine: I will move this to line 1000, as the connection might not be ready when the loop starts, and only show it when it dies.

Soulofdestiny: but in line 1000 it will show this everytime, no?

aaannz: Seems to me this will indeed print the string every time. I would move the inline condition to block one. I would also remove the "Possible solutions:" substring and just begin with "check network setttings..."

okurz:
On Wednesday, 1 March 2017 01:42:31 CET Ondřej Holeček wrote:
> Seems to me this will indeed print the string every time. I would move the
> inline condition to block one. I would also remove the "Possible
> solutions:" substring and just begin with "check network setttings..."

but it's very vague so it could also be another cause.

Soulofdestiny:
yup, will be executed everytime:
`
14:44:07.7472 28290 Connection to root@s390pb.suse.de established
14:44:07.7472 28290 Possible solutions: check network-settings e.g. VIRSH_CMDLINE or S390_NETWORK_PARAMS and VNC/SSH settings in there
Connected to domain openQA-SUT-21
`

I will update PR

Soulofdestiny:
oops - totally forgot about this one...

coolo:
I will close on monday

PR #728, opened 2017-03-10
Check if mouse pos var is defined before use

asdil12:
Fixes https://progress.opensuse.org/issues/17058


coolo:
I would rather init the mouse with -1, -1 instead of undef to fix this. Do you see a problem with that approach?

asdil12:
Patch updated

PR #721, opened 2017-02-10
Add Test::Warning where possible

okurz:

PR #729, opened 2017-03-10
Remove String::Random dependency

aaannz:
String::Random is not packaged in Tumbleweed and instead adding new
dependency redo filename generation by using File::Temp

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -57,7 +57,6 @@ on 'test' => sub {
   requires 'Test::MockModule';
   requires 'Pod::Coverage';
   requires 'Devel::Cover';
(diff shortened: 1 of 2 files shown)


aaannz:
See https://github.com/os-autoinst/os-autoinst/pull/721#discussion_r105402223

aaannz:
@foursixnine Any objections?

okurz:
LGTM

aaannz:
Added `OPEN => 0` to prevent `tempfile` creating the file. Now it is sort of equivalent to the original one. Except it also uses alphanum instead of just numeric random string.

PR #731, opened 2017-03-13
[WIP] Allow to run same module with different input

asmorodskyi:


asmorodskyi:
Currently it is just POC . Want to get a feedback before start full implementation . Idea actually inspired by  @DimStar77  PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/2492 . So my PR will allow to shrink his PR to one new test module instead of 30. 
@sysrich  what do you think about implementing your idea of infrastructure for well-defined package testing with this feature ? 
@coolo , @okurz  , @foursixnine please tell your opinion .

coolo:
This will most likely run into a wall as the webui maps the test name to a source (similiary as the "allow multiple loadtest of the same module" feature). But I general I agree with the requirement - and API-wise it makes sense.

But note that the first argument in run is still $self

autotest.pm
```diff
@@ -45,7 +45,7 @@ sub loadtest {
     my $category = $1;
     my $name     = $2;
     my $test;
-    my $fullname = "$category-$name";
+    my $fullname = "$category-$name-$arg";
```
okurz: @coolo this should take care to name the module unambiguously and correspondingly for the webui.

asmorodskyi:
@okurz I tried to do the run with custom isotovideo . Result is unpredictable - test hangs forever . Need to see why it happens , will continue working on this ...

autotest.pm
coolo: no, as the webui is not seeing this variable. This is only used to e.g. identify the SKIPTO and such. And as such it actually creates a problem. an empty arg shouldn't appear as $name- but as $name (leave alone you will get undefined warnings as it stands)

coolo:
please reopen on next hackweek :)

PR #733, opened 2017-03-14
Set PRJDIR to default if not provided by vars.json

aaannz:
if PRJDIR is not provided by vars.json try to set it to default
'/var/lib/openqa/share'. But do so only if CASEDIR is subdir of
this directory, otherwise report error and bail out

See https://progress.opensuse.org/issues/16362

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -124,18 +124,16 @@ sub init {
     $| = 1;
     select($oldfh);
 
-    unless ($vars{CASEDIR}) {
-        die "DISTRI undefined\n" . pp(\%vars) . "\n" unless $vars{DISTRI};
-        my @dirs = ("$scriptdir/distri/$vars{DISTRI}");
-        unshift @dirs, $dirs[-1] . "-" . $vars{VERSION} if ($vars{VERSION});
-        for my $d (@dirs) {
-            if (-d $d) {
-                $vars{CASEDIR} = $d;
-                last;
-            }
+    die "CASEDIR variable not set in vars.json, unknown test case directory" if !$vars{CASEDIR};
+
+    unless ($vars{PRJDIR}) {
+        if (index($vars{CASEDIR}, '/var/lib/openqa/share') != 0) {
+            die "PRJDIR not specified and CASEDIR ($vars{CASEDIR}) does not appear to be a
+                subdir of default (/var/lib/openqa/share). Please specify PRJDIR in vars.json";
         }
-        die "can't determine test directory for $vars{DISTRI}\n" unless $vars{CASEDIR};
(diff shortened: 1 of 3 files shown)


aaannz:
@okurz WDYT

bmwqemu.pm
```diff
@@ -153,7 +162,7 @@ sub init {
     ## env vars end
 
     ## some var checks
-    if (!-x $gocrbin) {
+    if ($gocrbin && !-x $gocrbin) {
```
okurz: unrelated. Please move to different commit (same PR is fine)

bmwqemu.pm
```diff
@@ -126,7 +126,7 @@ sub init {
 
     unless ($vars{CASEDIR}) {
         die "DISTRI undefined\n" . pp(\%vars) . "\n" unless $vars{DISTRI};
-        my @dirs = ("$scriptdir/distri/$vars{DISTRI}");
+        my @dirs = ("$scriptdir/tests/$vars{DISTRI}");
```
okurz: I don't get the meaning of "tests/" dir. Is this also only a nice "default here?

t/12-bmwqemu.t
```diff
@@ -0,0 +1,117 @@
...
+    open($varsfh, '<', 'vars.json') || BAIL_OUT('can not open vars.json for reading');
+    my $ret;
+    eval { $ret = JSON->new->relaxed->decode(<$varsfh>); };
+    die "parse error in vars.json:\n$@" if $@;
+    close($varsfh);
+    my %vars = %{$ret};
+    ok(!$vars{DISTRI}, 'DISTRI not supplied and not set');
+    is($vars{CASEDIR}, $dir, 'CASEDIR unchanged');
+    ok(!$vars{PRJDIR}, 'PRJDIR not supplied and not set');
+};
```
okurz: that test has a lot of duplication. Makes hard to see what are the differences among the three subtests. Can you extra helper function(s)?

bmwqemu.pm
aaannz: I don't get this check at all. We don't have `distri` directory AFAIK, but use tests instead. So this test IMO never passed. `$striptdir` is also set to some `/usr/lib/os-autoinst` instead of `/var/lib/openqa[/share]`. I'll remove this part and consider `CASEDIR` to be mandatory.

aaannz:
Updated and added one more commit to remove obsolete hack

okurz:
LGTM, more DRY in test would be good but ok

aaannz:
I did consider moving those `eval` blocks to sub, but I really didn't want to hide where the actual test is happening and what is being done.

okurz:
LGTM then

PR #734, opened 2017-03-17
Remove needle caching - we want to do this in the worker

coolo:


okurz:
we have yet to see if the worker caching works better than this one but I don't mind removing it here either way.

coolo:
what we know: this does not work as it leaves us with the backend accessing NFS.

PR #735, opened 2017-03-17
Only abort test on stall if assert_screen failed (not check)

AdamWill:
As discussed in #713, this code which aborts the test if a stall
is detected during a needle check has a problem. It will abort
the test even in `check_screen`, not just `assert_screen` - so
it can cause tests to fail which otherwise would not have.

This makes `check_asserted_screen` only register whether a stall
was detected or not, and makes `_check_backend_response` abort
the test if a stall was detected *and* it's in 'assert mode'. If
it's in 'check mode', it just logs a warning about the stall,
but will continue the test.

I'm a bit unsure what to do about the backend.crashed file.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -852,15 +852,14 @@ sub check_asserted_screen {
         # make sure we recheck later
         $self->assert_screen_last_check(undef);
 
-        if ($self->stall_detected) {
-            backend::baseclass::write_crash_file();
-            bmwqemu::mydie "assert_screen fails, but we detected a timeout in the process, so we abort";
-        }
         my $failed_screens = $self->assert_screen_fails;
         # store the final mismatch
         push(@$failed_screens, [$img, $failed_candidates, 0, 1000]);
         my $hash = $self->_failed_screens_to_json;
         $hash->{image} = encode_base64($img->ppm_data);
+        # store stall status
+        $hash->{stall} = $self->stall_detected;
+
         return $hash;
     }
 
--- a/testapi.pm
+++ b/testapi.pm
@@ -261,6 +261,16 @@ sub _check_backend_response {
                 );
             }
         }
+        # Handle case where a stall was detected: fail if this is an
+        # assert_screen, warn if it's a check_screen
+        if ($rsp->{stall}) {
+            if (!$check) {
+                record_info('Stall detected', 'Stall was detected during assert_screen fail', result => 'fail');
+            }
+            else {
+                bmwqemu::fctwarn("stall detected during check_screen failure!");
+            }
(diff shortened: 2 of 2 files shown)


AdamWill:
This is kind of a trial balloon - I'm really just floating it for comments. See #713 for more discussion. @okurz @coolo wdyt?

coolo:
the backend crash file makes sure the job is taken as incomplete. And once it's there, the worker will kill the job. So there is no gain for what you want.

coolo:
I.e. right now I see no way to have it as incomplete *and* handle it in the testapi. But possibly this is good enough - so we have it fail with a proper test result. Write a text result "Stall detected" as final failure, so we do not have to lookup the log file to know why.

okurz:
good approach. I also thought about this recently after I understood that people are complaining about this because it aborts their runs in `check_screen`, not `assert_screen`. Your approach looks good.

Regarding the "text result" as final failure: Will this help when the current test module details are never shown in incomplete jobs?

coolo:
Right. we would have a defined exit and a proper reporting

AdamWill:
Sorry, I don't quite follow. To be clear, is the suggestion to just drop writing the 'crash file' entirely, and make no other change to the pull request? And the expectation that this will result in the test failing 'normally' when a stall is detected in an `assert_screen`? Or did I misunderstand?

coolo:
leave out the crash file and make use of record_resultfile instead

AdamWill:
Still not totally sure I grok it - is that what you meant?

okurz:
looks reasonable to me.

AdamWill:
Er, yikes, okay, well I guess we'll find out if it works then :P I hadn't actually tested this yet. I'll throw it onto our staging instance tomorrow.

sysrich:
Thank you! Let us know :)

AdamWill:
So here's an example of the failure case, so you can check it looks how you expected:

https://openqa.stg.fedoraproject.org/tests/87836

note the failure was in a non-important module, so the next module still ran. But the module where the failure occurred is indeed shown as 'failed', and the post-fail hook ran (not sure if you were expecting that or not, and you can see 'Stall detected' on the thumbnail.

I haven't yet spotted a test where the 'warning' case happened, but of course it's a bit harder as you have to poke through the `os-autoinst.log` for every test pass to find one :)

AdamWill:
As a general thought, it might be nice if these kinds of messages could be read from the API. Then I could write a plugin which automatically restarts tests when stalls occur, cos I'd like that behaviour for our instance (because I can't figure out what the hell's causing the goddamn stalls to happen and fix it).

PR #736, opened 2017-03-20
Make qcow images self-contained by compression

okurz:
Related progress issue: https://progress.opensuse.org/issues/17790


foursixnine:
I would change the commit message to explicitly say that compression is being enabled by default.

okurz:
done

foursixnine:
LGTM

aaannz:
For uploading we have threshold 1000 second so compressing should not trigger dead worker issue. Fine by me.

PR #737, opened 2017-03-20
s390: ensure to copy the right image from nfs on zKVM host

Soulofdestiny:


Soulofdestiny:
@coolo @okurz 

related poo: https://progress.opensuse.org/issues/17796
local verifcation: http://opeth.suse.de/tests/4934

coolo:
definitely not. The hardcoding needs to be done in the caller

PR #738, opened 2017-03-21
Add console 'post hook' callback

okurz:
This allows to execute a command after every select_console.

Related progress issue: https://progress.opensuse.org/issues/14826


okurz:
see http://lord.arch/tests/6032 for a first verification how it can be used.

okurz:
updated name

PR #739, opened 2017-03-21
Relax a little

coolo:


coolo:
This is what I'm testing atm

okurz:
LGTM

coolo:
I changed the numbers a bit. I fear 500 kps are a bit too agressive.

coolo:
(even in my testing it was pretty fine)

coolo:
btw: we might want to make the big type_string optional in normal testing. it's a bit excessive :)

okurz:
When you tested and it was fine, why the limit? 
We can still make it less aggressive whenever we encounter a problem and who would ever test more with a more aggressive setting?

coolo:
Call it guts felling, call it experience. And as this morning I'm reminded on "QEMU: usb-kbd: warning: key event queue full" (with the old default) - so we have platforms where even the old 15 kps are a problem :(

PR #740, opened 2017-03-21
Forward custom arguments to console_selected callback

okurz:
Follow-up to https://github.com/os-autoinst/os-autoinst/pull/738

For the cases where a 'console_selected' callback is not desired but a test
distri specific one has been specified the callback can be configured to be
not called at all. For example when the callback is a check for the console to
be shown but the check should be done explicitly in the test and it is not
desired to overwrite the distribution callback.

Related progress issue: https://progress.opensuse.org/issues/14826

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1249,7 +1249,11 @@ our %testapi_console_proxies;
 
 =head2 select_console
 
-  select_console("root-console")
+    select_console($console [, @args]);
+
+Example:
+
+  select_console("root-console");
 
 Select the named console for further C<testapi> interaction (send_text,
 send_key, wait_screen_change, ...)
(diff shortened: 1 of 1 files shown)


coolo:
I'm not convinced. What is the use case? Can't the callback decide this just as well?

okurz:
I will show you what I have as tests so far, then we can decide how to continue

okurz:
changed to "Forward custom arguments to console_selected callback"

testapi.pm
```diff
@@ -1261,10 +1265,13 @@ object (in this example) - so it will ensure the console is active,
...
 to run test code.
 
+After the console selection the distribution callback
+C<$distri->console_selected> is called with C<%args>.
+
 =cut
 
 sub select_console {
-    my ($testapi_console) = @_;
+    my ($testapi_console, %args) = @_;
```
coolo: why a hash? If you make it an array it's completely up to the distribution to treat it as array or as a hash (I love perl's misfeatures :)

kraih: That's a feature (not a misfeature), but i agree that `@args` would seem appropriate here.

PR #741, opened 2017-03-22
Swtich to virtio-gpu on aarch64

foursixnine:
To avoid unreliable test results in our arm workers, this switch to
virtio-gpu which requires qemu version 2.8 will allow to use proper
video emulation.

This video shows the switch from terminal console to X back and forth (http://xgene2.arch/tests/158/file/video.ogv) in the autoinst-log.txt file the `-device virtio-gpu-pci` option can be seen also. 

This also requires two PR from gitlab defined in poo#17740

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -394,7 +394,7 @@ sub start_qemu {
         $arch_supports_boot_order = 0;
     }
     if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm') {
-        push @vgaoptions, '-device', 'VGA';
+        push @vgaoptions, '-device', 'virtio-gpu-pci';
         $arch_supports_boot_order = 0;
         $use_usb_kbd              = 1;
     }


okurz:
how about opensuse arm worker and the arm workers of *the whole world outside suse*?

coolo:
as long it's just qemu 2.8 and not some SUSE patch, I don't think there is a problem with the world

okurz:
Considering I have only qemu-2.6.2 on my openSUSE Leap 42.2 which is a recent system not quite there yet

foursixnine:
So, here's the upstream patch: https://lists.gnu.org/archive/html/qemu-devel/2015-02/msg04342.html and there's also (this)[https://lists.gnu.org/archive/html/qemu-devel/2016-06/msg01220.html] which in conjunction enable us to use -device virtio-gpu

coolo:
What you have on *your* 42.2 does not matter - requiring a new qemu for a new architecture isn't off limit. Your system is very likely not aarch64

okurz:
agreed, aarch64 can be considered non-production ready enough to cover this

PR #742, opened 2017-03-24
Copy HDD & ISO images from NFS to host's storage

mnowaksuse:


consoles/sshVirtsh.pm
```diff
@@ -541,9 +538,9 @@ sub run_cmd {
 
     my $chan = $self->{ssh}->channel();
     $chan->exec($cmd);
-    $chan->send_eof;
     bmwqemu::diag "Command executed: $cmd";
     get_ssh_output($chan);
+    $chan->send_eof;
```
okurz: unrelated

PR #743, opened 2017-03-27
Make it possible to define slow typing machines

coolo:
After a lot of forth and back, setting this to 10 is the only cure
for power and arm :(


okurz:
LGTM

PR #744, opened 2017-03-29
Fix assert_and_click mouse hiding

AdamWill:
See https://progress.opensuse.org/issues/17058#note-7 - #728
caused a problem with the cursor reposition that occurs at the
end of `assert_and_click`, in the case that the cursor had not
been explicitly positioned before the `assert_and_click`. In
this case, `assert_and_click` is meant to call `mouse_hide`.
However, #728 made `get_last_mouse_set` *always* return defined
values, while `assert_and_click` expected it to return undefined
values when the cursor had never been positioned, so it wound
up trying to position the cursor to -1, -1 instead of calling
`mouse_hide`.

The fix is pretty simple: just adapt the condition to the new
behaviour of `get_last_mouse_set`.

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -45,6 +45,9 @@ sub fake_read_json {
     elsif ($cmd eq 'backend_mouse_hide') {
         return {ret => 1};
     }
+    elsif ($cmd eq 'backend_get_last_mouse_set') {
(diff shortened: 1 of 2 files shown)


coolo:
I would have made it >= 0 to make it sound less negative, but it will do :)

PR #745, opened 2017-03-29
Die when validate_script_output called without $code

asmorodskyi:
No sense of using validate_script_output without callback code
for validation so function will fail with
proper error message when it is happening

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -912,7 +912,6 @@ sub validate_script_output($&;$) {
     $wait ||= 30;
 
     my $output = script_output($script, $wait);
-    return unless $code;
     my $res = 'ok';
 
     # set $_ so the callbacks can be simpler code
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -907,12 +907,12 @@ Wrapper around script_output, that runs a callback on the output. Use it as
 
 =cut
 
-sub validate_script_output($&;$) {
+sub validate_script_output {
```
okurz: good but unrelated. Please put into at least another commit if not PR where you try to remove the signature for all sub's where possible

coolo: Not good actually - did you test this?

coolo: So I think I want a test case

asmorodskyi:
it's good that @coolo suggested to add tests :) 
they show that with ($;$) you actually can't reach die message , so revert back ($;$) but still you can't reach "return unless $code" too , so this commit is about removing one useless line and add 100% test coverage for the method :)

okurz:
LGTM

coolo:
I knew your change wouldn't work - so thanks for the test case :)

PR #746, opened 2017-03-31
[WIP] Add test API call 'set_typing_speed'

okurz:
Related progress issue: https://progress.opensuse.org/issues/18200

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -200,6 +200,14 @@ subtest 'record_info' => sub {
     like(exception { record_info('my title', 'output', result => 'not supported', resultname => 'foo') }, qr/unsupported/, 'invalid result');
 };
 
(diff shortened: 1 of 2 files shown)


coolo:
I would think we better go with a generic backend_reload_vars command - and leave it to the testapi to save it first

coolo:
... because that can also be useful if switching scenarios

okurz:
and then you don't want the explicit set_typing_speed additionally?

coolo:
right. set_var would trigger a backend_reload - and we would have to make sure not to use set_var in backend :)

asmorodskyi:
poo#18200 marked as resolved is this PR is still relevant ?

okurz:
it serves as a reminder that we would like to have a backend command for saving/loading the variables covering all threads but I will close this PR.

PR #747, opened 2017-03-31
Prevent videoencoder from creating any video file on '-n'

okurz:
Verified locally with openQA.

Test: http://lord.arch/tests/6112#downloads, no video is uploaded as NOVIDEO=1
is set
Regression test: http://lord.arch/tests/6113#downloads, video exists

Related progress issue: https://progress.opensuse.org/issues/15896

### DIFF
--- a/videoencoder.cpp
+++ b/videoencoder.cpp
@@ -32,6 +32,7 @@ video.
 #define _FILE_OFFSET_BITS 64
 
 #include "theora/theoraenc.h"
+#include <cassert>
 #include <cstdio>
 #include <getopt.h>
 #include <ogg/ogg.h>
@@ -58,6 +59,7 @@ static int theora_write_frame(th_ycbcr_buffer ycbcr, int last)
 {
     ogg_packet op;
     ogg_page og;
+    assert(ogg_fp);
 
     /* Theora is a one-frame-in,one-frame-out system; submit a frame
    for compression and pull out the packet */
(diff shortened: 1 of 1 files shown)


videoencoder.cpp
```diff
@@ -173,11 +175,13 @@ int main(int argc, char* argv[])
...
-    ogg_fp = fopen(option_output, "wb");
-    if (!ogg_fp) {
-        fprintf(stderr, "%s: error: %s\n", option_output,
-            "couldn't open output file");
-        return 1;
+    if (output_video) {
+        ogg_fp = fopen(option_output, "wb");
+        if (!ogg_fp) {
+            fprintf(stderr, "%s: error: %s\n", option_output,
+                    "couldn't open output file");
```
asmorodskyi: Can we extend this with $? to get chance understand what is wrong ?

Soulofdestiny:
I'm missing the 'why' :/

videoencoder.cpp
coolo: $? is pretty much $? in C :)

videoencoder.cpp
```diff
@@ -72,11 +72,13 @@ static int theora_write_frame(th_ycbcr_buffer ycbcr, int last)
     }
 
     ogg_stream_packetin(&ogg_os, &op);
+    if (!ogg_fp) {
```
coolo: why would this function even called if not?

okurz: updated, not called anymore and the "if" replaced by "assert" at start of function

okurz:
updated:
* theora_write_frame is not called at all anymore
* made benefit even more explicit in commit message

PR #748, opened 2017-04-03
svirt: Do not ignore "0" when reading serial file

coolo:
See https://progress.opensuse.org/issues/15732


mnowaksuse:
Tested it with CaaSP Xen HVM images.

PR #750, opened 2017-04-05
Improve 'connection refused' error handling

okurz:
See http://lord.arch/tests/6127/file/autoinst-log.txt for an example.

Related progress issue: https://progress.opensuse.org/issues/18332

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -860,7 +860,7 @@ sub _receive_message {
     my $self = shift;
 
     my $socket = $self->socket;
-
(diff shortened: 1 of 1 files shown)


coolo:
that fixes one problem - but the main problem is that the QEMU errors are very well hidden in the log :(

okurz:
yes, I don't know how to make them more obvious unless using some ascii art

```
This is not the error message you are looking for, look further down for qemu errors: ----\
                                                                                          |
                                                                                          |
                                                                                          |
                                                                                          |
                                                                                          V
```

coolo:
see #753 for my version

PR #754, opened 2017-04-07
[WIP] Initial support for NVRAM uefi variable storage using QEMU_PFLASH variable

thehejik:
Refactoring of QEMU_PFLASH variable for future use.

Problem description: until now we miss a persistent storage for uefi vars which contains for eg. boot-order values for efi firmware. It means that if we want to use image generated on uefi machine in different job the machine will not boot from provided image but from mounted iso/DVD instead. If we transfer also uefi-vars.qcow2 together with hdd image we will be able to boot directly from the image.

https://wiki.linaro.org/LEG/UEFIforQEMU

TODO:
1) add support for publishing hdd image together with uefi-vars.qcow2  over PUBLISH_HDD variable
2) do not override uefi-vars.qcow2 for children jobs using QEMU_PFLASH and same BIOS value as parent job
3) make it compatible with x86_64 (current code compatible only with aarch64)

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -668,8 +668,17 @@ sub start_qemu {
 
         if ($vars->{UEFI_PFLASH}) {
             # Convert the firmware file into qcow2 format or savevm would fail
-            runcmd('qemu-img', 'convert', '-O', 'qcow2', $vars->{BIOS}, 'ovmf.bin');
-            push(@params, "-drive", "if=pflash,format=qcow2,file=ovmf.bin");
+            # Prepare clean 64M nvram image for uefi vars in raw format
+            runcmd("dd", "if=/dev/zero", "of=uefi-vars.bin", "iflag=fullblock", "bs=1M", "count=64");
+            # Convert uefi raw fw from BIOS var into 64M image for uefi code in raw format
+            runcmd("dd", "if=$vars->{BIOS}", "iflag=fullblock", "bs=1M", "of=uefi-code.bin");
+            runcmd("truncate", "-s", "64M", "uefi-code.bin");
+            # Convert both raw images into qcow2 format
+            foreach my $i (qw(uefi-vars uefi-code)) {
+                runcmd("qemu-img", "convert", "-O", "qcow2", "$i.bin", "$i.qcow2");
+            }
+            push(@params, "-drive", "if=pflash,format=qcow2,file=uefi-code.qcow2,unit=0,readonly=on");
+            push(@params, "-drive", "if=pflash,format=qcow2,file=uefi-vars.qcow2,unit=1");
         }
         elsif ($vars->{BIOS}) {
             push(@params, "-bios", $vars->{BIOS});


thehejik:
for x86_64 the size of uefi-vars.bin should be 8MB (64MB for aarch64):
```
            runcmd("dd", "if=$vars->{BIOS}", "iflag=fullblock", "bs=1M", "of=uefi-code.bin");
            runcmd("truncate", "-s", "8M", "uefi-code.bin");
```

okurz:
good idea in general. Unfortunately I can't help with the implementation :-/

foursixnine:
> But still no luck with uefi-vars.code where the size is unclear to me

If i remember correctly the thing is: that the sum of the bios + the vars needs to be exactly 8mb or 64mb... Maybe @agraf help will be great here:),

agraf:
Both need to be exactly 64MB each (on ARM). IIRC vars is just /dev/zero.

thehejik:
The code is working properly for ARM (64MB both files), but not for x86_64 with the same sizes. I found out that size of EFI ROM file on x86_64 have to be 8MB but no idea what value for VARS file.

agraf:
I don't think x86 has those strict sizing constraints. Can you just use the 128kb /usr/share/qemu/ovmf-x86_64-vars.bin as template? Also don't pad the code one either - just use it as is.

coolo:
any update here? this is pending for 3 weeks now

sysrich:
ping - any update here? @thehejik

thehejik:
@coolo, @sysrich no big progress yet, once I finish GRE tunnels I will continue with that.

richiejp:
I have decided to try incorporating this into the QEMU backend refactor: https://github.com/os-autoinst/os-autoinst/pull/942/commits/f9be186ca07c992536602bdf9309b56c5d5b427f

Instead of creating the firmware and vars file in os-autoinst I just allow the user to specify the file assets which will be used with the pflash drives. That way we don't have to hardcode how the images are created and the vars file can be published and reused like hard drives can.

thehejik:
@richiejp that's great. Please let me know once you incorporate this and I will close this PR, thanks

richiejp:
@thehejik I have it working on X86 with publishing, see: http://rpws.suse.cz/tests/398#downloads and the chained jobs. Still haven't tested it on ARM, I will try that next. Also the OpenQA UI needs a slight modification to recognise pflash images as hard disks (or we prepend the variable with HDD_ instead of UEFI_), but publishing already works if you provide absolute paths for the image file.

richiejp:
Started testing on aarch64: http://rpws.suse.cz/tests/407#, it at least boots so far.

thehejik:
Thanks @richiejp, it looks good - there is no sign of uefishell anymore \o/

From your x86 verification runs I can see that publishing and reusing pflash is working correctly. 

But the chained job http://rpws.suse.cz/tests/408 for aarch64 has failed because "Cannot find HDD_1 asset hdd/SLES-15-aarch64-652.1@aarch64-minimal_with_sdk652.1_installed.qcow2!".

Do you know the reason for? The qcow2 filenames match...

richiejp:
I probably did not have asset downloads setup correctly. I am running OpenQA on the worker now: http://centriq2.arch.suse.de/tests so should work OK.

richiejp:
It managed to boot from a saved vars file: http://centriq2.arch.suse.de/tests/5# so I guess you can close this. Surprisingly I haven't had to change the test boot module, it just worked.

richiejp:
spoke too soon... it used uefi menu.

thehejik:
uefi menu workaround came from wait_boot() from <tests>/sle/lib/opensusebasetest.pm especially from handle_uefi_boot_disk_workaround(), see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/3aaf825b28f5f94304f83a99729ff355524e945e/lib/opensusebasetest.pm#L251

thehejik:
Thanks, closing this PR

PR #755, opened 2017-04-07
Add support for older qemu versions on aarch64

foursixnine:
Since the switch to virtio-gpu-pci (64476a5) some of our workers have
stop working properly. This is due to the worker only having support
For VGA. (This is the counterpart of bsc#1023101)

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -395,7 +395,8 @@ sub start_qemu {
         $arch_supports_boot_order = 0;
     }
     if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm') {
-        push @vgaoptions, '-device', 'virtio-gpu-pci';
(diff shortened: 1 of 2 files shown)


coolo:
that's a shitty variable name, sorry. Please prefix it with QEMU_ as other variables specific to the qemu backend. And seperate nouns. The question is why would this be specific to aarch64.

Because we already have QEMUVGA (without _ unfortunately). Why not set this to vga for the old workers and in this case skip the virtio-gpu

foursixnine:
> Please prefix it with QEMU_ as other variables specific to the qemu backend. And seperate nouns. 
Done :) Also fixed the logic there, still not happy with the name of the variable either

> The question is why would this be specific to aarch64.
Tbh, i was wondering the same, i was doing some tests on my local worker... but graphics behave strange a bit when using device virtio-gpu-pci (works better if -vga cirrus is added, -vga std makes it slower), will keep on testing to see what i can come up with.

foursixnine:
Using -vga virtio ends up on this: http://deimos.suse.de/tests/559#step/bootloader/5).

coolo:
please tidy

PR #761, opened 2017-04-11
lockapi: Graceful retry of locking when locks not ready yet

okurz:
Related progress issue: https://progress.opensuse.org/issues/18184

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -25,19 +25,29 @@ our @EXPORT = qw(mutex_create mutex_lock mutex_unlock mutex_try_lock barrier_cre
 require bmwqemu;
 use mmapi 'api_call';
 
-sub _lock_action {
-    my ($name, $where) = @_;
-    my $param = {action => 'lock'};
-    $param->{where} = $where if $where;
-    my $res = api_call('post', "mutex/$name", $param)->code;
-    return 1 if ($res == 200);
-
-    bmwqemu::mydie "mutex lock '$name': lock owner already finished" if $res == 410;
-
+sub _try_lock {
+    my ($type, $name, $param) = @_;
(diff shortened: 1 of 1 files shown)


lockapi.pm
```diff
@@ -29,9 +29,16 @@ sub _lock_action {
     my ($name, $where) = @_;
     my $param = {action => 'lock'};
     $param->{where} = $where if $where;
-    my $res = api_call('post', "mutex/$name", $param)->code;
-    return 1 if ($res == 200);
-
+    # try up to 7 times
+    my $res = "";
+  retry:
```
aaannz: I'm confused. Why are you adding labels?

okurz: because of limited perl knowlegde of @Soulofdestiny and me where we encountered at least once a loop and some statement inside which abort the whole outer function and not just the inner loop. See the "last" further down. To my understanding of [last](https://perldoc.perl.org/functions/last.html) we don't need the label here but I am scared… can you help us?

aaannz: The `retry` label is indeed not used here and what's worse IMO is that the `retry` label is used twice. How perl should know which one to use?
But I have issue even with what is the patch trying to solve. `barrier_wait` and `mutex_lock` both waits already, for `*_try_*` function I thing we don't want to wait anyway. In the mentioned poo# I really didn't get why is this needed.

okurz: hm, actually it's https://progress.opensuse.org/issues/18184

okurz: updated without the label and extracted method for DRY

aaannz:
I'm not very comfortable with the inner waiting loop, but that's something I can live with. LGTM.

PR #764, opened 2017-04-19
In case of stalled VNC relogin instead of showing black

coolo:
The black screen was good enough for the use case we implemented it for (shut down
of s390 guest), but is very confusing in general. Now we try to relogin for 10 seconds
and will die if this fails. The backend is blocked during this period, so assert_screen
will not continue until we have a proper new picture

Tests better do not assert_screen on 'now rebooted' on platforms where the reboot drops
the VNC connection, but rather switch to a console that allows e.g. serial waits

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -826,15 +826,15 @@ sub send_update_request {
     my $time_since_last_update = time - $self->_last_update_received;
 
     # if there were no updates, send a forced update request
-    # to get a defined live sign. If that doesn't help, consider
-    # the framebuffer outdated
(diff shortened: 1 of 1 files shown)


coolo:
I tried to test this, but the only way I have to reproduce the problem is kill -STOP of qemu, but then qemu would accept the connect but does not react. So we are stalling for longer - which is a problem.

And I don't know if this is good enough for xen testing - waiting for michal to return, but we could test this with the s390 workers.

okurz:
sure, any way I could reproduce the issue on s390x to test?

coolo:
I wish I knew

PR #765, opened 2017-04-20
Fix svirt add_disk for remote files

coolo:
In case of patched systems, add_disk files are not readable on the
openqa worker but only on the remote system - and as such there is no
point in either checking nor calling rsync

See https://openqa.suse.de/tests/887899#step/bootloader_zkvm_sym/16


coolo:
@Soulofdestiny did you try?

Soulofdestiny:
@coolo Actually, it was blocked due to another issue in the test last week - but apparently it was working on the weekend... 
So I assume we can't expect a breakage there after deployment

PR #766, opened 2017-04-20
Remove all disabled debug statements

okurz:
"print-debugging" effects should not be visible in code.
If necessary, proper logging calls should be used.
Disabled code (with no good explanation) should not be
present in production code. That is what a VCS is for,
e.g. store in private development branches.


coolo:
I don't care for your rules

PR #771, opened 2017-04-25
Enable keepalive for IPMI SOL

sysrich:
The Serial over LAN for ipmiconsole is vulnerable to disconnects if it is not being used interactively. This will force the ipmiconsole to send regular NUL characters to keep the connection alive, which should make it a lot more robust.

This function was precisely intended for situations where an IPMI console might not be having user interactivity but is needed for long term logging, much like we need in some openQA scenarios.


okurz:
LGTM

you did not test it but just assume it should help, right? I mean, it sounds right :-)

PR #772, opened 2017-05-04
virtio_console: Add carry buffer for post match data

richiejp:
Avoid dropping data between calls to wait_serial, by saving surplus data read
from the socket to a persistent 'carry buffer'. Originally I thought that it
might be useful to drop data between calls, but it has only proven to be a
problem. If the old behavior proves to be desirable under some circumstances,
then another optional argument can be added to drop the carry buffer.


okurz:
LGTM

PR #773, opened 2017-05-05
Distinguish between failed and timed out in script run

okurz:


okurz:
@asmorodskyi

testapi.pm
```diff
@@ -726,6 +726,14 @@ sub x11_start_program {
     return $distri->x11_start_program($program, $timeout, $options);
 }
 
+sub _handle_script_run_ret {
+    my ($ret, $cmd, %args) = @_;
```
asmorodskyi: can we get even more info from $ret value ? like what was error code at least or something like this

okurz:
tests fixed

testapi.pm
okurz: the $ret value is visible in the info thumbnails within openQA already showing what the script returned.

PR #776, opened 2017-05-08
Clarify error message for undiscoverable needle

nicksinger:
This clarifies error messages like this:
`10:06:41.8585 17198 post_fail_hook failed: needle(s) 'password-prompt' not found`
(like seen in https://openqa.suse.de/tests/918407/file/autoinst-log.txt)

Proof run is in progress and will be added later. If not needed, feel free to merge and close.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -272,7 +272,7 @@ sub _check_backend_response {
             }
         }
         if (!$check && !$rsp->{saveresult}) {
-            OpenQA::Exception::FailedNeedle->throw(error => "needle(s) '$mustmatch' not found", tags => $mustmatch);
+            OpenQA::Exception::FailedNeedle->throw(error => "no candidate needle with tag(s) '$mustmatch' matched", tags => $mustmatch);
         }
         if ($rsp->{saveresult}) {
             $autotest::current_test->save_test_result();


okurz:
I am not quite sure right now how if it all it differs for the case when there are only non-matching candidates or the case where there are no candidates at all. Or even: Only blacklisted candidates so no candidates available for matching.

nicksinger:
@okurz Sorry i didn't quiet get your point there. Can you please rephrase this thought again?

coolo:
I think he was just thinking aloud :)

okurz:
kinda. Thoughts for the *next* improvement

coolo:
but let me add my 0.02€: 'unable to match any needles' sounds like it actually tried to match needles, but as a matter of fact there are no needles whatsoever to match. This might actually be because all needles are filtered out of the pool or because of a typo in the tag name.

nicksinger:
@coolo your 0.02€ miss a suggestion for a better message :)
Maybe: "Needle pool for tag(s) '$tag' is empty" ?

coolo:
well, "no needles for tag(s) '$mustmatch' found" would suit me :)

coolo:
(and it took me 5 minutes to figure out how to type € on a UK keyboard :)

okurz:
but wait. The job you referenced actually had needles that it tried to match but non matched so IMHO "error message for undiscoverable needle" is totally **wrong** here. This is what my comment was about. I need to check the local unit tests for that cases.

coolo:
hmm, is that possibly a regression due to #768 ? Because 'timeout' is the real error

testapi.pm
```diff
@@ -272,7 +272,7 @@ sub _check_backend_response {
             }
         }
         if (!$check && !$rsp->{saveresult}) {
-            OpenQA::Exception::FailedNeedle->throw(error => "needle(s) '$mustmatch' not found", tags => $mustmatch);
+            OpenQA::Exception::FailedNeedle->throw(error => "unable to match any needles with the tag(s) '$mustmatch'", tags => $mustmatch);
```
okurz: `no candidate needle with tag(s) '$mustmatch' matched` (discussed with @nicksinger in person)

nicksinger:
As discussed with @okurz, here is the updated message.  @coolo fine with you?

PR #777, opened 2017-05-08
Disable keepalive, restart ipmiconsole on non-zero death

sysrich:
Waiting validation run, should restart the ipmiconsole whenever it fails due to a non-zero error code

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -24,7 +24,7 @@ use IO::Select;
 use IO::Socket::UNIX 'SOCK_STREAM';
 use IO::Handle;
 use Data::Dumper;
-use POSIX qw(strftime :sys_wait_h);
+use POSIX qw(strftime :sys_wait_h _exit);
 use JSON;
 require Carp;
 use Fcntl;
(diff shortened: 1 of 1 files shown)


backend/ipmi.pm
```diff
@@ -150,25 +150,35 @@ sub start_serial_grab {
...
-
-        # our supermicro boards need workarounds to get SOL ;(
-        push(@cmd, qw(-W nochecksumcheck));
 
         exec(@cmd);
         die "exec failed $!";
     }
+
+    # monitor serial and restart cmd if it exited with a non-zero error code;
+    while (waitpid($self->{serialpid}, WNOHANG) > 0) {
```
coolo: This only runs once - WNOHANG will return 0 in about 100% of all cases and so you end the loop.

If it was not, it would be a busy loop in the parent, aka the backend - blocking all tests.

This loop has to be (without busy looping) in the serial grabbing child, not the parent.

sysrich:
No longer WIP - Validation run https://openqa.suse.de/tests/925222

Several successful attempts have been made to kill the ipmiconsole just for them to be restarted automatically.

PR #779, opened 2017-05-10
Handle expects expects number, gets string

mnowaksuse:
`wait_serial` returns matched string or undef. However
`_handle_script_run_ret` expects in $1 to have return number. This
change feeds `_handle_script_run_ret` with zero on match, or undef
otherwise.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -824,6 +824,7 @@ sub assert_script_sudo {
     my $str = hashed_string("ASS$cmd");
     script_sudo("$cmd; echo $str-\$?- > /dev/$serialdev", 0);
     my $ret = wait_serial("$str-\\d+-", $wait);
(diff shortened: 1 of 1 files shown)


coolo:
but hardcoding 0 is not a fix

coolo:
this is what script_run does:
```
        my $res = testapi::wait_serial(qr/$str-\d+-/, $wait);
        return unless $res;
        return ($res =~ /$str-(\d+)-/)[0];
```

okurz:
the travis checks fail trying to install the requirements. Can you try to rebase after the update in the .travis.yml file to trigger again? I assume this change would also fix 
https://github.com/os-autoinst/os-autoinst/pull/779
Now I am wondering about how to further DRY regarding the assert_script_sudo part in testapi.pm and the script_run part in distribution but this PR here should come first.

PR #786, opened 2017-05-15
Do not stop the debug viewer but keep running with old/black image

coolo:
It's quite annoying if the debug viewer closes on temporary glitches or
if the test did not yet create a last.png. Just keep it running - it's
for debugging after all


okurz:
+1

PR #789, opened 2017-05-16
Move IPMI SOL handling into a console so we can switch it

coolo:
This does not yet change functionality. But the goal is to have a ssh-col too that we can switch between


coolo:
I reviewed it twice now - let's see it in action

PR #792, opened 2017-05-18
Sort out duplicate needles on multiple tags (poo#19210)

okurz:
Related progress issue: https://progress.opensuse.org/issues/19210

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -33,6 +33,8 @@ use Net::SSH2;
 use feature 'say';
 use OpenQA::Benchmark::Stopwatch;
 use MIME::Base64 'encode_base64';
+use List::Util 'min';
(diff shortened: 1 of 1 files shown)


okurz:
Wanted to test on ppc64le but failed to build os-autoinst. Anyone wants to confirm it's fine? At least the local tests work.

coolo:
it builds in obs

coolo:
LGTM

okurz:
uniq is not yet available in perl(List::Util) package on SLE 12 SP2. I will use List-MoreUtils and test

okurz:
updated. See it in action in http://lord.arch/tests/6269#step/setup_zdup/1 vs. http://lord.arch/tests/6264#step/setup_zdup/1

PR #798, opened 2017-05-24
svirt: Hyper-V support

mnowaksuse:


backend/svirt.pm
```diff
@@ -67,22 +67,53 @@ sub do_stop_vm {
...
-        $self->run_cmd("virsh undefine $vmname");
+        if (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
+            my $ps = 'powershell -Command';
+            $self->run_cmd("$ps Stop-VM -Force -VMName $vmname -TurnOff");
+            $self->run_cmd("$ps Remove-VM -Force -VMName $vmname");
+        }
+        else {
+            $self->run_cmd("virsh destroy $vmname");
+            $self->run_cmd("virsh undefine $vmname");
+        }
```
okurz: Did you think about using OOP instead of if/else?

coolo: There is some truth to that, I will merge anyway

PR #805, opened 2017-05-27
Also compute and store needles git hash

okurz:
This allows easier investigation when the needles repository is its own and
one want so review what changes have been introduced between two test runs.

Example output:

```
17:21:32.5690 5175 git hash in /var/lib/openqa/share/tests/opensuse: 9302c275cdd6e8a55cbf0a34a9bf8bc5f8c176a3
[…]
17:21:33.4503 5175 git hash in /var/lib/openqa/share/tests/opensuse/products/opensuse/needles: 517190eb2793b3903b7eccd34650dae1c0e42e17
```

Obviously, if the needles directory is part of the same git repository it will
output the same git hash for both directories.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -196,27 +196,32 @@ sub signalhandler_chld {
     }
 }
 
+our $test_git_hash;
+our $needles_git_hash;
+
 sub init_backend {
     my ($name) = @_;
     $bmwqemu::vars{BACKEND} ||= "qemu";
 
     # make sure the needles are initialized
-    needle::init($bmwqemu::vars{PRODUCTDIR} . "/needles");
+    my $needles_dir = $bmwqemu::vars{PRODUCTDIR} . '/needles';
+    needle::init($needles_dir);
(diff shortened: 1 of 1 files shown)


isotovideo
```diff
@@ -435,6 +440,8 @@ while ($loop) {
         if ($rsp->{cmd} eq 'version') {
             my $result = {test_git_hash => $test_git_hash, version => $INTERFACE};
             myjsonrpc::send_json($r, $result);
+            $result = {needles_git_hash => $needles_git_hash, version => $INTERFACE};
+            myjsonrpc::send_json($r, $result);
```
coolo: having 2 answers doesn't really make sense

okurz: ok, so should the whole command be replaced by 'version_tests' and 'version_needles'?

coolo: question is if someone cares. The version call itself is pretty unused - but just adding the hash to the result would be good enough *if* someone cared

okurz: done, please recheck

coolo:
tested myself

PR #808, opened 2017-05-30
Improve ipmi

coolo:


coolo:
https://progress.opensuse.org/issues/19136

coolo:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/2916/commits/346b2bb39f4289b0b1631d062a3a810b984b286d contains the password typing. This needs to be aligned

coolo:
I will merge and deploy this, so we can move forward. This will be disruptive, but we have some time

CC @Soulofdestiny @mnowaksuse @alice-suse

PR #809, opened 2017-06-01
Add how to avoid warnings on wait_screen_change

SergioAtSUSE:
Extending the testapi documentation


SergioAtSUSE:
@coolo, what do you think about this?

coolo:
vale

PR #810, opened 2017-06-03
VNC: Handle temporarily unavailable socket during reconnect gracefully

okurz:
Verification run: http://lord.arch/tests/6557#step/reconnect_s390/4

Related progress issue: https://progress.opensuse.org/issues/19262

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -808,6 +808,10 @@ use POSIX ':errno_h';
 sub _send_frame_buffer {
     my ($self, $args) = @_;
 
+    if (!$self->socket) {
+        bmwqemu::diag 'undefined socket, ignoring request for _send_frame_buffer';
+        return;
+    }
     return $self->socket->print(
         pack(
             'CCnnnn',
@@ -878,7 +882,10 @@ sub _receive_message {
     my $self = shift;
 
     my $socket = $self->socket;
-    $socket or die 'socket does not exist. Probably your backend instance could not start or died.';
+    if (!$socket) {
+        bmwqemu::diag 'undefined socket, ignoring read request';
+        return;
+    }
     $socket->blocking(0);
     my $ret = $socket->read(my $message_type, 1);
     $socket->blocking(1);


okurz:
need to cover more locations for access to the socket, waitfor http://lord.arch/tests/6535

coolo:
But we are not supposed to be graceful - having a dead VNC for 8 seconds and counting is fatal.

okurz:
yes, but you turned it into a backend problem causing the job to be incomplete. This will make it not show up in list of previous jobs, make investigation harder and carry over of issues does not work. I am not fixing the dead VNC but turning it into a fail because I consider it an issue of either the tests or the product.

The VNC stall detection is undef'ing the socket variable and other methods try to access it. I consider that wrong from the perspective of the backend itself.

coolo:
You are not turning it into a fail, you are turning it into a pass. Your test just happens to fail later.

okurz:
The VNC stall detection is undef'ing the socket variable and other methods try to access it. Don't you consider this wrong?

coolo:
I didn't say the code is bugfree, I said your PR is wrong

okurz:
the usual coolo …

In https://github.com/os-autoinst/os-autoinst/pull/810/files#diff-1e0c5a0ad8e4c4789e337e7ac95822afL838 you added code to close (and delete) the socket and then IIUC call "login" which is reconnecting the VNC console. So the process shouldn't die here, am I wrong?

coolo:
The process should die if it can't relogin - if you can, handle it as failure. But in general it's an incomplete as something is wrong with the system.

The real failure is in the tests though as the dying VNC is not closed. So far this is only done in shutdown

okurz:
> The process should die if it can't relogin

so in case the relogin was successful the other sockets should have never been accessed?

okurz:
The fix for the failing s390x tests is in but I still consider this PR very useful to not have hidden incompletes with no useful carry over. This change would not fix the failures but should make them much more obvious and easier to fix.

sysrich:
The inability to reconnect to VNC puts the test into an inconsistent and undefinable state.

Ignoring that (as this PR does) and attempting to proceed regardless then relies on proceeding tests to fail as expected

I do not agree this PR is a good idea, it encorages bad behaviour in the backend and requires tests to hold additional logic to handle the potential variables.

It might work for the examples in the PR, but the code in question is used in countless other scenarios, and on that ground i agree with @coolo and give my -1

The job of the backend is to do what it can to provide a consistent framework for tests to run on top, when we detect a failure of the key part of that infrastructure, we shouldn't ignore it and expect the tests to be able to cope with such stupidity.

okurz:
ok, then I guess "dying" in all instances with a proper message is better. What is taking quite some time for reviewers is that incompletes don't show anything useful in the details tab. The whole last module is always missing and never uploaded and one needs to know what to look for in os-autoinst. At best we have a `record_info` screen burning into the user's eyes with an obvious "This is how the test is wrong"-frame :-)

coolo:
I agree, it's just not so easy to communicate between backend and testapi process. Perhaps some 'final words?' callback? :)

okurz:
Basically using the same die handler as for test code itself should work.

coolo:
if the die was in testapi process, yes.

PR #811, opened 2017-06-06
Allow to kill ssh connection

coolo:
We need to stop the ssh serial before we reboot


okurz:
LGTM

What issues does it help with? ipmi?

coolo:
reconnect on ipmi, yes

Before: http://147.2.212.149/tests/37
After: http://147.2.212.149/tests/51

okurz:
nice. As you know I like to turn incompletes into failures ;-)

PR #812, opened 2017-06-09
Needle test case

coolo:
CC @StefanBruens 

https://progress.opensuse.org/issues/19680

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -131,8 +131,9 @@ std::vector<char> str2vec(std::string str_in)
 
 /* we try to the find the best locations - possibly more and will
    weight in later */
-std::vector<Point> minVec(const Mat& m, float min)
+std::vector<Point> minVec(const Mat& m, double& min)
 {
+    min = INT_MAX;
     std::vector<Point> res;
     min += 10;
 
@@ -228,24 +229,11 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
     // Perform the matching. Info about algorithm:
     // http://docs.opencv.org/trunk/doc/tutorials/imgproc/histograms/template_matching/template_matching.html
     // http://docs.opencv.org/modules/imgproc/doc/object_detection.html
-    // Used metric is (sum of) squared differences
     matchTemplate(scene_roi, object_roi, result, CV_TM_SQDIFF);
 
-    // Use error at original location as upper bound
-    Point center = Point(x - scene_x, y - scene_y);
-    double sse = result.at<float>(center);
-    if (sse == 0) {
-        similarity = 1;
-        return { (int)(x), (int)(y) };
-    }
-
-    // We are only interested in points with a similarity > 0
-    //   similarity = .9 + (40 - mse) / 380  ==  0
-    //   <=>  mse = 40 + .9 * 380 = 382
-    sse = std::min(sse, (382.0 * object_roi.total()));
-
     // Localizing the points that are "good" - not necessarly the absolute min
-    std::vector<Point> mins = minVec(result, sse);
+    double minval;
+    std::vector<Point> mins = minVec(result, minval);
 
     if (mins.empty())
         return outvec;
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -89,8 +89,22 @@ ok(defined $res, "skip needles without png");
 
 $img1   = tinycv::read($data_dir . "console.test.png");
 $needle = needle->new($data_dir . "console.ref.json");
-$res    = $img1->search($needle);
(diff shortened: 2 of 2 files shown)


t/01-test_needle.t
```diff
@@ -214,7 +228,22 @@ ok(!defined $res, "the headline is completely different");
...
+            similarity => '0',
+            x          => 245,
+            w          => 312,
+            result     => 'fail',
+            y          => 219,
+            h          => 36
+
+        }
+    ],
+    'candidate total fail, but not at 0x0'
```
StefanBruens: Sorry, this check is just ...

The old code already returned 0,0 when the best match had an error > INT_MAX, i.e. an arbitrary value, depending on the needle size. Now requiring the method to always return the best match (which hopefully is better than error=INT_MAX), and the best match to be something different to 0,0 is completely bogus.

ppmclibs/tinycv_impl.cc
```diff
@@ -228,24 +229,11 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
     // Perform the matching. Info about algorithm:
     // http://docs.opencv.org/trunk/doc/tutorials/imgproc/histograms/template_matching/template_matching.html
     // http://docs.opencv.org/modules/imgproc/doc/object_detection.html
-    // Used metric is (sum of) squared differences
```
StefanBruens: Comment is still valid ....

ppmclibs/tinycv_impl.cc
```diff
@@ -228,24 +229,11 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
...
     matchTemplate(scene_roi, object_roi, result, CV_TM_SQDIFF);
 
-    // Use error at original location as upper bound
-    Point center = Point(x - scene_x, y - scene_y);
-    double sse = result.at<float>(center);
-    if (sse == 0) {
```
StefanBruens: This case/shortcut always holds ...

ppmclibs/tinycv_impl.cc
```diff
@@ -228,24 +229,11 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
...
-        similarity = 1;
-        return { (int)(x), (int)(y) };
-    }
-
-    // We are only interested in points with a similarity > 0
-    //   similarity = .9 + (40 - mse) / 380  ==  0
-    //   <=>  mse = 40 + .9 * 380 = 382
-    sse = std::min(sse, (382.0 * object_roi.total()));
```
StefanBruens: Now you replace an threshold, which is based on similarity and properly weighted by area by an arbitrary value (INT_MAX).

At least keep the SSE as lower bound, a good match *has* to have an error less than the error at the original location.

ppmclibs/tinycv_impl.cc
```diff
@@ -131,8 +131,9 @@ std::vector<char> str2vec(std::string str_in)
 
 /* we try to the find the best locations - possibly more and will
    weight in later */
-std::vector<Point> minVec(const Mat& m, float min)
+std::vector<Point> minVec(const Mat& m, double& min)
 {
+    min = INT_MAX;
```
StefanBruens: Arbitrary constant.

ppmclibs/tinycv_impl.cc
```diff
@@ -228,24 +229,11 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
...
-    //   similarity = .9 + (40 - mse) / 380  ==  0
-    //   <=>  mse = 40 + .9 * 380 = 382
-    sse = std::min(sse, (382.0 * object_roi.total()));
-
     // Localizing the points that are "good" - not necessarly the absolute min
-    std::vector<Point> mins = minVec(result, sse);
+    double minval;
+    std::vector<Point> mins = minVec(result, minval);
 
     if (mins.empty())
```
StefanBruens: Expanded, this is
if (min_error > INT_MAX)
  return Point(0, 0);

coolo: I just reverted the commit as a whole to avoid the regression - we can roll it back in a changed form

t/01-test_needle.t
coolo: Only for this very needle. I didn't want to add another test case so I picked one of the existant. But yes, I expect the needle to appear at best match. INT_MAX is a constant, true (and was chosen when we still checked the NORM, which was between 0 and 1), but it's still factor 10-100 larger than the largest needle error in the test cases. So in case the needle is huge and the  SQL_DIFF high, 0x0 becomes less of a problem - because huge needles will start close to 0x0 anyway. But the 0x0 is a big problem for other needles.

PR #813, opened 2017-06-09
Convert the needle tags to string when they aren't found

dasantiago:
The input parameter mustmatch can be a scalar or an array ref.
In case of failure to find the need the needles must be converted
to string.

Before:

> 16:08:44.9082 1527 post_fail_hook failed: no candidate needle with tag(s) 'ARRAY(0x37d3128)' matched

at [tragicbox job 528](http://tragicbox.suse.cz/tests/528/file/autoinst-log.txt)

After:

> 17:30:29.1704 5146 post_fail_hook failed: no candidate needle with tag(s) '3' matched

at [tragicbox job 536](http://tragicbox.suse.cz/tests/536/file/autoinst-log.txt)

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -275,7 +275,15 @@ sub _check_backend_response {
             }
         }
         if (!$check && !$rsp->{saveresult}) {
-            OpenQA::Exception::FailedNeedle->throw(error => "no candidate needle with tag(s) '$mustmatch' matched", tags => $mustmatch);
+            # Must match can be only scalar or array ref.
+            my $needletags = $mustmatch;
+            if (ref($mustmatch) eq 'ARRAY') {
+                $needletags = join(', ', @$mustmatch);
+            }
+            OpenQA::Exception::FailedNeedle->throw(
+                error => "no candidate needle with tag(s) '$needletags' matched",
+                tags  => $mustmatch
+            );
         }
         if ($rsp->{saveresult}) {
             $autotest::current_test->save_test_result();
--- a/t/99-full-stack.t
+++ b/t/99-full-stack.t
@@ -41,9 +41,29 @@ open($var, '>', 'live_log');
 close($var);
 system("perl $toplevel_dir/isotovideo -d 2>&1 | tee autoinst-log.txt");
 is(system('grep -q "\d*: EXIT 0" autoinst-log.txt'), 0, 'test executed fine');
-for my $result (glob("testresults/result*.json")) {
+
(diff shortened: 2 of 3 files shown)


testapi.pm
```diff
@@ -275,7 +275,12 @@ sub _check_backend_response {
             }
         }
         if (!$check && !$rsp->{saveresult}) {
-            OpenQA::Exception::FailedNeedle->throw(error => "no candidate needle with tag(s) '$mustmatch' matched", tags => $mustmatch);
+            # Must match can be only scalar or array ref.
+            local $" = ', ';
+            OpenQA::Exception::FailedNeedle->throw(
+                error => "no candidate needle with tag(s) '" . (ref($mustmatch) ? @$mustmatch : $mustmatch) . "' matched",
```
okurz: I think it would be an ever better idea to join all elements when we have an array. https://perldoc.perl.org/functions/join.html should work

coolo: And split the line by assiging a local variable

dasantiago:
Code changed to respect the changes suggested:

test: http://tragicbox.suse.cz/tests/542/file/autoinst-log.txt

> 13:34:55.7212 4044 post_fail_hook failed: no candidate needle with tag(s) 'tty2-selected, text-logged-in-root, text-login' matched

okurz:
great! It would be good to have a test in `t/03-testapi.t`. If possible, add it yourself. If you think that's not feasible we could also live without for now.

testapi.pm
```diff
@@ -275,7 +275,15 @@ sub _check_backend_response {
             }
         }
         if (!$check && !$rsp->{saveresult}) {
-            OpenQA::Exception::FailedNeedle->throw(error => "no candidate needle with tag(s) '$mustmatch' matched", tags => $mustmatch);
+            # Must match can be only scalar or array ref.
+            my $needletags = $mustmatch;
+            if (ref($mustmatch)) {
```
mudler: Not sure about what $mustmatch should be here, but i'd check strictly if it's an [array ref](https://perldoc.perl.org/functions/ref.html) before dereferencing it as an array. `ref(something) eq 'ARRAY'`

dasantiago: Thanks for the input.

That piece of code is either invoked by the check_screen or assert_screen. AFAIK, the tests are using only scalars and arrayrefs.

dasantiago:
Tests added.

okurz:
impressive :-)

PR #814, opened 2017-06-10
Prevent tiny resolution differences to fail the needle test

okurz:
See
https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Factory_ARM/aarch64
for an example.


okurz:
@coolo see https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Factory_ARM/aarch64

coolo:
I was kind of expecting this - I just didn't know in which form :)

PR #816, opened 2017-06-13
Add one more test case for keeping position in fails

coolo:


coolo:
perhaps in this example @StefanBruens can more easily understand the importance of having a fair position for failed needles. For reneedling this is key

PR #817, opened 2017-06-13
Use more robust QMP snapshot save command instead of HMP

okurz:


okurz:
for @foursixnine to continue with, cherry pick the commit or something

okurz:
@coolo @foursixnine I would have expected this change to be incorporated faster to fix annoying https://progress.opensuse.org/issues/19586 . I know it's not complete (only savevm replaced) but it should help nevertheless. Care to merge and deploy it?

coolo:
the message in the PR is 'for @foursixnine to continue with' - why would I merge it, leave alone deploy

foursixnine:
I wouldn't merge either :), since we have a nice mix of different versions of qemu this has to be tested well before deploying.

okurz:
The PR description is actually empty only the second comment mentions "to continue with…". There is a fallback to HMP in case QMP does not work. I tested it locally with qemu-2.6 and it still does what it should and how it did in before.

mnowaksuse:
I actually had the patch deployed to my openQA instance, where I saw the problem sporadically and I noticed it was not effective. Seen it fail once at least. What actually worked was reducing VM's memory from 4 to 2 gigs, then snapshot save worked.

coolo:
I actually dare to claim that any sync call will not help the issue - because the timeout happens between test process and backend.

okurz:
Sure, I don't think it will fix the issue but be a more robust approach meant for machine interfacing. Also qemu developers don't want to help as long as we use HMP here. Also I don't expect it to change any problems on x86_64 but only on aarch64 where we saw the original problem.

PR #819, opened 2017-06-15
New ipmi

coolo:
https://progress.opensuse.org/issues/19758

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -103,8 +103,8 @@ sub do_start_vm {
     open(my $sf, '>', $self->{serialfile});
     close($sf);
 
-    my $sol = $testapi::distri->add_console('sol', 'ipmitool', {serialfile => $self->{serialfile}});
-    $sol->activate;
+    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm');
+    $sol->backend($self);
     return {};
 }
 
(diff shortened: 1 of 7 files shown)


backend/ipmi.pm
```diff
@@ -91,43 +91,21 @@ sub restart_host {
...
-
-    return 1;
-}
-
 sub do_start_vm {
     my ($self) = @_;
 
     # remove backend.crashed
     $self->unlink_crash_file;
+
```
sysrich: NITPICK: Erroneous blank space?

backend/ipmi.pm
```diff
@@ -91,43 +91,21 @@ sub restart_host {
...
     $self->restart_host;
-    $self->relogin_vnc;
 
     # truncate the serial file
     open(my $sf, '>', $self->{serialfile});
     close($sf);
 
-    my $sol = $testapi::distri->add_console('sol', 'ipmi-sol', {serialfile => $self->{serialfile}});
-    $sol->activate;
+    my $sol = $testapi::distri->add_console('sol', 'ipmitool-xterm');
```
sysrich: Why 'ipmitool-xterm' instead of 'ipmi-xterm'? doesn't 'ipmi-sol' also use ipmitool? The weird naming confuses me.

coolo: I used ipmitool initially as name - and then perlcritic complained that I need to leave out the '' in the classname hash. And I thought for 2 seconds about a better name containing a - for consistency :)

coolo:
as discussed with Richard - merging for timely deployment

PR #820, opened 2017-06-15
Jplesnikova: add . to INC for coverage test (perl 5.26 doesn't)

AdamWill:
This patch is from Jitka Plesnikova at Red Hat. The os-autoinst
test suite fails on Perl 5.26.0 because . is no longer in INC
by default, so this test must add it explicitly.

### DIFF
--- a/t/06-pod-coverage.t
+++ b/t/06-pod-coverage.t
@@ -3,6 +3,10 @@
 use strict;
 use warnings;
 
+BEGIN {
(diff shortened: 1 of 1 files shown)


coolo:
I would rather find a way without it - I will enable perl-dev on travis to check

coolo:
ok, pod::coverage is really that dumb.

PR #823, opened 2017-06-21
Rename 'zlre' strings to 'zrle'

mnowaksuse:
Some refrences to RFB's run-length encoding protocol has the string
'zlre' instead of 'zrle'. Now it's unified to 'zrle'.

Those using os-autoinst from git may need to run `make clean && make`
twice (at least I had to).

Tested: http://assam.suse.cz/tests/27.


coolo:
yeah, I mistype that all the time :)

mnowaksuse:
coveralls "coverage" makes no sense to me... What should I do about it?

coolo:
travis deployed new images today - just ignore the drop

PR #824, opened 2017-06-29
[WIP] Connections hijack

mudler:
This work is still in progress, expecially regarding how to expose the functionalities in the testing side. Indeed the last commit (labeled with PoC) it's just an example of how a guest VM can access the two components that will be loaded. Feedbacks are really appreciated.

The components are in a separate namespace `backend::component`, and are lazy loaded on request by using the variables that are responsible to load the components. I tried to take the general approach and look at the components in a "service" perspective, this is why some changes on the backend's constructor were required.

E.g. components can be loaded during the backend instantiation:

	backend::driver->new(
        	backend    => $bmwqemu::vars{BACKEND},
        	components => [qw( foo bar )]); # Load foo and bar

or as well, from other backends

	backend::foo->new(
        	components => [qw( foo bar )]); # Load foo and bar

or, as it is now in the current implementation, components can "ask" to be loaded on start.

If the components have a method/attribute `load()` and the method returns 1, the component gets automatically loaded, even if not specified in the constructor (example https://github.com/mudler/os-autoinst/commit/832edf069862e6c9b101a198b208534cbf198d02#diff-ab6c7a4949ba5ea72135110fddc7f5c0R31)

I tried to keep the changes as much logically separated as possible:

- In https://github.com/mudler/os-autoinst/commit/56295c8496370cb0ef722b2d57df82fbf31eceb0 are present the changes required to the baseclass/backends/isotovideo to have generically loadable components.
- In https://github.com/mudler/os-autoinst/commit/832edf069862e6c9b101a198b208534cbf198d02 there are the two components (proxy and dns), with unit tests
- In https://github.com/mudler/os-autoinst/commit/f597b0c8a0d80dcc6c87b52d60ca25d10c04e1a3 options for qemu backend are generated to be able to redirect connections (semi-transparently) in user mode networking
- https://github.com/mudler/os-autoinst/commit/345a614f5680bfbbbf445938157403ad220e84ef it's just an example that i used to run my tests. Every test that have `CONNECTIONS_HIJACK` set to 1, will have exposed a proxy and a dns server that are reachable on the ports specified in `CONNECTIONS_HIJACK_DNS_SERVER_PORT` and `CONNECTIONS_HIJACK_PROXY_SERVER_PORT` on the testing side.

~Aside, there are some parts of the code that i'd like to get rid off, like the url write generation, that is mostly product-specific, but as for now resides in the proxy component: https://github.com/mudler/os-autoinst/commit/832edf069862e6c9b101a198b208534cbf198d02#diff-ab6c7a4949ba5ea72135110fddc7f5c0R74 .~ (Update: Generation of rewrite rules is now in distri, components have no product-specific code now) This is actually needed to setup automatically redirection from d.o.o. to openqa repo mirrored assets without user intervention, but the same can be accomplished manually, setting the DNS/PROXY entry tables accordingly.
Few changes will follow, ~since i want to move the code responsible for lazy-load of components in `osutils.pm`, so if necessary can be re-used in other parts of the code and not just for backend components~ (done).

wrt the DNS/Proxy components: they have different running options, e.g. the Proxy is able to forward, redirect, and rewrite each http request according to given rules; it can be configured thru `CONNECTIONS_HIJACK_PROXY_POLICY` to change it's behavior, but policies can be applied to single urls directly as well. (will update with more documentation once every aspect of the components are well-defined)  

Running test: [Leap to TW upgrade](http://10.100.51.88/tests/614#live)

Related Progress Issue: https://progress.opensuse.org/issues/17436


backend/baseclass.pm
```diff
@@ -51,7 +53,10 @@ __PACKAGE__->mk_accessors(
 
 sub new {
     my $class = shift;
-    my $self = bless({class => $class}, $class);
+
+    my $self = bless @_ ? @_ > 1 ? {@_} : {%{$_[0]}} : {}, $class;
```
okurz: what is that?

distribution.pm
```diff
@@ -104,6 +104,89 @@ sub become_root {
...
+        $arch = "arm"   if (testapi::check_var('ARCH', 'armhfp') || testapi::check_var('ARCH', 'arm') || testapi::check_var('ARCH', 'armv7'));
+
+        $os = "windows" if (testapi::check_var('DISTRI', 'windows'));
+        $os = "darwin"  if (testapi::check_var('DISTRI', 'macosx'));
+        $os = "openbsd" if (testapi::check_var('DISTRI', 'openbsd'));
+        $os = "freebsd" if (testapi::check_var('DISTRI', 'freebsd'));
+
+
+        # Unfortunately as a fallback, we need an agent that proxies UDP requests from the guest to the host machine
+        # since UDP guestfwd it's not supported, nor iptables is present on LIVECDs
```
okurz: how about installation? Does the same apply there?

backend/component/proxy.pm
```diff
@@ -0,0 +1,254 @@
...
+    };
+
+    $policy = "REDIRECT" if (!$policy && $hosts);
+
+    # Handle SUSEMIRROR and MIRROR_HTTP transparently
+    if ($bmwqemu::vars{MIRROR_HTTP} || $bmwqemu::vars{SUSEMIRROR}) {
+        $self->_diag("Use of mirror detected. Setting up Proxy configuration automatically according to MIRROR_HTTP");
+
+        my $rules = [
+            ($bmwqemu::vars{MIRROR_HTTP} || $bmwqemu::vars{SUSEMIRROR}), "/tumbleweed/repo/.*oss/repodata",
```
coolo: you will have to find a way to move all this into the distribution/tests - this won't go into os-autoinst

backend/baseclass.pm
coolo: good thing you asked first :)

mudler: This is the exactly the same approach used by `Mojo::Base` to bless the new object. Basically it's taking the options from `new()`, even if they are in the array form, or hashed ones.
It makes this calls equivalent:

      myobject->new( foo => "bar", baz => "foobaz")
      myobject->new( { foo => "bar", baz => "foobaz" })

While keeping passing the same arguments to the new blessed object

distribution.pm
mudler: Didn't tested installation yet, but i'm pretty sure will have the same problem.  i'll let you know when as soon as i setup a test for this.

backend/component/proxy.pm
mudler: indeed, i used it to test if the redirection goes as it should, but still have some doubts of the right place where rules should be generated

mudler:
@okurz : yes, it is messing on the networking layer, but few things should be considered aka 'reasons why i personally would keep such changes' :

- The approach is not invasive, what i mean is: if you don't enable them, the components are not loaded, and nothing changes from the backend perspective
- The test developer should still explictly set the machine to redirect connections to the DNS/Proxy components, otherwise networking is not handled (and that's why the new API call in testapi, trying to make it easier).
- It allows to run and perform tests that simulate bad network conditions, where a tester can also test software under domain obfuscations. E.g. you can supply custom answers to guest DNS queries in your tests, or even ask for drop parts of them. Same for http requests
- In future, we could move new components/code parts to a more modular approach

okurz:
> The approach is not invasive, what i mean is: if you don't enable them, the components are not loaded, and nothing changes from the backend perspective

Let me put it this way: Test reviewers will always ask *you* to fix their stupid tests because they can not grasp your code. Replacing the urls of zypper repositories everyone would understand even though it might be a lot of work

> It allows to run and perform tests that simulate bad network conditions, where a tester can also test software under, for example, domain obfuscations. You can supply custom answers to guest machine DNS queries in your tests, or even ask for drop parts of them. Same for http requests

Sure. But I guess if that would be necessary there should be solutions that act on the level of network bridges on the virtualization host, potentially using openvswitch and such.

mudler:
In that perspective - you are absolutely right - it's much easier to replace urls of zypper repositories. But the changes required to testers here in such case (once https://github.com/os-autoinst/os-autoinst/pull/824/commits/91589324e6f454cab46bf799a6db02c9596ba00b#diff-ab6c7a4949ba5ea72135110fddc7f5c0R74 gets a proper place ) will be to just set `CONNECTIONS_HIKJACK = 1` into vars and call `connection_hijack()` (or some equivalent methodology) inside the test code.

honestly i don't see how better can be messing with openvswitch and bridges in such cases, since requires changing states of the host machines (also not to mention we have already pending issues wrt ovs) from os-autoinst, which makes the things hard to test, debug and to maintain.  Having internal components that can be well-tested and documented it's what i'd prefer (and thus what i propose)

sysrich:
@okurz the problem with your approach, and the reason for this feature, is that testers can not always assume they can change the hosts in question

Cases in point:

SCC testing - if SLE testing will require the testing of a module before all of the SCC enablement is in place, a feature like this could be very useful

YaST update testing - The YaST update feature has a hard coded URL for receiving it's updates, with this feature we can start using it.

conncheck - the connection check features in Tumbleweed GNOME go to a specific URL which cannot be modified. With this feature we will have the capability of testing new conncheck functionality before it is deployed into production

And of course, the original usecase that jusified this, download.opensuse.org testing for Tumbleweed, where we do not want to have to change the system under the test.

In short, I would rather have this feature and have the burden of documentation and explanation, than having to hack endless hacks into tests that pollute the environment under test in an artificial way

coolo:
about scc testing: this is https traffic. About the yast update feature: this is coming from scc - and is a https URL. So before you sell this feature for it - please test proxying https

mudler:
https proxying is not supported - yet, there are few direction that could be taken into that, depending also on the SW that needs to be tested ( taking yast as an example in such case ). 
if yast it's pinning the cert, there is not much that we can do, except create a tunnel that is not decrypting the connection (and that would probably need another component, since would look like very different from the current proxy implementation) . Instead if we are not speaking about pinning certificates, we could provide an automatically crafted one, better based on a key that would be installed on the system (to avoid connection errors due to the not-be-able to authenticate certificate).

coolo:
So I summarize: this are almost 2000 lines of code to support exactly one feature: avoid to change the repositories in a Tumbleweed test. It's almost impossible to debug for a test writer, it will invalidate product testing because you never know if the bug is in the product, the server or the proxy.

Aka: I don't want that.

mudler:
It's ~1,900 LOC, but among them ~1000 LOC are for unit testing.
In my perspective (but i could be just wrong) I don't see it for just one feature, it helps to debug network issues in other SWs as well, take [Yast UI freeze](https://bugzilla.suse.com/show_bug.cgi?id=1044254) bug as an example, that now have been triaged in the [relative trello board](https://trello.com/c/Pvclh2O5) - Even if components would mangle the network leading to some "unpredictable" result, that is not necessary a bad thing because helps to highlight code parts that handle connections in the software that have to be tested , actually stressing them. 

What i see more as a useful feature is that helps to provide networking debug as part of the os-autoinst log (especially the DNS component), as i see now testers are used to use wireshark in their local machine to debug (hence they can't get the same kinda of information by remote builds): 

       07:42:13.3787 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 0.opensuse.pool.ntp.org
       07:42:13.3790 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 0.opensuse.pool.ntp.org
       07:42:13.3795 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 0.opensuse.pool.ntp.org
       07:42:13.3799 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 0.opensuse.pool.ntp.org
       07:42:13.3804 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 1.opensuse.pool.ntp.org
       07:42:13.3807 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 1.opensuse.pool.ntp.org
       07:42:13.3812 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 1.opensuse.pool.ntp.org
       07:42:13.3815 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 1.opensuse.pool.ntp.org
       07:42:13.3820 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 2.opensuse.pool.ntp.org
       07:42:13.3824 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 2.opensuse.pool.ntp.org
       07:42:13.3828 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 2.opensuse.pool.ntp.org
       07:42:13.3831 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 2.opensuse.pool.ntp.org
       07:42:13.3837 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 3.opensuse.pool.ntp.org
       07:42:13.3841 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 3.opensuse.pool.ntp.org
       07:42:13.3845 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 3.opensuse.pool.ntp.org
       07:42:13.3849 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for 3.opensuse.pool.ntp.org

or, as another example:

       07:58:14.9660 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for download.opensuse.org
       07:58:14.9662 21832 >> backend::component::dnsserver::__ANON__(): Answer download.opensuse.org.	IN	A	10.0.2.2
       07:58:14.9665 21832 >> backend::component::dnsserver::__ANON__(): Intercepting request for download.opensuse.org
       07:58:14.9699 21838 >> backend::component::proxy::_handle_request(): Request from: 127.0.0.1 method: GET to host: download.opensuse.org
       07:58:14.9701 21838 >> backend::component::proxy::_handle_request(): Requested url is: http://download.opensuse.org/distribution/leap/42.2/repo/non-oss/media.1/media
       07:58:14.9705 21838 >> backend::component::proxy::_build_tx(): Redirecting to: http://openqa.opensuse.org/assets/repo/openSUSE-Tumbleweed-oss-i586-x86_64-Snapshot20170625/media.1/media
       07:58:15.0527 21838 >> backend::component::proxy::_handle_request(): Request from: 127.0.0.1 method: GET to host: download.opensuse.org
       07:58:15.0529 21838 >> backend::component::proxy::_handle_request(): Requested url is: http://download.opensuse.org/distribution/leap/42.2/repo/non-oss/suse/noarch/yast2-nfs-client-3.1.17-1.2.noarch.rpm
       07:58:15.0533 21838 >> backend::component::proxy::_build_tx(): Redirecting to: http://openqa.opensuse.org/assets/repo/openSUSE-Tumbleweed-oss-i586-x86_64-Snapshot20170625/suse/noarch/yast2-nfs-client-3.1.17-1.2.noarch.rpm

In this way the tester can also obtain useful insights of what the machine is actually doing, in case they need more debug informations, and i'd guess sometimes would avoid them to stop and debug trivial networking issues locally.

wrt SCC, AFAICT from [lib/registration.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/registration.pm#L61) it is already accepting fake certs, so it could be actually possible to emulate that behavior as well.

coolo:
https://bugzilla.suse.com/show_bug.cgi?id=1044254#c1 nicely highlights my point. We have enough 'it works here, it must be openqa' discussions - so KISS is a key benefit.

Debugging or stressing are interesting use cases, but I don't see the value of having this system interception as part of isotovideo - even less do I see a value of having openSUSE specific routes in isotovideo.

When I heared you work on a generic proxy, I thought you did exactly that: a generic proxy. I.e. one that works independent of os-autoinst, can be started from the tests, configured from the distribution to intercept routes and provides a log file.

mudler:
As i said already - specific routes of openSUSE are meant to go away from the code (this is why it's WIP), but probably i wasn't clear enough. I opened the PR actually to have a feedback  - and a direction of where the work should go.

The Poo is related to the backend, not from the tests side - this is why the solution is in os-autoinst and not in tests; Anyway i don't see a good reason to develop a stand-alone Proxy - there are already quite a lot available in the FOSS world (Squid and HAProxy to name some).

sysrich:
After discussing with coolo and thinking about it, I'm starting to share the fears that this approach is too invasive.

Lets stick to a simple, single, user story - download.opensuse.org - what I want, is a simple solution that gives the VM SUT an environment within which instead of going to download.opensuse.org goes to the equivalent location on openqa.opensuse.org/assets/repo

To accomplish this the way I want it needs to be
1) done without changes within the SUT - if we can change the SUT, we might as well implement all of this into tests and not take this issue as a backend task
2) transparent to the SUT
3) simple, reliable, easy to debug as to avoid it becoming a potential cause of blame and INVALID bugs

This PR mostly accomplishes 2) but not 1) or 3)

When we originally talked about this, we bounced around ideas about using dnsmasq as a starting point of this solution, which would accomplish 1) and 2) for sure and most likely satisfy 3) also

That would only leave URL rewriting, which could be handled by a simple apache proxy.

However, openQA is using usermode networking by default so we do not have dnsmasq easily to hand.

I'd like to revisit that approach though. 

I think we have a potential better solution sitting with tying together our existing qemu usermode networking, with vde which should facilitate the use of dnsmasq.

Something along the lines of: http://alien.slackbook.org/dokuwiki/doku.php?id=slackware:vde#qemu_vde_and_dnsmasq

A solution along those lines should mean we could (selectively) create an environment that would be provided alternative DNS responses for specific domains (such as download.opensuse.org) which we could then forward to a simple rewrite proxy.

An alternative approach could be to selectively not use usermode networking for all tests, but I don't like that idea because I dislike the idea of openQA VM's running as root.

Really, it's a red-line to me when we start typing specific commands in the SUT to facilitate this feature - if we can't provide a solution without resorting to that, we might as well not do it and leave it to the tests to handle that complicated mess.

mudler:
That's what i meant [here](https://progress.opensuse.org/issues/17436#note-24) with "it's getting more as a big hack from the backend side" :)

IMHO i don't see a solution that fully supports qemu user mode and does not involve typing commands on the SUT. 
The "perfect"/cleaner solution could have been accomplished with qemu parameters, but sadly, the guestfwd option, that is available under qemu user mode supports just tcp streams - i've tried also redirecting everything to a chardev pointing to an udp endpoint, but had any luck with that (i saw no data coming from the guest to the host).

Unfortunately even the dnsmasq's instance used in qemu user mode can't be customized too much, the upstream reason (QEMU) is that would make it an unpredictable network, and thus leading also the machine to not be able to communicate with outside.

From the link you posted i can see the solution is involving using vde as a network model, which AFAIK won't work with qemu user mode together unless (as openQA default) we would create another network device on top of the one used in user mode to be able to serve the DNS requests (i think the same could be accomplished in tap mode as well). But still i see this solution a bit more invasive since would involve anyway adapting tests to generate the url-rewrite policies for the proxy (if we want to satisfy 2), an do it transaprently ) , and os-autoinst should be aware about which domains should be redirected, leaving then 1), and possibly even 2) unsatisfied.

Looking at the problem entirely from the test side, it's maybe easier to spawn a local DNS and proxy server and manage the rules from there, but at this point it won't involve changes in the backend side.

mudler:
@sysrich  rethinking on it, had some clue from your comment and the posted link that could lead to a less-invasive solution, that solves the problem of typing commands on the SUT - But with some caveats, and would need some time to work on it (still not sure if a solution like this will satisfy all the points you stated above). 

I'd go with creating a supplementary network device, handled with vde (or similar) when `CONNECTIONS_HIJACK` is set to 1, trying to force DNS Server from there. 
This removes the requirement of typing commands from the SUT,  but the caveat is that the rewrite rules for the urls will be supplied inside the tests less or more explictly, and would require some interaction with the proxy/dns components started in the backend (if we want to keep them still there, we would want them to handle e.g.  restart, stop, start and reload of rules).

But i see also some other drawback derived from this approach, depending if components are supplied internally, or not:

if not included, we need some service (similar to the D-Bus one we use for commanding openvswitch options) that drives external tools (like an apache server or dnsmasq) as well, not sure if we want this, because it makes it harder to test and debug. 
Alternatively, os-autoinst could directly start a separate/controlled instance of dmasq and apache server, but it would be less KISS in my perspective, since we will start requiring dnsmasq installed, and makes it difficult to test.

If components are instead kept internal, the current implementation needs to be changed, allowing those services to bind to a particular interface, and not just a port: in the current implementation, DNS servers are spawned in different ports than default 53 (to avoid collisions) , leading to "hacks" while supplying the DNS server to the SUT - on the other side - binding on interface makes everything easier, because then we could start the DNS server on the standard port, requiring less hacks.

mudler:
About SCC: just discovered https://github.com/SUSE/scc-qa-proxy - afaics it seems SCC can be used also over http, so this PR could possibly superseed the scc-qa-proxy approach.  scc-qa-proxy contains specific code about the products and other caveats, while with internal components the rules can be generated entirely from the tests side. 

Update - wrt recent changes: The *yet* remaining downside as for now it's the requirement of the SUT to redirect the dns and http traffic over the assigned components, the product-specific code was completely removed and now it's generated from distri.
Currently investigating solutions to avoid the SUT to redirect the connections with an additional network interface. 

small side-note: `backend::component::process`, at this stage, aside from the purpose of the dns/proxy components could be re-used to refactor some code parts

coolo:
So you want to replace a generic solution with a qemu-specific solution? That leaves s390 testing and others in the cold rain.

PR #825, opened 2017-07-04
Unconditionally set serial id on qemu drives (bsc#1041692)

okurz:
Normally one would expect a serial ID on each HDD drive. Previously we set
this only for NVME devices and not at all for other devices. To fix
bsc#1041692 we need to explicitly set it and it is making the case easier if
we just apply this in all cases. Probably we will not have a problem with that
in all currently used qemu versions.

Verification run: https://openqa.suse.de/tests/1044039


okurz:
Also checked locally a bit more, e.g. non-x86_64, so far no problems observed

okurz:
For now hotpatched our osd ppc64le workers malbec.arch and qa-power8-5-kvm with

```
ssh malbec.arch "sudo sed -i 's/my \$serial = .*nvme.*$/my \$serial = \"serial=\$i\";/' /usr/lib/os-autoinst/backend/qemu.pm"
```

PR #826, opened 2017-07-05
Add a new control variable to full stack tests

foursixnine:
Use INTEGRATION_TESTS to toggle the assert_screen_failt_test so that
when used by the full stack test of openQA, it can be enabled on demand.

This addresses the problems introduced by 126bf5ac on the openQA side.


foursixnine:
Boot test could fail depending on the load of the host :trollface:. And no... I'm not kidding... I'm guessing qemu is losing key events here even... lol. fun.

Boot test fails, due to the md5 sum mismatching, because all the text didn't match :D

coolo:
that's the whole purpose of this typing a lot test - to find the condition it fails under :)

PR #828, opened 2017-07-14
Free allocated vlan tags from tap devices

mudler:
the changes export a new dbus method to remove tags from tap devices, this is to leave the machine in a more "predictable" state. 

side note: it is freeing vlan tags just to the jobs that are stopped, not the killed one

Related Progress Issue: https://progress.opensuse.org/issues/19806


mudler:
A slight little change wrt ovs, now `OVS_DEBUG` controls the ovs show output

PR #829, opened 2017-07-18
[WIP] Process module and small refactorizations

mudler:
This is an extract from https://github.com/os-autoinst/os-autoinst/pull/824

Most of the LOC additions falls into the unit test of the `process` module (the module itself is ~200LOC).
Commits were separated to make it easier to pick up or discard one-by-one, to summarize: 

Not required (completely optional):
* [osutils functions to check for a valid ip address](https://github.com/mudler/os-autoinst/commit/1caa4b3e6a1cd86e91ea3659d6edd13855dcfaef)
* [osutils functions to lazy-load perl modules](https://github.com/mudler/os-autoinst/commit/c0db24f0c047dc2fb416c9cbac6252a05da76f33)
* [keep the constructor data among backends](https://github.com/mudler/os-autoinst/commit/dc34e72f92c6f938729b042312189c4f2f48375a) - it was just required for the components, now not anymore

Nitpicks/refactorizations:
* [backend::qemu nitpicks](https://github.com/mudler/os-autoinst/commit/cd87dfd0a03c329864fe2febfb9a0e0cf9c00c97)
* [backend::baseclass should also reap forked processes](https://github.com/mudler/os-autoinst/commit/f702b0d2c37d54e289afeb68e2b43c19b189e0a9)
* [use Mojo::Loader to load the backends in backend::driver and strictly check for failures](https://github.com/mudler/os-autoinst/commit/9a1c17b0ea864e420f418bd7a5fe27928360140c)
* [process module](https://github.com/mudler/os-autoinst/commit/37801dd807a914ee06e4b3ee2f07af409a4652ba) - most of the LOC additions falls into the unit test of the component

wrt [process module](https://github.com/mudler/os-autoinst/commit/37801dd807a914ee06e4b3ee2f07af409a4652ba) the intent was to create a common module to wrap both forked processes and external program execution in an uniform and convenient way (this module was used by both dns and proxy components).

The module exposes communication channels between the main process execution and the child, the child can be created from fork/system, but the way of fetching the data remains the same. It intercepts processes STDERR/STDIN/STDOUT by default (can be disabled) and in case of a forked process it exposes 2 additional channels to communicate with it (sorry for the indentation, but it's messy within github):

```perl
# Fork process:
my $process = process->new(sub { print "Hello\n" });
$process->start();
my $is_running = $process->is_running();
my $hello = $process->getline(); # just a convenient function around the handler
my $pid = $process->pid();
$process->stop();
             
#  $hello is "Hello\n"
# $pid will contain the process id
# $is_running will be 1 if the process is still in execution when called
```
So this works as well:
```perl
my $p =  process->new(sub { print "Hello\n" })->start()->stop();
$output = $p->getline();
```
Same for external process:

```perl
my $process = process->new(execute=> '/path/to/bin' );
$process->start();
my $hello = $process->getline();
my $pid = $process->pid();
$process->stop();
```
Errors are stored in the object:

```perl
my @errors = $process->error;
```
As well as the exit status (external process) or the result (if it's a function to be forked):
```perl
$process = process->new(
               separate_err => 0, # merge STDERR to STDOUT
               code => sub {
                              my ($self) = shift;

                              # Access to the channels
                              my $parent_output = $self->channel_out; 
                              my $parent_input  = $self->channel_in;

                              print "TEST normal print\n";
                              print STDERR "TEST error print\n";
                                                            
                              $self->channel_out->write("PING?");

                              return "256";
               })->start();
$process->stop unless $process->is_running; # We need to stop it to retrieve the exit status
my $return = $process->return_status;
# my $return = $process->exit_status; # equivalent
# $return is 256
```

Still we can access directly to handlers from the object:

```perl
my $stdout = $process->read_stream;
my $stdin = $process->write_stream;
my $stderr = $process->error_stream;
# So this works:
print $stdin "foo bar\n";
my @lines = <$stdout>;
```

There is also an alternative channel of communication (just for forked processes):

```perl
my $channel_in = $process->channel_in; # write to the child process
my $channel_out = $process->channel_out; # read from the child process

$process->channel_write("PING"); # convenience function
```

The process can be debugged with an environment variable `OSAUTOINST_PROCESS_DEBUG`. When setted to 1, it is possible to inspect all the spawned processes or codes, and will dump the process path+arguments or the code if the latter(to note, it will be printed the code as the perl backend **sees** it, not as we **wrote** it) e.g.:

                        Fork: {
                        use warnings;
                        use strict;
                        no feature ':all';
                        use feature ':5.16';
                        print STDERR "foobar\n" while 1;
                        }

while the code was `print STDERR "foobar\n" while 1;`

or in case of processes:

                        Execute: '/my/awesome/bin', 'FOO', 'BAZ'

To note, there is no buffer around the handlers, this is to keep things simple and more hackable: but this also means they must be used wisely when reading from them.

The process module started as a base for the components, but i think it could be used to refactorize part of the os-autoinst code that deals with process communication, giving more coverage and component re-usability - it will ease the debug (since we can turn it on/off) and would ease situations like https://progress.opensuse.org/issues/6958 .

If we decide to keep it, then i'd start to move step-by-step process spawned in os-autoinst with the new module, and document it deeply (there are also other functionalities and aliases)


coolo:
rebase and unWIP?

process.pm
```diff
@@ -0,0 +1,311 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package process;
```
richiejp: Seeing as you are using Mojo::Base, shouldn't this be OpenQA::Something::Process?

richiejp:
Looks like I maybe repeated some work done here when rewriting the QEMU backend. But that is not really an issue.

There is some other state which is closely related to the QEMU process, such as the QMP, VIRTIO_CONSOLE and VNC sockets. These are kept by various different classes which is probably correct. However we should probably be using an event-subscriber model so that these other classes don't have to be individually called when QEMU starts or stops. Currently whenever an event happens, like when a new test starts or we load a snapshot. Every component which is effected by that event has to be handled in a non standard way.

I see this is using EventEmitter so maybe this can be used for a few other things as well?

process.pm
mudler: Well, it's not a direct consequence, but sounds somewhat saner to slowly move module namespace's under OpenQA:: (as lowercase modules on first level of namespace are not reccomended, as are reserved to Perl)

mudler:
This is almost a quite old PR, it would require some work to readapt it now - but basically this would be https://progress.opensuse.org/issues/32293 . 

@richiejp  i agree with you, an observer pattern seems the way to go. The process.pm module now has been refactored (and generalized it furthermore) to [Mojo::IOLoop::ReadWriteProcess](https://metacpan.org/pod/Mojo::IOLoop::ReadWriteProcess) so potentially a class that represent a process could also simply inherit it

PR #830, opened 2017-07-21
Do not output a warning for only OCR areas

okurz:
This prevents warnings like

```
/var/lib/openqa/share/tests/opensuse/products/opensuse/needles/XFCE-desktop-20170522.json missing match area
/var/lib/openqa/share/tests/opensuse/products/opensuse/needles/yast2_tftp-server_configuration_newdir_typed-20170706.json missing match area
```

for two needles that only include OCR or exclude areas which should be
considered as valid.

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -83,7 +83,7 @@ sub new {
         $a->{type}   = $area->{type}   || 'match';
         $a->{margin} = $area->{margin} || 50;
 
-        $gotmatch = 1 if $a->{type} eq 'match';
+        $gotmatch = 1 if $a->{type} =~ /match|ocr/;
 
         $self->{area} ||= [];
         push @{$self->{area}}, $a;


coolo:
we use ocr on xfce-desktop?

okurz:
Both these needles never matched according to https://openqa.opensuse.org/admin/needles and I am going to delete them but that does not make them invalid per se, does it?

coolo:
and I wonder how this works with ocr needles:
    assert_screen 'yast2_tftp-server_configuration_newdir_typed';

okurz:
well, it does not. I don't know what @DimStar77 and @nilxam intended with these needles. Guess the just wanted to test *us* :-)

coolo:
these 2 needles are very much invalid. I never have seen a ocr test - and possibly we should just remove the support left

okurz:
@AdamWill had some experience with OCR. Adam, did it ever work for you? If not I would be fine to remove OCR support completely

AdamWill:
Well, I ditched that attempt after a test where it matched on a screen where it shouldn't match. But I didn't dig into it in any detail, I just went for an alternative approach...

okurz:
hm, does not convince me to fully delete it (although I just did that in a local patch to see the impact) but I rather want on of our interns to check it out see how useable it could be

coolo:
anyway, for the time being I disagree with shutting off the warning - as both us and @AdamWill have no pure ocr.

okurz:
what do you mean "both of us … have no pure ocr"? So you want to keep the warning and delete all needles which have only OCR because you think it does not make sense or can not work?

coolo:
as I said: I rather remove half baken and hardly tested ocr support - as such I keep the warning

PR #831, opened 2017-07-25
dos2unix - the prototype removal was still on my disk

coolo:


okurz:
LGTM, except for the PR subject line

PR #832, opened 2017-07-26
Do not try to take QEMU snapshots with virtio-gpu

Vogtinator:
virtio-gpu does not appear to support serialization yet.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -180,7 +180,7 @@ sub can_handle {
         return if $vars->{HDDFORMAT} eq 'raw';
 
         # virtio-gpu does not support saving yet
-        return if defined?($vars->{QEMUVGA}) && $vars->{QEMUVGA} eq 'virtio';
+        return if defined $vars->{QEMUVGA} && $vars->{QEMUVGA} eq 'virtio';
 
         # XXX: Temporary (hopefully) workaround for nvme, since snapshots fails.
         # See also https://github.com/os-autoinst/os-autoinst/pull/781
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -180,7 +180,7 @@ sub can_handle {
         return if $vars->{HDDFORMAT} eq 'raw';
 
         # virtio-gpu does not support saving yet
-        return if defined $vars->{QEMUVGA} && $vars->{QEMUVGA} eq 'virtio';
+        return if ($vars->{QEMUVGA} || '') eq 'virtio';
 
         # XXX: Temporary (hopefully) workaround for nvme, since snapshots fails.
         # See also https://github.com/os-autoinst/os-autoinst/pull/781
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -180,7 +180,7 @@ sub can_handle {
         return if $vars->{HDDFORMAT} eq 'raw';
 
         # virtio-gpu does not support saving yet
-        return if $vars->{QEMUVGA} eq 'virtio';
+        return if defined?($vars->{QEMUVGA}) && $vars->{QEMUVGA} eq 'virtio';
 
         # XXX: Temporary (hopefully) workaround for nvme, since snapshots fails.
         # See also https://github.com/os-autoinst/os-autoinst/pull/781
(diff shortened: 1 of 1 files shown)


foursixnine:
Only the aarch64 workers will be hit by this, but on the other hand it would render test developers unable to use test snapshots to speedup development of test which is a trade off

Vogtinator:
> Only the aarch64 workers will be hit by this, but on the other hand it would render test developers unable to use test snapshots to speedup development of test which is a trade off

This should not have any negative impact at all, as currently using QEMUVGA=virtio just causes every test to fail due to savevm aborting:

```
State blocked by non-migratable device '0000:00:02.0/virtio-gpu'
DIE Could not save snapshot 'lastgood': savevm lastgood
State blocked by non-migratable device '0000:00:02.0/virtio-gpu' at /usr/lib/os-autoinst/backend/qemu.pm line 241.
```

Or this is not the case on aarch64?

foursixnine:
That sounds like a kind of old version of qemu to me... Looks like you're being hit by bsc#1025025

Vogtinator:
> Looks like you're being hit by bsc#1025025

Indeed, that's probably it. I assume that's available on o3 as well?

foursixnine:
Yes, all of the aarch64 workers are using Qemu 2.9.x

Vogtinator:
This is mainly about x86_64, I guess that's the case for those workers as well?

foursixnine:
Indeed, they are running 2.6.2 on x86_64

okurz:
LGTM anyway

backend/qemu.pm
```diff
@@ -175,9 +175,17 @@ sub can_handle {
     my ($self, $args) = @_;
     my $vars = \%bmwqemu::vars;
 
-    # XXX: Temporary (hopefully) workaround for nvme, since snapshots fails.
-    # See also https://github.com/os-autoinst/os-autoinst/pull/781
-    if ($args->{function} eq 'snapshots' && $vars->{HDDFORMAT} ne 'raw' && $vars->{HDDMODEL} ne 'nvme') {
+    if ($args->{function} eq 'snapshots') {
+        # raw HDD format does not support snapshots
+        return unless $vars->{HDDFORMAT} ne 'raw';
```
coolo: Avoid double negation

Vogtinator: Done.

backend/qemu.pm
```diff
@@ -175,9 +175,17 @@ sub can_handle {
...
 
-    # XXX: Temporary (hopefully) workaround for nvme, since snapshots fails.
-    # See also https://github.com/os-autoinst/os-autoinst/pull/781
-    if ($args->{function} eq 'snapshots' && $vars->{HDDFORMAT} ne 'raw' && $vars->{HDDMODEL} ne 'nvme') {
+    if ($args->{function} eq 'snapshots') {
+        # raw HDD format does not support snapshots
+        return if $vars->{HDDFORMAT} eq 'raw';
+
+        # virtio-gpu does not support saving yet
+        return if $vars->{QEMUVGA} eq 'virtio';
```
coolo: if I read the code right QEMUVGA might be undef - this will give you a warning then.

Vogtinator: Fixed. I did not test my fix though.

backend/qemu.pm
```diff
@@ -180,7 +180,7 @@ sub can_handle {
         return if $vars->{HDDFORMAT} eq 'raw';
 
         # virtio-gpu does not support saving yet
-        return if $vars->{QEMUVGA} eq 'virtio';
+        return if defined?($vars->{QEMUVGA}) && $vars->{QEMUVGA} eq 'virtio';
```
okurz: that looks weird. It's intended to look like `defined?(…)`?

Vogtinator: Whoops, I totally mixed up ruby and perl there.

coolo: I prefer 
```
return if ($vars->{QEMUVGA} || '') eq 'virtio';
```

coolo:
So now please squash.

Vogtinator:
Can't you do that with the GitHub PR merge button directly?

coolo:
never tried - I only squash if there is just one commit. But I'll try :)

PR #836, opened 2017-08-11
Add option to disable qemu snapshots

foursixnine:
Due to the qemu hmp interface not being able to report status and the
combination with slow disks, we might end up having problems when a
system has been trying to save a snapshot but it takes way too long,
having the possibility to disable snapshots on slow workers comes in
handy for such situations

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -186,6 +186,8 @@ sub can_handle {
         # See also https://github.com/os-autoinst/os-autoinst/pull/781
         return if $vars->{HDDMODEL} eq 'nvme';
 
+        return if $vars->{QEMU_DISABLE_SNAPSHOTS};
+
         return {ret => 1};
     }
     return;
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -56,6 +56,7 @@ NICVLAN;integer;undef;network (vlan) number to which the NIC should be connected
 NUMDISKS;integer;1;Number of disks to be created and attached to VM
 OFW;;;
 QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64;boolean;undef;If set, for aarch64 systems use VGA as video adapter
+QEMU_DISABLE_SNAPSHOTS;boolean;undef;If set, disable snapshots in case the worker has slow disks to avoid save_vm calls failing due to timeouts (See [bsc#1035453](https://bugzilla.suse.com/show_bug.cgi?id=1035453))
 PATHCNT;integer;2;Number of paths in MULTIPATH scenario
 PXEBOOT;boolean;0;Boot VM from network
 QEMU;QEMU binary filename;undef;Filename of QEMU binary to use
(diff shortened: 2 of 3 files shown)


foursixnine:
My bad :) Fix is on it's way

okurz:
LGTM

AdamWill:
LGTM, would be helpful for me as we're having issues with snapshots on our new ppc64 workers (much like you're apparently seeing on aarch64).

sysrich:
@AdamWill out of interest - what kind of ppc64le workers are you running? have you correctly taking into account subcore allocation? we found the VM's & hosts in general performed a lot better when we took that into consideration, basically meaning using QEMUTHREADS=2 and QEMUCPUS=2 to make sure we're not hogging a whole core+8 hyperthreads with a VM wanting 1 virtual CPU (which, weirdly, is the behaviour on power)

AdamWill:
@sysrich here's everything I know about this box (it was donated by IBM):

    [root@openqa-ppc64le-01 adamwill][PROD]# cat /proc/cpuinfo
    processor	: 0
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 8
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 16
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 24
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 32
    cpu		: POWER8 (raw), altivec supported
    clock		: 2161.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 40
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 48
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 56
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 64
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    processor	: 72
    cpu		: POWER8 (raw), altivec supported
    clock		: 2061.000000MHz
    revision	: 2.0 (pvr 004d 0200)

    timebase	: 512000000
    platform	: PowerNV
    model		: 8348-21C        
    machine		: PowerNV 8348-21C        
    firmware	: OPAL
    MMU		: Hash

I have not tweaked `QEMUTHREADS` or `QEMUCPUS` at all, but the performance of the guests in general seems perfectly fine; it is *specifically* snapshot creation that seems to constantly take far, far, far longer than it should, and time out (I have actually bumped the os-autoinst timeout to 15 minutes and it still fails). I also tested downgrading qemu to 2.7.1 and it suddenly started working just fine. So I'm pretty happy to claim this is a bug in qemu for now, the trick of course is figuring out *what* bug...

PR #837, opened 2017-08-15
svirt: Extend time before providing password

mnowaksuse:
In 50 % of cases it takes more that 3 seconds to connect to Hyper-V, and
system gets stuck - not being able to connect.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -76,7 +76,7 @@ sub activate {
         die "cant' start xterm on $display (err: $! retval: $?)";
     }
     # FIXME: assert_screen('xterm_password');
-    sleep 3;
+    sleep 10;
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -76,7 +76,7 @@ sub activate {
         die "cant' start xterm on $display (err: $! retval: $?)";
     }
     # FIXME: assert_screen('xterm_password');
-    sleep 3;
+    sleep 10;
```
okurz: Isn't there anything we can assert before typing the password? E.g. output on a simulated serial console?

mnowaksuse:
Well, after a lot of investigation and working on workaround, I realized it's the Windows OpenSSH server's port is to blame. Updated it to the latest release and it's fixed now. Closing :).

PR #838, opened 2017-08-18
Get rid of wait_idle

okurz:
Related progress issue: https://progress.opensuse.org/issues/5830

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -998,35 +998,6 @@ sub check_socket {
     return $self->SUPER::check_socket($fh);
 }
 
-sub wait_idle {
-    my ($self, $args) = @_;
-    my $timeout       = $args->{timeout};
-    my $idlethreshold = $args->{threshold};
-    my $prev;
-    my $timesidle       = 0;
-    my $timesidleneeded = 1;
-
-    for my $n (1 .. $timeout) {
-        my ($stat, $systemstat) = @{$self->cpu_stat()};
-        $self->run_capture_loop(1);
-        next unless $stat;
-        $stat += $systemstat;
-        if ($prev) {
-            my $diff = $stat - $prev;
-            bmwqemu::diag("wait_idle $timesidle d=$diff");
-            if ($diff < $idlethreshold) {
-                if (++$timesidle > $timesidleneeded) {    # idle for $x sec
-                                                          #if($diff<2000000) # idle for one sec
-                    return {idle => 1};
-                }
-            }
-            else { $timesidle = 0 }
-        }
-        $prev = $stat;
-    }
-    return;
-}
-
 sub freeze_vm {
     my ($self) = @_;
     # qemu specific - all other backends will crash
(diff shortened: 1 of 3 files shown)


mnowaksuse:
This is an API change. I don't thing we want it.

okurz:
I understood we do want it. More than a year ago we added an ugly backtrace to each call of wait_idle to mark it as deprecated. As os-autoinst-distri-opensuse now forbids use of `wait_idle` I guess this is the way to go. Only open point is if we want to have it now or a bit later.

sysrich:
While I agree with the recommendation not to use wait_idle, I accept we live in a real world and within it sometimes people need to do things that are not recommended

wait_idle is one of those cases

life is not binary. Just because something is bad, doesn't mean it doesn't have a right to exist.

-1; Because wait_idle is useful for those times when it doesn't cause problems.

okurz:
@sysrich there is now even a rule in os-autoinst-distri-opensuse preventing to use `wait_idle` at all and we have https://progress.opensuse.org/issues/5830 that states we want to get rid of `wait_idle`.
This is also why @coolo in commit 7e44b202 added a `cluck` to track all uses of `wait_idle`. So you want to
1) go forward and clean it up properly
1) keep the removal in an eternal limbo state
2) rollback changes and go forward implementing wait_idle properly for all backends where I don't have any idea how that would work reliably

sysrich:
@okurz and you've checked with @AdamWill that fedora are not using the API call?
Or how about KDE neon?
Or the YaST team?

I am not comfortable with being cavalier about removing something which we, as an upstream, have a responsibility to provide until we're sure it's at least reasonably painless for our wider community

Your rules for your test repo are of zero relevance to anyone else, openQA is not the sole domain of SUSE QA SLE.

AdamWill:
[adamw@adam os-autoinst-distri-fedora (staging)]$ grep -R wait_idle *
    lib/fedoradistribution.pm:use testapi qw(send_key type_string wait_idle assert_screen);
    lib/fedoradistribution.pm:    wait_idle 5; # because of KDE dialog - SUSE guys are doing the same!
    lib/packagetest.pm:    wait_idle 20;
    tests/disk_guided_encrypted.pm:    wait_idle 5;
    tests/desktop_browser.pm:    wait_idle 5;
    tests/desktop_browser.pm:    wait_idle 3;
    tests/desktop_browser.pm:    wait_idle 2;
    tests/desktop_browser.pm:    wait_idle 2;

AdamWill:
our distri is in pagure: https://pagure.io/fedora-qa/os-autoinst-distri-fedora . pagure works a lot like github. if someone wants to send a PR to replace those `wait_idle` calls, we'll certainly look at it. doing it myself would not be my highest priority right now.

okurz:
I did not think about KDE neon, true. Do you know where their repo would be?

sysrich:
@okurz I don't know..ask jriddell?

coolo:
Suggested compromise: replace wait_idle with a sleep 19 as we do for !qemu backends and leave the carp.

okurz:
agreed, updated accordingly.

PR #844, opened 2017-08-25
Add control over resulted filename in upload_logs

asmorodskyi:
I need ability to upload same file several times in my opinion this is most convenient way to do this.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1791,14 +1791,14 @@ sub data_url($) {
 
 =for stopwords GiB failok OpenQA WebUI
 
-  upload_logs($file [, failok => 0, timeout => 90 ]);
+  upload_logs($file [, failok => 0, timeout => 90, log_name => "custom_name.log" ]);
 
 Upload C<$file> to OpenQA WebUI as a log file and
 return the uploaded file name. If failok is not set, a failed upload or
 timeout will cause the test to die. Failed uploads happen if the file does not
 exist or is over 20 GiB in size, so failok is useful when you just want
 to upload the file if it exists but not mind if it doesn't. Default
-timeout is 90s.
+timeout is 90s. C<log_name> parameter allow to control resulted job's attachment name.
 
 =cut
 
@@ -1810,7 +1810,7 @@ sub upload_logs {
 
     bmwqemu::log_call(file => $file);
     my $basename = basename($file);
-    my $upname   = ref($autotest::current_test) . '-' . $basename;
+    my $upname   = ($args{log_name} || ref($autotest::current_test)) . '-' . $basename;
     my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
     $cmd .= autoinst_url("/uploadlog/$basename");
     if ($failok) {


testapi.pm
```diff
@@ -1810,7 +1810,7 @@ sub upload_logs {
 
     bmwqemu::log_call(file => $file);
     my $basename = basename($file);
-    my $upname   = ref($autotest::current_test) . '-' . $basename;
+    my $upname   = $args{upname} || ref($autotest::current_test) . '-' . $basename;
```
okurz: better use `//` instead of `||` as you want to check if the parameter is appied at all

testapi.pm
```diff
@@ -1791,7 +1791,7 @@ sub data_url($) {
 
 =for stopwords GiB failok OpenQA WebUI
 
-  upload_logs($file [, failok => 0, timeout => 90 ]);
+  upload_logs($file [, failok => 0, timeout => 90, upname => "custom_name" ]);
 
 Upload C<$file> to OpenQA WebUI as a log file and
 return the uploaded file name. If failok is not set, a failed upload or
```
okurz: please add an explanation of `upname` in this section as well then

okurz:
can you give an example where you want this? E.g. a failing job that overwrites an existing file

testapi.pm
coolo: upname is a stupid paramater name.  what's wrong with using real words?

asmorodskyi: probably , but it is not me who added this name https://github.com/asmorodskyi/os-autoinst/commit/c490d2c721f0e21a101f27ed7a671b0b0cc33961 I just use whatever was before me. But if you have better ideas I open to change to anything what you think suites better

asmorodskyi: @okurz - explanation added

testapi.pm
asmorodskyi: just discussed with @okurz that suggestion is not relevant.

asmorodskyi:
@okurz there is no failure . Just when you do 
```
upload_logs("/var/log/messages");
upload_logs("/var/log/messages");
```

you will get only one attachment to test and because I am doing snapshot rollback in my tests I need to save same logs before each rollback

testapi.pm
coolo: but that upname is an internal parameter, nothing humans type

asmorodskyi:
@coolo I rename it to {destination_log_name}

coolo:
log_name would have done for me :)

testapi.pm
```diff
@@ -1791,14 +1791,14 @@ sub data_url($) {
...
-  upload_logs($file [, failok => 0, timeout => 90 ]);
+  upload_logs($file [, failok => 0, timeout => 90, destination_log_name => "custom_name.log" ]);
 
 Upload C<$file> to OpenQA WebUI as a log file and
 return the uploaded file name. If failok is not set, a failed upload or
 timeout will cause the test to die. Failed uploads happen if the file does not
 exist or is over 20 GiB in size, so failok is useful when you just want
 to upload the file if it exists but not mind if it doesn't. Default
-timeout is 90s.
+timeout is 90s. destination_log_name parameter allow to control resulted job's attachment name.
```
okurz: * break lines
* Use https://perldoc.perl.org/perlpod.html#Formatting-Codes -> `C<destination_log_name>`

testapi.pm
```diff
@@ -1791,14 +1791,14 @@ sub data_url($) {
 
 =for stopwords GiB failok OpenQA WebUI
 
-  upload_logs($file [, failok => 0, timeout => 90 ]);
+  upload_logs($file [, failok => 0, timeout => 90, destination_log_name => "custom_name.log" ]);
```
okurz: It seems we have a mixture in these summary lines (not only this method) mixing "default options", "examples", "possible values". To be at least a bit consistent I recommend to s/"custom_name.log"/$autotest::current_test/

I would appreciate if you update the whole file to be consistent. My suggestion would be to use possible values only in the summary line, e.g. `upload_logs($file [, failok => 0|1, timeout => $timeout, destination_log_name => $destination_log_name ]);`

asmorodskyi: I think you also mixing things here : 
1. adding additional parameter to existing function 
2. improving overall code / documentation quality
While this commit is just about first point you pushing me to do second. 
In anycase I can't understand what you meant by  "timeout => $timeout " ? what it suppose to say ? from string "timeout => 90" I can at least understand that I suppose to bypass timeout as digit and not boolean for example

testapi.pm
asmorodskyi: Done

asmorodskyi:
@coolo - done

testapi.pm
okurz: I am not pushing you to anything. I stated: I would appreciate. This is my opinion. But the real issue is that you are not following the existing convention but provide an example where previously it was stating the default value which is is something different. This is why I still suggest to write `$autotest::current_test` instead

okurz:
@coolo feel free to share your opinion about https://github.com/os-autoinst/os-autoinst/pull/844#discussion_r135800081 or merge regardless

testapi.pm
asmorodskyi: log_name don't have default value at all,  when log_name is undefined **another** variable  ($upname) will get autogenerated value based on :
```
my $upname   = $args{log_name} || ref($autotest::current_test) . '-' . $basename;
```
and defining in documentation "log_name => $log_name"  or even "log_name => $autotest::current_test" will not help to understand this

asmorodskyi:
so I accept and change all requests regarding my PR except 
`log_name => "custom_name.log" `  to 
log_name => "destination_log_name => $destination_log_name"
because I don't think it is good idea and I explain why above. Have I chance to get this PR merged ?

okurz:
sure but I can't push the button because I don't have merge rights on os-autoinst.

coolo:
sorry that I only come back to this every couple of days - but wouldn't you want to have the basetest still in the uploaded name for consistency?

asmorodskyi:
done

coolo:
not if I read the code correct (test cases would be cool) - I was talking about textinfo uploading /var/log/messages results in textinfo-messages - with log_name => 'messages2' I would expect textinfo-messages2.

You originally uploaded as messages2 and now as messages2-messages - which is even more confusing

coolo:
let's extend the API and worry about the resulting filename later

PR #845, opened 2017-08-28
Fix bug reference in docs/backend_vars.asciidoc

foursixnine:
Due to the qemu hmp interface not being able to report status and the
combination with slow disks, we might end up having problems when a
system has been trying to save a snapshot but it takes way too long,
having the possibility to disable snapshots on slow workers comes in
handy for such situations.


okurz:
LGTM, it's not markdown ;-)

foursixnine:
You're right, It's asciidoc :) and therefore rendered properly now

PR #847, opened 2017-09-05
testapi: Add 'no_wait' option to wait_still_screen

okurz:
As we already have the option 'no_wait' for check_screen/assert_screen this
patch adds the same functionality to `wait_still_screen`. With the option
'no_wait' the internal sleep period is reduced so that there is merely no
waiting time between two consecutive internal screenshot checks for "still
screen" detection.

This was a recommendation by @coolo based on IRC discussion.

Related progress issue: https://progress.opensuse.org/issues/19092

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -207,8 +207,8 @@ subtest 'record_info' => sub {
 subtest 'validate_script_output' => sub {
     my $mock_testapi = new Test::MockModule('testapi');
     $mock_testapi->mock(script_output => sub ($;$) { return 'output'; });
-    pass(validate_script_output('script', sub { m/output/ }));
-    pass(validate_script_output('script', sub { m/output/ }, 30));
+    ok(!validate_script_output('script', sub { m/output/ }), 'validating output with default timeout');
+    ok(!validate_script_output('script', sub { m/output/ }, 30), 'specifying timeout');
     like(
         exception {
             validate_script_output('script', sub { m/error/ });
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -5,6 +5,7 @@ use warnings;
 use Test::More;
 use Test::Output;
 use Test::Fatal;
+use Test::Mock::Time;
 use Test::Warnings;
 use File::Temp;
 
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -529,23 +530,29 @@ sub assert_screen_change(&@) {
...
 
 sub wait_still_screen {
-    my $stilltime        = shift || 7;
-    my $timeout          = shift || 30;
-    my $similarity_level = shift || (get_var('HW') ? 44 : 47);
-
-    bmwqemu::log_call(stilltime => $stilltime, timeout => $timeout, simlvl => $similarity_level);
-
-    $timeout = bmwqemu::scale_timeout($timeout);
+    my $stilltime = looks_like_number($_[0]) ? shift : undef;
```
mudler: I think it's safe to put `7` here as well, so you don't have to check if it's `undef` later

t/03-testapi.t
```diff
@@ -207,8 +208,8 @@ subtest 'record_info' => sub {
 subtest 'validate_script_output' => sub {
     my $mock_testapi = new Test::MockModule('testapi');
     $mock_testapi->mock(script_output => sub ($;$) { return 'output'; });
-    pass(validate_script_output('script', sub { m/output/ }));
-    pass(validate_script_output('script', sub { m/output/ }, 30));
+    is(validate_script_output('script', sub { m/output/ }), '', 'validating output with default timeout');
```
mudler: possible to replace it with `ok(!...)` instead of `is(..,'')` 😃

testapi.pm
```diff
@@ -529,23 +530,29 @@ sub assert_screen_change(&@) {
...
+    my $timeout = shift if (@_ % 2);
+    my %args = (stilltime => $stilltime // 7, timeout => $timeout // $bmwqemu::default_timeout, @_);
+    $args{similarity_level} //= 47;
+    bmwqemu::log_call(%args);
+    $timeout = bmwqemu::scale_timeout($args{timeout});
```
mudler: i think this would be equivalent as well:
`
my $timeout =  (@_ % 2) ? bmwqemu::scale_timeout(shift) : $bmwqemu::default_timeout;
`

okurz: well, not entirely because the scaling should happen after the log output but you gave me incentive nevertheless to simplify a bit more. Thanks. See updated commit

t/03-testapi.t
okurz: done

testapi.pm

okurz:
updated: Attended all comments

sysrich:
No_wait for wait_screen_change?

Doesn't this become a noop?

What next, assert_screen with no_assert?
type_string with no_type?

sysrich:
More seriously, what is the point of this PR? No description. No link to tickets. Who are you and what did you do to @okurz?

coolo:
this comment says it all: 
```
with 'no_wait' actually wait a little bit
```

This is like assert_screen 'apache', but_any_other_webserver_works_too => 1 :)

okurz:
yeah, yeah. Updated commit message and PR description, tidied, rebased.

PR #848, opened 2017-09-05
Add wait_screen_change option on send_key

okurz:
While doing this, deprecating do_wait which is calling wait_idle which we want
to avoid as it is too much relying on a backend specific implementation which
can not be made available on enough backends.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1184,7 +1184,6 @@ sub mouse_click {
     my $time   = shift || 0.15;
     bmwqemu::log_call(button => $button, cursor_down => $time);
     query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
-    # FIXME sleep resolution = 1s, use usleep
     sleep $time;
     query_isotovideo('backend_mouse_button', {button => $button, bstate => 0});
 }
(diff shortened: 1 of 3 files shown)


okurz:
checks fixed

testapi.pm
```diff
@@ -1025,11 +1031,24 @@ Special characters naming:
...
-    query_isotovideo('backend_send_key', {key => $key});
-    wait_idle() if $do_wait;
+    my $key = shift;
+    my %args;
+    if (@_ == 1) {
+        %args = (do_wait => $_[0]);
+    }
+    else {
+        %args = @_;
+    }
```
mudler: isn't this equivalent?
```perl
my %args = (@_ == 1) ? (do_wait => $_[0]) : @_;
```

mudler: since i see you are not using @_ anymore, you could also:
```perl
my %args = (@_ == 1) ? (do_wait => shift) : @_;
```

okurz: good idea, done. PR updated

testapi.pm
```diff
@@ -1025,11 +1031,18 @@ Special characters naming:
...
+    $args{do_wait}            //= 0;
+    $args{wait_screen_change} //= 0;
+    bmwqemu::log_call(key => $key, %args);
+    if ($args{wait_screen_change}) {
+        wait_screen_change { query_isotovideo('backend_send_key', {key => $key}) };
+    }
+    else {
+        query_isotovideo('backend_send_key', {key => $key});
+    }
+    wait_idle() if $args{do_wait};
```
foursixnine: If we want to slowly remove wait_idle, why not also cluck if do_wait is > 0?.

okurz: because that is already within wait_idle, isn't it?

PR #850, opened 2017-09-06
Add minimal wait time to wait_screen_change

dzedro:
Reason for this change is failing QAM updates, I noticed long time ago that wait_screen_change doesn't provide what is expected because of the button press animation
Default minimal_timeout can be also 0 if 0.5 is problem
Fail:
https://openqa.suse.de/tests/1146901#step/shotwell_edit/21
I could reproduce 3 ([1](http://10.100.12.155/tests/69) [2](http://10.100.12.155/tests/66) [3](http://10.100.12.155/tests/65)) from 6 runs without minimal timeout
Test:
Last 36 tests are with this change http://10.100.12.155/tests

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -452,11 +452,12 @@ sub assert_and_dclick {
 
 =head2 wait_screen_change
 
-  wait_screen_change(CODEREF [,$timeout [, similarity_level => 50]]);
+  wait_screen_change(CODEREF [,$timeout [,$minimal_timeout [, similarity_level => 50]]]);
 
 Wrapper around code that is supposed to change the screen.
 This is the opposite to C<wait_still_screen>. Make sure to put the commands to change the screen
 within the block to avoid races between the action and the screen change.
+Minimal timeout is time to wait before screen change due to button press animation.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -475,14 +476,15 @@ subroutine block.
...
+Default minimal_timeout is 0.5s. Default similarity_level is 50.
 
 =cut
 
 sub wait_screen_change(&@) {
-    my ($callback, $timeout, %args) = @_;
-    $timeout ||= 10;
+    my ($callback, $timeout, $minimal_timeout, %args) = @_;
+    $timeout         ||= 10;
+    $minimal_timeout ||= 0.5;
```
foursixnine: I would suggest to have a default timeout of 0, instead of 0.5, and add couple of tests in t/03-testapi.t

dzedro: Default 0 will be better, found out wait_screen_change is used also in another functions
Working on tests

foursixnine: Mr. @dzedro Any updates here? :)

dzedro: Mr. @foursixnine no updates now, had no time to work on tests

PR #853, opened 2017-09-17
Add option to exit early for test schedule evaluation

okurz:
By setting the newly introduced test variable _EXIT_AFTER_SCHEDULE to a true
value the execution of actual tests is never started so that only the
evaluated schedule can be checked with fast feedback.

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -7,6 +7,7 @@ Supported variables per backend
 |====================
 Variable;Values allowed;Default value;Explanation
 EXCLUDE_MODULES;string;;comma separated names or fullnames of excluded test modules
+_EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation of the test schedule, e.g. to check only which test modules would be executed
 NOVIDEO;boolean;0;Do not encode video if set
 VNC_TYPING_LIMIT;integer;50;Maximum number of keys per second
 NO_DEBUG_IO;boolean;0;Disable the I/O debug output in case of needle comparison times longer than expected
--- a/isotovideo
+++ b/isotovideo
@@ -262,6 +262,9 @@ unshift @INC, $bmwqemu::vars{CASEDIR} . '/lib';
 require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
 @INC = @oldINC;
 
+diag 'Early exit has been requested with _EXIT_AFTER_SCHEDULE. Only evaluating test schedule.';
+exit 0 if $bmwqemu::vars{_EXIT_AFTER_SCHEDULE};
+
 # set a default distribution if the tests don't have one
 $testapi::distri ||= distribution->new;
 
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -1,4 +1,4 @@
 AM_MAKEFLAGS = PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db"
-TESTS = 00-compile-check-all.t 01-test_needle.t 02-test_ocr.t 03-testapi.t 04-check_vars_docu.t 05-pod.t 06-pod-coverage.t 07-commands.t 08-autotest.t 09-lockapi.t 10-terminal.t 11-image-ppm.t 12-bmwqemu.t 13-osutils.t 20-openqa-benchmark-stopwatch-utils.t 99-full-stack.t
+TESTS = 00-compile-check-all.t 01-test_needle.t 02-test_ocr.t 03-testapi.t 04-check_vars_docu.t 05-pod.t 06-pod-coverage.t 07-commands.t 08-autotest.t 09-lockapi.t 10-terminal.t 11-image-ppm.t 12-bmwqemu.t 13-osutils.t 14-isotovideo.t 20-openqa-benchmark-stopwatch-utils.t 99-full-stack.t
 
 EXTRA_DIST = $(TESTS)
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -1,0 +1,27 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Test::More;
+use File::Basename;
+use Cwd 'abs_path';
+
+my $toplevel_dir = abs_path(dirname(__FILE__) . '/..');
+my $data_dir     = "$toplevel_dir/t/data/";
+my $pool_dir     = "$toplevel_dir/t/pool/";
+
+chdir($pool_dir);
+open(my $var, '>', 'vars.json');
+print $var <<EOV;
+{
+   "CASEDIR" : "$data_dir/tests",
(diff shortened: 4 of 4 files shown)


isotovideo
```diff
@@ -262,6 +262,10 @@ unshift @INC, $bmwqemu::vars{CASEDIR} . '/lib';
 require $bmwqemu::vars{PRODUCTDIR} . "/main.pm";
 @INC = @oldINC;
 
+# for evaluation of the test schedule only governed by "main.pm" early exit
+# can be requested
+exit 0 if $bmwqemu::vars{EXIT_AFTER_SCHEDULE};
```
foursixnine: Can you add the test for the full-stack test?

okurz: I thought about it but hestitated to do so because the full stack test is already taking quite a long time, is hard to debug and encourages people to execute the tests even less likely on local instances and instead waste their time waiting for travis to report results after some time. *But* I added "t/14-isotovideo.t" for that purpose taking only 1 second to come back with a useful test result.

foursixnine: Fair enough :), I wonder what's failing in travis... Just restarted the job...

doc/backend_vars.asciidoc
```diff
@@ -7,6 +7,7 @@ Supported variables per backend
 |====================
 Variable;Values allowed;Default value;Explanation
 EXCLUDE_MODULES;string;;comma separated names or fullnames of excluded test modules
+EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation of the test schedule, e.g. to check only which test modules would be executed
```
coolo: as this is pretty internal, I would welcome if this was prefixed with _ - otherwise you will have someone who has a get_var('EXIT_AFTER_SCHEDULE') in her main.pm to test a random app and will be very confused.

okurz: updated.

Doesn't this also apply for other variables? I find it hard to distinguish between internal and external variables. I guess this is also why you stated "pretty internal". I guess there is also "not so much internal" or "ugly internal" ;-)

okurz: I thought about it more and I don't like it. I still updated the code to convince you that the PR is mergeable but I suggest to go back to the version without the underscore. What about SKIPTO? Shouldn't it be also prepended with leading underscore? What about QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64? Will it not also have a potential big impact? The good thing about EXIT_AFTER_SCHEDULE is that the effect is drastic, people will learn faster than with a hideous change of a backend device setting. Maybe I should just add a diag message on exit to make it more explicit? I did that, updated (still with leading underscore). A diag message "Early exit has been requested with _EXIT_AFTER_SCHEDULE. Only evaluating test schedule." will be printed.

coolo: SKIPTO is very old - and for new variables I e.g. insist on a QEMU_ prefix

But there is a difference between a variable that belongs in a machine and one that developers use to debug something in their main.pm

okurz:
did anyone understand the travis failure in https://travis-ci.org/os-autoinst/os-autoinst/builds/277618674?utm_source=github_status&utm_medium=notification ?

I see

```
make PERL5OPT="-MDevel::Cover=-db,/home/travis/build/os-autoinst/os-autoinst/t/cover_db" check-TESTS
make[2]: Entering directory `/home/travis/build/os-autoinst/os-autoinst/t'
rm: invalid option -- '2'
Try 'rm --help' for more information.
```

foursixnine:
I understand that there is something failing, but no clue... Will try tomorrow and see. In the meantime `BACKEND_ EXIT_AFTER_SCHEDULE` or `BACKEND_SCHEDULE_ONLY` or even `BACKEND_PLAN_TESTS_ONLY`... Since I could think of situations of people saying `Yeah, use this setting BACKEND_EXIT_AFTER_SCHEDULE, and wait until the openqa scheduler assigns the job/test to a worker, and the worker schedules the test that it will run....` kind of scenario...

okurz:
But isn't everything in os-autoinst `BACKEND…`?

foursixnine:
We have variables that affect the backend or one part of the backend, to date, we don't differentiate other than per "hardware backend". This can be used as stepping stone to have a clear understanding on what's affecting the backend behaviour, the "machine" behaviour, and the test scheduling and execution behaviour (Such as SKIPTO and TESTDEBUG)

okurz:
@foursixnine for now I would like to keep the "_"-Prefix. Maybe even I myself can improve the grouping later on in os-autoinst. Can you retrigger travis tests and merge?

mudler:
few other suggestions : what about ```NO_TEST(S)``` , ```SKIP_TESTS``` or ```SKIP_ALL_TESTS```?

okurz:
Yes, I would like to avoid the word 'TEST' because it's used so ambiguous.

foursixnine:
@okurz Sorry I will have to insist to prepend `BACKEND_`.

okurz:
Sorry, I have to insist that this does not make sense because the variable does not have an effect any of the backends "baseclass.pm      driver.pm     ikvm.pm  pvm.pm   s390x.pm  virt.pm console_proxy.pm  generalhw.pm  ipmi.pm  qemu.pm  svirt.pm"
and there is no other variable with a BACKEND prefix. *Also* there is a variable which is just called "BACKEND". I strongly suggest to not introduce a prefix with the same name as a variable itself.

t/Makefile.am
```diff
@@ -1,4 +1,4 @@
 AM_MAKEFLAGS = PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db"
-TESTS = 00-compile-check-all.t 01-test_needle.t 02-test_ocr.t 03-testapi.t 04-check_vars_docu.t 05-pod.t 06-pod-coverage.t 07-commands.t 08-autotest.t 09-lockapi.t 10-terminal.t 11-image-ppm.t 12-bmwqemu.t 13-osutils.t 20-openqa-benchmark-stopwatch-utils.t 99-full-stack.t
+TESTS = 00-compile-check-all.t 01-test_needle.t 02-test_ocr.t 03-testapi.t 04-check_vars_docu.t 05-pod.t 06-pod-coverage.t 07-commands.t 08-autotest.t 09-lockapi.t 10-terminal.t 11-image-ppm.t 12-bmwqemu.t 13-osutils.t 14-isotovideo.t -20-openqa-benchmark-stopwatch-utils.t 99-full-stack.t
```
coolo: if you removed the - before the 20 I would be fine to merge

okurz: done, updated. For the reference for later: I don't mind if we group the test variables better but I suggest we come up with a convention first covering all openQA, os-autoinst and os-autoinst-distri-opensuse and don't try to fix it by discussing about a single variable (and then never end and never proceed).

coolo:
you didn't chmod a+x the new test

okurz:
updated, waiting for tests

PR #852, opened 2017-09-07
Use hash ref in query_isotovideo

mnowaksuse:
With Perl 5.18.2 I get:

```
18:50:30.2820 480 <<< testapi::wait_idle(timeout=19)
18:50:30.3698 480 # Test died: Can't use string ("timeout") as a HASH
ref while "strict refs" in use
```

Regressed in b9045f43a3c8186db352c7a5de8eea501d861a84.

PR #854, opened 2017-09-18
Use hash ref in query_isotovideo and add a test for it.

foursixnine:
Supersede  #852

### DIFF
--- a/t/data/tests/tests/shutdown.pm
+++ b/t/data/tests/tests/shutdown.pm
@@ -18,6 +18,7 @@ use base 'basetest';
 use testapi;
 
 sub run {
+    wait_idle 1;
     type_string "sudo su\n";
     type_string "poweroff\n";
     assert_shutdown;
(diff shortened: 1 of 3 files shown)


t/data/tests/tests/shutdown.pm
```diff
@@ -18,6 +18,7 @@ use base 'basetest';
 use testapi;
 
 sub run {
+    wait_idle 1;
```
okurz: IMHO a waste of time. wait_idle should be tested in t/03-testapi.t with mocked sleep

okurz:
* Please don't push merge commits, rebase
* Please don't add a dot at EOL of the git commit message subject line

foursixnine:
@okurz Yeah, happens when the gh editor is used :( That thing is evil for this kind of things...

coolo:
no travis?

foursixnine:
@coolo Travis seems to be drunk again :(

foursixnine:
Now we're talking!

t/data/tests/tests/shutdown.pm
okurz: So you don't want to comment on this?

PR #857, opened 2017-09-20
Remove typing tests from integration tests

foursixnine:
With changes from 517e20c5, we broke fullstack on openQA, this fixes the
problem.


foursixnine:
@coolo care to merge?. 

While the change is also tested in 03-testapi.t, I see no harm in having them inside the general test plan for os-autoinst.

PR #859, opened 2017-09-27
Fix scale_timeout call in wait_serial function

ldevulder:
Some VMs can take time to run, and setting TIMEOUT_SCALE can be needed.
This patch adds support for TIMEOUT_SCALE in assert_script_run function, as it not supported yet.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -724,7 +724,7 @@ sub wait_serial {
     my %nargs = (@_, (regexp => $regexp, timeout => $timeout));
 
     bmwqemu::log_call(%nargs);
-    $timeout = bmwqemu::scale_timeout($timeout);
+    $nargs{timeout} = bmwqemu::scale_timeout($nargs{timeout});
 
     my $ret = query_isotovideo('backend_wait_serial', \%nargs);
     my $matched = $ret->{matched};


okurz:
I don't think this should be necessary as assert_script_run should use wait_serial internally which scales the timeout already. Also it's strongly recommended you add a unit test, e.g. in t/03-testapi.t

ldevulder:
Hmm you're right, wait_serial already implement scale_timeout, but I had a the case that my assert_script_run call has been closed less than 120s (default is 90s) and so seems to not use the TIMEOUT_SCALE variable.
I will do the test again and investigate further.

ldevulder:
After many tests I can confirm that timeout scaling is not done as expected in ```wait_serial``` function.
This is because the ```timeout``` variable changed with ```scale_timeout``` is not correct. The timeout used in ```query_isotovideo``` function is passed through ```nargs```, and ```nargs``` is initialized with ```timeout``` variable *before* scaling it!

So, I changed the code to use ```$nargs{timeout}``` instead of ```${timeout}``` (see my new commit).

okurz:
that looks nice. Did you take a look if you could add a test for this scaling feature?

ldevulder:
No yet for the test case. I will try to do this, I just need to check testapi.t to see how tests work.

ldevulder:
As 03-testapi.t already contains a subtest for assert_script_run, which uses script_run and wait_serial, is this not enough?
My knowledge of subtest in Perl is pretty low :)

okurz:
well, it's apparently not enough when the bug you discovered doesn't make a test fail :)

I thought of the simple extreme programming rule [when a bug is found a test is created](http://www.extremeprogramming.org/rules/bugs.html) so I suggest the following:

* go back to the previous state of the source code
* add a test that fails because the scaling is not properly used
* add back your fix
* see the test passing and be happy :)

ldevulder:
I tried but for now failed...

The main issue I face is that in 03-testapi.t some commands are *fake* commands, like wait_serial.
And as the check is done during compilation time (so no openQA worker started) I really don't know how to implement this test...

My basic idea was to do this in ```subtest script_run```:
```
# Test if TIMEOUT_SCALE variable is well honored
$bmwqemu::vars{TIMEOUT_SCALE} = '4';
my $test_timeout = 20;
my $expected_timeout = $test_timeout * $bmwqemu::vars{TIMEOUT_SCALE};
$fake_exit = 0;
is(script_run("sleep $expected_timeout", $test_timeout), '0', 'script_run with TIMEOUT_SCALE value setting, returns exit code');
```

But this doesn't work...

I agree with you that adding a test case is a good idea, and I really want to it, but I don't how, and my Perl knowledge of the Test module is very poor... So I may need some help to do it :)

okurz:
yeah, I also don't know any better definite idea right now. Maybe one of use will have some cool idea anytime later, shouldn't stop PR from being merged -> LGTM

foursixnine:
I think the easiest way is to add it into the general test plan (t/data/tests/), there is a main.pm which could have another .pm file with this test... This would be the an easy way (at least the one that comes to mind) on how to test this properly. And as long it's inside the unless... It will not affect tests from the openQA project. I'm just not so sure that adding too many tests there is a wise thing :) (#857). 

Another way would be to [mock](https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L138) bmwqemu::scaletimeout as it is not really [a big](https://github.com/os-autoinst/os-autoinst/blob/master/bmwqemu.pm#L321) thing...

ldevulder:
@foursixnine: but in my case, the problem is not the bmwqemu::scaletimeout function itself, but the testapi::wait_serial function that not well called bmwqemu::scaletimeout. So testing it will not cover this case, no?

okurz:
You could mock `scale_timeout` and check within it's mock method if it's called, like https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L116

ldevulder:
Yes I could check if it is called. Iin my case ```scale_timeout``` was called, but the result was put in the wrong variable (```$timeout``` instead of ```$nargs{timeout}```. So a test like this would have been passed, no?

It's not that I don't want to add a test, it's just that it's better to add a useful test, like you said :)

ldevulder:
Sorry I clicked on the wrong button...

PR #862, opened 2017-10-06
[WIP] Improve User-feedback for VNC sockets is not existing

Soulofdestiny:
On ipmi and s390x sometimes jobs gets incomplete or fail without a module
This PR is meant to improve user feedback by showing a record_info box with a proper
error message

see related tickets:
https://progress.opensuse.org/issues/20022
https://progress.opensuse.org/issues/23650

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -9,6 +9,8 @@ use Time::HiRes qw( usleep gettimeofday time );
 use Carp;
 use List::Util 'min';
 
+use testapi 'record_info';
+
 use Crypt::DES;
 use Compress::Raw::Zlib;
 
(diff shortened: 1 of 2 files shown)


consoles/VNC.pm
```diff
@@ -878,7 +880,11 @@ sub _receive_message {
     my $self = shift;
 
     my $socket = $self->socket;
-    $socket or die 'socket does not exist. Probably your backend instance could not start or died.';
+    #$socket or die 'socket does not exist. Probably your backend instance could not start or died.';
```
okurz: should be deleted then but I would like to keep the text in the exception error message. We should give that hint to the test writer that probably we should not have run into that step at all, i.e. test code need to be changed or the SUT does not respond. And not "there is no socket because something is wrong in the backend"

rwx788: Fixed in https://github.com/os-autoinst/os-autoinst/pull/864

coolo:
superseded by 864

PR #863, opened 2017-10-08
Generalize x11_start_program with standard array argument

okurz:
This is a backward-compatible change. As the baseclass does not provide a real
implementation we should not try to parse any arguments. Having the last
argument as a generic array still allows for the old signature implicitly but
also other uses, for example a hash assignment for named parameters. This
change corrects the debug log output for os-autoinst-distri-opensuse after
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/3671

Use of old-style x11_start_program: http://lord.arch/tests/7652
New-style x11_start_program: http://lord.arch/tests/7653

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -72,10 +72,6 @@ sub add_console {
 }
 
 sub x11_start_program {
-    my ($program, $timeout, $options) = @_;
-    $timeout ||= 6;
-    $options ||= {};
-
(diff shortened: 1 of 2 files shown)


coolo:
such functions shouldn't be in testapi at all - but rather in the basetest :(

okurz:
I agree

PR #864, opened 2017-10-09
[WIP] Improve User-feedback for VNC sockets is not existing

rwx788:
Based on https://github.com/os-autoinst/os-autoinst/pull/862, simple fix of review comment:

On ipmi and s390x sometimes jobs gets incomplete or fail without a module
This PR is meant to improve user feedback by showing a record_info box with a proper
error message

see related tickets:
https://progress.opensuse.org/issues/20022
https://progress.opensuse.org/issues/23650


coolo:
we we have a test run?

rwx788:
@coolo not really. And we have discussed that with @nicksinger that it requires more changes and this change alone doesn't make much sense =(

PR #865, opened 2017-10-10
avoid accidentally cleaning by debhelper on Debian/Ubuntu

henrich:
debhelper packaging utility automatically cleans .orig files, so rename it as .original


okurz:
LGTM

PR #866, opened 2017-10-10
sshVirtsh: Hide confusing errors about undefined machines

okurz:
Before:
```
20:21:54.4386 22178 Command executed: virsh  destroy openQA-SUT-12
20:21:54.4593 22178 Command's stdout:

20:21:54.4594 22178 Command's stderr:
error: failed to get domain 'openQA-SUT-12'
error: Domain not found: no domain with matching name 'openQA-SUT-12'

20:21:54.4990 22178 Command executed: virsh  undefine --snapshots-metadata openQA-SUT-12
20:21:54.5195 22178 Command's stdout:

20:21:54.5196 22178 Command's stderr:
error: failed to get domain 'openQA-SUT-12'
error: Domain not found: no domain with matching name 'openQA-SUT-12'
```

After:
```
20:44:00.1185 24093 Command executed: virsh  destroy openQA-SUT-12 >/dev/null 2>&1
20:44:00.4177 24093 Command executed: virsh  undefine --snapshots-metadata openQA-SUT-12 >/dev/null 2>&1
```

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -462,6 +462,8 @@ sub get_ssh_output {
         return ($stdout, $errout);
     }
     else {
+        chomp($stdout);
+        chomp($errout);
         bmwqemu::diag "Command's stdout:\n$stdout" if length($stdout);
         bmwqemu::diag "Command's stderr:\n$errout" if length($errout);
     }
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -517,8 +519,8 @@ __END"
 
     # shut down possibly running previous test (just to be sure) - ignore errors
     # just making sure we continue after the command finished
-    $self->run_cmd("virsh $remote_vmm destroy " . $self->name);
-    $self->run_cmd("virsh $remote_vmm undefine --snapshots-metadata " . $self->name);
+    $self->run_cmd("virsh $remote_vmm destroy " . $self->name . ' >/dev/null 2>&1');
```
coolo: I'm torn here. We want to ignore expected errors - but hiding unexpected errors makes it harder to debug

okurz: so `grep -v 'failed to get domain' | grep -v 'Domain not found'`?

okurz: I don't want to ignore errors on other occassions just here. There is also the comment above stating "ignore errors", that's why :)

mnowaksuse: I'd prefer to see the `grep ...` here instead of ignoring everything.

okurz: done

mudler:
Please squash your commits since the changes seems regarding the same purpose - I would also avoid to put the full output (commented in markdown) in the commit message.

okurz:
squashed

PR #868, opened 2017-10-16
Rescan needles on set_var to re-register needles with the new variable

coolo:
This is a suggestion on the API could look like to implement 

https://progress.opensuse.org/issues/25504

This will fail as I plan to implement this test driven - so please review the test and not the test result.


backend/baseclass.pm
```diff
@@ -589,6 +589,16 @@ sub capture_screenshot {
     return;
 }
 
+sub reload_needles {
+    # called from testapi::set_var, so read the vars
+    bmwqemu::load_vars();
+
+    for my $n (needle->all()) {
+        $n->unregister();
+    }
```
okurz: you don't like something like `map { $_->unregister } needle->all()`? - if that even works. I know, you just moved that code

testapi.pm
```diff
@@ -611,15 +611,24 @@ sub get_required_var {
 
 =head2 set_var
 
-  set_var($variable, $value);
+  set_var($variable, $value [, reload_needles => 1] );
 
 Set test variable C<$variable> to value C<$value>.
 
+If you the reload_needles => 1 flag, the backend will reload the
```
okurz: If you what? I suggest

```perl
Specify a true value for the C<reload_needles> flag to trigger a reloading of needles in the backend and call …
```

backend/baseclass.pm
coolo: No, I don't like calling map for something that doesn't map

okurz: got it

testapi.pm
coolo: updated

okurz:
LGTM

PR #869, opened 2017-10-16
[WIP] -- why does this work? -- Fix needle loading if the needledir is a symlink itself

okurz:


okurz:
what if this is still needed to make openQA work for me?

coolo:
then you better switch to bind mounts

okurz:
ah, you mean maybe the needle loading is not permitted by apparmor but os-autoinst doesn't tell. Hm, that could be. Thanks for the pointer.

PR #874, opened 2017-10-19
[WIP] needs test on real instance -- Check all 'exec' calls (local and remote) for errors

okurz:
Related progress issue: https://progress.opensuse.org/issues/26094

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -116,7 +116,8 @@ sub run_cmd {
         username => 'root'
     );
     my $chan = $self->{ssh}->channel();
-    $chan->exec($cmd);
+    $chan->exec($cmd)
+      or $chan->die_with_error;
     get_ssh_output($chan);
     $chan->send_eof;
     my $ret = $chan->exit_status();
@@ -244,16 +245,19 @@ sub start_serial_grab {
         # libvirt esx driver does not support `virsh console', so
         # we have to connect to VM's serial port via TCP which is
         # provided by ESXi server.
-        $chan->exec('nc ' . get_var('VMWARE_HOST') . ' ' . get_var('VMWARE_SERIAL_PORT'));
+        $chan->exec('nc ' . get_var('VMWARE_HOST') . ' ' . get_var('VMWARE_SERIAL_PORT'))
+          or $chan->die_with_error;
     }
     elsif (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
         # Hyper-V does not support serial console export via TCP, just
         # windows named pipes (e.g. \\.\pipe\mypipe). Such a named pipe
         # has to be enabled by a namedpipe-to-TCP on HYPERV_SERVER application.
-        $chan->exec('nc ' . get_var('HYPERV_SERVER') . ' ' . get_var('HYPERV_SERIAL_PORT'));
+        $chan->exec('nc ' . get_var('HYPERV_SERVER') . ' ' . get_var('HYPERV_SERIAL_PORT'))
+          or $chan->die_with_error;
     }
     else {
-        $chan->exec('virsh console ' . $name);
+        $chan->exec('virsh console ' . $name)
+          or $chan->die_with_error;
     }
 }
 
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -83,7 +83,8 @@ sub activate {
         open(STDIN,  "<&", $peer);
         open(STDOUT, ">&", $peer);
         close($peer);
-        exec("Xvnc -depth 16 -inetd -SecurityTypes None -ac $display");
+        exec("Xvnc -depth 16 -inetd -SecurityTypes None -ac $display")
+          or die "Could not execute Xvnc: $!";
     }
     close($s);
     #print "$self->{testapi_console} -> $port\n";
--- a/consoles/sshIucvconn.pm
+++ b/consoles/sshIucvconn.pm
@@ -40,7 +40,8 @@ sub connect_remote {
     $chan->blocking(0);
     $chan->pty(1);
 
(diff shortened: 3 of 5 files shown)


okurz:
Seems like the error handling would only work properly with libssh2 version >= 1.7.0, see https://metacpan.org/pod/Net::SSH2#Error-handling . For openSUSE Leap 42.3 we have libssh2-1 1.4.3-18.3 so I think this will not lead us anywhere useful (for now).

Any ideas?

mnowaksuse:
We have `libssh2-1-1.8.0-3.1.x86_64` in `obs://build.opensuse.org/devel:openQA`. I guess we don't have this repo on workers?

To the code itself: It seems to me there are instances where `or die` would block waiting for the command to exit but I think in several cases we execute a command and let it run in the background.

okurz:
I think the check is acting upon the delivery of the command over the ssh channel. So even if a command "false &" would be executed over the ssh channel I think this is considered as "successful".

foursixnine:
@mnowaksuse we use that repo on the workers too... so I guess this wouldn't be that much of a problem

foursixnine:
@okurz if you package your change (Basically just branch the obs os-autoinst package, and change the _service file to point to your repo/branch), we could use the staging instance to test this.

okurz:
cool idea. But first, @mnowaksuse where did you find the package in devel:openQA? Could not find it.
It seems in openSUSE:Factory we have "libssh2_org" in version 1.8.0 as well as SLE:15 so we should be safe there. Still I would not expedite this fix when openSUSE Leap 42.3 does not have it available.

mnowaksuse:
@okurz I think @coolo added it per my request https://progress.opensuse.org/issues/15732#note-7 and it is on some of my systems since. However, I haven't found evidence of the package being in the openQA repos now.

coolo:
https://build.opensuse.org/package/show/devel:openQA:Leap:42.2/libssh2_org - we can do the same for 42.3

okurz:
Do you recommend to do that and rely on the more recent version in all cases by default,  should we scrap this idea or just wait it out until we have the new version in new stable SLE and openSUSE releases?

okurz:
Rebased. However the package situation did not change much:

```
$ osc se --version libssh2_org
SUSE:SLE-15:GA                                       libssh2_org  1.8.0
openSUSE:Factory                                     libssh2_org  1.8.0
openSUSE:Leap:15.0                                   libssh2_org  1.8.0
openSUSE:Leap:15.1                                   libssh2_org  1.8.0
openSUSE:Leap:42.3                                   libssh2_org  1.4.3
```

So I did `osc linkpac openSUSE:Factory libssh2_org devel:openQA:Leap:42.3` -> https://build.opensuse.org/package/show/devel:openQA:Leap:42.3/libssh2_org

Martchus:
```
# Compilation of consoles::sshVirtsh failed: Global symbol "$chan" requires explicit package name at consoles/sshVirtsh.pm line 358.
# Compilation failed in require at /usr/lib/perl5/vendor_perl/5.18.2/Test/Compile/Internal.pm line 197.
ok 30 - no (unexpected) warnings (via END block)
not ok 30 - Compile test for consoles/sshVirtsh.pm
```

okurz:
The linkpac'd package seems to fail the build of containers, e.g. in https://build.opensuse.org/package/live_build_log/devel:openQA/openqa_dev/containers/x86_64

```
  203s] Resolving package dependencies...
[  203s] 
[  203s] Problem: libssh2-devel-1.8.0-4.1.x86_64 requires libssh2-1 = 1.8.0, but this requirement cannot be provided
[  203s]   uninstallable providers: libssh2-1-1.8.0-4.1.x86_64[obs_repository]
[  203s]  Solution 1: install libssh2-1-1.8.0-4.1.x86_64 (with vendor change)
[  203s]   openSUSE  -->  obs://build.opensuse.org/devel:openQA
[  203s]  Solution 2: do not ask to install a solvable providing libssh2-devel
[  203s]  Solution 3: break libssh2-devel-1.8.0-4.1.x86_64 by ignoring some of its dependencies
[  203s] 
[  203s] Choose from above solutions by number or cancel [1/2/3/c] (c): c
```

however it seems I can not easily add the option `--allow-vendor-change` as that fails as in https://build.opensuse.org/package/live_build_log/home:okurz:branches:devel:openQA:fix_containers_libssh2/openQA/containers/x86_64 with `Unknown option '--allow-vendor-change'`

I do not know how to fix that for now. Hints would be appreciated :)

I deleted the package again for now to see if https://build.opensuse.org/package/show/devel:openQA/openqa_dev# works again.

backend/svirt.pm
```diff
@@ -116,7 +116,8 @@ sub run_cmd {
         username => 'root'
     );
     my $chan = $self->{ssh}->channel();
-    $chan->exec($cmd);
+    $chan->exec($cmd)
+      or $chan->die_with_error;
```
Martchus: Have you ever actually tested this? Considering the documentation it looks wrong. The `die_with_error` should be called on the `Net::SSH2` object and not on the channel.

---

But I'm going to improve the error handling for https://progress.opensuse.org/issues/44771 anyways. There are also more occurrences like the creation of the channel object itself which need to be covered.

Martchus:
I guess this is superseded by https://github.com/os-autoinst/os-autoinst/pull/1126.

By the way, the problem with *this* PR is that it calls `die_with_error` on the channel object. And it does not handle errors on channel creation (which also happen in practice).

backend/svirt.pm
okurz: Yes, I tested this. Could be that I had that fixed already locally or maybe I did not run into this yet. Thanks in general for taking over and bringing it to a good end.

PR #875, opened 2017-10-27
[WIP] Use ESXi server instead of vSphere

mnowaksuse:
In future we should probably use ESXi server directly not via vSpere cluster. Well but currently we don't have reasonable server wirh non-free license, and free one has API disabled:

error: Failed to define domain from /var/lib/libvirt/images/openQA-SUT-1.xml
error: internal error: HTTP response code 500 for call to 'RegisterVM_Task'. Fault: ServerFaultCode - Current license or ESXi version prohibits execution of the requested operation.


coolo:
it's almost half a year

mnowaksuse:
Will revisit once we have an ESXi server.

PR #876, opened 2017-11-02
Replacement of hot filehandlers for Mojo::Log

dasantiago:
This change will replace writing to the autoinst-log.txt with
a hot filehandler, with a proper logging mechanism: Mojo::Log

**This change will remove the color printing when printing to the terminal.**


bmwqemu.pm
```diff
@@ -53,7 +54,7 @@ our $screenshotpath = "qemuscreenshot";
 
 # global vars
 
-our $logfd;
+our $logger;
 
 our $istty;
```
coolo: this variable is not obsolete

bmwqemu.pm
```diff
@@ -182,55 +185,43 @@ sub get_timestamp {
...
 
-sub print_possibly_colored {
-    my ($text, $color) = @_;
-
-    if (($direct_output && !$istty) || !$direct_output) {
-        $logfd && print $logfd get_timestamp() . "$$ $text\n";
-    }
-    if ($istty || !$logfd) {
-        if ($color) {
-            print STDERR colored(get_timestamp() . "$$ " . $text, $color) . "\n";
```
coolo: as you remove colored calls, you can remove the Term::ANSIColor dependency as well

bmwqemu.pm
dasantiago: Good catch!

Change done :-)

bmwqemu.pm
dasantiago: Done.

Removed from the imports and from the cpanfile

okurz:
Hi, I think with this PR you changed the timestamp format missing out on the subsecond precision. Was this changed on purpose or maybe missed?

bmwqemu.pm
dasantiago: @okurz Only for STDOUT/STDERR, correct?

If yes, I'm fine with it. Please open a poo.

okurz: I did with https://progress.opensuse.org/issues/34978 , thx

PR #877, opened 2017-11-02
Add possibility to share files on worker directly

rwx788:
These changes are mainly caused by scenarios for autoyast installations.
For SLE15 we need to introduce information for registration, modules
into the profiles which differ for architectures. As an alternative we
could use chained test suites or support server. However, support server
does variables expansion using apache features, and would introduce
more load on multimachine capable workers. Chained jobs is better
solution than support server, and basic sharing modified autoyast profile
will work, whereas this approach won't work for class based autoyast
installations. These changes also will help in other scenarios allowing
avoidance of the extra resources usage.

Tested only manually so far using isotovideo.

See [poo#27076](https://progress.opensuse.org/issues/27076).

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -221,6 +221,14 @@ sub isotovideo_post {
     return isotovideo_command($c, [qw(interactive stop_waitforneedle continue_waitforneedle reload_needles)]);
 }
 
+sub get_temp_file {
+    my ($self)  = @_;
+    my $relpath = $self->param('relpath');
+    my $path    = testapi::hashed_string($relpath);
+    return _test_data_file($self, $path) if -f $path;
+    return $self->reply->not_found;
+}
+
 sub run_daemon {
     my ($port) = @_;
 
@@ -249,6 +257,9 @@ sub run_daemon {
 
     # to get the current bash script out of the test
     $token_auth->get('/current_script' => \&current_script);
+
+    # to get temporary files from the current worker
+    $token_auth->get('/files/*relpath' => \&get_temp_file);
 
     # get asset
     $token_auth->get('/assets/#assettype/#assetname'          => \&get_asset);
--- a/testapi.pm
+++ b/testapi.pm
@@ -21,7 +21,8 @@ use Carp;
 use Exporter;
 use 5.018;
 use warnings;
-use File::Basename 'basename';
+use File::Basename qw(basename dirname);
+use File::Path 'make_path';
 use Time::HiRes qw(sleep gettimeofday tv_interval);
 use autotest 'query_isotovideo';
 use Mojo::DOM;
@@ -62,6 +63,8 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   save_memory_dump save_storage_drives freeze_vm resume_vm
 
   diag hashed_string
+
+  save_tmp_file
 );
 our @EXPORT_OK = qw(is_serial_terminal);
 
@@ -959,6 +962,34 @@ sub script_output {
     $output =~ s/^\s+|\s+$//g;
 
     return $output;
+}
+
+
+=head2 save_tmp_file
+
+  save_tmp_file($relpath, $content)
+
+Saves content to the file in the worker pool directory using hash of the path,
+including file, so it can be fetched via http later on using
+ C< <autoinst_url>/files/#path_to_the_file> > url.
+Can be used to modify files for specific test needs, e.g. autoinst profiles.
+Dies if cannot open file for writing.
+
+Example:
+  save_tmp_file('autoyast/autoinst.xml', '<profile>Test</profile>')
+Then the file can be fetched using url:
+C< <autoinst_url>/files/autoyast/autoinst.xml> >
+
+=cut
+
+sub save_tmp_file {
+    my ($relpath, $content) = @_;
+    my $path = hashed_string($relpath);
+
+    bmwqemu::log_call(path => $relpath);
+    open my $fh, ">", $path;
+    print $fh $content;
+    close $fh;
 }
 
 =head2 validate_script_output
--- a/testapi.pm
+++ b/testapi.pm
@@ -64,7 +64,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
 
   diag hashed_string
 
-  save_tmp_file
+  save_tmp_file get_test_data
 );
 our @EXPORT_OK = qw(is_serial_terminal);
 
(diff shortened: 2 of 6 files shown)


commands.pm
```diff
@@ -221,6 +221,16 @@ sub isotovideo_post {
     return isotovideo_command($c, [qw(interactive stop_waitforneedle continue_waitforneedle reload_needles)]);
 }
 
+sub get_temp_file {
+    my ($self) = @_;
+    my $path = $self->param('relpath');
+    # do not allow .. in path
+    return $self->reply->not_found if defined $path && $path =~ /^(.*\/)*\.\.(\/.*)*$/;
```
foursixnine: I think you can rewrite the regex as: /\..*\./

rwx788: This won't be exactly same, but we can simplify it to simply look for /../

testapi.pm
```diff
@@ -949,6 +952,31 @@ sub script_output($;$) {
...
+sub share_tmp_file($$) {
+    my ($relpath, $content) = @_;
+    my $path     = File::Spec->rel2abs($relpath);
+    my $filename = basename($path);
+    my $dirname  = dirname($path);
+    # Create directory if doesn't exist
+    if (!-d $dirname) {
+        make_path $dirname;
+    }
+    open my $fh, ">", $path or croak("Could not open file. $!");
```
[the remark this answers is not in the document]
okurz: Why bother at all when we have `use autodie ':all';` that should catch failures in open

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
     return $output;
 }
 
+
+=head2 share_tmp_file
+
+  share_tmp_file($relpath, $content)
+
+Saves content to the file in the worker pool directory using relative path,
+to be fetched via http later on using <autoinst_url>/files/#relpath/#file url.
```
okurz: Better escape the URL with C<…>

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
...
+Can be used to modify files for specific test needs, e.g. autoinst profiles.
+Dies if cannot open file for writing.
+
+Example:
+  share_tmp_file(autoyast/autoinst.xml, "<profile>Test</profile>")
```
okurz: Don't you need ticks around the first argument as well? Also I recommend single ticks for the second argument as well

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
...
+
+
+=cut
+
+sub share_tmp_file($$) {
```
okurz: Please just use `sub share_tmp_file` without the argument signature, we want to avoid it for … perl reasons :)

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
...
+    my ($relpath, $content) = @_;
+    my $path     = File::Spec->rel2abs($relpath);
+    my $filename = basename($path);
+    my $dirname  = dirname($path);
+
+    bmwqemu::log_call(path => $path, filename => $filename);
+    # Create directory if doesn't exist
+    if (!-d $dirname) {
+        make_path $dirname;
+    }
```
okurz: Should be pretty obvious what's happening here, shouldn't need the comment ->

```perl
make_path $dirname if ! -d $dirname;
```

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
...
+    open my $fh, ">", $path or croak("Could not open file $filename located in $dirname. $!");
```
okurz: Again, use autodie

testapi.pm
rwx788: We have it for other functions and we have 2 mandatory parameters. But fine with me

testapi.pm
rwx788: fixed

testapi.pm
rwx788: Makes sense

testapi.pm
rwx788: Please, check if now it's fine

coolo:
As my comment is hidden in outdated - I copy it here (I somehow managed to review but not press the final button some days ago):

I'm somewhat reluctant to allow the tests to mess around in the worker directory directly. To support your use case (which really would benefit from test coverage) it would be better to save all such content in a single directory and use the hashsum of the relpath variable to lookup. This way you basically avoid any messing around - but only provide data under a given key.

rwx788:
@coolo I was considering this option as well, but decided maybe it's an overhead. Should we use tmp folder in the worker directory then?

coolo:
what folder you use, is not important then. Even the pool directory directly would work - but as tmp already exists, go for it.

testapi.pm
```diff
@@ -949,6 +952,38 @@ sub script_output($;$) {
     return $output;
 }
 
+
+=head2 share_tmp_file
+
+  share_tmp_file($relpath, $content)
```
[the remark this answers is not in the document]
foursixnine: @rwx788 ping :)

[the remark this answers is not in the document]
rwx788: Yeah, I didn't have a chance to work on that, hopefully will do so next week, should not take long =)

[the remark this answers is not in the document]
foursixnine: It took long apparently :)

testapi.pm
```diff
@@ -961,6 +964,34 @@ sub script_output {
     return $output;
 }
 
+
+=head2 save_tmp_file
+
+  save_tmp_file($relpath, $content)
+
+Saves content to the file in the worker pool directory using hash of the path,
+including file, so can be fetched via http later on using
```
okurz: so *it* can be …

testapi.pm
```diff
@@ -961,6 +964,34 @@ sub script_output {
...
+ C< <autoinst_url>/files/#path_to_the_file> > url.
+Can be used to modify files for specific test needs, e.g. autoinst profiles.
+Dies if cannot open file for writing.
+
+Example:
+  save_tmp_file('autoyast/autoinst.xml', '<profile>Test</profile>')
+Then file can be fetched using url:
```
okurz: Then *the* file …

okurz:
how about a test in either one of the t/*.t files or the full stack test? This can also serve as a good example how to use this feature.

rwx788:
@okurz .t test doesn't make much sense here, we've already discussed that with @foursixnine, he will help me a bit here

AdamWill:
This breaks build of os-autoinst in our (Fedora's) package build environment, because of some reliance on networking that is not present - Fedora packages are intentionally built in an environment with absolutely no networking:

    [2017-12-20T17:46:16.0043 UTC] [debug] # Test died: command 'wget -q http://10.0.2.2:15223/zGmFWfutOr/files/modified.xml' timed out at /builddir/build/BUILD/os-autoinst-25191d50d54eaded10b6b26199bb986728dcd5c2/t/data//tests/tests/modify_and_upload_file.pm line 45.

...

    not ok 9 - save_tmp_file test
    #   Failed test 'save_tmp_file test'
    #   at ./99-full-stack.t line 68.
    #          got: '256'
    #     expected: '0'

This may affect SUSE package builds too, I guess it depends how strongly the OBS build environment restricts network access...

okurz:
Hm, OBS builds IIRC don't have Internet but maybe they have networking locally

coolo:
our obs package removes the fullstack test for longer

AdamWill:
Ah, I see. I've been running the fullstack test on i686 and x86_64 for our package builds, as it has worked up until this commit.

rwx788:
I've added a check to identify if network is there: https://github.com/os-autoinst/os-autoinst/pull/900
But I wonder if we could use some env variables to define which tests to execute...

AdamWill:
Actually, it seems this was a false alarm: what I thought was the test failing due to lack of networking was actually just a consequence of a [different issue](https://bugs.launchpad.net/qemu/+bug/1738283?comments=all). The problem that issue causes is that, when the `boot.pm` test tries to do `cat > text <<EOF`, it actually winds up doing `cat > text >>EOF` - which means the `cat` command doesn't end when the test types `EOF`, and just keeps running. So anything any test *after* `boot.pm` types is not being interpreted by the shell but just appended to this still-running `cat`.

Once I figured out the cause of that bug and fixed it, the tests started passing, *including* these file tests. So apparently Fedora's build environment does allow this to work.

PR #878, opened 2017-11-07
Add offline mode to script_output

asmorodskyi:


asmorodskyi:
Unit tests are passed but I want to do check run with some real scenario before merge

mnowaksuse:
I guess those `doc/backend_vars.asciidoc` changes were not intended.

testapi.pm
```diff
@@ -919,6 +922,11 @@ sub script_output($;$) {
         type_string("\n", terminate_with => 'EOT');
         wait_serial("$suffix-0-");
     }
+    elsif ($args{offline}) {
+        my $cat = "cat - > /tmp/script$suffix.sh << EOF;";
+        type_string($cat . "\n");
```
okurz: Don't you need to type `$current_test_script` here?

testapi.pm
```diff
@@ -903,12 +903,15 @@ fetches the script through HTTP into the SUT and execs it with C<bash -xe> and d
 C<stdout> (I<not> C<stderr>!) to the serial console and returns the output I<if> the script
 exists with 0. Otherwise the test is set to failed.
 
-The default timeout for the script is 30 seconds. If you need more, pass a second parameter
+The default timeout for the script is 30 seconds. If you need more, pass a second parameter.
+By passing as third parameter "offline => 1" you workaround case when SUT don't have network -
```
okurz: Better use `C<…>` for escaping and also update L900. Also for better english I recommend "Pass as third parameter C<offline => 1> in case the SUT does not have network. Instead of downloading the script this function will just use C<type_string> to type the passed script string."

asmorodskyi: "C<oofline => 1>" - will not work because "=>" need to be escaped somehow , do you know how to escape here ? 
And how I can check resulted text BTW ?

testapi.pm
```diff
@@ -897,18 +897,21 @@ sub script_sudo {
...
-  script_output($script [, $wait])
+  script_output($script [, $wait, offline => 1])
 
 fetches the script through HTTP into the SUT and execs it with C<bash -xe> and directs
 C<stdout> (I<not> C<stderr>!) to the serial console and returns the output I<if> the script
 exists with 0. Otherwise the test is set to failed.
 
-The default timeout for the script is 30 seconds. If you need more, pass a second parameter
+The default timeout for the script is 30 seconds. If you need more, pass a second parameter.
+Pass as third parameter "offline => 1" in case the SUT does not have network.
```
coolo: nitpick: these are actually parameter 3 and 4, => is just an alias for ,

coolo:
I like the idea, but I would like to have it more generic.

Don't name that parameter offline, but type_command and make the default for it depend on the length of the script.

Because script_output('ls /home') takes much more typing if we curl it.

testapi.pm
```diff
@@ -897,19 +897,24 @@ sub script_sudo {
...
-fetches the script through HTTP into the SUT and execs it with C<bash -xe> and directs
-C<stdout> (I<not> C<stderr>!) to the serial console and returns the output I<if> the script
-exists with 0. Otherwise the test is set to failed.
+Executing script inside SUT with C<bash -eox> (in case of serial console with C<bash -eo>)
+and directs C<stdout> (I<not> C<stderr>!) to the serial console and returns
+the output I<if> the script exists with 0. Otherwise the test is set to failed.
 
-The default timeout for the script is 30 seconds. If you need more, pass a second parameter
+Script content based on C<current_test_script> value and can be fetched  through HTTP
+into the SUT or typed with C<type_string>. This depends on C<current_test_script> length,
```
coolo: that's a bit technical for a documentation - even for API documentation. I would leave it at 'if the command is typed or fetched through HTTP depends on various parameters, so if you need to have it typed because your SUT is offline, pass type_command => 1'

testapi.pm
```diff
@@ -897,19 +897,24 @@ sub script_sudo {
...
+The default timeout for the script is 30 seconds. If you need more, pass a second parameter.
 
 =cut
 
-sub script_output($;$) {
-    my ($current_test_script, $wait) = @_;
+sub script_output {
+    my ($current_test_script, $wait, %args) = @_;
     my $suffix = hashed_string("SO$current_test_script");
+    $args{type_command} ||= length($current_test_script) < 80;
```
coolo: To avoid having @okurz complain about magic numbers - add a comment above stating what you counted to be the default command line for curl

okurz: ☺

testapi.pm
```diff
@@ -919,6 +926,12 @@ sub script_output($;$) {
         type_string("\n", terminate_with => 'EOT');
         wait_serial("$suffix-0-");
     }
+    elsif ($args{type_command}) {
+        my $cat = "cat - > /tmp/script$suffix.sh << EOF;";
```
asmorodskyi: this approach is not working :( http://kimball.arch.suse.de/tests/279#step/wicked_before_test/28. 
Code which I am trying to execute is 
```
my $var1 =  script_output('cat /proc/stat',30,type_command => 1);
script_output("echo $var1",30,type_command => 1);
```
Any ideas what I am doing wrong ?

asmorodskyi:
Finally ready to be merged , at least for me :) 

Here is proof run http://kimball.arch.suse.de/tests/290#step/wicked_before_test/24 . 
Workers using testapi with this PR and in test code I have such lines :

```
my $var1 =  script_output('cat /proc/fb',30,type_command => 1);
script_output("echo $var1",30,type_command => 1);
```

WDYT ?

ldevulder:
This modification breaks my code: http://moon.qa.suse.cz/tests/2103#step/drbd_passive/5

I use this command:
`my $node_01_ip = script_output "host -t A $node_01 | awk '{ print \$NF }'";`

And now the `$NF` variable disappears... The same code is working well with the "old" version of script_output (which has other issues, I know).

ldevulder:
I tried with this syntax:
`my $node_01_ip = script_output("host -t A $node_01 | awk '{ print \$NF }'", 10, type_command => 1);`

But same issue.

coolo:
please add the quotes in your local instance to verify

asmorodskyi:
@ldevulder to really check that issue is in this patch please  
`my $node_01_ip = script_output("host -t A $node_01 | awk '{ print \$NF }'", 10, type_command => 0);`

then you will get into "old way" (first 'if')

coolo:
the bug *is* in this patch - see #886 for the fix

ldevulder:
Thanks! I will try the cooling fix this morning.

ldevulder:
Fix #886 confirmed: http://moon.qa.suse.cz/tests/2121#step/drbd_passive/5

PR #879, opened 2017-11-13
svirt: set NUMDISK if RAIDLEVEL is defined, Xen: switch from IDE to SCSI devices

mnowaksuse:


mnowaksuse:
Fixed typo in commit title.

backend/svirt.pm
```diff
@@ -37,7 +37,7 @@ sub do_start_vm {
     my ($self) = @_;
 
     my $vars = \%bmwqemu::vars;
-    $vars->{NUMDISKS} ||= 1;
+    $vars->{NUMDISKS} ||= defined($vars->{RAIDLEVEL}) ? 4 : 1;
```
foursixnine: I was about to ask what would happen when RAIDLEVEL=0 since we have that for functional tests until I noticed that this is about the svirt backend :P and we have kind of the same in the qemu backend

PR #882, opened 2017-11-23
[blocked] Add support for a timeout parameter for memory dumps

foursixnine:


testapi.pm
```diff
@@ -1539,7 +1539,7 @@ sub eject_cd {
 
 =head2 save_memory_dump
 
-  save_memory_dump(filename => undef, migration_speed => "4096m");
+  save_memory_dump([$filename, $max_downtime]);
```
coolo: this looks wrong. It's expecting a hash in my reading of the code

foursixnine: You are correct :), it's wrong

testapi.pm
```diff
@@ -1539,7 +1539,7 @@ sub eject_cd {
...
 
 Saves the SUT memory state using C<$filename> as base for the memory dump
 filename,  the default will be the current test's name.
```
okurz: I recommend to mention the default `max_downtime` in the description. Yes, it's annoying that we have to repeat that but if we insist on using perl we are stuck with POD then IMHO we have to do that

foursixnine:
So I have rebased locally, and found out that I'm getting:

`[Mon Dec  4 15:39:09 2017] [687:debug] capture loop failed panic: attempt to copy freed scalar 5579bc5a37b8 to 5579ba49c2d0 at /usr/lib/perl5/5.26.1/Carp.pm line 229.
`

foursixnine:
see #892

coolo:
hopefully no longer needed

PR #884, opened 2017-11-23
Distinguish modules name in a test run if it's a duplicate

mudler:
~~WIP because needs testing (currently changes are applied in staging instance) and i'd like to add unit tests.~~

[Success run](http://e122.suse.de/tests/2871) - [hostname](http://e122.suse.de/tests/2871#step/hostname%232/11) and [enable_usb_repo](http://e122.suse.de/tests/2871#step/enable_usb_repo%231/2) test
[Failed run](http://e122.suse.de/tests/2598#downloads) - logs/assets uploaded are unique
[Tests not using duplicate modules](http://e122.suse.de/tests/overview?distri=sle&version=12-SP3&build=0473&groupid=28) are unaffected by the change.

See related Progress issue: https://progress.opensuse.org/issues/10514

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -67,6 +67,7 @@ sub loadtest {
         # to all perl hardcore hackers: fuck off!
         $nr = $nr eq '' ? 1 : $nr + 1;
         bmwqemu::diag($fullname . ' already scheduled');
+        $test->{name} = join("#", $name, $nr);
     }
     $tests{$fullname . $nr} = $test;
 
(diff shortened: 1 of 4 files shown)


okurz:
> WIP because needs testing (currently changes are applied in staging instance) and i'd like to add unit tests.

I don't understand why you are so keen on this staging instance for so many cases? It certainly makes sense for stuff like scheduling, performance, multi-worker setup, architecture specific things but for single test cases that shouldn't be a necessity

foursixnine:
@okurz, so that you don't complain on stuff being broken, the staging was introduced.

mudler:
Danger

This change is not for a single test :)

PR #885, opened 2017-11-24
Fix snd2png check on TW: don't use PNM but PNG

coolo:
(the program's name doesn't imply it can do PNM anyway :), PNM contains
now a comment about the opencv version - png does not (for now)


okurz:
LGTM

PR #886, opened 2017-11-27
Fix quoting bug in type_command support

coolo:
See https://github.com/os-autoinst/os-autoinst/pull/878#issuecomment-347260739


okurz:
LGTM

PR #887, opened 2017-11-29
Add context to loadtest

richiejp:
Allow arbitrary parameters to be sent to the test from main.pm

https://progress.opensuse.org/issues/27124

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -26,13 +26,55 @@ use Socket;
 use IO::Handle;
 use POSIX '_exit';
 use cv;
+use Scalar::Util 'blessed';
 
 our %tests;        # scheduled or run tests
 our @testorder;    # for keeping them in order
 our $isotovideo;
 
+=head1 Introduction
+
+OS Autoinst decides which test modules to run based on a distribution specific
+script called main.pm. This is either located in $vars{PRODUCTDIR} or
+$vars{CASEDIR} (e.g. <distribution>/products/<product>/main.pm).
+
+This script does not actually run the tests, but queues them to be run by
+autotest.pm. A test is queued by calling the loadtest function which is also
+located in autotest.pm. The test modules are executed in the same order that
+loadtest is called.
+
+=cut
+
+=head2 loadtest
+
+	loadtest(<string>, [ name => <string>, run_args => <OpenQA::Test::RunArgs> ]);
+
+Queue a test module for execution by the test runner. The first argument is
+mandatory and specifies the Perl module name containing the test code to be run.
+
+The next two arguments are optional and rarely used. First there is name which
+can be used to give the test a different display name from the Perl source
+file.
+
+Then there is the run_args object, which must be a subclass of
+OpenQA::Test::RunArgs. This is passed to the run() method of the test module
+when it is executed. This is useful if you need to load the same test module
+multiple times within a single test, but with different parameters each time.
+
+Usually get_var and set_var are used to pass parameters to a test. However if
+you use set_var multiple times inside main.pm then the final value you set
+will be the one seen by all tests. Regardless of whether the tests were loaded
+before or after the variable was set.
+
+Both optional arguments were created for integrating a third party test suites
+or test runners into OpenQA. In such cases the same test module may be
+dynamically queued multiple times to execute different test cases within the
+third party test suite.
+
+=cut
+
 sub loadtest {
-    my ($script) = @_;
+    my ($script, %args) = @_;
     my $casedir = $bmwqemu::vars{CASEDIR};
 
     unless (-f join('/', $casedir, $script)) {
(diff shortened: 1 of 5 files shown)


autotest.pm
```diff
@@ -32,7 +32,7 @@ our @testorder;    # for keeping them in order
 our $isotovideo;
 
 sub loadtest {
```
okurz: I guess that method deserves a documentation now and I know you are good at it :)

coolo: I had the same idea thought - that this needs extension to the documentation. And then I realized we have no documentation about main.pm at all :(

coolo: So I would merge this without docu right now - but I first need to understand the change myself. So docu *would* help :)

richiejp: I actually don't mind documenting this once I have addressed @mudler's comments here: https://progress.opensuse.org/issues/27124

richiejp:
Coveralls is wrong: this is definitely covered by the unit tests:
https://coveralls.io/builds/14547955/source?filename=%2Fhome%2Ftravis%2Fbuild%2Fos-autoinst%2Fos-autoinst%2Fautotest.pm#L103

coolo:
coveralls being wrong is the reason we switched to codecov - but so far only set it up for openQA as os-autoinst coverage is suboptimal anyway :)

coolo:
thanks for the writeup. There is a small problem though: we only have http://open.qa/api/testapi/ published - shall we fake the autotest documention into the testapi? Or anyone another idea?

t/08-autotest.t
```diff
@@ -84,6 +86,34 @@ is($died,      0, 'start+next+start should not die');
 is($completed, 1, 'start+next+start should complete');
 @sent = [];
 
+my $ctx = OpenQA::Test::Context->new();
+stderr_like(
+    sub {
+        autotest::loadtest("tests/context.pm", name => 'alt_name', context => $ctx);
```
mudler: I think _'name'_ should be a context attribute, and not outside of it - otherwise it's not clear if _'name'_ is an attribute of the **test** or the **context**.

Aside: @coolo wdyt about allowing 'everything' to be passed into loadtest?

t/fake/tests/context.pm
```diff
@@ -0,0 +1,11 @@
+use 5.018;
+use base 'basetest';
+
+sub run {
+    my ($self) = @_;
```
mudler: Wouldn't be practical if we can access to the context either by:

```perl 
sub run {
     my ($self, $ctx) = @_;
     ....
}
```

or, if inside the test object:


```perl 
sub run {
     my ($self) = @_;
     $ctx = $self->ctx(); # or also with an alias to ->context();
}
```

t/08-autotest.t
richiejp: It is the test's display name.

It is also the LTP test name, but they could be different.

t/fake/tests/context.pm
richiejp: I am happy to go with whatever people want.

t/08-autotest.t
richiejp: >  i'm not following what a freedom/empty implementation of context class could lead to

@foursixnine wanted type checking? It should stop some irritating Perl errors by making sure you have passed the right thing to the test. I can't think of any methods to add to it at this time though.

mudler:
What about a function inside testapi, let's say _context()_ that returns the context object that is inside the ```$autotest::current_test``` ?

coolo:
I'm a bit worried that it becomes just too unclear what this context is all about. The documentation of @richiejp basically says "don't use it" :)

In every other circumstance we pass context in %vars - and the documentation doesn't make it clear enough when you would need this. Possibly context() is just a misleading word for it. What we try to express here are in normal circumstances just arguments to run(), no? Can't we map it like that?

richiejp:
Well, maybe we could call it Procedural Context/Args or Dynamic Context/Args? Because it is only necessary when something needs to be calculated inside main.pm at runtime.

I think autotest probably just needs to have its own documentation generated. Then we could link to it in the test writing guidelines section on main.pm (which might not exist yet).

mudler:
context was suggested just to _avoid_ passing arbitrary data to loadtest - if we accept everything it looses it's original purpose (and you can pass whatever object from tests)

mudler:
i think it's just a matter of deciding if we want to proxy the arguments feeded to loadtest to https://github.com/os-autoinst/os-autoinst/blob/master/basetest.pm#L321 or not

coolo:
I guess I just don't understand what makes up a 'context' for you. It's a very vague phrase in my book.

richiejp:
But it _is_ arbitrary data. I can't enumerate every possible use of this interface and create a class definition for it because I only know how the LTP test suite needs to use it. Other users may need to pass completely different things.

We could just pass (proxy?) any trailing arguments from loadtest() to run(), but then it is difficult to add more optional arguments to loadtest. So I would prefer to put them in an object; we could call it RunArgs which is more descriptive than Context.

coolo:
this might be a small change, but I tell you honestly: now it suddenly 900% more sense. I'm fine with merging

richiejp:
OK, well I will rebase then to remove the context property history.

PR #888, opened 2017-12-01
[POC] Wrap HMP commands with QMP's human-monitor-command

richiejp:
It is possible to call HMP commands from QMP using the human-monitor-command
wrapper. This is not recommended by upstream, but may be an improvement over
calling HMP commands directly.

See: https://git.qemu.org/?p=qemu.git;a=blob;f=qapi-schema.json;h=18457954a841ae4b1de98c0b4687636196d7fb7f;hb=HEAD#l1226

If you would like this to be merged quickly then someone else will need to pick it up, test it and clean it up. It seems to work on Tumbleweed at least as well as calling loadvm and savevm over an actual HMP socket.

The alternative is to implement the commands as native QMP commands upstream, which may not be too difficult (from my brief look), but would probably take a while.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -33,6 +33,7 @@ use autodie ':all';
 use Try::Tiny;
 use osutils qw(find_bin gen_params qv);
 use List::Util 'max';
+use Data::Dumper;
 
 sub new {
     my $class = shift;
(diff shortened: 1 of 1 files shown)


okurz:
so what is the benefit of calling HMP commands wrapped in QMP over native HMP?

coolo:
for me there would be an advantage if we could stop monitoring hmp and live with one socket for all. I understand this PR is a POC, but it does look promising

richiejp:
Also with QMP, the transport is at least well defined even if the actual command names[1] and responses are not. HMP could start sending new control characters or add autocomplete which might confuse our parsing.

[1] I am specifically talking about HMP commands being sent over QMP. All QMP commands are well defined.

coolo:
how about https://github.com/os-autoinst/os-autoinst/pull/889 then?

coolo:
superseded

PR #890, opened 2017-12-06
Fix error checking on load_snapshot and remove sleep

richiejp:
Fail if $rsp is a falsy value. Also drop the stop, cont and sleep because
there does not appear to be any reason for them. They just cause a 1 second
load to take 10 seconds instead.

@coolo

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -265,12 +265,9 @@ sub load_snapshot {
     my ($self, $args) = @_;
     my $vmname = $args->{name};
     my $rsp    = $self->handle_qmp_command(_wrap_hmc("loadvm $vmname"));
-    if ($rsp || $rsp->{return} ne '') {
+    if (!$rsp || $rsp->{return} ne '') {
         die "Could not load snapshot \'$vmname\': " . Dumper($rsp);
     }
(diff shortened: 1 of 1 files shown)


coolo:
the sleep was added by a08de6a0daad78520d71869a0b284978c101d288 - looks more like a debugging aid back then.

the stop+cont was there from the beginning, no idea if it was necessary back then.

okurz:
Wasn't the `sleep 10` there to give the SUT some settling time before trying to hammer it with the next test executions? I suggest this change to be properly tested with some real-world openQA tests where loading from snapshots appears

coolo:
So a08de6a0daad78520d71869a0b284978c101d288 just moved it indeed

coolo:
2fd59d6d6338708d69b4cfc509920395d7d8419e added it - also without satisfying explanation. If the SUT needs time to settle after loadvm, this would be a qemu problem, no?

okurz:
hm, don't know. Can the SUT know about it? Maybe it behaves like a suspend-resume cycle? or a time-jump for the SUT? It *could* be a linux problem or even an openSUSE/SLE one.

coolo:
and it could be a stale sleep call :)

PR #891, opened 2017-12-08
Remove empty miniatures.

dasantiago:
poo#23422.

When an invocation to assert_script_run doesn't change the screen, the test result would
allocate the miniature for the screen capture, but since the screen didn't change, it
would became a empty space, causing the webinterface to overlap both areas.

The change, now "allocates" the space only when the screen can be captured.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -409,16 +409,16 @@ sub record_serialresult {
 
     $string //= '';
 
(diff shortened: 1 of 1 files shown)


dasantiago:
Test run: http://tragicbox.suse.cz/tests/329#step/userspace_postfix/7

PR #892, opened 2017-12-08
Memory dumps are only callable from post_fail-hook

foursixnine:
Forbid calling save_memory_dump from testapi due to boo#1072008 and workaround for boo#1072000

See comments on https://progress.opensuse.org/issues/19390 for more details.
See also  https://git.io/vb4lq and https://travis-ci.org/foursixnine/os-autoinst/builds/313704345#L2032

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -218,19 +218,28 @@ sub save_memory_dump {
 
     die(sprintf("Migration failed: desc: %s, class: %s, stopped", $rsp->{error}->{desc}, $rsp->{error}->{class})) if ($rsp->{error});
 
-
+    my $migration_starttime = gettimeofday;
+    my $execution_time      = gettimeofday;
+    my $max_execution_time  = 240;            # We need to wait for qemu, since it will not honor timeouts
     do {
 
-        sleep 0.5;    #We want to wait a decent amount of time, a file of 1GB will be
-                      # migrated in about 40secs with an ssd drive. and no heavy load.
+        sleep 0.5;                            #We want to wait a decent amount of time, a file of 1GB will be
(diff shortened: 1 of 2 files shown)


coolo:
If the call will still timeout/fail - what is the point in offering it at all?

foursixnine:
@coolo I'm adding the cancel-migration command if the call is taking > 120 secs.

backend/qemu.pm
```diff
@@ -218,19 +218,27 @@ sub save_memory_dump {
...
 
         sleep 0.5;    #We want to wait a decent amount of time, a file of 1GB will be
                       # migrated in about 40secs with an ssd drive. and no heavy load.
+        $execution_time = gettimeofday - $migration_starttime;
         $rsp = $self->handle_qmp_command({execute => "query-migrate"});
 
         diag "Migrating total bytes:     \t" . $rsp->{return}->{ram}->{total};
         diag "Migrating remaining bytes:   \t" . $rsp->{return}->{ram}->{remaining};
+        # 240 seconds should be ok for 4GB
+        if ($execution_time > 240) {
```
foursixnine: I don't like magic numbers, but :/

foursixnine:
@agraf :)

backend/qemu.pm
okurz: better extract it into a constant with a helpful name

coolo: I suggest a $wtf_why_is_this_still_not_done = 240

okurz: worksforme :)

foursixnine: set it to $max_execution_time :)

PR #896, opened 2017-12-14
Retidy - perltidy 20171214 changed output

coolo:


coolo:
Unfortunately this breaks with old tidy

asmorodskyi:
https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Leap_42.3/x86_64

coolo:
I expected nothing else

PR #897, opened 2017-12-16
Add option to upload assets on demand even in failed jobs

okurz:
Example log output:

```
[Sat Dec 16 08:20:34 2017] [18363:debug] >>> testapi::_check_backend_response: match=bootloader timed out after 30
[Sat Dec 16 08:20:35 2017] [18363:debug] test boot failed
[Sat Dec 16 08:20:35 2017] [18357:debug] isotovideo done
[Sat Dec 16 08:20:35 2017] [18357:debug] BACKEND SHUTDOWN
[Sat Dec 16 08:20:36 2017] [18368:debug] sending magic and exit
[Sat Dec 16 08:20:36 2017] [18357:debug] received magic close
[Sat Dec 16 08:20:36 2017] [18357:debug] preparing hdd 1 for upload as opensuse-Tumbleweed-x86_64-20170505-textmode@64bit.qcow2 in qcow2
[Sat Dec 16 08:20:36 2017] [18357:debug] running nice ionice qemu-img convert -O qcow2 raid/l1 assets_public/opensuse-Tumbleweed-x86_64-20170505-textmode@64bit.qcow
```

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -448,42 +448,56 @@ if (!$return_code) {
 # read calculated variables from backend and tests
 bmwqemu::load_vars();
 
-# mark hard disks for upload if test finished
-if (!$return_code && $command_handler->test_completed && $bmwqemu::vars{BACKEND} eq 'qemu') {
+sub store_asset {
+    my ($index, $name, $dir) = @_;
+    $name =~ /\.([[:alnum:]]+)$/;
+    my $format = $1;
+    return {hdd_num => $index, name => $name, dir => $dir, format => $format};
+}
+
+sub handle_generated_assets {
+    # mark hard disks for upload if test finished
+    return unless !$return_code && $command_handler->test_completed && ($bmwqemu::vars{BACKEND} eq 'qemu');
     my @toextract;
-    for my $i (1 .. $bmwqemu::vars{NUMDISKS}) {
-        my $dir  = 'assets_private';
-        my $name = $bmwqemu::vars{"STORE_HDD_$i"} || undef;
-        unless ($name) {
-            $name = $bmwqemu::vars{"PUBLISH_HDD_$i"} || undef;
-            $dir  = 'assets_public';
-        }
-        next unless $name;
-        $name =~ /\.([[:alnum:]]+)$/;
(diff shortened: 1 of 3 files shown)


coolo:
I have my doubts that this works. Because if the test suite is not complete, the vm is most likely not shutdown.

okurz:
This is why in this case I do not check for the VM to be shutdown. It might not work in all cases but in more than none :) I did the check that you can see above where it kind of worked. At least it really uploaded a qcow file.

coolo:
but boot in this case was fatal? Because non-fatal failures are rolled back - and publishing the hdd for debugging purposes as such worthless.

okurz:
Depends on the milestone and what the investigator actually wants to debug, e.g. If it's not the last failing module but often just the installed system with all the packages,  versions,  patterns etc.

coolo:
I still would opt to extend the usability of this feature - and make all tests fatal, so they end up uploading their failed status - and as we can't shutdown the VM, but you have to save *something* - we could save a snapshot 'lastfail'. So that you can 'boot' the VM with -loadvm 'lastfail'.

okurz:
Well, sounds useful. Well take me [some_time:eternity[ though ;)

foursixnine:
@coolo I take that you still would like this?

coolo:
the story is different after @richiejp's changes. So it needs to be reevaluated from scratch :)

Martchus:
So it likely doesn't even make sense to rebase this branch. Should we close it to keep the list of open PR shorter?

okurz:
I thought it made sense to rebase. Please take a look again in the current updated state if you like it better. I also added an automatic test.

okurz:
Hm, the tests fail to execute because qemu is called in 14-isotovideo and kvm is not available. How do we handle that for the full stack test?

Martchus:
We set the test variable `QEMU_NO_KVM` to `1`.

okurz:
all set. Ready to merge?

Martchus:
Since @coolo had initially some objections I leave clicking the merge button to him.

PR #898, opened 2017-12-19
Check runcmd return status

mitiao:
Especially care return status for extract assets
See: https://progress.opensuse.org/issues/17902

### DIFF
--- a/backend/pvm.pm
+++ b/backend/pvm.pm
@@ -22,7 +22,7 @@ require IPC::System::Simple;
 use autodie ':all';
 use File::Basename;
 use Digest::MD5 'md5_hex';
-use osutils qw(dd_gen_params gen_params);
+use osutils qw(dd_gen_params gen_params runcmd);
 
 sub new {
     my $class      = shift;
@@ -43,11 +43,6 @@ sub do_start_vm {
     $self->unlink_crash_file;
     $self->start_lpar();
     return {};
-}
-
-sub runcmd {
-    diag "running " . join(' ', @_);
-    return system(@_);
 }
 
 sub do_extract_assets {
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -31,7 +31,7 @@ use bmwqemu qw(fileContent diag save_vars);
 require IPC::System::Simple;
 use autodie ':all';
 use Try::Tiny;
-use osutils qw(find_bin gen_params qv);
+use osutils qw(find_bin gen_params qv runcmd);
 use List::Util 'max';
 use Data::Dumper;
 
@@ -280,12 +280,6 @@ sub load_snapshot {
     return $rsp;
 }
 
-sub runcmd {
-    diag "running " . join(' ', @_);
-    local $SIG{CHLD} = 'IGNORE';
-    return CORE::system(@_);
-}
-
 sub do_extract_assets {
     my ($self, $args) = @_;
     my $hdd_num = $args->{hdd_num};
@@ -293,7 +287,7 @@ sub do_extract_assets {
     my $img_dir = $args->{dir};
     my $format  = $args->{format};
     if (!$format || $format !~ /^(raw|qcow2)$/) {
-        bmwqemu::diag "do_extract_assets: only raw and qcow2 formats supported $name $format";
+        die "do_extract_assets: only raw and qcow2 formats supported $name $format";
     }
     elsif (-f "raid/l$hdd_num") {
         bmwqemu::diag "preparing hdd $hdd_num for upload as $name in $format";
@@ -314,7 +308,7 @@ sub do_extract_assets {
         }
     }
     else {
-        bmwqemu::diag "do_extract_assets: hdd $hdd_num does not exist";
+        die "do_extract_assets: hdd $hdd_num does not exist";
     }
 }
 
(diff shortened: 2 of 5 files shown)


osutils.pm
```diff
@@ -74,4 +75,18 @@ sub qv($) {
...
+    bmwqemu::diag "running " . join(' ', @_);
+    my %children;
+    local $SIG{CHLD} = sub {
+        local ($!, $?);
+        while ((my $pid = waitpid(-1, WNOHANG)) > 0) {
+            delete $children{$pid};
+            cleanup_child($pid, $?);
+        }
+    };
+    return CORE::system(@_);
```
okurz: Wouldn't it be safer to throw an exception if the call fails so we don't need to check the exit code explicitly on every call?

backend/qemu.pm
```diff
@@ -271,20 +271,14 @@ sub load_snapshot {
     return $rsp;
 }
 
-sub runcmd {
-    diag "running " . join(' ', @_);
-    local $SIG{CHLD} = 'IGNORE';
-    return CORE::system(@_);
```
coolo: You are reverting https://github.com/os-autoinst/os-autoinst/pull/534 basically - so I object

mitiao: I will take `system` from CORE package back and avoid autodie

osutils.pm
```diff
@@ -74,4 +77,20 @@ sub qv($) {
...
 
+# runcmd use SIGCHLD handler to wait all child processes at exit and reap zombies
+sub runcmd {
+    diag "running " . join(' ', @_);
+    my %children;
+    local $SIG{CHLD} = sub {
+        local ($!, $?);
+        while ((my $pid = waitpid(-1, WNOHANG)) > 0) {
+            delete $children{$pid};
+            cleanup_child($pid, $?);
```
mudler: I can't find ```cleanup_child()``` in the [codebase,](https://github.com/os-autoinst/os-autoinst/search?utf8=%E2%9C%93&q=cleanup_child&type=) am i missing something?

isotovideo
```diff
@@ -535,7 +535,11 @@ if (!$r && $completed && (my $nd = $bmwqemu::vars{NUMDISKS}) && ($bmwqemu::vars{
     }
     else {
         for my $asset (@toextract) {
-            $bmwqemu::backend->extract_assets($asset);
+            eval { $bmwqemu::backend->extract_assets($asset); };
```
mudler: you might want ```local ($@)``` as well

backend/pvm.pm
```diff
@@ -45,11 +45,6 @@ sub do_start_vm {
     return {};
 }
 
-sub runcmd {
```
mudler: +1 for grouping common function, but afaics from the changes it's used in backends, probably best place for it should be in [baseclass](https://github.com/os-autoinst/os-autoinst/blob/master/backend/baseclass.pm)

mudler: Note aside: as functions in osutils are meant to be re-used, they should come up with unit tests

osutils.pm
mitiao: seems I can't get pid of `system` with `waitpid`, is it possible that Perl reap system subprocesses itself?
if so to use `local $SIG{CHLD} = 'DEFAULT'` should be ok here?

backend/pvm.pm
mitiao: I moved it to baseclass

t/13-osutils.t
```diff
@@ -130,4 +130,15 @@ subtest find_bin => sub {
 
 };
 
+subtest runcmd => sub {
+    use osutils 'runcmd';
+
+    my @cmd = ('qemu-img', 'create', '-f', 'qcow2', 'image.qcow2', '30G');
```
foursixnine: 30G is kind of a big number, even if it takes very short time :P can you switch it to 1G?

mitiao: Fine. I just thought the initial size of qcow2 format is quite small

mitiao:
I had a test and seems it didn't display exit status of reaped childs if `system` called :( 
Also I had a few runs of failure with logs:
http://10.67.17.71/tests/109/file/autoinst-log.txt
http://10.67.17.71/tests/112/file/autoinst-log.txt

mudler:
Yeah, because maybe what you are trying to call with ```system()``` is furthermore forking - such case you should use process groups. But that won't help if a process among them it's also declaring another process group and detaching more further.

mitiao:
Thanks for suggestion that break my hesitation to use different approach, work to fix.

mitiao:
runcmd failed test: http://10.67.17.71/tests/197/file/autoinst-log.txt
runcmd ok test: http://10.67.17.71/tests/196/file/autoinst-log.txt

osutils.pm
```diff
@@ -21,13 +21,19 @@ use warnings;
 use Carp;
 use base 'Exporter';
 use Mojo::File 'path';
+use bmwqemu 'diag';
+use POSIX ':sys_wait_h';
+use Symbol 'gensym';
+use IPC::Open3;
```
okurz: Do we have that package in our distris? Couldn't find it. Is it included in perl?

foursixnine: @okurz Yes, it is part of the core perl distribution

mudler: Take your monday morning coffee, take a breath and then, luke, use the [force](http://www.letmegooglethat.com/?q=Perl+core+module+IPC%3A%3AOpen3)

foursixnine:
@coolo @mudler I think we can merge this already right?

PR #900, opened 2017-12-21
Skip save_tmp_file full stack test if no network

rwx788:
Changes introduced in
[PR#877](https://github.com/os-autoinst/os-autoinst/pull/877#) break
build for fedora, as no network is there. Skip test which relies on
network connection if cannot ping default gateway.

### DIFF
--- a/t/data/tests/tests/modify_and_upload_file.pm
+++ b/t/data/tests/tests/modify_and_upload_file.pm
@@ -41,10 +41,15 @@ sub run {
     my $url = autoinst_url . '/files/modified.xml';
     $content =~ s/PASSWORD/nots3cr3t/g;
     save_tmp_file('modified.xml', $content);
-    # Verify that correct file is downloaded
-    assert_script_run("wget -q $url");
-    script_run "echo '72d2c15cb10535f36862d7d2eecc8a79  modified.xml' > modified.md5";
-    assert_script_run("md5sum -c modified.md5");
-
-    type_string("echo save_tmp_file returned expected file\n");
+    # Ping default gateway to see if network is available and skip test otherwise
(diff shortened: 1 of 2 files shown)


t/data/tests/tests/modify_and_upload_file.pm
```diff
@@ -41,10 +41,14 @@ sub run {
...
     $content =~ s/PASSWORD/nots3cr3t/g;
     save_tmp_file('modified.xml', $content);
-    # Verify that correct file is downloaded
-    assert_script_run("wget -q $url");
-    script_run "echo '72d2c15cb10535f36862d7d2eecc8a79  modified.xml' > modified.md5";
-    assert_script_run("md5sum -c modified.md5");
-
-    type_string("echo save_tmp_file returned expected file\n");
+    # Ping default gateway to see if network is available and skip test otherwise
+    if (script_run("ping -q -w 1 -c 1 `ip r | grep default | cut -d ' ' -f 3`")) {
```
nicksinger: Depending on the definition of "no network" this could also break. E.g. if no NIC is available or the interface does not get configured then the "ip r s" command will not contain a default gateway.
IMHO it'd be better to first check the output of `ip r | grep default | cut -d ' ' -f 3` before pinging it.

However, the main purpose of this ("test does not get executed on no-network-systems") should still work.

rwx788: I guess if this command breaks for any reason, it means no network is there, if there is no default route, ping will return non-zero code, which does the trick. But I don't like this solution much anyway, we should not run tests at all then skipping it almost silently =)

rwx788:
Original issue resolved, no changes required

PR #902, opened 2017-12-29
Polling the read socket in the capture loop

dasantiago:
poo#16418

Before, you can check in the autoinst file (**NOTE:** Huge autoinst files), that it got stuck in the restart 8 mins until i restarted:
- http://tragicbox.suse.cz/tests/204/

After: 

- http://tragicbox.suse.cz/tests/240
- http://tragicbox.suse.cz/tests/242

[After the fix patch installed:](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4107)

- http://tragicbox.suse.cz/tests/248

This change will make the capture loop more robust by polling the read socket to
detect half-open connections
When a connection is closed only on the other end if we try to read from it, an
error isn't yielded (not like how it happens with writes), creating an infinite
capture loop

There are three ways to solve this problem:
+ Send a message either to the application protocol (null message) or to the
  application protocol framing (an empty message)
  Disadvantages: Requires changes on both ends of the communication.

+ Polling the connections (Note: This is how HTTP servers work when dealing with
   persistent connections)
   Disadvantages: False positives

+ Change the keepalive packet settings
  Disadvantages: TCP/IP stacks are not required to support keepalives.

The first and last option, are a bit troublesome as we need to deal with restarts.
The first option also requires changes in both ends - Besides having to
enable/disabe the _SIGPIPE_ capture on demand, when we have a SSH connection we
start hitting the error "bad packet length".

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -154,6 +154,8 @@ sub run_capture_loop {
     }
 
     eval {
+        # Time slot buckets
+        my $buckets = {};
         while (1) {
 
             last if (!$self->{cmdpipe});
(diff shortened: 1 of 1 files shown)


coolo:
your description is sound, your tests look good, but give me some time to review the code - I'm on vacation and this is complex :)

Soulofdestiny:
fyi:
I've tried this patch to debug the s390 shutdown issue -> https://progress.opensuse.org/issues/19350
The test incompletes now even earlier, during the initial reboot with 
`DIE The console isn't responding correctly. Maybe half-open socket? at /home/os-autoinst/backend/baseclass.pm line 238.`, see http://opeth.suse.de/tests/6192

not sure though if the patch is doing the correct thing but the tests need adoption

Soulofdestiny:
increasing the Timeout in L237 from 10 to 20 helped to pass the initial reboot 
(` 237                     if (update_time_bucket($buckets, 20, fileno $fh) > 5_000) { `)

now it's failing in shutdown with the same expected error message `Maybe half-open socket?`

so 10 seconds might be a bit too low

dasantiago:
That's the point. The point is to die quickly and not to hang in an infinite loop.

You need to deploy Michal's PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4107

Since the problem seems to happen randomly, maybe you hit it on the first try? All my tests shown that 10 seconds were more than enough.

PR #903, opened 2018-01-02
Ignore WORKER_HOSTNAME for qemu explicitly

asmorodskyi:


okurz:
In commit message and SR subject s/excplicitly/explicitly/ . You are not stating what this commit provides. Please describe in more detail what problem is fixed by this change.

asmorodskyi:
@okurz - done

SergioAtSUSE:
LGTM, but PR subject line still has the typo.

PR #904, opened 2018-01-12
Display interface version when using --version

dasantiago:


isotovideo
```diff
@@ -87,15 +87,15 @@ sub usage {
 
 sub version {
     my $thisversion = qx{git rev-parse HEAD};
```
foursixnine: Just for future reference, this line is modified by a spec file when we're packaging: https://build.opensuse.org/package/view_file/devel:openQA/os-autoinst/os-autoinst.spec

dasantiago: @foursixnine THanks! I didn't knew that...

PR #906, opened 2018-01-12
Increase the limits for checking the VNC console. 

dasantiago:
The limits, although were big enough for most of the tests, were not big enough
for xen tests or when the machines were under heavy load.

Limits increased 3 times.

It's now possible to control this values by using the vars _CHKSEL_RATE_WAIT_TIME and _CHKSEL_RATE_HITS

The checking function was also changed to make sure that:

-  will wait the CHKSEL_RATE_WAIT_TIME. Before if it hit the _CHKSEL_RATE_HITS, it would consider dead

- All the FD in the bucket need to achieve the _CHKSEL_RATE_HITS, not just one. This in conjunction with the previous point will allow the system to recover if another FD suddenly starts responding.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -155,7 +155,10 @@ sub run_capture_loop {
 
     eval {
         # Time slot buckets
-        my $buckets = {};
+        my $buckets         = {};
+        my $wait_time_limit = $bmwqemu::vars{_CHKSEL_RATE_WAIT_TIME} // 30;
+        my $hits_limit      = $bmwqemu::vars{_CHKSEL_RATE_HITS} // 15_000;
+
         while (1) {
 
             last if (!$self->{cmdpipe});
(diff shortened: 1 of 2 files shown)


mnowaksuse:
Great idea to have it configurable. I'd like the variable name to be trimmed of the `OS_AUTOINST_` prefix, it's quite unusual to have it and it's quite long anyways. Also document the variables somewhere, e.g. in `docs/backend_vars.asciidoc`.

coolo:
But as I mentioned in other PRs before, I want (new) variables that affect isotovideo's internals to be prefixed. For EXIT_AFTER_SCHEDULE we agreed on a _ - so I'd be fine with _BUCKET_TIME_SIZE

The question I have though is: how many more buckets will we have over time. So perhaps instead of _BUCKET, make the variable short and hard to spell, but document it. As in _SKBTS - ok, a bit longer :)

coolo:
I have second thoughts, if making it configurable is smart at all. 30 seconds are a lot - still not enough for xen?

coolo:
But let's merge it with the _ prefix it should be clear it's for rare cases

dasantiago:
Just to inform that the tests are now passing:
https://openqa.suse.de/tests/1394398

backend/baseclass.pm
```diff
@@ -234,7 +237,8 @@ sub run_capture_loop {
                 if (fileno $fh && fileno $fh != -1) {
                     # Very high limits! On a working socket, the maximum hits per 10 seconds will be around 60.
                     # The maximum hits per 10 seconds saw on a half open socket was >100k
-                    if (update_time_bucket($buckets, 10, fileno $fh) > 5_000) {
+                    # if (update_time_bucket($buckets, $bucket_time_size, fileno $fh) > $bucket_hit_size) {
```
okurz: what's with this line now?

dasantiago: it's commented, it shouldn't be there.

PR #908, opened 2018-01-15
Add variables to change ipmi behavior in test

dzedro:
Servers in Prague assigned for SES testing have configuration that needs this modifications.
e.g.
* hornet is master server and has 2 network cards and pretty fast boot and tent to fail with existing sleep 2 after status check http://10.100.12.155/tests/493/file/autoinst-log.txt
* 4 nodes have 7 network cards, boot takes much longer and there are windows without change, much longer than 4 seconds stall time. Configurable stall time is useful anyway
http://10.100.12.155/tests/523
* Shutting down the machines after installation is not needed, after installation continues manual testing or openQA testing in future. Configurable shutdown of ipmi machine after the test is done is also useful. Right ? :smile: 

Added variables leave default values if unset.
Tests: http://10.100.12.155/tests

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -111,7 +111,9 @@ sub do_start_vm {
 sub do_stop_vm {
     my ($self) = @_;
 
-    $self->ipmitool("chassis power off");
+    if (!$bmwqemu::vars{IPMI_DO_NOT_POWER_OFF}) {
+        $self->ipmitool("chassis power off");
+    }
     $self->deactivate_console({testapi_console => 'sol'});
     return {};
 }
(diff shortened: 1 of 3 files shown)


coolo:
Please document the new variables

dzedro:
sorry, completely forgot about the variables docu. updated

backend/ipmi.pm
```diff
@@ -76,18 +76,18 @@ sub restart_host {
 
     $self->ipmitool("chassis power off");
     while (1) {
+        sleep(3);
```
okurz: I know you are just moving a sleep but could you please add a comment why it's necessary?

PR #910, opened 2018-01-18
Try to load OpenQA::Parser if present and parse extra test files. Add also support for xUnit and LTP

mudler:
This adds ```parse_extra_log``` to testapi, it accepts the format as first argument, and the file to parse as second. e.g. ```parse_extra_log( XUnit => 'result.xml' );```
In this PR there is no deprecation of ```parse_junit_log```, but ```OpenQA::Parser``` ( https://github.com/os-autoinst/openQA/pull/1545 ) is used to generate test results, so it will have impact on e.g. slenkins tests that are using this testapi function.

Caveat: we have to depend on the openQA libpath for that. Since this is loaded dynamically, only tests using this testapi function would die if openQA-common package is not installed on the machines running the test.

I didn't directly tested XUnit parsing with a format from avocado or such similar frameworks - the format spec was infered by [xunit viewer tests files](https://github.com/lukejpreston/xunit-viewer/tree/master/data), and specific unit tests were written from that,  so it might require changes to the XUnit parser, or create specific implementations per-framework as fits our goals.

Test runs with this change(slenkins) : http://e122.suse.de/tests/overview?distri=sle&version=12-SP3&build=0473&groupid=28 - ping @thehejik 

See related Progress issue: https://progress.opensuse.org/issues/16076

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2017 SUSE LLC
+# Copyright © 2012-2018 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -35,6 +35,7 @@ use MIME::Base64 'decode_base64';
 use Scalar::Util 'looks_like_number';
 
 require bmwqemu;
+use constant OPENQA_LIBPATH => '/usr/share/openqa/lib';
 
 our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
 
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -35,6 +35,7 @@ use MIME::Base64 'decode_base64';
 use Scalar::Util 'looks_like_number';
 
 require bmwqemu;
+use constant OPENQA_LIBPATH => '/usr/share/openqa/lib';
```
foursixnine: I wonder if there was a better way, but in the long run, since it's being cached if the lib is not found, shouldn't then be a major problem

mudler: This is how it looks like currently when it can't find the module : http://e122.suse.de/tests/5576#step/slenkins_control/16

Yeah, it's not an elegant solution, wondering if at this point we could just add openqa libpath by default in @INC, but that means the whole os-autoinst would depend on openQA-common

thehejik:
So you've moved all the logic from parse_junit_log into new OpenQA::Parser. Cool, the testapi now seems to be way cleaner than before. I'm totally ok with this change if our slenkins tests keeps working :-)

mudler:
Thanks @thehejik for confirming results are looking good - @foursixnine care to merge?

foursixnine:
Let's merge

PR #913, opened 2018-01-25
Extract function which will retrun VM's host IP

asmorodskyi:


testapi.pm
```diff
@@ -1806,6 +1806,14 @@ sub diag {
     return bmwqemu::diag(@_);
 }
 
+=head2 get_host_ip
+    Return VM's host IP
+    in a kvm instance you reach the VM's host under 10.0.2.2
+=cut
+sub get_host_ip {
```
coolo: as you have no set_ function, the get_ prefix is just noise.

asmorodskyi: Fixed

testapi.pm
```diff
@@ -1806,6 +1806,14 @@ sub diag {
     return bmwqemu::diag(@_);
 }
 
+=head2 host_ip
+    Return VM's host IP
+    in a kvm instance you reach the VM's host under 10.0.2.2
```
mnowaksuse: I am a bit confused here. If we are looking for VM's host, than in case of `svirt` it is not `WORKER_HOSTNAME` but `VIRSH_HOSTNAME`. If you are actually looking for an entry point to worker, then the name `host_ip` does not fit.

okurz: But asmorodskyi is just extracting the method, right?

PR #915, opened 2018-01-30
[WIP] qemu: Increase memory dump performance

richiejp:
This is mainly to help determine if migration can be made reliable for the benefit of snapshots. I will probably follow this up with another commit which enables performance counters during snapshots and memory dumps. Currently this requires a new package which I will try avoiding, but it is a pain.

Makes three changes to the memory dump:

1) Uses the file descriptor URI with the migrate command instead of exec. This
means that QEMU is writing the data directly to a file instead of writing it
to a pipe to /bin/sh which then relays it to gzip which tries to re-compress it
and write it back to a pipe to sh which then writes it to a file.

Note that QEMU compresses the image internally by default using zlib. The
compression level can be increased, but it did not result in a smaller file
size during my testing.

2) Set max bandwidth to 2^31 to make sure that the transfer speed is not being
limited by QEMU. It seems unlikely that this is a problem for us, but I am not
sure if QEMU can set this value automatically.

3) Make sure the VM is frozen before beginning the migration. QEMU will try to
copy most of the RAM while the VM is still running which requires it to track
dirtied RAM and resend it. Eventually it freezes the machine to finish the
copy. However if the guest is busy then this could increase the size of the
dump and cause it to take longer. It also means QEMU has to maintain a bitmap
of the guest RAM to track modified regions.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -44,6 +44,7 @@ requires 'base';
 requires 'constant';
 requires 'strict';
 requires 'warnings';
+requires 'Socket::MsgHdr';
 
 on 'test' => sub {
   requires 'Devel::Cover';
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -18,9 +18,12 @@ package backend::qemu;
 use strict;
 use base 'backend::virt';
 use File::Path 'mkpath';
+use File::Basename;
 use Time::HiRes qw(sleep gettimeofday);
 use IO::Select;
 use IO::Socket::UNIX 'SOCK_STREAM';
+use Socket;
+use Socket::MsgHdr;
 use IO::Handle;
 use POSIX qw(strftime :sys_wait_h);
 use JSON;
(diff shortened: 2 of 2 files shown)


cpanfile
```diff
@@ -44,6 +44,7 @@ requires 'base';
 requires 'constant';
 requires 'strict';
 requires 'warnings';
+requires 'Socket::MsgHdr';
```
coolo: at least we have a package available for it

richiejp: FYI the package generates its class methods inside the INIT block which doesn't work with something we are doing:
```
Too late to run INIT block at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Socket/MsgHdr.pm line 15.
Too late to run INIT block at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Socket/MsgHdr.pm line 59.
```

richiejp:
I will start a new pull request on a different branch

PR #917, opened 2018-01-31
WIP: Make sure that we don't leave running processes

dasantiago:
poo#30388

We need to make sure that the qemu dies.

For example if isotovideo backend dies, then his kid qemu would survive.
This ensures that all the descendants will be dead.

Related changes:
https://github.com/os-autoinst/openQA/commit/5aa41337736490fe2bdc880de2aaf75b1579612c

Verification run: http://tragicbox.suse.cz/tests/473/file/autoinst-log.txt

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -173,9 +173,13 @@ sub run_all {
         warn $@;
         $died = 1;    # test execution died
     }
-    bmwqemu::save_vars();
-    myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
-    close $isotovideo;
+    eval {
+        # When the disk runs out of space there's nothing else we can do. We just can't allow to jump the exit
+        bmwqemu::save_vars();
+        myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
+        close $isotovideo;
+    };
+    warn "Exception while processing the send_json call: $@" if $@;
     Devel::Cover::report() if Devel::Cover->can('report');
     _exit(0);
 }
--- a/isotovideo
+++ b/isotovideo
@@ -146,6 +146,8 @@ sub kill_autotest {
 
 sub kill_backend {
     if (defined $bmwqemu::backend && $bmwqemu::backend->{backend_pid}) {
+        # Make sure that the backend child processes (eg. qemu) dies
+        _kill_all_descendants($bmwqemu::backend->{backend_pid});
         # save the pid in a scalar - signal handlers will reset it
         my $bpid = $bmwqemu::backend->{backend_pid};
         diag "killing backend process $bpid";
@@ -181,6 +183,7 @@ sub signalhandler_chld {
         }
         if ($bmwqemu::backend->{backend_pid} && $child == $bmwqemu::backend->{backend_pid}) {
             diag("backend $child died");
+            _kill_all_descendants($bmwqemu::backend->{backend_pid});
             $bmwqemu::backend->{backend_pid} = 0;
             $loop = 0;
             next;
@@ -323,6 +326,30 @@ our $tags = undef;
 my $backend_requester = undef;
 
 my ($last_check_seconds, $last_check_microseconds) = gettimeofday;
+
+# Kill all the descendantes of a PID - A grandchild process is not considered a child
+# We loop through the proc instead of checking the task/tid/children, so we can invoke this when processing SIGCHLD.
+# Since the orphans are adopted by init (pid 1), we need to check if the process group is the same as the actual group (since forks have the same group)
+# To know more about procfs, please check man(5) proc
+sub _kill_all_descendants {
+    no autodie;
+    my ($ppid) = @_;
+    my $group = getpgrp(0);
+    for my $proc_folder (glob("/proc/*")) {
+        next if (!-d $proc_folder || !-f "$proc_folder/stat" || $proc_folder eq "/proc/$$");
+        open my $ifh, '<', "$proc_folder/stat" or next;
+        my ($pid, $executable, $state, $parent, $pgroup) = split(' ', <$ifh>);
+        close $ifh;
+        if ($executable !~ /videoenc/ && (($parent == $ppid) || ($parent == 1 && $pgroup == $group)) && $state =~ /R|S/) {
+            if (kill 'TERM', $pid) {
+                diag "awaiting death of descendant $executable $pid";
+                my $ret = waitpid($pid, 0);
+                diag "descendant process $executable exited: $ret";
+            }
+        }
+    }
+}
+
 sub _calc_check_delta {
     # an estimate of eternity
     my $delta = 100;


autotest.pm
```diff
@@ -173,9 +173,13 @@ sub run_all {
...
     }
-    bmwqemu::save_vars();
-    myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
-    close $isotovideo;
+    eval {
+        # When the disk runs out of space there's nothing else we can do. We just can't allow to jump the exit
+        bmwqemu::save_vars();
+        myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
+        close $isotovideo;
+    };
```
dasantiago: In the logs, there were some exceptions in this step. I decided to put it in an eval to avoid, the propagation of the exception skipping the _exit (this is a fork process we don't want to leave kids loose).

coolo:
wouldn't it be safer to to this all in the worker? It's the one setting the process group after all. I might be off here, but I guess if you run isotovideo from a shell you don't get an own process group and your code would take down all of my desktop :)

isotovideo
```diff
@@ -323,6 +326,29 @@ our $tags = undef;
...
+# Kill all the descendantes of a PID - A grandchild process is not considered a child
+# We loop through the proc instead of checking the task/tid/children, so we can invoke this when processing SIGCHLD.
+# Since the orphans are adopted by init (pid 1), we need to check if the process group is the same as the actual group (since forks have the same group)
+# To know more about procfs, please check man(5) proc
+sub _kill_all_descendants {
+    my ($ppid) = @_;
+    my $group = getpgrp(0);
+    for my $proc_folder (glob("/proc/*")) {
+        next if (!-d $proc_folder || !-f "$proc_folder/stat" || $proc_folder eq "/proc/$$");
+        open my $ifh, '<', "$proc_folder/stat";
```
mudler: What would happen here if the filehandle fails to be opened?

dasantiago:
@coolo 

When you launch a process from from a shell, it will be put in a new process-group by using setsid. Any child process of the isotovideo (which is the qemu) will get the same group as the isotovideo.

If you check, the isotovideo already as a different process group from the worker:

```
tragicbox:~ # ps xua | grep -i perl | grep worker

_openqa+  3988  0.4  0.3 134908 50164 ?        Ss   09:48   0:00 /usr/bin/perl /usr/share/openqa/script/worker --instance 1

tragicbox:~ # cat /proc/3988/stat | awk '{print $5}'

3988

tragicbox:~ # ps xua | grep -i isotovi

_openqa+  3996  1.3  0.9 241904 154488 ?       S    09:48   0:02 perl /usr/bin/isotovideo -d

_openqa+  3998  0.0  0.3 121280 51612 ?        S    09:48   0:00 /usr/bin/isotovideo: comman

_openqa+  4000  0.5  0.6 252900 101492 ?       S    09:48   0:01 /usr/bin/isotovideo: autote

_openqa+  4002 12.8  1.3 984340 227780 ?       Sl   09:48   0:23 /usr/bin/isotovideo: backen

root      4062  0.0  0.0   9272  1640 pts/0    R+   09:51   0:00 grep --color=auto -i isotovi

tragicbox:~ # cat /proc/4002/stat | awk '{print $5}'

3996

tragicbox:~ # ps xua | grep -i qemu

_openqa+  4006 27.9  6.8 1780220 1116432 ?     Sl   09:48   2:57 /usr/bin/qemu-system-x86_64 -machine 
accel=kvm -serial file:serial0 -soundhw ac97 -vga cirrus -global isa-fdc.driveA= -m 1024 -cpu qemu64 -netdev user,id=qanet0 -device virtio-net,netdev=qanet0,mac=52:54:00:12:34:56 -device virtio-scsi-pci,id=scsi0 -device virtio-blk,drive=hd1,serial=1 -drive file=raid/l1,cache=unsafe,if=none,id=hd1,format=qcow2 -drive media=cdrom,if=none,id=cd0,format=raw,file=/var/lib/openqa/share/factory/iso/SLE-12-SP2-Desktop-DVD-x86_64-GM-DVD1.iso -device scsi-cd,drive=cd0,bus=scsi0.0 -boot once=d,menu=on,splash-time=5000 -device usb-ehci -device usb-tablet -smp 1 -enable-kvm -no-shutdown -vnc :91,share=force-shared -device virtio-serial -chardev socket,path=virtio_console,server,nowait,id=virtio_console,logfile=virtio_console.log -device virtconsole,chardev=virtio_console,name=org.openqa.console.virtio_console -qmp unix:qmp_socket,server,nowait -monitor unix:hmp_socket,server,nowait -S -monitor telnet:127.0.0.1:20012,server,nowait

root      4084  0.0  0.0   9272  1604 pts/0    R+   09:59   0:00 grep --color=auto -i qemu

tragicbox:~ # cat /proc/4006/stat | awk '{print $5}'

3996

tragicbox:~ #

```
Worker has group 3988 while isotovideo and qemu have group 3996.

It won't matter how you call the isotovideo (unless you change manually the process group), all his kids will have the same process group.

Putting this code in worker it would work the same, you would need to check the isotovideo process group as the qemu parent process would already be the init. Unless there's a way that i don't know of identifying process descendants.

Also this is the second part of the fix. The first part, mentioned above, will already check if the qemu is running before deleting the pool folder (which is the final part before of a job). And in case it is, it will shutdown.

isotovideo
dasantiago: @mudler 

you are interested only in the processes you own, and in those you always have permissions to read, but the the stat file is always available to be read (the /proc/[pid]/status one is protected).

But you have a point. Let's say that a process dies, between the `next`and the `open` :-)

Let me fix it :-)

mudler: Whatever you are reading, Perl can fail opening a filehandle

dasantiago: @mudler done

dasantiago:
Example:

```
           └─worker(6288)───perl(6311)─┬─/usr/bin/isotov(6313)
                                       ├─/usr/bin/isotov(6315)
                                       └─/usr/bin/isotov(6317)─┬─qemu-system-x86(6321)─┬─{qemu-system-x86}(6324)
                                                               │                       ├─{qemu-system-x86}(6325)
                                                               │                       ├─{qemu-system-x86}(6326)
                                                               │                       └─{qemu-system-x86}(6328)
                                                               ├─videoencoder(6318)
                                                               ├─{/usr/bin/isotov}(6331)
                                                               ├─{/usr/bin/isotov}(6332)
                                                               ├─{/usr/bin/isotov}(6333)
                                                               ├─{/usr/bin/isotov}(6334)
                                                               ├─{/usr/bin/isotov}(6335)
                                                               ├─{/usr/bin/isotov}(6336)
                                                               ├─{/usr/bin/isotov}(6337)
                                                               └─{/usr/bin/isotov}(6338)

```

This is the process tree of a normal worker.

These are the pgroups:
pgroups:
6288 - 6288
6311 - 6311
6313 - 6311
6315 - 6311
6317 - 6311
6321 - 6311
6318 - 6311
...



These changes will work for when 6313,6315 and 6317 die.

**The problem is when 6311 dies**. When the 6311 dies with a SIGSEGV, all the three processes (6313,6315,6317) will keep running. However when this happens the 6288 will get a SIGCHLD.

After a while i thought that maybe i could add:

> last if getppid() == 1;

in https://github.com/os-autoinst/os-autoinst/blob/master/isotovideo#L385

But this isn't a good solution. So after rambling a bit in the irc channel, it was decided to close this, and implement this in the openqa project code.

PR #918, opened 2018-02-02
qemu: Increase memory dump performance 

richiejp:
This follows on from #915, I have removed Socket::MsgHdr as a dependency and done some other cleanup.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -18,6 +18,7 @@ package backend::qemu;
 use strict;
 use base 'backend::virt';
 use File::Path 'mkpath';
+use File::Which;
 use Time::HiRes qw(sleep gettimeofday);
 use IO::Select;
 use IO::Socket::UNIX 'SOCK_STREAM';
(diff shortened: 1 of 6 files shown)


backend/qemu.pm
```diff
@@ -201,22 +201,54 @@ sub can_handle {
...
-    bmwqemu::diag("Migrating the machine.");
+    my $rsp = $self->handle_qmp_command({execute => 'query-status'}, fatal => 1);
+    bmwqemu::diag("Migrating the machine (Current VM state is $rsp->{return}->{status}).");
+    my $was_running = $rsp->{return}->{status} eq 'running';
+
+    # Ensure slow dump times are not due to a transfer rate cap
+    $rsp = $self->handle_qmp_command(
+        {
+            execute   => 'migrate-set-parameters',
+            arguments => {'max-bandwidth' => $vars->{QEMU_MAX_BANDWIDTH} || 2147483648}
```
okurz: where does this magic number come from? Better have a constant with an explanation?

backend/qemu.pm
```diff
@@ -855,25 +894,41 @@ sub start_qemu {
     $self->{select}->add($self->{qemupipe});
 }
 
-# runs in the thread to bounce QMP
+# Send a QMP command and wait for the result
+#
+# Pass fatal => 1 to die on failure.
+# Pass send_fd => $fd to send $fd to QEMU using SCM rights. Probably only
+# useful with the getfd command
```
okurz: I guess you could simply write POD here for better extractability :)

PR #919, opened 2018-02-02
Avoid letting the kids get loose.

dasantiago:
When there's no disk space available, it's unable to write to disk
and it will throw an exception.

These changes avoid the exception to propagate and so it won't let the child
to skip death.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -173,8 +173,10 @@ sub run_all {
         warn $@;
         $died = 1;    # test execution died
     }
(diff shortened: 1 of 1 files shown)


coolo:
the exception is in save_vars, then? I think the other 2 lines should still be done - even if the save_vars threw

dasantiago:
@coolo 

There's also the possibility of the syswrite fail due to a broken pipe.

Moved the close handle to outside the loop.

PR #920, opened 2018-02-04
testapi: Ensure script_output only returns the actual ouput of the executed script

okurz:
This prevents additional data on the used serial device to confuse
script_output callers which do not expect script_output to implicitly consume
and return that additional "noise" data.

The test data mocking in the unit is certainly not perfect because it relies
on the actual definition of context markers used within the implementation of
script_output but I have not found a better way.

Related progress issue: https://progress.opensuse.org/issues/30613

### DIFF
--- a/t/data/tests/tests/boot.pm
+++ b/t/data/tests/tests/boot.pm
@@ -50,6 +50,12 @@ END
     script_run "echo '924095f2cb4d622a8796de66a5e0a44a  text' > text.md5";
     assert_script_run 'md5sum -c text.md5';
 
+    # TinyCore busybox sh acts as bash but does not provide it so we do here
+    script_run 'alias bash=sh', 0;
+    my $out = script_output('mount');
+    die "mount does not show any mount points? output: $out" unless $out =~ /.*\/.*on/;
+    die "^rootfs not found. output: $out"                    unless $out =~ qr{^rootfs};
+    die "tmpfs on /dev/shm not found. output: $out"          unless $out =~ qr{tmpfs on /dev/shm};
 }
 
 sub test_flags {
--- a/testapi.pm
+++ b/testapi.pm
@@ -905,64 +905,71 @@ sub script_sudo {
 
 Executing script inside SUT with C<bash -eox> (in case of serial console with C<bash -eo>)
 and directs C<stdout> (I<not> C<stderr>!) to the serial console and returns
-the output I<if> the script exists with 0. Otherwise the test is set to failed.
+the output I<if> the script exits with 0. Otherwise the test is set to failed.
 
-Script content based on C<current_test_script> value and typed or fetched
-through HTTP depends on various parameters,so if you need to have it typed because
-your SUT is offline, pass type_command => 1.
+The script content is based on the variable content of C<current_test_script>
+and is typed or fetched through HTTP depending on various parameters. Typing
+can be forced by passing C<type_command => 1> for example when the SUT does
+not provide a usable network connection.
 
-The default timeout for the script is 30 seconds. If you need more, pass a second parameter.
+The default timeout for the script is based on the default in C<wait_serial>
+and can be tweaked by setting the C<$wait> positional parameter.
 
 =cut
 
 sub script_output {
     my ($current_test_script, $wait, %args) = @_;
-    my $suffix = hashed_string("SO$current_test_script");
+    my $marker = hashed_string("SO$current_test_script");
     # 80 is approximate quantity of chars typed during 'curl' approach
     # if script length is lower there is no point to proceed with more complex solution
-    $args{type_command} ||= length($current_test_script) < 80;
+    $args{type_command} //= length($current_test_script) < 80;
+    my $script_path = "/tmp/script$marker.sh";
 
     if (is_serial_terminal) {
-        my $cat = "cat - > /tmp/script$suffix.sh; echo $suffix-\$?-";
+        my $cat = "cat - > $script_path; echo $marker-\$?-";
         type_string($cat . "\n");
         wait_serial("$cat", undef, 0, no_regex => 1);
         type_string($current_test_script);
         type_string("\n", terminate_with => 'EOT');
-        wait_serial("$suffix-0-");
+        wait_serial("$marker-0-");
     }
     elsif ($args{type_command}) {
-        my $cat = "cat - > /tmp/script$suffix.sh << 'EOF';\n";
+        my $cat = "cat - > $script_path;\n";
         type_string($cat);
-        type_string($current_test_script . "\nEOF\n");
+        type_string($current_test_script . "\n");
+        send_key('ctrl-d');
     }
     else {
         open my $fh, ">", 'current_script' or croak("Could not open file. $!");
         print $fh $current_test_script;
         close $fh;
-        assert_script_run "curl -f -v " . autoinst_url("/current_script") . " > /tmp/script$suffix.sh";
+        assert_script_run "curl -f -v " . autoinst_url("/current_script") . " > $script_path";
         script_run "clear";
     }
 
-    my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
+    # Surround the actual script output with special markers so that we can
+    # unambiguously separate the expected output from other content that we
+    # might encounter on the serial device depending on how it is used in the
+    # SUT
+    my $shell_cmd = is_serial_terminal() ? 'bash -oe pipefail' : 'bash -eox pipefail';
+    my $run_script = "echo $marker; $shell_cmd $script_path ; echo SCRIPT_FINISHED$marker-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -oe pipefail $run_script\n");
+        type_string("$run_script\n");
         wait_serial($run_script, undef, 0, no_regex => 1);
     }
     else {
-        type_string "(/bin/bash -eox pipefail $run_script)| tee /dev/$serialdev\n";
+        type_string "($run_script) | tee /dev/$serialdev\n";
     }
-    my $output = wait_serial("SCRIPT_FINISHED$suffix-\\d+-", $wait, 0, record_output => 1)
+    my $output = wait_serial("SCRIPT_FINISHED$marker-\\d+-", $wait, 0, record_output => 1)
       || croak "script timeout";
 
-    croak "script failed" if $output !~ "SCRIPT_FINISHED$suffix-0-";
+    croak "script failed" if $output !~ "SCRIPT_FINISHED$marker-0-";
 
-    # strip the internal exit catcher
-    $output =~ s,SCRIPT_FINISHED$suffix-0-,,;
-
+    # and the markers including internal exit catcher
+    my $out = $output =~ /$marker(?<expected_output>.+)SCRIPT_FINISHED$marker-0-/s ? $+ : '';
     # trim whitespaces
-    $output =~ s/^\s+|\s+$//g;
-
-    return $output;
+    $out =~ s/^\s+|\s+$//g;
+    return $out;
 }
 
 
--- a/testapi.pm
+++ b/testapi.pm
@@ -954,6 +954,7 @@ sub script_output {
     my $shell_cmd = is_serial_terminal() ? 'bash -oe pipefail' : 'bash -eox pipefail';
     my $run_script = "echo $marker; $shell_cmd $script_path ; echo SCRIPT_FINISHED$marker-\$?-";
     if (is_serial_terminal) {
+        wait_serial('# ', undef, 0, no_regex => 1);
         type_string("$run_script\n");
         wait_serial($run_script, undef, 0, no_regex => 1);
     }
--- a/testapi.pm
+++ b/testapi.pm
@@ -621,7 +621,7 @@ Set test variable C<$variable> to value C<$value>.
 
 Specify a true value for the C<reload_needles> flag to trigger a reloading
 of needles in the backend and call the cleanup handler with the new variables
-to make sure that possibly unselected needles are now taken into account
+to make sure that possibly deselected needles are now taken into account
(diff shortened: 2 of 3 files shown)


testapi.pm
```diff
@@ -921,6 +923,11 @@ sub script_output {
     # 80 is approximate quantity of chars typed during 'curl' approach
     # if script length is lower there is no point to proceed with more complex solution
     $args{type_command} ||= length($current_test_script) < 80;
+    # Surround the actual script output with special markers so that we can
+    # unambiguously separate the expected output from other content that we
+    # might encounter on the serial device depending on how it is used in the
+    # SUT
+    $current_test_script = "echo '>>>'$($current_test_script)'<<<'";
```
coolo: this only works for one lined output. try with script_output('mount') in full-stack.

But we already guard the end of the script with SCRIPT_FINISHED. So if you prepend an echo command to the test script and strip that too - it should work.

okurz: Well, I'm trying to run `script_output` in the full stack test but TinyCore does not have bash which `script_output` currently relies on … Let's see what I can come up with

coolo: One more argument to have script_output actually not in os-autoinst but in a linux test library :(

coolo: anyway, then test manually - all I'm saying it's buggy :)

okurz:
ok, did the following changes:

* using `sh` instead of `bash` now in `script_output` to enable testing in full stack test based on tinycore
* added to full stack test next to unit test
* using SCRIPT_FINISHED as already existing end marker and multi-line regex

mudler:
i'm not sure at end will make ```script_output``` bulletproof, but maybe a 'bit' more reliable.

Just for reference my 2c when i was looking into it.

i couldn't think at a safer way of fixing this with the current mechanisms we use, it's going always to be tricky at some point:

Even if it means almost reinventing a new protocol, and a new shiny wheel,  adding another layer of check would maybe help here - we could send special control sequence along with a cksum of the script output generated during invocation,  so we can extract it on the other side, generate the cksum of the received output and compare.

okurz:
@mudler yeah, what you suggest sounds like what @nicksinger also told me. I think we would re-invent [ZMODEM](https://en.wikipedia.org/wiki/ZMODEM) with this approach ;)

coolo:
and remember the more you want the SUT to do, the more you have to type with the current approach - and the higher the chances are you mistype :)

testapi.pm
```diff
@@ -945,24 +952,22 @@ sub script_output {
 
     my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -oe pipefail $run_script\n");
+        type_string("/bin/sh -oe pipefail $run_script\n");
```
coolo: yeah, this still relies on /bin/sh being bash - and as such will fail on debian based systems.

t/data/tests/tests/boot.pm
```diff
@@ -50,6 +50,8 @@ END
     script_run "echo '924095f2cb4d622a8796de66a5e0a44a  text' > text.md5";
     assert_script_run 'md5sum -c text.md5';
 
+    my $out = script_output('mount');
+    die "mount does not show any mount points?" unless $out =~ /.*\/.*on/;
```
coolo: well, sure you get the mount output, but not as lines

```
coolo@f84#openQA>echo $(mount)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime) proc on /proc type proc (rw,nosuid,nodev,noexec,relatime) devtmpfs on /dev type devtmpfs (rw,nosuid,size=16422316k,nr_inodes=4105579,mode=755) securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime) tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev) devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000) tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755) tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755) cgroup on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime)

File: testapi.pm (hunk 2 above)
Comment (okurz) [reply testapi.pm]: IMHO not true because TinyCore provides POSIX sh from busybox and it works there.

File: t/data/tests/tests/boot.pm (hunk 3 above)
Comment (okurz) [reply t/data/tests/tests/boot.pm]: I guess I have that fixed now with /…/s instead of /…/m

File: testapi.pm [hunk 4]
```diff
@@ -905,64 +905,69 @@ sub script_sudo {
...
-    my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
+    my $run_script = "/tmp/script$marker.sh ; echo SCRIPT_FINISHED$marker-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -oe pipefail $run_script\n");
+        type_string("/bin/sh -oe pipefail $run_script\n");
         wait_serial($run_script, undef, 0, no_regex => 1);
     }
     else {
-        type_string "(/bin/bash -eox pipefail $run_script)| tee /dev/$serialdev\n";
+        type_string "(/bin/sh -eox pipefail $run_script)| tee /dev/$serialdev\n";
```
coolo: busybox sh supports the option - dash does not.

test case: put 'false | cat' in /tmp/sh.
```
> bash  t.sh; echo DONE-$?
DONE-0
> bash -eox pipefail t.sh; echo DONE-$?
+ false
+ cat
DONE-1

(looks the same on the busybox ash)

and on ubuntu:
> sh /tmp/t.sh; echo DONE-$?
DONE-0
> sh -eox pipefail /tmp/t.sh; echo DONE-$?
sh: 0: Illegal option -o pipefail
DONE-2

File: testapi.pm [hunk 5]
```diff
@@ -905,64 +905,70 @@ sub script_sudo {
...
         type_string("\n", terminate_with => 'EOT');
-        wait_serial("$suffix-0-");
     }
     elsif ($args{type_command}) {
-        my $cat = "cat - > /tmp/script$suffix.sh << 'EOF';\n";
+        my $cat = "cat - > $script_path;\n";
         type_string($cat);
-        type_string($current_test_script . "\nEOF\n");
+        type_string($current_test_script . "\n");
+        send_key('ctrl-d');
```
okurz: @asmorodskyi are you ok with this change? IMHO 'ctrl-d' is less typing than the double EOF and replicates better what `is_serial_terminal` is doing

testapi.pm
```diff
@@ -905,64 +905,70 @@ sub script_sudo {
...
+    my $script_path = "/tmp/script$marker.sh";
 
     if (is_serial_terminal) {
-        my $cat = "cat - > /tmp/script$suffix.sh; echo $suffix-\$?-";
+        my $cat = "cat - > $script_path";
         type_string($cat . "\n");
         wait_serial("$cat", undef, 0, no_regex => 1);
         type_string($current_test_script);
         type_string("\n", terminate_with => 'EOT');
-        wait_serial("$suffix-0-");
```
okurz: @richiejp I removed that `wait_serial` because further down we already have a start and end marker now. Do you think this is ok?

t/data/tests/tests/boot.pm
```diff
@@ -50,6 +50,10 @@ END
     script_run "echo '924095f2cb4d622a8796de66a5e0a44a  text' > text.md5";
     assert_script_run 'md5sum -c text.md5';
 
+    # TinyCore busybox sh acts as bash but does not provide it so we do here
+    script_run 'alias bash=sh', 0;
```
okurz: @coolo this makes it work for tinycore. Unfortunately I can't add a test for downloading as we have no curl in tinycore

testapi.pm
```diff
@@ -905,64 +905,70 @@ sub script_sudo {
...
         script_run "clear";
     }
 
-    my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
+    # Surround the actual script output with special markers so that we can
+    # unambiguously separate the expected output from other content that we
+    # might encounter on the serial device depending on how it is used in the
+    # SUT
+    my $shell_cmd = is_serial_terminal() ? 'bash -oe pipefail' : 'bash -eox pipefail';
+    my $run_script = "echo $marker; $shell_cmd $script_path ; echo SCRIPT_FINISHED$marker-\$?-";
```
okurz: @rwx788 in your zypper_lifecycle workaround you used `echo '>>>';echo $(<cmd>)`. Instead I am just using the equivalence of `echo >>>; <cmd>`. To my and @nicksinger's understanding that should be enough and the tests confirm this. If you see any other particular reason why we should do your approach feel open to express your opinion :)

testapi.pm
```diff
@@ -905,64 +905,70 @@ sub script_sudo {
...
+        assert_script_run "curl -f -v " . autoinst_url("/current_script") . " > $script_path";
         script_run "clear";
     }
 
-    my $run_script = "/tmp/script$suffix.sh ; echo SCRIPT_FINISHED$suffix-\$?-";
+    # Surround the actual script output with special markers so that we can
+    # unambiguously separate the expected output from other content that we
+    # might encounter on the serial device depending on how it is used in the
+    # SUT
+    my $shell_cmd = is_serial_terminal() ? 'bash -oe pipefail' : 'bash -eox pipefail';
```
okurz: @coolo I am using 'bash' instead of '/bin/bash' here. Less typing and allows to define bash for the tinycore based test. Any reason for /bin/bash?

testapi.pm
richiejp: It is hard to say, but probably not, because the next type_string might send data before the EOT has been transformed into a signal by the TTY layer or cat. In which case that data might be dropped because it was read into a buffer after the EOT and the owner of that buffer does not bother to preserve it (or in cat's case, it can't do anything with it). Also the shell can try to send more data to cat after it has received the signal, but before it has closed STDIN. I don't really know where characters get dropped, but they disappear quite a lot if you are not careful.

As a general rule it is best to introduce a wait after every command with the serial terminal.

testapi.pm
rwx788: I guess that's more readable than having marker in command. My intention was to have single echo, so we print atomic line. If we have 2 echo calls, there can be something in between if there is another activity in the background but I haven't seen that output of the command was broken in the middle.

testapi.pm
okurz: I see, I will change that specific part back then. Thank you

okurz: ok, I put it back.

okurz:
updated to put back the extra synchronisation for `is_serial_terminal`.

okurz:
comment on what? I don't want to describe all implementation details in the documentation. Don't know what to add.

testapi.pm
foursixnine: @okurz NVM, I guess I kind of skip this comment in the code... It's more than enough

okurz:
conducting integration test with openQA and real tests

okurz:
Tested against openQA+os-autoinst-distri-opensuse as well, see http://lord.arch/tests/480

asmorodskyi:
Currently I am having some problems with script_output + VIRTIO_CONSOLE=1 . I decided to test behavior with this branch. This detect regression :
- http://kimball.arch.suse.de/tests/443 ( current master version , basic module fails but after 2 tests passed ) 
- http://kimball.arch.suse.de/tests/456 ( worker running with this patch applied , basic module fails much earlier )

okurz:
I doubt I would be able to find the time to integrate a proper virtio_console test within os-autoinst soon but I guess this is what we should do. If someone needs this PR urgently feel free to investigate further and propose changes and I will incorporate them.

richiejp:
@asmorodskyi are your tests using the version where Oliver re-added the `wait_serial`?

I was worried about text being dropped, but in this case the terminal and/or shell appears to have echoed `/bin/bash -oe pipefail /tmp/scriptFUuEK.sh ; echo SCRIPT_FINISHEDFUuEK-$?-` twice and then possibly copied it into your `$iface` variable.

@okurz There are tests in 10-terminal.t, but they use the `screen` interface directly instead of testapi. However I did leave a stub for adding tests which go through testapi.

asmorodskyi:
@richiejp I checkout latest version available at time when I did run . According to git log @okurz did latest update 1 day before my message so it seems like I used latest version.

testapi.pm
```diff
@@ -905,64 +905,71 @@ sub script_sudo {
...
+    my $run_script = "echo $marker; $shell_cmd $script_path ; echo SCRIPT_FINISHED$marker-\$?-";
     if (is_serial_terminal) {
-        type_string("/bin/bash -oe pipefail $run_script\n");
+        type_string("$run_script\n");
```
richiejp: There should probably by a `wait_serial('# ', undef, 0, no_regex => 1)` here. It looks like on @asmorodskyi's test it is typing in `$run_script` before the shell prompt appears which causes the shell to echo it twice.

okurz:
@asmorodskyi could you please test the current version with the third commit named "wip…" ?

testapi.pm
```diff
@@ -932,6 +932,7 @@ sub script_output {
         type_string($current_test_script);
         type_string("\n", terminate_with => 'EOT');
         wait_serial("$marker-0-");
+        wait_serial('# ', undef, 0, no_regex => 1);
```
richiejp: Please put this just before the `type_string` which requires it. In the future, someone may decide to replace the below `type_string` with `script_run` which contains a `wait_serial('# ' ...`. Also there is no need to synchronise the test script and os-autoinst at this point.

okurz:
uhm, shouldn't have merged with the "wip" commit included. Sorry to not mark the PR as WIP :-/

coolo:
we will see :)

PR #921, opened 2018-02-05
[RFC] Do sanity check for validate code

bmwiedemann:
to avoid cases like zypper_moo returning a constant string
instead of doing a pattern match
This created confusion for people making new tests based on existing ones like in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4346

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1037,6 +1037,15 @@ sub validate_script_output($&;$) {
     my $output = script_output($script, $wait);
     my $res = 'ok';
 
+    if ($ENV{SANITYCHECK}) {
+        # quick sanity check for check functions:
+        $_ = '';
+        my $ret1 = $code->();
+        $_ = 'foobar';
+        if ($ret1 && $code->()) {
+            confess("sanity check failed for $script validator code - always returns true");
+        }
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -1037,6 +1037,15 @@ sub validate_script_output($&;$) {
     my $output = script_output($script, $wait);
     my $res = 'ok';
 
+    if ($ENV{SANITYCHECK}) {
+        # quick sanity check for check functions:
+        $_ = '';
+        my $ret1 = $code->();
+        $_ = 'foobar';
```
coolo: I don't think this is worth it. This should actually be possible to check by means of static analysis - but calling the function just so is a no-go. And if you hide it behind a variable, it's even more no-go as we would still do this only by chance.

bmwiedemann: we can certainly drop the if.
But I recently saw http://perlmonks.org/?node_id=663393 which seems to suggest that static analysis is impossible - at least in the general case.

coolo: I don't care about the general case

coolo:
I say no

PR #923, opened 2018-02-12
Don't add floppy for arm/aarch

cwh42:
Fixing https://progress.opensuse.org/issues/25622

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -563,7 +563,10 @@ sub start_qemu {
     {
         push(@params, @vgaoptions);
 
-        gen_params @params, "global", "isa-fdc.driveA=" unless $vars->{QEMU_NO_FDC_SET};
+        # Avoid qemu's error message about floppy controller when using arm/aarch64:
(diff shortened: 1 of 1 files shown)


backend/qemu.pm
```diff
@@ -563,7 +563,9 @@ sub start_qemu {
     {
         push(@params, @vgaoptions);
 
-        gen_params @params, "global", "isa-fdc.driveA=" unless $vars->{QEMU_NO_FDC_SET};
+        unless ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm' || $vars->{QEMU_NO_FDC_SET}) {
```
SergioAtSUSE: Why `$vars->{ARCH} eq 'arm'`?
I think we don't use this name.

okurz: Please add a comment describing why the parameter does not apply for aarch64 (or arm)

cwh42: Used that to be consistent compared to #L433

SergioAtSUSE: You are right.

This is os-autotinst. We don't use this name in os-autoinst-distri*.

cwh42:
Added an explaining comment to the change.

okurz:
Awesome, LGTM

PR #926, opened 2018-02-20
testapi: Refactor script_output into distribution.pm and fix race

richiejp:
script_output's implementation is very specific to the Operating System and
even to a particular Linux distribution. This moves it into a place where it
can be overridden by the distribution and can also use distribution specific
variables more easily.

This also fixes a race condition with the serial terminal console by waiting
for the shell prompt to appear.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -19,6 +19,7 @@ use strict;
 use warnings;
 
 use testapi ();
+use Carp 'croak';
 
 sub new {
     my ($class) = @_;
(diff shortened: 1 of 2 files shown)


richiejp:
It probably needs to be implemented in distribution.pm where it can be overridden.

FYI @asmorodskyi

richiejp:
@asmorodskyi maybe it fixes your problem now?

distribution.pm
```diff
@@ -171,6 +172,74 @@ sub script_sudo {
...
+        close $fh;
+        testapi::assert_script_run("curl -f -v " . testapi::autoinst_url("/current_script") . " > $script_path");
+        testapi::script_run "clear";
+    }
+
+    # Surround the actual script output with special markers so that we can
+    # unambiguously separate the expected output from other content that we
+    # might encounter on the serial device depending on how it is used in the
+    # SUT
+    my $shell_cmd = testapi::is_serial_terminal ? 'bash -oe pipefail' : 'bash -eox pipefail';
```
asmorodskyi: this is right way :
`my $shell_cmd = testapi::is_serial_terminal() ? 'bash -oe pipefail' : 'bash -eox pipefail';`
currently it is not compiling with your version of code

richiejp: Works on my computer... but I will change it :-p

PR #928, opened 2018-02-27
Add support to override HDDMODEL per disk

frankenmichl:
HDDMODEL can be set to 'nvme' to make qemu use this disk type. However,
sometimes it is desired to just have this extra NVMe disk in addition to
a classical setup.
This commit adds a a new variable HDDMODEL_$i available for all disks.
When it is set, the specified disk will be made availabe as NVMe-disk,
and the snapshot-feature will be disabled, like with the global
HDDMODEL='nvme'.
The HDDMODEL_$i will override the global HDDMODEL, but does not replace
the parameter in order to not break anything.

Signed-off-by: Michael Moese <mmoese@suse.de>

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -202,7 +202,7 @@ sub can_handle {
         return if $vars->{QEMU_DISABLE_SNAPSHOTS};
 
         for my $i (1 .. $vars->{NUMDISKS}) {
-            return if $vars->{"HDDMODEL_$i"} eq 'nvme';
+            return if (defined $vars->{"HDDMODEL_$i"} && $vars->{"HDDMODEL_$i"} eq 'nvme');
         }
 
         return {ret => 1};
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -718,7 +718,8 @@ sub start_qemu {
                 my $diskmodel;
                 if (defined $vars->{"HDDMODEL_$i"}) {
                     $diskmodel = $vars->{"HDDMODEL_$i"};
-                } else {
+                }
+                else {
                     $diskmodel = $vars->{HDDMODEL};
                 }
                 # when booting from disk on UEFI, first connected disk gets ",bootindex=0"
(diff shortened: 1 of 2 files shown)


frankenmichl:
[autoinst-log.txt](https://github.com/os-autoinst/os-autoinst/files/1762091/autoinst-log.txt)

backend/qemu.pm
```diff
@@ -711,10 +715,16 @@ sub start_qemu {
                 }
             }
             else {
+                my $diskmodel;
+                if (defined $vars->{"HDDMODEL_$i"}) {
```
foursixnine: Since it's a new variable, it must be documented in doc/backend_vars.asciidoc file

backend/qemu.pm
```diff
@@ -201,6 +201,10 @@ sub can_handle {
 
         return if $vars->{QEMU_DISABLE_SNAPSHOTS};
 
+        for my $i (1 .. $vars->{NUMDISKS}) {
+            return if $vars->{"HDDMODEL_$i"} eq 'nvme';
```
mitiao: From your log, I could see uninitialized value warning here, guess if not define all HDDMODEL_$i in settings, better to add `defined` statement to avoid this warning.

frankenmichl:
[autoinst-log.txt](https://github.com/os-autoinst/os-autoinst/files/1775152/autoinst-log.txt)

backend/qemu.pm
```diff
@@ -711,10 +715,17 @@ sub start_qemu {
                 }
             }
             else {
+                my $diskmodel;
+                if (defined $vars->{"HDDMODEL_$i"}) {
+                    $diskmodel = $vars->{"HDDMODEL_$i"};
+                }
+                else {
```
coolo: I would write this as 

```
my $diskmodel = $vars->{"HDDMODEL_$i"} || $vars->{HDDMODEL};
```

There are various other examples of such a usage in qemu.pm. But I won't block merging for this

PR #929, opened 2018-03-03
qemu: enable discard support for disks

marmarek:
Add 'discard=on' option to disks. This allows using `fstrim` in SUT to
make its disk image smaller - especially helpful when using STORE_HDD_*
or PUBLISH_HDD_*. In extreme cases it helps avoiding 20GB upload size
limit.

To make it really working, one need also HDDMODEL=scsi-hd, qemu
apparently ignore it for virtio-blk devices.


coolo:
Now that you highlighted it, I wonder if cache=none or cache=unsafe are the better option. I don't think it's justified that we use different modes there

coolo:
I would go with cache=unsafe - which means it's using cache but doesn't care for writeback

marmarek:
But there is already `cache=unsafe` (for non-multipath case).

coolo:
yeah, as you added the discard option to both lines I stumbled over the different cache= cases. I can't justify it (trying to make sense out of the man page of qemu if discard is problematic or not).

marmarek:
I guess `cache=none` is needed for multipath for data consistency across paths.

coolo:
now that makes sense. 

I resolved the conflict on command line

PR #931, opened 2018-03-05
Tries to guess which qemu executable should use via WORKER_CLASS

dasantiago:
related poo#14674

There are two ways to define which qemu to use:
- via environment variable QEMU
- via setting the variable QEMU (it will be appended to qemu-system-)

However some templates don't have this variable defined, inducing
the worker to choose the wrong qemu.
So if any of the previous mechanisms work, it will use the
WORKER_CLASS to guess which executable should use.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -418,8 +418,23 @@ sub start_qemu {
     my $qemubin = $ENV{QEMU};
 
     my $qemuimg = find_bin('/usr/bin/', qw(kvm-img qemu-img));
+
     unless ($qemubin) {
-        $qemubin = find_bin('/usr/bin/', $vars->{QEMU} ? ('qemu-system-' . $vars->{QEMU}) : qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64));
+        if ($vars->{QEMU}) {
+            $qemubin = find_bin('/usr/bin/', 'qemu-system-' . $vars->{QEMU});
+        }
+        else {
+            (my $class = $vars->{WORKER_CLASS} || '') =~ s/qemu_/qemu-system\-/g;
+            my @execs = qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64);
+            my %allowed = map { $_ => 1 } @execs;
+            for (split(/\s*,\s*/, $class)) {
+                if ($allowed{$_}) {
+                    $qemubin = find_bin('/usr/bin/', $_);
+                    last;
+                }
+            }
+            $qemubin = find_bin('/usr/bin/', @execs) unless $qemubin;
+        }
     }
 
     die "no kvm-img/qemu-img found\n" unless $qemuimg;


backend/qemu.pm
```diff
@@ -418,8 +418,15 @@ sub start_qemu {
...
 
     my $qemuimg = find_bin('/usr/bin/', qw(kvm-img qemu-img));
+
     unless ($qemubin) {
-        $qemubin = find_bin('/usr/bin/', $vars->{QEMU} ? ('qemu-system-' . $vars->{QEMU}) : qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64));
+        if ($vars->{QEMU}) {
+            $qemubin = find_bin('/usr/bin/', 'qemu-system-' . $vars->{QEMU});
+        }
+        else {
+            (my $class = $vars->{WORKER_CLASS}) =~ s/qemu_/qemu-system\-/g;
```
coolo: won't this cause warnings on undefined?

dasantiago: @coolo 

Yes it will. :-)

Hence the 
> (split(/\s*,\s*/, $class // '')

in the next line

dasantiago: @coolo true.

Fixed.

backend/qemu.pm
```diff
@@ -418,8 +418,15 @@ sub start_qemu {
...
     my $qemuimg = find_bin('/usr/bin/', qw(kvm-img qemu-img));
+
     unless ($qemubin) {
-        $qemubin = find_bin('/usr/bin/', $vars->{QEMU} ? ('qemu-system-' . $vars->{QEMU}) : qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64));
+        if ($vars->{QEMU}) {
+            $qemubin = find_bin('/usr/bin/', 'qemu-system-' . $vars->{QEMU});
+        }
+        else {
+            (my $class = $vars->{WORKER_CLASS} || '') =~ s/qemu_/qemu-system\-/g;
+            $qemubin = find_bin('/usr/bin/', (split(/\s*,\s*/, $class), qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64)));
```
coolo: I have one more question: if WORKER_CLASS=tap,qemu_x86_64 will you start /usr/bin/tap?

dasantiago: @coolo You do good questions :+1:  :-D

got the point.

Fixed

backend/qemu.pm
```diff
@@ -418,8 +418,19 @@ sub start_qemu {
...
+        if ($vars->{QEMU}) {
+            $qemubin = find_bin('/usr/bin/', 'qemu-system-' . $vars->{QEMU});
+        }
+        else {
+            (my $class = $vars->{WORKER_CLASS} || '') =~ s/qemu_/qemu-system\-/g;
+            my %allowed = map { $_ => 1 } qw(kvm qemu-kvm qemu qemu-system-x86_64 qemu-system-ppc64);
+            for (split(/\s*,\s*/, $class)) {
+                $qemubin = find_bin('/usr/bin/', $_) if $allowed{$_};
+            }
+            $qemubin = find_bin('/usr/bin/', keys %allowed) unless $qemubin;
```
coolo: this drops order - and we would run qemu-system-ppc64 before qemu by chance (perl hash randomization)

dasantiago: @coolo done.

coolo:
you thought it's an easy one, didn't you? :)

PR #932, opened 2018-03-12
Search for patterns in the SERIAL file and raise failures

dasantiago:
Search for patterns in the SERIAL file and raise failures

With these changes the tester can now define if he wants to raise a soft or a hard
failure when a pattern appears in a test.

There's two ways to use this functionality:
1- On main.pm the developer after setting the distribution, can set the serial failures interface
   Example:

> $testapi::distri->set_serial_failures(soft=>['gcc version 4.8.5 (SUSE Linux)'], hard=>[qr/No iBFT detected/]);

2- On the tests the developer needs to set the serial_failures variable of the test
   object

The patterns defined in main.pm file are going to be available to all tests, the
developer can then override those patterns by setting the instance variable.

The hashref (case 2) needs to have this structure:
> \%SERIAL_FAILURES = (
>                  soft => [quotemeta "pattern 1", qr/pattern2(?!pattern3)/],
>                  hard => [qr/pattern 3/]
> );

The SERIAL_FAILURES keys needs to be soft or hard.
The "soft" key will raise a softfail warning in the test.
The "hard" key will raise a hard fail in the test.

Both keys need to have an arrayref containing strings.
The strings represent the patterns to look in the serial file.

The comparation will be done line by line (please pay attention to the $/ var)
and comparing the pattern string/regexp as-is with the line.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -104,6 +104,7 @@ sub loadtest {
     $test             = $name->new($category);
     $test->{script}   = $script;
     $test->{fullname} = $fullname;
+    $test->{serial_failures} = $testapi::distri->{serial_failures} // {};
 
     if (defined $args{run_args}) {
         unless (blessed($args{run_args}) && $args{run_args}->isa('OpenQA::Test::RunArgs')) {
@@ -270,6 +271,7 @@ sub runalltests {
     my $vmloaded            = 0;
     my $snapshots_supported = query_isotovideo('backend_can_handle', {function => 'snapshots'});
     bmwqemu::diag "Snapshots are " . ($snapshots_supported ? '' : 'not ') . "supported";
+    my $serial_file_pos = 0;
 
     write_test_order();
 
@@ -298,7 +300,10 @@ sub runalltests {
                 make_snapshot($t->{fullname});
             }
 
-            eval { $t->runtest; };
+            eval {
+                $t->runtest;
+                $serial_file_pos = parse_serial_output($serial_file_pos, $t);
+            };
             $t->save_test_result();
 
             if ($@) {
@@ -343,6 +348,51 @@ sub loadtestdir {
     foreach my $script (glob "$dir/*.pm") {
         loadtest($script);
     }
+}
+
+sub parse_serial_output {
+    if ($bmwqemu::vars{BACKEND} eq 'qemu') {
+        parse_serial_output_qemu(@_);
+    }
+}
+
+sub parse_serial_output_qemu {
+    my ($serial_file_pos, $t) = @_;
+    # serial failures defined in distri (test can override them)
+    my $failures = $t->{serial_failures};
+
+    myjsonrpc::send_json($isotovideo, {cmd => 'read_serial', position => $serial_file_pos});
+    my $json = myjsonrpc::read_json($isotovideo);
+
+    # loop line by line
+    open my $ifh, '<', \$json->{serial} or die "Unable to process serial text: $!";
+    my $die = 0;
+    while (defined(my $line = <$ifh>)) {
+        chomp $line;
+        # only two keys allowed
+        for my $type (qw(soft hard)) {
+            for my $regexp (@{$failures->{$type}}) {
+                # If you want to match a simple string please be sure that you create it with quotemeta
+                if ($line =~ /$regexp/) {
+                    if ($type eq 'soft') {
+                        $t->record_testresult('softfail');
+                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'softfail');
+                        $t->{result} = 'softfail';
+                    }
+                    else {
+                        $t->record_testresult('fail');
+                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'fail');
+                        $t->{result} = 'fail';
+                        $die = 1;
+                    }
+                }
+            }
+        }
+    }
+
+    close $ifh;
+    die if $die;
+    return $json->{position};
 }
 
 1;
--- a/autotest.pm
+++ b/autotest.pm
@@ -302,7 +302,7 @@ sub runalltests {
 
             eval {
                 $t->runtest;
-                $serial_file_pos = parse_serial_output($serial_file_pos, $t);
+                $serial_file_pos = $t->search_for_expected_serial_failures($serial_file_pos);
             };
             $t->save_test_result();
 
@@ -348,49 +348,6 @@ sub loadtestdir {
     foreach my $script (glob "$dir/*.pm") {
         loadtest($script);
     }
-}
-
-sub parse_serial_output {
-    if ($bmwqemu::vars{BACKEND} eq 'qemu') {
-        parse_serial_output_qemu(@_);
-    }
-}
-
-sub parse_serial_output_qemu {
-    my ($serial_file_pos, $t) = @_;
-    # serial failures defined in distri (test can override them)
-    my $failures = $t->{serial_failures};
-
-    myjsonrpc::send_json($isotovideo, {cmd => 'read_serial', position => $serial_file_pos});
-    my $json = myjsonrpc::read_json($isotovideo);
-
-    my $die = 0;
-    # loop line by line
-    for my $line (split(/^/, $json->{serial})) {
-        chomp $line;
-        # only two keys allowed
-        for my $type (qw(soft hard)) {
-            for my $regexp (@{$failures->{$type}}) {
-                # If you want to match a simple string please be sure that you create it with quotemeta
-                if ($line =~ /$regexp/) {
-                    if ($type eq 'soft') {
-                        $t->record_testresult('softfail');
-                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'softfail');
-                        $t->{result} = 'softfail';
-                    }
-                    else {
-                        $t->record_testresult('fail');
-                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'fail');
-                        $t->{result} = 'fail';
-                        $die = 1;
-                    }
-                }
-            }
-        }
-    }
-
-    die "Got serial hard failure" if $die;
-    return $json->{position};
 }
 
 1;
--- a/autotest.pm
+++ b/autotest.pm
@@ -364,10 +364,9 @@ sub parse_serial_output_qemu {
     myjsonrpc::send_json($isotovideo, {cmd => 'read_serial', position => $serial_file_pos});
     my $json = myjsonrpc::read_json($isotovideo);
 
-    # loop line by line
-    open my $ifh, '<', \$json->{serial} or die "Unable to process serial text: $!";
     my $die = 0;
-    while (defined(my $line = <$ifh>)) {
+    # loop line by line
+    for my $line (split(/^/, $json->{serial})) {
         chomp $line;
         # only two keys allowed
         for my $type (qw(soft hard)) {
@@ -390,7 +389,6 @@ sub parse_serial_output_qemu {
         }
     }
 
-    close $ifh;
     die if $die;
     return $json->{position};
 }
--- a/autotest.pm
+++ b/autotest.pm
@@ -389,7 +389,7 @@ sub parse_serial_output_qemu {
         }
     }
 
-    die if $die;
+    die "Got serial hard failure" if $die;
     return $json->{position};
 }
 
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -699,13 +699,26 @@ Returns the output on the serial device since the last call to set_serial_offset
 
 sub serial_text {
     my ($self) = @_;
+    return ($self->read_serial($self->{serial_offset}))[0];
+}
+
+=head2 read_serial
+
+Returns the output and the offset after reading on the serial device from position
+
+=cut
+
+sub read_serial {
+    my ($self, $position, $whence) = @_;
 
     open(my $SERIAL, "<", $self->{serialfile});
-    seek($SERIAL, $self->{serial_offset}, 0);
+    seek($SERIAL, $position, $whence // 0);
     local $/;
-    my $data = <$SERIAL>;
+    my $data   = <$SERIAL>;
+    my $offset = tell $SERIAL;
     close($SERIAL);
-    return $data;
+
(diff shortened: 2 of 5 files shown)


autotest.pm
```diff
@@ -104,6 +104,7 @@ sub loadtest {
     $test             = $name->new($category);
     $test->{script}   = $script;
     $test->{fullname} = $fullname;
+    $test->{serial_failures} = \%::SERIAL_FAILURES // {};
```
coolo: that's a bit strange API - I would prefer if this was set explicitly through the distribution object.

autotest.pm
```diff
@@ -345,6 +350,44 @@ sub loadtestdir {
     }
 }
 
+sub parse_serial_output {
+    return -1 if !-e 'serial0';
```
richiejp: I think there is variable for this file name

autotest.pm
```diff
@@ -345,6 +350,44 @@ sub loadtestdir {
...
+    
+    open my $ifh, '<', 'serial0' or die "Unable to open file: $!";
+    seek $ifh, $serial_file_pos, 0;
+    my $die = 0;
+    while (defined(my $line = <$ifh>)) {
+        chomp $line;
+        for my $type (keys %$failures) {
+            for my $regexp (@{$failures->{$type}}) {
+                my $re = quotemeta $regexp;
+                if ($line =~ /$regexp/ || $line =~ /$re/) {
```
richiejp: The user should store the regex in the hash with `qr/../` so they are compiled once

dasantiago: @richiejp 

Compiling regexp doesn't mean it's faster and **not** using qr doesn't mean multiple regexp compilations . Using qr **might be faster** if we have variables interpolated. If there's no variables to be interpolated (which i assume is going to be most of the cases - since we will be looking for specific strings) the regular expression is compiled once and cached.

autotest.pm
dasantiago: @coolo 
Not sure if this is what you mean:

> $test->{serial_failures} = $testapi::distri->{serial_failures} // {};

autotest.pm
dasantiago: There's a serialfile instance variable in the backend but since is an instance variable is in a different scope.

Maybe i should make it class variable...

autotest.pm
coolo: I mean instead of that global variable a specific call to $distri - how you handle it internally is not my problem.

autotest.pm
foursixnine: I would suggest to expect the user to write the expressions with qr/exp/ already, because of compiling only once, and because you should not try to cover for an user, giving another point of possible failure by adding quotemeta. Unless you intention is giving the possibility of having support for both cases... Which imo, not the best idea

autotest.pm
```diff
@@ -345,6 +350,44 @@ sub loadtestdir {
...
+                    if ($type eq 'soft') {
+                        $t->record_testresult('softfail');
+                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'softfail');
+                        $t->{result} = 'softfail';
+
+                    }
+                    else {
```
foursixnine: if you are only allowing `soft` or `hard`, this block could be simplified, since you can already map:
hard = fail && $die = 1;
soft = softfail && $die = 0;

autotest.pm
```diff
@@ -345,6 +350,44 @@ sub loadtestdir {
...
+
+    my ($serial_file_pos, $t) = @_;
+    my $failures = $t->{serial_failures};
+
+    open my $ifh, '<', 'serial0' or die "Unable to open file: $!";
+    seek $ifh, $serial_file_pos, 0;
+    my $die = 0;
+    while (defined(my $line = <$ifh>)) {
+        chomp $line;
+        for my $type (keys %$failures) {
```
foursixnine: See my comment below regarding soft & hard

autotest.pm
dasantiago: @coolo Done.

Created a function in the DIstribution.pm:

> $testapi::distri->set_serial_failures(soft=>['gcc version 4.8.5 (SUSE Linux)'], hard=>[qr/No iBFT detected/]);

autotest.pm
dasantiago: done

autotest.pm
richiejp: Another idea would be to move this to the backend somehow. One way would be to put it in the backend base class then it can be overridden by backends which use a socket instead of a file or multiplex multiple serial streams through one file etc.

distribution.pm
```diff
@@ -240,6 +241,14 @@ sub script_output {
     return $out;
 }
 
+sub set_serial_failures {
```
foursixnine: POD? ^^

basetest.pm
```diff
@@ -458,7 +459,7 @@ sub record_testresult {
         $res = 'unk';
     }
 
-    if ($res eq 'fail') {
+    if ($res eq 'fail' || $res eq 'softfail') {
```
okurz: I would be careful with this change. Looks unrelated

distribution.pm
okurz: yes, doc would be appreciated. Also I wonder if there can not be a better name. How about `set_expected_serial_failures`?

autotest.pm
dasantiago: @richiejp 

It's a bit tricky to move it to the backend. Please check now the current solution. :-)

dasantiago:
All the changes mentioned were addressed.

Please confirm everything's fine now :-)

distribution.pm
```diff
@@ -240,6 +241,25 @@ sub script_output {
...
 
+=head2 set_expected_serial_failures
+
+    set_expected_serial_failures(%failures)
+
+Define the patterns to look for in the serial console.
+The patterns can be either I<hard> or I<soft>.
+
+Example:
+
```
okurz: Uhm, shouldn't there be an example?

autotest.pm
richiejp: Sorry to be a pain, but this still does not look right to me. :-)

Maybe it would be better to send a list of regular expressions to the backend and the backend sends back any matches. Sending the regular expressions over JSON is annoying (it makes my previous comment about regexs irrelevant), but that is basically what we are doing with every wait_serial at least while using virtio_console.

Or you could just send all the serial output over JSON since the last request. The JSON serialiser will probably cope.

Sending the file name and file position to the front end only works for regular files which support seek. If it is a socket or something else then it won't work and we will need a workaround. This basically tightly couples the backend with autotest.

dasantiago: @richiejp 

I'm not understanding your point... I'm sending the ouptut via JSON: https://github.com/os-autoinst/os-autoinst/pull/932/files#diff-07616258592e792ab88f484012b38370R483


$serial is the serial data, returned by the function read_serial:
https://github.com/os-autoinst/os-autoinst/pull/932/files#diff-e9f8b7109dd3e4e2e6ce710cd417d82fR711

distribution.pm
dasantiago: XD

autotest.pm
richiejp: Ah, OK, my bad. The other thing which is confusing me is that you are sending the file position. It would be nice if the backend stores the file position and just sends all the text since the previous call to `read_serial(..., stream_id => 'kernel_log')` where `stream_id` allows us to specify which serial output we need without any implementation specific information. 

However implementing that correctly might be a much bigger job and it won't take long to refactor this feature once it is implemented, so this looks OK for now, but only because it is a very nice feature. :-)

dasantiago: There's already a mechanism in place to store the current position of the serial, however since tests can read from serial as well, we would need to implement a mechanism to identify reads.

I also think that a function to tail the serial file is useful :-)

dasantiago:
Tests and changes done in e212:

1- worker12 with /var/lib/openqa/share unmounted

2- Files changed on e122 (changes not visible in OSD e212):
/var/lib/openqa/share/tests/sle/products/sle/main.pm
/var/lib/openqa/share/tests/sle/tests/console/hostname.pm
/var/lib/openqa/share/tests/sle/tests/slenkins/login.pm

these files are transfered to the cache folder on worker12:
/var/lib/openqa/cache/tests/sle/products/sle/main.pm
/var/lib/openqa/cache/tests/sle/tests/console/hostname.pm
/var/lib/openqa/cache/tests/sle/tests/slenkins/login.pm

Normal test (no changes in the test) with the update os-autoinst:

- http://e212.suse.de/tests/11108 - SLENKINS: select_virtio_console
- http://e212.suse.de/tests/11109 - CREATE_HD
- http://e212.suse.de/tests/11110 - MAU


Triggering the error:

>         type_string "echo 1 > /proc/sys/kernel/sysrq\n";
>         type_string "echo c > /proc/sysrq-trigger\n";


on the files:

- /var/lib/openqa/cache/tests/sle/tests/console/hostname.pm
- /var/lib/openqa/cache/tests/sle/tests/slenkins/login.pm


Putting the exceptions to look for on `/var/lib/openqa/cache/tests/sle/products/sle/main.pm` :

With: 

`$testapi::distri->set_expected_serial_failures(soft=>[quotemeta "Kernel panic"], hard=> [quotemeta "This is a hard failure"] );`

- http://e212.suse.de/tests/11121#step/slenkins_node/3 - SLENKINS: select_virtio_console

On a normal console the kernel crash just kills the test, so there's nothing i can do:

- http://e212.suse.de/tests/11123 - MAU
- http://e212.suse.de/tests/11122#step/hostname/9 - CREATE_HD


With: 
`$testapi::distri->set_expected_serial_failures(hard=>[quotemeta "Kernel panic"], soft=> [quotemeta "This is a hard failure"] );`


- http://e212.suse.de/tests/11125#step/slenkins_node/3 - SLENKINS: select_virtio_console

mudler:
looks good to me in general :+1: , i would just add some minor style changes note which are not relevant to the functionality, so feel free to ignore them

autotest.pm
```diff
@@ -345,6 +350,49 @@ sub loadtestdir {
...
+                        $t->record_resultfile('Serial Failure', "Serial error: $line", result => 'fail');
+                        $t->{result} = 'fail';
+                        $die = 1;
+                    }
+                }
+            }
+        }
+    }
+
+    die if $die;
```
mudler: I'd add a message indicating the failure (just for sake of completeness)

autotest.pm
```diff
@@ -345,6 +350,49 @@ sub loadtestdir {
...
+    myjsonrpc::send_json($isotovideo, {cmd => 'read_serial', position => $serial_file_pos});
+    my $json = myjsonrpc::read_json($isotovideo);
+
+    my $die = 0;
+    # loop line by line
+    for my $line (split(/^/, $json->{serial})) {
+        chomp $line;
+        # only two keys allowed
+        for my $type (qw(soft hard)) {
+            for my $regexp (@{$failures->{$type}}) {
```
mudler: I guess you can get rid of the double loop (and the double nesting) by looping like this:
 ```perl
for(@{$failures->{hard}}, @{$failures->{soft}}) {
....
 }
``` 

or alternatively just using hash slices if you want to group even further.

File: autotest.pm [hunk 11]
```diff
@@ -298,7 +300,10 @@ sub runalltests {
                 make_snapshot($t->{fullname});
             }
 
-            eval { $t->runtest; };
+            eval {
+                $t->runtest;
+                $serial_file_pos = parse_serial_output($serial_file_pos, $t);
```
mudler: Since you are passing the test reference, i guess this method can actually go to the test object, which would call the backend specific function (if available, or the opposite way around)

backend/baseclass.pm
```diff
@@ -699,13 +699,26 @@ Returns the output on the serial device since the last call to set_serial_offset
...
+=head2 read_serial
+
+Returns the output and the offset after reading on the serial device from position
+
+=cut
+
+sub read_serial {
+    my ($self, $position, $whence) = @_;
 
     open(my $SERIAL, "<", $self->{serialfile});
```
mudler: I know this is not part of the changes, but i see no handling of error in case open fails :(

autotest.pm
dasantiago: @mudler done :+1:

autotest.pm
dasantiago: How would i then identify the type of failure, required for the next IF ? I would then to store more info, like the ammount of exceptions in each type, before starting looping... To be honest i don't think it's worth it as it would make the algorithm confusing.

backend/baseclass.pm
okurz: there is, it's the `use autodie ':all';`

autotest.pm
dasantiago: Done.

validation runs:

1-
Code:

> $testapi::distri->set_expected_serial_failures(soft=>[
>                                                    quotemeta("gcc version 4.8.5 (SUSE Linux)")],
>  );

http://tragicbox.suse.cz/tests/1208#step/welcome/10

2- 
Code: 

> $testapi::distri->set_expected_serial_failures(soft=>[
>                                                    quotemeta("gcc version 4.8.5 (SUSE Linux)")],
>                                                hard=>['ERROR']
>  );

http://tragicbox.suse.cz/tests/1215#step/welcome/14


3- 
Code: 

> $testapi::distri->set_expected_serial_failures(
>                                                hard=>['ERROR']
>  );

http://tragicbox.suse.cz/tests/1216#step/welcome/12

okurz:
I wonder is the only real documentation about this feature in the PR description? I suggest looking into extending the POD you wrote with the example code from the PR description or os-autoinst documentation or  extending os-autoinst-distri-opensuse

rwx788:
Somehow this didn't work for me. I tried to set in main.pm and then echoed text, which matches the pattern, to tty, but nothing happened. Then I also tried to do that in the test modifying serial_failures hash, but that didn't work either. Am I missing something?

dasantiago:
Where are you trying? Can you confirm that your openqa isn't updating the tests via rsync overwriting your changes?

jknphy:
Just to note, that the name of the function is not what's written on the first initial message. It's `set_expected_serial_failures` and not `set_serial_failures` ;-)

PR #933, opened 2018-03-16
Allow setting console tty and other args

Martchus:
This allows to alter the console's arguments after
the initialization via add_console():

* console('x11')->set_arg(tty => 7);
* console('x11')->set_tty(7);

See https://progress.opensuse.org/issues/25702

---

Contains test-cases, but I haven't tested whether this works in a test run.

### DIFF
--- a/backend/console_proxy.pm
+++ b/backend/console_proxy.pm
@@ -37,6 +37,9 @@ sub DESTROY {
     # nothing to destroy but avoid AUTOLOAD
 }
 
+# handles the attempt to invoke an undefined method on the proxy console object
+# using query_isotovideo() to invoke the method on the actual console object in
+# the right process
 sub AUTOLOAD {
 
     my $function = our $AUTOLOAD;
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -90,4 +90,23 @@ sub is_serial_terminal {
     return 0;
 }
 
+sub set_args {
+    my ($self, %args) = @_;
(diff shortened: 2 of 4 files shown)


Martchus:
@okurz I actually already did this to debug/improve the interactive mode (although it didn't help that much).

---

But I have really no idea how to convince autotools to use the right include path. Changing the line `unshift @INC, '..';` also doesn't help.

Only starting the test directly with `PERL5LIB=.:$PERL5LIB prove -v t/03-testapi.t` works.

Martchus:
Fixed the include using `PERL5LIB="..:$$PERL5LIB"`.

coolo:
prove -l . sounds saner - did you try?

Martchus:
So this autotools check target just calls prove, too? How would I add additional arguments such as `-I` to the prove call of the check target then? In CMake I would know, but in autotools I haven't even a clue where to look it up.

coolo:
No, I was wrong. They call the .t file. So the only option you have is the BEGIN {} block in the .t file to add INCPATH.

Martchus:
> So the only option you have is the BEGIN {} block in the .t file to add INCPATH.

Which didn't do the trick when I tried it last Friday. In fact
```
BEGIN {
    unshift @INC, '..';
}
```
is already present but I doesn't work. I also tried to add a few other directories ('.', '../..', the absolute directory) and it changed nothing.

PR #934, opened 2018-03-22
Remove specific qemu code

dasantiago:
Remove the specific code for qemu

### DIFF
--- a/backend/driver.pm
+++ b/backend/driver.pm
@@ -143,16 +143,7 @@ sub mouse_hide {
     my ($self, $border_offset) = @_;
     $border_offset ||= 0;
 
-    # TODO: come up with a better solution - this is qemu specific.
-    my $counter = 0;
-    my $rsp;
-    while ($counter < 10) {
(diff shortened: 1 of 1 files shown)


backend/driver.pm
```diff
@@ -143,16 +143,8 @@ sub mouse_hide {
     my ($self, $border_offset) = @_;
     $border_offset ||= 0;
 
-    # TODO: come up with a better solution - this is qemu specific.
     my $counter = 0;
```
coolo: looks like you don't need $counter then

dasantiago:
Just some comment on mouse hidding:

Unfortunately there's no direct way to hide the mouse or to verify if the mouse is hidden. 

For Example:
If we are at the display manager and we hide the mouse, our internal state will change to have the mouse "hidden" in the bottom right corner. However when we login into gnome the mouse automatically jumps to the middle of the screen, so our internal state is not correct anymore :-(

At the moment, i believe that the best way to check if the mouse is present or not would be to have a mouse needle and search for it around the screen.

backend/driver.pm
dasantiago: Right. Removed.

PR #909, opened 2018-01-16
WIP: Introduce a slim pvm backend using ssh

coolo:
This requires ssh access to a novalink installation, but currently does
little around it.

PR #935, opened 2018-03-23
Add username parameter for sshCommand

nicksinger:
This partly contains the changes from #909 and enables consoles to use a different user then root when using sshCommand. See it as preparation for the future where we eventually want to user other users then root.

### DIFF
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -52,9 +52,9 @@ sub screen {
 
 # helper function
 sub sshCommand {
-    my ($self, $host, $gui) = @_;
+    my ($self, $username, $host, $gui) = @_;
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no root\@$host";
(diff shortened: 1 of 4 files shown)


consoles/console.pm
```diff
@@ -52,9 +52,9 @@ sub screen {
 
 # helper function
 sub sshCommand {
-    my ($self, $host, $gui) = @_;
+    my ($self, $username, $host, $gui) = @_;
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no root\@$host";
+    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
```
mnowaksuse: Fixing all instances of `sshCommand` is good but making sure `$username` defaults to `root` - just for being on the safe side - is better in my mind.

foursixnine: @mnowaksuse xD Sorry

coolo: as the username is the new first argument, it's a bit hard to leave it out

PR #938, opened 2018-03-30
Add test option _SKIP_POST_FAIL_HOOKS to save time on test development

okurz:
Tested manually with a test on openQA.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -302,6 +302,7 @@ sub post_run_hook {
 
 sub run_post_fail {
     my ($self, $msg) = @_;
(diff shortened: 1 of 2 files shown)


mnowaksuse:
Maybe singular, i.e. `_SKIP_POST_FAIL_HOOK` fits better in the sense "skip `post_fail_hook` sub". Otherwise LGTM.

coolo:
I can see arguments for both singular and plural - so let's merge as it is.

coolo:
Please add a snippet to the Writing Tests documentation

PR #940, opened 2018-04-04
Add OFFLINE mode to QEMU backend

asdil12:
This mode will disable all networkinterfaces for the VM.

https://progress.opensuse.org/issues/33262

Required by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4767

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -570,6 +570,10 @@ sub start_qemu {
     my $num_networks = 1;
     $num_networks = max($num_networks, scalar @nicmac, scalar @nicvlan, scalar @tapdev);
 
+    if ($vars->{OFFLINE_SUT}) {
+        $num_networks = 0;
+    }
+
     for (my $i = 0; $i < $num_networks; $i++) {
         # ensure MAC addresses differ globally
         # and allow MAC addresses for more than 256 workers (up to 16384)
@@ -680,6 +684,7 @@ sub start_qemu {
         gen_params @params, 'machine', $vars->{QEMUMACHINE} if $vars->{QEMUMACHINE};
         gen_params @params, 'cpu',     $vars->{QEMUCPU}     if $vars->{QEMUCPU};
         gen_params @params, 'device',  'virtio-rng-pci'     if $vars->{QEMU_VIRTIO_RNG};
+        gen_params @params, 'net',     'none'               if $vars->{OFFLINE_SUT};
 
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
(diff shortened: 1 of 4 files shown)


backend/qemu.pm
```diff
@@ -570,6 +570,10 @@ sub start_qemu {
     my $num_networks = 1;
```
Martchus: What about:
```
my $num_networks = max($vars->{OFFLINE} ? 0 : 1, scalar @nicmac, scalar @nicvlan, scalar @tapdev);
```

coolo: not sure this is an improvement

asdil12: For some reason it doesn't seem to work this way.

asdil12:
set to `OFFLINE_SUT`

distribution.pm
```diff
@@ -192,6 +192,12 @@ sub script_output {
     $args{type_command} //= length($script) < 80;
     my $script_path = "/tmp/script$marker.sh";
 
+    # prevent use of network for offline installations
+    if (testapi::get_var 'OFFLINE_SUT') {
+        testapi::record_info('forced type_cmd', "Forced typing in the command as we are offline");
```
coolo: I think typing already includes the 'in' - there is little out typing at least.

testapi.pm
```diff
@@ -1847,19 +1847,24 @@ sub upload_logs {
...
 
-    bmwqemu::log_call(file => $file, %args);
-    my $basename = basename($file);
-    my $upname   = ($args{log_name} || $autotest::current_test->{name}) . '-' . $basename;
-    my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
-    $cmd .= autoinst_url("/uploadlog/$basename");
-    if ($failok) {
-        # just use script_run so we don't care if the upload fails
-        script_run($cmd, $timeout);
+    if (get_var 'OFFLINE_SUT') {
```
coolo: use get_var()

testapi.pm
```diff
@@ -1847,19 +1847,24 @@ sub upload_logs {
...
+        record_info('upload skipped', "Skipped uploading log file '$file' as we are offline");
```
coolo: return here - or refactor to avoid yet another if/else

testapi.pm
```diff
@@ -1891,16 +1896,21 @@ C<$nocheck> parameter:
...
     my ($file, $public, $nocheck) = @_;
 
-    bmwqemu::log_call(file => $file, public => $public, nocheck => $nocheck);
-    my $cmd = "curl --form upload=\@$file ";
-    $cmd .= "--form target=assets_public " if $public;
-    my $basename = basename($file);
-    $cmd .= autoinst_url("/upload_asset/$basename");
-    if ($nocheck) {
-        type_string("$cmd\n");
+    if (get_var 'OFFLINE_SUT') {
```
coolo: same here. no else but return

asdil12:
@coolo updated.

PR #942, opened 2018-04-12
[WIP] Qemu refactor

richiejp:
The primary motivation for this is to allow QEMU to use external snapshots through the native QMP interface which should result in much higher reliability. Currently we use internal snapshots through a QMP wrapper for the HMP.

Note that the new method is actually a combination of a migration to file and block device snapshots. In the old method this complexity was hidden at the expense of control. The new method requires restarting QEMU which is a lot of work, but has some advantages regardless of the snapshot method.

The following are some reasons why external snapshots and restarting QEMU are better:
1) Some of the code in the internal snapshot mechanism is not widely used or tested. Upstream developers claim that the external migration code is much more mature and is likely to remain so. It seems that we have suffered from a number of these internal snapshot bugs.
2) In theory devices in QEMU can retain state from before a snapshot. There is no way to know that all devices are in a consistent state except by restarting QEMU.
3) Internal snapshots provide no control over where the virtual machine RAM and state are saved to or what method is used to save it (this once resulted in install_ltp publishing an image with 100GB of snapshot data in it).
4) The HMP interface is synchronous and lacks proper error reporting. (the QMP wrapper does not solve these issues)

The downside of this method is that it is more complicated, but it seems QEMU prefers to give control over ease-of-use.

In order to restart QEMU I had to create a model of QEMU's block device state which can be modified and used to dynamically generate the QEMU command line arguments and `qemu-img` commands. I think that this model should be extended to cover all QEMU command line arguments, but for now it just covers block devices and related devices such as controllers.

Some advantages of the model are:
1) It allows os-autoinst to track what it thinks QEMU's state should be and recreate a previous QEMU instance based on that state. Of course this is what snapshots require.
2) It can easily be used for unit testing.
3) It allows serialisation and deserialisation into arbitrary formats for debugging or as an alternative to os-autoinsts vars configuration.
4) It splits the configuration code into modules making it easier to figure out what code is related to each other.
5) Other backends could adopt the same model to help unify them

The disadvantage is that it introduces some abstraction which may be more difficult to initially understand. An alternative to all of this may be to use libvirt although it may not release us from having to track state ourselves and also I wonder if it exposes all the functionality we will ever need. Someone should probably try it.

Also, while trying to figure out what would break when restarting QEMU I discovered a whole load of unrelated problems. Some of which I have tried to fix, but I haven't implemented the deep changes which are really necessary in the long run.

Before attempting to refactor the QEMU backend I tried modifying QEMU instead so that it does not require a restart. This worked fine in the limited test I did, but after discovering how devices work in QEMU I decided it was a lot safer to restart QEMU. Probably a lot of people upstream thought I was crazy for wanting to patch QEMU instead of just restarting it, but then they haven't seen os-autoinst.


okurz:
I would like to ask about the disadvantages, the advantages are all fine:
* *Q1*: What is the impact on time behaviour, how does it change vs. old behaviour?
  * *Q1.1*: When loading a snapshot within the test, e.g. module fails, test falls back to lastgood snapshot
  * *Q1.2*: When starting a test with SKIPTO=
* *Q2*: Can you test interactive mode? Unless you did already but I doubt it because you did not even mention openQA anywhere ;)
* *Q3*: Does this have any implications on used space for the snapshot on the storage of the worker? E.g. could it happen that something depletes more storage than in before?
* *Q4*: Requirements regarding the qemu version or architecture?

richiejp:
Q1.1: So far there is no noticeable difference, in theory there will be no difference with OpenQA's usage because we do not use live migrations. If we did then the new method would be faster because it makes better use of asynchronous I/O. We have more control with the new method so it is probably easier to optimize. In the future we might be able to use differential RAM images which will be a significant improvement, but I am not sure whether that really exists yet.
Q1.2: I am guessing the new method will be faster as starting QEMU in a paused state and issuing loadvm is likely to be less efficient than starting in a migration state and loading the RAM from disk. However it is probably negligible compared to the time it takes to read bytes from your disk.

Note, that I previously profiled migrations to disk and found that QEMU is fairly efficient at migrating to disk. With internal snapshots, they seem to work quickly most of the time, then break completely. So I don't think the average performance is important for now.

Q2: Yes, but I am not sure what scenario(s) you want me to test. I admit I am fairly ignorant of this feature although I am using OpenQA for all testing so far.

Q3: Maybe (in terms of temporary storage), but this is configurable depending on whether we are more concerned about speed or (ram)disk space (i.e. /var/lib/openqa/pool/). Currently the snapshots are created in a differential chain and the old ones are not merged (except when the disk is published (not implemented yet)). On the other hand internal snapshots seem to be fucked, so it might actually take up less space already. Like I said above, one of our drives increased by 100GB due to an internal snapshot which is difficult to explain.

Q4: The new method uses migrations to file which are better supported than internal snapshots. However I intend to enable snapshots for things like virtio GPU which will probably increase the necessary version. However I have added the `-only-migratable` switch which could allow us to start QEMU with migrations enabled and if it fails to start, disable migrations and remove `-only-migratable`, then try it again.

okurz:
> Q2: Yes, but I am not sure what scenario(s) you want me to test. I admit I am fairly ignorant of this feature although I am using OpenQA for all testing so far.

I suggest a test module which has some needles. E.g. take one from https://openqa.opensuse.org/tests?match=yast2_gui , delete some needles used in this scenario locally and try to use the interactive mode to create the needles.

richiejp:
Information regarding PFLASH: https://github.com/os-autoinst/os-autoinst/pull/754

richiejp:
`[2018-06-05T07:39:57.0165 UTC] [debug] running /usr/bin/qemu-img create -f qcow2 -o nocow=on raid/hd0 10G
[2018-06-05T07:39:57.0177 UTC] [debug] Unknown option 'nocow'
qemu-img: raid/hd0: Invalid options for file format 'qcow2'.`

I guess this is a newer feature.

richiejp:
QEMU 2.0.0 is far too ancient, Travis should be on 2.9+

coolo:
See https://github.com/os-autoinst/openQA/pull/1673

coolo:
I consider integrating the test suites of both repos - if not even merging both repos.

richiejp:
Well I can definitely see some advantages to merging. Although it might be difficult to undo in the future.

For now I would rather someone just copy and paste the docker solution into another PR and I can rebase onto that. Depending on how the merging is done, it seems likely to be more problematic which could keep this PR in flight for longer.

PR #944, opened 2018-04-15
Fix missing text thumbnail after needle related test failures

okurz:
This fixes a regression that needle related test failures did not create a
message "# Test died: ..." in the logfile as well as no text thumbnail with
information and what happened. This regression must have been introduced about
at least a year ago. I could not go further in my bisecting without needing to
come up with test changes.

It feels like there should be a better solution than checking for
"InternalException" but then excluding "FailedNeedle" again.

### DIFF
--- a/OpenQA/Exceptions.pm
+++ b/OpenQA/Exceptions.pm
@@ -1,4 +1,4 @@
-# Copyright © 2016 SUSE LLC
+# Copyright © 2016-2018 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


okurz:
@asdil12 I think this is what you mentioned as well. AFAIK we had this feature some time ago but then nobody ever raised it again properly. You mentioned it, thank you :)

basetest.pm
```diff
@@ -328,7 +330,8 @@ sub runtest {
     };
     if ($@) {
         # copy the exception early
-        my $internal = Exception::Class->caught('OpenQA::Exception::InternalException');
+        my $exception = Exception::Class->caught('OpenQA::Exception::InternalException');
+        my $internal = blessed $exception && !$exception->isa('OpenQA::Exception::FailedNeedle');
```
coolo: hmm, why is FailedNeedle an InternalException. I think that's the oversight here

okurz: Well, it has been like this since commit 1759b5e5 with the description
    
    For specific exceptions (like Needle Failed) we don't record an extra
    test result, only for exceptions thrown unexpectedly

so could you think of why you did it like this?

coolo:
let's try

PR #945, opened 2018-04-16
Allow passing parameters to activate_console method

rwx788:
Motivated by [poo#32926](https://progress.opensuse.org/issues/32926),
there we type user name, even though have not switched to correct tty
yet. For SLE15 and SLE12 we even have different tty for the same console
(e.g. root-console). To improve this scenario and not to break existing
tests we want to pass special parameter to assert selected tty.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1357,7 +1357,7 @@ Select the named console for further C<testapi> interaction (send_text,
 send_key, wait_screen_change, ...)
 
 If this the first time, a test selects this console, the distribution
-will get a call into activate_console('root-console', $console_obj) to
+will get a call into activate_console('root-console', $console_obj, @args) to
 make sure to actually log in root. For the backend it's just a C<tty>
 object (in this example) - so it will ensure the console is active,
 but to setup the root shell on this console, the distribution needs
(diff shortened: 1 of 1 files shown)


coolo:
I guess we should have increased the interface version for such a thing?

testapi.pm
```diff
@@ -1383,7 +1383,7 @@ sub select_console {
         if ($autotest::last_milestone) {
             push(@{$autotest::last_milestone->{activated_consoles}}, $testapi_console);
         }
-        $testapi::distri->activate_console($testapi_console);
+        $testapi::distri->activate_console($testapi_console, @args);
```
okurz: wait, you forward the same args into activate_console as well as console_selected? I wonder if this is not confusing.

rwx788: @okurz That's the thing I also do not like, but it will be more mess if we try to get two hashes of settings and separate them. And in our distri specific implementations we still can have parameters which are valid for both callback functions.
@coolo I actually checked some other PRs and they do not contain changes to version. But it makes sense.

rwx788:
@coolo I actually checked some other PRs and they do not contain changes to version. But it makes sense.

coolo:
It's still pretty recent and we're still experimenting with a balance when to bump :)

PR #947, opened 2018-04-18
Increase timeout send_key_until_needlematch for PowerPC

michelmno:
because timing issue on o3
https://progress.opensuse.org/issues/35185


coolo:
This is the wrong place, sorry. We provide a scale factor for timeouts - and I don't want hardware specifics hacked into the API

coolo:
TIMEOUT_SCALE=3 then

PR #948, opened 2018-04-19
Get distri value and avoid import cycles

mudler:
See: https://progress.opensuse.org/issues/34522


coolo:
we decided against that

PR #949, opened 2018-04-20
Do not quote -append qemu parameter

rwx788:
For some reason quoting doesn't work well with kernel parameters we pass
to iPXE kernel. We were able to see the error which was "' dhcp" command
not found. This regression was introduced in [5467a37f8](https://github.com/os-autoinst/os-autoinst/commit/5467a37f8) when iBFT test stopped working.

Also, when running qemu from console with quotes, it's able to configure network.

We used openqaworker12 and were able to reproduce issue there and tried hot patch. 

[Verification run](http://g226.suse.de/tests/1720).
[Run with old code](http://g226.suse.de/tests/1721#).
See [poo#23554](https://progress.opensuse.org/issues/23554).


okurz:
LGTM

PR #952, opened 2018-04-22
Add wait_screen_change option on send_key

okurz:
While doing this, deprecating do_wait which is calling wait_idle which we want
to avoid as it is too much relying on a backend specific implementation which
can not be made available on enough backends.

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -90,7 +90,7 @@ sub fake_read_json {
         return {ret => {x => 100, y => 100}};
     }
     else {
-        print "not implemented \$cmd: $cmd\n";
+        print "mock method not implemented \$cmd: $cmd\n";
     }
     return {};
 }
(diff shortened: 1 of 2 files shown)


okurz:
should have been adressed already more than one year ago. I don't know what happened on previous branches. https://github.com/os-autoinst/os-autoinst/pull/848 was the last branch for this change in 2017-09, let's retry.

okurz:
I don't understand what happens in the travis CI job, "Can't open 'testresults/basetest-2.txt' for writing: 'No such file or directory' at (eval 574)[/home/travis/perl5/perlbrew/perls/5.18-extras/lib/5.18.4/Fatal.pm:1164] line 155"? Seems to be reproducible on travis but not locally for me. Don't have any good idea now …

Martchus:
Can you rebase? Let's see what the CI says now.

okurz:
Also here the tests are fine just that the coverage again complains which is weird because I added quite some tests

PR #954, opened 2018-04-23
Add proceed_on_failure option

asmorodskyi:


distribution.pm
```diff
@@ -175,7 +175,7 @@ sub script_sudo {
 
 =head2 script_output
 
-    script_output($script, [ $wait, type_command => ? ]) 
+    script_output($script, [ $wait, type_command => ?, die_on_failure => ?])
```
foursixnine: Update the POD explaining the parameter, and you also need to increase the [isotovideo interface](https://github.com/os-autoinst/os-autoinst/blob/master/isotovideo#L55)

distribution.pm
```diff
@@ -191,6 +191,9 @@ sub script_output {
     # if script length is lower there is no point to proceed with more complex solution
     $args{type_command} //= length($script) < 80;
     my $script_path = "/tmp/script$marker.sh";
+    # in some cases client wants validate output of failed script
```
foursixnine: Who is the client?, Can you rephrase this?

distribution.pm
```diff
@@ -238,7 +241,7 @@ sub script_output {
     my $output = testapi::wait_serial("SCRIPT_FINISHED$marker-\\d+-", $wait, 0, record_output => 1)
       || croak "script timeout";
 
-    croak "script failed" if $output !~ "SCRIPT_FINISHED$marker-0-";
+    croak "script failed with : $output" if $output !~ "SCRIPT_FINISHED$marker-0-" && $args{die_on_failure};
```
coolo: this deserves a proper if

distribution.pm
asmorodskyi: sure I can update isotovideo ver number but I thought that you updating it only in case of not backward compatible changes ?

distribution.pm
asmorodskyi: this is API function , usually they called by consumers (or clients) of API . If I will replace "client" with "consumer" it will be ok ? or can you give your version which will suit you for sure ?

distribution.pm
asmorodskyi: done

asmorodskyi: done

Martchus: I think we increase it on every change which adds new stuff.

BTW, do we ever non-backward-compatible changes on the test API? If yes, we maybe better use [semantic versioning](https://semver.org/).

asmorodskyi:
@foursixnine , @coolo - your requests implemented

testapi.pm
```diff
@@ -903,7 +903,7 @@ sub script_sudo {
 
 =head2 script_output
 
-  script_output($script [, $wait, type_command => 1])
+  script_output($script [, $wait, type_command => 1, die_on_failure => 1])
```
coolo: I know I'm late in the game but we usually show case how to pass non-default parameters. And if I understand it correctly, 1 is default. You want to allow the opposite by adding this paramter.

But having a die_on_failure => 0 sounds double negative. So give this a positive meaning - and make 0 the default. From what I get: proceed_on_failure => 1

asmorodskyi: the problem that `proceed_on_failure` which really ease understanding on function input param level , complicate things in actual place of usage - this is why I decided to go with `die_on_failure`. But let's try to follow you suggestion ...

asmorodskyi:
@coolo - your request implemented

testapi.pm
```diff
@@ -903,7 +903,7 @@ sub script_sudo {
 
 =head2 script_output
 
-  script_output($script [, $wait, type_command => 1])
+  script_output($script [, $wait, type_command => 1, proceed_on_failure => 0])
```
coolo: this should be => 1 to be consistent with the type_command where we list the non-default usage

asmorodskyi:
@coolo - your request implemented

distribution.pm
okurz: Yes, we do backward-incompatible changes. This one is not

okurz:
Even more late to the game I wonder about the benefit of this change. Now by setting the new option one can not easily determine if the command actually failed or not. If one would really like to ignore the failure then in before it could have been just handled in the sut, e.g. by writing something like `script_output 'failing_command; true'` so I wonder if this new option is actually needed

coolo:
I find the proceed_on_failures => 1 easier to spot

foursixnine:
now even later to this game: proceed_on_failure => 1 confuses certain [calls](https://progress.opensuse.org/issues/38606)

asmorodskyi:
@foursixnine  false alarm, I just called a function in not documented way

PR #955, opened 2018-04-23
Set distribution earlier in isotovideo

foursixnine:
Cover the corner cases where a test suite is not setting the test
distribution.

This is related [poo#34522](https://progress.opensuse.org/issues/34522) and  #948


isotovideo
```diff
@@ -260,6 +260,9 @@ $bmwqemu::vars{TEST_GIT_HASH} = $test_git_hash;
 # is not supposed to talk to the backend directly
 ($cpid, $cfd) = commands::start_server($bmwqemu::vars{QEMUPORT} + 1);
 
+# set a default distribution if the tests don't have one
+$testapi::distri ||= distribution->new;
```
coolo: There is no point in doing || - the only place that sets it normally is the main.pm

PR #956, opened 2018-04-24
svirt KVM: Set <on_reboot> to 'destroy'

mnowaksuse:
For restarts on KVM we don't strictly need to destroy & reinitialise the
VM as we do on Xen, but it's convenient to have it as the test code
already expects this.


coolo:
trusting you there

PR #958, opened 2018-05-03
Enhanced live

coolo:


coolo:
the HTML and javascript is not supposed to stay - but it's way easier to develop if you only need to restart one service

coolo:
Now providing status to WS clients as it happens

Martchus:
Actually the developer mode looks like a nice feature to me. Maybe you can make it stay but disable it by default?

coolo:
For the time being sure, but the end goal is definitely to maintain this in openQA instead

coolo:
superseded by #966

PR #959, opened 2018-05-03
Allow environment variable to point to OPENQA_LIBPATH

foursixnine:
To help with new a parser development, an environment variable comes in handy to specify where the current openQA code is

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1699,7 +1699,7 @@ sub parse_extra_log {
         local $@;
         # We need to touch @INC as specific supported format are split
         # in different classes and dynamically loaded by OpenQA::Parser
-        local @INC = (OPENQA_LIBPATH, @INC);
+        local @INC = ($ENV{OPENQA_LIBPATH} // OPENQA_LIBPATH, @INC);
         eval {
             require OpenQA::Parser;
             OpenQA::Parser->import('parser');


testapi.pm
```diff
@@ -1699,7 +1699,7 @@ sub parse_extra_log {
         local $@;
         # We need to touch @INC as specific supported format are split
         # in different classes and dynamically loaded by OpenQA::Parser
-        local @INC = (OPENQA_LIBPATH, @INC);
+        local @INC = ($ENV{OPENQA_LIBPATH}, OPENQA_LIBPATH, @INC);
```
coolo: wouldn't an empty value be troublesome?

Martchus: Maybe make it `local @INC = ($ENV{OPENQA_LIBPATH} // OPENQA_LIBPATH, @INC);` then? `OPENQA_LIBPATH` is already supposed to be the default anyways.

And this way it would also override the default which is better. Because if you set the env variable incorrectly it would otherwise silently fallback to the packaged version and you would wonder why your changes have no effect.

foursixnine: Not really: http://phobos.suse.de/tests/11075,

It will only complain when the parser cannot be found: 

Use of uninitialized value in require at (eval 130) line 1 [in the autoinst-log.txt](http://phobos.suse.de/tests/11074#)

coolo: is this opposing @Martchus ? Because his suggestion makes sense to me

foursixnine: @coolo Didin't see @Martchus comment :P, changing that... In any case, as it is, is not troublesome.

testapi.pm
```diff
@@ -1694,12 +1694,12 @@ sub parse_extra_log {
 
     $file = upload_logs($file);
     my @tests;
-
+    use Data::Dumper;
```
coolo: you slipped in unrelated changes

foursixnine: Holy!

foursixnine: Done xD

PR #961, opened 2018-05-09
localXvnc: Output the PIDs of the involved processes

coolo:
see poo#34003


coolo:
yeah - the ticket is still open.

PR #966, opened 2018-05-25
Allow to pause test execution at specific test 

Martchus:
This is basically following [@coolo's comment](https://progress.opensuse.org/issues/36442#note-5).

It causes autotest to block in `myjsonrpc::read_json()` simply by isotovideo not being responsive. This way the test execution is paused.

I also tried whether it would be possible to resume form that situation again but it didn't work. See the FIXME in the code. But note that it also didn't work without that 2nd read in `autotest::query_iso_to_video` and isotovideo not answering commands in paused state at all.

To test it, see the commit message. I used https://github.com/os-autoinst/openQA/pull/1661 to send commands to the isotovideo command server.


coolo:
you misunderstood - but we clarify that on monday :)

Martchus:
I removed handling the postponed command in autotest. In isotovideo I still save that postponed command to be able to resume. However, it doesn't work. Currently it looks like this:

1. I send `{"cmd":"set_pause_at_test","name":"installation-welcome"}` to command server.
2. When autotest calls `set_current_test`, isotovideo is actually set into paused mode. `pausing test execution of installation-welcome` is logged and the next command isn't passed to the backend anymore: `paused, so not passing backend_save_snapshot to backend`.
3. autotest is actually blocked.
4. I send `{"cmd":"resume_test_execution"}`  to the command server to resume.
5. isotovideo actually continues processing the last command: `resuming, continue passing backend_save_snapshot to backend`
6. Then I only get:
    ```
    BYTES {"save_snapshot":{"json_cmd_token":"pfEsEcpp","name":"installation-welcome"},"json_cmd_token":"lAoCcwIo"}
    [2018-05-28T12:32:23.0986 CEST] [debug] EVENT {"timestamp":{"seconds":1527503465,"microseconds":249613},"event":"NIC_RX_FILTER_CHANGED","data":{"path":"/machine/peripheral-anon/device[0]/virtio-backend"}}
    [2018-05-28T12:32:24.0019 CEST] [debug] EVENT {"timestamp":{"seconds":1527503544,"microseconds":19036},"event":"STOP"}
    [2018-05-28T12:32:24.0355 CEST] [debug] GET "/MR6OaPkvzVtO2P5t/isotovideo/status"
    [2018-05-28T12:32:24.0355 CEST] [debug] Routing to a callback
    [2018-05-28T12:32:24.0355 CEST] [debug] 200 OK (0.000637s, 1569.859/s)
    [2018-05-28T12:32:24.0855 CEST] [debug] GET "/MR6OaPkvzVtO2P5t/isotovideo/status"
    [2018-05-28T12:32:24.0855 CEST] [debug] Routing to a callback
    [2018-05-28T12:32:24.0856 CEST] [debug] 200 OK (0.000652s, 1533.742/s)
    [2018-05-28T12:32:25.0333 CEST] [debug] EVENT {"event":"RESUME","timestamp":{"microseconds":333375,"seconds":1527503545}}
    BYTES {"ret":0,"json_cmd_token":"UDIfkwov"}
    [2018-05-28T12:32:25.0356 CEST] [debug] GET "/MR6OaPkvzVtO2P5t/isotovideo/status"
    [2018-05-28T12:32:25.0356 CEST] [debug] Routing to a callback
    ...
    ```
    But the blocking call in autotest never returns (there's no `autotest: resumed` in the log).

Likely the way the delayed answer is sent is wrong. Also, it would maybe make sense to let autotest wait directly after `set_current_test`.

coolo:
> Also, it would maybe make sense to let autotest wait directly after set_current_test. 
> This way it wouldn't be required to remember at which command we stopped 
> because it would always be set_current_test.

This would work for this feature, but the 'needle refresh' use case is the next stop where we wonder how to resume normal execution.

Martchus:
Found the reason why it didn't continue. I sent the reply not to the right `$r`. I assume that one letter variable name stands for read (which doesn't make that much sense because we also write to it).

Martchus:
Extended openQA fullstack test: https://github.com/Martchus/openQA/commit/5cd833900936658d7137276f736a33d8f1b523a7

coolo:
you're allowed to rename variables :)

isotovideo
```diff
@@ -386,133 +395,185 @@ sub check_asserted_screen {
...
+    if ($rsp->{cmd} eq 'resume_test_execution') {
+        # print/send debug info
+        print($stop_passing_commands_to_backend ?
+              "isotovideo: test execution will be resumed\n"
+            : "isotovideo: resuming test execution requested but not paused anyways\n"
+        );
+        myjsonrpc::send_json($cmd_srv_fd, {resume_test_execution => $postponed_command});
+
+        # no longer stop passing commands to backend
+        $stop_passing_commands_to_backend = 0;
```
foursixnine: How about renaming it to `pass_commands_to_backend` it's a bit less confusing when reading sentences/comments like this.

isotovideo
```diff
@@ -386,133 +395,185 @@ sub check_asserted_screen {
...
+        $bmwqemu::backend->_send_json({cmd => 'set_serial_offset'});
+        $current_test_name      = $rsp->{name};
+        $current_test_full_name = $rsp->{full_name};
+
+        # send debugging info for ws clients
+        myjsonrpc::send_json($cmd_srv_fd, {set_current_test => $current_test_name});
+        if (   $pause_test_name
+            && $current_test_name
+            && $current_test_full_name
+            && ($pause_test_name eq $current_test_name || $pause_test_name eq $current_test_full_name))
```
foursixnine: Why `$current_test_name` and `$current_test_full_name`? what's the difference between both?

isotovideo
```diff
@@ -322,18 +322,27 @@ if ($ENV{RUN_DEBUGVIEWER}) {
...
-my $needinput   = 0;
-
+# store the name of the current test
 my $current_test_name;
+my $current_test_full_name;
+
+# store the name of the test where the test execution should be paused
+my $pause_test_name = $bmwqemu::vars{PAUSE_AT};
+
+# when paused, stop passing commands to the backend
```
foursixnine: I wonder if this should be a property of the test variables? so that the worker/webui comunication can easily tell if isotovideo is accepting or not more commands (on top of the job state...)

Martchus: Ok, maybe later. Not sure how I can add/update test variables from `isotovideo` so the changes are promoted to the child processes and the web UI.

isotovideo
Martchus: Mh... I made this negative because if the variable is set, that's the special case.

PR #967, opened 2018-05-25
Correct return values for match_has_tag

AdamWill:
Happened to notice today the explanation of `match_has_tag`'s
return values is, strictly speaking, incorrect. It only returns
undef if no needle match has yet taken place when it is called.
If there is a most recent needle match but it does not have the
specified tag, it returns 0, not undef.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -380,7 +380,9 @@ sub check_screen {
 
   match_has_tag($tag);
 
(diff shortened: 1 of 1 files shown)


coolo:
not that the difference matters in real life perl, but correct is correct

AdamWill:
#RealLifePerl

coolo:
sounds like $you are not \&invited to our @family_dinners.

PR #968, opened 2018-05-30
ppmclibs/tinycv_impl.cc code improvements

aliffey:
- Remove using std to avoid namespace clash with cv
- Pass by reference for speeding up
- Include c++ libraries instead of c
- Adding const for unwanted changes

Builds fine on all platforms [1] and passes tests.

[1] https://build.opensuse.org/package/show/home:aliffey:branches:devel:openQA/os-autoinst

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -16,8 +16,8 @@
 #include <cerrno>
 #include <exception>
 #include <iostream>
-#include <stdint.h>
-#include <stdio.h>
+#include <cstdint>
+#include <cstdio>
 #include <sys/time.h>
 
 #include <algorithm> // std::min
@@ -37,14 +37,13 @@
 #define VERY_SIM 1000000.0
 
 using namespace cv;
-using namespace std;
 
 struct Image {
-    cv::Mat img;
-    mutable cv::Mat _preped;
-    mutable cv::Rect _prep_roi;
+    Mat img;
+    mutable Mat _preped;
+    mutable Rect _prep_roi;
 
-    cv::Mat prep(const Rect& roi) const
+    Mat prep(const Rect& roi) const
     {
         if (!_preped.empty()) {
             // Check if ROI is contained in the current ROI
@@ -160,9 +159,9 @@ std::vector<Point> minVec(const Mat& m, float min)
 // Used to sort a list of points to find the closest to the original
 struct SortByClose {
     SortByClose(int _x, int _y) { orig.x = _x, orig.y = _y; }
-    bool operator()(Point a, Point b)
+    bool operator()(const Point& a, const Point& b) const
     {
-        return cv::norm(orig - a) < cv::norm(orig - b);
+        return norm(orig - a) < norm(orig - b);
     }
     Point orig;
 };
@@ -333,21 +332,21 @@ Image* image_read(const char* filename)
 
 Image* image_from_ppm(const unsigned char* data, size_t len)
 {
-    vector<uchar> buf(data, data + len);
+    std::vector<uchar> buf(data, data + len);
     Image* image = new Image;
     image->img = imdecode(buf, CV_LOAD_IMAGE_COLOR);
     return image;
 }
 
-bool image_write(Image* s, const char* filename)
+bool image_write(const Image* const s, const char* filename)
 {
     return imwrite(filename, s->img);
 }
 
-vector<uchar>* image_ppm(Image* s)
+std::vector<uchar>* image_ppm(Image* s)
 {
     // reuse memory
-    static vector<uchar> buf;
+    static std::vector<uchar> buf;
     imencode(".ppm", s->img, buf);
     return &buf;
 }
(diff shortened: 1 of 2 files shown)


ppmclibs/tinycv_impl.cc
```diff
@@ -160,11 +157,11 @@ std::vector<Point> minVec(const Mat& m, float min)
 // Used to sort a list of points to find the closest to the original
 struct SortByClose {
     SortByClose(int _x, int _y) { orig.x = _x, orig.y = _y; }
-    bool operator()(Point a, Point b)
+    bool operator()(const cv::Point& a, const cv::Point& b) const
```
Martchus: `cv::Point` is just two ints so it is as big as a pointer. So I would actually prefer passing them by value.

ppmclibs/tinycv_impl.cc
```diff
@@ -333,22 +330,22 @@ Image* image_read(const char* filename)
...
-    vector<uchar> buf(data, data + len);
+    std::vector<uchar> buf(data, data + len);
     Image* image = new Image;
-    image->img = imdecode(buf, CV_LOAD_IMAGE_COLOR);
+    image->img = cv::imdecode(buf, CV_LOAD_IMAGE_COLOR);
     return image;
 }
 
-bool image_write(Image* s, const char* filename)
+bool image_write(Image* const s, const char* filename)
```
Martchus: Making arguments `const` is actually not common. Or did you intend making the pointer target const (`const Image*  s`)?

ppmclibs/tinycv_impl.cc
Amynka: passing by reference generalise better

ppmclibs/tinycv_impl.cc
Amynka: I think preventing arguments from being changed is a good practice also good point I think making the pointer const would be usefull as well..

Amynka: I think in this case removing using std and cv is good idea especially because you use std::vector and opencv vectors. Since this is all in one file it is making code better readable, understandable preventing programing errors in the future as well as namespace clashes.

mudler: I agree with @Amynka - namespace clashes aren't fun

ppmclibs/tinycv_impl.cc
mudler: Usually indeed it's opposite - prefer passing by reference over value, the latter only when you don't care about performance / memory, or you actually need a data copy.

ppmclibs/tinycv_impl.cc
Martchus: > because you use std::vector and opencv vectors

Ok, that's a good reason.

---

> making the pointer const would be usefull as well

Making the pointer *itself* const is actually not bad, but it is common exception to make function arguments not const (still talking about the pointer *itself*): https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable  
If you want to make it const nevertheless, at least be consistent and make the pointer to filename also const.

ppmclibs/tinycv_impl.cc
Martchus: There's no general 'pass things by reference' rule. The C++ Core Guidelines always distinguish between cheap to copy and ... just have a look at this table: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information

So I was basically just following this guideline: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

mudler: That would work, if we were not trying to optimize this code. So 'prefer simple' it's not what i'm looking for in this specific case.

How many points of that guide you think we are following in tinycv?

coolo:
I don't consider prefixing every function as 'code cleanup' - I would have understood if you removed the cv:: prefixes left over as 'cleanup'. 

So harmonize the parameter passing and give it a proper commit log - but don't *add* prefixes.

Amynka:
Dear Stephan,
thank you for the warm welcome on contributing to your project! :) As we discussed earlier with @Martchus use of namespaces in this code makes code more readable and prevents a lot of bugs. 

Usually, I would agree with you that clean up should be the opposite way but since this code doesn't even have proper comments and I truly doubt everyone who wants to contribute know what functions and vectors belong to cv or std. 

After all it is your project and I did this in my free time out of good will. If you think it was useless feel free to discard it :)

Thank you and have a nice day

aliffey:
Dear Stephan,
thank you for the warm welcome on contributing to your project! :) As we discussed earlier with @Martchus use of namespaces in this code makes code more readable and prevents a lot of bugs.

Usually, I would agree with you that clean up should be the opposite way but since this code doesn't even have proper comments and I truly doubt everyone who wants to contribute know what functions and vectors belong to cv or std.

After all it is your project and I did this in my free time out of good will. If you think it was useless feel free to discard it :)

Thank you and have a nice day

coolo:
Hi Amy,

This is not my problem, but feel welcome anyway :)

I guess more readable is very subjective, I personally find tons of :: less readable. If you feel the clash between cv and std is the problem, I can live with the std usages referenced - they are far less. But this code is opencv code - if someone has doubts where a function unknown comes from, the developer can pretty reliably guess it's cv. 

What I dislike (and this might be your reason why you reacted so ironically as I tend to express my disliking very directly) is people coming out of nowhere with 'cleanup' commits because it makes *their* reading of the code easier.

As such: take it as advice to improve your commit log messages. A 'Avoid implicit usage of namespace to improve readability' would have caused a different reaction at least from my side.

ppmclibs/tinycv_impl.cc
aliffey: In general on amd64 you are right it is the same size but as you can see below on arm it is not anymore.

   hell@beaglebone:~/test/build$ ./test-p
   Size of cv::Point: 8
   Size of ref to cv::Point: 4

If you are working with the structure before you call this function it is in cache and passing by reference is even more quicker than copy as it is in the cache and processor can work with it fast and lot of other reasons.

aliffey:
I am sorry I totally agree that my commit messages should be more explanatory and the compromise with using std:: sounds good. I will look into it.

Thanks for the feedback :)

coolo:
thx

PR #969, opened 2018-06-04
Allow where parameter for mutex unlocking

kravciak:
Any cluster job can be locked by jobid ($where) parameter.
Only children & parent jobs can be unlocked ($where not implemented).

Outcome is that we can't lock & unlock resource if it's sibling job.
First job in queue gets lock and it won't release until it dies.

Related PR: https://github.com/os-autoinst/openQA/pull/1677

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -72,10 +72,13 @@ sub mutex_try_lock {
 }
 
 sub mutex_unlock {
-    my ($name) = @_;
+    my ($name, $where) = @_;
+    my $param = {action => 'unlock'};
+    $param->{where} = $where if $where;
+
     bmwqemu::mydie('missing lock name') unless $name;
     bmwqemu::diag("mutex unlock '$name'");
-    my $res = api_call('post', "mutex/$name", {action => 'unlock'})->code;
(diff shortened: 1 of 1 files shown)


lockapi.pm
```diff
@@ -72,10 +72,13 @@ sub mutex_try_lock {
...
 
 sub mutex_unlock {
-    my ($name) = @_;
+    my ($name, $where) = @_;
+    my $param = {action => 'unlock'};
+    $param->{where} = $where if $where;
+
     bmwqemu::mydie('missing lock name') unless $name;
     bmwqemu::diag("mutex unlock '$name'");
-    my $res = api_call('post', "mutex/$name", {action => 'unlock'})->code;
```
coolo: where is all this documented btw?

kravciak: I think all documentation we have is [here](https://github.com/os-autoinst/openQA/blob/master/docs/WritingTests.asciidoc#test-synchronization-and-locking-api).

kravciak: small improvement in https://github.com/os-autoinst/openQA/pull/1686

PR #970, opened 2018-06-07
Qemu refactor travis experiment

richiejp:
Experiment with running tests inside a docker image: see #942


coolo:
Why not pull @dasantiago 's image?

richiejp:
It is an unnecessary external dependency from someone's personal repository. If the required dependencies changes then the image will need to be updated independently of the source. It probably makes sense to cache the image somehow, but I think a better solution is required.

coolo:
It makes sense to use the same image openQA uses and not duplicate this. Ignore that it's hosted in a personal account for now.

richiejp:
There are other problems as well, the openQA image needs a lot of work. Mine is much simpler, so it wouldn't take much to update or replace it. It also appears to be working.

docker/travis_test/prepare.sh
```diff
@@ -27,6 +27,7 @@ zypper in -y -C \
        qemu-tools \
        qemu-kvm \
        tar \
+       opencv-devel \
```
coolo: as much as you enjoy duplicating efforts - I will not merge this. We will only have one docker image

richiejp:
Well, whatever Perl critic found some problems in the QEMU refactor so I will go fix those for now.

PR #971, opened 2018-06-08
Wait for command typed in serial terminal before running it

aaannz:
- inspired by the way script_output waits for typed in command
- helps with reading results and also in my case it dramatically improve reliability of assert_script_run call

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -127,7 +127,10 @@ sub script_run {
     if ($wait > 0) {
         my $str = testapi::hashed_string("SR$cmd$wait");
         if (testapi::is_serial_terminal) {
-            testapi::type_string " ; echo $str-\$?-\n";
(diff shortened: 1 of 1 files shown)


aaannz:
I don't know what's wrong with travis, but job based on mine branch passed: https://travis-ci.org/aaannz/os-autoinst/builds/389633151

okurz:
Looks like a good idea to me. travis checks are passed fine now.

@aaannz @richiejp I wonder if this could be put into proper tests now. t/03-testapi.t does not use the serial terminal emulation but t/10-terminal.t does. IIUC it only does not yet test any testapi call using the serial terminal. Would it work to explore this area now?

aaannz:
@okurz I'll look into writing a test for this. But I'm out of time currently and I cannot give any ETA.

PR #972, opened 2018-06-12
Install OpenQA/Commands.pm

Martchus:
That's the reason why the ws server isn't responding on e212.


Martchus:
Btw, no need to update the package because it includes the OpenQA directory already.

On openqaworker12 with my already updated package:
```
[martchus@openqaworker12 ~]$ rpm -ql os-autoinst | grep -i command
/usr/lib/os-autoinst/OpenQA/Commands.pm
/usr/lib/os-autoinst/commands.pm
```

PR #973, opened 2018-06-12
Make ssh failures to remote systems debuggable

coolo:
In case ssh fails, the xterm just closes and leaves the reviewer
clueless. So adding a blocking 'read' call afterwards make sure
the xterm stays open with the ssh error.

https://progress.opensuse.org/issues/34003

### DIFF
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -60,7 +60,7 @@ sub sshCommand {
         $sshopts = "-X $sshopts";
     }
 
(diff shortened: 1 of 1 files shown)


okurz:
Perfect, just what the doctor ordered ;)

Now, does it work?

coolo:
It's live since may 11th - I just created a PR for it now :)

okurz:
May 11 already, time flies

PR #976, opened 2018-06-14
Add extra information to the status command in isotovideo

foursixnine:


isotovideo
```diff
@@ -519,7 +519,8 @@ sub process_command {
 
     # handle HTTP commands
     if ($rsp->{cmd} eq 'status') {
-        my $result = {tags => $tags, running => $current_test_name};
+        my $result
+          = {tags => $tags, running => $current_test_name, test_execution_paused => $stop_passing_commands_to_backend, pause_test_name => $pause_test_name};
```
Martchus: It would be nice if you would add `$current_test_full_name` as well. This one includes the prefix, eg. 'installation-welcome' instead of just 'welcome'.

coolo: Please use this route only for initial status, so I guess you should put updates to test_execution_paused too

foursixnine: @coolo What do you mean by initial status? Since it's served via websocket it can be used to get current state information. But If I get what you mean correctly:

I could move it to a separate command and subscribe to that one over websockets, and leave status command as is.

coolo: So my arguments aren't against this change but pro having a message about  test_execution_paused towards the client if the tab is already open.

foursixnine: Ah I get it now! Indeed It's gonna be through the [ws proxy](https://github.com/foursixnine/openQA/commit/b24eab7728deb1333661cccc8325cf53a4f2f30d#diff-d607a0943ac197c97f5def97f636753aR309)

Martchus: @coolo Of course. Getting the initial state is the use-case for this.

Further updates are already emitted anyways - at least in my web socket proxy route for the developer mode. This is also already tested. The web socket proxy route which is only for status updates should do the same of course.

Martchus: Btw, both web socket proxy routes should behave the same to simplify the code (on both, server and client). The only difference is that the status-only route doesn't open a developer session and doesn't allow to pass commands to os-autoinst.

Martchus:
I'll check this out locally and make the UI using it.

Martchus:
It works as expected, but as I said the full test name should be there, too. So maybe you like to cherry-pick and squash this commit: https://github.com/Martchus/os-autoinst/commit/c4d908ba5cc96f61d80295692a402b35d4c273a6

PR #977, opened 2018-06-15
[WIP] Add label message to serial failures

jknphy:
Add label message to serial failures:
Calling subroutine with:
`$testapi::distri->set_expected_serial_failures(soft=>{"AWESOME SOFT MSG 1"=>[qr/gcc version/], "AWESOME SOFT MSG 2"=>[qr/insmod error: 1/]}, hard=>{"AWESOME HARD MSG"=>[qr/No iBFT detected/]});`
local corresponding verification run is: http://dhcp254.suse.cz/tests/1694#step/welcome/12


dasantiago:
If there were tests using the previous implementation, then they need to be fixed, otherwise this will break their capture.

jknphy:
I didnt' find any test in #932.

dasantiago:
I meant tests from os-autoinst-distri :-)

jknphy:
I think you're refering to this repo for the tests, https://github.com/os-autoinst/os-autoinst-distri-opensuse. This features was not used yet (probably because nobody knew the correct name of the function to be called that we discovered).

dasantiago:
The change was described in the PR comments  ;-)

I see in the test you made that it's showing the same soft exception twice. 
So I have a suggestion: Do not show consecutive screens with exceptions that are the same.

basetest.pm
```diff
@@ -645,19 +645,22 @@ sub parse_serial_output_qemu {
...
-                if ($line =~ /$regexp/) {
-                    my $fail_type = 'softfail';
-                    if ($type eq 'hard') {
-                        $die       = 1;
-                        $fail_type = 'fail';
+            for my $regexp_table ($failures->{$type}) {
+                for my $msg (keys %$regexp_table) {
+                    for my $regexp (@{$regexp_table->{$msg}}) {
+                        # If you want to match a simple string please be sure that you create it with quotemeta
+                        if ($line =~ /$regexp/) {
```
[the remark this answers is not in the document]
rwx788: I don't really get how we ended up with O(n^3) here (considering first loop is a constant). What's about changing structure, I also think it will be enough to have one regexp per message, as for OR scenarios regexp can be easily written. Reason to have array in previous implementation is to be able to match multiple pattern for different use-cases.
So I suggest following structure:
```perl
my %failures = (
    soft => [
            { message => '1', pattern => qr/Pattern1/  },
            { message => '2', pattern => qr/Pattern2/  },
    ],
    hard => [
            { message => '3', pattern => [qr/Pattern3/] },
    ],
);
```
Then we can iterate like:
```perl
for my $type (qw(soft hard)) {
    for my $regexp_table (@{$failures{$type}}) {
        print $regexp_table->{message} . " " . $regexp_table->{pattern} . "\n";
    }
}
```
Or using same approach with message as a key and pattern as value and having something like:
```perl
while(my($message, $pattern) = each %hash) { ... }
```

WDYT?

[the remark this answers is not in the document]
dasantiago: @rwx788 That's a good idea. But, imho, if we want to make the code clearer i think we should use the regular expression as key. Then we could loop the file with something like:

`
while(<>)  {
 my $match = $hash{$_};
 print "found $match->{$message} with $match->{type}\n" if $match;
}
`

If i'm correct this supports all the possible cases without the need to change anything in the future, if we want to extend.

Although this would bring a new cpan dep (Tie::RegexpHash). :-(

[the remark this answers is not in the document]
rwx788: Yeah, main thing which we got on top is array or regexps per message, but it will be enough to have 1 to 1, whether message or regexp is the key. Having same regexp twice with different messages and behavior is indeed wrong. 4 nested for loops looks like overkill. On top we can set type of failure as a property of failure, having just array of hashes in the end. It only depends if we plan to extend it even further. But I like your proposal, for now it's perfect (except extra dependency, but that's minor).
E.g.
```perl
my %failures = (
    [
            { type => soft, message => '1', pattern => qr/Pattern1/ },
            { type => soft, message => '2', pattern => qr/Pattern2/ },
            { type => hard, message => '3', pattern => qr/Pattern3/ },
    ]
);
```

jknphy:
I made work the first approach mentioned by @rwx788 changing the internal structure and removing some loops. Besides I added a hash for not repeating matches when recording the results as suggested by @dasantiago.
Verification run: http://dhcp254.suse.cz/tests/1723#

basetest.pm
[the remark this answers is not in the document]
jknphy: I made it work with the first approach mentioned by @rwx788 changing the internal structure and removing some loops. Besides I added a hash for not repeating matches when recording the results as suggested by @dasantiago.
Verification run: http://dhcp254.suse.cz/tests/1723#

basetest.pm
```diff
@@ -640,24 +640,26 @@ sub parse_serial_output_qemu {
...
         chomp $line;
         # only two keys allowed
         for my $type (qw(soft hard)) {
-            for my $regexp (@{$failures->{$type}}) {
+            for my $regexp_table (@{$failures->{$type}}) {
                 # If you want to match a simple string please be sure that you create it with quotemeta
-                if ($line =~ /$regexp/) {
+                my $regexp = $regexp_table->{pattern};
+                if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
+                    $regexp_matched{$regexp} = 1;
```
rwx788: I guess this mechanism is required for soft failures only. @dasantiago is it the case?. And advantage of having regexp as a key is that all of them will be unique. So, please, consider using it as a key and put $type as a part of hash. Then we can iterate over keys and values of hash in single loop like `while(my($k, $v) = each %h)`, but that's optional.

distribution.pm
```diff
@@ -265,7 +265,13 @@ Define the patterns to look for in the serial console.
 The patterns can be either I<hard> or I<soft>.
 
 Example:
-    set_expected_serial_failures(soft=>[qr/Pattern1/], hard=>[qr/Pattern2/]);
+    set_expected_serial_failures(soft => [
```
rwx788: I guess it would be beneficial to have test for this, as it will show how to use it. Not sure if we can assert soft-failures though. See https://github.com/os-autoinst/os-autoinst/blob/master/t/99-full-stack.t

basetest.pm
dasantiago: @rwx788 
My suggestion was for **consecutive** exceptions that are the same, independently of soft / hard.

IMHO, having 30 consecutive exceptions that are the same, only creates noise.

This doesn't address the consecutive part ;-)

basetest.pm
```diff
@@ -640,25 +640,26 @@ sub parse_serial_output_qemu {
...
+            if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
+                $regexp_matched{$regexp} = 1;
+                my $fail_type = 'softfail';
+                if ($type eq 'hard') {
+                    $die       = 1;
+                    $fail_type = 'fail';
                 }
+
+                $self->record_testresult($fail_type);
+                $self->record_resultfile('Serial Failure', $regexp_table->{message} . " - Serial error: $line", result => $fail_type);
```
rwx788: We need to check if message key is defined, as if not, we'll get an error as try to concatenate uninitialized value.

jknphy:
VR with last changes required: http://dhcp254.suse.cz/tests/1739#step/welcome/12

basetest.pm
```diff
@@ -640,25 +640,29 @@ sub parse_serial_output_qemu {
...
-
+        for my $regexp_table (@{$failures}) {
+            my $regexp  = $regexp_table->{pattern};
+            my $message = $regexp_table->{message};
+            my $type    = $regexp_table->{type};
+            die "Wrong type defined for serial failure. Only soft or hard allowed." if $type !~ /soft|hard/;
+            die "Not defined message key. Please, define one for each regular expression" if !defined $message;
+
+            # If you want to match a simple string please be sure that you create it with quotemeta
+            if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
```
dasantiago: If you want to avoid consecutive regular exceptions then you need to save the exception position as well.

rwx788: Hmm, good point. Question is if we want to see same soft-failures even in case of multiple matches or just once and then it's possible to check serial output in case more details are required.

distribution.pm
```diff
@@ -262,18 +262,21 @@ sub script_output {
     set_expected_serial_failures(%failures)
```
okurz: please also update this one here, you changed the variable format

jknphy:
Might  this be merged? I guess we can find more advanced use case in future and revisit it. The current code enhance the existing functionality/feature, which was not eve used, so using it there is opportunity to improve it later.

rwx788:
@foursixnine @coolo whenever you have time, could you please take a look.

PR #978, opened 2018-06-20
Record info for barriers and mutexes

kravciak:
Local run: http://dhcp165.suse.cz/tests/6056#step/stack_bootstrap/2

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -52,6 +52,25 @@ sub _lock_action {
     return _try_lock('mutex', $name, $param);
 }
 
+# Log info about event and it's location
+sub _log {
+    my ($name, %args) = @_;
+
+    # Generate log message
+    my $job = $args{where} ? get_job_info($args{where})->{settings}->{TEST} . " #$args{where}" : 'parent job';
+    my $msg = "Wait for $name (on $job)";
+    $msg .= " - $args{info}" if $args{info};
+
+    if (defined $args{amend}) {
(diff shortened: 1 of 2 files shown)


coolo:
perhaps rebase to fix the tests?

coolo:
looks like your test fails

lockapi.pm
```diff
@@ -52,6 +52,25 @@ sub _lock_action {
...
+    my ($name, %args) = @_;
+
+    # Generate log message
+    my $job = $args{where} ? get_job_info($args{where})->{settings}->{TEST} . " #$args{where}" : 'parent job';
+    my $msg = "Wait for $name (on $job)";
+    $msg .= " - $args{info}" if $args{info};
+
+    if (defined $args{amend}) {
+        # amend log info with wait duration
+        $autotest::current_test->remove_last_result;
```
coolo: Can't call method "record_resultfile" on an undefined value at ../testapi.pm line 213.

kravciak: Right, I think it is missing $autotest::current_test. I will look at it during the week.
```perl
ok(barrier_wait('barrier1'), 'registered for waiting and released immideately');
  barrier_wait: _log $name, where => $where;
  _log: record_info 'Paused '
  record_info: $autotest::current_test->record_resultfile($title, $output, %nargs);
```

kravciak: Fixed

PR #979, opened 2018-06-21
Add basic spvm backend

asdil12:


okurz:
LGTM. If the test can at least do *something* with it I would even merge it and extend later.

asdil12:
@okurz so let's do it then.

backend/spvm.pm
```diff
@@ -0,0 +1,113 @@
+# Copyright © 2009-2013 Bernhard M. Wiedemann
+# Copyright © 2012-2016 SUSE LLC
```
Martchus: Isn't this a new file?

asdil12: yes, it is

asdil12: updated (C)

backend/spvm.pm
```diff
@@ -0,0 +1,112 @@
...
+use testapi qw(get_var get_required_var check_var);
+
+use IO::Select;
+
+# supporting the minimal command set of NovaLink through a ssh tunnel
+
+sub new {
+    my $class = shift;
+    my $self  = $class->SUPER::new;
+    get_required_var('WORKER_HOSTNAME');
```
ldevulder: This variable is not used... Should not be `NOVALINK_HOSTNAME` instead?

backend/spvm.pm
```diff
@@ -0,0 +1,112 @@
...
+        'novalink-ssh',
+        'ssh-xterm',
+        {
+            hostname => get_required_var('NOVALINK_HOSTNAME'),
+            password => get_required_var('NOVALINK_PASSWORD'),
+            username => get_var('NOVALINK_USERNAME', 'root')});
+    $ssh->backend($self);
+
+    # remove backend.crashed
+    $self->unlink_crash_file;
```
ldevulder: Test on my lab failed with unfound `unlink_crash_file`, I searched it but was not able found it...

backend/spvm.pm
coolo: it is used by testapi

backend/spvm.pm
coolo: that function is dead by now - the commits have been lying around for a while

asdil12: So shall I remove that line?

coolo: yes, see #930

backend/spvm.pm
```diff
@@ -0,0 +1,110 @@
...
+sub stop_serial_grab {
+    my ($self) = @_;
+
+    $self->stop_ssh_serial;
+    return;
+}
+
+1;
+
+# vim: set sw=4 et:
```
okurz: I think we have this line removed from all files already, please delete

backend/spvm.pm
```diff
@@ -0,0 +1,108 @@
...
+use IO::Select;
+
+# supporting the minimal command set of NovaLink through a ssh tunnel
+
+sub new {
+    my $class = shift;
+    my $self  = $class->SUPER::new;
+    get_required_var('WORKER_HOSTNAME');
+
+    return $self;
```
ldevulder: No so important, but it could be simplify like this: `return $class->SUPER::new;` (seen in others backend).

okurz:
LGTM

backend/spvm.pm
foursixnine: It would end up being the same.

foursixnine:
Aside from the package not building, it seems that I forgot to check for  backend/spvm.pm being added into the makefile @asdil12 can you take a look?

asdil12:
@foursixnine --> https://github.com/os-autoinst/os-autoinst/pull/989

PR #980, opened 2018-06-25
Revert of commit 0b557ea

asmorodskyi:
Waiting for whole command additionally to marker introduce some regressions
in case of complex multiline commands. Also from commit 0b557ea it was not clear
what benefits we getting when waiting for more complex things

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -127,10 +127,7 @@ sub script_run {
     if ($wait > 0) {
         my $str = testapi::hashed_string("SR$cmd$wait");
         if (testapi::is_serial_terminal) {
-            my $marker = " ; echo $str-\$?-";
-            testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, undef, 0, no_regex => 1);
-            testapi::type_string("\n");
+            testapi::type_string " ; echo $str-\$?-\n";
         }
         else {
             testapi::type_string " ; echo $str-\$?- > /dev/$testapi::serialdev\n";


asmorodskyi:
https://openqa.suse.de/tests/1772606#step/setup/11 - example of failed test. If you open same test several builds before you will not see this failure

foursixnine:
If you read the commit message, you get a briefing on what are the benefits. But in deed there's something odd here

aaannz:
I assume shell added '>' characters when issuing multiline commands. I don't know yet how to properly fix it, but given the test module didn't failed I would like you not to revert it. I had serious troubles with `script_run` before and none after this change. I don't know if there was some race or test machine too slow or what, but after command was sent to SUT, `select` call returned there are some data, which was the command typed, and then consistently returned 0 as no new data present and the actual result of command was not received on os-autoinst side. This never happened when using script_output which always read back what was typed. Thus my change.

asmorodskyi:
@aaannz , you mention in commit message that you doing like in script_output but it is not true actually because script_output is waiting for command and marker in two separate wait_serials while you doing this in one. In worst case scenario I suggesting to try out to do it exactly like script_output doing. WDYT ?

aaannz:
@asmorodskyi Are you sure? There already was one wait_serial call on [line 138](https://github.com/os-autoinst/os-autoinst/commit/0b557ea528fbf1cc3ce6143331f929299ece5ef1#diff-fca689375b4b83edbf1f8a673fab105fR138). The wair_serial I added waits for typed in command before sending '\n' to execute it.
What I see to fix this is either do it correctly - postprocess inserted command to remove '\' and replace '\n' -> '\n>' before waiting. Or ugly hack limit wait time to some minimal defined amount (let's say 5s to handle slow machines and long commands) and signal wait_serial this can fail and not to report it.

aaannz:
Aha, I know what you mean. I misread your question. Well, sure, I have nothing against using two wait_serial calls, but I don't think it will solve this particular issue with multiline commands. The typed command will be read back with shell marks anyway IMO.

asmorodskyi:
after some discussion with @richiejp  decide to fix this on test side simply by not passing something which can cause such behavior.

PR #983, opened 2018-06-26
amt: initial support

marmarek:
Initial version of amt backend. Right now it's capable of power control,
and using VNC. There is (commented out) code for setting up VNC
redirection - needed once for particular machine.
There is also incomplete code for IDE redirection, but tests shows that
(at least Linux implementation) is very unreliable. So, right now there
is no support for ISO or HDD_* variables.
Serial console isn't fully functional - amtterm command line is correct,
but right now is connected as 'sol' console - worth considering
switching it to serial console (the one writing to 'serial0' file). We
have only one such console, so this needs more evaluation which one is
more useful.

New backend-specific variables:
  - AMT_HOSTNAME - hostname/IP of target host
  - AMT_PASSWORD - password for 'admin' AMT user on target host

Dependencies:
 - wsmancli
 - amtterm

Discussion here: https://progress.opensuse.org/issues/37853

### DIFF
--- a/backend/amt.pm
+++ b/backend/amt.pm
@@ -1,0 +1,268 @@
+# Copyright © 2018 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


backend/amt.pm
```diff
@@ -0,0 +1,263 @@
+# Copyright © 2018 Marek Marczykowski-Górecki
```
okurz: Please make that SUSE LLC if you don't mind :)

marmarek: Done.

coolo:
I will merge it as is - as it might be a good starting point for people interested in this. It's not added to the Makefile though

PR #986, opened 2018-06-28
Add a note for script_output method when is not using dedicated serial

rwx788:
script_output is designed to work with dedicated serial console. In case
it is polluted with anything else, content will be returned as a command
execution result.

See [poo#30613](https://progress.opensuse.org/issues/30613).


okurz:
good, thx

PR #987, opened 2018-06-29
Add tests for web socket route of command server

Martchus:


Martchus:
The coverage is unaffected. So we already test the web socket server somewhere else?

coolo:
I doubt it, I would rather assume that the extra report of command process is not working

mudler:
you have to make the process call Devel::Report if it's a different process hitting that codepath. Or create unit tests directly targetting OpenQA::Commands

Martchus:
@mudler Yes, but I thought the already present line `Devel::Cover::report() if Devel::Cover->can('report');` in `commands::start_server()` would do it. But apparently there's something missing.

foursixnine:
@mudler Do you have any clue on why the code is shown as partially covered?. (Specially OpenQA::Commands, it does not show anywhere)

PR #988, opened 2018-07-06
Introduce check_shutdown for testapi

OleksandrOrlov:
The check_shutdown function is added to have ability to decide on further actions in case of shutdown failure, instead of marking a test as failed immediately.

In tests it might be used to make soft failure checks, for example.

Related ticket: [poo#35215](https://progress.opensuse.org/issues/35215)

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -9,6 +9,7 @@ use Test::Fatal;
 use Test::Mock::Time;
 use Test::Warnings;
 use File::Temp;
+use Test::Exception;
 
 BEGIN {
     unshift @INC, '..';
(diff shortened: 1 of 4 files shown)


testapi.pm
```diff
@@ -1524,18 +1524,17 @@ sub power {
...
 Periodically check backend for status until C<'shutdown'>. Does I<not> initiate shutdown.
-Default timeout is 60s
 
-Returns C<undef> on success, throws exception on timeout.
+Returns true on success and false if C<$timeout> timeout is hit. Default timeout is 60s.
 
 =cut
 
-sub assert_shutdown {
+sub check_shutdown {
```
foursixnine: You will need to use Test::MockModule, mock the query_isotovideo method from autotest to return the expected values. An example on how to do that can be seen in the [testapi tests](https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L157)

testapi.pm
```diff
@@ -1524,18 +1524,17 @@ sub power {
...
+=head2 check_shutdown
 
-  assert_shutdown([$timeout]);
+  check_shutdown([$timeout]);
 
 Periodically check backend for status until C<'shutdown'>. Does I<not> initiate shutdown.
-Default timeout is 60s
 
-Returns C<undef> on success, throws exception on timeout.
+Returns true on success and false if C<$timeout> timeout is hit. Default timeout is 60s.
```
foursixnine: An update to the pod wouldn't hurt :)

Martchus:
Looks good now that you've added tests.

This changes the meaning of `assert_shutdown` which is a breaking API change. So I suppose `$INTERFACE` should be incremented (https://github.com/os-autoinst/os-autoinst/blob/master/isotovideo#L66)?

t/03-testapi.t
```diff
@@ -291,6 +292,30 @@ subtest 'set console tty and other args' => sub {
...
+    # Test cases, when shutdown is finished before timeout is hit
+    $mod->mock(
+        read_json => sub {
+            return {ret => 1};
+        });
+    ok(check_shutdown, 'check_shutdown should return "true" if shutdown finished before timeout is hit');
+    is(assert_shutdown, undef, 'assert_shutdown should return "undef" if shutdown finished before timeout is hit');
+    $mod->mock(
+        read_json => sub {
+            return {ret => -1};
```
foursixnine: why -1 specifically?

OleksandrOrlov: It is because I just wanted to cover the following part in `check_shutdown` function:
```
if ($is_shutdown < 0) {
    bmwqemu::diag("Backend does not implement is_shutdown - just sleeping");
    sleep($timeout);
}
```

t/03-testapi.t
```diff
@@ -9,6 +9,7 @@ use Test::Fatal;
 use Test::Mock::Time;
 use Test::Warnings;
 use File::Temp;
+use Test::Exception;
```
foursixnine: Please add this dependency to the cpanfile.

Martchus:
@foursixnine If the things you've mentioned are fixed now, we should merge this.

PR #992, opened 2018-07-18
Add current_console method

foursixnine:
With changes introduced in c670720 testapi no longer holds what is the
current console, leading to problems when test distributions rely on
this.

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -318,7 +318,7 @@ sub run_daemon {
             }
 
             # Process input chunk
-            say "BYTES $bytes";
+            app->log->debug("BYTES $bytes");
         });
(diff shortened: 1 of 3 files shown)


richiejp:
You could also refactor `is_serial_terminal`, `console` etc. to use the new function :-p

foursixnine:
@richiejp good point!

foursixnine:
I wonder why codecov is complaining that there are no lines being covered...

mudler:
We need to look at it, but it should boil down to Devel::Cover that must be called from each involved process :(

foursixnine:
\o/ Coverage slowly going up!

PR #993, opened 2018-07-18
Allow to pause on assert screen timeout

Martchus:
https://progress.opensuse.org/issues/38510

---

This is very WIP. I'm just trying out things.

My current approach is to call a certain isotovideo command from autotest on timeout:

* It will block in case the test should be paused. In this case the regular assertion handling in autotest is aborted. Instead the assertion is repeated after resuming from the blocking isotovideo command.
* It will not block in case the test shouldn't be paused. In this case autotest continues with the assertion as usual.

---

PR for updating the UI accordingly: https://github.com/os-autoinst/openQA/pull/1727

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -381,7 +381,7 @@ sub handle_paused_test_execution {
     my ($io_handle, $rsp) = @_;
     my $cmd = $rsp->{cmd};
 
-    # write debugging info
+    # emit info
     myjsonrpc::send_json($cmd_srv_fd, {paused => $rsp});
     print("isotovideo: paused, so not passing $cmd to backend\n");
 
@@ -409,6 +409,23 @@ sub process_command {
         myjsonrpc::send_json($backend_process->channel_in, {cmd => $cmd, arguments => $rsp});
         return;
     }
+    if ($rsp->{cmd} eq 'report_timeout') {
+        if (!$pause_on_assert_screen_timeout) {
+            myjsonrpc::send_json($io_handle, {ret => 0});
+            return;
+        }
+
+        $stop_passing_commands_to_backend = 1;
+
+        # emit info
+        myjsonrpc::send_json($cmd_srv_fd, {paused => $rsp, reason => 'timeout'});
+        print("isotovideo: pausing test execution at $current_test_full_name because a timeout occured\n");
+
+        # postpone sending the reply
+        $postponed_io_handle = $io_handle;
+        $postponed_command   = undef;
+        return;
+    }
     if ($rsp->{cmd} eq 'set_pause_at_test') {
         $pause_test_name = $rsp->{name};
 
@@ -422,7 +439,7 @@ sub process_command {
     if ($rsp->{cmd} eq 'set_pause_on_assert_screen_timeout') {
         $pause_on_assert_screen_timeout = $rsp->{flag};
 
-        # print/send debugging info
+        # send debugging info
         myjsonrpc::send_json($cmd_srv_fd, {set_pause_on_assert_screen_timeout => $pause_on_assert_screen_timeout});
 
         myjsonrpc::send_json($io_handle, {ret => 1});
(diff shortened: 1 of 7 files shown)


Martchus:
Apparently the new unit test fails on Travis:

```
Can't locate tinycv.pm in @INC (you may need to install the tinycv module) (@INC contains: /opt/run/t/blib/arch /opt/run/t/blib/lib .. /usr/lib/perl5/site_perl/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/site_perl/5.18.2 /usr/lib/perl5/vendor_perl/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/vendor_perl/5.18.2 /usr/lib/perl5/5.18.2/x86_64-linux-thread-multi /usr/lib/perl5/5.18.2 /usr/lib/perl5/site_perl .) at /usr/lib/perl5/vendor_perl/5.18.2/Test/MockModule.pm line 26.
```

I can not reproduce this locally, but likely I should mock tinycv in a better way anyways.

Martchus:
Ok, the tests pass now using `PERL5LIB="..:../ppmclibs:../ppmclibs/blib/arch/auto/tinycv:$$PERL5LIB"`. Unfortunately codecov is still stuck with 13.95 %.

Martchus:
Note: If we merge this, we must add `perl(Test::Exception)` to dependencies of os-autoinst package: https://build.opensuse.org/package/rdiff/home:mkittler:branches:devel:openQA/os-autoinst?linkrev=base&rev=3

coolo:
No reason to wait with a submit request now

t/03-testapi.t
```diff
@@ -21,6 +22,16 @@ use Test::MockModule;
...
 my $fake_exit = 0;
 
+# define variables for 'fake_read_json'
+my $report_timeout_called         = 0;
+my $fake_pause_on_timeout         = 0;
+my $fake_needle_found             = 1;
+my $fake_needle_found_after_pause = 0;
+
+# define 'write_with_thumbnail' to fake image
+sub write_with_thumbnail {
```
foursixnine: Just an empty sub?

isotovideo
```diff
@@ -438,6 +441,23 @@ sub process_command {
...
+        if (!$pause_on_assert_screen_timeout) {
+            myjsonrpc::send_json($io_handle, {ret => 0});
+            return;
+        }
+
+        $test_execution_paused = $rsp->{msg};
+
+        # emit info
+        myjsonrpc::send_json($cmd_srv_fd, {paused => $rsp, reason => $test_execution_paused});
+        print("isotovideo: pausing test execution at $current_test_full_name because a timeout occurred\n");
```
foursixnine: You should  use the diag function instead of print whenever possible, so that it ends on the logs properly on the other hand: 
i'd write the message as: 'pausing test execution on timeout as requested at $current_test_full_name'

isotovideo
```diff
@@ -477,14 +509,17 @@ sub process_command {
...
+                current_test_full_name => $current_test_full_name,
+        });
         if ($pause_test_name
             && $current_test_name
             && $current_test_full_name
             && ($pause_test_name eq $current_test_name || $pause_test_name eq $current_test_full_name))
         {
-            print("isotovideo: pausing test execution of $pause_test_name\n");
-            $stop_passing_commands_to_backend = 1;
+            print("isotovideo: pausing test execution of $pause_test_name because we're supposed to pause there\n");
```
foursixnine: Same as above, diag instead of print. On the other hand, I wonder if the reason to pause should be more clear.

isotovideo
Martchus: Ok, I'll change that.

t/03-testapi.t
Martchus: It is just to mock that method (in this method we don't care about writing thumbnails).

testapi.pm
```diff
@@ -237,7 +237,23 @@ sub _check_backend_response {
         return _handle_found_needle($foundneedle, $rsp, $tags);
     }
     elsif ($rsp->{timeout}) {
-        bmwqemu::fctres("match=" . join(',', @$tags) . " timed out after $timeout");
+        my $status_message = "match=" . join(',', @$tags) . " timed out after $timeout";
```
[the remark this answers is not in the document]
Martchus: I'm not changing the behavior of the test API here. At least not for test writers.

Only the interface of the command server is *extended*. But I suppose that is only relevant for openQA's internal developer mode. Should I still increase the interface version considering that?

Martchus:
> I'm just not 100% clear if the status route of the command server will report the type of pause that we have, but I don't think knowing the type is that important for now

It will and the reason is also already displayed in the web UI (https://github.com/os-autoinst/openQA/pull/1727/commits/89ece8a81e1d907d61788cbf78fe4fc5bf31bebc).

testapi.pm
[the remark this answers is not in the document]
foursixnine: Yes :)

PR #994, opened 2018-07-20
Revert "Add label message to serial failures"

foursixnine:
Reverts os-autoinst/os-autoinst#977

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -645,29 +645,25 @@ sub parse_serial_output_qemu {
     my $json = myjsonrpc::read_json($autotest::isotovideo);
 
     my $die = 0;
-    my %regexp_matched;
     # loop line by line
     for my $line (split(/^/, $json->{serial})) {
         chomp $line;
-        for my $regexp_table (@{$failures}) {
-            my $regexp  = $regexp_table->{pattern};
-            my $message = $regexp_table->{message};
-            my $type    = $regexp_table->{type};
-            die "Wrong type defined for serial failure. Only soft or hard allowed." if $type !~ /soft|hard/;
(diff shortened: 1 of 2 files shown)


foursixnine:
This breaks the CI:

`[2018-07-20T15:10:59.0144 CEST] [debug] Not an ARRAY reference at /home/foursixnine/Projects/suse.com/github.com/os-autoinst/os-autoinst/basetest.pm line 652.`

not ok 26 - image of hdd uploaded
#   Failed test 'image of hdd uploaded'
#   at t/full-stack.t line 218.
Use of uninitialized value in subroutine entry at t/full-stack.t line 219 (#1)
    (W uninitialized) An undefined value was used as if it were already
    defined.  It was interpreted as a "" or a 0, but maybe it was a mistake.
    To suppress this warning assign a defined value to your variables.
    
    To help you figure out what was undefined, perl will try to tell you
    the name of the variable (if any) that was undefined.  In some cases
    it cannot do this, so it also tells you what operation you used the
    undefined value in.  Note, however, that perl optimizes your program
    anid the operation displayed in the warning may not necessarily appear
    literally in your program.  For example, "that $foo" is usually
    optimized into "that " . $foo, and the warning will refer to the
    concatenation (.) operator, even though there is no . in
    your program.
    
not ok 27 - exported image has correct permissions (420 -> 0644)
#   Failed test 'exported image has correct permissions (420 -> 0644)'
#   at t/full-stack.t line 220.
#          got: '0'
#     expected: '420'

foursixnine:
@rwx788 @okurz fyi

okurz:
@jknphy

PR #996, opened 2018-07-25
Avoid rolling back to undefined console

richiejp:
We can only roll back to the last selected console if the user selected one
before the snapshot. If they didn't then we have to rely on the backend's
default console.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -622,9 +622,11 @@ sub rollback_activated_consoles {
         autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
     }
 
-    my $ret = autotest::query_isotovideo('backend_select_console',
-        {testapi_console => $autotest::last_milestone_console});
-    die $ret->{error} if $ret->{error};
+    if (defined($autotest::last_milestone_console)) {
+        my $ret = autotest::query_isotovideo('backend_select_console',
+            {testapi_console => $autotest::last_milestone_console});
+        die $ret->{error} if $ret->{error};
+    }
 
     return;
 }
--- a/autotest.pm
+++ b/autotest.pm
@@ -18,7 +18,7 @@ package autotest;
 use strict;
 use bmwqemu;
 use Exporter 'import';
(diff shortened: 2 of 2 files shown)


basetest.pm
```diff
@@ -622,9 +622,11 @@ sub rollback_activated_consoles {
         autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
     }
 
-    my $ret = autotest::query_isotovideo('backend_select_console',
-        {testapi_console => $autotest::last_milestone_console});
-    die $ret->{error} if $ret->{error};
+    if (defined($autotest::last_milestone_console)) {
```
foursixnine: This makes sense to me, but @coolo is usually good at having hunches :)
@richiejp did you manage to get a job passing with this fix? Just to be on the safe side :)

foursixnine:
@okurz: https://progress.opensuse.org/issues/38813#note-12

basetest.pm
richiejp: This job was crashing after failing, but now it reverts the image successfully... then ends. It is king of dump, but it proves this change works: http://10.160.67.78/tests/561

richiejp:
@okurz Your assuming I haven't tested it because you can't see the testing and what is more you are suggesting that more effort would solve this, which implies the problem is a lack of effort.

In reality I actually spent longer testing than figuring out what is wrong and writing the solution. Obviously more testing would be better, especially when changing default parameters in the API...

okurz:
I was not assuming that you did not test but I did not know what you test so I struggle to review and find out what potentially was missed. And I know that there are people that also provide PRs which are just based on best guesses and are merely a proposal but have never been tested. Happened to me as well … ;)

basetest.pm
foursixnine: @richiejp but we still would need to set up a default console at a test distribution level, right?

richiejp:
> I was not assuming that you did not test but I did not know what you test so I struggle to review and find out what potentially was missed

Yes, the problem is you used too many words :-p and I should have said what testing I had done.

PR #997, opened 2018-07-26
Use bootindex with non-uefi systems

richiejp:
Even non-uefi bios's support the bootindex device parameter now, which is more
flexible than the old '-boot order=<drive>' command line option. In particular
it supports setting USB devices as the first boot drive which eliminates the
need for selecting them in the bios boot menu (which is unreliable as the
order can change).

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -142,7 +142,6 @@ sub configure_blockdevs {
     my ($self, $bootfrom, $basedir, $vars) = @_;
     my $bdc       = $self->blockdev_conf;
     my @scsi_ctrs = $self->controller_conf->get_controllers(qr/scsi/);
-    my $uefi      = $vars->{UEFI};
 
     $bdc->basedir($basedir);
 
@@ -163,7 +162,7 @@ sub configure_blockdevs {
             $drive = $bdc->add_new_drive($node_id, $hdd_model, $size);
         }
 
-        if ($i == 1 && $uefi && $bootfrom eq 'disk') {
+        if ($i == 1 && $bootfrom eq 'disk') {
             $drive->bootindex(0);
         }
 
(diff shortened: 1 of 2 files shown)


OpenQA/Qemu/Proc.pm
```diff
@@ -163,7 +163,7 @@ sub configure_blockdevs {
             $drive = $bdc->add_new_drive($node_id, $hdd_model, $size);
         }
 
-        if ($i == 1 && $uefi && $bootfrom eq 'disk') {
+        if ($i == 1 && $bootfrom eq 'disk') {
```
foursixnine: We don't care about UEFI anymore? Hmm?

richiejp: It no longer requires UEFI, but will be still used with it, unless I have got the logic wrong?

richiejp:
multipath works: http://10.160.67.78/tests/571

Tried to test it with chained jobs and pflash, but it ran into a product bug (kernel soft lockup):
http://10.160.67.78/tests/565

OpenQA/Qemu/Proc.pm
foursixnine: Doesn't look like it's used elsewhere, at least within this method

richiejp: ah, removed.

foursixnine: :)

foursixnine:
Chained used to be a big annoyance when it came to bootindex...

richiejp:
I restarted the test with MAKETESTSNAPSHOTS, so I can hopefully run it until it passes.

http://10.160.67.78/tests/576

richiejp:
OK, it managed to boot a chained job: http://10.160.67.78/tests/589

foursixnine:
@richiejp at least from what I see, PFLASH variable has to be full path?

richiejp:
Yep and for reference this is the line which should make it work: https://github.com/os-autoinst/openQA/blob/86bc960627436824b432eae4e2198dfb03329e96/lib/OpenQA/Utils.pm#L520

richiejp:
@foursixnine I think I found the reason why the PFLASH variable doesn't behave like a HDD: https://github.com/os-autoinst/openQA/blob/34630305090c10c31949a6704ecf658a678c8a8b/lib/OpenQA/Worker/Engines/isotovideo.pm#L98

foursixnine:
@richiejp as long as the code looks like this
```
$res{UEFI_PFLASH_VARS} = 'hdd' if $vars->{UEFI_PFLASH_VARS};
```
The worker [cache code](
https://github.com/os-autoinst/openQA/blob/34630305090c10c31949a6704ecf658a678c8a8b/lib/OpenQA/Worker/Cache.pm#L104) will parse it properly and properly cache the file.

I fear there's still a bit more to this...

foursixnine:
lets see how it goes with  https://github.com/os-autoinst/openQA/pull/1728

PR #998, opened 2018-07-27
Add messages to serial failures

rwx788:
Resurrection of #977 with the fix with proper initialization + unit tests for parse_serial_output_qemu.

[Verification run](http://g226.suse.de/tests/2079#step/welcome/8).

See [poo#36448](https://progress.opensuse.org/issues/36448).

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -50,7 +50,7 @@ sub new {
     $self->{timeoutcounter}         = 0;
     $self->{activated_consoles}     = [];
     $self->{name}                   = $class;
-    $self->{serial_failures}        = {};
+    $self->{serial_failures}        = [];
     return bless $self, $class;
 }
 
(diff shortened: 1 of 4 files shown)


t/17-basetest.t
```diff
@@ -11,19 +12,70 @@ BEGIN {
...
+        {type => 'soft', message => 'DoNotMatch', pattern => qr/DoNotMatch/},
+    ];
+    $basetest->parse_serial_output_qemu(0);
+    is($basetest->{result}, undef, 'test result untouched without match');
+    is($message,            undef, 'test details do not have extra message');
+
+    $basetest->{serial_failures} = [
+        {type => 'soft', message => 'SimplePattern', pattern => qr/Serial/},
+    ];
+    # push @$expected_details, {title => 'Serial Failure', result => 'softfail', text => ''};
```
okurz: Disabled code?

rwx788:
tinycv compilation fails. @foursixnine could you please give me a hint?

Martchus:
@rwx788 Note that the messages after the tinycv `g++` line are just warnings which are also printed on successful test runs. So tinycv doesn't fail to compile here. The same counts for the message about `Module::Pluggable`. Unfortunately I have no clue what's going wrong so far except that it is something in target `check-local`.

BTW: We can not easily fix the source of the distracting C++ warnings because the relevant code is in a header provided by Perl itself. Maybe just disable them by adding `-Wno-literal-suffix` to compile flags?

rwx788:
@Martchus thanks for the comment, I guess I've found root cause of the failure.

Martchus:
And what was it? (Since you amended your existing commits it is hard to figure from the diff.)

rwx788:
I had quotes for hash keys entry. Error was visible when I executed check-local on my side.

basetest.pm
```diff
@@ -638,34 +638,44 @@ sub search_for_expected_serial_failures {
...
-
-                    $self->record_testresult($fail_type);
-                    $self->record_resultfile('Serial Failure', "Serial error: $line", result => $fail_type);
-                    $self->{result} = $fail_type;
-
+        for my $regexp_table (@{$failures}) {
+            my $regexp  = $regexp_table->{pattern};
+            my $message = $regexp_table->{message};
+            my $type    = $regexp_table->{type};
+            die "Wrong type defined for serial failure. Only soft or hard allowed." if $type !~ /soft|hard/;
```
foursixnine: Please add the type in the message in both cases of the `die`  call

rwx788: Fixed, thanks!

PR #999, opened 2018-07-30
Create dummy pflash vars file to avoid asset caching failure

richiejp:
Temporary hack until better asset handling can be implemented.

https://progress.opensuse.org/issues/38813#note-24

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -628,11 +628,19 @@ if (!$return_code && $completed && $bmwqemu::vars{BACKEND} eq 'qemu') {
         my $format = $1;
         push @toextract, {hdd_num => $i, name => $name, dir => $dir, format => $format};
     }
-    if ($bmwqemu::vars{UEFI} && $bmwqemu::vars{PUBLISH_PFLASH_VARS}) {
-        push(@toextract, {pflash_vars => 1,
-                name   => $bmwqemu::vars{PUBLISH_PFLASH_VARS},
-                dir    => 'assets_public',
-                format => 'qcow2'});
+    if ($bmwqemu::vars{PUBLISH_PFLASH_VARS}) {
+        if ($bmwqemu::vars{UEFI}) {
+            push(@toextract, {pflash_vars => 1,
+                    name   => $bmwqemu::vars{PUBLISH_PFLASH_VARS},
+                    dir    => 'assets_public',
+                    format => 'qcow2'});
+        }
+        else {
+            # Temporary workaround for poo#38813 note 24
+            mkpath('assets_public');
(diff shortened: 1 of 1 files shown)


isotovideo
```diff
@@ -628,11 +628,19 @@ if (!$return_code && $completed && $bmwqemu::vars{BACKEND} eq 'qemu') {
...
-                format => 'qcow2'});
+    if ($bmwqemu::vars{PUBLISH_PFLASH_VARS}) {
+        if ($bmwqemu::vars{UEFI}) {
+            push(@toextract, {pflash_vars => 1,
+                    name   => $bmwqemu::vars{PUBLISH_PFLASH_VARS},
+                    dir    => 'assets_public',
+                    format => 'qcow2'});
+        }
+        else {
+            # Temporary workaround for poo#38813 note 24
```
foursixnine: I would add this also as content to that file... I just wonder if there's a better workarround for this.

richiejp: I guess a basic script could read the webapi And create any 'missing' pflash assets or create placeholders.... I think that would be problematic though.

richiejp:
So far this appears to work:
http://10.160.67.78/tests/609#settings
http://10.160.67.78/tests/607

isotovideo
foursixnine: Nee, just print it to the file :) But it's just a random idea...

mudler: A script that reads webapi and recreate assets ... that would mean basically re-implementing cache again ( so at this point, better make it a service, no? ) :)

mudler:
This special case eventually needs to be abstracted or we keep stacking workarounds.. but if this temp. solves the issue i'm fine with it. If the deploy doesn't goes as expected...  i guess we will have a fun week :)

isotovideo
coolo: Actually I think this solution of always publishing a pflash scales much better than all what you propose as proper.

mudler: i really don't see how specific bits to overcome a specific situation can actually scale

coolo: KISS

richiejp:
I'm already having a fun week :-D

isotovideo
richiejp: I think a service which handles all asset caching and uploads etc. could be simpler in the long run. Because it will be more DRY and there will a be a single source of truth. However I don't think it could be written and tested in the next few hours because it will requiring ripping out a lot of code to do it properly.

mudler: ...i don't think it's really appropriate to *KISS* collegues on GitHub :smiley:

PR #1001, opened 2018-07-30
[wip] Write data to the fake vars file in case it is ignored when empty

richiejp:
This is probably not a good idea, but it may be useful for testing.


coolo:
I thought you guys tested this on staging?

richiejp:
It works, but Santi found another way, so I will close this.

PR #1003, opened 2018-08-01
No console reset

richiejp:
Fix problem with logging into consoles after reverting to a snapshot.

https://progress.opensuse.org/issues/38813

Verification (previously zbar and weechat were failing):
http://10.160.67.78/tests/646
http://10.160.67.78/tests/647

"All consoles need to be "disabled", but not "reset". Resetting consoles causes
the SUT to try logging into the them again, but they may already be logged in
at the point the snapshot was taken. Disabling just disconnects os-autoinst
from the console which is necessary while restarting QEMU. After restarting
QEMU we can reconnect by calling activate directly.

It is assumed that consoles which can't be disabled don't need to be
reënabled."

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -569,9 +569,19 @@ sub disable_consoles {
 
     for my $console (keys %{$testapi::distri->{consoles}}) {
         my $console_info = $self->console($console);
-        $console_info->reset();
         if ($console_info->can('disable')) {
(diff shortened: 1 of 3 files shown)


foursixnine:
If we manage to find another reason to deploy, this change would go in

foursixnine:
Found a reason, but for o3.

PR #1005, opened 2018-08-02
Avoid enabling unused consoles after reverting to a snapshot

richiejp:
Only reënable consoles which are currently active. Any other consoles are
either not being used or will be reënabled during reactivation when next
selected. This avoids activating consoles which are not used in the test.


richiejp:
http://10.160.67.78/tests/652
http://10.160.67.78/tests/653

PR #1006, opened 2018-08-02
Unset vlan on qemu process turndown and runcmd fixes

mudler:
This is part and continuation of work on top of the QEMU backend rewrite (https://github.com/os-autoinst/os-autoinst/pull/985) and a slightly retake on https://github.com/os-autoinst/os-autoinst/pull/829

Boils down to:
- Move qemu process to event emitter model ( untag of vlan with this change happens when qemu process status is finally gathered - although it might still not yet cover all edge cases, i need more feedback )
- Replace IPC::Open3 logic in runcmd_* in osutils.pm
- Refactor pipe communication among internal processes - move internal communication to channels and create classes for the relevant involved processes
- Depends on [Mojo::IOLoop::ReadWriteProcess](https://metacpan.org/pod/Mojo::IOLoop::ReadWriteProcess), but we are already using it in the worker code
- With this change, to ease debug os-autoinst processes, you can enable [MOJO_PROCESS_DEBUG=1](https://metacpan.org/pod/Mojo::IOLoop::ReadWriteProcess#DEBUGGING) and see all the processes-related events, you can also enable [MOJO_EVENTEMITTER_DEBUG=1](https://metacpan.org/pod/Mojo::EventEmitter#DEBUGGING) to inspect all the emitted events as well (bit noisy)

This will require still work later, but it may allows us to cut more code in the long-term as well if we will slowly move towards an Event based approach ( See also discussion https://github.com/os-autoinst/os-autoinst/pull/829#issuecomment-379685717). 
There is still a large room for improvement, this is just a little step towards it.

This is in a unique PR because the changes are tightly related to each other. 

Not ready yet, i need to polish it and double check on real instances - it's working on staging and fullstack tests are passing, who is ready for another haircut? ahem, nice ride? :smile: 

haircut

See: https://progress.opensuse.org/issues/35895, https://progress.opensuse.org/issues/32293, https://progress.opensuse.org/issues/33514

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -46,7 +46,10 @@ use constant STATE_FILE => 'qemu_state.json';
 
 has qemu_bin     => 'qemu-kvm';
 has qemu_img_bin => 'qemu-img';
-has '_process';
+has _process     => sub { process(
+        pidfile       => 'qemu.pid',
+        separate_err  => 0,
+        blocking_stop => 1) };
 
 has _static_params => sub { return []; };
 has _mut_params    => sub { return []; };
@@ -60,11 +63,6 @@ has snapshot_conf   => sub { return OpenQA::Qemu::SnapshotConf->new(); };
 sub new {
     my $self = shift->SUPER::new(@_);
 
-    $self->_process(process(
-            pidfile       => 'qemu.pid',
-            separate_err  => 0,
-            blocking_stop => 1)
-    );
     $self->_push_mut($self->controller_conf);
     $self->_push_mut($self->blockdev_conf);
     $self->_push_mut($self->snapshot_conf);
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -356,7 +356,14 @@ sub exec_qemu {
         });
 
     session->enable_subreaper;
-    $self->_process->code(sub { exec(@params) })->start();
+    $self->_process->code(sub {
+            $SIG{__DIE__} = undef;    # overwrite the default - just exit
+            system $self->qemu_bin, '-version';
+            bmwqemu::diag("starting: " . join(" ", @params));
+            # don't try to talk to the host's PA
+            $ENV{QEMU_AUDIO_DRV} = "none";
+
+            exec(@params) })->separate_err(0)->start();
     fcntl($self->_process->read_stream, Fcntl::F_SETFL, Fcntl::O_NONBLOCK) or die "can't setfl(): $!\n";
 
     return $self->_process->read_stream;
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -37,12 +37,16 @@ use OpenQA::Qemu::BlockDevConf;
 use OpenQA::Qemu::ControllerConf;
 use OpenQA::Qemu::SnapshotConf;
 use osutils qw(gen_params runcmd runcmd_output);
+use Mojo::IOLoop::ReadWriteProcess 'process';
+use Mojo::IOLoop::ReadWriteProcess::Session 'session';
+
 use POSIX ();
 
 use constant STATE_FILE => 'qemu_state.json';
 
 has qemu_bin     => 'qemu-kvm';
 has qemu_img_bin => 'qemu-img';
+has '_process';
 
 has _static_params => sub { return []; };
 has _mut_params    => sub { return []; };
@@ -56,6 +60,11 @@ has snapshot_conf   => sub { return OpenQA::Qemu::SnapshotConf->new(); };
 sub new {
     my $self = shift->SUPER::new(@_);
 
+    $self->_process(process(
+            pidfile       => 'qemu.pid',
+            separate_err  => 0,
+            blocking_stop => 1)
+    );
     $self->_push_mut($self->controller_conf);
     $self->_push_mut($self->blockdev_conf);
     $self->_push_mut($self->snapshot_conf);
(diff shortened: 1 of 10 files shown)


OpenQA/Qemu/Proc.pm
```diff
@@ -56,6 +60,11 @@ has snapshot_conf   => sub { return OpenQA::Qemu::SnapshotConf->new(); };
 sub new {
     my $self = shift->SUPER::new(@_);
 
+    $self->_process(process(
+            pidfile       => 'qemu.pid',
+            separate_err  => 0,
+            blocking_stop => 1)
+    );
```
richiejp: can you create this lazily instead of in new?

autotest.pm
```diff
@@ -212,33 +214,31 @@ sub start_process {
...
-    $SIG{INT}  = 'DEFAULT';
-    $SIG{HUP}  = 'DEFAULT';
-    $SIG{CHLD} = 'DEFAULT';
+            cv::init;
+            require tinycv;
 
-    cv::init;
-    require tinycv;
+            $0 = "$0: autotest";
+            my $line = <$isotovideo>;
```
richiejp: Is it safe to read in this way from a/this socket? Although I see below that it has always been read like this. I guess this type of read is buffered, so there better not be any unbuffered IO on this socket elsewhere in the code.

backend/driver.pm
```diff
@@ -45,6 +47,12 @@ sub new {
     $self->{backend}      = "backend::$name"->new();
     $self->{backend_name} = $name;
 
+    session->on(
+        collected_orphan => sub {
+            my ($session, $p) = @_;
+            printf STDERR "Driver backend collected unknown process with pid " . $p->pid . " and exit status: " . $p->exit_status . "\n";
+        });
+
```
richiejp: This seems to be a common pattern.

autotest.pm
mudler: Well, nope if you ask me, but as you said it always been like this - as this will already shake things a lot, better do in a separate PR imho (hence why i said there is a lot of room for improvement :) )

OpenQA/Qemu/Proc.pm
mudler: Sure

backend/driver.pm
mudler: Yup, but it is also in os-autoinst/openQA as well

OpenQA/Qemu/Proc.pm
mudler: But next step, would be proper inheritance here (also in Qemu::Proc)

commands.pm
```diff
@@ -340,24 +342,22 @@ sub start_server {
...
-        close($child);
-        $0 = "$0: commands";
-        run_daemon($port, $isotovideo);
-        Devel::Cover::report() if Devel::Cover->can('report');
-        _exit(0);
-    }
     close($isotovideo);
-
-    return ($pid, $child);
+    $process->on(collected => sub { print STDERR "commands process exited: " . shift->exit_status . "\n"; });
```
foursixnine: Why not use the diag function here?

backend/driver.pm
```diff
@@ -53,41 +61,34 @@ sub new {
...
 
-        # now initialize opencv
-        require cv;
+            $self->{backend}->run(fileno($process->channel_in), fileno($process->channel_out));
+            _exit(0);
+    })->blocking_stop(1)->separate_err(0)->subreaper(1)->start;
 
-        cv::init();
-        require tinycv;
+    $backend_process->on(collected => sub { diag "backend process exited: " . shift->exit_status; });
```
foursixnine: Somehow this message scapes the logger, and ends up all naked, the same as the other message above, perhaps the logger facility goes out of scope?

osutils.pm
```diff
@@ -92,79 +92,38 @@ sub quote {
...
-    die "couldn't open: $!" unless defined $pid;
-    close($wtr) or die "couldn't close fh: $!";
-
-    $sub->($rdr, $err);
+sub _run {
+    diag "running " . join(' ', @_);
+    my $e    = pop;
+    my @args = @_;
+    my $out;
+    my $p = process(sub { exec(@args) })->separate_err($e)->start;
```
foursixnine: Hopefully this is a good safeguard for the bitshift madness...

backend/qemu.pm
```diff
@@ -111,7 +106,7 @@ sub eject_cd {
 
 sub cpu_stat {
     my $self = shift;
-    my $stat = bmwqemu::fileContent("/proc/" . $self->{pid} . "/stat");
+    my $stat = bmwqemu::fileContent("/proc/" . $self->{proc}->_process->pid . "/stat");
```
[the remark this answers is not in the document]
mudler: Yeah, slowly we should move to it, also in os-autoinst/openQA we should be bit more consistent on it's usage

commands.pm
mudler: forgot :)

backend/driver.pm
mudler: Possible, have to look at that still

mudler: Checked now.. actually it's not using bmwqemu::diag, but the one in the driver (which is naked) https://github.com/os-autoinst/os-autoinst/blob/master/backend/driver.pm#L37 . wanna use bmwqemu::diag instead? i thought we wanted to get rid of bmwqemu at some point

commands.pm
mudler: Just, checked, also here. we didn't used to import bmwqemu here - see https://github.com/os-autoinst/os-autoinst/pull/1006#discussion_r207459425

backend/driver.pm
foursixnine: Yeah, but We'd rather have it all centralized so that we don't end up with this kind of confusion :) I mean, bmwqemu::diag is already adding the stuff to the logging (maybe wrap it?). And we can remove it later (and easier) when we revamp the logging functions...

commands.pm
foursixnine: makes me torn between both xD

foursixnine:
/o/ http://loewe.arch.suse.de/tests/1059
\o\ http://loewe.arch.suse.de/tests/1058
Looks like we're on a good track then

backend/driver.pm
richiejp: It would be useful to know the original code location where  the call back was set as well

mudler: @richiejp argh, missed the comment, good point

PR #1007, opened 2018-08-03
virtio_terminal: Avoid connecting to already connected socket

richiejp:
Another day another os-autoinst PR :-). This only effects one test to my knowledge, but it is quite an important test because it is for an important product bug.

The console can be reset (e.g. when rebooting the SUT) without closing the
console socket because QEMU is not restarted. So keep using the currently open
socket if there is one.

Verification: http://10.160.67.78/tests/669

### DIFF
--- a/consoles/virtio_terminal.pm
+++ b/consoles/virtio_terminal.pm
@@ -133,7 +133,7 @@ sub open_socket {
 sub activate {
     my ($self) = @_;
     if (get_var('VIRTIO_CONSOLE')) {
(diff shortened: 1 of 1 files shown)


foursixnine:
Looks ok so far, let's wait for the test to finish, as http://10.160.67.78/tests/669#step/asyncio02/6 would prove the point (since it would revert the snapshot afterwards right?)

richiejp:
yeah

PR #1008, opened 2018-08-06
osutils: Avoid racing to get output

mudler:
Instead, just read on stop

PR #1009, opened 2018-08-07
Add log output for sim level in wait_still_screen

DrMullings:
Now prinitng the detected similarity level in wait_still_screen

As enhancement to https://progress.opensuse.org/issues/15310
Verification: http://pinky.arch.suse.de/tests/1372/file/autoinst-log.txt
`[2018-08-07T09:51:59.0025 CEST] [debug] >>> testapi::wait_still_screen: detected same image for 1 seconds, last similarity was 1000000`


DrMullings:
@okurz wdyt?

okurz:
+1 . Does it really work? I mean, does it also show values lower than the specified limit, e.g. in the one example you originally worked on?

DrMullings:
Since I have fixed that before I cannot tell for sure, but according to the code I see no reason why it should not

DrMullings:
I freshly cloned it yesterday.
Regarding the danger of an uninitialized variable: shouldn't the chance be the same as before, meaning this would only happen if L#570 fails somehow?

foursixnine:
@coolo I think the coverage drop might be related to https://github.com/os-autoinst/os-autoinst/pull/1008#issuecomment-410687077 to some extent

DrMullings:
I added an initialization for $sim, just to go sure

coolo:
where?

DrMullings:
Locally in L#568 -_-
It should be online now, sorry for that

DrMullings:
At that time I have rebased it to the current version, yes

PR #1010, opened 2018-08-07
qemu: Allow user to set HDD serial number

richiejp:
Should 'fix' migrating from SLE11 and Leap 13 by allowing the user to set the drive serial to '1', '2', etc.

Verification: http://10.160.67.78/tests/685


richiejp:
@rwx788

richiejp:
Also a verification test without the variable set: http://10.160.67.78/tests/686

okurz:
+1, makes sense :)

rwx788:
Great, thanks!

PR #1011, opened 2018-08-08
Fix ppc64le flaky exit status codes

mudler:
This was a fallout of https://github.com/os-autoinst/os-autoinst/pull/1006 - it highlighted the problem we was having already to get the exit statuses, but the fallout was regarding only ppc64le (which made the thing bit more harsh to debug).  

This removes also runcmd_output (just used in OpenQA::Qemu::Proc) as now we can slim a bit more the duplicated code.

It requires also new RWP to treat the specific arches case (and also to make code bit slimmer) and to fix a minor bug i found while looking at this.

Also added on the Docker image the possibility to install the modules from CPAN (default is disabled)  when INSTALL_FROM_CPAN=1 is supplied to env variables.

( Travis tests will fail as new RWP still didn't hit repos )

### DIFF
--- a/docker/travis_test/run.sh
+++ b/docker/travis_test/run.sh
@@ -2,11 +2,20 @@
 
 set -e
 
+INSTALL_FROM_CPAN="${INSTALL_FROM_CPAN:-0}"
+
 # Prepare dir and chdir into it before executing the wanted action
 sudo cp -rd /opt/repo /opt/run
 sudo chown -R $NORMAL_USER:users /opt/run
 
 pushd /opt/run
+[ "$INSTALL_FROM_CPAN" -eq 1 ] && \
+  (cpanm --local-lib=~/perl5 local::lib && cpanm -n --installdeps . ) || \
+  cpanm -n --mirror http://no.where/ --installdeps .
+
+[ "$INSTALL_FROM_CPAN" -eq 1 ] && eval $(perl -I ~/perl5/lib/perl5/ -Mlocal::lib)
+[ $? -eq 0 ] || echo "Missing dependencies. Please check output above"
+
 ./autogen.sh
 
 /bin/bash -c "$*"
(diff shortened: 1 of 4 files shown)


docker/travis_test/run.sh
```diff
@@ -2,11 +2,20 @@
...
 set -e
 
+INSTALL_FROM_CPAN="${INSTALL_FROM_CPAN:-0}"
+
 # Prepare dir and chdir into it before executing the wanted action
 sudo cp -rd /opt/repo /opt/run
 sudo chown -R $NORMAL_USER:users /opt/run
 
 pushd /opt/run
+[ "$INSTALL_FROM_CPAN" -eq 1 ] && \
```
foursixnine: So for now this will be deactivated right?

mudler: Yup, it's still bit tricky to use if you run the tests from Makefile, but if you use docker run manually, you need to supply -e INSTALL_FROM_CPAN=1 to get there

rwx788:
Seems that it's related to the failure I get when run worker with local copy of os-autoinst:
http://g226.suse.de/tests/2264/file/autoinst-log.txt
Could you please take a quick look. Process we start in autotest exits immediately.
And I have perl-Mojo-IOLoop-ReadWriteProcess-0.22 installed. Thanks!

rwx788:
Seems that it's related to the failure I get when run worker with local copy of os-autoinst:
http://g226.suse.de/tests/2264/file/autoinst-log.txt
Could you please take a quick look. Process we start in autotest exits immediately.
And I have perl-Mojo-IOLoop-ReadWriteProcess-0.22 installed. Thanks!
When I've reverted this one, it works fine, so that's the only proof I have so far, let me know if you are interested in more details.

rwx788:
@mudler I've installed 0.23 from cpan and it works fine now. Thanks!

mudler:
Thanks for investigating :+1:

PR #1012, opened 2018-08-10
[WIP] Add ssh backend

cfconrad:
It is a first proposal to discuss, if I'm on the right track.

Related ticket: https://progress.opensuse.org/issues/38744

Why do we need a SSH only backend:
We need it in public cloud release chain, to launch openqa tests on an already 
running instance. Only SSH credentials and not CSP framework
credentials get shared between MESH and openQA.

Abstract:
Create a named pipe on SUT. Use it as $testapi::serialdev so exit codes can parsed.
Use a second SSH connection to execute scripts.

Further I would like discuss the following:
* How to log executed scripts and its output
* Could this be used with localXvnc console
* Log kernel messages
* Where should we locate the named pipe

thx for help

### DIFF
--- a/backend/ssh.pm
+++ b/backend/ssh.pm
@@ -20,10 +20,15 @@ use testapi qw(get_required_var set_var get_var);
 use Carp 'cluck';
 use Mojo::IOLoop::ReadWriteProcess 'process';
 use IO::Select;
+use Errno ':POSIX';
 
 sub new {
     my $class = shift;
     my $self  = $class->SUPER::new;
+
+    # overwrite WORKER_HOSTNAME, use a SSH reverse tunnel.
+    $bmwqemu::vars{WORKER_HOSTNAME} = 'localhost';
+    bmwqemu::save_vars();
 
     return $self;
 }
@@ -63,6 +68,23 @@ sub do_start_vm {
     return {};
 }
 
+sub save_syswrite
+{
+    my ($fd, $buf, $len) = @_;
+
+    my $wr = 0;
+    do {
+        my $i = syswrite($fd, $buf);
+        if (!defined($i)) {
+            next if ($!{EAGAIN});
+            return;
+        }
+        $wr += $i;
+        $buf = substr($buf, $i);
+    } while ($wr < $len);
+    return $wr;
+}
+
 sub start_reverse_tunnel {
     my ($self) = @_;
 
@@ -95,19 +117,20 @@ sub start_reverse_tunnel {
                 );
                 if ($socket) {
                     my $buf = '';
-                    my $len = 512;
+                    my $len = 4096;
                     my $s   = IO::Select->new();
                     $s->add($socket);
                     while (1) {
                         last if ($channel->eof);
-                        while (sysread($channel, $buf, $len)) {
-                            last unless syswrite($socket, $buf);
+                        while (my $read = sysread($channel, $buf, $len)) {
+                            goto OUT unless save_syswrite($socket, $buf, $read);
                         }
-                        if ($s->can_read(0.1)) {
-                            last unless sysread($socket, $buf, $len);
-                            last unless syswrite($channel, $buf);
+                        while ($s->can_read(0.1)) {
+                            goto OUT unless (my $read = sysread($socket, $buf, $len));
+                            goto OUT unless save_syswrite($channel, $buf, $read);
                         }
                     }
+                  OUT:
                     $socket->close();
                 }
                 $channel->close();
--- a/backend/ssh.pm
+++ b/backend/ssh.pm
@@ -1,0 +1,143 @@
+# Copyright © 2018 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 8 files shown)


backend/ssh.pm
```diff
@@ -0,0 +1,143 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package backend::ssh;
+use strict;
+use base 'backend::baseclass';
```
richiejp: As discussed verbally, I think it is nice to use the Mojo base classes, they provide some nice setting and getter functions amongst other things (see OpenQA/Qemu/*). However this class must derive from backend::basclass so it may be a good idea to move some of the functionality here to the library OpenQA/ssh/SomeClass.pm where you can derive from the Mojo base class.

backend/ssh.pm
```diff
@@ -0,0 +1,143 @@
...
+    $ssh_console->backend($self);
+    $self->select_console({testapi_console => 'sut'});
+
+    # TODO use unique filename or exit if file exists
+    my $serial_file = "/dev/" . $testapi::serialdev;
+    $self->run_cmd("mkfifo $serial_file", $hostname, $password, $username);
+
+    # Listen on serial file
+    my $chan = $self->start_ssh_serial(hostname => $hostname, password => $password, username => $username);
+    $chan->exec("tail -F $serial_file");
```
richiejp: As discussed, we now have a separate code path for when we are communicating over a duplex serial connection. So there is no need to create a serial file for this purpose. You probably just want the serial file (serial0.txt) for logging kernel messages over a separate connection to the one used for sending commands to the SUT. Hopefully the cloud provider gives you a dedicate serial feed for this which doesn't require an ssh daemon to be running on the SUT, but otherwise you can figure out some workaround.

backend/ssh.pm
```diff
@@ -0,0 +1,143 @@
...
+        return ($stdout, $errout);
+    }
+    else {
+        bmwqemu::diag "Command's stdout:\n$stdout" if length($stdout);
+        bmwqemu::diag "Command's stderr:\n$errout" if length($errout);
+    }
+}
+
+# TODO used from svirt.pm, move to baseclass?!
+sub run_cmd {
```
richiejp: Possibly you could connect the console and use (assert_)script_run even internally. Otherwise you can indeed move this to OpenQA/Ssh/SomeLib.pm as mentioned above.

backend/ssh.pm
cfconrad: Thx for pointing this explicitly. I will go that advice and use  also the duplex serial connection for SSH.

consoles/ssh_console.pm
```diff
@@ -0,0 +1,71 @@
...
+
+    my $hostname = $self->{hostname} || die('we need a hostname to ssh to');
+    my $password = $self->{password};
+    my $username = $self->{username};
+
+    $self->{ssh} = $self->backend->new_ssh_connection(hostname => $hostname, password => $password, username => $username);
+    $self->{shell} = $self->{ssh}->channel();
+    $self->{shell}->shell();
+
+    $self->{screen} = consoles::virtio_screen->new($self->{shell});
```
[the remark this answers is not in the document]
cfconrad: The channel, which is used from NET::SSH2 for communication, need a other way of reading data then using fileno(fd) and select()
I'm thinking in extending virtio_screen and overwrite the read_until so it fits to NET:SSH2::Channel.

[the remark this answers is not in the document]
cfconrad: I took this idea, thx.  But I still needed the NET::SSH2:Channel object to write/read data from the remote shell. But I could use select() with NET::SSH2->sock, which makes the change in virtio_screen very little. The NET::SSH2 object should not be used with multiple channels.

consoles/ssh_console.pm
```diff
@@ -0,0 +1,74 @@
...
+    my ($self, $args) = @_;
+
+    my $hostname = $self->{hostname} || die('we need a hostname to ssh to');
+    my $password = $self->{password};
+    my $username = $self->{username};
+
+    $self->{ssh} = $self->backend->new_ssh_connection(hostname => $hostname, password => $password, username => $username);
+    my $chan = $self->{shell} = $self->{ssh}->channel();
+    $chan->pty(1);
+    $chan->shell();
```
richiejp: Perhaps blocking should be explicitly disabled here?

richiejp:
> 
> NET::SSH2 sock can be used with select, but we still need
> the NET::SSH2:Channel for read/write.
> Added a optional paramter to specify the fd which will be used
> in select.

Interesting, I guess if you have multiple channels then select will sometimes exit prematurely because a different channel received data, but this is OK because we only rely on select to reduce the amount of work we do while waiting for data.

consoles/ssh_console.pm
cfconrad: What is the reason to use it as non-blocking? Is it because of method peak() ?

richiejp: Yes, but more importantly we need to timeout if we receive no data in a sensible time. The POSIX read syscall has no timeout option, so it would block until the socket is closed or the process receives a signal.

There are various ways to handle this, however using a non-blocking socket and reading in a loop is probably the easiest. To save CPU cycles while waiting for bytes to appear, we are blocking using select() which does have a timeout.

cfconrad:
I think a native ssh-backend doesn't make sense anymore.
A ssh-connection is more a console which is the case in svirt and other backends.
A lot of the things I did here are already moved or will be moved with https://github.com/os-autoinst/os-autoinst/pull/1166 to existing implementations.

PR #1013, opened 2018-08-14
Allow to pause on check_screen timeout & refactor isotovideo's command processing

Martchus:
A start to implement https://progress.opensuse.org/issues/39566.

After adding one further of those `if ($rsp->{cmd} eq '...') {` blocks I guess it is time to refactor this. This would also ease writing unit tests for the particular command handler. So I have splitted the if's in `process_command` into separate functions and move them also to a separate module.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -305,8 +305,9 @@ my $current_test_full_name;
 # store the name of the test where the test execution should be paused
 my $pause_test_name = $bmwqemu::vars{PAUSE_AT};
 
-# store whether to stop on next assert_screen timeout
+# store whether to stop on next assert_screen/check_screen timeout
 my $pause_on_assert_screen_timeout = $bmwqemu::vars{PAUSE_ON_ASSERT_SCREEN_TIMEOUT} // 0;
+my $pause_on_check_screen_timeout  = $bmwqemu::vars{PAUSE_ON_CHECK_SCREEN_TIMEOUT} // 0;
 
 # when paused, the reason for the pause (as a string); otherwise 0
 my $test_execution_paused = 0;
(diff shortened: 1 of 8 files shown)


Martchus:
I'll deploy this when `openqaworker11` is available again. So for now I started to add some first unit tests.

Martchus:
Deployed on e212/openqaworker11: https://e212.suse.de/tests/12330

Martchus:
The acceptance test passed: https://e212.suse.de/tests/12330

Martchus:
@foursixnine I think it makes sense to merge this. Tested the refactored code multiple times locally and the acceptance test on e212 also passed.

isotovideo
```diff
@@ -335,14 +296,9 @@ sub _calc_check_delta {
...
-    if ($delta > 0) {
-        # sleep the remains of one second
-        $timeout = 1 - $delta;
-        $timeout = 0 if $timeout < 0;
-    }
-    else {
-        $timeout = 0;
-    }
+    # sleep the remains of one second if $delta > 0
+    my $timeout = $delta > 0 ? 1 - $delta : 0;
```
StefanBruens: `my $timeout = max(0, min(1 - $delta, 1))` ?

or even add a helper, `clamp(0, 1 - $delta, 1)` - I think this construct is used in several places.

Martchus: In my opinion the max/min version is even harder to read. Adding a helper would make sense, but only if we give it a meaningful name. (I don't think `clamp` is meaningful enough.)

Martchus:
Don't merge this yet. It breaks the fullstack test in openQA. Luckily only 2 tests fail and in a reproducible way.

PR #1015, opened 2018-08-21
Prevent double timestamps on QEMU command line log entry

richiejp:
By avoiding nesting calls to Mojo log. This may also prevent error messages
from QEMU being dropped and the command line being truncated.

https://progress.opensuse.org/issues/40046


richiejp:
@okurz http://10.160.67.78/tests/694/file/autoinst-log.txt

PR #1016, opened 2018-08-23
Reduce impact of VM snapshots on ARM platform (bsc#1105302)

dirkmueller:
The compression of 8 threads with level 9 is excessively slow on a 4-6
core machine, so we lower this a bit which helps it cope better
with the load.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -283,6 +283,8 @@ sub _migrate_to_file {
             arguments => {
                 # This is ignored if the compress capability is not set
                 'compress-level' => $compress_level,
+                # Default of 8 causes high load spikes for concurrent runs
+                'compress-threads' => 2,
                 # Ensure slow dump times are not due to a transfer rate cap
                 'max-bandwidth' => $max_bandwidth,
               }
@@ -402,8 +404,10 @@ sub save_snapshot {
     });
 
     mkpath(VM_SNAPSHOTS_DIR);
+    my $migrate_compress_level = 6;
+
     $self->_migrate_to_file(filename => VM_SNAPSHOTS_DIR . '/' . $snapshot->name,
-        compress_level => 9);
+        compress_level => $migrate_compress_level);
     diag('Snapshot complete');
 
     $self->cont_vm() if $was_running;


backend/qemu.pm
```diff
@@ -402,8 +404,13 @@ sub save_snapshot {
     });
 
     mkpath(VM_SNAPSHOTS_DIR);
+    my $migrate_compress_level = 9;
+    # ARM is slow, and high compression level is much slower cpu wise than
+    # disk speed, so we tradeoff faster migration over higher compression
+    $migrate_compress_level = 2 if if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm');
```
foursixnine: I wonder if `$vars->{QEMU_SNAPSHOT_COMPRESSION} // 2` would be better and have that setting defined in the workers.ini, as this might affect other architectures/machines where more workers are running and there's not so much space (and document it on backend_vars.asciidoc file in the meantime)

backend/qemu.pm
```diff
@@ -283,6 +283,8 @@ sub _migrate_to_file {
             arguments => {
                 # This is ignored if the compress capability is not set
                 'compress-level' => $compress_level,
+                # Default of 8 causes high load spikes for concurrent runs
+                'compress-threads' => 2,
```
foursixnine: @richiejp wdyt? imo compress-threads would then/should be configurable too...with the same or similar treatment as compress_level below...

richiejp: yeah, sure. I guess I assumed QEMU would choose a sensible default for the arch or just overlooked.

backend/qemu.pm
richiejp: I think 9 should be the default because x86_64 and Power8 have no issue with it, but yeah, it should be configurable.

Also there appears to be two ifs.

backend/qemu.pm
dirkmueller: the default of 8 is documented, and I think it isn't a sensible default for openqa in any case, no matter which architecture we're talking about because it suddenly eightfolds the cpu load on a load that is anyway cpu bound.

backend/qemu.pm
dirkmueller: sure they don't have "an issue" but both x86 and power 8 have significantly faster I/o than this compression is ever gonna save. 

note that this is compressing 4k pages at a time, not a stream compression. while I haven't been able to do elaborate comparisons, a stream gzip on compression level of 1 (the least best but most fast compression) is ten fold (!) better compressing than qemu's builtin compression at level 9. I would argue that it isn't worth the cpu overhead on any architecture, but since that is probably contentious by the openqa maintainers I limited it to aarch64.

foursixnine: @dirkmueller I get the thing of x86 not having an issue, but the actual change makes sense to me, just with $vars->{QEMU_SNAPSHOT_COMPRESSION} // 2 :)

backend/qemu.pm
richiejp: We could limit it to QEMUCPU so the compression can only produce as much load as the VM/SUT itself could. Users should not give their machines more cores than the worker can handle.

During my testing it didn't make a lot of difference what compression level is being used so setting it to 6 is probably fine.

> on a load that is anyway cpu bound

This isn't true in all cases. For example, most deployments are actually a single worker on a test developer's machine. On all my development machines the workload is I/O bound.

Also a lot of the server grade ARM machines I have seen have 48-96 cores, but poor single core performance, so it makes more sense to me to reduce the number of workers on these machines, but give each worker more cores.

coolo:
superseded by @richiejp from what I can see

PR #1018, opened 2018-08-29
qemu: Add knobs and dials for tuning snapshotting

richiejp:
Follow on from https://github.com/os-autoinst/os-autoinst/pull/1016

I will test the various compression levels on ARM.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -267,10 +267,11 @@ sub _wait_for_migrate {
 
 sub _migrate_to_file {
     my ($self, %args) = @_;
-    my $fdname         = 'dumpfd';
-    my $compress_level = $args{compress_level} || 0;
-    my $filename       = $args{filename};
-    my $max_bandwidth  = $args{max_bandwidth} // LONG_MAX;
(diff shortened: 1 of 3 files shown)


richiejp:
https://bugzilla.opensuse.org/show_bug.cgi?id=1105302#c12

coolo:
please update the docs - otherwise fine

richiejp:
OK, added variable documentation.

backend/qemu.pm
```diff
@@ -282,9 +283,10 @@ sub _migrate_to_file {
             execute   => 'migrate-set-parameters',
             arguments => {
                 # This is ignored if the compress capability is not set
-                'compress-level' => $compress_level,
+                'compress-level'   => $compress_level + 0,
```
foursixnine: Why are you forcing to integers? Shouldn't block from merging though

richiejp: Because I found that QEMUCPUS was set to a string in one case and then QEMU throws an error because it is expecting a JSON integer, but it got a string. So I decided to just coerce everything to integers.

PR #1019, opened 2018-08-30
Add SCHEDULE for manual override; Prefer openQA override test modules (poo#36994)

okurz:
* Prefer override test module files if found in the openQA "other" repo path
* isotovideo: Require explicit 'tests/' prefix within SCHEDULE again for openQA use
* Allow to start tests without relying on a vars.json file at all
* Add variable SCHEDULE to define the full schedule by variable


okurz:
tests fail for the same reason the test currently fails on master as can be seen on https://travis-ci.org/os-autoinst/os-autoinst : "not ok 19 - Snapshots are not supported" in the full stack test. I hope e.g. @richiejp is on it?

richiejp:
If you look carefully you will see that no actual test cases are failing on master, it is codecov which is failing.

foursixnine:
@richiejp https://travis-ci.org/os-autoinst/os-autoinst/builds/423295319#L1886 :) sounds like true story to me But indeed, codecov seems to be failing too to load results

richiejp:
Yeah, it fails on Oliver's branch, I don't see it failing on master. I tried running it locally and it passed.

Looking at Oliver's code he has moved bmwqemu::init to after the forced variables are set (which appears to be where the disable snapshots variable will be set on this test). The init function calls load_vars which appears to overwrite the vars hash map with the contents of vars.json. So all the forced variables will be dropped.

okurz:
Thanks for investigating, I will take a look

okurz:
https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Leap_15.0/x86_64 shows the same problem I had when trying to execute tests locally on Leap 15.0, os-autoinst does not depend on a specific version of perl-Mojo-IOLoop-ReadWriteProcess but should probably depend on >= 0.23 . Created a submission https://build.opensuse.org/request/show/633699 now

okurz:
Updated, fixed more perl warnings that were there in before and made sure that `make check` passes locally

okurz:
seems like no tests have been executed now whatsoever. I doubt these can explicitly be triggered. I rebased and pushed again. This triggered new tests.

PR #1020, opened 2018-09-04
Ballooning

richiejp:
Built on top of #1018 this allows the user to set a balloon target before a snapshot is called. In theory this should force the guest kernel to give up some RAM and make the snapshots smaller. In practice the snapshots don't appear to get any smaller although it seems there is some performance improvement taking the snapshot.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -370,6 +370,42 @@ sub save_storage_drives {
     return;
 }
 
+sub inflate_balloon {
+    my $self = shift;
+    my $vars = \%bmwqemu::vars;
+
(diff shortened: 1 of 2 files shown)


richiejp:
https://www.youtube.com/watch?v=dQEmaj9C6ko

okurz:
hi, would you mind rebasing and bringing it up again for review?

richiejp:
OK, but I don't want to test it right now. However I think you can safely merge it. Just don't enable it for all tests at the same time.

okurz:
understood. Thanks.

PR #1022, opened 2018-09-11
Prevent isotovideo <=> command server IPC to occasionally lose messages

Martchus:
see https://progress.opensuse.org/issues/40793 and https://github.com/os-autoinst/openQA/pull/1782


Martchus:
The fullstack test passed 2 times in a row and the developer fullstack test 6 times (on Travis and already with the new scheduler). There were actually no failures at all so far (beside another random one in the UI tests). So it looks like this change actually helped.

It failed now here because of perlcritic. I fixed it and if the CI finishes and passes I'd say this is ready to be merged.

PR #1023, opened 2018-09-11
Treat $OpenQA::Isotovideo::Interface::version as test API

Martchus:
See https://progress.opensuse.org/issues/40823#note-2


mudler:
I think the comment was refering to have a specific api call on tests side to get the interface number - instead of accessing it directly

Martchus:
Yes, it would be a chicken/egg problem indeed. Since the tests are now already using the variable directly I would just keep it that way for the time. And that's actually not a big deal. I've just didn't know about that usage when refactoring.

PR #1024, opened 2018-09-12
Die when remote backend programs are missing

SergioAtSUSE:
Testing, but PR open for feedback.

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -34,6 +34,9 @@ sub callxterm {
     $command = "TERM=xterm $command";
     my $xterm_vt_cmd = which "xterm-console";
     die "Missing 'xterm-console'" unless $xterm_vt_cmd;
+    die('Missing "Xvnc"')         unless which('Xvnc');
+    die('Missing "icewm"')        unless which('icewm');
+    die('Missing "xterm"')        unless which('xterm');
     eval { system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\""); };
     if (my $E = $@) {
         die "cant' start xterm on $display (err: $! retval: $?)";


consoles/localXvnc.pm
```diff
@@ -96,7 +96,8 @@ sub activate {
     sleep 1;
 
     # we need a window manager for fullscreen apps to work
-    system("DISPLAY=$display icewm -c $bmwqemu::scriptdir/consoles/icewm.cfg & echo \"icewm PID is \$!\"");
+    $icewm_return_code = system("DISPLAY=$display icewm -c $bmwqemu::scriptdir/consoles/icewm.cfg & echo \"icewm PID is \$!\"");
+    die('sh: icewm: command not found') if ($icewm_return_code == 32512);
```
Martchus: You check for the exact return code `32512` here? I doubt that this return code will always be that arbitrary number. Why not checking `!= 0` instead?

coolo: this is exit status 127 if you pass it through WEXITSTATUS :)

mudler: maybe is safer to check if the binary is actually there before calling it or even  you can also use run_cmd https://github.com/os-autoinst/os-autoinst/search?q=runcmd&unscoped_q=runcmd

SergioAtSUSE: $? == @127
But, system(...) == 32512

I didn't want to let it die on other cases.

SergioAtSUSE:
What do you think now?

okurz:
shouldn't be WIP I guess

coolo:
Is this really the only program left to be checked? You better take the opportunity to review all tools used without check

coolo:
then please:
- run ./tools/tidy
- squash your commits
- remove WIP from title

and we should be good to merge

SergioAtSUSE:
@coolo, it should be ready.

The code coverage complaints. Should I do something about it?

PR #1025, opened 2018-09-13
script_output: increase logging in case of failure

asmorodskyi:


Martchus:
Looks good, beside that it isn't formatted with `tools/tidy`.

distribution.pm
```diff
@@ -245,10 +245,16 @@ sub script_output {
         testapi::type_string("($run_script) | tee /dev/$testapi::serialdev\n");
     }
     my $output = testapi::wait_serial("SCRIPT_FINISHED$marker-\\d+-", $wait, 0, record_output => 1)
-      || croak "script timeout";
+      || croak "script timeout: $script";
 
     if ($output !~ "SCRIPT_FINISHED$marker-0-") {
-        croak "script failed with : $output" unless $args{proceed_on_failure};
+        my $log_message = 'script failed with :' . $output;
```
Martchus: The space should go after the colon.

Martchus:
I would ignore the coverage checks (eg. the check for patch shows lots of red lines which aren't actually part of the patch).

distribution.pm
asmorodskyi: Done

okurz:
Could you provide a link to an openQA test which shows the difference in action?

asmorodskyi:
http://kimball.arch.suse.de/tests/1761/file/autoinst-log.txt - here you go , look for "distribution::script_output(script failed with :"

okurz:
+1

PR #1026, opened 2018-09-17
Add hard_reset API to the console sshXtermVt

jerrytt:
When using ipmi backend , installation will performed in ssh-root console which need ssh.
(also installation will reboot in himself)
prepare_system_shutdown function will break this connection ,  which break the reboot .

reboot should not triggered in installation self, use this api after prepare_system_shutdown.

### DIFF
--- a/consoles/sshXtermVt.pm
+++ b/consoles/sshXtermVt.pm
@@ -61,4 +61,10 @@ sub kill_ssh {
     $self->backend->stop_ssh_serial;
 }
 
+# be be called when normal reboot is no avaiable
+sub hard_reset {
+    my ($self) = @_;
+    $self->backend->restart_host;
+}
+
 1;


consoles/sshXtermVt.pm
```diff
@@ -61,4 +61,10 @@ sub kill_ssh {
     $self->backend->stop_ssh_serial;
 }
 
+# when use ipmi backend with ssh , reboot code should not be invoked in ssh
```
okurz: I don't understand this comment. So when *should* this be used? Why should this console also be special in this regard? I suggest to extend the comments for the methods in this file with proper POD format to describe what in details is necessary and special. How does this method relate to the above "kill_ssh"?

alice-suse:
@coolo @foursixnine  @mitiao  Would you please help to make a review?

alice-suse:
The PR is intended for https://progress.opensuse.org/issues/32746#change-150167. You can get full story there.

consoles/sshXtermVt.pm
jerrytt: > I don't understand this comment. So when _should_ this be used? 

in the context of installation on remote ssh connection .
1. backend monitor the socket rate .
2. ssh connection killed may cause half-open
3. ssh connection will be killed after install 

here , basically no stander way close ssh connectison . because last step of installation is reboot , which kill ssh

4. kill-ssh  method will remove the sock from monitor .



  
> Why should this console also be special in this regard?

Well , I'm not sure , i hope you can tell where to add.
 but this issue is about ssh connection can not disconnect normally ,
 only this type console have this issue. maybe should add to all ssh relate console. 

>I suggest to extend the comments for the methods in this file with proper POD format to describe what in details is necessary and special. How does this method relate to the above "kill_ssh"?

refer to 4 in above

coolo:
this sounds wrong to me. You jump layers up and down. In the tests you check if BACKEND is ipmi to then use a rather random console to trigger a backend action.

To me it sounds like you want the power_action to trigger the backend action directly - because that's what the tests already do.

consoles/sshXtermVt.pm
jerrytt: update new comment

coolo:
Sorry, this is wrong.

jerrytt:
> Sorry, this is wrong.

I coolo, we need track and fix this issue 
https://progress.opensuse.org/issues/32746

could you explain why add a api  to consoles/sshXtermVt.pm is wrong ?

and give some direction how to fix this ?

coolo:
https://github.com/os-autoinst/os-autoinst/pull/1026#issuecomment-422703219 is the direction

jerrytt:
> [#1026 (comment)](https://github.com/os-autoinst/os-autoinst/pull/1026#issuecomment-422703219) is the direction

but i don't understand . 
power action is kind of frontend  code.
it can't create magic hard_reset  function without backend supported

coolo:
Well, I don't have time to solve it for you. I'm saying the tests already call power_action, so isotovideo has all informations to close the ssh channels properly.

jerrytt:
but it's not about close ssh channels.
it's about want use ssh reboot .but u kill it before reboot, and expect ssh reboot works anyway.

jerrytt:
I understand @coolo,
i will find another developer help me.

jerrytt:
@foursixnine @mitiao  
can you guys help me on  understanding coolo's suggestion?

jerrytt:
As coolo mention above,  I move all  action to prepare_system_shutdown.
but still need this backend api.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5778#issuecomment-429227141

PR #1027, opened 2018-09-17
Download new needles from openQA on resume

Martchus:
see https://progress.opensuse.org/issues/41075
openQA: https://github.com/os-autoinst/openQA/pull/1793

### DIFF
--- a/OpenQA/Isotovideo/NeedleDownloader.pm
+++ b/OpenQA/Isotovideo/NeedleDownloader.pm
@@ -35,25 +35,33 @@ has openqa_url        => sub {
     my $url  = Mojo::URL->new($bmwqemu::vars{OPENQA_URL});
     my $host = $bmwqemu::vars{OPENQA_HOSTNAME};
 
-    # assume 'http' if scheme is missing
-    if (!$url->scheme) {
-        $url->scheme('http');
-    }
-
     # determine host if not present in OPENQA_URL
     if (!$url->host) {
         my $path_parts = $url->path->parts;
         if (scalar @$path_parts == 1) {
-            # treat URLs like just 'e212.suse.de' in the way that 'e212.suse.de' is the hostname (and not a relative path)
-            $url->host($path_parts->[0]);
+            if ($url->scheme) {
+                # treat URLs like 'localhost:9526' in the way that 'localhost' is the hostname (and not the protocol)
+                $url->host($url->scheme);
+                $url->port($path_parts->[0]);
+                $url->scheme('');
+            }
+            else {
+                # treat URLs like just 'e212.suse.de' in the way that 'e212.suse.de' is the hostname (and not a relative path)
+                $url->host($path_parts->[0]);
+            }
             $url->path(Mojo::Path->new);
         }
         elsif ($host) {
             # build a default URL from OPENQA_HOSTNAME if no host in OPENQA_URL is missing
             $url = Mojo::URL->new();
-            $url->scheme('http');
+            $url->scheme('');
             $url->host($host);
         }
+    }
+
+    # assume 'http' if scheme is missing
+    if (!$url->scheme) {
+        $url->scheme('http');
     }
 
     return $url;
--- a/OpenQA/Isotovideo/NeedleDownloader.pm
+++ b/OpenQA/Isotovideo/NeedleDownloader.pm
@@ -60,6 +60,7 @@ has openqa_url        => sub {
     return $url;
 };
 has ua => sub { Mojo::UserAgent->new };
+has download_limit => 150;
 
 sub _add_download {
     my ($self, $needle, $extension, $path_param) = @_;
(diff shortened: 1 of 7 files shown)


Martchus:
BTW, this is supposed to work like this:

1. On resume, the liveviewhandler checks for needles which have been created/updated since the running test has been started. That list is attached to the resume command passt to os-autoinst.
2. os-autoinst will go through the list and download missing/outdated needles before resuming. The download is done in isotovideo. The download is done though new web API routes (*not* served by liveviewhandler).

This approach seems to work in staging where caching is enabled: https://e212.suse.de/tests/3

So I'll keep polishing it. TODOs:

* Fix warnings
* Modularize download and add unit tests
* Prevent downloading SLE needles for an openSUSE test
* Prevent downloading non-existing needles

OpenQA/Isotovideo/NeedleDownloader.pm
```diff
@@ -0,0 +1,105 @@
...
+use Mojo::File;
+use File::stat;
+use Try::Tiny;
+use POSIX 'strftime';
+use testapi 'diag';
+use bmwqemu;
+
+has files_to_download => sub { [] };
+has needle_dir        => sub { needle::default_needle_dir() };
+has openqa_url        => sub { 'http://' . $bmwqemu::vars{OPENQA_URL} };
```
okurz: As I understood the variable "OPENQA_URL" normally includes the prefix, e.g. see https://openqa.opensuse.org/tests/758251/file/vars.json . Maybe better use an URL parsing function which I am sure we have used somewhere else already.

OpenQA/Isotovideo/NeedleDownloader.pm
```diff
@@ -0,0 +1,105 @@
...
+    my $needle_name     = $needle->{name};
+    my $latest_update   = $needle->{t_updated};
+    my $needle_dir      = $self->needle_dir;
+    my $download_target = "$needle_dir/$needle_name.$extension";
+
+    if (my $target_stat = stat($download_target)) {
+        if (my $target_last_modified = $target_stat->[9] // $target_stat->[8]) {
+            $target_last_modified = strftime('%Y-%m-%dT%H:%M:%SZ', gmtime($target_last_modified));
+            if ($target_last_modified ge $latest_update) {
+                diag("skip downloading new needle: $download_target seems already up-to-date (last update: $target_last_modified > $latest_update)");
```
okurz: I see imperative tense in logging messages as misleading. -> s/skip/Skipping/

OpenQA/Isotovideo/NeedleDownloader.pm
Martchus: Ok - than the name of the variable would actually make sense. Strange that in my case the 'URL' was actually just the hostname. I'll improve that to handle both cases. We usually use `Mojo::URL` for this.

OpenQA/Isotovideo/NeedleDownloader.pm
Martchus: Right, I'll change that.

Martchus:
Not sure how to make it ignore unneeded needles with reasonable effort. But just downloading any new or updated needle since the test has been started is not too wasteful, isn't it?

coolo:
I don't think so. Just add a sanity max - to avoid something going busted, e.g. at max 100

Martchus:
Ok, that should be simple enough.

I also suppose it isn't the best idea to do the blocking downloads in isotovideo which will be unresponsive in that time. Do you the download is better done in autotest?

coolo:
autotest is blocked in check_screen - hooking a download into that sounds like a spaghetti code nightmare. Possibly easiest is really a temporary child process of isotovideo. But that's just brain dump - do your own experiments.

Martchus:
> autotest is blocked in check_screen - hooking a download into that sounds like a spaghetti code nightmare.

I just checked out the code again. I suppose moving the download to autotest would be as simple as moving the download [before the reload is triggered](https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L340). Only disadvantage is that it wouldn't work if the test is paused for another reason than check/assert screen timeout. (Btw, in this case the needle reload isn't triggered so far.)

Martchus:
After adding limit and doing the download in autotest it still works on staging: https://e212.suse.de/tests/9

Martchus:
Rebased and fixed perlcritic error. Let's see whether the tests pass now.

okurz:
Broke the build on all repos in [devel:openQA](https://build.opensuse.org/package/show/devel:openQA/os-autoinst)

* https://build.opensuse.org/request/show/640378 adds the new dependency to os-autoinst.spec
* https://build.opensuse.org/request/show/640381 and https://build.opensuse.org/request/show/640382 adds perl-File-Touch to the according SLE subprojects of devel:openQA

PR #1028, opened 2018-09-21
Fix pseudo encoding -261

coolo:
qemu supports 'led state' through this pseudo encoding, but if we claim
we support -261 other vnc servers like tigervnc send us wildly encoded
PNG because they don't support led state (or if they do under a different
pseudo encoding)

As we only supported it to get some debug in the log (hoping to notice
kernel panics), we can easily remove it without loosing functionality
and tigervnc works again

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -945,7 +945,7 @@ sub _receive_update {
         #bmwqemu::diag "UP $x,$y $w x $h $encoding_type";
 
         # work around buggy addrlink VNC
-        next if ($w * $h == 0);
+        next if $encoding_type > 0 && $w * $h == 0;
 
         my $bytes_per_pixel = $self->_bpp / 8;
 


coolo:
See https://progress.opensuse.org/issues/39503

okurz:
I thought we used it for the "stall detection"?

coolo:
You were wrong

coolo:
It was implemented in 53aa6c5f1e41c8e00f6c7bf5d1f7dbe44ed3bd65 to workaround qemu bug - hopefully this won't bite us :(

coolo:
This is still current qemu code, so we might actually regress

```
  /* Turn off the lock state sync logic if the client support the led
       state extension.
    */
    if (down && vs->vd->lock_key_sync &&
        !vnc_has_feature(vs, VNC_FEATURE_LED_STATE) &&
        keycode_is_keypad(vs->vd->kbd_layout, keycode)) {
        /* If the numlock state needs to change then simulate an additional
           keypress before sending this one.  This will happen if the user
           toggles numlock away from the VNC window.
        */
        if (keysym_is_numlock(vs->vd->kbd_layout, sym & 0xFFFF)) {
            if (!vs->modifiers_state[0x45]) {
                trace_vnc_key_sync_numlock(true);
                vs->modifiers_state[0x45] = 1;
                press_key(vs, 0xff7f);
            }
        } else {
            if (vs->modifiers_state[0x45]) {
                trace_vnc_key_sync_numlock(false);
                vs->modifiers_state[0x45] = 0;
                press_key(vs, 0xff7f);
            }
        }
    }

**Comment by coolo** [comment]
And https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#id94 lists it as led state - so tigervnc must get something wrong (or just doesn't care)

**Comment by coolo** [comment]
From tigervnc changelog:
  * Alternative, "raw" keyboard mode in the native client and all servers
  * CapsLock/NumLock/ScrollLock synchronisation in the native client and all servers

**Comment by coolo** [comment]
So it's not wild PNGs as I feared
```
+  const int pseudoEncodingLEDState = -261;
```

coolo:
So better not remove it - but debug and fix it: done in ef6d671

okurz:
awesome :)

PR #1029, opened 2018-09-21
Use here tag in script_out

cfconrad:
Use the "cat > FILE << 'HERE_TAG'" approach to write the script.
The here tag is build from 'EOT_' plus the marker, this
should make collisions less possible with the
content of the script itself.

Related ticket: https://progress.opensuse.org/issues/40913
Testrun: http://cfconrad-vm.qa.suse.de/tests/1008

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -213,8 +213,8 @@ sub script_output {
         testapi::wait_serial($self->{serial_term_prompt}, undef, 0, no_regex => 1);
         testapi::type_string($cat . "\n");
         testapi::wait_serial("$cat", undef, 0, no_regex => 1);
-        testapi::type_string($script);
-        testapi::type_string("\n$heretag\n");
+        testapi::type_string("$script\n$heretag\n");
+        testapi::wait_serial("> $heretag", undef, 0, no_regex => 1);
         testapi::wait_serial("$marker-0-");
     }
     elsif ($args{type_command}) {
--- a/distribution.pm
+++ b/distribution.pm
@@ -208,12 +208,13 @@ sub script_output {
     }
 
     if (testapi::is_serial_terminal) {
-        my $cat = "cat - > $script_path; echo $marker-\$?-";
+        my $heretag = 'EOT_' . $marker;
+        my $cat     = "cat > $script_path << '$heretag'; echo $marker-\$?-";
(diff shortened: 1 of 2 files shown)


asmorodskyi:
@richiejp could you please take a look ?

okurz:
Did you try to come up with a test for this?

asmorodskyi:
> Did you try to come up with a test for this?

Testrun: http://cfconrad-vm.qa.suse.de/tests/1008

distribution.pm
```diff
@@ -208,12 +208,13 @@ sub script_output {
...
     if (testapi::is_serial_terminal) {
-        my $cat = "cat - > $script_path; echo $marker-\$?-";
+        my $heretag = 'EOT_' . $marker;
+        my $cat     = "cat > $script_path << '$heretag'; echo $marker-\$?-";
         testapi::wait_serial($self->{serial_term_prompt}, undef, 0, no_regex => 1);
         testapi::type_string($cat . "\n");
         testapi::wait_serial("$cat", undef, 0, no_regex => 1);
         testapi::type_string($script);
-        testapi::type_string("\n", terminate_with => 'EOT');
+        testapi::type_string("\n$heretag\n");
```
richiejp: Please combine this with the line above and use `wait_serial` to suck up the echo (unless there is some reason we can't do that?). Then we are not risking matching some random occurrence of `$marker-0-`.

richiejp: What you could do is create a regex out of the script text, but where newlines are substituted with `[>\s]+` or whatever the correct sequence is.

okurz:
> > Did you try to come up with a test for this?
> 
> Testrun: http://cfconrad-vm.qa.suse.de/tests/1008

Sorry, not what I meant. I meant os-autoinst tests, add some code under `t/`

richiejp:
LGTM.

cfconrad:
> 
> Sorry, not what I meant. I meant os-autoinst tests, add some code under `t/`

Changed the unit test for script_output to cover as much as possible. Didn't try to cover explicit
script creation implementation, as I think its an internal of that function and not related to the API.
Maybe this could be done in integration tests of openQA.

Thx @rwx788 for help!

okurz:
@cfconrad test looks cool, thx :+1:

PR #1030, opened 2018-09-24
testapi: Relax regex in script_output filter out what to return

asmorodskyi:
Initially script_output was designed to return output only
from commands which pass ( return zero ). Later proceed_on_failure flag was introduced
This patch fixing last place which not allow to return
command output when command finished with non-zero result

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -259,7 +259,7 @@ sub script_output {
     }
 
     # and the markers including internal exit catcher
-    my $out = $output =~ /$marker(?<expected_output>.+)SCRIPT_FINISHED$marker-0-/s ? $+ : '';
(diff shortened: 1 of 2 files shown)


asmorodskyi:
@Martchus it was actually both ( I had extra backslash which needs to be removed and also one test expected to return empty string when it is actually what I am trying to fix by this change )

asmorodskyi:
prove run with patch applied on worler - http://kimball.arch.suse.de/tests/1763/file/autoinst-log.txt ( search for "!!!!!!!" you will see output of command which returned non-zero exit code previosuly

asmorodskyi:
@cfconrad currently we have `script_run` which returns exit code and `script_output` which returns output of command. I think it is sane distinction. In my personal experience I never face situation when I need both usually one of them is good enough to do judgement what happened and how to react.

PR #1031, opened 2018-09-24
Set script when using extra test results

cfconrad:
When using parse_extra_log(), the "test-module-links" in details
ui view pointing to 404.
This happens cause the Parser do not know to which test-module
the extra-log corresponding to.
Patch the script value, when the extra-test is added to the
basetest.

Testrun: http://cfconrad-vm.qa.suse.de/tests/1558

### DIFF
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -135,4 +135,36 @@ subtest record_testresult => sub {
     is(scalar @{$basetest->{details}}, 8, '8 details added');
 };
 
+subtest 'register_extra_test_results' => sub {
+    my $test = basetest->new('foo');
+    $test->{script} = '/tests/foo/bar.pm';
+
+    my $extra_tests = [
+        {
+            category => 'foo',
+            name     => 'extra1',
+            flags    => {},
+            script   => 'unk'
+        },
+        {
+            category => 'foo',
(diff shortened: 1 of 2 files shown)


okurz:
wow, impressive :)

cfconrad:
I thought, it might be handy to fill the script in basetest, cause if someone calls `register_extra_test_results()` with results created by someone different then a Parser, it's handled too.

t/17-basetest.t
```diff
@@ -133,6 +133,38 @@ subtest record_testresult => sub {
 
     is($basetest->{test_count},        8, 'test_count accumulated');
     is(scalar @{$basetest->{details}}, 8, '8 details added');
+}
```
mudler: > Semicolon seems to be missing at ./17-basetest.t line 137.
> "my" variable $test masks earlier declaration in same statement at ./17-basetest.t line 140.
> syntax error at ./17-basetest.t line 138, near "subtest "
> syntax error at ./17-basetest.t line 168, near "}"
> Execution of ./17-basetest.t aborted due to compilation errors.

```;``` is missing here

cfconrad: oh, sorry. Thx for pointing that out!

PR #1032, opened 2018-09-28
Avoid thread crashes

coolo:


coolo:
thx

backend/driver.pm
```diff
@@ -72,8 +72,23 @@ sub start {
             require cv;
 
             cv::init();
+
+            # opencv forks a lot of threads and the TERM signal we
+            # may get from the parent process would be delivered
+            # to an undefined thread. But as those threads do not
+            # have a perl interpreter, the perl signal handler
```
coolo: it's still happening :(

```
#0  0x000000000048f9be in Perl_csighandler ()
#1  <signal handler called>
#2  0x00007f6dcdcba0cf in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#3  0x00007f6dc833574d in ?? () from /usr/lib64/libopencv_core.so.3.1
#4  0x00007f6dcdcb5724 in start_thread () from /lib64/libpthread.so.0
#5  0x00007f6dcd9f5c1d in clone () from /lib64/libc.so.6

AdamWill:
For the record, we see something similar in Fedora (traceback doesn't look identical to the one @coolo posted but it runs through libtbb, which makes opencv the obvious suspect): https://bugzilla.redhat.com/show_bug.cgi?id=1667163

PR #1034, opened 2018-09-28
Add test flag to revert to last good when test module didn't fail

rwx788:
Documentation update is [here](https://github.com/os-autoinst/openQA/pull/1812).

Here is local VR, all tests passed, [here](http://g226.suse.de/tests/2657#step/zypper_lr_validate/1) we have created file, and [here](http://g226.suse.de/tests/2657#step/openvswitch/2) tested that it doesn't exist.

See [poo#30775](https://progress.opensuse.org/issues/30775).

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -354,12 +354,16 @@ sub runalltests {
                     bmwqemu::stop_vm();
                     return 0;
                 }
-                elsif (!$flags->{norollback} && $last_milestone) {
+                elsif (!$flags->{no_rollback} && $last_milestone) {
                     load_snapshot('lastgood');
                     $last_milestone->rollback_activated_consoles();
                 }
             }
             else {
+                if (!$flags->{no_rollback} && $last_milestone && $flags->{always_rollback}) {
+                    load_snapshot('lastgood');
+                    $last_milestone->rollback_activated_consoles();
+                }
                 if ($snapshots_supported && ($flags->{milestone} || $bmwqemu::vars{TESTDEBUG})) {
                     make_snapshot('lastgood');
                     $last_milestone         = $t;
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -86,6 +86,65 @@ is($died,      0, 'start+next+start should not die');
 is($completed, 1, 'start+next+start should complete');
 @sent = [];
 
+# Test loading snapshots with always_rollback flag. Have to put it here, before loading
+# runargs test module, as it fails.
+subtest 'test always_rollback flag' => sub {
+    # Test that no rollback is triggered when flag is not explicitly set to true
+    $mock_basetest->mock(test_flags => sub { return {milestone => 1}; });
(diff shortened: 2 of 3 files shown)


coolo:
we obviously lack an icon for that flag - because it will make reviewing tests quite puzzling

rwx788:
Good point, we also don't have it when tests fails, which also may get puzzling if we have reverted to milestone snapshot, plus it's only for QEMU. Any suggestions for it? E.g. some circled arrow e.g. 
image
?

coolo:
any will do IMO as long as we can differentiate

t/08-autotest.t
```diff
@@ -86,6 +86,35 @@ is($died,      0, 'start+next+start should not die');
...
+($died, $completed) = get_tests_done;
+is($died,      0, 'start+next+start should not die when force_reset flag is set');
+is($completed, 1, 'start+next+start should complete when force_reset flag is set');
+is($reverts_done, 2, "Snapshots are loaded even when tests succeed");
+@sent = [];
+
+# Revert mocks
+$mock_basetest->unmock('test_flags');
+$mock_autotest->unmock('load_snapshot');
+$mock_autotest->unmock('query_isotovideo');
```
okurz: hm, could you put the whole changed section into a subtest? That would also make it more obvious that you mock something and then revert. Basically you would not even have to unmock at the end of a subtest section when you create locally scoped variables. Compare to https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L193

t/08-autotest.t
```diff
@@ -86,6 +86,35 @@ is($died,      0, 'start+next+start should not die');
...
+$reverts_done = 0;
+@sent = [];
+# Test with snapshot available
+$mock_basetest->mock(test_flags => sub { return { force_reset => 1, milestone => 1 }; });
+autotest::run_all;
+($died, $completed) = get_tests_done;
+is($died,      0, 'start+next+start should not die when force_reset flag is set');
+is($completed, 1, 'start+next+start should complete when force_reset flag is set');
+is($reverts_done, 2, "Snapshots are loaded even when tests succeed");
+@sent = [];
```
okurz: hm, do we have a test now that tests that *without* `force_reset` *no* revert is done on success?

autotest.pm
```diff
@@ -360,6 +360,10 @@ sub runalltests {
                 }
             }
             else {
+                if(!$flags->{norollback} && $last_milestone && $flags->{force_reset}) {
```
okurz: Could you please test (locally, manually, once) what happens when you execute this on non-qemu backends?

rwx788: Here we go: http://g226.suse.de/tests/2666#step/openvswitch/4 
So test failed, as file was still there (no revert done), no error in the logs.

coolo:
please tidy so we can move forward with this change

coolo:
thx

PR #1035, opened 2018-09-30
Reset serial console to ttyAMA0 for aarch64 test

waynechen55:
Reset serial console to ttyAMA0 for aarch64 test

    Related ticket: Virtualization Test
    Needles: n/a
    Verification run: http://10.162.187.154/tests/78

@alice-suse @XGWang0 @xguo @Julie-CAO

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -125,6 +125,7 @@ sub init {
         $serialdev = 'ttyS0';
     }
     $serialdev = 'ttyS1' if check_var('BACKEND', 'ipmi');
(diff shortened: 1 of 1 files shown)


coolo:
Don't. Set SERIALDEV on the MACHINE

waynechen55:
> Don't. Set SERIALDEV on the MACHINE

I need to do some verification to ensure SERIALDEV on MACHINE always work as the same.

coolo:
I think this way is wrong at least

PR #1038, opened 2018-10-11
Remove dead Code from s390x zkvm backend

DrMullings:
Removed the complete login handling from SshVirtsh.pm since it is now handled in the distribution to be able to use assert_screen for syncing 

Related ticket: https://progress.opensuse.org/issues/41504
See also: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/5962

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -51,7 +51,6 @@ sub activate {
 
     my $hostname = $args->{hostname} || die('we need a hostname to ssh to');
     my $password = $args->{password};
-
     $self->{ssh} = $self->backend->new_ssh_connection(hostname => $hostname, password => $password);
     if ($self->vmm_family eq 'vmware') {
         $self->{sshVMwareServer} = $self->backend->new_ssh_connection(
@@ -62,14 +61,6 @@ sub activate {
     # start Xvnc
     $self->SUPER::activate;
 
-    my $testapi_console = $self->{testapi_console};
-    my $ssh_args        = $self->{args};
-
-    my $command = $self->sshCommand('root', $hostname);
-    $self->callxterm($command, "ssh:$testapi_console");
-    # FIXME: assert_screen('xterm_password');
-    sleep 3;
-    $self->type_string({text => $password . "\n"});
     $self->_init_xml();
 }
 


DrMullings:
@okurz still needs abstraction of the key path and I am unsure if we can remove the second SSH command in sshVirtsh

okurz:
the key authentication approach looks ok-ish but it looks like we did not handle the original problem of "sleep 3".

@mnowaksuse do *you* have any idea why we need to handle what looks like two different approaches for ssh connection establishing?

DrMullings:
@okurz we have one approach via Net::SSH because its too early for openQA to handle this and the other via XTerm so we can see it in openQA if I understood that correctly

mnowaksuse:
Nothing obvious on my mind.

DrMullings:
@Soulofdestiny FYI

consoles/sshXtermVt.pm
```diff
@@ -35,11 +35,14 @@ sub activate {
...
-    my $sshcommand = $self->sshCommand($username, $hostname, $gui);
+    my $key      = "~/.ssh/id_rsa_s390";
+
+    my $sshcommand = $self->sshCommand($username, $hostname, $gui, $key);
     my $serial = $self->{args}->{serial};
 
     $self->callxterm($sshcommand, "ssh:$testapi_console");
-
+    # FIXME we currently have no better way of synchronizing
+    sleep 3;
```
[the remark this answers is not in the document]
DrMullings: Is there anything that prevents us from using assert_screen?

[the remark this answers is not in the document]
mnowaksuse: I once tried to replace such a `sleep` in sshVirsh console with `assert_screen` and it didn't play out as console wasn't initialized yet, the error was something like "json socket blah". Try and see :).

consoles/sshXtermVt.pm
```diff
@@ -35,11 +35,14 @@ sub activate {
     my $hostname = $ssh_args->{hostname} || die('we need a hostname to ssh to');
     my $password = $ssh_args->{password} || $testapi::password;
     my $username = $ssh_args->{username} || 'root';
-    my $sshcommand = $self->sshCommand($username, $hostname, $gui);
+    my $key      = "/home/jora/.ssh/id_rsa_s390";
```
coolo: nope

consoles/sshXtermVt.pm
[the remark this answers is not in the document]
DrMullings: Turned out exactly as mnowak said, the console is not yet initialized

DrMullings:
@foursixnine fyi

consoles/console.pm
```diff
@@ -63,9 +63,17 @@ sub screen {
...
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
+    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no";
+
+    if ($key) {
+        $sshopts = $sshopts . " -i $key";
+    }
+    else {
+        $sshopts .= " -o PubkeyAuthentication=no";
+    }
```
okurz: Please see my further comments in https://github.com/os-autoinst/os-autoinst/pull/1038#pullrequestreview-164239585

consoles/sshVirtsh.pm
```diff
@@ -51,8 +51,8 @@ sub activate {
 
     my $hostname = $args->{hostname} || die('we need a hostname to ssh to');
     my $password = $args->{password};
-
     $self->{ssh} = $self->backend->new_ssh_connection(hostname => $hostname, password => $password);
+
```
okurz: No unrelated whitespace changes please

consoles/sshVirtsh.pm
```diff
@@ -62,14 +62,6 @@ sub activate {
...
     $self->SUPER::activate;
 
-    my $testapi_console = $self->{testapi_console};
-    my $ssh_args        = $self->{args};
-
-    my $command = $self->sshCommand('root', $hostname);
-    $self->callxterm($command, "ssh:$testapi_console");
-    # FIXME: assert_screen('xterm_password');
-    sleep 3;
-    $self->type_string({text => $password . "\n"});
```
okurz: I think this deserves its own PR

DrMullings: It is now its own PR

DrMullings:
Changed the title and the commit message
Please merge in order to avoid breaking stuff by only merging the other PR

foursixnine:
@DrMullings I think that the other pr was prematurely merged, I'm not sure when the next deploy will happen. cc @coolo?

PR #1039, opened 2018-10-11
Ensure main.pm is still used for proper initialization on SCHEDULE

okurz:
For os-autoinst-distri-opensuse the main.pm does the initialization of the
distribution which we need to execute most of the test modules successfully.
When defining the schedule using the test variable SCHEDULE we therefore need
to load the main.pm but ignore all "loadtest" calls in there.

Verification run: http://lord.arch/tests/1886


okurz:
*argh* @Martchus @coolo @mudler could you help me understand how this codecov works and how we could fix it? If you have a hint I am willing to invest some time into it. Thanks in advance.

coolo:
this is not about codecov, this is about devel:cover not reliably reporting subprocesses.

okurz:
I remember a time when coverage was reliable. What happened since then?

coolo:
Tempus fugit.Tempus fugit.

PR #1041, opened 2018-10-16
Add switch on/off api for ssh half-open detection

jerrytt:
Add A switch to half-open detect.

when we can predict ssh-half-open will be happen expected, we switch off detection and restore it after it happened.

- Related ticket: https://progress.opensuse.org/issues/32746
- Needles: None
- Verification run: http://10.67.132.86/tests/353#

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -46,7 +46,7 @@ __PACKAGE__->mk_accessors(
       last_screenshot last_image assert_screen_check
       reference_screenshot assert_screen_tags assert_screen_needles
       assert_screen_deadline assert_screen_fails assert_screen_last_check
-      stall_detected
+      stall_detected ssh_half_open_expected
       ));
 
 sub new {
@@ -238,6 +238,11 @@ sub run_capture_loop {
                     # Very high limits! On a working socket, the maximum hits per 10 seconds will be around 60.
                     # The maximum hits per 10 seconds saw on a half open socket was >100k
                     if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
+                        if ($self->ssh_half_open_expected) {
+                            bmwqemu::diag "expected half-open sockets in ssh";
+                            $self->stop_ssh_serial;
+                            last;
+                        }
                         die "The console isn't responding correctly. Maybe half-open socket?";
                     }
                 }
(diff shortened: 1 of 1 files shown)


okurz:
s/action , restore/action, restore/

mnowaksuse:
Looks like a too big brush to me. Can't you rather `stop_ssh_serial` where it's expected to be unusable (restart, shutdown, ...)?

backend/baseclass.pm
```diff
@@ -238,6 +238,11 @@ sub run_capture_loop {
                     # Very high limits! On a working socket, the maximum hits per 10 seconds will be around 60.
                     # The maximum hits per 10 seconds saw on a half open socket was >100k
                     if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
+                        if ($self->ssh_half_open_expected) {
+                            bmwqemu::diag "expected half-open sockets in ssh";
+                            $self->stop_ssh_serial;
+                            last;
+                        }
```
okurz: Hm, the indendation / complexity is IMHO too high now, too convoluted code. Also, I do not think that any "ssh" specific code should be hidden in here. How about just raising this event as a proper exception and handling it on a higher level in a specific way?

jerrytt: Thanks @okurz ,
This code is in a while loop, Raising this event and handling on higher level may need more carefully .
can you give example ? thanks

jerrytt:
> Looks like a too big brush to me. Can't you rather `stop_ssh_serial` where it's expected to be unusable (restart, shutdown, ...)?
sorry , i don't understand what you mean .
do you mean directly call stop_ssh_serial where restart/shutdown ?

mnowaksuse:
> do you mean directly call stop_ssh_serial where restart/shutdown ?

Yes.

backend/baseclass.pm
okurz: Look for https://github.com/os-autoinst/os-autoinst/search?q=Exception&unscoped_q=Exception , maybe that helps

jerrytt:
Unfortunately no, please read the issue ticket @okurz comments.

jerrytt:
@mitiao  please take care of this issue. 
I may not have time focus on this.

okurz:
I assume with https://github.com/os-autoinst/os-autoinst/pull/1232 we have found a better, competing approach so closing this ticket. If you feel you can provide a better, updated version feel welcome to re-open

PR #1042, opened 2018-10-16
Don't attach screenshot when recording softfailure

Martchus:
The screenshot was attached as result leading to display
issues in openQA (see https://progress.opensuse.org/issues/41510).

So the test API function record_soft_failure will now really
only store the given $reason like record_info with `result =>
'softfail'`.


okurz:
+1

Martchus:
It only fails because of codecov/path check. But as you can see in the diff I adjusted tests accordingly to cover the change.

PR #1046, opened 2018-10-30
Prevent record_testresult to override failed with softfailed

Martchus:
See https://progress.opensuse.org/issues/43100


Martchus:
Only the tests are failing here. I wouldn't wait forever to merge this fix because otherwise we easily miss bugs (see http://lord.arch/tests/1915#step/gnucash/27). So any feedback?

okurz:
LGTM

PR #1047, opened 2018-11-02
[AARCH64 Automation] [WIP] More reliable way to activate sol session

waynechen55:
1.Before activating sol session, bmc needs to be reset and then deactivated. This way can help guarantee active sol session is always available anytime during test run.
2.The while loop here is to make sure the least amount of time elapsed before re-activate sol session following bmc comes back online.
3.New global variable BMC_IP has been defined. If to be used, it should be added into MACHINE.
4.No ARCH dependent code has been added into this change. If needed, it should only be applied to aarch64.

- Related ticket: aarch64 ipmi backend
- Needles: n/a
- Verification run: http://10.162.187.154/tests/349

### DIFF
--- a/consoles/sshXtermIPMI.pm
+++ b/consoles/sshXtermIPMI.pm
@@ -22,6 +22,7 @@ use testapi 'get_required_var';
 require IPC::System::Simple;
 use autodie ':all';
 use File::Which;
+use testapi 'get_var';
 
 sub activate {
     my ($self) = @_;
@@ -31,10 +32,40 @@ sub activate {
 
     my $testapi_console = $self->{testapi_console};
 
-    my @command = $self->backend->ipmi_cmdline;
-    push(@command, qw(sol activate));
+    my $cstr              = "";
+    my @command_reset_tmp = $self->backend->ipmi_cmdline;
+    push(@command_reset_tmp, qw(mc reset cold));
+    my $command_reset = join(' ', @command_reset_tmp);
+
+    my @command_sleep_tmp = "sleep";
+    push(@command_sleep_tmp, qw(15));
+    my $command_sleep = join(' ', @command_sleep_tmp);
+
+    my $bmcipaddr = get_var('BMC_IP');
+    if ($bmcipaddr eq '') {
+        die "BMC IP address is not defined.";
+    }
+    my @command_ping_tmp = "(while true; do /usr/bin/ping -c1";
+    push(@command_ping_tmp, $bmcipaddr);
+    my $command_ping_tmp_part2 = "; if [";
+    push(@command_ping_tmp, $command_ping_tmp_part2);
+    push(@command_ping_tmp, qw($?));
+    my $command_ping_tmp_part3 = "-eq '0' ]; then break; fi; done;)";
+    push(@command_ping_tmp, $command_ping_tmp_part3);
+    my $command_ping = join(' ', @command_ping_tmp);
+
+    my @command_deactivate_tmp = $self->backend->ipmi_cmdline;
+    push(@command_deactivate_tmp, qw(sol deactivate));
+    my $command_deactivate = join(' ', @command_deactivate_tmp);
+
+    my @command_activate_tmp = $self->backend->ipmi_cmdline;
+    push(@command_activate_tmp, qw(sol activate));
+    my $command_activate = join(' ', @command_activate_tmp);
+    $cstr = join(' ; ', $command_reset, $command_sleep, $command_ping, $command_deactivate, $command_activate);
+
+    bmwqemu::diag "ipmicmdline is $cstr";
+
     my $serial = $self->{args}->{serial};
-    my $cstr = join(' ', @command);
     $self->callxterm($cstr, "ipmitool:$testapi_console");
 }
 
--- a/consoles/sshXtermIPMI.pm
+++ b/consoles/sshXtermIPMI.pm
@@ -32,36 +32,53 @@ sub activate {
 
     my $testapi_console = $self->{testapi_console};
 
-    my $cstr              = "";
+    my $cstr = "";
+
+
+    ############## Following snippet will be splitted into separate functions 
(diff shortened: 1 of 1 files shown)


Martchus:
You need to run `tools/tidy`.

waynechen55:
> You need to run `tools/tidy`.

Done

consoles/sshXtermIPMI.pm
```diff
@@ -22,6 +22,7 @@ use testapi 'get_required_var';
 require IPC::System::Simple;
 use autodie ':all';
 use File::Which;
+use testapi 'get_var';
```
okurz: Please add that in line 21. However IMHO using the test API from within the consoles is bad design

consoles/sshXtermIPMI.pm
```diff
@@ -31,11 +32,40 @@ sub activate {
...
-    my @command = $self->backend->ipmi_cmdline;
-    push(@command, qw(sol activate));
-    my $serial = $self->{args}->{serial};
+    my $cstr              = "";
+    my @command_reset_tmp = $self->backend->ipmi_cmdline;
+    push(@command_reset_tmp, qw(mc reset cold));
+    my $command_reset = join(' ', @command_reset_tmp);
+
+    my @command_sleep_tmp = "sleep";
+    push(@command_sleep_tmp, qw(15));
```
okurz: Why do we need a sleep of 15? I assume it is seconds, 15 seconds would be a lot of time wasted

consoles/sshXtermIPMI.pm
```diff
@@ -31,11 +32,40 @@ sub activate {
...
+    my $command_sleep = join(' ', @command_sleep_tmp);
+
+    my $bmcipaddr = get_var('BMC_IP');
```
okurz: Can we not work with hostnames? Why need to work with IPs?

consoles/sshXtermIPMI.pm
```diff
@@ -31,11 +32,40 @@ sub activate {
...
+    if ($bmcipaddr eq '') {
+        die "BMC IP address is not defined.";
+    }
+    my @command_ping_tmp = "(while true; do /usr/bin/ping -c1";
```
okurz: What if the ping never succeeds? How about using a for loop to abort after a certain amount of retries

consoles/sshXtermIPMI.pm
waynechen55: I will think whether this can be avoided.

consoles/sshXtermIPMI.pm
waynechen55: To make sure bmc already offline before starting ping.

consoles/sshXtermIPMI.pm
waynechen55: Yes. There is already IPMI_HOSTNAME. Will change this.

consoles/sshXtermIPMI.pm
waynechen55: Will improve.

consoles/sshXtermIPMI.pm
okurz: Yes, clearly. My question is: Can you ensure that with a proper check, not sleeping?

waynechen55:
Need more time to figure best solution out.

waynechen55:
I will simplify implementation by moving the large patch of code out of activate function step by step.

okurz:
@alice-suse this looks like going in the direction of what we would could use to reset the MC from tests properly

waynechen55:
> @alice-suse this looks like going in the direction of what we would could use to reset the MC from tests properly

https://progress.opensuse.org/issues/36027
I read it through. It seems that we both share something in common although having been tried to solve different problems if say accurately. Virtualization PRJ2 test suites require multiple times of host power reset, during which sol session can be interrupted or lost somehow. The reasons behind this could be dozens, which I think are already covered in your discussion. So I tried to do "mc reset" before each time of host reboot to put "mc" into its original state, which I think can help increase the success rate when test suites switch back to sol console to capture sol session output. We converged in this point literally. As far as the specific implementation is concerned:
1) If you finally do it  jenkin way or openqa-not-related way, we do not have too much in common anymore.
2) If you finally implement the "mc reset" or similar function abstractly in a separate module or class or subroutine, which then is called upon or propagated into other test modules, I can take advantage of that as well. The only difference is I may need to insert it right just before every times of host reboot.

waynechen55:
> @okurz @waynechen55 , Yes, I agree with you both mostly. The purpose is to do 'mc reset', the solutions can be slightly different according to where(backend or test?) and when(at the beginning of every test or every build or specific points like before reboot or when unstability detected) to do it.
> But the, core code will always be "do mc reset and wait it finish successfully or retry within given times".
> 
> Let's try the agreed on solution in poo#36027 first, if it can not achieve our target to be on stable ipmi through the tests, or it hurts stability with so repeated reset, then let's decide the next step.

I think flexibility is necessary, which then can be leveraged by other test suites or extended into other scenarios, whatever tests, builds or specific points. Just like the openQA itself.

alice-suse:
@waynechen55 , the proposal to do mc reset from backend is merged, please refer to https://github.com/os-autoinst/os-autoinst/pull/1078. I guess this can meet your requirement for aarch64 :-)

okurz:
With https://github.com/os-autoinst/os-autoinst/pull/1232 we should have a better approach to handle consoles that should be persistent. The idea mentioned in this PR could be still valid in case of recovering from instabilities but as there has not been any update by you within the last 10 months I assume you do not plan to continue. Please reopen if I am wrong.

waynechen55:
> With #1232 we should have a better approach to handle consoles that should be persistent. The idea mentioned in this PR could be still valid in case of recovering from instabilities but as there has not been any update by you within the last 10 months I assume you do not plan to continue. Please reopen if I am wrong.

You can close. I will re-implement it better if necessary.

PR #1048, opened 2018-11-05
Add serial console for svirt backend

Martchus:
Allows to add an additional serial console (beside the one used for the serial log) when using the svirt backend.

<s>Still WIP, see https://progress.opensuse.org/issues/36457#note-21</s> see latest comments

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -49,6 +49,7 @@ sub do_start_vm {
         'ssh-virtsh',
         {
             hostname => get_required_var('VIRSH_HOSTNAME'),
+            username => get_var('VIRSH_USERNAME'),
             password => get_var('VIRSH_PASSWORD'),
         });
 
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -194,7 +194,27 @@ sub load_snapshot {
     return $rsp;
 }
 
-# Open another ssh connection to grab the serial console.
+sub read_credentials_from_virsh_variables {
+    my ($self) = @_;
+
+    my ($hostname, $username, $password);
+    if (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
+        $hostname = get_required_var('VIRSH_GUEST');
+        $password = get_var('VIRSH_GUEST_PASSWORD');
+    }
+    else {
+        $hostname = get_required_var('VIRSH_HOSTNAME');
+        $username = get_var('VIRSH_USERNAME');
+        $password = get_var('VIRSH_PASSWORD');
+    }
+    return {
+        hostname => $hostname,
+        username => ($username // 'root'),
+        password => $password,
+    };
+}
+
+# opens another SSH connection to grab the serial console for the serial log
 sub start_serial_grab {
     my ($self, $name) = @_;
 
@@ -209,7 +229,8 @@ sub start_serial_grab {
         $hostname = get_required_var('VIRSH_HOSTNAME');
         $password = get_var('VIRSH_PASSWORD');
     }
-    my $chan = $self->start_ssh_serial(hostname => $hostname, password => $password, username => 'root');
+    my $credentials = $self->read_credentials_from_virsh_variables;
+    my $chan        = $self->start_ssh_serial(%$credentials);
     if (check_var('VIRSH_VMM_FAMILY', 'vmware')) {
         # libvirt esx driver does not support `virsh console', so
         # we have to connect to VM's serial port via TCP which is
@@ -225,6 +246,35 @@ sub start_serial_grab {
     else {
         $chan->exec('virsh console ' . $name);
     }
+}
+
+# opens another SSH connection to grab the serial console with the specified port
+sub open_serial_console_via_ssh {
+    my ($self, $name, $port) = @_;
+
+    bmwqemu::diag("Starting SSH connection to connect to libvirt domain $name via serial port $port");
+    my $credentials = $self->read_credentials_from_virsh_variables;
+    my $ssh         = $self->new_ssh_connection(%$credentials);
+    my $chan        = $ssh->channel();
+    die 'No channel found' unless $chan;
+    $chan->blocking(0);
+    $chan->pty('vt100', {echo => 1});
+    $chan->pty_size(1024, 24);
+    $chan->shell();
+    print($chan "PS1='# '\n");
+
+    # note: see comments in start_serial_grab for the special handling of vmware/hyperv
+    if (check_var('VIRSH_VMM_FAMILY', 'vmware')) {
+        $chan->exec('nc ' . get_var('VMWARE_SERVER') . ' ' . $port);
+    }
+    elsif (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
+        $chan->exec('nc ' . get_var('HYPERV_SERVER') . ' ' . $port);
+    }
+    else {
+        $chan->exec("virsh console \"$name\" \"serial$port\"");
+    }
+
+    return ($ssh, $chan);
 }
 
 sub check_socket {
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -51,7 +51,11 @@ sub activate {
 
     my $hostname = $args->{hostname} || die('we need a hostname to ssh to');
     my $password = $args->{password};
-    $self->{ssh} = $self->backend->new_ssh_connection(hostname => $hostname, password => $password);
+    $self->{ssh} = $self->backend->new_ssh_connection(
+        username => $args->{username},
+        hostname => $hostname,
+        password => $password,
+    );
     if ($self->vmm_family eq 'vmware') {
         $self->{sshVMwareServer} = $self->backend->new_ssh_connection(
             hostname => get_required_var('VMWARE_SERVER'),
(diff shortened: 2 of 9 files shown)


backend/svirt.pm
```diff
@@ -226,6 +249,36 @@ sub start_serial_grab {
...
+    $read_serial_script =~ s/\>/\\\>/g;
+
+    bmwqemu::diag("Starting SSH connection to connect to libvirt domain $name via serial port $port");
+    my $credentials = $self->read_credentials_from_virsh_variables;
+    my $ssh = $self->new_ssh_connection(%$credentials);
+    my $chan = $ssh->channel();
+    die 'No channel found' unless $chan;
+    $chan->blocking(0);
+    $chan->pty(1);
+    my $command = "screen \"\$(virsh dumpxml \"$name\" | perl -e \"$read_serial_script\" \"$port\")\"";
```
richiejp: I think `socat` is more appropriate here than screen. We just want the bytes to be transferred from the serial device/socket to the SSH socket.

backend/svirt.pm
```diff
@@ -226,6 +249,36 @@ sub start_serial_grab {
     }
 }
 
+# opens another SSH connection to grab the serial console with the specified port
+sub open_serial_console_via_ssh {
+    my ($self, $name, $port) = @_;
+
+    # FIXME: don't hardcode path
+    my $read_serial_script = Mojo::File->new('/hdd/openqa-devel/repos/os-autoinst/read_serial_dev_from_virsh_xml.pm')->slurp;
```
richiejp: Isn't there a `virsh` command which lists the serial devices for a 'domain'?

Martchus: I looked for such a command in the first place (before writing this Perl script) but didn't find anything - except dumping the XML of course. I'm going to look again... because passing this script via SSH is really messy and so far I don't know a better way.

backend/svirt.pm
richiejp: Infact in QEMU you can expose the serial device as a socket which is accessible directly from the network. I wonder if this is also possible here?

Martchus: I didn't know that tool. I'm going to read the man page. Since this runs on the svirt host, we might need to install it there first. This leads to the question: How are dependencies on those "jump-hosts" handled?

consoles/virtio_screen.pm
```diff
@@ -197,7 +199,7 @@ sub read_until {
     bmwqemu::log_call(%nargs);
 
     my $rin = '';
-    vec($rin, fileno($fd), 1) = 1;
+    vec($rin, fileno($sel_fd), 1) = 1;
```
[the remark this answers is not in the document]
Martchus: Or we just merge https://github.com/os-autoinst/os-autoinst/pull/1012 first. I would know how to resolve the conflict when rebasing.

backend/svirt.pm
richiejp: This is better. However I wonder if you can change the domain config so that the serial device is exposed on the host as a socket instead of a TTY device? Then virtio/serial_screen can connect to it directly with no SSH or virsh.

I think this is probably the way VNC works, libvirt just gets the hypervisor to create a socket and the VNC console connects to it, the same as for QEMU.

consoles/sshVirtsh.pm
```diff
@@ -242,7 +249,9 @@ sub add_vnc {
     $graphics->setAttribute(autoport    => 'no');
     $graphics->setAttribute(listen      => '0.0.0.0');
     $graphics->setAttribute(sharePolicy => 'force-shared');
-    $graphics->setAttribute(passwd      => $testapi::password);
+    if (my $vnc_password = $testapi::password) {
+        $graphics->setAttribute(passwd => $vnc_password);
+    }
```
mnowaksuse: Likely a blank password is set. Anyway, I'd like to see `passwd` attribute go, we don't use it in qemu.

distribution.pm
```diff
@@ -54,6 +54,7 @@ sub add_console {
         'tty-console'     => 'ttyConsole',
         'ssh-xterm'       => 'sshXtermVt',
         'ssh-virtsh'      => 'sshVirtsh',
+        'ssh-virtsh-sut'  => 'sshVirtshSUT',
```
mnowaksuse: I am not sure about the console class name. Currently we use `sut` console via `ssh-virtsh`, and in other backends, and it is about VNC, this one is not.

consoles/virtio_screen.pm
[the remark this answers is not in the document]
richiejp: My concern is that you may have to revert the virtio_screen/console changes which is more complicated if the SSH console is also merged. However it is probably quite a low risk.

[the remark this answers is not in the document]
Martchus: In fact the changes I made here are contained by https://github.com/os-autoinst/os-autoinst/pull/1012. Resolving the conflicts would simply mean to discard all of my changes in `virtio_screen.pm` and use `consoles::serial_screen` instead of `consoles::virtio_screen` in `sshVirtshSUT.pm`.

backend/svirt.pm
Martchus: I changed this now to use `virsh console`.

backend/svirt.pm
Martchus: Ok, but I'd like to go for something that works and maybe improve this later. So I changed this now to use `virsh console` which already works for the serial log.

richiejp: OK, if `virsh console` passes the data through unmolested then this is probably the best solution anyway.

distribution.pm
Martchus: To my understanding `ssh-virtsh` **is** SSH and not VNC. The SSH connection is used to configure VNC among other things, though. So I thought it would make sense to call it `ssh-virtsh-sut` because it is a SSH connection using `virsh console` to provide console access to the SUT.

But I'm not familiar with your conventions. So just tell me how to call it and I rename it (if it make some sort of sense).

Martchus:
It basically seems to work:
screenshot_20181106_130038

I did the following modifications in the test distribution for testing it locally: https://github.com/os-autoinst/os-autoinst-distri-opensuse/compare/master...Martchus:svirt3  
So beside `add_console` and `select_console`, `systemctl start serial-getty\@ttyS1` has to be executed in the testsuite to enable the serial console.

And it turns out I can use `vrish console` after all. So there is no need to deal with the XML anymore after the initial configuration.

I haven't tested the code for handling vmware/hyperv where `virsh console` is supposed to be unavailable. In these cases the serial port for the serial log is specified via the test variables `VMWARE_SERIAL_PORT`/`HYPERV_SERIAL_PORT` by the test writer. The additional serial consoles introduced in this PR use the port passed to `add_console`. So the test writer can/needs to pass the expected port as well.

So until I get some feedback I'll add some documentation about the svirt backend and try to add some unit tests.

mnowaksuse:
Xen on svirt backend uses `virsh console` as well. Do there and on Hyper-V regression testing, please. I am happy to help you setup a suitable workers.

Yes, for Hyper-V & VMware we need additional variables.

distribution.pm
richiejp: maybe `ssh-virtsh-serial`?

mnowaksuse: I like this one.

Martchus:
Do I need to add `XML::SemanticDiff` to the cpan file in the openQA repository, too?

distribution.pm
Martchus: Ok, I'll change it to `ssh-virsh-serial`.

Martchus:
Adding the dependency via openQA worked.

I removed the WIP because it basically works when testing locally and it didn't break other things (did the Xen and Hyper-V validation).

The unit tests test only the console initialization, passing the credentials and generating the XML. To tests the actual console, I needed to mock the SSH connection which would be some effort so I didn't do it. (Say if want me to take the effort.)

There's adaption/integration with the other PRs required. I'll do that when those are merged before this one.

Martchus:
I think we should merge this before https://github.com/os-autoinst/os-autoinst/pull/1012 because it likely takes some time till it is ready. And this PR is ready in the sense I think it won't break existing things and allows me to get feedback. I'm glad to help rebasing https://github.com/os-autoinst/os-autoinst/pull/1012 (has already conflicts anyways).

doc/backends.md
```diff
@@ -0,0 +1,49 @@
+# Backends in openQA
```
okurz: It's nice that you add a new documentation file but I wonder how it is integrated into the documentation how it is visualized on http://open.qa

Martchus: Ok, so I'll move it to openQA.

Martchus: But then I likely need to rewrite it as asciidoc. So can I just leave it for now here? Maybe we can figure out a solution later which also covers the other documentation files in os-autoinst like architecture.md and networking.md.

okurz: Yes, I would also prefer if you keep the documentation in os-autoinst. I would be happy if you can tackle the documentation aggregation.

As a side-note, converting from md to asciidoc should not be a problem, you already know [pandoc](https://pandoc.org/), right? ;)

Martchus: Ok, let's see whether generating the documentation can be improved to include files from os-autoinst, too.

foursixnine: @Martchus you can simply rename it to asciidoc, I think the rendering would be ok, as there's no weird stuff

richiejp:
FYI just in case you are waiting for it @pevik

PR #1050, opened 2018-11-12
svirt/Xen: Expect the image to be located on the svirt host

Martchus:
rather than on the worker

See https://progress.opensuse.org/issues/32281

---

required for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6191

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -320,7 +320,6 @@ sub add_disk {
         my $dir = "/var/lib/libvirt/images";
         die "No file given" unless $args->{file};
         if ($args->{cdrom} or $args->{backingfile}) {
-            die "File $args->{file} not readable" unless -r $args->{file};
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -320,7 +320,7 @@ sub add_disk {
         my $dir = "/var/lib/libvirt/images";
         die "No file given" unless $args->{file};
         if ($args->{cdrom} or $args->{backingfile}) {
-            die "File $args->{file} not readable" unless -r $args->{file};
+            $self->run_cmd('ls ' . $args->{file}) && die "File $args->{file} does not exist on svirt host";
             $self->run_cmd(sprintf("rsync -av '$args->{file}' '${dir}/%s'", basename($args->{file}))) && die 'rsync failed';
```
coolo: Why not just let rsync fail? I don't see any value on one more remote command

Martchus: yes, I'll just remove the line

PR #1051, opened 2018-11-13
qemu: Add QEMU_APPEND option

ggardet:


ggardet:
I testsed it locally with isotovideo, with and without QEMU_APPEND in vars.json

okurz:
+1

PR #1052, opened 2018-11-13
Add option to override status of test modules with soft-fail

okurz:
This allows to mark a failing test as a "known issue" from a
post_fail_hook, for example when the post_fail_hook executes a generic
method to investigate failures.

https://github.com/os-autoinst/openQA/pull/1837 is the according
proposal to extend the actual definition of "soft-fail".

Related progress issue: https://progress.opensuse.org/issues/39719

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -59,7 +59,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   upload_asset data_url check_shutdown assert_shutdown parse_junit_log parse_extra_log upload_logs
 
   wait_idle wait_screen_change assert_screen_change wait_still_screen wait_serial
-  record_soft_failure record_info
+  record_soft_failure record_info force_soft_failure
   become_root x11_start_program ensure_installed eject_cd power
 
   save_memory_dump save_storage_drives freeze_vm resume_vm
@@ -164,11 +164,16 @@ sub save_screenshot {
 
 =head2 record_soft_failure
 
+=for stopwords softfail
+
   record_soft_failure([$reason]);
 
 Record a soft failure on the current test modules result. The result will
 still be counted as a success. Use this to mark where workarounds are applied.
-Takes an optional C<$reason> string which is recorded in the log file.
+Takes an optional C<$reason> string which is recorded in the log file. See
+C<force_soft_failure> to forcefully override a failed test module status from
+a C<post_fail_hook> or C<record_info> when the status should not be
+influenced.
 
 =cut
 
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -164,19 +164,25 @@ sub save_screenshot {
 
 =head2 record_soft_failure
 
-  record_soft_failure([$reason]);
+=for stopwords softfail
+
+  record_soft_failure([$reason] [, force_status => $force_status]);
```
coolo: for boolean values we normally put => 1 in the documentation

testapi.pm
```diff
@@ -164,19 +164,25 @@ sub save_screenshot {
...
 
 Record a soft failure on the current test modules result. The result will
 still be counted as a success. Use this to mark where workarounds are applied.
-Takes an optional C<$reason> string which is recorded in the log file.
+Takes an optional C<$reason> string which is recorded in the log file.  Set
+C<$force_status> to a true value to override the test module status to a
+softfail status even if the module would be set to fail otherwise. This can be
```
coolo: I find the use of this flag a little undermotivated here. Because what you write here makes *only* sense in post_fail_hooks, actually the whole force_status does. Because you need to call it *after* the failure - which can't be in run() as failures are exceptions throwing you out of the run.

And IMO this should be an extra API call - and not interfered into record_soft_failure.

okurz: I see. We already have `record_info` with various options but to my understanding it *only* informs and does not change the status. So how about a third explicit API call `record_known_failure`?

coolo: We're getting into troubled water here - if a soft fail is defined as known failure, what's the difference between record_soft_failure and record_known_failure? None to the reader.

So I wouldn't introduce this other word for the same - but make the reader aware we're doing more than recording in the post fail hook: force_soft_failure

okurz: sure, works for me, updated

testapi.pm
```diff
@@ -213,6 +218,26 @@ sub record_info {
     $autotest::current_test->record_resultfile($title, $output, %nargs);
 }
 
+=head2 force_soft_failure
+
+=for stopwords softfail
+
+  force_soft_failure([$reason]);
+
+Like C<record_soft_failure> but overrides the test module status to a softfail
```
coolo: your wording still sounds like you can pass this before the fail - but it's not true as it only overwrites past failures.

coolo: I'll merge so we can deploy it today, but I would still like you to update the docu about it (or remove it if your experiments fails :)

okurz: https://github.com/os-autoinst/os-autoinst/pull/1062

PR #1055, opened 2018-11-14
qemu: Handle files compressed as *.xz

ggardet:
This allow to test JeOS images from OBS, which are XZ compressed.
Tested with isotovideo with JeOS-efi on aarch64.

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -30,6 +30,8 @@ parameters which are represented as complex objects.
 
 =cut
 package OpenQA::Qemu::Proc;
+use File::Basename;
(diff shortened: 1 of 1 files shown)


AdamWill:
Well I only just noticed it, but - this might actually make the `_DECOMPRESS_URL` functionality in openQA redundant. At least I think with this it's redundant for the use case we wrote it for. I'll have to try it out.

okurz:
gosh, how do you find all these old PRs? ;) But I am happy to hear from you after your experiments

AdamWill:
I just happened to see the compression handling code while I was looking up something else and thought waaaaait a minute :P

PR #1021, opened 2018-09-06
[sshXtermIPMI] Overrides reset function for ipmi console

XGWang0:
Make sure sol connection is deactivated after reset (rest_consoles) operation, which  guarantees it is successful for sol connecting again.

    Related ticket:
    poo#40544

    Verification run:
    test on super-micro machine : http://10.67.19.191/tests/209
    test on dell machine : http://10.67.19.191/tests/201
    test for xen on dell machine: http://10.67.132.86/tests/170

PR #1057, opened 2018-11-15
[remoteVnc] Override reset method for remoteVnc

alvarocarvajald:
Ensure `remoteVnc` connection is disconnected/disabled when `->reset()` is called (reset_consoles), to prevent backend crashes when connected to certain hardware manufacturers.

Crash: http://mango.suse.de/tests/596
Verification Run: http://mango.suse.de/tests/600
Backend: IPMI
Server: Dell PowerEdge R640

This is a follow up to https://github.com/os-autoinst/os-autoinst/pull/1021, which performed a similar fix on `sshXtermIPMI`.

### DIFF
--- a/consoles/remoteVnc.pm
+++ b/consoles/remoteVnc.pm
@@ -38,6 +38,16 @@ sub activate {
         });
 }
 
+sub reset {
+    my ($self) = @_;
+
+    if ($self->{activated}) {
+        $self->SUPER::disable();
+        $self->{activated} = 0;
+    }
+    return;
+}
+
 # override
 sub select {
 }


consoles/remoteVnc.pm
```diff
@@ -38,6 +38,16 @@ sub activate {
         });
 }
 
+sub reset {
+    my ($self) = @_;
+
+    if ($self->{activated}) {
+        $self->SUPER::disable();
```
coolo: but if this code is a thing now in console, the base class should call a deactivate function - which is then different in ipmi and vnc.

And it makes you wonder if this shouldn't be in vnc_base if it's important

coolo:
aehm - which of your consoles in that test run is actually remotevnc?

alvarocarvajald:
> aehm - which of your consoles in that test run is actually remotevnc?

This one:

```
[2018-11-15T17:54:35.0936 CET] [debug] <<< testapi::select_console(testapi_console='x11')
/usr/lib/os-autoinst/consoles/vnc_base.pm:71:{
  'hostname' => 'holmes.qa.suse.de',
  'password' => 'nots3cr3t',
  'port' => 5901
}
```

In the failing test, the backend is dying with SIGSEGV after a couple of calls to `select_console` and `assert_screen` on that console after reboot, which is pretty much with what was happening with `sshXtermIPMI` before https://github.com/os-autoinst/os-autoinst/pull/1021.

consoles/remoteVnc.pm
alvarocarvajald: Should I move this to `vnc_base` then? I really haven't seen problems with the other consoles, but don't think this will do harm.

coolo:
the 'x11' console is defined as:

```
       $self->add_console(
            'x11',
            'vnc-base',
            {
                hostname => $hostname,
                port     => 5901,
                password => $testapi::password
            });

```

So whatever you did to removeVnc shouldn't matter much. Especially as remoteVNC console is about PARM file - and I doubt you have one. sshXtermIPMI only died with SIGSEGV for you. And I don't think a SOL disable fixes that either.

alvarocarvajald:
Closing this to do more research on the issue.

Had a successful test when moving the `->reset()` code to `vnc_base` as suggested by @coolo  (http://mango.suse.de/tests/610), but then removed the method from both `vnc_base` and from `remoteVnc` and its also working (http://mango.suse.de/tests/612). In both cases `/dev/sshserial` is not working after reboot.

Sorry for the noise.

PR #1058, opened 2018-11-16
Include full path in category name

Martchus:
Eg. a file with the path `tests/sle/tests/x11/toolkits/xterm.pm` will be
put into category `x11/toolkits` instead of just `toolkits`.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -98,11 +98,12 @@ sub loadtest {
     my ($script, %args) = @_;
     my $casedir     = $bmwqemu::vars{CASEDIR};
     my $script_path = find_script($script);
-    unless ($script_path =~ m,(\w+)/([^/]+)\.pm$,) {
-        die "loadtest: script path '$script_path' does not match required pattern \\w.+/[^/]+.pm\n";
+    my $pattern     = qr,(tests/[^/]+/)?tests/([\w/]+)/([^/]+)\.pm$,;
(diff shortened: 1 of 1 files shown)


Martchus:
Testrun which would make use of this change: http://artemis.suse.de/tests/898

@asdil12 If you test it, you can send a link to a successful testrun.

asdil12:
It breaks tests with manually scheduled assets:
```
[2018-11-16T14:42:52.290 CET] [debug] Found override test module for sysrq.pm: /home/dheidler/.local/lib/openqa/share/factory/other/sysrq.pm
loadtest: script path '../../factory/other/sysrq.pm' does not match required pattern (?^:(tests/[^/]+/)?tests/([\w/]+)/([^/]+)\.pm$)
12032: EXIT 1
```

Test was scheduled by:

```
openqa_clone_job_artemis_from_osd --skip-chained-deps 2261417 TEST=toolkits SCHEDULE=tests/boot/boot_to_desktop,tests/console/consoletest_setup,tests/x11/toolkits/xterm,sysrq ASSET_1_URL=https://w3.suse.de/~okurz/sysrq.pm
```

This tests automatically put the additional test modules in the category `other`.

Martchus:
Ok - this would have been to easy otherwise :-)

So we need to tweak the regex a little bit further or maybe use 2 different ones for the different cases.

PR #1060, opened 2018-11-19
qemu: Enhance QEMU_APPEND option to handle multiple options

ggardet:
with or without params

Tested locally with os-autoinst with:
* no QEMU_APPEND in vars.json
* QEMU_APPEND with a single option
* QEMU_APPEND with 2 options: 1 without param, 1 with a single param

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -851,7 +851,13 @@ sub start_qemu {
 
     # Add parameters from QEMU_APPEND var, if any.
     # The first item will have '-' prepended to it.
-    sp("$vars->{QEMU_APPEND}") if $vars->{QEMU_APPEND};
+    if ($vars->{QEMU_APPEND}) {
+        # Split multiple options, if needed
+        my @spl = split(' -', $vars->{QEMU_APPEND});
+        foreach my $i (@spl) {
+            sp(split(' ', $i));
+        }
+    }
 
     $self->{qemupipe}  = $self->{proc}->exec_qemu();
     $self->{qmpsocket} = $self->{proc}->connect_qmp();


backend/qemu.pm
```diff
@@ -851,7 +851,13 @@ sub start_qemu {
 
     # Add parameters from QEMU_APPEND var, if any.
     # The first item will have '-' prepended to it.
-    sp("$vars->{QEMU_APPEND}") if $vars->{QEMU_APPEND};
+    if ($vars->{QEMU_APPEND}) {
+        # Split multiple options, if needed
+        my @spl = split(' -', $vars->{QEMU_APPEND});
+        foreach my $i (@spl) {
+            sp(split(' ', $i));
+        }
```
okurz: would it work to apply the map function here? Something like:

```perl
map { sp(' ', $_) } split(' -', $vars->{QEMU_APPEND}) if $vars->{QEMU_APPEND};
```

ggardet: @okurz No, since there are multiple splits. And IMO, the 1st solution is better for understanding.
For the unit test, could you point me to existing file(s) or example(s), please?

Martchus: You needed to use 2nd split inside the map. But I would just keep the loop for better readability.

For unit tests there's already `t/18-qemu.t`. Not sure whether we already execute this `start_qemu` inside a unit test. If not, it might be easier to move your actual code into a separate function and just test that.

ggardet: I am looking at adding test, but it will take time to finalize.
@okurz As aarch64 needs this patch, could you please merge this one now, and wait for another PR for the test, please?

ggardet: Tests added

foursixnine: @ggardet Please remove the tests from the fullstack as it would increase a bit the time for the tests to run. I guess the right place to add unit tests would be https://github.com/os-autoinst/os-autoinst/blob/master/t/18-qemu.t, maybe @richiejp can shed some light here... since the sp() method may not be that easy to test.

foursixnine:
@ggardet please see my comment about the changes in the fullstacktest

okurz:
I'm sorry, I actually do not have permissions to merge in this repo even if I wanted to

backend/qemu.pm
ggardet: @foursixnine I tried to add it to 18-qemu.t, but I am not able to check output values. And we need to do_start_vm to check it, anyway.
Moreover, there are 3 tests out of 4 which are very fast since they just output qemu version, list qemu machines and test an invalid option, so does not start a full real test.
Just tell me if I need to remove the tests to get it committed or if it is fine like this.

ggardet:
> @ggardet please see my comment about the changes in the fullstacktest, You also ned to run tidy :)

@foursixnine See my answer ;) And I run tidy.

richiejp:
This won't actually append the command line items, it will insert them after the other static params, but before the dynamic parameters (i.e. block devices and controllers, etc.). This is probably what you want, but the name is a bit misleading.

Also this feature should come with a warning because if you use this on a permanent basis then your instance is more likely to be broken by changes to the default set of parameters and parameter order. Otherwise it looks good to me.

ggardet:
> This won't actually append the command line items, it will insert them after the other static params, but before the dynamic parameters (i.e. block devices and controllers, etc.). This is probably what you want, but the name is a bit misleading.

Yes, but the name is already there ;) This is just an update/fix.
> 
> Also this feature should come with a warning because if you use this on a permanent basis then your instance is more likely to be broken by changes to the default set of parameters and parameter order. 

IMO, this will be used for:
a) testing new params easily
b) make use of special config, such as on aarch64 with hugepages config
So, breakage is unlikely to me.

> Otherwise it looks good to me.

Ok, thanks.

foursixnine:
LGTM

ggardet:
> Temporary fixes and workarounds are OK, but any permanent config I would assume is useful to other people until proven otherwise.

I agree. For now we will test the hugepages option and if it turns to be useful and needed permanently, I will submit a PR to add a new QEMU option to enable hugepages easily.

backend/qemu.pm
okurz: I understand that you actually want to start qemu to make it work but how much longer does it actually take? the full stack test has a problem that it already runs for quite some time without good feedback to the waiting developer what is going on so what I think would be good if we can achieve is that we have the tests you added in a separate test and as fast as possible, e.g. just startup qemu, check that it does not fail on invalid command line parameters and then exit.

ggardet: @okurz , I can move it to another file, and maybe remove the real run to be faster if really needed. 
Any suggestion for the new file name? XX-qemu-options.t? 
I would like to get it in ASAP. We can tune things afterwards, if needed.

okurz:
Yes sounds good, also the name for the file

backend/qemu.pm
ggardet: All comments addressed in latest push.

okurz:
Looks like this caused some build failures, e.g. see in https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/SLE_15/x86_64

```
[  295s] failed to start VM at /home/abuild/rpmbuild/BUILD/os-autoinst-4.5.1543566133.1c333b52/backend/driver.pm line 134.
[  295s] 6344: EXIT 1
[  295s] Deep recursion on subroutine "B::Deparse::_pessimise_walk_exe" at /usr/lib/perl5/5.26.1/B/Deparse.pm line 439.
[  295s]     not ok 12 - -broken option added
[  295s]     #   Failed test '-broken option added'
[  295s]     #   at ./18-qemu-options.t line 148.
[  295s]     #          got: '256'
[  295s]     #     expected: '0'
[  295s]     not ok 13 - invalid option detected
[  295s]     #   Failed test 'invalid option detected'
[  295s]     #   at ./18-qemu-options.t line 149.
[  295s]     #          got: '256'
[  295s]     #     expected: '0'
[  295s]     1..13
[  295s]     # Looks like you failed 10 tests of 13.
[  295s] not ok 1 - qemu_append_option
[  295s] #   Failed test 'qemu_append_option'
[  295s] #   at ./18-qemu-options.t line 151.
```

ggardet:
It passed just fine on my x86_64 machine. Looking.

ggardet:
@okurz, the problem is related to  ```no Qemu/KVM found``` error message in OBS.
I think OBS worker try to use ```qemu-kvm``` bin which will fails anyway. It should rather use ```qemu-system-x86_64``` which will handle QEMU_NO_KVM flag properly.

ggardet:
Ok, I got it. See os-autoinst.spec , you have:
```
# don't require qemu within OBS
cp t/05-pod.t t/99-full-stack.t
```
So, the same thing must be done in spec file for this test!

PR #1061, opened 2018-11-19
Prevent cmd srv being busy in loop consuming isotovideo messages

Martchus:
See https://progress.opensuse.org/issues/43949

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -306,7 +306,6 @@ sub run_daemon {
     # process json messages from isotovideo
     Mojo::IOLoop->singleton->reactor->io($isotovideo => sub {
             my ($reactor, $writable) = @_;
-            return if ($writable);
 
             my $isotovideo_response = myjsonrpc::read_json($isotovideo);
             my $clients             = app->defaults('clients');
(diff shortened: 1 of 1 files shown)


commands.pm
```diff
@@ -306,7 +306,7 @@ sub run_daemon {
     # process json messages from isotovideo
     Mojo::IOLoop->singleton->reactor->io($isotovideo => sub {
             my ($reactor, $writable) = @_;
-            return if ($writable);
+            return $reactor->watch($isotovideo, 1, 0) if ($writable);    # only watch for reading
```
kraih: While this would work just fine, it might be slightly more efficient to add a `->watch($isotovideo, 1, 0);` below (like `->reactor->io($isotovideo => sub {...})->watch($isotovideo, 1, 0);`).

Martchus: Ok, and I suppose *below* is the keyword here. Because I tried to execute it at the top and it didn't work. That wasn't a surprise because the documentation says 'Note that this method requires an active I/O watcher'. I just figured that 'active I/O watcher' then must mean 'at any time after the loop has been started with IO callbacks registered'. Maybe the documentation should explain what 'active I/O watcher' means.

kraih: Perhaps, "for the handle you're passing to this method" is kind of implied.

Martchus:
@kraih Changed it in the way you proposed. (And tested again, seems to work.)

PR #1062, opened 2018-11-20
basetest: Fix force_soft_fail being able to override the module status

okurz:
Verification run: http://lord.arch/tests/1944#step/force_soft_failure/6

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -281,12 +281,6 @@ sub done {
     autotest::set_current_test(undef);
 }
 
-sub fail_if_running {
-    my $self = shift;
-    $self->{result} = 'fail' if $self->{result};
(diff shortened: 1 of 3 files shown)


doc/backend_vars.asciidoc
```diff
@@ -1,36 +1,71 @@
 Supported variables per backend
 -------------------------------
 
+.AMT backend
```
coolo: why?

okurz: https://progress.opensuse.org/issues/34303 is the reason

okurz: updated PR

Martchus:
14-isotovideo.t and 99-full-stack.t are failing.

PR #1065, opened 2018-11-22
[WIP] Add 'no_wait' option to 'send_key_until_needlematch'

ldevulder:
An issue on aarch64 uefi boot led to me to change a `send_key_until_needlematch` by a `check_screen` loop with `send_key` and `no_wait` options added to `check_screen`.

At the end it's better to directly add support of `no_wait` inside `send_key_until_needlematch`.

This was discussed in this PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4761

Ticket: https://progress.opensuse.org/issues/34501

Verification run: http://1b147.qa.suse.de/tests/4124#step/uefi_bootmenu/10 (test fails for another reason, uefi_bootmenu using UEFI bootmanager is working well).

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1144,22 +1144,27 @@ sub release_key {
 
 =head2 send_key_until_needlematch
 
-  send_key_until_needlematch($tag, $key [, $counter, $timeout]);
+  send_key_until_needlematch($tag, $key [, [$counter] | [timeout => $timeout]]);
 
 Send specific key until needle with C<$tag> is not matched or C<$counter> is 0.
-C<$tag> can be string or C<ARRAYREF> (C<['tag1', 'tag2']>)
-Default counter is 20 steps, default timeout is 1s
+C<$tag> can be string or C<ARRAYREF> (C<['tag1', 'tag2']>).
+Default counter is 20 steps, default timeout is 1s.
+
+As this function uses check_screen, all extra arguments are passed as-is to it.
 
 Throws C<FailedNeedle> exception if needle is not matched until C<$counter> is 0.
 
 =cut
 
 sub send_key_until_needlematch {
-    my ($tag, $key, $counter, $timeout) = @_;
+    my $tag     = shift;
+    my $key     = shift;
+    my $counter = shift;
+    $counter = 20 unless looks_like_number($counter);
+    my $timeout = looks_like_number($_[0]) ? shift : 1;
+    my %args = (timeout => $timeout, @_);
 
-    $counter //= 20;
-    $timeout //= 1;
-    while (!check_screen($tag, $timeout)) {
+    while (!check_screen($tag, %args)) {
         wait_screen_change {
             send_key $key;
         };
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -451,6 +451,19 @@ subtest 'check_assert_shutdown' => sub {
 
 };
 
+subtest 'send_key_until_needlematch' => sub {
+    $mod->mock(
+        read_json => sub {
+            return {ret => {sim => 999}};
+        });
+    $fake_needle_found = 1;
+    ok(send_key_until_needlematch('foo', 'spc'), 'simple test');
+    ok(send_key_until_needlematch('foo', 'spc', 2). 'counter specified');
+    ok(send_key_until_needlematch('foo', 'spc', 2, 5). 'counter and old-fashion timeout specified');
+    ok(send_key_until_needlematch('foo', 'spc', 2, timeout => 10). 'counter and new-fashion timeout specified');
+    ok(!send_key_until_needlematch('foo', 'spc', 2, no_wait => 1). 'check_screen option specified');
+};
+
 done_testing;
 
 # vim: set sw=4 et:


testapi.pm
```diff
@@ -1144,22 +1144,28 @@ sub release_key {
...
+screen which can change in a range faster than 1-2 seconds not to miss the
+screen to check for.
 
 Throws C<FailedNeedle> exception if needle is not matched until C<$counter> is 0.
 
 =cut
 
 sub send_key_until_needlematch {
-    my ($tag, $key, $counter, $timeout) = @_;
+    my ($tag, $key, $counter, $timeout, $no_wait) = @_;
```
okurz: Actually what you could do is just consume all additional parameters inside %args including $timeout and forward that one to the internal `check_screen` call

Martchus: But wouldn't that suggestion break the existing API? E.g. `send_key_until_needlematch('foo', 'bar', 5, 20)` would then need to be adjusted to `send_key_until_needlematch('foo', 'bar', counter => 5, timeout => 20)`.

okurz:
Well, then don't require named parameters

testapi.pm
ldevulder: I think that compatibility can be kept like in `check_screen` or `assert_screen`. I will try this.

ldevulder: I change the code to be able to use `%args` syntax with additional parameters and this code also keep the old API.

I tested both calls in this run (test failure is not because of this change): http://1b147.qa.suse.de/tests/4105

testapi.pm
```diff
@@ -1144,22 +1144,30 @@ sub release_key {
...
 
 =cut
 
 sub send_key_until_needlematch {
-    my ($tag, $key, $counter, $timeout) = @_;
+    my $tag     = shift;
+    my $key     = shift;
+    my $counter = looks_like_number($_[0]) ? shift : 20;
+    my $timeout = looks_like_number($_[0]) ? shift : 1;
+    my %args    = (counter => $counter, timeout => $timeout, @_);
```
alvarocarvajald: I like the approach, but the changes made to the parsing of the method arguments could lead it to set a timeout of 1  if the method is called as `send_key_until_needlematch('mytag', 'mykey', counter => 5, 34)` (which granted, is not in use right now, but could be valid per the comment in L1147). I would recommend to either change the comment, or make the argument parsing more thorough.

testapi.pm
```diff
@@ -1144,22 +1144,30 @@ sub release_key {
...
 
-    $counter //= 20;
-    $timeout //= 1;
-    while (!check_screen($tag, $timeout)) {
+    while (!check_screen($tag, %args)) {
         wait_screen_change {
             send_key $key;
         };
```
alvarocarvajald: Where else is `$args{counter}` needed? Checking in the rest of the file, it seems like neither `check_screen()` nor `_check_or_assert()` use it, so it appears there is no need to add it to `%args`.

testapi.pm
```diff
@@ -1144,22 +1144,30 @@ sub release_key {
...
+  send_key_until_needlematch($tag, $key [, [$counter] | [counter => $counter], [$timeout] | [timeout => $timeout], [no_wait => $no_wait]]);
+
 
 Send specific key until needle with C<$tag> is not matched or C<$counter> is 0.
-C<$tag> can be string or C<ARRAYREF> (C<['tag1', 'tag2']>)
-Default counter is 20 steps, default timeout is 1s
+C<$tag> can be string or C<ARRAYREF> (C<['tag1', 'tag2']>).
+Default counter is 20 steps, default timeout is 1s.
+
+Specify C<$no_wait> to run the screen check as fast as possible that is
```
okurz: I dislike copy-paste duplication. I would not even mention the "no_wait" parameter in the synopsis in L1147 but rather describe that all further arguments are forwarded to the internal check_screen call so you can give a hint that check_screen options can be used here as well :)

testapi.pm
alvarocarvajald: My suggestion for this:

```
my $counter = ($_[0] eq 'counter') ? $_[1] : looks_like_number($_[0]) ? $_[0] : 20;
$counter == $_[0] ? splice(@_, 0, 1) : splice(@_, 0, 2);
my $timeout = ($_[0] eq 'timeout') ? $_[1] : looks_like_number($_[0]) ? $_[0] : 1;
$timeout == $_[0] ? splice(@_, 0, 1) : splice(@_, 0, 2);
my %args    = (timeout => $timeout, @_);
```

Yeah, I know. Not too pretty. Perhaps will be easier to read with if-else blocks.

testapi.pm
```diff
@@ -1144,22 +1144,27 @@ sub release_key {
 
 =head2 send_key_until_needlematch
 
-  send_key_until_needlematch($tag, $key [, $counter, $timeout]);
+  send_key_until_needlematch($tag, $key [, [$timeout] | [timeout => $timeout]]);
```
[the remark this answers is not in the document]
ldevulder: Yes, corrected!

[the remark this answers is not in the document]
okurz: I think you are still supporting using timeout as positional argument, don't you? In that case I would write the help like this:

```perl
… [, [$counter] | [, $timeout] | [timeout => $timeout]]
```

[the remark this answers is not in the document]
Martchus: I would not include legacy-usage into the help.

foursixnine:
I wonder if tidy will complain...

okurz:
Could you add a test e.g. to t/03-testapi.t ?

Martchus:
I rebased and ran tidy again. No diff so we're good to merge.

okurz:
tests failed:

```
not ok 35 - No tests run for subtest "send_key_until_needlematch"
#   Failed test 'No tests run for subtest "send_key_until_needlematch"'
#   at ./03-testapi.t line 485.
unexpected response {
  'sim' => 999
} at ../testapi.pm line 359.
```

ldevulder:
> tests failed:

Yes I saw, don't had time to check further. I never do test like this, so it's new to me... And I'm pretty new to Perl too :) (openQA tests is easier to write because of the API).

ldevulder:
I still have to create the test for this PR... I need to check #952 as an example for this.

okurz:
please reopen when you continue the work. Thanks. Closing to clean up old open PRs.

PR #1066, opened 2018-11-22
Keep only the 30 most recently used needle images

Martchus:
See https://progress.opensuse.org/issues/15438

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -849,10 +849,8 @@ sub set_tags_to_assert {
     my $mustmatch = $args->{mustmatch};
     my $timeout   = $args->{timeout} // $bmwqemu::default_timeout;
 
-    # free all needle images (https://progress.opensuse.org/issues/15438)
-    for my $n (needle->all()) {
-        $n->{img} = undef;
-    }
+    # keep only the most recently used images (https://progress.opensuse.org/issues/15438)
+    needle::clean_image_cache();
 
     # get the array reference to all matching needles
     my $needles = [];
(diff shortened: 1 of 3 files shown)


backend/baseclass.pm
```diff
@@ -849,10 +849,8 @@ sub set_tags_to_assert {
     my $mustmatch = $args->{mustmatch};
     my $timeout   = $args->{timeout} // $bmwqemu::default_timeout;
 
-    # free all needle images (https://progress.opensuse.org/issues/15438)
-    for my $n (needle->all()) {
-        $n->{img} = undef;
-    }
+    # keep only 30 most recently used images (https://progress.opensuse.org/issues/15438)
```
okurz: magic number in comment?

t/01-test_needle.t
```diff
@@ -337,6 +337,33 @@ is($res->{needle}->{name}, 'login_sddm.ref.perfect', "perfect match should win")
 $res = $img1->search([$imperfect, $workaround], 0.9, 0);
 is($res->{needle}->{name}, 'login_sddm.ref.workaround.imperfect', "workaround match should win");
 
+# test caching via needle->get_image
```
okurz: because you add this header for this section, why not move it into a subtest?

backend/baseclass.pm
Martchus: I'll remove that number.

t/01-test_needle.t
okurz: ok, will you?

Martchus: Not as part of this PR. The benefit wouldn't be that high in my opinion, e.g. the different sub tests would actually still be intertwined.

PR #1067, opened 2018-11-26
WIP: Allow to skip assert/check_screen timeout for developer mode

Martchus:
* See https://progress.opensuse.org/issues/44249


Martchus:
This now really needs to be merged simultaneously with https://github.com/os-autoinst/openQA/pull/1894. Otherwise we increase the duration for the developer full-stack test by almost 10 minutes and it will likely timeout.

okurz:
@Martchus wouldn't that be a good argument for bumping the internal API version?

backend/baseclass.pm
```diff
@@ -900,6 +901,12 @@ sub set_tags_to_assert {
     return {tags => \@tags};
 }
 
+sub set_assert_screen_timeout {
+    my ($self, $timeout) = @_;
+    return bmwqemu::diag('set_assert_screen_timeout called with non-numeric timeout') unless looks_like_number($timeout);
```
okurz: shouldn't we die here?

t/19-isotovideo-command-processing.t
```diff
@@ -6,6 +6,7 @@ use Test::More;
 use Test::MockModule;
 use Test::Warnings;
 use OpenQA::Isotovideo::CommandHandler;
+use bmwqemu;
```
okurz: Wouldn't all the changes fit into another test module better when we now actually require bmwqemu?

Martchus: This test module is about command processing of isotovideo and the new tests are exactly about that.

Some commands might use `bmwqemu` internally e. g. to access the backend. So I'm using it here to assign a fake backend. That shouldn't be a problem, right?

okurz: yeah, sure. But if we need to override bmwqemu and in former tests we didn't it feels like we conduct tests on different levels where I would suggest to move to other test files. But with this stated I leave the decision to you and trust you.

Martchus: I thought about it and added another commit to get rid of `use bmwqemu` as it is really only used for mocking the backend but not by itself. This hopefully clarifies the code.

Martchus:
The project's coverage decrease doesn't seem to be valid.

t/19-isotovideo-command-processing.t
```diff
@@ -31,6 +31,24 @@ $rpc_mock->mock(read_json => sub {
         fail('we do not expect anything to be read here');
 });
 
+# mock bmwqemu/backend
```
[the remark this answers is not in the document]
Martchus: ok, good idea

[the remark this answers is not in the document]
okurz: That's a treacherous path. A "fake backend" might help for some testing but specific mocking for real unit tests is something different nevertheless, not really the same.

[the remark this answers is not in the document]
foursixnine: @okurz I did not understand a word of understanding of whatever you were trying to make me understand. :) But a fake backend is only meant to improve testing, including helping with feature tests or integration. Unit tests are a different thing.

backend/baseclass.pm
okurz: *bump*

Martchus: Yes. Or return 0/1 to distinguish wrong/successful calls. I'll add this to my TODOs for refactoring and general improvements.

PR #1069, opened 2018-11-30
Enhance VMware support

mnowaksuse:
https://trello.com/c/O104uCeQ/178-p1-vmware-guest-install-support-in-openqa

This brings VMware via svirt support up-to-date.

Validation runs (it's enough that VM boots):
* default_install@svirt-vmware: http://nilgiri.suse.cz/tests/1951
* jeos-extratest@svirt-vmware: http://nilgiri.suse.cz/tests/1999
* jeos-main@svirt-vmware: http://nilgiri.suse.cz/tests/1998

Regression testing (it's enough that VM boots):
* jeos-extratest@svirt-xen-pv: http://nilgiri.suse.cz/tests/2003
* allpatterns@svirt-xen-hvm: http://nilgiri.suse.cz/tests/2002

Test code changes (but this can be merged independently): https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6326.

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -73,8 +73,9 @@ sub do_stop_vm {
             $self->run_cmd("$ps Remove-VM -Force -VMName $vmname");
         }
         else {
-            $self->run_cmd("virsh destroy $vmname");
-            $self->run_cmd("virsh undefine $vmname");
+            my $libvirt_connector = get_var('VMWARE_REMOTE_VMM');
+            $self->run_cmd("virsh $libvirt_connector destroy $vmname");
+            $self->run_cmd("virsh $libvirt_connector undefine --snapshots-metadata $vmname");
         }
     }
     return {};
@@ -126,7 +127,7 @@ sub can_handle {
     my $vars = \%bmwqemu::vars;
     if ($args->{function} eq 'snapshots' && !check_var('HDDFORMAT', 'raw')) {
         # Snapshots via libvirt are supported on KVM and, perhaps, ESXi. Hyper-V uses native tools.
-        if (check_var('VIRSH_VMM_FAMILY', 'kvm') || check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
+        if (check_var('VIRSH_VMM_FAMILY', 'kvm') || check_var('VIRSH_VMM_FAMILY', 'hyperv') || check_var('VIRSH_VMM_FAMILY', 'vmware')) {
             return {ret => 1};
         }
     }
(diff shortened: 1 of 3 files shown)


Martchus:
I'm not familiar with VMware but the code looks good in general. Let's see whether the tests are passing. Maybe change the commit message "get_cmd_output() use other connections than 'ssh' " to "Make get_cmd_output() use other connections than 'ssh'" for consistency.

backend/svirt.pm
```diff
@@ -73,8 +73,9 @@ sub do_stop_vm {
             $self->run_cmd("$ps Remove-VM -Force -VMName $vmname");
         }
         else {
-            $self->run_cmd("virsh destroy $vmname");
-            $self->run_cmd("virsh undefine $vmname");
+            my $vmware_remote_vmm = get_var('VMWARE_REMOTE_VMM');
+            $self->run_cmd("virsh $vmware_remote_vmm destroy $vmname");
+            $self->run_cmd("virsh $vmware_remote_vmm undefine --snapshots-metadata $vmname");
```
alice-suse: IMHO, this $vmware_remote_vmm usage will make the svirt backend looks like supporting only hyperv and vmware. Its real default support on general linux will be blurred. So does the latter similar changes for vmware. 

What about changes 'vmware_remote_vmm' to a general name like 'vmm_connection' or whatever you prefer (no matter it is a local or remote connection to hypervisor), and just sets it to VMWARE_REMOTE_VMM when VIRSH_VMM_FAMILY is vmware?

mnowaksuse: Good idea @alice-suse, I changed the code according to your idea.

mnowaksuse:
Thanks for the review @Martchus. I changed the commit message and hopefully tamed perlcritic.

consoles/sshVirtsh.pm
```diff
@@ -330,43 +330,98 @@ sub add_interface {
 sub add_disk {
     my ($self, $args) = @_;
 
-    my $file = $self->name . $args->{dev_id} . (($self->vmm_family eq 'vmware') ? ".vmdk" : ".img");
+    my $backingfile             = $args->{backingfile};
+    my $cdrom                   = $args->{cdrom};
+    my $name                    = $self->name;
+    my $file                    = $name . $args->{dev_id} . ($self->vmm_family eq 'vmware' ? '.vmdk' : '.img');
+    my $basedir                 = '/var/lib/libvirt/images/';
```
foursixnine: I wonder if it would make sense to move this to a test/worker setting... But I see no reason to block from merging...

mnowaksuse:
codecov ain't happy, should I do something about it? (I don't even understand what's wrong.)

consoles/sshVirtsh.pm
foursixnine: @coolo does my comment make sense to you?

coolo: to my personal taste this whole vmm_family thingy became unmaintainable long ago - there are so many specific commands in rather unexpected places. I wouldn't single out single commands. This should be rather refactored on a whole - but as long as @mnowaksuse is the only user of all of this there is no real need :(

PR #1070, opened 2018-12-04
Use Perl::Critic::Freenode to determine good coding style

kraih:
This pull request does the same for os-autoinst as https://github.com/os-autoinst/openQA/pull/1902 did for openQA.

Progress: https://progress.opensuse.org/issues/44597.

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -72,22 +72,22 @@ sub new {
     $self->{properties} = $json->{properties} || [];
 
     my $gotmatch;
-    for my $area (@{$json->{area}}) {
-        my $hash = {};
+    for my $area_from_json (@{$json->{area}}) {
+        my $area = {};
         for my $tag (qw(xpos ypos width height)) {
-            $hash->{$tag} = $area->{$tag} || 0;
+            $area->{$tag} = $area_from_json->{$tag} || 0;
         }
         for my $tag (qw(processing_flags max_offset)) {
-            $hash->{$tag} = $area->{$tag} if $area->{$tag};
+            $area->{$tag} = $area_from_json->{$tag} if $area_from_json->{$tag};
         }
-        $hash->{match} = $area->{match} if $area->{match};
-        $hash->{type}   = $area->{type}   || 'match';
-        $hash->{margin} = $area->{margin} || 50;
+        $area->{match} = $area_from_json->{match} if $area_from_json->{match};
+        $area->{type}   = $area_from_json->{type}   || 'match';
+        $area->{margin} = $area_from_json->{margin} || 50;
 
-        $gotmatch = 1 if $hash->{type} =~ /match|ocr/;
+        $gotmatch = 1 if $area->{type} =~ /match|ocr/;
 
         $self->{area} ||= [];
-        push @{$self->{area}}, $hash;
+        push @{$self->{area}}, $area;
     }
 
     # one match is mandatory
@@ -115,12 +115,12 @@ sub save {
     my ($self, $fn) = @_;
     $fn ||= $self->{file};
     my @area;
-    for my $area (@{$self->{area}}) {
-        my $hash = {};
+    for my $area_from_json (@{$self->{area}}) {
+        my $area = {};
         for my $tag (qw(xpos ypos width height max_offset processing_flags match type margin)) {
-            $hash->{$tag} = $area->{$tag} if defined $area->{$tag};
+            $area->{$tag} = $area_from_json->{$tag} if defined $area_from_json->{$tag};
         }
-        push @area, $hash;
+        push @area, $area;
     }
     my $json = JSON->new->pretty->utf8->canonical->encode(
         {
--- a/needle.pm
+++ b/needle.pm
@@ -22,7 +22,8 @@ use autodie ':all';
 
 use File::Find;
 use File::Spec;
-use JSON;
+use Mojo::JSON 'decode_json';
+use Cpanel::JSON::XS ();
 use File::Basename;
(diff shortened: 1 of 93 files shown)


needle.pm
```diff
@@ -72,21 +73,21 @@ sub new {
 
     my $gotmatch;
     for my $area (@{$json->{area}}) {
-        my $a = {};
+        my $hash = {};
```
Martchus: When cleaning up anyways, maybe rename `$area` to `$area_from_json` or `$raw_area` and `$hash`/`$a` to `$area`. Same for the loop in `sub save`.

ppmclibs/tinycv.pm
```diff
@@ -75,16 +76,16 @@ sub search_ {
...
+        push @exclude, $area if $area->{type} eq 'exclude';
+        push @match,   $area if $area->{type} eq 'match';
+        push @ocr,     $area if $area->{type} eq 'ocr';
     }
 
     if (@exclude) {
         $img = $self->copy;
-        for my $a (@exclude) {
-            $img->replacerect($a->{xpos}, $a->{ypos}, $a->{width}, $a->{height});
+        for my $e (@exclude) {
```
Martchus: Please don't go from one bad variable name to the next. I'd rather call this `$exclude_area`.

ppmclibs/tinycv.pm
```diff
@@ -120,9 +121,9 @@ sub search_ {
 
     $ret->{error} = mean_square_error($ret->{area});
     if ($ret->{ok}) {
-        for my $a (@ocr) {
+        for my $o (@ocr) {
```
Martchus: ... `$ocr_area`.

t/04-check_vars_docu.t
```diff
@@ -57,18 +57,18 @@ sub read_doc {
     open($docfh, '<', VARS_DOC);
     my $backend;
     my $reading;
-    while (<$docfh>) {
-        if (!$backend && /^\.([^ ]+) backend$/) {
+    for my $line (<$docfh>) {
+        if (!$backend && $line =~ /^\.([^ ]+) backend$/) {
```
Martchus: This really helps understanding what this code does.

ppmclibs/tinycv.pm
kraih: Short variable names are perfectly fine if the scope is very limited, which is the case here. The greater the scope, the longer the variable name is one of the rules i personally follow. I'm not opposed to using longer variable names everywhere, but then we should be sure to enforce the rule consistently.

needle.pm
kraih: I'm not really cleaning up here, `$a` and `$b` are built-in Perl variables with special meaning. Using them outside their intended purpose is a bug.

kraih:
Once the Travis image has been updated this should pass. It already passes local tests on Tumbleweed.

okurz:
As we already use the perltidyrc from os-autoinst in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/Makefile#L26 maybe it makes sense to also reuse the perlcriticrc file?

@jknphy recently invested quite some effort to update perlcritic in os-autoinst-distri-opensuse , so we do not want to loose that though :)

kraih:
@okurz I think it would, running `Perl::Critic::Freenode` against os-autoinst-distri-opensuse i do see a lot of things that should really be avoided.
```
...
lib/apachetest.pm: Missing strict or warnings at line 1, column 1.  The strict and warnings pragmas are important toavoid common pitfalls and deprecated/experimental functionality. Make sure each script or module contains "use strict; use warnings;" or a module that does this for you.  (Severity: 4)
lib/ctcs2_to_junit.pm: Subroutine "new" called using indirect syntax at line 48, column 20.  See page 349 of PBP.  (Severity: 5)
...
lib/ipmi_backend_utils.pm: Subroutine called with "&" sigil at line 230, column 34.  See page 175 of PBP.  (Severity: 4)
...
lib/kdump_utils.pm: Using $a or $b outside sort() at line 50, column 16.  $a and $b are special package variables for use in sort() and related functions. Declaring them as lexicals like "my $a" may break sort(). Use different variable names.  (Severity: 4)
...
lib/mm_network.pm: <>/<<>>/readline/readdir/each result not explicitly assigned in while condition at line 29, column 5.  When used alone in a while condition, the <>/<<>> operator, readline, readdir, and each functions assign their result to $_, but do not localize it. Assign the result to an explicit lexical variable instead (my $line = <...>, my$dir = readdir ...).  (Severity: 4)
...
```
In some regards the rules i'm introducing here are also more relaxed. They do for example allow string evals, which have way too many sensible uses.

kraih:
Since we have to wait for the Travis test image to be fixed anyway, i'm also cleaning up the uses of `JSON` instead of `Mojo::JSON`/`Cpanel::JSON::XS`.

kraih:
There are a few cases where it looks like JSON is not properly UTF-8 encoded, which would crash and burn as soon as someone enters Unicode characters somewhere. But since i'm not familiar enough yet with how pipes are used in os-autoinst i will have to experiment a bit first, so those fixes might follow in a later pull request.

kraih:
Anyway, this pull request as it stands should be ready to merge and still passes all tests locally.

coolo:
First make the module available in devel:openQA. I gave you maintainer rights, so you can linkpac the needed packages from openSUSE:Factory into the subprojects

okurz:
+1

kraih:
Ok, travis tests are working again.

PR #1071, opened 2018-12-05
Avoid the prefix tests/ when using SCHEDULE

SergioAtSUSE:
See documentation: https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc


okurz:
The "tests/" subdirectory within os-autoinst-distri-opensuse is only a convention but not prescribed by os-autoinst

SergioAtSUSE:
The documentation should be updated instead

PR #1074, opened 2018-12-11
Don't write _SECRET_* vars in vars.json

cfconrad:
If a variable name starts with `_SECRET_*`, it is excluded from
writing in to the final vars.json file.
During the JOB, the variable is accessible via the normal functions
e.g. get_var().

Why: This is needed to store sensitive data (credentials) in workers.ini
and prevent them to be visible via the web-ui.

This change is related to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6346. The discussion on how to provide credentials, turns in the direction of better having hidden vars, then introduce a new configuration file.

Variables which are passed via job configuration or openqa-client are not affected and not in the focus of this PR. These variables still  visible in the settings tab.

Ticket: https://progress.opensuse.org/issues/42155
Verification run: http://cfconrad-vm.qa.suse.de/tests/3134#step/extra_log_test/14

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -217,7 +217,7 @@ sub run_all {
         $died = 1;    # test execution died
     }
     eval {
-        bmwqemu::save_vars();
+        bmwqemu::save_vars(no_secret => 1);
         myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
     };
     close $isotovideo;
--- a/t/12-bmwqemu.t
+++ b/t/12-bmwqemu.t
@@ -103,6 +103,40 @@ subtest 'test PRJDIR default' => sub {
     is($vars{PRJDIR},  $data_dir, 'PRJDIR set to default');
 };
 
+subtest 'save_vars' => sub {
+    my $dir = "$data_dir/tests";
+    create_vars({CASEDIR => $dir, _SECRET_TEST => 'my_credentials'});
+    $bmwqemu::openqa_default_share = $data_dir;
+
+    eval {
+        use bmwqemu ();
+        bmwqemu::init;
+        bmwqemu::save_vars();
+    };
+    ok(!$@, 'init successful');
+
+    my %vars = %{read_vars()};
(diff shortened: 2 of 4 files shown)


autotest.pm
```diff
@@ -217,7 +217,7 @@ sub run_all {
         $died = 1;    # test execution died
     }
     eval {
-        bmwqemu::save_vars();
+        bmwqemu::save_vars(no_secret => 1);
```
asmorodskyi: it is not the only place where `save_vars` actually called. Did you check others ? For example `save_vars` called on each `set_var` call so looks like your secrets will still leak after this  PR. 

Generally I don't see why we need to have another flavor of `save_vars` which actually saving them

foursixnine: @asmorodskyi It need to be the last part where the save_vars method is called, as this is the one that saves the final vars.json file that will be updated in the end to the webUI.

Martchus: @asmorodskyi Good point. Besides, would it ever make sense to call this with `no_scret => 0`? If not, I would not make it an option at all.

coolo:
I would also recommend to drop a note in `set_var` documentation about this

autotest.pm
coolo: the problem is that we use vars.json dump and save to transfer it between backend and autotest processes.

cfconrad:
Dropped note in set_var().

autotest.pm
cfconrad: Yes. If I would always avoid writing of such variables, the first test-module sees them, but not the second one during one job.

cfconrad:
Test cases for bmwqemu::save_vars() added.

t/12-bmwqemu.t
```diff
@@ -103,6 +103,40 @@ subtest 'test PRJDIR default' => sub {
     is($vars{PRJDIR},  $data_dir, 'PRJDIR set to default');
 };
 
+subtest 'save_vars' => sub {
```
Martchus: Perlcritic won't like the quotes.

cfconrad: thx

PR #1075, opened 2018-12-13
Add an option of mousehide for assert_and_click

mitiao:
This option always move mouse to the 'hidden' place after clicking to prevent to hide area where user
wants to assert/click in second step.

Also see: https://progress.opensuse.org/issues/35395
Verification run: http://10.67.19.14/tests/1341

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 13;
+our $version = 14;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -83,7 +83,10 @@ sub fake_read_json {
         return {ret => 1};
     }
     elsif ($cmd eq 'backend_get_last_mouse_set') {
-        return {ret => {x => -1, y => -1}};
+        return {ret => {x => 100, y => 100}};
+    }
+    elsif ($cmd eq 'backend_mouse_set') {
+        return {ret => {x => 100, y => 100}};
     }
     else {
         print "not implemented \$cmd: $cmd\n";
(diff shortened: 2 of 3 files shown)


Martchus:
This is a breaking API change so we need to merge it at the same time as https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6402.

mitiao:
@Martchus so add a unit test in https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/t ?

Martchus:
Likely. Maybe @okurz knows for sure.

foursixnine:
@mitiao There's already a test for it: https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L331

Martchus:
@foursixnine I suppose the question related to checking the incremented version. But extending the existing unit tests for that function would be great, too.

okurz:
hm, have you thought about a backward-compatible implementation? Shouldn't be too hard to support the old positional parameters as well as we did that also for other test methods.

mitiao:
@okurz the parameters are getting more and limited by position that is why i am using hash as parameters. I also changed the affected usage in the test module as referenced https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6402

okurz:
ok, the problem is only that with this and the according test change we would need to sync the deployment of tests + backend on both o3 and osd at least and additionally inform usual users of os-autoinst-distri-opensuse that they need to update os-autoinst to have their tests not fail.

Martchus:
> we would need to sync the deployment of tests + backend on both o3 and osd

Sounds annoying but not impossible.

> inform usual users of os-autoinst-distri-opensuse that they need to update os-autoinst to have their tests not fail

Should be as simple as writing a mail to the openQA and team mailing list.

If they were Arch users there wouldn't be a problem at all: https://external-preview.redd.it/37IvX3Ix0CGzlMdXEudtzQr3uyGe7Ep1NwceQbah4Kc.png?auto=webp&s=6245ebae88444ef5f1b6cde9eb90d36ea2a6cd30

Martchus:
@nicksinger Seems like I need to rebase this again. We should really not forget to finally merge this.

AdamWill:
this really should have accepted the old argument signature for a while before deprecating it. Just switching over makes it extremely awkward to have one distri work for both the old code (for a stable, production deployment) and the new code (for a staging, testing deployment).

AdamWill:
In case anyone else gets bitten, here's a convenience method that handles both:

    sub _assert_and_click {
        # this is a wrapper around assert_and_click which handles this:
        # https://github.com/os-autoinst/os-autoinst/pull/1075/files
        # it changed the signature without any backward compatibility, so
        # earlier os-autoinsts require an *array* of args, but later ones
        # require a *hash* of args. This works with both.
        my $version = $OpenQA::Isotovideo::Interface::version;
        my ($mustmatch, %args) = @_;
        if ($version > 13) {
            return assert_and_click($mustmatch, %args);
        }
        else {
	    $args{timeout}   //= $bmwqemu::default_timeout;
            $args{button}    //= 'left';
            $args{dclick}    //= 0;
            $args{mousehide} //= 0;
            return assert_and_click($mustmatch, $args{button}, $args{timeout}, 0, $args{dclick});
        }
    }

PR #1076, opened 2018-12-14
Reformat with latest version of perltidy

Martchus:


okurz:
If you want that then I think you need to update https://github.com/os-autoinst/os-autoinst/blob/master/cpanfile#L53 where we hardcode the version of perltidy. I wonder, why do you do this? To fix the new tidy problems in the openQA repo? Actually that could help me because I also have a failing test in https://travis-ci.org/okurz/openQA/jobs/467969386 but when we hardcode the version in os-autoinst I think we should do the same for openQA instead?

Martchus:
> where we hardcode the version of perltidy

I wasn't aware of that. I wanted to fix the same problem we see in the openQA repository here, too. Apparently it is not the case due to hard-coding the version.

> I wonder, why do you do this?

It would still make sense to apply the latest tidy for consistency with openQA where we already use it.

> we hardcode the version in os-autoinst I think we should do the same for openQA instead?

Too late. I wouldn't undo all those changes in openQA again. Besides, code formatted with the new tidy version actually looks better.

It also makes sense to use the latest perl tidy because sooner or later distributions will ship that version and it would be additional effort for developers to get an older version.

foursixnine:
@Martchus we could bring the change to the same as openQA. I think we stick to it, because it was/would and will affect os-autoinst-distri-opensuse. 

Also we had a lot of surprises of stuff failing due to tidy version changing (and last change prior to the fixed version, broke us for some time until we figured out https://github.com/os-autoinst/os-autoinst/pull/927).

Martchus:
I updated the PR. It now hardcodes the new version which is also used in openQA. There we should hardcode that version, too (see https://github.com/os-autoinst/openQA/pull/1929).

mitiao:
Does it need to add new version of perltidy in os-autoinst-distri-opensuse? since @GraceWang571 got travis fail:
https://travis-ci.org/os-autoinst/os-autoinst-distri-opensuse/jobs/470321811

okurz:
Ah, this got merged. I think, yes, everybody needs to update the version of os-autoinst checked out for the use of perltidy in tests

PR #1078, opened 2018-12-20
IPMI: support mc reset for sol stability

alice-suse:
This PR aims to increase ipmi job's stability via resetting ipmi main board thru 'mc reset' command.

* Background:
Detailed discussions are in https://progress.opensuse.org/issues/36027.
Shortly speaking, currently openqa jobs on ipmi backend is unstable because ipmi sol console itself is unstable after long time of job running or high work load. Resetting the ipmi main board can recover the sol console stability. 

Where to reset and how to reset sol?
Reset via external tools like jenkins jobs or reset based on given time or given frequency is very likely to break running openqa jobs because syncing is not easy. So doing it in openqa jobs and backend becomes preferred.

How often to reset?
Reset once before test start, and reset at weak points where sol unstability is likely to happen  like around reboot or power on.

* What it provides:
- backend method to handle mc reset in ipmi.pm
- console method to be exposed to testapi to call

* Related tickets:
test code PR: 

* Verification run:
  -http://10.67.18.220/tests/494:  guest installation test
 -  http://10.67.18.220/tests/498: offline host upgrade
  Both with mc reset before test and mc reset during test (before reboot,  in reboot_and_wait_up_normal.pm and reboot_and_wait_up_upgrade.pm)


alice-suse:
@coolo  @okurz  Would you please help to review?

alice-suse:
Travis check failed, can anyone help to see how to fix it?

backend/ipmi.pm
```diff
@@ -142,6 +149,57 @@ sub get_mc_status {
...
+    # deactivate sol console before doing mc reset because it breaks sol connection
+    if (defined $testapi::distri->{consoles}->{'sol'}) {
+        bmwqemu::diag("Before doing mc reset, sol console exists, so cleanup it");
+        $testapi::distri->{consoles}->{'sol'}->reset();
+        bmwqemu::diag("sol console reset done");
+        $self->deactivate_console({testapi_console => 'sol'});
+        bmwqemu::diag("deactivate console sol done");
+    }
+
+    # during the eval execution of following commands, SIG{__DIE__} will definitely be triggered, let it go
```
coolo: why? the eval is supposed to catch anyway

backend/ipmi.pm
```diff
@@ -142,6 +149,57 @@ sub get_mc_status {
     $self->ipmitool("mc selftest");
 }
 
+sub do_mc_reset {
+    my ($self) = @_;
+
+    # deactivate sol console before doing mc reset because it breaks sol connection
+    if (defined $testapi::distri->{consoles}->{'sol'}) {
+        bmwqemu::diag("Before doing mc reset, sol console exists, so cleanup it");
+        $testapi::distri->{consoles}->{'sol'}->reset();
```
coolo: the travis job fails because you use useless quotes here. Just {sol} will do

backend/ipmi.pm
```diff
@@ -97,6 +97,11 @@ sub restart_host {
 sub do_start_vm {
     my ($self) = @_;
 
+    # reset ipmi main board if switch on
+    if ($bmwqemu::vars{IPMI_BACKEND_MC_RESET} eq 'yes') {
```
coolo: So we set this in the workers.ini for the less stable machines?

backend/ipmi.pm
alice-suse: This class inherits the baseclass's signal handler for DIE, and will eventually trigger the die operation without this change, although when $@ catched we have a if branch to handle.

backend/ipmi.pm
alice-suse: got it, will change.

backend/ipmi.pm
alice-suse: yes, it is supposed to set in workers.ini, and better for all ipmi machines(all ipmi machines are unstable under long time working).

coolo: So why have a variable at all? For developing?

coolo: In any case: Add comments here and in the docs/ on when and why to use this variable with what content.

alice-suse: Thanks for your suggestion. I will add them.

Why having the variable is mainly that this solution to improve stability via mc reset needs to be proved by large practical openqa jobs statistics. We *may* need to switch on/off on some ipmi machines according to real run results and maybe different behaviors on different machines. So having this option provided can easily tune.

alice-suse:
All changes done. Please help to review!

okurz:
+1

\o/

alice-suse:
@coolo @okurz  Thank you!

PR #1079, opened 2018-12-20
Fix "tidy" use within os-autoinst-distri-opensuse

okurz:
See https://travis-ci.org/os-autoinst/os-autoinst-distri-opensuse/jobs/470365309

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -22,7 +22,8 @@ test -e tools/tidy || exit 1
 
 find -name '*.tdy' -delete
 
-find . \( -name '*.p[lm]' -o -name '*.t' \) -not -path '*/.git/*' -print0 | xargs -0 perltidy --pro=.../.perltidyrc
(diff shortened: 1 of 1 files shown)


tools/tidy
```diff
@@ -22,7 +22,8 @@ test -e tools/tidy || exit 1
 
 find -name '*.tdy' -delete
 
-find . \( -name '*.p[lm]' -o -name '*.t' \) -not -path '*/.git/*' -print0 | xargs -0 perltidy --pro=.../.perltidyrc
+# Exclude any "os-autoinst" subdirectory which for example is used for "os-autoinst-distri-opensuse" which symlinks "tools/tidy" into their own tests
+find . \( -name '*.p[lm]' -o -name '*.t' \) -not -path '*/.git/*' -not -path '*/os-autoinst/*' -print0 | xargs -0 perltidy --pro=.../.perltidyrc
```
foursixnine: i'm using it like `-not -path 'os-autoinst'` and also works, but I guess like this is also fine

PR #1081, opened 2019-01-02
Do not incomplete on connection error with ssh based consoles

Soulofdestiny:
[Verification run, bad case](http://opeth.suse.de/tests/3667#step/bootloader_s390/33)
[Verification run, good case](http://opeth.suse.de/tests/3668)

### DIFF
--- a/consoles/network_console.pm
+++ b/consoles/network_console.pm
@@ -1,4 +1,5 @@
-# Copyright © 2016-2018 SUSE LLC
+# Copyright © 2009-2013 Bernhard M. Wiedemann
+# Copyright © 2019 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


consoles/network_console.pm
```diff
@@ -0,0 +1,43 @@
+# Copyright © 2009-2013 Bernhard M. Wiedemann
+# Copyright © 2012-2019 SUSE LLC
```
Martchus: Looks like a new file. So `# Copyright © 2019 SUSE LLC` should be sufficient.

consoles/network_console.pm
```diff
@@ -0,0 +1,43 @@
...
+        $self->connect_remote($self->{args});
+        return $self->SUPER::activate;
+    }
+    catch {
+        die $_ unless blessed $_ && $_->can('rethrow');
+        return {error => $_->error};
+    };
+}
+
+# to be overloaded
```
Martchus: I suppose it should be over*written*.

Soulofdestiny: copy paste fail - fixed

consoles/network_console.pm
Soulofdestiny: done

okurz:
And now the coverage dropped ?!?

Soulofdestiny:
> And now the coverage dropped ?!?

weird indeed... I pushed again since the year for the testfile was still wrong - let's see

PR #1083, opened 2019-01-07
Add support to checkout git repos+refspecs

okurz:
As a test developer I want to execute an openQA test on a production
server based on a pull request to a test distribution to not need a
local openQA instance.

So far I tested this with the included integration test as well as
manually by specifying a real valid remote test distribution and branch
which has shown to be successful. The next steps to test would be
including the full openQA stack with remote workers including their
caching setup.

Related progress issue: https://progress.opensuse.org/issues/44327

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -15,9 +15,9 @@
 
 package OpenQA::Isotovideo::Utils;
 use Mojo::Base -base;
+use Mojo::URL;
 
 use File::Spec;
-use File::Path;
 use Cwd;
 use testapi 'diag';
 use bmwqemu;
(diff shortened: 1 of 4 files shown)


isotovideo
```diff
@@ -78,6 +78,7 @@ use File::Spec;
 use File::Path;
 use Mojo::IOLoop::ReadWriteProcess 'process';
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
+use Mojo::URL;
```
kraih: The module doesn't appear to be used.

t/14-isotovideo.t
```diff
@@ -47,4 +47,16 @@ EOV
     is_in_log('\d* scheduling.*shutdown', 'schedule has been evaluated');
 };
 
+subtest 'isotovideo with custom git repo parameters specified' => sub {
+    chdir($pool_dir);
+    unlink('vars.json');
+    mkdir('repo.git');
+    qx{git init -q --bare repo.git};
+    remove_tree('repo');
```
kraih: Where did "repo" come from?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -33,4 +32,35 @@ sub calculate_git_hash {
...
+
+Takes a test or needles distribution directory parameter and checks out the
+referenced git repository into a local working copy with an additional,
+optional git refspec to checkout. The git clone depth can be specified in the
+argument C<clone_depth> which defaults to 1.
+
+=cut
+sub checkout_git_repo_and_branch {
+    my ($dir, %args) = @_;
+    my $url = Mojo::URL->new($bmwqemu::vars{$dir});
```
kraih: `Mojo::URL` has not been imported.

isotovideo
okurz: fixed

t/14-isotovideo.t
okurz: I added a comment in file. This is to ensure the checkout folder does not exist so that git clone tries to create a new checkout on every test run.

OpenQA/Isotovideo/Utils.pm
okurz: fixed

okurz:
@kraih attended all your comments

asmorodskyi:
+1 , I was dreaming about that feature for a long time !

PR #1084, opened 2019-01-07
Add "error" value to matches

Martchus:
* Add a test for basetest::record_screenmatch
* Add the "error" value not only to the candidate list but also to the actual match
* See https://progress.opensuse.org/issues/45746

---

I've imported the test details from the example o3 test locally (see progress ticket) to see how the web UI behaves when adding the error. It appears to work out of the box.


okurz:
+1

PR #1085, opened 2019-01-08
Add missing network_console.pm to Makefile

asdil12:


asdil12:
@Soulofdestiny @foursixnine

PR #1086, opened 2019-01-14
qemu: use /dev/urandom as source for virtio RNG

AdamWill:
qemu by default uses the host's /dev/random as the source for
virtio-rng, which means we still *can* run out of entropy if a
bunch of guests demand a lot and exhaust /dev/random on the
host. Since nothing run inside os-autoinst should be used in
any kind of security-sensitive production context, let's just
use host /dev/urandom as the entropy source for os-autoinst VMs
instead.

For a case where this caused some problems, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1663318

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -745,11 +745,12 @@ sub start_qemu {
             sp('global', 'isa-fdc.driveA=');
         }
 
-        sp('m',       $vars->{QEMURAM})     if $vars->{QEMURAM};
-        sp('machine', $vars->{QEMUMACHINE}) if $vars->{QEMUMACHINE};
-        sp('cpu',     $vars->{QEMUCPU})     if $vars->{QEMUCPU};
-        sp('device',  'virtio-rng-pci')     if $vars->{QEMU_VIRTIO_RNG};
-        sp('net',     'none')               if $vars->{OFFLINE_SUT};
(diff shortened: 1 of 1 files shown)


AdamWill:
Note: if anyone thinks just unconditionally using `/dev/urandom` is too much, we can make this yet *another* option (have a `QEMU_VIRTIO_RNG_DEVICE`, or just allow setting `QEMU_VIRTIO_RNG to a path and treat a value that looks like a path as one, something like that). But I thought I'd propose the simplest approach first, if everyone is fine with this, it's the easiest thing to do and adds very little complexity.

ggardet:
In OBS we use ```-object rng-random,filename=/dev/hwrng,id=rng0 -device virtio-rng-device,rng=rng0``` not sure how different/similar it is.

AdamWill:
`/dev/hwrng` is - IIUC - what an actual hardware RNG in the system, supported by one of the `hw_random` drivers, will show up as. Any such hardware RNGs *should* be used by the kernel as an entropy source for `/dev/random`, again IIUC, so I think the effect of this is to limit OBS to *only* using entropy from the HW RNG, as opposed to using entropy from it *and also from any other source used by `/dev/random`*. I don't know if that's what is intended. In any case, it's not what this does, and it wouldn't really help this case, I don't think.

PR #1087, opened 2019-01-14
openvswitch: change vlan_mode to dot1q-tunnel

cfconrad:
Setting vlan_mode=dot1q-tunnel makes the VLAN configuration, which is used in
SUT, independent from the VLAN configuration needed by openqa-workers.
The problem was discovered during a VLAN test for wicked, where two VMs
setup a VLAN interface and try to ping each other.

Related Ticked: https://progress.opensuse.org/issues/45461
Verification run:
  http://fromm.arch.suse.de/tests/4794 (SUT) [worker on  fromm.arch.suse.de]
  http://fromm.arch.suse.de/tests/4793 (REF) [worker on cfconrad-vm.qa.suse.de]

Attached a pcap from the GRE tunnel between fromm and cfconrad-vm, which shows the ping of two parallel running wicked_startandstop testsuites. Each test has it's own tunnel VLAN (2 or 4), after this we see the VLAN:42 which comes from the wicked testsuite.
[gre_wicked_vlan_test.pcapng.zip](https://github.com/os-autoinst/os-autoinst/files/2757016/gre_wicked_vlan_test.pcapng.zip)

### DIFF
--- a/os-autoinst-openvswitch
+++ b/os-autoinst-openvswitch
@@ -146,7 +146,7 @@ sub set_vlan {
     }
 
     # Connect tap device to given vlan
-    ($return_code, $ovs_vsctl_error, $ovs_vsctl_output) = _cmd( 'ovs-vsctl', 'set', 'port', $tap, "tag=$vlan" );
+    ($return_code, $ovs_vsctl_error, $ovs_vsctl_output) = _cmd( 'ovs-vsctl', 'set', 'port', $tap, 'vlan_mode=dot1q-tunnel', "tag=$vlan" );
 
     print STDERR $ovs_vsctl_error if length($ovs_vsctl_error) > 0;
     print $ovs_vsctl_output if length($ovs_vsctl_output) > 0;


thehejik:
> @asmorodskyi, @jlausuch, @thehejik WDYT?

Is there any way how can we set this `vlan_mode` option only if really needed? I'm bit afraid it can influence all multi-machine tests because you are setting `vlan_mode` to every tap device in use.

side-note: by following command we can verify if the settings for tap device in ovs switch on worker are correctly set `ovs-vsctl get port tap1 vlan_mode tag`

---
### MTU problem
I would like to know if also TCP communication (and not only ping/ICMP) between SUTs is passing through the GRE tunnel when using this additional `vlan_mode` and VLAN set on interface within SUTs.

_Note: GRE is using DF (don't fragment) flag on IP packets and MTU 1500 is too high - it needs to do a fragmentation which is disabled by GRE and then the TCP communication will fail. Regarding to MTU see last Note [here]( https://github.com/os-autoinst/openQA/blob/master/docs/Networking.asciidoc#gre-tunnels)_

So please try following, first with default MTU=1500 and in case it will not work (it should not) try also with MTU=1458 set on eth0 interfaces on both VMs (MTU=1458 is used by support_server's dhcpd):
- on SUT1 run sshd and try to connect from SUT2
- on SUT2 run apache2 and try to use curl http://sut2 from SUT1

BTW in your [test](http://fromm.arch.suse.de/tests/4794#step/t11_vlan_ifdown_modify_one_config/147) I see that you are using default MTU 1500 which means that TCP communication will most likely not pass through the GRE.

jlausuch:
Full explanation of the problem here: https://progress.opensuse.org/issues/45461#note-9

cfconrad:
> Is there any way how can we set this `vlan_mode` option only if really needed? I'm bit afraid it can influence all multi-machine tests because you are setting `vlan_mode` to every tap device in use.

Do you mean for production, or to test it before going in production?
In case of production, it would not make sense to me to have different behaviors depending on the test-module type. As this makes trouble shooting even more worse.

> side-note: by following command we can verify if the settings for tap device in ovs switch on worker are correctly set `ovs-vsctl get port tap1 vlan_mode tag`

wonne add `ovs-vsctl list Port [name|uuid]` to the helpfull commands

> ### MTU problem
Right, we just discovered a problem with TCP and MTU. Decrease the MTU solve it.

asmorodskyi:
@thehejik can you suggest some multi-machine job ( from existing or synthetic ) which we could run to prove that our PR not breaking anything ?

jlausuch:
After testing TCP data transfer between both VMs, we saw that SCP transfer gets stalled. We need to change the MTU size in the VM to make it work. 

However, this problem is not new. This problem is there always when GRE tunnels are used (different workers). It has nothing to do with the addition of having vlan interface in the VM. We actually tried to do SCP using normal eth0 in the VM with default 1500 MTU size and it failed, and had to change it to 1458 to make it work. So we can say that this PR doesn't create a MTU problem.

The good thing is that currently there are no Multi-machine test which transfer files between VMs, only ICMP packets, which is fine.

thehejik:
> After testing TCP data transfer between both VMs, we saw that SCP transfer gets stalled. We need to change the MTU size in the VM to make it work.
> 
> However, this problem is not new. This problem is there always when GRE tunnels are used (different workers). It has nothing to do with the addition of having vlan interface in the VM. We actually tried to do SCP using normal eth0 in the VM with default 1500 MTU size and it failed, and had to change it to 1458 to make it work. So we can say that this PR doesn't create a MTU problem.

Cool so we don't need to change MTU again and we can stay on MTU=1458

> The good thing is that currently there are no Multi-machine test which transfer files between VMs, only ICMP packets, which is fine.

True, I don't know about any testsuite transferring files between VMs BUT there are mm tests using TCP between VMs intensively -  for example CaaSP-* stack test - where we use for eg zeroMQ(salt), http, ssh and probably even much more needed for k8s.

So once the CaaSP-* stack test will pass in your test env we are safe :)

jlausuch:
> So once the CaaSP-* stack test will pass in your test env we are safe :) The test is running http://fromm.arch.suse.de/tests/4828

It seems it's green
http://fromm.arch.suse.de/tests/overview?distri=caasp&version=3.0&build=GM&groupid=5

thehejik:
> > So once the CaaSP-* stack test will pass in your test env we are safe :) The test is running http://fromm.arch.suse.de/tests/4828
> 
> It seems it's green
> http://fromm.arch.suse.de/tests/overview?distri=caasp&version=3.0&build=GM&groupid=5

then LGTM :-)

foursixnine:
@coolo any further comments?

PR #1088, opened 2019-01-15
Send Ctrl-Alt-Fx more times

coolgw:
Sometimes the console can not switch after send Ctrl-Alt-Fx in openqa
env, but can not easy reproduce in local env, so maybe caused by worker
performance or network issue, so try to send key more times, this is a
workround. 

Related ticket: https://progress.opensuse.org/issues/45575
Needles: None
Verification run: http://10.67.17.238/tests/29#

### DIFF
--- a/consoles/ttyConsole.pm
+++ b/consoles/ttyConsole.pm
@@ -39,7 +39,10 @@ sub trigger_select {
         $key = "ctrl-alt-f" . $self->{args}->{tty};
     }
 
-    $self->screen->send_key({key => $key});
+    # Sometimes the screen can not change after send key, so try to send key more times
+    for my $i (1 .. 3) {
+        $self->screen->send_key({key => $key});
+    }
     return;
 }
 


coolo:
in all cases where people tried to send more keys it didn't help either. The key isn't lost - but the system just doesn't react because it's busy. So sending 3 keys won't help. 

This needs to be handled from within the tests - not with force

coolgw:
@coolo 
How to do retry action within os-autoinst-distri-opensuse in stead of os-autoinst? After some investigation,  only  select_console(testapi.pm) can be called within test module, i can not recall select_console again within test cases since this function do a lot of things(start call $testapi::distri->activate_console and check the needles etc..).
So could you help give some suggestion using which API to do this retry select_console task within test cases.

coolo:
activating consoles triggers a callback into the tests - you can use this.

coolgw:
@coolo 
callback indeed in tests, and callback will try to check needle, when needle not match means the system not response so i need redo the select_console action again, so i still need an API to do following action again, correct? But currently no this kind of API in os-autoinst.

query_isotovideo('backend_select_console', {testapi_console => $testapi_console});

coolo:
no, repeating select_console doesn't change a thing - as the console is already selected. You need to repeat the key presses you need.

coolgw:
@coolo 
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6650 submitted, could you have a look? Thanks

PR #1089, opened 2019-01-15
Add python binding/import/wrapping code for testapi

okurz:
This is based on bmwiedemann's work, see for example the initial idea in
https://github.com/os-autoinst/os-autoinst/pull/364 .
The idea for "python bindings" was recently brought up as well in
See https://trello.com/c/GSXjk2wZ/35-python-bindings-for-openqa

With this test modules can written in python code importing the testapi
module from python. All perl objects not within the testapi can still be
accessed using the dynamic python module "perl".

Detailed changes:

* 08-autotest.t: Add test for python test modules
* t: Add test for python test modules using perl testapi
* Add python testapi interface

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -53,7 +53,7 @@ loadtest is called.
 sub find_script {
     my ($script) = @_;
     my $casedir = $bmwqemu::vars{CASEDIR};
-    my $script_override_path = join('/', $bmwqemu::vars{PRJDIR}, 'factory/other', $script);
+    my $script_override_path = join('/', $bmwqemu::vars{PRJDIR} // '', 'factory/other', $script);
     if (-f $script_override_path) {
         bmwqemu::diag("Found override test module for $script: $script_override_path");
         return File::Spec->abs2rel($script_override_path, $casedir);
--- a/autotest.pm
+++ b/autotest.pm
@@ -110,7 +110,21 @@ sub loadtest {
     $code .= "use lib '$casedir/lib';";
     my $basename = dirname($script_path);
     $code .= "use lib '$basename';";
-    $code .= "require '$script_path';";
+    if ($script =~ m/\.pm$/) {
+        $code .= "require '$script_path';";
+    }
+    elsif ($script =~ m/\.py$/) {
+        # import the test API and all methods from the test API by default
+        # into python context
+        $code .= "
+            use base 'basetest';
+            use Inline Python => <<'END_OF_PYTHON_CODE';\n";
+        $code .= `cat $casedir/$script`;
+        $code .= "\nEND_OF_PYTHON_CODE\n";
+    }
+    else {
+        die "impossible codepath";
+    }
     eval $code;
     if ($@) {
         my $msg = "error on $script: $@";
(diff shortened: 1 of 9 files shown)


okurz:
@lnussel you added crop.py to be installed in 587bae06 . I wonder what good is a python file in "/usr/lib/os-autoinst". Shouldn't it be in `/usr/lib/python*/…`?

okurz:
Ok, I think I need to submit perl-Inline-Python first to openSUSE:Factory build.opensuse.org/request/show/666217 and then devel:openQA

coolo:
I'm 100% against mixing languages in the tests. If you want python, port all tests and rewrite the documentation.

okurz:
@coolo we already mix languages, e.g. with the scripts https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data , right?

Do you think it's realistic to port all test modules to python and convince everybody?

coolo:
No, I don't think it's realistic. As such this should be closed - the reasons in 364 still apply today.

coolo:
And the scripts in data are called in itself - they aren't openQA tests.

coolo:
there are 15T lines of code in distri-opensuse/lib - opening all this up to duplication is something I veto. And as such I'm against 'having people experiment with it'.

Either all or nothing. This is not a binding, this is a mixture.

okurz:
I don't think there is a need to rewrite any perl module as they can just be imported, so no duplication at all should be necessary. Otherwise I would not have done it anyway.

okurz:
@coolo does my last statement change your mind? We do *not* need to rewrite all files in lib to python

coolo:
> @coolo does my last statement change your mind? 

No. I find mixing languages in one tool is road to madness.

aaannz:
My 2c, there can still be rules wrt. individual test projects. So even though os-autoinst would support python, distri-opensuse would not allow mixing languages. I personally think that if this solution works, this can be good for the project and its growth.

okurz:
> mixing languages in one tool is road to madness

like javascript, html, sql, perl, C, C++ and bash? ;)

coolo:
Right.

okurz:
So I wonder what else we can do to convince coolo.

On top I wonder what else is the *road to madness*:

```
[Wednesday 20 January 2016] [11:31:36 A.M. CET] <okurz>>coolo: can we please keep issues open until the issuer is happy? we don't want to intimidate newcomers, do we?
[Wednesday 20 January 2016] [11:32:50 A.M. CET] <coolo>>no. that's the road to madness
```

;P

coolo:
So you too keep a record of quotes of mine? Tss, welcome in the club :)

coolo:
So: I'm vetoing the use of Inline::Python for all I have.

I agree that having a way to write your test distribution in python could lead to growth of the project, but Inline::Python is no solution. Neither is mixing python and perl helping in any way to untangle the main.pm issue as we have it.

So if you want python: do it properly. Have a custom API and a python way to load tests - don't mimik perl.

okurz:
What do you mean with "custom API"?

coolo:
A testapi.py that talks RPC like testapi.pm does.

okurz:
@coolo my approach was trying to not duplicate manually all the testapi calls in python. Do you think we could "talk RPC like testapi.pm" does but without duplicating all the code in python?

coolo:
well, @aaannz once tried by implementing a testapi layer with minimal interface. By now this would be in isotovideo itself. And this is what I tried to achieve, minimizing the code in testapi.pm replacing it with RPC. But other code is just syntactic sugar for linux distributions that I would indeed replicate in python, because if I started with a new language, I would also seperate the linux distribution handling more clearly.

okurz:
So you would change the meaning of the API when implementing other language bindings? You seem to have very specific expectations which I doubt I or anyone else can fulfill so can you do it yourself?

PR #1090, opened 2019-01-18
openvswitch: check dot1q-tunnel support

cfconrad:
dot1q-tunnel was introduced with openvswitch version 2.8.1+.
Avoid this feature when openvswitch is to old.

Update to https://github.com/os-autoinst/os-autoinst/commit/3747133a4831bc297e990b71bde8768a9158fb89


cfconrad:
Maybe we could use https://metacpan.org/pod/version to compare version strings. But I avoided it for now, to not increase the dependencies.

okurz:
+1

PR #1091, opened 2019-01-27
doc: Add PAUSE_AT

okurz:


okurz:
@Martchus I understood `PAUSE_AT` makes sense to be used to trigger a job with so that the developer mode directly pauses at the specified module. What about PAUSE_ON_SCREEN_MISMATCH and PAUSE_ON_NEXT_COMMAND. Do these variable make sense to be specified manually?

Martchus:
I would say yes. But I'd like to note that none of these variables trigger the developer mode in the sense that such a session is opened and tracked and the responsible developer could be comprehend.

okurz:
Do you mean not even PAUSE_AT? Because I thought it worked for me that the according openQA job started and then waited until I interacted with it with the developer mode

okurz:
> in the sense that such a session is opened and tracked

do you mean that just specifying this variable on a job does not magically open a browser in the face of the developer? Because otherwise I don't know what you mean with "session"

okurz:
Alright, updated as discussed

PR #1092, opened 2019-01-28
Prevent warning in command handler

Martchus:
The variable is set to undef to disable pausing. So it should not be
printed unconditionally.

See https://progress.opensuse.org/issues/46721


okurz:
+1

Martchus:
I added a unit test which covers both cases.

PR #1099, opened 2019-02-06
Silence complain about undefined $libvirt_connector

foursixnine:


foursixnine:
> But it makes still sense to execute the command above because `$VMWARE_REMOTE_VMM` is only supposed to contain a global option, right?

Correct, other backends using virsh will [still work](https://openqa.suse.de/tests/2440012#)

backend/svirt.pm
```diff
@@ -145,7 +145,7 @@ sub is_shutdown {
         $rsp = $self->run_cmd("powershell -Command \"if (\$(Get-VM -VMName $vmname \| Where-Object {\$_.state -eq 'Off'})) { exit 1 } else { exit 0 }\"");
     }
     else {
-        my $libvirt_connector = get_var('VMWARE_REMOTE_VMM');
+        my $libvirt_connector = get_var('VMWARE_REMOTE_VMM') || '';
```
okurz: When already using testapi methods I suggest to use them also here to handle the value feedback, so

```
get_var('too', '')
```

foursixnine: Good catch :)

pevik:
We've just discussed with @mnowaksuse, that it'll be better to refactor [consoles/sshVirtsh.pm](https://github.com/os-autoinst/os-autoinst/tree/master/consoles/sshVirtsh.pm) (`add_disk()`) [backend/svirt.pm](https://github.com/os-autoinst/os-autoinst/tree/master/backend/svirt.pm) to move vmware specific functions to it's own functions (but it a question when some of us find a time for it):
https://progress.opensuse.org/issues/47258

PR #1100, opened 2019-02-06
Fix table documenting svirt backend variables

Martchus:
I've just noticed that the table is completely messed. Since we might want to point to the documentation in the upcoming training I thought it would be worth fixing it.


Martchus:
I've just used the GitHub renderer to check this. It also works on PR branches.

okurz:
stupid me. Of course I could have just clicked on https://github.com/Martchus/os-autoinst/blob/c3dc6cfe20421daf01de5209f041d4d87d6c1114/doc/backend_vars.asciidoc :D

PR #1102, opened 2019-02-12
Fix skipping needles with missing PNGs

Martchus:
The error handling was just done too late - after attempting to the the (non-existent) image. I also fixed a memory leak in the case this error happens.

(see https://progress.opensuse.org/issues/47372)


okurz:
+1

PR #1103, opened 2019-02-13
qemu: Don't use deprecated/removed vlan option for multinet

Martchus:
Instead, just increase the number of network devices to be added to get another device (with different MAC address).

See https://progress.opensuse.org/issues/29419

@richiejp Am I'm missing something or is it really that easy? At least `ip addr` now shows the additional device with IP.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -533,7 +533,6 @@ sub start_qemu {
 
     die "no kvm-img/qemu-img found\n" unless $qemuimg;
     die "no Qemu/KVM found\n"         unless $qemubin;
-    die "MULTINET is not supported with NICTYPE==tap\n" if ($vars->{MULTINET} && $vars->{NICTYPE} eq "tap");
 
     $self->{proc}->qemu_bin($qemubin);
     $self->{proc}->qemu_img_bin($qemuimg);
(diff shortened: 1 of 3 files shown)


richiejp:
It is not clear to me whether the two NICs will be on the same VLAN. @asmorodskyi says he will look into it. I'm not familiar with this code so I don't have much useful to say without doing a lot of research.

Martchus:
I don't think those would be in the same VLAN. Since the previous code explicitly used `vlan=1` I thought that this might be what's wanted.

asmorodskyi:
in this case I suggest to drop this variable at all. I don't see actual use case where such combination could be tested. Basically if you got two NIC which not able to connect each other this means that you can only cover same thing which you could cover with "normal" one NIC machine. 
Also after reading code I realized that same thing should be achievable by use of `NICMAC`   or `TAPDEV` variables

Martchus:
It is true, the change `my $num_networks = $vars->{MULTINET} ? 2 : 1;` would be just a shortcut for what could be achieved by passing multiple values via `NICMAC` or `TAPDEV` variables.

So I removed the support for that variable at all and updated the documentation to use `NICMAC` or `TAPDEV` variables. I don't think anybody is using that `MULTINET` code path containing obsolete and likely wrong code anyways, right?

Martchus:
So I assume nobody has any objections?

PR #1104, opened 2019-02-14
Enhance documentation for testapi::wait_serial()

ricardobranco777:
- Add the `no_regex` paramenter to the documentation of `wait_serial`.
- Document the default for `timeout`.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -789,7 +789,7 @@ sub is_serial_terminal {
 
 =head2 wait_serial
 
-  wait_serial($regex or ARRAYREF of $regexes [[, $timeout], $expect_not_found]);
+  wait_serial($regex or ARRAYREF of $regexes [, $timeout [, $expect_not_found [, @args ]]]);
 
 Wait for C<$regex> or anyone of C<$regexes> to appear on serial output.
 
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -789,15 +789,18 @@ sub is_serial_terminal {
 
 =head2 wait_serial
 
-  wait_serial($regex or ARRAYREF of $regexes [[, $timeout], $expect_not_found]);
+  wait_serial($regex or ARRAYREF of $regexes [[[, $timeout], $expect_not_found],
```
Martchus: Not sure about our convention but `[[[` looks wrong.

okurz: Actually I think it should be the other way around. The only required argument is `$regexp` but one can only specify additional arguments if $expect_not_found and $timeout are both specified as they are `shift`-ed in first. A different thing on top: `no_regex` is no parameter of `wait_serial` but rather all additional arguments are forwarded. So I guess it should read

```suggestion
  wait_serial($regex or ARRAYREF of $regexes [, $timeout [, $expect_not_found [, @args ]]]);
```

similar to what we have for `select_console($console [, @args ])`

However we can still keep the explanation that `no_regex` as one of these parameters can be specified and what it does.

testapi.pm
```diff
@@ -789,15 +789,18 @@ sub is_serial_terminal {
 
 =head2 wait_serial
 
-  wait_serial($regex or ARRAYREF of $regexes [[, $timeout], $expect_not_found]);
+  wait_serial($regex or ARRAYREF of $regexes [, $timeout [, $expect_not_found [, @args ]]]);
+                                                no_regex => 0]);
```
okurz: you need to delete this line which is dangling around :)

ricardobranco777: Yes.  Done!

okurz:
:+1:

PR #1106, opened 2019-02-14
Extend arguments for type_string

mloviska:
I would like to make return values from ``wait_still_screen`` somehow useful in our test suites. If I correctly understand ``wait_still_screen``, it's basically letting us know whether the SUT had still screen for some certain amount of time, then it returns success. Otherwise it reports a failure which we tend to ignore. I have found it useful use case in our yast2 gui tests, when we are filling some forms using ``type_string``.

* failure: http://eris.suse.cz/tests/11922#step/yast2_hostnames/31
* success: http://eris.suse.cz/tests/12008#step/yast2_hostnames/27

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1215,7 +1215,8 @@ sub send_key_until_needlematch {
 
 =head2 type_string
 
-  type_string($string [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, secret => 1 ] );
+  type_string($string [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, secret => 1 ] 
+  [, timeout => <num>] [, similarity_level => <num>] );
 
 send a string of characters, mapping them to appropriate key names as necessary
 
@@ -1227,8 +1228,12 @@ C<max_interval> the slower the typing.
 C<wait_screen_change> if set, type only this many characters at a time
 C<wait_screen_change> and wait for the screen to change between sets.
 
-C<wait_still_screen> if set, C<wait_still_screen> for the given seconds
-after the whole string is typed.
+C<wait_still_screen> if set, C<wait_still_screen> returns true if screen is not 
+changed for given C<$wait_still_screen> seconds or false if the screen is not still
+for the given seconds within defined C<timeout> after the whole string is typed.
+Default timeout is 30s, default stilltime is 0s.
+
+C<similarity_level> can be passed as argument for wrapped C<wait_still_screen> calls.
 
 C<secret (bool)> suppresses logging of the actual string typed.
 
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -1252,25 +1252,32 @@ sub type_string {
...
+        timeout => $wait_timeout, similarity_level => $wait_sim_level);
     if ($wait) {
         # split string into an array of pieces of specified size
         # https://stackoverflow.com/questions/372370
         my @pieces = unpack("(a${wait})*", $string);
         for my $piece (@pieces) {
             wait_screen_change { query_isotovideo('backend_type_string', {text => $piece, max_interval => $max_interval}); };
             if ($wait_still) {
-                wait_still_screen($wait_still);
+                unless (wait_still_screen(stilltime => $wait_still, timeout => $wait_timeout, similarity_level => $wait_sim_level)) {
```
Martchus: I'd combine the `if ... unless` into a single `if`.

mloviska:
Alright, I will add a unit test for it. Thanks!

testapi.pm
```diff
@@ -1252,25 +1252,30 @@ sub type_string {
...
         # split string into an array of pieces of specified size
         # https://stackoverflow.com/questions/372370
         my @pieces = unpack("(a${wait})*", $string);
         for my $piece (@pieces) {
             wait_screen_change { query_isotovideo('backend_type_string', {text => $piece, max_interval => $max_interval}); };
-            if ($wait_still) {
-                wait_still_screen($wait_still);
+            if ($wait_still && !wait_still_screen(stilltime => $wait_still,
+                    timeout => $wait_timeout, similarity_level => $wait_sim_level)) {
+                die "wait_still_screen timed out after ${wait_timeout}s!!!";
```
okurz: Please use single `!`'s only :)

testapi.pm
```diff
@@ -1252,25 +1252,30 @@ sub type_string {
...
             }
         }
     }
     else {
         query_isotovideo('backend_type_string', {text => $string, max_interval => $max_interval});
-        if ($wait_still) {
-            wait_still_screen($wait_still);
+        if ($wait_still && !wait_still_screen(stilltime => $wait_still,
+                timeout => $wait_timeout, similarity_level => $wait_sim_level)) {
+            die "wait_still_screen timed out after ${wait_timeout}s!!!";
```
okurz: It seems we can extract this block from the if/else as it is called in both places. This would prevent the additional duplication. However I recommend you extend the unit tests first, then you can extract that part and see if tests still work or if I was wrong :)

mloviska: yup, that is my plan for today. Thanks for hints

mloviska:
Test cases have been added

testapi.pm
```diff
@@ -1252,25 +1252,27 @@ sub type_string {
...
 
-    my $max_interval = $args{max_interval}       // 250;
-    my $wait         = $args{wait_screen_change} // 0;
-    my $wait_still   = $args{wait_still_screen}  // 0;
-    bmwqemu::log_call(string => $log, max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still);
+    my $max_interval   = $args{max_interval}       // 250;
+    my $wait           = $args{wait_screen_change} // 0;
+    my $wait_still     = $args{wait_still_screen}  // 0;
+    my $wait_timeout   = $args{timeout}            // 30;
+    my $wait_sim_level = $args{similarity_level}   // 47;
```
rwx788: Could you please, also update perldoc above for your changes, so it's visible for other who wants to use this feature.

mloviska: Done, please check!

rwx788: looks good!

testapi.pm
```diff
@@ -1252,25 +1257,27 @@ sub type_string {
...
+        @pieces = unpack("(a${wait})*", $string);
     }
     else {
-        query_isotovideo('backend_type_string', {text => $string, max_interval => $max_interval});
-        if ($wait_still) {
-            wait_still_screen($wait_still);
+        push @pieces, $string;
+    }
+    for my $piece (@pieces) {
+        wait_screen_change { query_isotovideo('backend_type_string', {text => $piece, max_interval => $max_interval}); };
```
okurz: I think we have a problem here. In http://lord.arch.suse.de/tests/2443/file/autoinst-log.txt looking for `wait_screen_change` I found that `wait_screen_change` is called but I never requested it. The problem is that we always loop over `@pieces` now and call `wait_screen_change` even when not asked. In most cases this will not fail but just waste time when there is no actual screen change visible. In some cases this might fail in obscure ways, e.g. when no graphical console is actually connected, e.g. when typing `reboot` on a remote backend where the graphical session dies as soon as the command is effectively rebooting the SUT

okurz: reported as https://progress.opensuse.org/issues/54935

okurz: fix in https://github.com/os-autoinst/os-autoinst/pull/1183

PR #1107, opened 2019-02-15
[WIP][AARCH64 Automation] Save the original value of SERIALDEV for grub configuration

waynechen55:
This commit should be merged together with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800

There is a defect found during openQA integration with aarch64 automation. Firstly, not all aarch64 machines use ttyAMA0 as their serial console. Some vendor meachines like xgene has only ttyS0 as their serial console. Secondly, the SERIALDEV variable is overridden by some other functionalities
during automation run. So it is necessary to save its orignal value to SERIALDEVORIG variable which is only for internal use. It should not be referred to from outside of live test run, for example, assign a value to it explicitly in test suite or machine settings.

Verification run: http://10.162.187.154/tests/1050

@alice-suse @xguo @Julie-CAO

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -121,6 +121,8 @@ sub init {
     }
     elsif (get_var('SERIALDEV')) {
         $serialdev = get_var('SERIALDEV');
+        set_var('SERIALDEVORIG', $serialdev);
+        bmwqemu::save_vars();
     }
     else {
         $serialdev = 'ttyS0';


testapi.pm
```diff
@@ -121,6 +121,8 @@ sub init {
     }
     elsif (get_var('SERIALDEV')) {
         $serialdev = get_var('SERIALDEV');
+        set_var('SERIALDEVORIG', $serialdev);
+        bmwqemu::save_vars();
```
okurz: I don't think this is a good idea, also see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800/files#r257496910 please:

1. The backend should not need to save variables which are only relevant in test context. This needs to be handled in tests
2. When coming up with new variable names better use a word separation character, e.g. `_` so `SERIAL_DEV_ORIG` or at least `SERIALDEV_ORIG`
3. Define all new variables in the documentation, the file https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc

But as stated in 1. I do not think this should be here at all and the PR should be closed.

waynechen55: As stated in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800#discussion_r257538430
The original SERIALDEV is overwritten, at least, at one place, for example:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a5371a62e9829efd54b547160111c2699da61900/lib/Utils/Backends.pm#L50
It is not a good design and good practice in the first place. And because of this bad design, I have to save its original value again to another parameter.  Why didn't you prevent this bad design from happening in the first place ?  @okurz

okurz: > Why didn't you prevent this bad design from happening in the first place ? @okurz

:D Because I am not almighty ;)

waynechen55: @okurz The most suitable solution coming to my mind currently is introducing a new parameter called SERIALDEV_AARCH64 or similar or more meaningful. We can not solve this problem all at once, because the original SERIALDEV is already being used for multiple purposes(at least, sshserial and serial console at the same time).  If the original implementation got refined more carefully, there should have been two parameters already, one for sshserial and one for serial console, so I do not need to bother add another one by myself. But in light of the existing fact, and  we do need one more to continue our work, my proposal is keep the original implementation and add another one SERIALDEV_AARCH64 for aarch64 machine serial console usage. Then this PR can be closed, and the other one can be updated with this new proposal. Sounds reasonable ?

okurz: I do not see where aarch64 is special here. I would rather use a dedicated variable for the serial-/ssh-terminal

waynechen55: Some aarch64 use ttyAMA0, some aarch64 use ttyS0, and these machines can not be configured to use other serial consoles. They only have one to use. I will see what I can do. So you suggest one parameter for serial terminal, one for ssh terminal, right ?

alice-suse: I agree with okurz. SERIALDEV is better to be set or changed or saved in test code, just as the way we do for xen serial console setting on x86 testing.

waynechen55: @okurz  I am really willing to hear from you with your opinion. Which of the following suggestions that you support:
1. Introduce another PARAMETER for AARCH64 serial console
2. Change the existing implementation, which means do not save 'sshserial' to SERIALDEV, save it to local variable or save it to new SSHSERIALDEV parameter
3. Your own proposal

foursixnine: @alice-suse Can you point me on how do you do it with XEN? I see the point of @waynechen55 about different consoles for aarch64, thus should be a variable that changes where the console is.

Since IPMI workers are usually set as jump hosts, we can easily put that into a new variable in the workers.ini file. Unless there's a difference in the worker setup that I'm unable to see?

Could you guys provide a bit more context? @alice-suse @waynechen55

waynechen55: @foursixnine For the worker setup, there is no difference. I am really curious about and interested in your "easily put that into a new variable in the workers.ini file". Could you help show me an example ? The workers.ini file and how to use it. 
And you said "Since IPMI workers are usually set as jump hosts", are there any prerequisites for setting up such a jump host ? Hardware and software requirements ? Let's say there are totally 6 openqa-workers need to be established.

okurz: Regarding https://github.com/os-autoinst/os-autoinst/pull/1107#discussion_r258373931 I would go with your proposal 2.

alice-suse:
@foursixnine @okurz @waynechen55 When we were integrating xen tests, we used to work on serialdev/SERIALDEV. Xen is a special case where serial changes during test, which during host installation it is ttyS1, and after booting to xen, the correct serial is xvc0/hvc0 according to different product. So to always get serial output, we changes SERIALDEV accordingly. That's why you see such codes.

* Shall SERIALDEV change? Yes
From my understanding of os-autoinst code and experience, the suggested way is also to ALWAYS keep SERIALDEV correct according to current system state, because some codes query this variable to get the serial to perform serial actions.  So SERIALDEV can be floating, just as VERSION does.

If the original SERIALDEV setting is also needed after it changes, then SAVE it to other variables. For example, host upgrade tests, VERSION is floating as system real version is, and we save the original VERSION to others for future usage. THIS IS BETTER TO BE DONE IN TEST CODE, rather than os-autoinst.

* Where to set SERIALDEV?
If it is HW specific, then better to set in worker.ini, just like other worker settings. If not, but different from ttyS1, set it in testsuites, or machines, or test code according to your specific case.

So I would personally suggest to move this save original value operation to where it needs in test code, actually only a few places in test code changes SERIALDEV.

testapi.pm
waynechen55: Already made the change based on proposal 2.

waynechen55:
> @foursixnine @okurz @waynechen55 When we were integrating xen tests, we used to work on serialdev/SERIALDEV. Xen is a special case where serial changes during test, which during host installation it is ttyS1, and after booting to xen, the correct serial is xvc0/hvc0 according to different product. So to always get serial output, we changes SERIALDEV accordingly. That's why you see such codes.
> 
>     * Shall SERIALDEV change? Yes
>       From my understanding of os-autoinst code and experience, the suggested way is also to ALWAYS keep SERIALDEV correct according to current system state, because some codes query this variable to get the serial to perform serial actions.  So SERIALDEV can be floating, just as VERSION does.
>       If the original SERIALDEV setting is also needed after it changes, then SAVE it to other variables. For example, host upgrade tests, VERSION is floating as system real version is, and we save the original VERSION to others for future usage. THIS IS BETTER TO BE DONE IN TEST CODE, rather than os-autoinst.
> 
>     * Where to set SERIALDEV?
>       If it is HW specific, then better to set in worker.ini, just like other worker settings. If not, but different from ttyS1, set it in testsuites, or machines, or test code according to your specific case.
> 
> 
> So I would personally suggest to move this save original value operation to where it needs in test code, actually only a few places in test code changes SERIALDEV.
> 
> @coolo Please correct me if I am wrong.

You can have a look at https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800 firstly. Then can comment with your concern there.

waynechen55:
Let's move any following discussion to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800, in which the actual change has been and would have been made. This PR#1107 had already approached the end of its life.  Thanks. @All

okurz:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6800 merged. I hope you don't mind if I close this PR then – as you suggested.

PR #1111, opened 2019-02-20
Configure bootindex for first path only

czerw:
Fix poo#48155: When is MULTIPATH variable enabled, qemu command line
options contain bootindex for each device and qemu will not start.
Option bootindex needs to be configured only once.

Progress : https://progress.opensuse.org/issues/48155
Verification: http://10.100.12.105/tests/1546

### DIFF
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -111,8 +111,9 @@ sub gen_cmdline {
         if (defined $path->controller) {
             push(@params, 'bus=' . $path->controller->id . '.0');
         }
-        $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex);
-        $self->_push_ifdef(\@params, 'serial=',    $self->serial);
+        # Configure bootindex only for first path
+        $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex) if (($path->id eq 'path0') || (!$path->id));
+        $self->_push_ifdef(\@params, 'serial=', $self->serial);
         push(@cmdln, ('-device', join(',', @params)));
     }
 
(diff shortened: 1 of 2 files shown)


czerw:
@richiejp Could you please check?

OpenQA/Qemu/DriveDevice.pm
```diff
@@ -111,7 +112,8 @@ sub gen_cmdline {
         if (defined $path->controller) {
             push(@params, 'bus=' . $path->controller->id . '.0');
         }
-        $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex);
+	# Configure bootindex only for first path
+        $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex) if ($current_path++ == 1);
```
kalikiana: Is the first path always the bootable one? Do we know that `@paths` is always ordered that way?

Small note wrt coding style: code should only be indented with spaces.

czerw: yes, i suppose  according to `my $paths = @{$self->paths} < 1 ? [OpenQA::Qemu::DrivePath->new()] : $self->paths;`, but @richiejp can comment it further

richiejp: You could check that `$path->id == '0'` instead so that the order of `@paths` is not significant (assuming it starts at zero, I can't remember).

I don't think it matters which path to a particular device is chosen as bootindex zero (so long as it is deterministic) unless we are testing QEMU itself.

czerw: code updated to `$path->id eq 'path0'`, verification run in description updated as well

czerw:
checking that travis failure

OpenQA/Qemu/DriveDevice.pm
czerw: extended condition with empty path-id  to fix travis failures for non-multipath scenarios

richiejp:
LGTM

PR #1112, opened 2019-02-22
Use rsync with checksum to copy installation mediums

rwx788:
rsync always checks if copied files' checksum matches with original
file, but we might have the case that when we retrigger disk creation,
there is medium with matching modification date and size, but with
different content. Decision based on checksum is more reliable.

See [poo#41414](https://progress.opensuse.org/issues/41414).

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -384,7 +384,7 @@ sub add_disk {
                       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
                       'done;' .
                       'else ' .
-                      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
+                      "rsync --checksum -av /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
                       'fi;'
                 );
                 $vmware_chan->send_eof;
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -384,7 +384,7 @@ sub add_disk {
                       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
                       'done;' .
                       'else ' .
-                      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
+                      "rsync --checksum -av /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
```
mnowaksuse: There's a misunderstanding. There's no `rsync` on ESXi and it's not an trivial enterprise to get it there.

foursixnine: Ahh It sounded too easy to be true :)

rwx788: I had a hope =)

foursixnine:
@rwx788 We can build on the shoulders of giants and use the cache + test distribution for this, maybe we can grab a coffee later and talk about this

rwx788:
@foursixnine I've checked there, problem is that we have approached the problem from one angle. We have checksum set in the job settings, so I guess I will just inject it to the distribution initialization.

foursixnine:
That's the first step :)

PR #1113, opened 2019-02-23
Extract common run_cmd to prevent duplication

okurz:
Verification run: http://lord.arch.suse.de/tests/2037

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -72,13 +72,13 @@ sub do_stop_vm {
         bmwqemu::diag "Destroying $vmname virtual machine";
         if (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
             my $ps = 'powershell -Command';
-            $self->run_cmd("$ps Stop-VM -Force -VMName $vmname -TurnOff");
-            $self->run_cmd("$ps Remove-VM -Force -VMName $vmname");
+            $self->run_ssh_cmd("$ps Stop-VM -Force -VMName $vmname -TurnOff");
+            $self->run_ssh_cmd("$ps Remove-VM -Force -VMName $vmname");
         }
         else {
             my $libvirt_connector = get_var('VMWARE_REMOTE_VMM', '');
-            $self->run_cmd("virsh $libvirt_connector destroy $vmname");
-            $self->run_cmd("virsh $libvirt_connector undefine --snapshots-metadata $vmname");
+            $self->run_ssh_cmd("virsh $libvirt_connector destroy $vmname");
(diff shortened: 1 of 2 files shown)


okurz:
Need to remove "[WIP] Add debugging script for remote ssh command execution" again

okurz:
> Makes sense. So what's missing form your point of view (WIP is still in title)?

Manual integration test. Let's see how http://lord.arch.suse.de/tests/2037 turns out.

> When I previously looked at this I spotted a test script which you seem to have removed. 
> Maybe you can actually share this somehow? On the first look it seemed useful.

Well, I used it only to debug but I made it available on https://gist.github.com/okurz/1e82cdecf6e5c478119a58964117e119 if you are interested.

consoles/sshVirtsh.pm
```diff
@@ -505,17 +506,20 @@ sub add_disk {
     return;
 }
 
+sub virsh {
+    my $libvirt_connector = get_var('VMWARE_REMOTE_VMM', ' ');
+    return "virsh$libvirt_connector";
```
mnowaksuse: `VMWARE_REMOTE_VMM`, if defined, may not start with space:
```suggestion
    return "virsh $libvirt_connector";
```

backend/svirt.pm
```diff
@@ -72,13 +72,13 @@ sub do_stop_vm {
...
+            $self->run_ssh_cmd("$ps Stop-VM -Force -VMName $vmname -TurnOff");
+            $self->run_ssh_cmd("$ps Remove-VM -Force -VMName $vmname");
         }
         else {
-            my $libvirt_connector = get_var('VMWARE_REMOTE_VMM', '');
-            $self->run_cmd("virsh $libvirt_connector destroy $vmname");
-            $self->run_cmd("virsh $libvirt_connector undefine --snapshots-metadata $vmname");
+            my $libvirt_connector = get_var('VMWARE_REMOTE_VMM', ' ');
+            $self->run_ssh_cmd("virsh$libvirt_connector destroy $vmname");
+            $self->run_ssh_cmd("virsh$libvirt_connector undefine --snapshots-metadata $vmname");
```
mnowaksuse: Space as well?

consoles/sshVirtsh.pm
okurz: true, updated. I hope I got it right.

pevik:
I don't like the naming `run_cmd` and `run_ssh_cmd`: `run_cmd` is also working with `ssh` via ssh channel.
\+ `run_cmd` uses actually only ssh channel, so it might be better to pass it instead of `ssh`

PR #1114, opened 2019-02-25
Add test for logging to t/Makefile.am

Martchus:


Martchus:
Seems like the `15-logging.t` just wasn't flagged as executable.

okurz:
failed now with `FAIL: 15-logging.t` and no further information in the logs.

Martchus:
But now it passes :-)

PR #1116, opened 2019-02-26
Run tidy on the correct directory

rwx788:
In #1097 we got the change to fix Perl::Tidy version check, but it runs
on the wrong directory in case of tests repo.

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -23,7 +23,7 @@ if [ "$perltidy_version_found" != "$perltidy_version_expected" ]; then
     exit 1
 fi
 
-cd $dir
+cd "${0%/*}/.."
 
 # just to make sure we are at the right location
 test -e tools/tidy || exit 1


rwx788:
@foursixnine @Martchus it's kinda urgent, as travis checks wrong thing for tests repo =)))

okurz:
"kinda urgent" to literally revert https://github.com/rwx788/os-autoinst/commit/ef838b427d5a39f00b00315a54c3ee41d46f9605#diff-c1b043a4f3b172277d8aef7852a95c26 without evening pinging @pevik . Aren't we breaking again the problem that this commit tried to fix?

@pevik please take a look

pevik:
@okurz @rwx788 @foursixnine Sorry for silly error causing problems.
I'd personally prefer `cd "$(dirname $0)/..` (IMHO more readable than `${0%/*}/..`), but don't dare to open another PR :).

rwx788:
@pevik actually, you've fixed the issue I had as I run openQA in container and don't have all dependencies installed. `cd "$(dirname $0)/..` is indeed more readable, I was not confident enough to change the way we get current directory of the executable and not directory it was called from =)

rwx788:
@okurz so you assume that I didn't get what Petr tried to do in his PR and make something which breaks it and doesn't fix anything? You were wrong then ;)

pevik:
@rwx788 What dependency do you miss in your container?
BTW I dared to add second PR on tidy, this one should be without bug (https://github.com/os-autoinst/os-autoinst/pull/1118).

okurz:
@rwx788 Well, I assumed that all of us make mistakes so it's better to build upon the shoulders of giants, i.e. the authors of the git history ;) No offense intended, just suggesting to invite others to take a second look as I have seen many examples in our recent history already where people missed what was done in before and break the stuff of others again – unintendedly of course.

rwx788:
@pevik sorry for the confusion, I've just meant that I hot-patched tidy script so it doesn't break when cannot find expected cpan dependencies file. #1118 makes it more readable, thanks ;)

pevik:
Well, thanks not for fix itself, but also for pinging me, so I did second review of my work :).

PR #1120, opened 2019-03-01
Improve feedback on "half-open socket"

okurz:
The die-message for "half-open socket" is replaced by an exception and
extended with hints as well as the output of the actual console that is the
culprit.

Verification run: http://lord.arch/tests/2039/file/autoinst-log.txt shows the
additional output in the log files with helpful pointers into what went
wrong and how to fix it.

Related progress issue: https://progress.opensuse.org/issues/48260


okurz:
This should be an improvement even though I would like to turn it into a test failure instead of incomplete as the backend did nothing wrong here.

PR #1122, opened 2019-03-06
Prevent recording empty soft failure test detail

Martchus:
See https://progress.opensuse.org/issues/48758

---

os-autoinst actually recorded 2 test details for one soft-failure. The first one was empty and causing the issue linked above.

I extended the unit tests and did a local verification run. I've also tested whether the extended unit test fails without that change.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -27,6 +27,7 @@ use POSIX;
 use testapi  ();
 use autotest ();
 use MIME::Base64 'decode_base64';
+use Mojo::File 'path';
 
 my $serial_file_pos = 0;
 
@@ -417,9 +418,7 @@ sub next_resultname {
 sub write_resultfile {
     my ($self, $filename, $output) = @_;
 
-    open my $fh, '>', bmwqemu::result_dir() . "/$filename";
-    print $fh $output;
-    close $fh;
+    path(bmwqemu::result_dir(), $filename)->spurt($output);
 }
 
 =head2 record_resultfile
--- a/basetest.pm
+++ b/basetest.pm
@@ -414,6 +414,14 @@ sub next_resultname {
     }
 }
 
+sub write_resultfile {
+    my ($self, $filename, $output) = @_;
+
+    open my $fh, '>', bmwqemu::result_dir() . "/$filename";
+    print $fh $output;
+    close $fh;
+}
+
 =head2 record_resultfile
 
     $self->record_resultfile($title, $output [, result => $result] [, resultname => $name]);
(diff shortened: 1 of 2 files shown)


basetest.pm
```diff
@@ -449,17 +455,19 @@ sub record_serialresult {
     $output .= "# Result:\n";
     $output .= "$string\n";
     $self->record_resultfile('wait_serial', $output, result => $res);
-    return;
+    return undef;
```
aaannz: Just wondering, why is this needed?

coolo: in short: because @kraih said so :)

Martchus: I assume the intention behind the `return;` was to always return `undef` (regardless of the last expression). However, using only `return;` would actually return an empty list if the function is executed where an array is wanted. We likely don't want that behavior.

aaannz: That's why I asked. I always preferred to call plain `return;` as then you get empty list, where with `return undef;` you get a list with one `undef` element IIUC. And if I greped source correctly I found that `record_serialresult` and `record_soft_failure_result` are never called in tests nor is their return actually used. I see only `register_extra_test_results` is used somewhere in other returns.
So I was wondering if there is something new I missed.

kraih: This change is correct, bare `return;` is not considered good Perl anymore. In recent years we've seen too many high profile security issues with functions/methods that behave differently depending on the context in which they were used. This is usually caused by `wantarray`, but bare `return;` has similar risks. So just don't make functions/methods that mix contexts, return a list or a scalar, never both. (The Perl Best Practices book does recommend bare `return;`, but the author now also considers this an error.)

Martchus: @aaannz Yes, the return value isn't actually used here. But I'd like to introduce the currently preferred best practices so hopefully they are copied/used more often.

kraih: The [Modern Perl Best Practices](https://learning.oreilly.com/videos/modern-perl-best/9781491910719) training also mentions this.

aaannz: @kraih Perfect. This is what I didn't know and I still lived in the old world where `return;` was recommended. Thanks.

Martchus:
Any feedback on the actual change?

basetest.pm
```diff
@@ -414,6 +414,14 @@ sub next_resultname {
     }
 }
 
+sub write_resultfile {
+    my ($self, $filename, $output) = @_;
+
+    open my $fh, '>', bmwqemu::result_dir() . "/$filename";
+    print $fh $output;
+    close $fh;
```
kraih: This would look a bit nicer with `Mojo::File`. (At first i wanted to complain about the lack of exception handling here... but then noticed the `autodie` at the top of the file 😃)
```perl
use Mojo::File 'path';
...
path(bmwqemu::result_dir(), $filename)->spurt($output);
```

Martchus: os-autoinst depends on Mojo anyways so I'll change that.

PR #1123, opened 2019-03-07
Add QEMU_HUGE_PAGES_PATH option

ggardet:


ggardet:
The tidy error does not look right. Code is barely readable with tidy suggestion. WDYT?

Martchus:
Not sure whether there's something like `// clang-format off` for pertidy. But I would format it differently anyways:   
```
if (my $path = $vars->{QEMU_HUGE_PAGES_PATH}) {
    sp('mem-prealloc');
    sp('mem-path', $path);
}
```

Because having the same if twice and repeating the hash value even 3 times is just not very nice.

ggardet:
@Martchus updated with your suggestion.

ggardet:
> Not sure about the options itself but the per code looks good.

It is in use since months in o3, with QEMU_APPEND="mem-prealloc -mem-path /dev/hugepages/" so it should be fine. :)

I added a test for this option now.

PR #1126, opened 2019-03-12
Improve error handling when using Net::SSH2 library

Martchus:
The library unfortunately does not throw any exceptions. So the return value of most function invocations must be checked manually and the error handled.

I made it die in most cases. The notable exception is the `get_cmd_output` function of the sshVirsh console where I implemented a reconnect to prevent test failures due to connection issues (see https://progress.opensuse.org/issues/44771 for that).

This is still completely untested and hence WIP. I'll do some tests on my own but it would be nice if someone could give it a try, too.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1204,14 +1204,15 @@ sub start_ssh_serial {
 
     $self->stop_ssh_serial;
 
-    $self->{serial} = $self->new_ssh_connection(%args);
-    my $chan = $self->{serial}->channel();
-    die "No channel found" unless $chan;
-    $self->{serial_chan} = $chan;
+    my $ssh  = $self->{serial}      = $self->new_ssh_connection(%args);
+    my $chan = $self->{serial_chan} = $ssh->channel();
+    if (!$chan) {
+        $ssh->die_with_error("Unable to establish SSH channel for serial console");
+    }
     $chan->blocking(0);
     $chan->pty(1);
     $self->{select}->add($self->{serial}->sock);
-    return $chan;
+    return ($ssh, $chan);
 }
 
 sub check_ssh_serial {
--- a/backend/spvm.pm
+++ b/backend/spvm.pm
@@ -67,14 +67,14 @@ sub run_cmd {
     $hostname ||= get_required_var('NOVALINK_HOSTNAME');
     $password ||= get_required_var('NOVALINK_PASSWORD');
 
-    $self->{ssh} = $self->new_ssh_connection(
+    my $ssh = $self->{ssh} = $self->new_ssh_connection(
         hostname => $hostname,
         password => $password,
         username => get_var('NOVALINK_USERNAME', 'root'));
-    my $chan = $self->{ssh}->channel();
-    $chan->exec($cmd);
+    my $chan = $ssh->channel() || $ssh->die_with_error();
+    $chan->exec($cmd) || $ssh->die_with_error();
     get_ssh_output($chan);
-    $chan->send_eof;
+    $chan->send_eof();
     my $ret = $chan->exit_status();
     bmwqemu::diag "Command executed: $cmd, ret=$ret";
     $chan->close();
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -115,8 +115,8 @@ sub get_ssh_output {
 sub run_cmd {
     my ($ssh, $cmd) = @_;
 
-    my $chan = $ssh->channel();
-    $chan->exec($cmd);
+    my $chan = $ssh->channel() || $ssh->die_with_error("Unable to create SSH channel for executing \"$cmd\"");
+    $chan->exec($cmd) || $ssh->die_with_error("Unable to execute \"$cmd\"");
     get_ssh_output($chan);
     $chan->send_eof;
     my $ret = $chan->exit_status();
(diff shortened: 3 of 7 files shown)


Martchus:
Note that the problem regarding the error handling with older versions of libssh2 (discussed [in the previous PR](https://github.com/os-autoinst/os-autoinst/pull/874#issuecomment-338044906)) shouldn't matter much.

The documentation says:

> That means that after any of those methods fails, "error" would not return the real code but just some bogus result when an older version of the library is used.

So we are able to detect errors also in previous version. Only the error code for some failures is not set correctly. So the usefulness of some error messages is impaired but at least the error is handled at all. That is already a plus compared to the current situation where errors are just ignored and it dies later with a not so nice message like "Can't call method "exec" on an undefined value".

Martchus:
Seems like I've already ran into an error case here:

```
# Test died: {
  'cmd' => 'backend_proxy_console_call',
  'args' => [],
  'console' => 'svirt',
  'function' => 'define_and_start',
  'json_cmd_token' => 'gmYJZFaj'
}
Unable to start serial console: Operation would block (-37 LIBSSH2_ERROR_EAGAIN) at /hdd/openqa-devel/repos/os-autoinst/backend/svirt.pm line 272.
```

Not sure why blocking operations are forbidden, though.

Martchus:
When I make the channel blocking the error is gone but then of course it blocks at this point and the test never actually starts. So making it blocking is not an option. Any ideas how to handle this?

By the way, I assume the failing `$chan->exec($command);` at the end of `start_serial_grab` has always been failing. It just hasn't been discovered because the error hasn't been handled so far.

Martchus:
For now I've changed the `die` to a `bmwqemu::diag`. The serial console also seems to work later on so it is not that critical.

backend/baseclass.pm
```diff
@@ -1204,14 +1204,15 @@ sub start_ssh_serial {
...
+    my $ssh  = $self->{serial}      = $self->new_ssh_connection(%args);
+    my $chan = $self->{serial_chan} = $ssh->channel();
+    if (!$chan) {
+        $ssh->die_with_error("Unable to establish SSH channel for serial console");
+    }
     $chan->blocking(0);
     $chan->pty(1);
     $self->{select}->add($self->{serial}->sock);
-    return $chan;
+    return ($ssh, $chan);
```
pevik: I guess you probably conflict with https://github.com/os-autoinst/os-autoinst/pull/1109. Good idea, but we both (+ @okurz did as well) are working on SSH code refactoring (addressing similar issues, but not quite the same). After rebasing https://github.com/os-autoinst/os-autoinst/pull/1113 I hope I won't need to rebase this one as well :).

BTW, during https://github.com/os-autoinst/openQA/pull/2003 I wasn't sure, whether returning `$ssh` wouldn't be enough in functions creating SSH connection. I mean anybody can call `$ssh->channel()`, so why to complicate returning with list. But probably having a test on `$chan` make justify it. 
\+1 for using `die_with_error()` from `Net::SSH2`, I was also thinking about it.

Martchus: This *must* return `$ssh` to be useful. Otherwise gathering error information e.g. in case of a failed `$chan->exec` is not possible. I can help rebasing if required.

Note that this PR is supposed to supersede the one created by @okurz.

pevik: Sorry, I meant return *only* `$ssh` (without `$chan`), sure `$ssh` should returned. My other concerns is speed (is it `$chan` a new connection or reusing already existed one; because speed is one of 2 benefits for serial console) + whether we need sometimes other SSH instance than `$self->{ssh}`).

> Note that this PR is supposed to supersede the one created by @okurz.

I'm changing that code differently in https://github.com/os-autoinst/os-autoinst/pull/1109 :(. Any objections?

Martchus: Feel free to change it to return only `$ssh`.

pevik: If you don't require it, I'd prefer to keep it as it is in order to have it merged and address these issues later. This PR does quite a lot of cleanup already :).

Martchus:
I've ran some tests locally: VMWare seems to work in general (it failed in `prepare_test_data` on some `curl` command, though) and Hyper-V even passed. Xen tests fail for me with:

screenshot_20190319_140239

Not sure whether this is related to the error handling, though. Any ideas?

KVM via svirt is also broken on my machine due to an issue which is likely unrelated (hostname isn't passed correctly) so I couldn't test it so far. @foursixnine had the same problem on his instance.

mnowaksuse:
@Martchus If there are links to failed jobs I am happy to help.

Martchus:
I've also ran an spvm test now and ran into the [same problem as with the svirt backend](https://github.com/os-autoinst/os-autoinst/pull/1126#issuecomment-472055212). I worked around it in the same way by making the error non-fatal and just logging it. The spvm test still failed in `logs_from_installation_system` but that is likely not related.

I've also checked whether the version of `Net::SSH2` deployed on the OSD workers already has the methods I'm using here (and it has).

So I'd say it is ready to be merged unless e.g. @mnowaksuse wants to do some further testing.

Martchus:
> @Martchus If there are links to failed jobs I am happy to help.

Thanks. Unfortunately test results are just on my laptop which I'm shutting down in an hour. The failing `curl` commands e.g. `# Test died: command 'curl -L -v -f http://127.0.0.1:20033/QFUGmZgtWnVmyXK1/data > test.data' failed at /hdd/openqa-devel/openqa/share/tests/sle/tests/console/prepare_test_data.pm line 24.` are likely just because my local setup is not sufficient and very likely unrelated. The spvm test I've [cloned from OSD](https://openqa.suse.de/tests/2549198) fails here with `# Test died: no candidate needle with tag(s) 'install-shell' matched` in `logs_from_installation_system`. Maybe it was just typing too slow.

Martchus:
@mnowaksuse Did you see the links to my test runs I've sent you via IRC?

PR #1127, opened 2019-03-14
Move virtio rng option _after_ Ethernet setup to keep it as eth0

ggardet:
Related ticket: https://progress.opensuse.org/issues/49292

Tested locally on aarch64 with `isotovideo`.


ggardet:
Could it be merged, please?

PR #1128, opened 2019-03-15
Prevent rendering unknown results

Martchus:
If no screenshot is available at the point one is taken os-autoinst previously rendered an empty/invalid result.

This change prevents that which should also fix the question marks in the openQA web UI (or rendering issues in previous version of the web UI), e.g. http://cfconrad-vm.qa.suse.de/tests/3541#step/sle15_workarounds/1.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -249,7 +249,6 @@ sub record_screenfail {
     return $result;
 }
 
-# for interactive mode
 sub remove_last_result {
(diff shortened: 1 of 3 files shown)


Martchus:
For now I've changed the `die` to a `bmwqemu::diag`. The serial console also seems to work later on so it is not that critical.

PR #1129, opened 2019-03-20
Use always $bmwqemu::vars{NEEDLES_DIR} directly

Martchus:
This variable is initialized at the beginning to a default path if not specified explicitely. So let's just use that and get rid of the function `default_needle_dir`.

That should prevent https://progress.opensuse.org/issues/48452.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -247,7 +247,7 @@ OpenQA::Isotovideo::Utils::checkout_git_repo_and_branch('CASEDIR');
 # multiple distributions or flavors in one repository.
 $bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
 
-$bmwqemu::vars{NEEDLES_DIR} //= $bmwqemu::vars{PRODUCTDIR} . '/needles';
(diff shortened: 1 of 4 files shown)


Martchus:
Tests have passed despite being still shown as pending.

isotovideo
```diff
@@ -247,7 +247,7 @@ OpenQA::Isotovideo::Utils::checkout_git_repo_and_branch('CASEDIR');
 # multiple distributions or flavors in one repository.
 $bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
 
-$bmwqemu::vars{NEEDLES_DIR} //= $bmwqemu::vars{PRODUCTDIR} . '/needles';
+$bmwqemu::vars{NEEDLES_DIR} //= "$bmwqemu::vars{PRODUCTDIR}/needles/";
```
coolgw: Minor comment: why add extra "/" compare with origin line? 
BTW: When this PR can be merged? I need this PR merge quickly and i can validate some issue on OpenQa env.

Martchus: I removed the extra "/". No usage of that variable seems to expect it.

The only reason this has not been merged is that nobody gave a positive review.

PR #1130, opened 2019-03-21
Update to handle xz compressed images

ccret:
This change uncompresses the image and sets the name of the image to the uncompressed one.

Progress ticket: https://progress.opensuse.org/issues/49523
Validation runs: 
  - Compressed image: http://ccret.suse.cz/tests/3184, http://ccret.suse.cz/tests/3181
  - Uncompressed image: http://ccret.suse.cz/tests/3183

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -412,6 +412,10 @@ sub add_disk {
             }
             else {
                 $self->run_cmd(sprintf("rsync -av '$args->{file}' '$basedir/%s'", $file_basename)) && die 'rsync failed';
+                if ($file_basename =~ /(.*)\.xz$/) {
+                    $self->run_cmd(sprintf("nice ionice unxz -f -k '$basedir/%s'", $file_basename)) unless -e "$basedir$1";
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -412,6 +412,10 @@ sub add_disk {
             }
             else {
                 $self->run_cmd(sprintf("rsync -av '$args->{file}' '$basedir/%s'", $file_basename)) && die 'rsync failed';
+                if ($file_basename =~ /(.*)\.xz/) {
+                    $self->run_cmd(sprintf("nice ionice unxz -f -k '$basedir/%s'", $file_basename)) unless (-e $basedir . $1);
```
coolo: who/what will delete this later?

ccret: The image gets uncompressed inside /var/lib/libvirt/images so I don't think I have to remove it from there since it's done through a service on the host. If this is not what you asked I am not sure I understood the question.

ccret:
Updated.
New validation run: http://ccret.suse.cz/tests/3191

PR #1131, opened 2019-03-22
[wip] Use public key authentication

foursixnine:
Attempt to allow key based authentication for ssh based backends.

There's still the VNC part of this that I haven't quite found the way how to make it work, as it's still using password auth.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1169,7 +1169,7 @@ sub new_ssh_connection {
     my ($self, %args) = @_;
     $args{username} ||= 'root';
 
-    my $ssh = Net::SSH2->new;
+    my $ssh             = Net::SSH2->new;
     my $privatekey_path = '/home/foursixnine/.openqa';
     my $publickey_path  = '/home/foursixnine/.openqa.pub';
     # Retry 5 times, in case of the guest is not running yet
@@ -1179,10 +1179,10 @@ sub new_ssh_connection {
             # Attempt always key authentication
             eval {
                 $ssh->auth(
-                    username => $args{username},
-                    password => $args{password},
+                    username   => $args{username},
+                    password   => $args{password},
                     privatekey => $privatekey_path,
-                    publickey => $publickey_path,
+                    publickey  => $publickey_path,
                     passphrase => undef
                 );
             };
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1170,24 +1170,40 @@ sub new_ssh_connection {
     $args{username} ||= 'root';
 
     my $ssh = Net::SSH2->new;
-
+    my $privatekey_path = '/home/foursixnine/.openqa';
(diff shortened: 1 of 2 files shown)


backend/baseclass.pm
```diff
@@ -1170,24 +1170,40 @@ sub new_ssh_connection {
     $args{username} ||= 'root';
 
     my $ssh = Net::SSH2->new;
-
+    my $privatekey_path = '/home/foursixnine/.openqa';
```
foursixnine: This would be a worker setting

consoles/console.pm
```diff
@@ -66,13 +67,15 @@ sub screen {
...
-    my ($self, $username, $host, $gui) = @_;
-
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
-
-    if ($gui) {
-        $sshopts = "-X $sshopts";
+    my ($self, $username, $host, $gui, $privatekey) = @_;
+    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no";
+    $sshopts   .= " -o PubkeyAuthentication=no" unless defined $privatekey;
+    if (defined $privatekey && -e $privatekey){
```
foursixnine: @coolo @mnowaksuse I'm tempted to use a job variable (say `VIRSH_SSH_PRIVATEKEY` and use that if it's set instead of expecting the private key to be passed somehow (Something I haven't figured out yet how to)

mnowaksuse:
> There's still the VNC part of this that I haven't quite found the way how to make it work, as it's still using password auth.

Perhaps make it password-less? I don't see much security in this.

pevik:
@foursixnine What is a benefit of pub key authentication for our testing? Is it problematic to pass the password (I guess it's working everywhere, isn't it?) or security (why bother for testing)?
It'd be nice to briefly explain it in commit message.

foursixnine:
> @foursixnine What is a benefit of pub key authentication for our testing? Is it problematic to pass the password (I guess it's working everywhere, isn't it?) or security (why bother for testing)?
> It'd be nice to briefly explain it in commit message.

Actually it's a byproduct of another issue I was working with, regarding timeouts on svirt backends, main benefit here is just to not need to input the password, and avoid the risk of having keys missing when there's high network load (which correlates with connection timeout messages) which was an issue in the past. You can give a look to [poo#41504](https://progress.opensuse.org/issues/41504)

Security wise, it's not my concern atm but rather stability and safety that the connection can be stablished...

pevik:
@foursixnine make sense, thanks for an explanation. I'd still explain the motivation in commit message or pointed out poo#41504. (if you search in git, you usually don't check PR on the web)

foursixnine:
@pevik makes sense tbh :) I'll do that at a later point

okurz:
please reopen when you continue the work. Thanks. Closing to clean up old open PRs.

PR #1132, opened 2019-03-26
Make failing exec in sshIucvconn non-fatal

Martchus:
See https://progress.opensuse.org/issues/49676

### DIFF
--- a/consoles/sshIucvconn.pm
+++ b/consoles/sshIucvconn.pm
@@ -36,11 +36,12 @@ sub connect_remote {
 
     # start agetty to ensure that iucvconn is not killed
     my $chan = $ttyconn->channel() || $ttyconn->die_with_error();
-
     $chan->blocking(0);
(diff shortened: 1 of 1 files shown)


Martchus:
It seems that the svirt tests succeed in production with that workaround, e.g. https://openqa.suse.de/tests/2734790 (look for "Unable to grab serial console" in the log). So I'm confident this change will help here, too.

In the mentioned svirt test we also see the impaired error message because the libssh2 version in production is too old. We only get `Unable to grab serial console at this point: 0` instead of `Unable to grab serial console at this point: Operation would block (-37 LIBSSH2_ERROR_EAGAIN)` but it is better than nothing.

Martchus:
@Soulofdestiny Looks good?

@nicksinger I guess you should deploy this as soon as it is merged and packaged.

pevik:
LGTM

PR #1133, opened 2019-03-28
make get_var_array return values consistent

badboywj:
if $var value exists, return a array referrence "\@var", but return a string if $var doesn't exist. This is ugly. I just make the returned value consistent, both array referrence. And this alsoconforms to the document of "get_var_array".

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -739,7 +739,8 @@ Return the given variable as array reference (split variable value by , | or ; )
 sub get_var_array {
     my ($var, $default) = @_;
     my @vars = split(/,|;/, $bmwqemu::vars{$var} || '');
-    return $default if !@vars;
+    my @default = split(/,|;/, $default || '');
+    return \@default if !@vars;
     return \@vars;
 }
 


Martchus:
You need to run `tools/tidy`. Besides, did you ensure that any usage of `get_var_array` in our testsuite is updated?

badboywj:
> You need to run `tools/tidy`. Besides, did you ensure that any usage of `get_var_array` in our testsuite is updated?

Sorry, I don't understand, why need to run "tools/tidy"? I commit the code on the
github webpage, I don't know how to run "tools/tidy", can you help me?
I make sure we don't need to update anything after accepting this change of "get_var_array".

Martchus:
In this repository there's a directory `tools` with a script called `tidy` in it. You need to run it to format the perl code according our conventions.

badboywj:
> In this repository there's a directory `tools` with a script called `tidy` in it. You need to run it to format the perl code according our conventions. I'm afraid you need to checkout the repo locally for that any maybe install missing dependencies like `perl-Perl-Tidy`.

OK, I see. Thank you very much

PR #1134, opened 2019-03-28
make get_var_array return values consistent

badboywj:
if $var value exists, return a array referrence "@var", but return a string if $var doesn't exist. This is ugly. I just make the returned value consistent, both array referrence. And this alsoconforms to the document of "get_var_array".


badboywj:
make sure that we don't need to update anything after accepting this patch of get_var_array.

Martchus:
The same tidy problem as mentioned before. And please don't create multiple PRs for the same thing. It is possible to re-open PRs and to force-push new changes.

Martchus:
And a small nit-pick: The commit message needs to start with a capital letter by our conventions.

PR #1135, opened 2019-03-28
Allow checking journal for known bugs

asdil12:
https://progress.opensuse.org/issues/46988


coolo:
I don't want to have package specific details in the test api, find a way to hook this in the distribution.

coolo:
(and yes, systemd is a package)

coolo:
Actually I see 0 reason to have this in the testapi. it's plainly in the test itself.

Martchus:
I'm afraid @coolo is right. Or prevents you something from having this in the test distribution?

PR #1136, opened 2019-03-28
Fix force_soft_failure

asdil12:
Allow overwriting the test result

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -320,11 +320,12 @@ sub post_run_hook {
 sub run_post_fail {
     my ($self, $msg) = @_;
     return if $bmwqemu::vars{_SKIP_POST_FAIL_HOOKS};
+    $self->{result}                 = 'fail' if $self->{result};
     $self->{post_fail_hook_running} = 1;
     eval { $self->post_fail_hook; };
     bmwqemu::diag("post_fail_hook failed: $@") if $@;
     $self->{post_fail_hook_running} = 0;
-    $self->fail_if_running();
+    autotest::set_current_test(undef);
     die $msg . "\n";
 }
 


asdil12:
Normal fail: http://artemis.suse.de/tests/1274
Forced softfail: http://artemis.suse.de/tests/1271#step/prepare_test_data/42

Seems to work fine.

asdil12:
The only thing that I found is that the overall status will still be set to *failed*:
http://artemis.suse.de/tests/1287

@okurz WDYT? Is this expected? The documentation doesn't cover this one.

basetest.pm
```diff
@@ -320,11 +320,12 @@ sub post_run_hook {
 sub run_post_fail {
     my ($self, $msg) = @_;
     return if $bmwqemu::vars{_SKIP_POST_FAIL_HOOKS};
+    $self->{result}                 = 'fail' if $self->{result};
```
SergioAtSUSE: you added the result 'fail' here. Maybe you can use 'softfail'.

See https://github.com/os-autoinst/os-autoinst/blob/2bd444f4840a223602021e564d092fd6dc4ac69b/basetest.pm#L460

asdil12: I don't want to change the existing behaviour.
I just want to allow overwriting the result from post fail hook.

asdil12:
@coolo You can already merge - the overall status issue can be discussed and improved later - maybe with a parameter `force_overall_result`.

okurz:
Sorry, I'm not really getting what you want to achieve here. I guess unit tests would help to explain this :)

asdil12:
@okurz At the moment `force_soft_failure` will not set the result of a test module to `softfail` insted of `fail` as the result will be overwritten to `fail` a few lines later.
My patch will prevent this by executing that line earlier so that an overwrite via `force_soft_failure` will have an effect.

okurz:
ok, then please add a test for that

okurz:
@asdil12 can you please add tests?

PR #1137, opened 2019-03-28
Print error message returned by VNC server if security types missing

Martchus:
See https://progress.opensuse.org/issues/49793

---

So we would get at least a more helpful error message. It is WIP because I couldn't test this so far. I'd grab an svirt host from OSD to test this (when one is free).

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -225,7 +225,11 @@ sub _handshake_security {
             $number_of_security_types = unpack('C', $number_of_security_types);
         }
         if ($number_of_security_types == 0) {
-            die 'Error authenticating';
+            my $generic_error_message = 'Error authenticating: server returned no "security types"';
+            $socket->read(my $error_message_length, 4) || die $generic_error_message;
+            $error_message_length = unpack('N', $error_message_length) || die $generic_error_message;
+            $socket->read(my $error_message, $error_message_length) || die $generic_error_message;
+            die "Error authenticating: $error_message";
         }
 
         my @security_types;


consoles/VNC.pm
```diff
@@ -225,7 +225,11 @@ sub _handshake_security {
             $number_of_security_types = unpack('C', $number_of_security_types);
         }
         if ($number_of_security_types == 0) {
-            die 'Error authenticating';
+            my $generic_error_message = 'Error authenticating: server returned not "security types"';
```
kalikiana: What does that mean? Maybe it's a typo but my head can't really parse the message

Martchus: It should be no instead of not. I actually already fixed that but forgot to push the amended commit.

Martchus:
Now I'm only getting `Error authenticating: server returned no "security types"`. To it didn't work or the server simply didn't return an error message in my test.

Martchus:
After reading RFC 6143 and the documentation of the `unpack` function again I'd say the code is supposed to work and the VNC on s390 is so broken that it also doesn't provide an error message. But I still couldn't do a real test to receive the error message.

Despite I'd say there's no risk merging this because in the case of this error the test dies anyways.

If you think it is worth it I could try to write a basic unit test for `VNC.pm` and cover receiving the error message.

coolo:
What makes you think there comes an error message after no security types?

coolo:
I wouldn't try to read anything from a broken VNC server - but trying to make this a test failure sounds like the better plan (reading the ticket). "Failed to select_console 'vnc' - failure on authentification" would have lead the reviewers in the right direction.

Martchus:
> What makes you think there comes an error message after no security types?

> If number-of-security-types is zero, then for some reason the
>  connection failed (e.g., the server cannot support the desired
>  protocol version).  This is followed by a string describing the
>  reason (where a string is specified as a length followed by that many
>  ASCII characters):

(see https://tools.ietf.org/html/rfc6143#page-9)

> I wouldn't try to read anything from a broken VNC server

If the read fails one gets the generic error message - so I see no problem in at least attempting to read.

Martchus:
> trying to make this a test failure sounds like the better plan (reading the ticket)

But we can not assume in general that this type of error is always caused by the SUT, can we? I mean it could also be simply a network issue causing the VNC connection to die unexpectedly.

coolo:
Sure, but still a test failure is preferred for broken things. If we can recover, we should.

Martchus:
> If we can recover, we should.

I don't think we can not recover from broken VNC except trying to connect again a few times.

> test failure is preferred for broken things

Ok, but then we should be consistent and not only do this for the few lines changed/added by this PR.

---

Seems like the error handling is already a bit inconsistent because in some place we use `OpenQA::Exception::VNCSetupError->throw(error => $error_message);`. I was wondering what is supposed to do (assumptions were either incomplete or test failure). So I put a similar line into the code to be executed unconditionally. Apparently the error is just ignored and the test gets stuck in `assert_screen`. That is certainly not what we want.

okurz:
@Martchus I guess by now you are also more confident to snatch svirt workers. Do you plan to continue?

Martchus:
The effort for testing this change is likely not worth the benefit. One needed to setup a fake VNC server which emits that error. Besides, I wouldn't be sure that my fake VNC server behaves like a real one did.

PR #1138, opened 2019-03-29
testapi: Make get_var_array return values consistent

badboywj:
If "$var" value exists, return a array referrence "@var", but return a string if "$var" doesn't exist. This is ugly. I just make the returned value consistent, both array referrence. And this also conforms to the document of "get_var_array".

NOTE: sorry for re-creating a PR because I make a mistake to delete my os-autoinst repository. Sorry, I will be carefull.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -738,8 +738,9 @@ Return the given variable as array reference (split variable value by , | or ; )
 
 sub get_var_array {
     my ($var, $default) = @_;
-    my @vars = split(/,|;/, $bmwqemu::vars{$var} || '');
-    return $default if !@vars;
+    my @vars    = split(/,|;/, $bmwqemu::vars{$var} || '');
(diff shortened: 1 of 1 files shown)


badboywj:
> The English in the commit message is not the best but I'd say it is mergeable.
> 
> I've been checking the openSUSE test distribution and it seems that the default is not used so far. So breaking the existing behavior shouldn't matter.

My English is NOT good : \, Thank you for checking this PR.

I also checked the openSUSE test distribution, and I found that everyone made
"get_var_array" to return  a array referrence, so everything is OK, but once
"get_var_array" return "string", some scripts in the project would fail. Actually
I encountered such a issue("get_var_array" return "string" type) when I plan to
improve some function of the openSUSE test distribution.

In the end, thank you to approve this changes.

PR #1139, opened 2019-03-29
Fix error message in consoles::network_console

Martchus:
Turns e.g.:
```
Can't locate object method "blessed" via package "actual error message
" (perhaps you forgot to load "actual error message
"?) at /hdd/openqa-devel/repos/os-autoinst/consoles/network_console.pm line 34.
```

into

```
actual error message
```

See https://progress.opensuse.org/issues/49793

---

Tested by playing around with the perl debugger.

### DIFF
--- a/consoles/network_console.pm
+++ b/consoles/network_console.pm
@@ -22,6 +22,7 @@ use warnings;
 use base 'consoles::console';
 
 use Try::Tiny;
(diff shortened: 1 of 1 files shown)


Soulofdestiny:
what's actually missing here?
affects all zVM jobs in 12SP5

PR #1140, opened 2019-04-08
Declare wait_screen_change earlier

foursixnine:
When ::wait_screen_change it is translated internally to
&main::wait_screen_change which means that the symbol will be looked
into the current program instead of the proper library where the method
is defined

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -86,7 +86,7 @@ sub send_key;
 sub check_screen;
 sub type_string;
 sub type_password;
-
+sub wait_screen_change;
 
 =head1 introduction
 
@@ -601,7 +601,7 @@ Example:
 =cut
 
 sub assert_screen_change(&@) {
-    ::wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
+    wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
 }
 
 


rwx788:
https://github.com/os-autoinst/os-autoinst/pull/689

rwx788:
Hmm, I wonder if your change could actually break all the tests (doubt that). Could you please restart them?

foursixnine:
@rwx788 Actually it's my change that breaks all tests :) I guess a bit of reading is in order... 
Quickly updating the prototype of the function results in:

```
Press ENTER or type command to continue
03-testapi.t .. Type of arg 1 to testapi::wait_screen_change must be block or sub {} (not array dereference) at ../testapi.pm line 604, near "@_) "
Bareword "is_serial_terminal" not allowed while "strict subs" in use at ../testapi.pm line 827.
Type of arg 1 to testapi::wait_screen_change must be block or sub {} (not reference constructor) at ../testapi.pm line 1214, near "};"
Bareword "is_serial_terminal" not allowed while "strict subs" in use at ../testapi.pm line 1265.
Type of arg 1 to testapi::wait_screen_change must be block or sub {} (not reference constructor) at ../testapi.pm line 1282, near "};"
Compilation failed in require at 03-testapi.t line 98.
BEGIN failed--compilation aborted at 03-testapi.t line 98.
03-testapi.t .. Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 

```

rwx788:
So it's even more broken =)))

foursixnine:
Closing in favor of #1142

PR #1142, opened 2019-04-09
Fix assert_screen_change method call

rwx788:
See [poo#50093](https://progress.opensuse.org/issues/50093).

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -601,7 +601,11 @@ Example:
 =cut
 
 sub assert_screen_change(&@) {
-    ::wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
+    # Need to parse code reference and pass to the method explicitly as
+    # wait_screen_change uses prototype which expects code block as an argument
+    # This resolves compile time issues
+    my ($coderef, @args) = @_;
+    wait_screen_change(\&{$coderef}, @_) or die 'assert_screen_change failed to detect a screen change';
 }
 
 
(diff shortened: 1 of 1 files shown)


rwx788:
@foursixnine, could you please take a look?

testapi.pm
```diff
@@ -601,7 +601,7 @@ Example:
 =cut
 
 sub assert_screen_change(&@) {
-    ::wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
+    wait_screen_change(\&{shift()}, @_) or die 'assert_screen_change failed to detect a screen change';
```
foursixnine: I guess you will still need to declare the method, or declare assert_screen_change after wait_screen_change so that the tests don't fail (but let's see)

rwx788: `wait_screen_change` is declared above `assert_screen_change`, so should be fine.

rwx788: Done. Was to motivated to keep it as single line =)))

foursixnine (approved):

testapi.pm
```diff
@@ -601,7 +601,11 @@ Example:
 =cut
 
 sub assert_screen_change(&@) {
-    ::wait_screen_change(@_) or die 'assert_screen_change failed to detect a screen change';
+    # Need to parse code reference and pass to the method explicitly as
+    # wait_screen_change uses prototype which expects code block as an argument
+    # This resolves compile time issues
+    my ($coderef, @args) = @_;
+    wait_screen_change(\&{$coderef}, @_) or die 'assert_screen_change failed to detect a screen change';
```
perlpunk, line 608: This is basically now:
```
wait_screen_change(\&{$coderef}, $coderef, @args) or die 'assert_screen_change failed to detect a screen change';
```
Which results in `$coderef` being read as a timeout, which results in the memory address, so usually a high number.

It should have been

```suggestion
    wait_screen_change(\&{$coderef}, @args) or die 'assert_screen_change failed to detect a screen change';
```

PR #1143, opened 2019-04-15
script_output: log script for execution into log

asmorodskyi:
script_output: log script for execution into log

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -212,6 +212,7 @@ sub script_output {
         my $heretag = 'EOT_' . $marker;
         my $cat     = "cat > $script_path << '$heretag'; echo $marker-\$?-";
         testapi::wait_serial($self->{serial_term_prompt}, undef, 0, no_regex => 1);
+        bmwqemu::log_call("Content of $script_path :\n \"$cat\" \n");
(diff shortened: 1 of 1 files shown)


asmorodskyi:
Do I need verification run for this ?

distribution.pm
```diff
@@ -212,6 +212,7 @@ sub script_output {
         my $heretag = 'EOT_' . $marker;
         my $cat     = "cat > $script_path << '$heretag'; echo $marker-\$?-";
         testapi::wait_serial($self->{serial_term_prompt}, undef, 0, no_regex => 1);
+        bmwqemu::log_call("Content of $script_path : $cat");
```
foursixnine: I'd vote for some markers at least, or enclose the $cat inside double quotes or double brackets or something in that fashion?

foursixnine:
+1

distribution.pm
asmorodskyi:
```suggestion
        bmwqemu::log_call("Content of $script_path : \n \"$cat\" \n ");
```

something like this ?

foursixnine: Somehow, yes :)

PR #1144, opened 2019-04-23
Enable assert_and_click to use an area's click point 

Martchus:
* See https://progress.opensuse.org/issues/49943
* See https://github.com/os-autoinst/openQA/pull/2054

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -349,9 +349,53 @@ is(match_has_tag,        undef, 'match_has_tag on no value -> undef');
 is(match_has_tag('foo'), undef, 'match_has_tag on not matched tag -> undef');
 subtest 'assert_and_click' => sub {
     my $mock_testapi = Test::MockModule->new('testapi');
-    $mock_testapi->mock(assert_screen => {area => [{x => 1, y => 2, w => 3, h => 4}]});
+    my @areas        = ({x => 1, y => 2, w => 10, h => 20});
+    $mock_testapi->mock(assert_screen => {area => \@areas});
+
+    $cmds = [];
     ok(assert_and_click('foo'));
-    is_deeply($cmds->[-1], {cmd => 'backend_mouse_hide', offset => 0}, 'assert_and_click succeeds and hides mouse again -> undef return');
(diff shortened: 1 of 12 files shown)


Martchus:
It is actually still WIP because I need to test it on a real test run.

t/03-testapi.t
```diff
@@ -349,9 +349,53 @@ is(match_has_tag,        undef, 'match_has_tag on no value -> undef');
...
+
+    $cmds  = [];
+    @areas = ({x => 50, y => 60, w => 22, h => 20, click_point => 'center'}, {x => 0, y => 0, w => 0, h => 0});
     ok(assert_and_click('foo'));
-    is_deeply($cmds->[-1], {cmd => 'backend_mouse_hide', offset => 0}, 'assert_and_click succeeds and hides mouse again -> undef return');
+    is_deeply($cmds->[1], {
+            cmd => 'backend_mouse_set',
+            x   => 61,
+            y   => 70,
+    }, 'assert_and_click clicks at the  click point specified as "center"') or diag explain $cmds;
```
kalikiana: There should be one more case with two click points to demonstrate what happens in that case.

Martchus: Than always the last click point is picked. Since it might not be predictable which one will be the last in practice I didn't care much about it, though. I would even forbid this case within the needle editor.

Martchus: In fact, having only one - well specified - click point is where the idea for this feature comes from.

kalikiana: Perhaps it should actually fail then, and that's what the test needs to verify.

Martchus: I added that validation. However, I moved the validation to the place when the needle is actually parsed to catch the error earlier.

Martchus:
I now also did a local test run and it worked. Since it didn't work from the start I also know that my test run actually relied on the correct behavior.

Martchus:
I needed to add a small fix in 'Take click point from area over to match' so the click point is actually serialized to matches and available in the needle editor when creating a new needle based on that match. (Tested in a local test-run.)

PR #1145, opened 2019-04-29
Extend type_password arguments

czerw:
Fix#50045: Function `type_password` use `type_string` for text input, but
not all arguments from type_string are supported. Arguments of `type_password`
function were extended to cover all from `type_string`.

Original reason was https://progress.opensuse.org/issues/49655. Function `type_password`
don't support `wait_still_screen`, although it is supported in `type_string`. 
* Old use:
```
type_password "password123";
wait_still_screen(1);
```
* New use:
```
type_password "password123", wait_still_screen => 1;
```

Progress: https://progress.opensuse.org/issues/50045
Verification: http://10.100.12.105/tests/1928

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1313,18 +1313,21 @@ sub type_string {
 
 =head2 type_password
 
-  type_password([$password]);
+  type_password($password [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, timeout => <num>]
+  [, similarity_level => <num>] );
 
 A convenience wrapper around C<type_string>, which doesn't log the string.
 
 Uses C<$testapi::password> if no string is given.
+
+You can pass same optional parameters as for C<type_string> function.
 
 =cut
 
 sub type_password {
     my ($string, %args) = @_;
     $string //= $password;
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -1324,7 +1324,13 @@ Uses C<$testapi::password> if no string is given.
 sub type_password {
     my ($string, %args) = @_;
     $string //= $password;
-    type_string $string, secret => 1, max_interval => ($args{max_interval} // 100);
+    type_string($string,
```
kalikiana: Can you make `type_password` an actual alias to `type_string` that passes all `%args` so the defaults etc. don't need to be re-defined here? The only difference would be that `secret => 1` is set.

czerw: I was thinking about it, but there is already `max_interval` with different default for `type_password` and I wanted to follow same logic, although with same defaults for other arguments.

kalikiana: Alright, that makes sense. I'd still suggest passing other arguments as-is since there's no requirement to have different defaults for everything.

czerw:
> Looks good but I guess tidy will complain.

Tidy was fine, had short talk with @kalikiana and I will change it to pass all variables, without naming them there.

PR #1146, opened 2019-04-30
Revert "Add an option of mousehide for assert_and_click"

nicksinger:
Reverts os-autoinst/os-autoinst#1075 since OBS refused to publish the package in the repository.
I could deploy the fetched RPMs manually but this only helps if OBS publishes until midnight (when o3 will update itself).


nicksinger:
We've disabled transactional updates on o3 (https://progress.opensuse.org/issues/50951) to avoid diverging os-autoinst versions. Therefore this can be closed and doesn't need to be reverted.

PR #1147, opened 2019-05-08
Add explanation to error regarding command server

Martchus:
So people don't make wrong deductions from this message as in https://progress.opensuse.org/issues/50225 and https://progress.opensuse.org/issues/49961.

The alternative would be removing the message completely but I find it helpful for debugging the developer mode.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -165,12 +165,20 @@ sub kill_commands {
         my $url       = "http://127.0.0.1:$cmd_srv_port/$job_token/broadcast";
         diag('isotovideo: informing websocket clients before stopping command server: ' . $url);
 
-        my $ua = Mojo::UserAgent->new(request_timeout => 15);
-        my $tx = $ua->post($url, json => {stopping_test_execution => $reason});
+        my $timeout = 15;
+        my $ua      = Mojo::UserAgent->new(request_timeout => $timeout);
+        my $tx      = $ua->post($url, json => {stopping_test_execution => $reason});
         try {
             my $res = $tx->result;
             if ($res->code != 200) {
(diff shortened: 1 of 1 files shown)


isotovideo
```diff
@@ -165,12 +165,18 @@ sub kill_commands {
...
+        my $ua      = Mojo::UserAgent->new(request_timeout => $timeout);
+        my $tx      = $ua->post($url, json => {stopping_test_execution => $reason});
         try {
             my $res = $tx->result;
             if ($res->code != 200) {
-                diag('isotovideo: unable to inform websocket clients about stopping command server: ' . $res->to_string);
+                my $error_message = $res->to_string;
+                diag('isotovideo: unable to inform websocket clients about stopping command server: ' . $error_message);
+                if ($error_message =~ qr/.*timeout.*/i) {
+                    diag("This is likely NOT related to the job possibly exceeding its timeout as it takes max. $timeout seconds. This indicates that the command server is under heavy load or has already terminated.");
```
foursixnine:
```suggestion
                    diag("isotovideo: The command server could be under heavy load and the timeout of $timeout has been exceeded. This message might not be related to the MAX_JOB_TIME of the current job being exceeded");
```

isotovideo
```diff
@@ -165,12 +165,18 @@ sub kill_commands {
...
+                }
+                diag('The command server might have already been killed by the worker after the user has aborted the job or the job timeout has been exceeded.');
```
foursixnine: This would be a bit confusing, don't you think?... I mean, if the message above is shown, and also this one... it would look a bit awkward

PR #1149, opened 2019-05-10
Allow failing for matching regex in autoinst log

asdil12:
https://progress.opensuse.org/issues/45011

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -28,8 +28,10 @@ use testapi  ();
 use autotest ();
 use MIME::Base64 'decode_base64';
 use Mojo::File 'path';
+use Try::Tiny;
 
-my $serial_file_pos = 0;
+my $serial_file_pos  = 0;
+my $autoinst_log_pos = 0;
 
 # enable strictures and warnings in all tests globaly
 sub import {
@@ -55,6 +57,7 @@ sub new {
     $self->{activated_consoles}     = [];
     $self->{name}                   = $class;
     $self->{serial_failures}        = [];
+    $self->{autoinst_failures}      = [];
     $self->{fatal_failure}          = 0;
     return bless $self, $class;
 }
@@ -366,13 +369,26 @@ sub runtest {
     }
 
     # Detect serial failures and override result if die
-    eval { $self->search_for_expected_serial_failures(); };
-    # Process serial dectection failure
-    if ($@) {
-        bmwqemu::diag($@);
-        $self->record_resultfile('Failed', $@, result => 'fail');
+    try {
+        $self->search_for_expected_serial_failures();
+    }
+    catch {
+        # Process serial dectection failure
+        bmwqemu::diag($_);
+        $self->record_resultfile('Failed', $_, result => 'fail');
         $died = 1;
     }
+
+    # Detect autoinst failures and override result if die
+    try {
+        $self->search_for_expected_autoinst_failures();
+    }
+    catch {
+        # Process autoinst dectection failure
+        bmwqemu::diag($_);
+        $self->record_resultfile('Failed', $_, result => 'fail');
+        $died = 1;
+    };
 
     $self->run_post_fail("test $name died") if ($died);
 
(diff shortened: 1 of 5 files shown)


basetest.pm
```diff
@@ -715,6 +726,47 @@ sub parse_serial_output_qemu {
...
+                    $die                   = 1;
+                    $fail_type             = 'fail';
+                    $self->{fatal_failure} = $type eq 'fatal';
+                }
+                $self->record_resultfile($message, $message . " - autoinst-log error: $line", result => $fail_type);
+                $self->{result} = $fail_type;
+            }
+        }
+    }
+    $autoinst_log_pos = tell $AUTOINSTLOG;
```
Martchus: If the previous open/seek or reads fails this will assign `-1` to that global variable. When calling `search_for_expected_autoinst_failures` again this would be used for the initial seek. This behavior is likely not the easiest to debug. Since it is not nice to suppress errors silently anyways I suggest to handle possible errors.

distribution.pm
```diff
@@ -289,6 +290,28 @@ sub set_expected_serial_failures {
     $self->{serial_failures} = $failures;
 }
 
+=head2 set_expected_autoinst_failures
+
+    set_expected_autoinst_failures($failures)
+
+Define the patterns to look for in the os-autoinst.log.
```
Martchus: s/os-autoinst.log/autoinst-log.txt/

basetest.pm
```diff
@@ -715,6 +726,47 @@ sub parse_serial_output_qemu {
...
+    seek($AUTOINSTLOG, $autoinst_log_pos, 0);
+    while (my $line = <$AUTOINSTLOG>) {
+        chomp $line;
+        for my $regexp_table (@{$failures}) {
+            my $regexp  = $regexp_table->{pattern};
+            my $message = $regexp_table->{message};
+            my $type    = $regexp_table->{type};
+
+            # Input parameters validation
+            die "Wrong type defined for serial failure. Only 'soft' or 'hard' allowed. Got: $type" if $type !~ /^soft|hard|fatal$/;
```
Martchus: I don't care much but wouldn't it be better to do the validation already in `set_expected_autoinst_failures`?

basetest.pm
```diff
@@ -374,6 +376,15 @@ sub runtest {
         $died = 1;
     }
 
+    # Detect autoinst failures and override result if die
+    eval { $self->search_for_expected_autoinst_failures(); };
```
Martchus: I'd prefer use of `Try::Tiny` here over eval. (Using a return code would be sufficient here as well.)

basetest.pm
asdil12: Leaving for now as discussed.

basetest.pm
asdil12: Checking error for seek and open.

Martchus: But not for the reading the lines. If a read fails the while loop ends and I think `tell` would also return `-1` then.

asdil12: fixed that.

basetest.pm
```diff
@@ -715,6 +731,51 @@ sub parse_serial_output_qemu {
...
+                    $die                   = 1;
+                    $fail_type             = 'fail';
+                    $self->{fatal_failure} = $type eq 'fatal';
+                }
+                $self->record_resultfile($message, $message . " - autoinst-log error: $line", result => $fail_type);
+                $self->{result} = $fail_type;
+            }
+        }
+    }
+    $fp_pos = tell $AUTOINSTLOG;
```
Martchus: Global symbol "$fp_pos" requires explicit package name at ../basetest.pm line 769.

asdil12: I hate perl - should be fixed now.

Martchus:
Not sure whether the failing unit tests are related. I guess you have to look into it.

basetest.pm
```diff
@@ -28,8 +28,10 @@ use testapi  ();
 use autotest ();
 use MIME::Base64 'decode_base64';
 use Mojo::File 'path';
+use Try::Tiny;
```
foursixnine: You will need to update the cpanfile and likely the os-autoinst.spec file too

Martchus: We're already using this all over the place - also in os-autoinst. `Try::Tiny` is missing in the cpan and spec file so it appears it is already pulled by some dependency.

asdil12:
Fixed the testsuite.

SergioAtSUSE:
@foursixnine, can you take a look?

asdil12:
As discussed on friday my code will properly not get executed.
As the job result page doesn't show any test module status in this case
there would be major changes to os-autoinst needed to change that.

PR #1150, opened 2019-05-17
Don't modify backend vars when running test

asdil12:
Unless `t/04-check_vars_docu.t` is run with `--write`.

Ticket: https://progress.opensuse.org/issues/34303

### DIFF
--- a/t/04-check_vars_docu.t
+++ b/t/04-check_vars_docu.t
@@ -84,7 +84,7 @@ sub read_doc {
 
 sub write_doc {
     my $docfh;
-    open($docfh, '>', VARS_DOC);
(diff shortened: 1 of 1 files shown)


t/04-check_vars_docu.t
```diff
@@ -44,6 +44,8 @@ my $error_found = 0;
 # ignore errors for now
 my $ignore_errors = 1;
 
+my $do_write = (defined $ARGV[$#ARGV] && $ARGV[$#ARGV] eq '--write') ? 1 : 0;
```
coolo: tests aren't programs. Let it create a different filename, so it appears in git status - for those that want to check what's different

asdil12: done.

PR #1151, opened 2019-05-17
Add quiet option to wait_serial()

cfconrad:
When using utility functions, it's not needed to show status squares of wait_expect and so on. As these are not related to the actual test. Adding the `quiet` option do avoid this.

Also use common function parameters for equal functions like
script_run(), assert_script_run(), script_output() and validate_script_output().

Validation run testscript:
  * http://cfconrad-vm.qa.suse.de/tests/4367

If this goes in the right direction I will add tests and update the perlpod.

@asmorodskyi @foursixnine @Martchus WDYT?

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -106,38 +106,47 @@ sub become_root {
 
 =head2 script_run
 
-script_run($program, [$wait_seconds])
+  script_run($cmd [, timeout => $timeout] [,quiet => $quiet])
 
-Run I<$program> (by assuming the console prompt and typing the command). After that, echo
+Deprecated mode
+
+  script_run($program, [$timeout])
+
+Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
 hashed command to serial line and wait for it in order to detect execution is finished.
-To avoid waiting, use I<$wait_seconds> 0.
+To avoid waiting, use I<$timeout> 0.
 
-<Returns> exit code received from I<$program>, or 0 in case of C<not> waiting for I<$program>
+Use C<quiet> to avoid recording serial_results.
+
+<Returns> exit code received from I<$cmd>, or C<undef> in case of C<not> waiting for I<$cmd>
 to return.
 
 =cut
 
 sub script_run {
-    # start console application
-    my ($self, $cmd, $wait) = @_;
-    $wait //= $bmwqemu::default_timeout;
+    my ($self, $cmd) = splice(@_, 0, 2);
+    my %args = testapi::compat_args(
+        {
+            timeout => $bmwqemu::default_timeout,
+            quiet   => undef
+        }, ['timeout'], @_);
 
     if (testapi::is_serial_terminal) {
-        testapi::wait_serial($self->{serial_term_prompt}, undef, 0, no_regex => 1);
+        testapi::wait_serial($self->{serial_term_prompt}, no_regex => 1, quiet => $args{quiet});
     }
     testapi::type_string "$cmd";
-    if ($wait > 0) {
-        my $str = testapi::hashed_string("SR$cmd$wait");
+    if ($args{timeout} > 0) {
+        my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
         if (testapi::is_serial_terminal) {
             my $marker = " ; echo $str-\$?-";
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, undef, 0, no_regex => 1);
+            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
             testapi::type_string("\n");
         }
         else {
             testapi::type_string " ; echo $str-\$?- > /dev/$testapi::serialdev\n";
         }
-        my $res = testapi::wait_serial(qr/$str-\d+-/, $wait);
+        my $res = testapi::wait_serial(qr/$str-\d+-/, timeout => $args{timeout}, quiet => $args{quiet});
         return unless $res;
         return ($res =~ /$str-(\d+)-/)[0];
     }
@@ -180,6 +189,10 @@ sub script_sudo {
 
 =head2 script_output
 
+    script_output($script [, timeout => ?] [, type_command => ?] [, proceed_on_failure => ?] [, quiet => ?])
+
+Deprecated mode
+
     script_output($script, [ $wait, type_command => ?, proceed_on_failure => ?])
 
 Execute $script on the SUT and return the data written to STDOUT by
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -835,21 +835,28 @@ Setting C<$no_regex> will cause it to do a plain string search.
...
+            timeout          => 90,
+            expect_not_found => 0,
+            quiet            => 0,
+            no_regex         => 0,
+            buffer_size      => undef,
+            record_output    => undef,
+            exclude_match    => undef
+        }, ['timeout', 'expect_not_found'], @_);
+    # hacky, but previous version converted 0 timeout to default
+    $args{timeout} ||= 90;
```
cfconrad: I'm not sure, if we wanna keep that behavior. A quick `grep` didn't found such call with `timeout == 0`.

distribution.pm
```diff
@@ -118,26 +118,29 @@ to return.
 =cut
 
 sub script_run {
-    # start console application
-    my ($self, $cmd, $wait) = @_;
-    $wait //= $bmwqemu::default_timeout;
+    my ($self, $cmd) = splice(@_, 0, 2);
+    my %args = OpenQA::Isotovideo::Utils::compat_args(
+        {
+            'timeout' => $bmwqemu::default_timeout,
```
Martchus: Perlcritic likely don't like the "unnecessary" quotes around the keys.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -64,4 +65,27 @@ sub checkout_git_repo_and_branch {
...
+Helper function to create backward compatible function arguments when moving
+from positional arguments to named one.
+
+    compat_args( $hash_ref_defaults, $arrayref_old_fixed, [ $arg1, $arg2, ...])
+
+A typical call would look like:
+
+    my %args = compat_args({timeout => 60, .. }, ['timeout'], @_);
+=cut
+sub compat_args {
```
Martchus: I like that you created a separate function for this recurring and not that simple code. Maybe it is worth creating a dedicated unit test for it.

testapi.pm
Martchus: Then I'm inclined not to keep it. But test writers should decide that.

foursixnine: Isn't the default or whatever was assigned going to be assigned here? basically making this line not needed anymore

testapi.pm
```diff
@@ -835,21 +835,28 @@ Setting C<$no_regex> will cause it to do a plain string search.
 =cut
 
 sub wait_serial {
+    my $regexp = shift;
+    my %args   = OpenQA::Isotovideo::Utils::compat_args(
+        {
+            regexp           => $regexp,
+            timeout          => 90,
```
foursixnine: ~~Why not `$bmwqemu::default_timeout` again?~~ I wonder if $bmwqemu::default_timeout should be increased and used here instead (But that's ouside the scope of this pr)

OpenQA/Isotovideo/Utils.pm
```diff
@@ -64,4 +65,27 @@ sub checkout_git_repo_and_branch {
     $bmwqemu::vars{$dir} = File::Spec->rel2abs($local_path);
 }
 
+=head2
```
foursixnine: I sort of wish there was a better namespace to put this routine... But I don't wanna say bmwqemu...

distribution.pm
cfconrad: ups left over... thx

cfconrad:
I needed to move `compat_args()` to `testapi.pm` as I had cycling include problems with `Utils.pm`.

testapi.pm
cfconrad: The important thing with this line is that `undef` is mapped to default value. But the old implementation also mapped `0` to default value.

OpenQA/Isotovideo/Utils.pm
cfconrad: I needed to put it in testapi. See comment below:
>  I needed to move compat_args() to testapi.pm as I had cycling include problems with Utils.pm.

testapi.pm
cfconrad: From my point the wired thing with line is that timeout equal to `0` was mapped to default/90 as well and not only `undef`.

OpenQA/Isotovideo/Utils.pm
cfconrad: I'm on it now.

testapi.pm
```diff
@@ -1113,7 +1126,7 @@ sub validate_script_output($&;$) {
         bmwqemu::diag("output does not pass the code block:\n$output");
     }
     # abusing the function
-    $autotest::current_test->record_serialresult($output, $res, $output);
+    $autotest::current_test->record_serialresult($output, $res, $output) unless ($args{quiet});
```
foursixnine: Looking at the code, and the test run, this is information that might be needed from a test review perspective. Perhaps the quiet parameter should only keep calls beyond this point, but not the final result (I guess this would apply to the *_script functions from the testapi but not to those on the backend...

cfconrad:
actually this is the idea of `quiet`. Imagine you have a 100 cnt loop doing 3 script_run() in it. I would like to avoid spamming the gui with such things, when a single validation afterwards would do it as well.

foursixnine:
> actually this is the idea of `quiet`. Imagine you have a 100 cnt loop doing 3 script_run() in it. I would like to avoid spamming the gui with such things, when a single validation afterwards would do it as well.

Hands down, I like the idea :P, but there might be something lost... but I guess it's a tradeoff...

cfconrad:
What I have in my pocket is something like this http://cfconrad-vm.qa.suse.de/tests/4280, where I'm using serial terminal to  do  wired file transfer (beside if this is a good idea or not :). And this makes a lot of noise, so it could be much nicer in the end.

OpenQA/Isotovideo/Utils.pm
cfconrad: done for compat_args(). Need to do it for *_run() calls

testapi.pm
cfconrad: Ok I removed that line now. So we have a none backward compatible change. But I didn't found such call in os-autoinst-distri-opensuse.

jlausuch:
LGTM. This is needed for [poo#51098](https://progress.opensuse.org/issues/51098) to allow upload logs in network tests which might not have network configured (can't use curl/scp)

cfconrad:
@foursixnine I see, so you would just pass `quiet=>1` to `script_output()`. I like that idea, but lets do that in a different step.

PR #1152, opened 2019-05-21
Ignore files generated by the testsuite

Martchus:
Complements https://github.com/os-autoinst/os-autoinst/pull/1150 and also covers 2 other files I have usually laying around after running tests locally.


.gitignore
```diff
@@ -42,3 +42,8 @@ t/*.trs
 t/pool/*
 t/cover_db/*
 .*.swp
+
+# files generated by running the testsuite
+doc/backend_vars.asciidoc.newvars
```
coolo: well, the idea was that the file appears in git status if it changed. Perhaps the test failed to implement the 'if changed' part :)

Martchus: Yes - because so far the tests just create yet another file lying around in my local checkout which I need to take care not to check in.

Martchus: @asdil12 What was your intention which the PR creating that file?

asdil12: The test itself should not modify the existing files.
I would rather list that new file in gitignore as in 95% of the cases you properly don't care about this file and it will just be annoying.

Martchus: I just had a brief look at the test and (at least in its current state) it returns with a non-zero exit code in case of a failure. That should be sufficient. The file `doc/backend_vars.asciidoc.newvars` is always created.

So I'd say we're good to ignore the file. I'm only wondering why the test fails for me locally.

okurz: I reopened https://progress.opensuse.org/issues/34303 because we are *not* good to ignore the file. The idea was to have tests fail when someone introduces an undocumented variable.

okurz:
See discussion in https://github.com/os-autoinst/os-autoinst/pull/1152/files#r285966387

PR #1153, opened 2019-05-21
Modify workaround property to support adding reason

Amrysliu:
See https://progress.opensuse.org/issues/33706
Related: https://github.com/os-autoinst/openQA/pull/2070

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -163,10 +163,15 @@ sub record_screenmatch {
 
     # When the needle has the workaround property,
     # mark the result as dent and increase the dents
-    if ($foundneedle->has_property('workaround')) {
-        $result->{dent} = 1;
-        $self->{dents}++;
-        bmwqemu::diag("needle '$serialized_match->{name}' is a workaround");
+    if (my $reason = $foundneedle->get_property_value('workaround')) {
+        $result->{dent}   = 1;
+        $result->{result} = "softfail";
+
+        # write a test result file
+        my $reason = $foundneedle->get_property_value('workaround');
+        $self->record_soft_failure_result($reason);
+
+        bmwqemu::diag("needle '$serialized_match->{name}' is a workaround. The reason is $reason");
     }
 
     # also include the not matched needles
@@ -464,6 +469,7 @@ sub record_soft_failure_result {
     $result->{title} = 'Soft Failed';
     $result->{text}  = $filename;
     $self->write_resultfile($filename, "# Soft Failure:\n$reason\n");
+    $self->{dents}++;
     return undef;
 }
 
--- a/needle.pm
+++ b/needle.pm
@@ -148,7 +148,7 @@ sub save {
         {
             tags       => [sort(@{$self->{tags}})],
             area       => \@area,
-            properties => [sort(@{$self->{properties}})],
+            properties => [$self->{properties}],
         });
     open(my $fh, '>', $fn);
     print $fh $json;
@@ -276,11 +276,25 @@ sub has_tag {
 }
 
 sub has_property {
-    my ($self, $tag) = @_;
-    for my $t (@{$self->{properties}}) {
-        return 1 if ($t eq $tag);
+    my ($self, $property_name) = @_;
+
+    return grep { ref($_) eq "HASH" ? $_->{name} eq $property_name : $_ eq $property_name } @{$self->{properties}};
+}
+
+sub get_property_value {
+    my ($self, $property_name) = @_;
+
+    for my $property (@{$self->{properties}}) {
+        if (ref($property) eq "HASH") {
+            return $property->{value} if ($property->{name} eq $property_name);
+        }
     }
-    return 0;
+    if ($property_name eq "workaround") {
+        if ($self->{name} =~ /\S+\-(bsc|poo|bnc)(\d+)\-\S+/) {
+            return $1 . "#" . $2;
+        }
+    }
+    return undef;
 }
 
 sub TO_JSON {
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -413,6 +413,31 @@ subtest 'click point' => sub {
     is_deeply($needle, undef, 'multiple click points not accepted');
 };
 
+subtest 'workaround property' => sub {
+    my $workaround_string_needle     = needle->new($misc_needle_dir . 'check-workaround-bsc1234567-20190522.json');
+    my $workaround_hash_needle       = needle->new($misc_needle_dir . 'check-workaround-hash-20190522.json');
+    my $no_workaround_needle         = needle->new($misc_needle_dir . 'click-point-center.json');
+    my $mix_workaround_string_needle = needle->new($misc_needle_dir . 'check-workaround-mix-bsc987321-20190617.json');
+    my $mix_workaround_hash_needle   = needle->new($misc_needle_dir . 'check-workaround-hash-mix-20190617.json');
+
+    ok($workaround_string_needle->has_property("workaround"),     "workaround property found when it is recorded in string");
+    ok($workaround_hash_needle->has_property("workaround"),       "workaround property found when it is recorded in hash");
+    ok($mix_workaround_string_needle->has_property("workaround"), "workaround property found in mixed properties");
+    ok($mix_workaround_hash_needle->has_property("workaround"),   "workaround property found in mixed properties");
+    ok($no_workaround_needle->has_property("glossy"),             "glossy property found");
+    ok(!$no_workaround_needle->has_property("workaround"),        "workaround property not found");
+    ok(!$workaround_string_needle->has_property("glossy"),        "glossy property not found");
+    ok(!$workaround_hash_needle->has_property("glossy"),          "glossy property not found");
+
+    is($workaround_string_needle->get_property_value("workaround"), "bsc#1234567", "get correct value when workaround is recorded in string");
+    is($workaround_hash_needle->get_property_value("workaround"), "bsc#7654321: this is a test about workaround.", "get ccorrect value when workaround is recorded in hash");
+    is($mix_workaround_string_needle->get_property_value("workaround"), "bsc#987321",                                         "workaround value is correct");
+    is($mix_workaround_hash_needle->get_property_value("workaround"),   "bsc#123789: This is a test for workaround property", "workaround value is correct");
+    is($workaround_hash_needle->get_property_value("test"),             undef,                                                "no test value");
+    is($no_workaround_needle->get_property_value("workaround"),         undef,                                                "no workaround property");
+    is($no_workaround_needle->get_property_value("glossy"), undef, "glossy property is a string, has no value");
+};
+
 done_testing();
 
 # vim: set sw=4 et:
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -5,6 +5,7 @@ use warnings;
 use Test::MockModule;
 use Test::More;
 use Test::Fatal;
+use File::Basename;
 
 BEGIN {
     unshift @INC, '..';
@@ -220,6 +221,63 @@ subtest record_screenmatch => sub {
             }
     ], 'screenmatch detail recorded as expected')
       or diag explain $basetest->{details};
+
+    # check a needle has workaround property
+    my $basetest_for_workaround = basetest->new();
+    my $misc_needle_dir         = dirname(__FILE__) . '/misc_needles/';
+    my $needle_file             = $misc_needle_dir . "check-workaround-hash-20190522.json";
+    my %workaround_match        = (
+        area => [
+            {x => 1, y => 2, w => 3, h => 4, result => 'ok'},
+        ],
+        error  => 0.25,
+        needle => {
+            name       => 'check-workaround-hash-20190522',
+            file       => $needle_file,
+            properties => [
+                {
+                    name  => 'workaround',
+                    value => 'bsc#7654321: this is a test about workaround.',
+                }
+            ],
+        },
+    );
+
+    $basetest_for_workaround->record_screenmatch($image, \%workaround_match, ['check-workaround-hash'], [], $frame);
+    is_deeply($basetest_for_workaround->{details}, [
+            {
+                result => 'softfail',
+                title  => 'Soft Failed',
+                text   => 'basetest-3.txt'
+            },
+            {
+                area => [
+                    {
+                        x          => 1,
+                        y          => 2,
+                        w          => 3,
+                        h          => 4,
+                        similarity => 0,
+                        result     => 'ok',
+                    },
+                ],
+                error      => 0.25,
+                frametime  => [qw(1.00 1.04)],
+                needle     => 'check-workaround-hash-20190522',
+                json       => $needle_file,
+                properties => [
+                    {
+                        name  => 'workaround',
+                        value => 'bsc#7654321: this is a test about workaround.',
+                    }
+                ],
+                screenshot => 'basetest-1.png',
+                tags       => [qw(check-workaround-hash)],
+                result     => 'softfail',
+                dent       => 1,
+            }
+    ], 'screenmatch detail with workaround property recorded as expected')
+      or diag explain $basetest_for_workaround->{details};
 };
 
 subtest 'register_extra_test_results' => sub {
--- a/testapi.pm
+++ b/testapi.pm
@@ -181,7 +181,6 @@ sub record_soft_failure {
     bmwqemu::log_call(reason => $reason);
 
     $autotest::current_test->record_soft_failure_result($reason);
-    $autotest::current_test->{dents}++;
 }
 
 sub _is_valid_result {
@@ -234,7 +233,6 @@ sub force_soft_failure {
     bmwqemu::log_call(reason => $reason);
 
     $autotest::current_test->record_soft_failure_result($reason, force_status => 1);
-    $autotest::current_test->{dents}++;
 }
 
 sub _handle_found_needle {
--- a/t/misc_needles/check-workaround-bsc1234567-20190522.json
+++ b/t/misc_needles/check-workaround-bsc1234567-20190522.json
@@ -1,0 +1,14 @@
+{
+    "area": [
+        {
+            "height": 100,
+            "match": 80,
+            "type": "match",
+            "width": 200,
+            "xpos": 10,
+            "ypos": 10
+        }
+    ],
+    "tags": ["check-workaround"],
+    "properties": ["workaround"]
+}
--- a/t/misc_needles/check-workaround-bsc1234567-20190522.png
+++ b/t/misc_needles/check-workaround-bsc1234567-20190522.png
@@ -1,0 +1,1 @@
+check-workaround-bsc1234567-20190522.json
--- a/t/misc_needles/check-workaround-hash-20190522.json
+++ b/t/misc_needles/check-workaround-hash-20190522.json
@@ -1,0 +1,19 @@
+{
+    "area": [
+        {
+            "height": 100,
+            "match": 80,
+            "type": "match",
+            "width": 200,
+            "xpos": 10,
+            "ypos": 10
+        }
+    ],
+    "tags": ["check-workaround-hash"],
+    "properties": [
+        {
+            "name": "workaround",
+            "value": "bsc#7654321: this is a test about workaround."
+        }
+    ]
+}
--- a/t/misc_needles/check-workaround-hash-20190522.png
+++ b/t/misc_needles/check-workaround-hash-20190522.png
@@ -1,0 +1,1 @@
+check-workaround-hash-20190522.json
--- a/t/misc_needles/check-workaround-hash-mix-20190617.json
+++ b/t/misc_needles/check-workaround-hash-mix-20190617.json
@@ -1,0 +1,30 @@
+{
+    "area": [
+        {
+            "height": 100,
+            "match": 80,
+            "type": "match",
+            "width": 200,
+            "xpos": 10,
+            "ypos": 10
+        }
+    ],
+    "tags": ["check-workaround"],
+    "properties": [
+        "glossy",
+        {
+            "name": "test",
+            "value": "this is a test"
+        },
+        {
+            "name": "workaround",
+            "value": "bsc#123789: This is a test for workaround property"
+        },
+        "test2",
+        "test3",
+        {
+            "name": "test4",
+            "value": "this is a test for mix properties"
+        }
+    ]
+}
--- a/t/misc_needles/check-workaround-hash-mix-20190617.png
+++ b/t/misc_needles/check-workaround-hash-mix-20190617.png
@@ -1,0 +1,1 @@
+check-workaround-hash-mix-20190617.json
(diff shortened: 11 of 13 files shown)


needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
+        else {
+            return 1 if ($t eq $pro);
+        }
     }
     return 0;
 }
 
+sub get_property_value {
+    my ($self, $pro) = @_;
+    for my $t (@{$self->{properties}}) {
```
Martchus: I'd prefer a better variable name than `$t`.

needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
+        if (ref($t) eq "HASH") {
+            return $t->{value} if ($t->{name} eq $pro);
+        }
+        else {
+            return "" if ($pro ne "workaround");
```
Martchus: Wouldn't this function fail to find a property value if a non-hash comes before? So shouldn't it be `next` instead of `return`?

needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
 }
 
 sub has_property {
-    my ($self, $tag) = @_;
+    my ($self, $pro) = @_;
```
Martchus: `s/$pro/$property_name/`

(same for the other function)

needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
 
+sub get_property_value {
+    my ($self, $pro) = @_;
+    for my $t (@{$self->{properties}}) {
+        if (ref($t) eq "HASH") {
+            return $t->{value} if ($t->{name} eq $pro);
+        }
+        else {
+            return "" if ($pro ne "workaround");
+            if ($self->{name} =~ /\S+\-(bsc|poo)(\d+)\-\S+/) {
```
Martchus: Having this fallback inside the loop body (although it does not use `$t`) looks wrong, too. Shouldn't this only be executed if there is no matching HASH?

Amrysliu:
> Maybe I completely misread the `get_property_value` function or it is wrong. In any case, you could try to add unit tests.

I trim the code logic, and modify them according to your comments. Also add some test code.

t/01-test_needle.t
```diff
@@ -413,6 +413,22 @@ subtest 'click point' => sub {
...
+
+    is($workaround_string_needle->has_property("workaround"), 1, "workaround property found when it is recorded in string");
+    is($workaround_hash_needle->has_property("workaround"),   1, "workaround property found when it is recorded in hash");
+    is($no_workaround_needle->has_property("workaround"),     0, "workaround property not found");
+
+    is($workaround_string_needle->get_property_value("workaround"), "bsc#1234567", "get correct value when workaround is recorded in string");
+    is($workaround_hash_needle->get_property_value("workaround"), "bsc#7654321: this is a test about workaround.", "get ccorrect value when workaround is recorded in hash");
+    is($workaround_hash_needle->get_property_value("test"),     "", "no test value");
+    is($no_workaround_needle->get_property_value("workaround"), "", "no workaround property");
+    is($no_workaround_needle->get_property_value("glossy"),     "", "glossy property is a string, has no value");
```
Martchus: You could also add a test for `has_property` and `glossy`.

t/misc_needles/check-workround-hash-bsc7654321-20190522.json
```diff
@@ -0,0 +1,19 @@
...
+            "height": 100,
+            "match": 80,
+            "type": "match",
+            "width": 200,
+            "xpos": 10,
+            "ypos": 10
+        }
+    ],
+    "tags": ["check-workaround-hash"],
+    "properties": [
```
Martchus: Seems like none of these JSON files contain multiple properties. Maybe you can add some other text-only and HASH-properties before and after the workaround property. It shouldn't prevent the workaround property to be found and it would make sense to test that.

needle.pm
```diff
@@ -148,7 +148,7 @@ sub save {
         {
             tags       => [sort(@{$self->{tags}})],
             area       => \@area,
-            properties => [sort(@{$self->{properties}})],
+            properties => [$self->{properties}],
```
Martchus: @os-autoinst/tools-team I'm wondering whether we still want to have the properties sorted. A custom sort function which handles the hashes shouldn't be hard to implement.

basetest.pm
```diff
@@ -164,7 +164,18 @@ sub record_screenmatch {
     if ($foundneedle->has_property('workaround')) {
         $result->{dent} = 1;
         $self->{dents}++;
-        bmwqemu::diag("needle '$serialized_match->{name}' is a workaround");
+
+        # get the workaround reason and write a test result file.
```
Martchus: Either start with a small letter xor end with a dot.

basetest.pm
```diff
@@ -164,7 +164,18 @@ sub record_screenmatch {
...
+        my $reason = $foundneedle->get_property_value('workaround');
+
+        $result->{title} = 'Soft Failed';
```
Martchus: This is copied and pasted from `record_soft_failure`. Why don't you simply make use of that function? Of course you need to remove `$result->{dent} = 1;` and `$self->{dents}++;` at the top of the `if`-block because `$self->record_soft_failure($reason)` would already take care of that.

needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
-    for my $t (@{$self->{properties}}) {
-        return 1 if ($t eq $tag);
+    my ($self, $property_name) = @_;
+    for my $property (@{$self->{properties}}) {
+        if (ref($property) eq "HASH") {
+            return 1 if ($property->{name} eq $property_name);
+        }
+        else {
+            return 1 if ($property eq $property_name);
+        }
```
okurz:
```perl
return 1 if ($property->{name} // $property) eq $property_name);
```

should work or even to replace the whole for-loop and return block as well:

```perl
return grep { ($_->{name} // $_) eq $property_name } @{$self->{properties}};

File: needle.pm [hunk 11]
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
 }
 
+sub get_property_value {
+    my ($self, $property_name) = @_;
+
+    for my $property (@{$self->{properties}}) {
+        if (ref($property) eq "HASH") {
+            return $property->{value} if ($property->{name} eq $property_name);
+        }
+    }
```
okurz: see above, same as there :)

needle.pm
```diff
@@ -276,13 +276,34 @@ sub has_tag {
...
+    if ($property_name eq "workaround") {
+        if ($self->{name} =~ /\S+\-(bsc|poo)(\d+)\-\S+/) {
```
okurz: hm, this will not be long-term scalable but IIUC this is merely backward compatibility support, right?

t/01-test_needle.t
```diff
@@ -413,6 +413,22 @@ subtest 'click point' => sub {
     is_deeply($needle, undef, 'multiple click points not accepted');
 };
 
+subtest 'workround property' => sub {
```
okurz: s/workround/workaround/, also in the following lines

t/01-test_needle.t
```diff
@@ -413,6 +413,22 @@ subtest 'click point' => sub {
...
+    my $workaround_string_needle = needle->new($misc_needle_dir . 'check-workround-bsc1234567-20190522.json');
+    my $workaround_hash_needle   = needle->new($misc_needle_dir . 'check-workround-hash-bsc7654321-20190522.json');
+    my $no_workaround_needle     = needle->new($misc_needle_dir . 'click-point-center.json');
+
+    is($workaround_string_needle->has_property("workaround"), 1, "workaround property found when it is recorded in string");
```
okurz: If you just want to test for a "true" value I suggest to use `ok` instead of `is`

t/17-basetest.t
```diff
@@ -220,6 +221,60 @@ subtest record_screenmatch => sub {
             }
     ], 'screenmatch detail recorded as expected')
       or diag explain $basetest->{details};
+
+    # check a needle has workaround property
+    my $basetest_for_workaround = basetest->new();
+    my $misc_needle_dir         = dirname(__FILE__) . '/misc_needles/';
+    my $needle_file             = $misc_needle_dir . "check-workround-hash-bsc7654321-20190522.json";
```
okurz: again typo s/workround/workaround/

t/17-basetest.t
```diff
@@ -220,6 +221,60 @@ subtest record_screenmatch => sub {
...
+            name       => 'check-workround-hash-bsc7654321-20190522',
+            file       => $needle_file,
+            properties => [
+                {
+                    name  => 'workaround',
+                    value => 'bsc#7654321: this is a test about workaround.',
+                }
+            ],
+        },
+    );
```
okurz: I suggest to just check what we care about here, i.e. not `is_deeply` but rather just check for the right keys and values being present

t/misc_needles/check-workround-bsc1234567-20190522.json
```diff
@@ -0,0 +1,14 @@
+{
```
okurz: typo in the filename

basetest.pm
Amrysliu: record_soft_failure is a function in testapi.pm, could I call this in basetest.pm? There is a function called "record_soft_failure_result", maybe you mean it? but it doesn't do $self->{dents}++ and $result->{dent}. So there is a little different.

needle.pm
Amrysliu: If I modify the code as your suggestion, there will be a warning message "Cannot use a string as HASH...". 
Current code maybe looks a bit stupid, but I think there is no warning message and simpler to read. Can I keep them?

basetest.pm
Martchus: I mean `record_soft_failure_result` and my previous comment already contains a note about `$self->{dents}++ and $result->{dent}`.

Amrysliu: I still have some questions about those code:
a. In you previous comment, you said I need to remove `result->{dent} = 1` and `$self->{dents}++;`, I do not understand clearly what 'dent' and 'dents' used for.   I cannot find out where `record_soft_failure_result` has handled them.
b. function `record_soft_failure_result` calls `record_testresult`, it does `++self->{test_count}`, as a result, the next result file is 1 more than expected (In my opinion). For example, In my testcase, when I call `record_soft_failure_result`, the result file name is 'basetest-3.txt'; when I do not call it, the result file name is 'basetest-2.txt'.
c.  Function `record_soft_failure_result` calls `record_testresult`, and `record_testresult` push a result hash to `$self->{details}`. In the end of function `record_screenmatch` also push a result hash to `$self->{details}`, it repeats.
I really did test for these. This is why I did not simply call `record_soft_failure_result` in my code. Maybe I made some mistakes in understanding the 'record_screenmatch' function. If I did, please let me know.

t/17-basetest.t
Amrysliu: Except for area and error, almost others need to be checked. so I prefer to check all hash.

t/01-test_needle.t
Amrysliu: Added

t/misc_needles/check-workround-hash-bsc7654321-20190522.json
Amrysliu: added

basetest.pm
Amrysliu: modified

basetest.pm
Martchus: > I cannot find out where record_soft_failure_result has handled them.

Those are actually handled in `testapi::record_softfailure` and `testapi::force_soft_failure` and *not* inside `basetest`. But maybe we want to change this inconsistency and increase the dents directly in `basetest::record_soft_failure_result`? That would make sense to me. Otherwise you actually need to keep the dents in place.

---

> I do not understand clearly what 'dent' and 'dents' used for.

To count the number of issues which have been occurred during the test run. Dents is actually (in my opinion) a poorly chosen word in that context. I think someone was a bit too eager using sophisticated vocabulary.

---

So I admit that using `basetest::record_soft_failure_result` is not that easily possible. Beside from that: Your code marks the screen match itself as the softfailure which make indeed sense. To store the reason you then use `record_serialresult`. That on the other hand is actually not so nice because it looks like `wait_serial` was used in the test code.

What you actually want here is to combine different result types (screenshot and text) but to my knowledge that's not possible (without adjusting also the frontend and making things even more complicated). Maybe the simplest solution is to mark the screenmatch result as softfail by setting `$result->{result} = 'softfail';` as your code already does and then record the reason as a separate text result just as if `record_softfailure` was called. That could be implemented by making use of `record_soft_failure_result`.

If you decide to move increment the dents to basetest which I would advocate you can get rid of increasing the dents for softfailure screenmatch.

---

> function record_soft_failure_result calls record_testresult, it does ++self->{test_count}, as a result, the next result file is 1 more than expected (In my opinion).

That *might* be a problem, but have you tested how the fronted handles actully that? That the frontend can display the results correctly is what matters.  
By the way, it would make sense to share a test run with such a softfailure (e.g. just a screenshot so we see how it looks like).

basetest.pm
```diff
@@ -166,7 +166,16 @@ sub record_screenmatch {
     if ($foundneedle->has_property('workaround')) {
```
Martchus: Maybe one could just write `if (my $reason = $foundneedle->get_property_value('workaround')) {` instead of running the (linear) search algorithm twice?

needle.pm
okurz: What's the problem with the old code?

needle.pm
okurz: IMHO the explicit for loop should be avoided, hence I suggested grep. You can put the check for hash within the grep if it helps

needle.pm
okurz: Have you seen this comment?

Amrysliu: Sorry, I forgot to reply.
Yes, you are right, This is to support the existed needles those have workaround property and the related ticket number was recorded in needle name. Maybe I should add a array to store those ticket head, such as bsc, poo and so on(I did not find out other ticket head in existed needles.)

needle.pm
Amrysliu: Why should the explicit for loop be avoided?  I do not know how is `grep` implemented in Perl.  If there is also a for in its implement, I prefer to keep the `for` loop.  In my opinion, it is more readable.

needle.pm
Martchus: `$self->{properties}` now contains not only a list of strings but also references (to hashes). This must be taken into account when sorting. I don't know how `sort` handles references by default.

Amrysliu:
workaround_call_record_softfailure
workaround_Not_call_record_softfailure

needle.pm
kraih: `sort` on references is pointless, they just get stringified to their address.

Amrysliu:
@Martchus  Please see the above picture.
I did test for two situations:
1. The first picture is the result when code is 
   `$result->{dent} = 1;
    $self->{dents}++;
   my $reason = $foundneedle->get_property_value('workaround');
   $self->record_soft_failure_result($reason);`

2. The second picture is the result when code is 
   `$result->{dent} = 1;
    $self->{dents}++;
    $result->{title} = 'Soft Failed';
    my $filename = $self->next_resultname('txt');
    $result->{text} = $filename;
    $self->write_resultfile($filename, "#Soft Failure: \n$reason\n");
    $result->{result} = "softfail";`
In first picture, user can see the needle's png file by click it. But the png and 'Soft Failure' location seem opposite. Uses can click the 'Soft Failed' to see the workaround property.
In second picture, user can not see the needle's png file when it has workaround property. Because when users click the picture, it will show the workaround reason as the second picture show.

Seems those two forms of expression are not very good.  I prefer the form of expression is that not only users can see the workaround needle, but also show the workaround reason.

Martchus:
I prefer the first version because the user can see the PNG. That is also what I suggested in the previous review comment. But I also said:

> Maybe the simplest solution is to mark the screenmatch result as softfail by setting $result->{result} = 'softfail'; as your code already does [...]

It looks like you didn't do that because otherwise the softfailed screenmatch should have been rendered with a yellow border. Then it would be more clear that this screen mismatch caused the softfailure. So I suggest to do that as well as the things I've mentioned previously.

The 2nd version only works that way because the frontend gives text results precedence over images if a step contains both result types. (By the way, the full order is 1. audio, 2. text, 3. image.)

Amrysliu:
Here is a test result screenshot after I modified the code.
When users click 'Soft Failed', it will show the workaround reason.
When users click the needle file (PNG) with workaround property, it will show the PNG.
Is this form of expression ok? @Martchus 
Screenshot_20190619_164641

basetest.pm
Amrysliu: I modified as your comments.

Martchus:
I think the screenshot looks good now. (I'm only wondering why the font sizes are messed up but I guess that's a different problem.)

testapi.pm
```diff
@@ -181,7 +181,6 @@ sub record_soft_failure {
     bmwqemu::log_call(reason => $reason);
 
     $autotest::current_test->record_soft_failure_result($reason);
-    $autotest::current_test->{dents}++;
```
Martchus: There is another occurrence of this line in `force_soft_failure` which needs to be removed as well.

needle.pm
Martchus: To be honest: The `grep` line looks harder to read than the more verbose but simpler for-loop. Especially for those who are not that familiar with Perl. So from my point of view: Keep it simple. (Actually I don't care much but I wouldn't require you to change anything here.)

okurz:
Now the problem with a text popup referencing the bug id as well as the yellow border around the screenshot is that show the same soft failure twice IMHO that should be avoided. I fear that without changing the way we display the popup for screenshots we can not easily solve the problems you mentioned in before but still ensure that one soft-fail only pops up once. IMHO the whole "feature" seems to get a bit out of hands ;) If you want to go forward I guess the screenshot popup should show the bugref in e.g. a single text line, not in a separate text popup next to it.

needle.pm
okurz: I don't think we can get by learning nice perl without using `grep` or `map` where it fits but I would trust more experienced persons here of course:

@kraih what's your take on this? Explicit for-loop or grep? :)

kraih: Yes, understanding`map` and `grep` is considered an important milestone in learning Perl. I would generally use a `grep` one-liner over a bigger for loop here.

Amrysliu:
> Now the problem with a text popup referencing the bug id as well as the yellow border around the screenshot is that show the same soft failure twice IMHO that should be avoided. I fear that without changing the way we display the popup for screenshots we can not easily solve the problems you mentioned in before but still ensure that one soft-fail only pops up once. IMHO the whole "feature" seems to get a bit out of hands ;) If you want to go forward I guess the screenshot popup should show the bugref in e.g. a single text line, not in a separate text popup next to it.

"In web page only show the PNG with yellow border, when click the PNG, the needle picture will show, and in its popup area, there should be a text line that used to shown workaround reason (such as bsc#123456)", this form of expression is you expected?  In order to achieve this effect, I need to modify the UI code.  What's your opinion? Is that form better? @Martchus

Martchus:
For simplicity I thought rendering 2 boxes - one for the text and one for the screenshot - would be sufficient.

> In order to achieve this effect, I need to modify the UI code.

That is true. And then you actually need the os-autoinst code which produces a result which has text *and* screenshot attached at the same time.

Not sure whether it is worth taking the effort but if it makes @okurz happy....

okurz:
As I stated "IMHO the whole 'feature' seems to get a bit out of hands". Maybe everything will be fine but I am just careful with changing how results are displayed knowing that we also have some tooling reading this data. It is not about "making me happy". I am just providing my concerns but will not stop you from going ahead either way.

Amrysliu:
> As I stated "IMHO the whole 'feature' seems to get a bit out of hands". Maybe everything will be fine but I am just careful with changing how results are displayed knowing that we also have some tooling reading this data. It is not about "making me happy". I am just providing my concerns but will not stop you from going ahead either way.

I agree with your opinion and thanks for your suggestion. I would like to investigate how to achieve this effect. If it needs much code to be modified, I suggest consider it as an enhancement as this PR to avoid this PR is too long. Do you think it is ok?

Amrysliu:
> I think the screenshot looks good now. (I'm only wondering why the font sizes are messed up but I guess that's a different problem.)

I compare the font sizes with openQA, I did not see any different.

okurz:
> I suggest consider it as an enhancement as this PR to avoid this PR is too long

Well I suggest to not introduce a breaking change and then afterwards introduce the backward-compatible handling again so to my understanding it would be better to handle it in the same PR.

Amrysliu:
> > I suggest consider it as an enhancement as this PR to avoid this PR is too long
> 
> Well I suggest to not introduce a breaking change and then afterwards introduce the backward-compatible handling again so to my understanding it would be better to handle it in the same PR.

ok, I knew it.  I did think it is not worth to do this if it need to spend much more time. I will investigate the code and decide to do it or not.

testapi.pm
Amrysliu: I fixed it.

needle.pm
Amrysliu: I used `grep` instead of the `for` loop.

needle.pm
Amrysliu: I would not like to change those code to a `grep`. I think this logic of this piece of code is a little complicated than above, so I prefer to keep its readability.

okurz:
Unfortunately I want to come back to https://github.com/os-autoinst/os-autoinst/pull/1153#issuecomment-503494256 and this time it is not about backward compatibility :)
One major concern: Take for example http://10.67.18.227:9526/tests/6#step/disable_grub_timeout/13 : In case of a needle with the new workaround property there is a text popup with a nice bug link and the needle thumbnail shows up with yellow border. However it is not obvious where the soft-fail comes from, if it is a call in the test code or the needle. So still my suggestion would be to combine the text thumbnail and the needle preview and show the soft-fail text within the needle view to make it obvious that the soft-failure was triggered by the needle not the test code.

Amrysliu:
@okurz Thanks for your suggestion. I also think the form you mentioned is better. But it needs more code to be modified and spend more time. I just think if it is worth to do that.

coolo:
> Unfortunately I want to come back to [#1153 (comment)](https://github.com/os-autoinst/os-autoinst/pull/1153#issuecomment-503494256) and this time it is not about backward compatibility :)
> One major concern: Take for example http://10.67.18.227:9526/tests/6#step/disable_grub_timeout/13 : In case of a needle with the new workaround property there is a text popup with a nice bug link and the needle thumbnail shows up with yellow border. However it is not obvious where the soft-fail comes from, if it is a call in the test code or the needle. So still my suggestion would be to combine the text thumbnail and the needle preview and show the soft-fail text within the needle view to make it obvious that the soft-failure was triggered by the needle not the test code.

Don't try to squeeze everything into one PR please.

okurz:
@coolo sure, that's not what we want. I am just saying I am not feeling confident with this changed behaviour (hence not merging myself).

Amrysliu:
> So if it looks like this I would say this is sufficient and I would not put more effort into it (at least for the time being).

Thanks for your review again and approval.

needle.pm
perlpunk: This actually could have never worked correctly.
Before it was an array reference `properties => [@{ ... }]`, now it is an array reference with one array reference in it.
I cannot see this function being used, and it seems the last usage was removed in 525b71d5c8d86cc009af730003a6ebea47660f11

PR #1158, opened 2019-05-27
Add recursive get data call

cfconrad:
when requesting `curl autoinst_url . "/data"` and the requested URI
points to an directory, we do now retrieve the directory recursive.
The returned data is a cpio-achive.

In action: http://cfconrad-vm.qa.suse.de/tests/4620/file/serial_terminal.txt

### DIFF
--- a/t/07-commands.t
+++ b/t/07-commands.t
@@ -36,10 +36,15 @@ use Test::More;
 use Test::Warnings;
 use Test::Mojo;
 use Devel::Cover;
+use Mojo::File qw(path tempfile tempdir);
+use File::Which;
+use Data::Dumper;
 use POSIX '_exit';
 
 our $mojoport = Mojo::IOLoop::Server->generate_port;
-my $base_url = "http://localhost:$mojoport";
+my $base_url     = "http://localhost:$mojoport";
+my $toplevel_dir = path(__FILE__)->dirname->realpath;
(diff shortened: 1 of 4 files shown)


t/07-commands.t
```diff
@@ -119,6 +123,31 @@ subtest 'web socket route' => sub {
...
 
+subtest 'data api' => sub {
+    my $job = 'Hallo';
+    $t->get_ok("$base_url/$job/data");
+    $t->status_is(200);
+    $t->content_type_is("application/x-cpio");
+
+    $t->get_ok("$base_url/$job/data/mod1");
+    $t->status_is(200);
+    $t->content_type_is("application/x-cpio");
```
cfconrad: I would like to add a check, if the achieve really contains all files.
How do I check make sure, that cpio is available?

okurz: do you mean if the cpio command line tool is available in the environment that runs the tests? As "cpio" is included by default even in the docker containers "opensuse/tumbleweed" or even "mini/base" (alpine) I would not be concerned about environments missing it but still check that the command exists

okurz: Probably better:

```
use File::Which;
ok(-x which 'cpio');
```

cfconrad: exact that was my question. Thanks a lot!

PR #1159, opened 2019-05-28
Always log to autoinst-log.txt

asdil12:
Even when also logging to stderr

https://progress.opensuse.org/issues/52235

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -127,21 +127,25 @@ sub init {
     mkdir result_dir;
     mkdir join('/', result_dir, 'ulogs');
 
+    $logger = Mojo::Log->new(level => 'debug', path => 'autoinst-log.txt');
+    $logger->format(\&format_log);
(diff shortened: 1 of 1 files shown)


asdil12:
Actually nothing but running the testsuite

asdil12:
Ok - I now managed to test it and it works.

pevik:
@Martchus Not only that this PR duplicate logging on every single output line (as noted in https://github.com/os-autoinst/os-autoinst/pull/1161) it also breaks (at least some) s390x builds:
http://quasar.suse.cz/tests/2853#step/bootloader_zkvm/28

pevik:
@Martchus  with manual change from https://github.com/os-autoinst/openQA/pull/2092/files it's working again also for s390x builds. So no need to revert it (e.g. https://github.com/os-autoinst/os-autoinst/pull/1161 is not needed).

PR #1160, opened 2019-05-29
Use _QUIET_SCRIPT_CALLS variable for script_run/output

jlausuch:
This complements PR https://github.com/os-autoinst/os-autoinst/pull/1151 
which adds the flag `quiet` to script_run and script_output.
Using variables, we can make all the calls to those functions in quiet mode without
specifying the flag `quiet => 1` in every call.

Verification RUNs: 
With       : http://fromm.arch.suse.de/tests/6259
Without :  http://fromm.arch.suse.de/tests/6152

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -25,6 +25,7 @@ TIMEOUT_SCALE;integer;1;This scale parameter can be used based on performance of
 PAUSE_AT;string;;Test module (name or fullname) to pause test execution at. To be used together with the openQA developer mode to continue test execution. Be aware that the openQA web UI will only reflect this within the developer mode after confirming to control the test.
 PAUSE_ON_SCREEN_MISMATCH;boolean;0;Pause test execution on the next screen mismatch. Same notes as for `PAUSE_AT` apply.
 PAUSE_ON_NEXT_COMMAND;boolean;0;Pause test execution on the next test API command. Same notes as for `PAUSE_AT` apply.
+_QUIET_SCRIPT_CALLS;boolean;0;Add quiet flag to all the calls to script_run, script_output and validate_script_output. It will omit all the squares "wait_serial expected" on the Details view of the test case. This option might be useful for serial terminal tests.
 
 |====================
 
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -608,6 +608,32 @@ subtest 'compat_args' => sub {
     like(warning { testapi::compat_args(\%def_args, [], 'outch') }->[0], qr/^Odd number of arguments/, 'Warned on Odd number 2');
 };
 
+subtest 'check quiet option on script runs' => sub {
+    $bmwqemu::vars{_QUIET_SCRIPT_CALLS} = 1;
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->mock(wait_serial => sub {
+            my ($regex, %args) = @_;
+            is($args{quiet}, 1, 'Check default quiet argument');
+            return "XXXfoo\nSCRIPT_FINISHEDXXX-0-";
+    });
+    is(script_output('echo foo', 30), 'foo', 'script_output with _QUIET_SCRIPT_CALLS=1 expects command output');
+    is(script_run('true'),        '0',   'script_run with _QUIET_SCRIPT_CALLS=1');
+    is(assert_script_run('true'), undef, 'assert_script_run with _QUIET_SCRIPT_CALLS=1');
+    ok(!validate_script_output('script', sub { m/output/ }), 'validate_script_output with _QUIET_SCRIPT_CALLS=1');
+
+    $mock_testapi->mock(wait_serial => sub {
+            my ($regex, %args) = @_;
+            is($args{quiet}, 0, 'Check default quiet argument');
+            return "XXXfoo\nSCRIPT_FINISHEDXXX-0-";
+    });
+    is(script_output('echo foo', quiet => 0), 'foo', 'script_output with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
+    is(script_run('true', quiet => 0), '0', 'script_run with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
+    is(assert_script_run('true', quiet => 0), undef, 'assert_script_run with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
+    ok(!validate_script_output('script', sub { m/output/ }, quiet => 0), 'validate_script_output with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
+    delete $bmwqemu::vars{_QUIET_SCRIPT_CALLS};
+};
+
+
 done_testing;
 
 # vim: set sw=4 et:
--- a/testapi.pm
+++ b/testapi.pm
@@ -904,7 +904,7 @@ sub _handle_script_run_ret {
 
 =head2 assert_script_run
 
-  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message]);
+  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
 
 Deprecated mode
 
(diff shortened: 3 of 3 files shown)


jlausuch:
@cfconrad @asmorodskyi @foursixnine

distribution.pm
```diff
@@ -129,7 +129,7 @@ sub script_run {
     my %args = testapi::compat_args(
         {
             timeout => $bmwqemu::default_timeout,
-            quiet   => undef
+            quiet   => testapi::get_var('QUIET', 0)
```
foursixnine: add the variable to backend_vars.md (https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc) but also in the meantime, rename it to _QUIET_SCRIPT_CALLS

jlausuch: done

distribution.pm
```diff
@@ -213,8 +213,8 @@ sub script_output {
     my %args = testapi::compat_args(
         {
             timeout            => undef,
-            proceed_on_failure => 0,       # fail on error by default
-            quiet              => undef,
+            proceed_on_failure => 0,                                                # fail on error by default
+            quiet              => testapi::get_var('_QUIET_SCRIPT_CALLS', undef),
```
cfconrad: `undef` is the default, if it's not present.
```suggestion
            quiet              => testapi::get_var('_QUIET_SCRIPT_CALLS'),
```

distribution.pm
```diff
@@ -129,7 +129,7 @@ sub script_run {
     my %args = testapi::compat_args(
         {
             timeout => $bmwqemu::default_timeout,
-            quiet   => undef
+            quiet   => testapi::get_var('_QUIET_SCRIPT_CALLS', undef)
```
cfconrad:
```suggestion
            quiet   => testapi::get_var('_QUIET_SCRIPT_CALLS')
```

jlausuch:
@cfconrad WDYT? I just modified testapi.pm in the end

testapi.pm
```diff
@@ -1049,7 +1049,16 @@ and can be tweaked by setting the C<$wait> positional parameter.
 =cut
 
 sub script_output {
-    return $distri->script_output(@_);
+    my $script = shift;
+    my %args   = testapi::compat_args(
+        {
+            timeout            => undef,
+            proceed_on_failure => 0,                                         # fail on error by default
```
cfconrad:
```suggestion
            proceed_on_failure => undef,                                         # fail on error by default
```

testapi.pm
```diff
@@ -1049,7 +1049,16 @@ and can be tweaked by setting the C<$wait> positional parameter.
...
+            quiet              => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            type_command       => length($script) < 80,
```
cfconrad: I would let the defaults up to the distri implementation for that option.
```suggestion
            type_command       => undef,
```

t/03-testapi.t
```diff
@@ -608,6 +608,21 @@ subtest 'compat_args' => sub {
...
+    $bmwqemu::vars{_QUIET_SCRIPT_CALLS} = 1;
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->mock(wait_serial => sub {
+            my ($regex, %args) = @_;
+            is($args{quiet}, 1, 'Check default quiet argument');
+            return "XXXfoo\nSCRIPT_FINISHEDXXX-0-";
+    });
+    is(script_output('echo foo', 30), 'foo', '');
+    is(script_run('true'),        '0',   'script_run with no check of success, returns exit code');
+    is(assert_script_run('true'), undef, 'nothing happens on success');
```
cfconrad: maybe add `validate_script_output()`, too.

What do you think in testing `_QUIET_SCRIPT_CALLS` is set, but specify `quiet => 0` explicit. I would guess, that then we actually wont `wait_serial()` called with `quiet=>0` as well.

jlausuch: > maybe add `validate_script_output()`, too.

sure


> What do you think in testing `_QUIET_SCRIPT_CALLS` is set, but specify `quiet => 0` explicit. I would guess, that then we actually wont `wait_serial()` called with `quiet=>0` as well.
Yes, that is a good point and I would like that the flag overrides the env variable (_QUIET_SCRIPT_CALLS=1 but quiet=>0). I haven't checked that case, but will do it soon.

doc/backend_vars.asciidoc
```diff
@@ -25,6 +25,7 @@ TIMEOUT_SCALE;integer;1;This scale parameter can be used based on performance of
 PAUSE_AT;string;;Test module (name or fullname) to pause test execution at. To be used together with the openQA developer mode to continue test execution. Be aware that the openQA web UI will only reflect this within the developer mode after confirming to control the test.
 PAUSE_ON_SCREEN_MISMATCH;boolean;0;Pause test execution on the next screen mismatch. Same notes as for `PAUSE_AT` apply.
 PAUSE_ON_NEXT_COMMAND;boolean;0;Pause test execution on the next test API command. Same notes as for `PAUSE_AT` apply.
+_QUIET_SCRIPT_CALLS;boolean;0;Add quiet flag to all the calls to script_run and script_output.
```
asmorodskyi: as far as I can see _QUIET_SCRIPT_CALLS has influence only for case `testapi::is_serial_terminal` returning true , might be usefull to mention this in variable description

cfconrad: From my perspective, `is_serial_terminal` doesn't belong to the `quiet` switch. 
The `quiet` switch avoid the call of `record_serialresult()` in `wait_serial()` and validate_script_output()`. And this apply to  all kinds of consoles.

asmorodskyi: @jlausuch have you tried to run it with some VNC test  case to see if it really make some difference and not produce any errors ?

foursixnine:
@jlausuch @asmorodskyi any last changes here?

jlausuch:
> @jlausuch @asmorodskyi any last changes here?

give me 1 day or so to add UT.

testapi.pm
jlausuch: done

testapi.pm
jlausuch: done

jlausuch:
without flag: http://fromm.arch.suse.de/tests/6364#
with flag: http://fromm.arch.suse.de/tests/6367

jlausuch:
can this be merged?

doc/backend_vars.asciidoc
okurz: can you please also explain here what the benefit of the flag is? Stating that it just adds a flag is certainly correct but also a bit boring, don't you think? ;)

t/03-testapi.t
```diff
@@ -608,6 +608,32 @@ subtest 'compat_args' => sub {
     like(warning { testapi::compat_args(\%def_args, [], 'outch') }->[0], qr/^Odd number of arguments/, 'Warned on Odd number 2');
 };
 
+subtest 'check_quiet_option' => sub {
```
okurz: no need to escape any whitespace within the string. You can use something more explanatory here if you like

testapi.pm
```diff
@@ -932,7 +932,7 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => undef
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
```
okurz: I don't think it's a good idea to access the test API within the methods itself. Do you think it would work to just access the `$bmwqemu::VARS` hash here?

testapi.pm
```diff
@@ -1049,7 +1049,16 @@ and can be tweaked by setting the C<$wait> positional parameter.
...
 
 sub script_output {
-    return $distri->script_output(@_);
+    my $script = shift;
+    my %args   = testapi::compat_args(
+        {
+            timeout            => undef,
+            proceed_on_failure => undef,                                     # fail on error by default
+            quiet              => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            type_command       => undef,
```
okurz: I guess you should extend the doc string on top for this

testapi.pm
foursixnine: Why not?

okurz: because the external API is the contract we give to the customers + error checking, etc. The internal calls we can decide within the context of os-autoinst how to use and skip some indirection.

testapi.pm
jlausuch: done

t/03-testapi.t
jlausuch: ok

doc/backend_vars.asciidoc
jlausuch: done

PR #1161, opened 2019-06-04
Fix missing or double output in autoinst-log.txt with partial revert

okurz:
This reverts commit 85a4003cb5c716744a487d92c5b74977ea1973ab and
partially reverts commit fe70160d7134dfa499cae7d47fafd95e5321dcf8 as
problems in the output handling in conjunction with openQA has been
encountered. The general code interface is preserved and should be
overhauled when the implementation is completed.
 
Related ticket: https://progress.opensuse.org/issues/52235

Was: Reverts os-autoinst/os-autoinst#1159 to prevent the problem as described in https://progress.opensuse.org/issues/52235#note-5 until a proper solution has been found.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -28,7 +28,6 @@ use testapi  ();
 use autotest ();
 use MIME::Base64 'decode_base64';
 use Mojo::File 'path';
-use Try::Tiny;
 
 my $serial_file_pos  = 0;
 my $autoinst_log_pos = 0;
@@ -368,27 +367,13 @@ sub runtest {
         }
     }
 
-    # Detect serial failures and override result if die
-    try {
-        $self->search_for_expected_serial_failures();
-    }
-    catch {
-        # Process serial dectection failure
-        bmwqemu::diag($_);
-        $self->record_resultfile('Failed', $_, result => 'fail');
+    eval { $self->search_for_expected_serial_failures(); };
+    # Process serial dectection failure
+    if ($@) {
+        bmwqemu::diag($@);
+        $self->record_resultfile('Failed', $@, result => 'fail');
         $died = 1;
     }
-
-    # Detect autoinst failures and override result if die
-    try {
-        $self->search_for_expected_autoinst_failures();
-    }
-    catch {
-        # Process autoinst dectection failure
-        bmwqemu::diag($_);
-        $self->record_resultfile('Failed', $_, result => 'fail');
-        $died = 1;
-    };
 
     $self->run_post_fail("test $name died") if ($died);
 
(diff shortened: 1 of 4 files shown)


okurz:
@asdil12 @foursixnine @Martchus

foursixnine:
@asdil12 could you work on improving the fullstack tests too while fixing this part of the problem?

coolo:
Actually I would prefer reverting the complete feature - until it's done correctly

foursixnine:
> Actually I would prefer reverting the complete feature - until it's done correctly

+1

asdil12:
> Actually I would prefer reverting the complete feature - until it's done correctly

-1

Too much work and it would break tests if we don't also revert it in the test distri.
Also it is not the feature that is broken but isotovideo being broken and just noone noticing before.

coolo:
No. your feature broke it by depending on a debug log

coolo:
and we can leave the API in place - as NOP

Martchus:
Seems like we need to revert https://github.com/os-autoinst/os-autoinst/pull/1159 and the changed introducing the bug that PR is attempting to fix. I guess that's also what @coolo means.

coolo:
No, 1159 was already an attempt to cover up. Even earlier

pevik:
Thanks oliver for this PR. This revert is really needed as https://github.com/os-autoinst/os-autoinst/pull/1159 breaks (at least some) s390x builds:
http://quasar.suse.cz/tests/2853#step/bootloader_zkvm/28

asdil12:
Reverting doesn't seem to be needed anymore with https://github.com/os-autoinst/openQA/pull/2092

asdil12:
> Thanks oliver for this PR. This revert is really needed as #1159 breaks (at least some) s390x builds:
> http://quasar.suse.cz/tests/2853#step/bootloader_zkvm/28

How do you relate this fail to that PR?
Is there something in that error message that I overlooked?

pevik:
@asdil12 OK, with manual change from https://github.com/os-autoinst/openQA/pull/2092/files it's working again also for s390x builds. Thanks for pointing it out.

okurz:
I don't know what I am doing but I tried to "disarm #1149" as suggested in https://github.com/os-autoinst/openQA/pull/2093#issuecomment-498738520

okurz:
yeah I tried to keep the "Try::Tiny" changes in place but this would fail the existing implementation and tests of the serial log error detection.

@asdil12 I hope you are also ok with it

okurz:
ok, all included tests are passed. I could reproduce the duplication problem using a local openQA instance, e.g. by running `sudo -u _openqa-worker ~/local/os-autoinst/openQA/script/worker --instance 1 --verbose --no-cleanup` and monitoring /var/lib/openqa/pool/1/autoinst-log.txt and verified that this PR is fine regarding output by running it as worker with `sudo -u _openqa-worker ~/local/os-autoinst/openQA/script/worker --instance 1 --isotovideo ~/local/os-autoinst/os-autoinst/isotovideo  --verbose --no-cleanup` confirming that the logfile looks fine, calling `isotovideo -d` on a local `vars.json` file asserting that the output is there as well as `isotovideo` without `-d` asserting that no debug output is there.

PR #1163, opened 2019-06-05
Format using Perl::Tidy 20190601

Martchus:
For consistency with https://github.com/os-autoinst/openQA/pull/2095

### DIFF
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -92,9 +92,9 @@ sub _gen_node_name {
 
 sub gen_cmdline {
     my ($self) = @_;
-    my @cmdln = ();
-    my $paths = @{$self->paths} < 1 ? [OpenQA::Qemu::DrivePath->new()] : $self->paths;
-    my $pathn = scalar @$paths;
+    my @cmdln  = ();
+    my $paths  = @{$self->paths} < 1 ? [OpenQA::Qemu::DrivePath->new()] : $self->paths;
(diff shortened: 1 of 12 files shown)


okurz:
just keep in mind that os-autoinst-distri-opensuse uses the tidy from os-autoinst hence you should ensure that every QA engineer has access to the corresponding tidy version

foursixnine:
@Martchus @okurz check linked pr.

okurz:
on my openSUSE Leap 15.1 I do not yet have perl-Perl-Tidy-2019… from RPM packages so others might have a problem as well. I mean, I have it from cpanm but we should be nice to other people

Martchus:
We should not require `cpanm`. I'll add perl-Perl-Tidy to our repo for Leap 15.1 as we have it already for Leap 15.0. I guess we should also wait for Tumbleweed (the latest version is in Factory but seems not published yet).

kraih:
For future reference i made [a gist](https://gist.github.com/kraih/01ee1d7900c8bf776b93e7652b84f519) with instructions for how to use `cpanm` to temporarily install and use any version of perltidy in a local lib should the need arise.

Martchus:
@kraih I turned it into scripts which are easy enough to work with so I can remember their usage: https://github.com/Martchus/openQA-helper#installing-a-custom-version-of-perl-dependencies-like-perltidy

Martchus:
Tumbleweed has this tidy version now and we have it in our devel:openQA Leap repos. So I'd say this should be merged now.

PR #1165, opened 2019-06-06
 Add script_output_with_status() 

cfconrad:
This function is like script_output(). But it always return a array
wich contains the output and the exit code of the script.

This is useful, when calling this function with `proceed_on_failure=>1`.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -20,6 +20,7 @@ use warnings;
 
 use testapi ();
 use Carp 'croak';
+use Mojo::Util 'trim';
 
 sub new {
     my ($class) = @_;
@@ -188,16 +189,12 @@ sub script_sudo {
     return;
 }
 
-=head2 script_output
+=head2 script_output_with_status
 
-    script_output($script [, timeout => ?] [, type_command => ?] [, proceed_on_failure => ?] [, quiet => ?])
+    script_output_with_status($script [, timeout => ?] [, type_command => ?] [, proceed_on_failure => ?] [, quiet => ?])
 
-Deprecated mode
-
-    script_output($script, [ $wait, type_command => ?, proceed_on_failure => ?])
-
-Execute $script on the SUT and return the data written to STDOUT by
-$script. See script_output in the testapi.
+Execute C<$script> on the SUT and return an array with the data written to
+STDOUT by C<$script> and the exit value.
 
 C<proceed_on_failure> - allows to proceed with validation when C<$script> is
 failing (return non-zero exit code)
@@ -208,17 +205,12 @@ You may be able to avoid overriding this function by setting
 $serial_term_prompt.
 
 =cut
-sub script_output {
-    my ($self, $script) = splice(@_, 0, 2);
-    my %args = testapi::compat_args(
-        {
-            timeout            => undef,
-            proceed_on_failure => 0,       # fail on error by default
-            quiet              => undef,
-            # 80 is approximate quantity of chars typed during 'curl' approach
-            # if script length is lower there is no point to proceed with more complex solution
-            type_command => length($script) < 80,
-        }, ['timeout'], @_);
+sub script_output_with_status {
+    my ($self, $script, %args) = @_;
+    $args{proceed_on_failure} //= 0;
+    # 80 is approximate quantity of chars typed during 'curl' approach
+    # if script length is lower there is no point to proceed with more complex solution
+    $args{type_command} //= length($script) < 80;
 
     my $marker      = testapi::hashed_string("SO$script");
     my $script_path = "/tmp/script$marker.sh";
@@ -270,8 +262,8 @@ sub script_output {
     }
     my $output = testapi::wait_serial("SCRIPT_FINISHED$marker-\\d+-", timeout => $args{timeout}, record_output => 1, quiet => $args{quiet})
       || croak "script timeout: $script";
-
-    if ($output !~ "SCRIPT_FINISHED$marker-0-") {
+    my ($retval) = $output =~ /SCRIPT_FINISHED$marker-(\d+)-/;
+    if ($retval != 0) {
         my $log_message = 'script failed with : ' . $output;
         if ($args{proceed_on_failure}) {
             bmwqemu::log_call($log_message);
@@ -282,10 +274,29 @@ sub script_output {
     }
 
     # and the markers including internal exit catcher
-    my $out = $output =~ /$marker(?<expected_output>.+)SCRIPT_FINISHED$marker-\d+-/s ? $+ : '';
-    # trim whitespaces
-    $out =~ s/^\s+|\s+$//g;
-    return $out;
+    my $out;
+    $out = $+{output} if ($output =~ /$marker(\r?\n|\025)(?<output>.*)SCRIPT_FINISHED$marker-$retval-/s);
+    return ($out, $retval);
+}
+
+=head2 script_output
+
+Wrap C<script_output_with_status()> and only return the trimmed output.
+Returns empty string, unless start- and end-markers were found.
+=cut
+sub script_output {
+    my ($self, $script) = splice(@_, 0, 2);
+    my %args = testapi::compat_args(
+        {
+            timeout            => undef,
+            proceed_on_failure => undef,
+            quiet              => undef,
+            type_command       => undef,
+        }, ['timeout'], @_);
+
+    my ($out, $reval) = $self->script_output_with_status($script, %args);
+    $out //= '';
+    return trim($out);
 }
 
 =head2 set_expected_serial_failures
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -439,16 +439,19 @@ sub script_output_test {
     $mock_testapi->mock(hashed_string      => sub { return 'XXX' });
     $mock_testapi->mock(is_serial_terminal => sub { return $is_serial_terminal });
 
-    $mock_testapi->mock(wait_serial => sub { return "XXXfoo\nSCRIPT_FINISHEDXXX-0-" });
+    $mock_testapi->mock(wait_serial => sub { return "XXX\nfoo\nSCRIPT_FINISHEDXXX-0-" });
     is(script_output('echo foo'), 'foo', 'sucessfull retrieves output of script');
+    my @array = script_output('echo foo');
+    is(@array,    1,     'No auto array context 1');
+    is($array[0], 'foo', 'No auto array context 2');
 
     $mock_testapi->mock(wait_serial => sub { return 'SCRIPT_FINISHEDXXX-0-' });
     is(script_output('foo'), '', 'calling script_output does not fail if script returns with success');
 
-    $mock_testapi->mock(wait_serial => sub { return "This is simulated output on the serial device\nXXXfoo\nSCRIPT_FINISHEDXXX-0-\nand more here" });
+    $mock_testapi->mock(wait_serial => sub { return "This is simulated output on the serial device\nXXX\nfoo\nSCRIPT_FINISHEDXXX-0-\nand more here" });
     is(script_output('echo foo'), 'foo', 'script_output return only the actual output of the script');
 
-    $mock_testapi->mock(wait_serial => sub { return "XXXfoo\nSCRIPT_FINISHEDXXX-1-" });
+    $mock_testapi->mock(wait_serial => sub { return "XXX\nfoo\nSCRIPT_FINISHEDXXX-1-" });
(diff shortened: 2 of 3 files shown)


cfconrad:
I see and actually wondered why it's not used that often.
What about something like this https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L157 ?
Or passing a scalar as ref, so I could write to it... But this isn't really common in perl or?

Martchus:
I think @kraih can answer this in detail :-)

The short version is that context sensitive returns values often caused unintended behavior of the program leading to (security) bugs.

I think https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L157 is ok. Or you return always a hash or an array.

kraih:
Yes, i could answer that in great detail if necessary. 😉 But i think @Martchus's explanation should be sufficient. It's a common source of security issues, and therefore considered bad practice in the Perl community.

cfconrad:
Ok, I will change it.

> I think https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L157 is ok. Or you return always a hash or an array.
I think it's bad to return always an array, I would need to change ~300 `script_output()` calls in test repo.

I go for the svirt.pm approach.

pevik:
> I see and actually wondered why it's not used that often.
> What about something like this https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L157 ?

@cfconrad Yes, that's my code. I also wanted to use `wantarray` and was asked to use `$args{wantarray}` :).

> Or passing a scalar as ref, so I could write to it... But this isn't really common in perl or?

 I'm still not convinced that `wantarray` is evil, but agree `$args{wantarray}` is safer.

pevik:
@Martchus @kraih is `$args{wantarray}` the only concern in this PR?

Martchus:
From my side it is the only concern. Just don't forget to update the POD documentation, too.

kraih:
`$args{wantarray}` is actually also considered bad practice. Just pick one context and stick with it for the function/method.

Martchus:
But `$args{wantarray}` is at least *explicit* content switching unlike `wantarray` and `return;`. It might not be a nice API but I see not danger here.

cfconrad:
I retrieve the exit code via parameter now. I'm not 100% happy with it, for me it looks more like perl in a C coat. Anyway, for me it's important to get the exit code and this would work.

kraih:
If you need multiple contexts the best practice these days is to use two functions/methods. Like `$object->to_list` and `$object->to_string`.

cfconrad:
I see, but we already have: `script_run`, `script_output`, `validate_script_output` and so on. 
I'm actually unable to find a suitable name :-/ . Maybe `script()` and let all other using it?

General question: We have multiple implementations of that script_XXX things. And in general we could break them down to one and only return what is expected. The detail is a bit different, as we are using `here tag` or native execution. Is there any problem when we use always `here tag` ?

pevik:
@cfconrad I'd merge these into single commit (changing something you defined previously in this PR does not make much sense).

cfconrad:
@pevik yeah I will squash them, once we have agreed :)

t/03-testapi.t
```diff
@@ -441,23 +441,50 @@ sub script_output_test {
...
     is(script_output('foo'), '', 'calling script_output does not fail if script returns with success');
+    $retval = undef;
+    $out    = script_output('foo', exit_code => \$retval);
+    is($out,    '', 'calling script_output does not fail if script returns with success (wantarray)');
+    is($retval, 0,  'calling script_output does not fail if script returns with success (wantarray)');
 
     $mock_testapi->mock(wait_serial => sub { return "This is simulated output on the serial device\nXXXfoo\nSCRIPT_FINISHEDXXX-0-\nand more here" });
     is(script_output('echo foo'), 'foo', 'script_output return only the actual output of the script');
+    $retval = undef;
+    $out    = script_output('foo', exit_code => \$retval);
```
pevik: Oh, this is really C like approach :). I'm not sure about it, I'd prefer previous version which returns array or scalar, even it's "considered bad practice" :).

kraih:
Since it would be related to `script_output` a name like `script_output_and_status` would make sense. And they can share the same code internally.

pevik:
Uh, *and* in function name looks ugly to me. But that's certainly does not follow C approach :).

cfconrad:
But the idea is clear. Maybe we find a better name...

cfconrad:
I chose the name `script_output_with_status()`. The implementation is slightly different from `script_output()`:

1. I do not trim the output. So the output is (except from extra output which might appear on serial) what the script really return, even if it is empty.
2. If there is a matching error on start- or end-marker, the output is `undef` instead of `empty-string`.

cfconrad:
See it in action: http://cfconrad-vm.qa.suse.de/tests/4679

distribution.pm
```diff
@@ -282,12 +283,24 @@ sub script_output {
...
-    return $out;
+    my $out;
+    $out = $+{output} if ($output =~ /$marker(\r?\n|\025)(?<output>.*)SCRIPT_FINISHED$marker-$retval-/s);
+    return ($out, $retval);
+}
+
+=head2 script_output
+
+Wrap C<<script_output_with_status()>> and only return the
+script the trimmed output.
```
okurz: that sentence reads weirdly now: "only returns the script the trimmed output"?

coolo: > that sentence reads weirdly now: "only returns the script the trimmed output"?

[side note]
your use of an adverb makes it sound like the sentense is reading to you in a weird way. I fail to find if your english is correct or not though :)

okurz: you mean it sounds to you like the sentence as an active object is reading something out to me? ;) https://forum.wordreference.com/threads/the-sentence-reads.2106906/#post-10541813

btw, there are double `<<` where should be single ones only, i.e. `Wrap C<script_output_with_status> and only …`

cfconrad: Changed to 
```
Wrap C<script_output_with_status()> and only return the trimmed output.
Return empty string, if start- or end-markers where not found accordingly.
```

okurz: s/where/were/

distribution.pm
```diff
@@ -282,10 +274,29 @@ sub script_output {
...
+}
+
+=head2 script_output
+
+Wrap C<script_output_with_status()> and only return the trimmed output.
+Return empty string, if start- or end-markers were not found accordingly.
+=cut
+sub script_output {
+    my ($self, $script) = splice(@_, 0, 2);
+    my %args = testapi::compat_args(
```
cfconrad: I would like to avoid the deprecated argument style in `distribution.pm`. Of course we keep it in `testapi.pm`.WDYT?

cfconrad:
Run with latest version: http://cfconrad-vm.qa.suse.de/tests/4738

distribution.pm
cfconrad: thx for pointing that out!

distribution.pm
okurz: well, yes. The `testapi.pm` should be considered our external API. Unfortunately this `script_output` might be used directly. Have you checked os-autoinst-distri-opensuse?

cfconrad: There might be a problem in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm
It is based on `distribution` and calls sometime `$self->script_run()` https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm#L314

So I would keep it for now.

cfconrad: There might be a problem in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm
It is based on `distribution` and calls sometime `$self->script_run()` https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm#L314

So I would keep it for now.

cfconrad:
From my side, this PR is ready to merge. Any comments?

distribution.pm
```diff
@@ -188,16 +189,12 @@ sub script_sudo {
...
-
-Deprecated mode
+=head2 script_output_with_status
 
-    script_output($script, [ $wait, type_command => ?, proceed_on_failure => ?])
+    script_output_with_status($script [, timeout => ?] [, type_command => ?] [, proceed_on_failure => ?] [, quiet => ?])
 
-Execute $script on the SUT and return the data written to STDOUT by
-$script. See script_output in the testapi.
+Execute C<$script> on the SUT and return a array with the data written to STDOUT
```
Martchus: an array

distribution.pm
```diff
@@ -188,16 +189,12 @@ sub script_sudo {
...
+by C<$script> and  the exit value.
```
Martchus: double space between and and the

testapi.pm
```diff
@@ -1061,6 +1049,36 @@ sub script_output {
     return $distri->script_output($script, %args);
 }
 
+=head2 script_output_with_status
+
+  my ($out, $retval) = script_output_with_status($script [, timeout => ?] [, type_command => ?] [, proceed_on_failure => ?] [, quiet => ?])
+
+Executing script inside SUT with C<bash -eox> (in case of serial console with C<bash -eo>)
```
Martchus: Executes a script inside the SUT ...

testapi.pm
```diff
@@ -1061,6 +1049,36 @@ sub script_output {
...
+and directs C<stdout> (I<not> C<stderr>!) to the serial console and returns
```
Martchus: Make a fullstop after console and start a new sentence.

cfconrad:
Added documentation changes with commit https://github.com/os-autoinst/os-autoinst/pull/1165/commits/3501183001d3136fdede4a9119806d91ba462155 
I'm going to squash them, when we are aligned.

okurz:
It seems we have a regression in the code coverage now. Can you please try to bump it? codecov reports only 5.88% of diff hit and the overall coverage seems to be below the target of 39%. It might be that there is some flakyness in the code coverage analysis or that actually other changes have caused this but I still suggest to take it serious.

cfconrad:
I'm not sure how. I added the function `script_output_with_status()` and the corresponding tests in `t/03-testapi.t`.
Also this https://codecov.io/gh/os-autoinst/os-autoinst/compare/bca214ee1dfc257c694cf1b4650dec0fa7d984a1...ac6724d66954ffd416f8a56a1e415ad7f514a504/src/testapi.pm#L1078 shows a red line, but isn't it covered by this test https://github.com/os-autoinst/os-autoinst/pull/1165/files#diff-28df6c42c0c36431d855e299c59f56f9R507
@okurz do you have a hint?

okurz:
The following ideas come to mind:

* Could be that our tests have a inconsistent behaviour. Try to retrigger the tests multiple times and see if the coverage result changes
* Also check the numbers reported in https://travis-ci.org/os-autoinst/os-autoinst/builds/554166645#L1009 against the report from codecov
* Try to gather the coverage results locally, e.g. just with the html based coverage output and look into the html report if the lines are covered

pevik:
Uh, codecov makes os-autoinst development more difficult and I have doubts about its beneficial effect. It's important to have good test coverage, but IMHO this must be reviewed, not depending on amount of lines being added.

testapi.pm
```diff
@@ -1061,6 +1049,37 @@ sub script_output {
...
+
+The default timeout for the script is based on the default in C<wait_serial>
+and can be tweaked by setting the C<$timeout> parameter.
+
+Use C<quiet> to avoid recording serial_results or test variable C<_QUIET_SCRIPT_CALLS>.
+=cut
+
+sub script_output_with_status {
+    my ($script, %args) = @_;
+    $args{quiet} //= testapi::get_var('_QUIET_SCRIPT_CALLS');
```
pevik: Interesting => setting new standard to use underscore prefix for library variables. I like it, but it could be noted also in git commit message (I guess as it's in perldoc it will be in the docs).

Also adding new dependency (`Mojo::Util`) could be noted.

okurz:
> It's important to have good test coverage, but IMHO this must be reviewed, not depending on amount of lines being added.

sure, we take it as additional data to consider, not a hard rule.

testapi.pm
cfconrad: This _QUIET_SCRIPT_CALLS comes from https://github.com/os-autoinst/os-autoinst/pull/1160
I thought `Mojo::Util` is a common mojo library and we are using it. So I thought it isn't a new one.

coolgw:
> This function is like script_output(). But it always return a array
> wich contains the output and the exit code of the script.
> 
> This is useful, when calling this function with `proceed_on_failure=>1`.

typo ? wich :)

okurz:
Please open again when you want to continue your work. Thanks. Cleaning up old open PRs.

PR #1166, opened 2019-06-17
Generalize Net::SSH2 usage in svirt backend

cfconrad:
Moved methods like run_ssh_cmd() to backend::baseclass(). So each
backend can use it in the same way.

All other run_cmd() functions, which use Net::SSH2 in the background, are
based on run_ssh() and run_ssh_cmd().

By default each call to run_ssh_cmd() with equal credentials, will use
the same SSH Connection. A caller can force a new connection with
`keep_open => 0`.
Each ssh connection is stored in `$self->{ssh_connection}->{$key}`

Ensure none-blocking for permanent connections, which use select() in the
read/write loop.

Hyperv really use 2 different host. One host for running
powershell commands and the other as an intermediate host for serial
connection.
Also there are sequential calls on the hyperv-intermediate host which
must use the same ssh-connection.
Commands like:
  `if not exist N: ( mount \\openqa.suse.de\var\lib\openqa\share\factory N: )`
and
  `if not exist C:\cache\SLE-12-SP5-Server-DVD-x86_64-Build0201-Media1.iso \
       copy /Z /Y N:\iso\SLE-12-SP5-Server-DVD-x86_64-Build0201-Media1.iso \
       :\cache\ )`
And if we disconnect after the first command, the windows usersession-clean
umount the device and the later command fail.

Progress Ticked: https://progress.opensuse.org/issues/45863

The PR is set to WIP as I wont to make further testings. @pevik can you help me with that?

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -65,6 +65,7 @@ sub new {
     $self->{min_image_similarity} = 10000;
     $self->{min_video_similarity} = 10000;
     $self->{children}             = [];
+    $self->{ssh_connections}      = {};
 
     return $self;
 }
@@ -348,6 +349,7 @@ sub stop_vm {
         $self->{encoder_pipe} = undef;
         $self->{started}      = 0;
     }
+    $self->close_ssh_connections();
     $self->close_pipes();    # does not return
     return;
 }
@@ -1179,7 +1181,20 @@ sub retry_assert_screen {
 # shared between svirt and s390 backend
 sub new_ssh_connection {
     my ($self, %args) = @_;
+    bmwqemu::log_call(%args);
+    my %credentials = $self->get_ssh_credentials;
+    $args{$_} //= $credentials{$_} foreach (keys(%credentials));
     $args{username} ||= 'root';
+    $args{keep_open} //= 0;
+    my $connection_key;
+
+    # e.g. using hyperv_intermediate host which is running Windows need to keep the connection.
+    # Otherwise a mount point doesn't exists within the next command.
+    if ($args{keep_open}) {
+        $connection_key = join(',', map { $_ . "=" . $args{$_} } qw(hostname username password));
+        my $con = $self->{ssh_connections}->{$connection_key};
+        return $con if (defined($con));
+    }
 
     my $ssh = Net::SSH2->new;
 
@@ -1195,7 +1210,7 @@ sub new_ssh_connection {
                 # this relies on agent to be set up correctly
                 $ssh->auth_agent($args{username});
             }
-            bmwqemu::diag "Connection to $args{username}\@$args{hostname} established" if $ssh->auth_ok;
+            bmwqemu::diag "SSH connection to $args{username}\@$args{hostname} established" if $ssh->auth_ok;
             last;
         }
         else {
@@ -1207,13 +1222,21 @@ sub new_ssh_connection {
     }
     OpenQA::Exception::SSHConnectionError->throw(error => "Error connecting to <$args{username}\@$args{hostname}>: $@") unless $ssh->auth_ok;
 
+    $self->{ssh_connections}->{$connection_key} = $ssh if ($args{keep_open});
     return $ssh;
+}
+
+=head2 get_ssh_credentials
+Should return a hash with the keys: C<host, username, password>
+=cut
+sub get_ssh_credentials {
+    return;
 }
 
 # open another ssh connection to grab the serial console
 sub start_ssh_serial {
     my ($self, %args) = @_;
-
+    bmwqemu::log_call(%args);
     $self->stop_ssh_serial;
 
     my $ssh  = $self->{serial}      = $self->new_ssh_connection(%args);
@@ -1260,6 +1283,59 @@ sub check_ssh_serial {
 
     bmwqemu::diag("svirt serial: unable to read: $error_string (error code: $error_code)");
     return 1;
+}
+
+=head2 run_ssh_cmd
+
+   $ret = run_ssh_cmd($cmd [, username => ?][, password => ?][,host => ?]);
+   ($ret, $stdout, $stderr) = run_ssh_cmd($cmd [, username => ?][, password => ?][,host => ?], wantarray => 1);
+
+=cut
+sub run_ssh_cmd {
+    my ($self, $cmd, %args) = @_;
+    my ($stdout, $stderr) = ('', '');
+    $args{wantarray} //= 0;
+    $args{keep_open} //= 1;
+
+    bmwqemu::log_call(cmd => $cmd, %args);
+    my ($ssh, $chan) = $self->run_ssh($cmd, %args);
+    $chan->send_eof;
+
+    while (!$chan->eof) {
+        if (my ($o, $e) = $chan->read2) {
+            $stdout .= $o;
+            $stderr .= $e;
+        }
+    }
+
+    bmwqemu::diag("[run_ssh_cmd($cmd)] stdout:$/$stdout") if length($stdout);
+    bmwqemu::diag("[run_ssh_cmd($cmd)] stderr:$/$stderr") if length($stderr);
+    my $ret = $chan->exit_status();
+    bmwqemu::diag("[run_ssh_cmd($cmd)] exit-code: $ret");
+    $ssh->disconnect() unless ($args{keep_open});
+
+    return $args{wantarray} ? ($ret, $stdout, $stderr) : $ret;
+}
+
+sub run_ssh {
+    my ($self, $cmd, %args) = @_;
+    bmwqemu::log_call(cmd => $cmd, %args);
+    $args{blocking} //= 1;
+    my $ssh  = $self->new_ssh_connection(%args);
+    my $chan = $ssh->channel() || $ssh->die_with_error("Unable to create SSH channel for executing \"$cmd\"");
+    $chan->exec($cmd) || $ssh->die_with_error("Unable to execute \"$cmd\"");
+    $ssh->blocking($args{blocking});
+    return ($ssh, $chan);
+}
+
+sub close_ssh_connections {
+    my $self = shift;
+    my $cons = $self->{ssh_connections} // {};
+    for my $key (keys(%{$cons})) {
+        bmwqemu::diag("SSH disconnect $key");
+        $cons->{$key}->disconnect();
+        delete($cons->{$key});
+    }
 }
 
 sub stop_ssh_serial {
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1181,7 +1181,7 @@ sub retry_assert_screen {
 # shared between svirt and s390 backend
 sub new_ssh_connection {
     my ($self, %args) = @_;
-    bmwqemu::log_call(%args);
+    bmwqemu::log_call(%{$self->hide_password(%args)});
     my %credentials = $self->get_ssh_credentials;
     $args{$_} //= $credentials{$_} foreach (keys(%credentials));
     $args{username} ||= 'root';
@@ -1191,7 +1191,7 @@ sub new_ssh_connection {
     # e.g. using hyperv_intermediate host which is running Windows need to keep the connection.
     # Otherwise a mount point doesn't exists within the next command.
     if ($args{keep_open}) {
-        $connection_key = join(',', map { $_ . "=" . $args{$_} } qw(hostname username password));
+        $connection_key = join(',', map { $_ . "=" . $args{$_} } qw(hostname username));
         my $con = $self->{ssh_connections}->{$connection_key};
         return $con if (defined($con));
     }
@@ -1236,7 +1236,7 @@ sub get_ssh_credentials {
 # open another ssh connection to grab the serial console
 sub start_ssh_serial {
     my ($self, %args) = @_;
-    bmwqemu::log_call(%args);
+    bmwqemu::log_call(%{$self->hide_password(%args)});
     $self->stop_ssh_serial;
 
     my $ssh  = $self->{serial}      = $self->new_ssh_connection(%args);
@@ -1297,7 +1297,7 @@ sub run_ssh_cmd {
     $args{wantarray} //= 0;
     $args{keep_open} //= 1;
 
-    bmwqemu::log_call(cmd => $cmd, %args);
+    bmwqemu::log_call(cmd => $cmd, %{$self->hide_password(%args)});
     my ($ssh, $chan) = $self->run_ssh($cmd, %args);
     $chan->send_eof;
 
@@ -1319,7 +1319,7 @@ sub run_ssh_cmd {
 
 sub run_ssh {
     my ($self, $cmd, %args) = @_;
-    bmwqemu::log_call(cmd => $cmd, %args);
+    bmwqemu::log_call(cmd => $cmd, %{$self->hide_password(%args)});
     $args{blocking} //= 1;
     my $ssh  = $self->new_ssh_connection(%args);
     my $chan = $ssh->channel() || $ssh->die_with_error("Unable to create SSH channel for executing \"$cmd\"");
@@ -1347,6 +1347,12 @@ sub stop_ssh_serial {
     $self->{select_read}->remove($ssh->sock);
     $ssh->disconnect;
     return $self->{serial} = undef;
+}
+
+sub hide_password {
+    my ($self, %args) = @_;
+    $args{password} = 'SECRET' if ($args{password});
+    return \%args;
 }
 
 # Send TERM signal to any child process
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -106,81 +106,12 @@ sub do_stop_vm {
 }
 
 # Log stdout and stderr and return them in a list (comped).
-sub get_ssh_output {
-    my ($chan) = @_;
-    die 'No channel found' unless $chan;
-
-    my ($stdout, $stderr) = ('', '');
-    while (!$chan->eof) {
-        if (my ($o, $e) = $chan->read2) {
-            $stdout .= $o;
-            $stderr .= $e;
-        }
-    }
-    chomp($stdout, $stderr);
-    bmwqemu::diag("Command's stdout:\n$stdout") if length($stdout);
-    bmwqemu::diag("Command's stderr:\n$stderr") if length($stderr);
-    return ($stdout, $stderr);
-}
-
-=head2 run_cmd($ssh, $cmd, %args);
-Runs command to libvirt host over SSH, logs stdout and stderr of the command.
-
-Returns either exit code itself or with stdout and stderr (both in blocking
-mode. C<$args{nonblock}> is for long running process, returns $ssh and $sock
-for handling the process.
-
-# Examples:
-    my $ret = $svirt->run_cmd($ssh, "virsh snapshot-create-as snap1");
-    die "snapshot creation failed" unless $ret == 0;
-
-    my ($ret, $stdout, $stderr) = $svirt->run_cmd($ssh, "grep -q '$marker' $log", wantarray => 1);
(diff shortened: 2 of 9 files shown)


backend/baseclass.pm
```diff
@@ -1219,21 +1228,72 @@ sub start_ssh_serial {
...
     my ($self, $fh) = @_;
+    my $read_size = 1024;
+    my $buffer;
 
     if ($self->{serial} && $self->{serial}->sock == $fh) {
-        my $chan = $self->{serial_chan};
-        my $line = <$chan>;
-        if (defined $line) {
-            print $line;
+        if (defined($self->{serial_chan}->read($buffer, $read_size))) {
```
cfconrad: Not sure if this could cause problems in other places, my tests didn't show any. 
For me it is much more usual to read the output in blocks, then line by line. We also avoid a warning with this change:
```
Calling Net::SSH2::Channel::readline in non-blocking mode is usually a programming error
```

backend/baseclass.pm
```diff
@@ -1219,21 +1228,72 @@ sub start_ssh_serial {
...
         }
         return 1;
     }
     return;
 }
 
+=head2 run_ssh_cmd
+
+   $ret = run_ssh_cmd($cmd [, username => ?][, password => ?][,host => ?]);
+   ($ret, $stdout, $stderr) = run_ssh_cmd($cmd [, username => ?][, password => ?][,host => ?], wantarray => 1);
```
cfconrad: Maybe we should avoid `wantarray` and create a method called `run_ssh_cmd_with_output()` ?

pevik:
@cfconrad Generally I'd appreciate more atomic commits then one big, that's easier to review.
Or do these interfere that much that it must be a single commit? At least move to `backend.pm` could be separated.

cfconrad:
@pevik do you mean more then one commit in this PR or multiple PR's.
I thought also how to split them up. Or at least make smaller steps. But I actually just did "one" cleanup, which was using only one `SSH->exec()` and one `SSH->channel->read()` related function.
The rest is more or less cleanups to use this implementation.

The less thing is the change in `check_ssh_serial()` which could be done in a separate PR. But for me it's so close, that is reason I added it here. No problem if we wont to exclude it here...

pevik:
@cfconrad I meant more commits (1 PR would be enough). There are other changes: using `read()`, removing default hostname and password in `run_ssh_cmd()`, etc. The latter one is probably wrong as Hyper-V and/or VMware tests needs it. I'll try to test it for you, but currently the tests are broken for another reason (http://quasar.suse.cz/tests/2903#step/system_role/3)

I've learned from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6841 that smaller steps helps during rebasing :). svirt based changes are needed to be tested (at least) on s390x (`zkvm` and `s390x-kvm-sle1{2,5}`), Hyper-V ( at least `textmode@svirt-hyperv`, `media_upgrade*` and `proxyscc_upgrade*`) and VMware (`textmode@svirt-vmware`).

cfconrad:
>  removing default hostname and password in `run_ssh_cmd()`, etc. The latter one is probably wrong as Hyper-V and/or VMware tests needs it. I'll try to test it for you, but currently the tests are broken for another reason (http://quasar.suse.cz/tests/2903#step/system_role/3)
>

Can you point me to the code where the default password was placed. I only found placed where it was read from vars.json variables.
My implementation should use the same variables as the old one. `run_ssh_cmd()` is calling `get_ssh_credentials()` to get the current one. And each backend which is using SSH can override this function to make there "default".
The problem with consoles, you need to pass the credentials, cause they do not extend the backend.
But for instance in `sshVirtsh.pm` has there own wrappers for `run_ssh_cmd()`, which exactly does this for you.

Regarding the read thing. I will extract it in a separate commit. Or maybe different PR.

pevik:
> Regarding the read thing. I will extract it in a separate commit. Or maybe different PR.

@cfconrad If it's close enough, I'd keep it in the same PR. svirt requires huge validation, so it's better to validate all in once.

consoles/sshVirtsh.pm
```diff
@@ -636,29 +625,8 @@ sub get_cmd_output {
...
-    }
-
-    # read output and close channel
-    bmwqemu::diag "Command executed: $cmd";
-    my @cmd_output = backend::svirt::get_ssh_output($chan);
-    $chan->send_eof();
-    $chan->close();
-    return $wantarray ? \@cmd_output : $cmd_output[0];
+    my @ret       = $self->backend->run_ssh_cmd($cmd, $self->get_ssh_credentials($domain), wantarray => 1);
+    return $wantarray ? ($ret[1], $ret[2]) : $ret[1];
```
cfconrad: This is wrong. The previous version did return a Array ref instead of array.

backend/svirt.pm
```diff
@@ -244,18 +174,17 @@ sub load_snapshot {
     my $rsp;
     my $post_load_snapshot_command = '';
     if (check_var('VIRSH_VMM_FAMILY', 'hyperv')) {
-        my $ps          = 'powershell -Command';
-        my %credentials = {hostname => get_required_var('VIRSH_GUEST'), password => get_var('VIRSH_GUEST_PASSWORD')};
```
mnowaksuse: Previously (https://github.com/os-autoinst/os-autoinst/pull/1166/files#diff-447b3fde9d20bc30cc5c2ab06c1184deL166) `VIRSH_HOSTNAME`/`VIRSH_PASSWORD` was the default tuple for SSH connections, but on Hyper-V - due to a different architecture -, we need to connect to jump-host, hence the `VIRSH_GUEST`/`VIRSH_GUEST_PASSWORD` combination instead. Unless the defaults are no more, this should break on Hyper-V.

cfconrad: Good catch. Thanks!
My fault was, that I only checked on VIRSH_VMM_FAMILY but didn't realized that we do have two different SSH credentials inside one family.
Can you give me some background why we sometimes need that jump-host and sometimes not?

cfconrad: argh, github compare confused me and hide the relevant lines. 
Take a look to https://github.com/os-autoinst/os-autoinst/pull/1166/files#diff-447b3fde9d20bc30cc5c2ab06c1184deR209 which is called when a SSH-Connection gets created. It has the `VIRSH_GUEST/VIRSH_GUEST_PASSWORD` credentials for `VIRSH_VMM_FAMILY eq 'hyperv'`.

So I think we use  `VIRSH_GUEST/VIRSH_GUEST_PASSWORD` here.

pevik:
> Can you give me some background why we sometimes need that jump-host and sometimes not?

@cfconrad ones are for connecting to the host and second to guest. Or am I wrong, @mnowaksuse ?

> I've learned from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/6841 ...

@cfconrad Actually more info about issues I bump into with hyperv/vmware, isn't in test repo, but in this PR https://github.com/os-autoinst/os-autoinst/pull/1109 in os-autoinst.

backend/svirt.pm
mnowaksuse: Ah, right, I didn't know this one got it.

The background info about Hyper-V is that openQA requires VNC but Hyper-V only provides RDP. Jump host is the place where we feed RDP data to Xvnc so openQA can consume it.

cfconrad:
VR x390: http://cfconrad-vm.qa.suse.de/tests/4786
VR Hyperv: http://cfconrad-vm.qa.suse.de/tests/4787

mnowaksuse:
> VR Hyperv: http://cfconrad-vm.qa.suse.de/tests/4787

Thanks for doing the run. Please, run some Hyper-V migration scenario in the _Virtualization-Milestone_ group as well as one Hyper-V scenario from the _JeOS_ group. Running Xen PV/HVM jobs from the _JeOS_ and _Functional: Server_ group which leverage `svirt` backend is a good idea as well. Thanks!

cfconrad:
@mnowaksuse thx for the hints.
I realized that it will not run with eachtime a new SSH connection.
I run into trouble with the hyperv-intermediate host. The problem was that one `run_cmd()` call executed `if not exist N: ( mount \\openqa.suse.de\var\lib\openqa\share\factory N: )` and the next a `if not exist C:\cache\SLE-12-SP5-Server-DVD-x86_64-Build0201-Media1.iso ( copy /Z /Y N:\iso\SLE-12-SP5-Server-DVD-x86_64-Build0201-Media1.iso C:\cache\ )`.
But this failed, cause if the SSH connections get disconnected, windows cleanup the usersession and also the mount point...

So I keep now the connection for all run_ssh_cmd() calls within the same credentials.

cfconrad:
I didn't found a Xen HVM job which I could clone. 

But I run some other tests:
* http://cfconrad-vm.qa.suse.de/tests/4823 (jeos-main@svirt-xen-pv)
* http://cfconrad-vm.qa.suse.de/tests/4819 (proxyscc_upgrade_sles12sp4_allpatterns_hyperv@svirt-hyperv)
* http://cfconrad-vm.qa.suse.de/tests/4821 (install_ltp+sle+Server-DVD@s390x-kvm-sle12) *result fail only in case of "not enough free space on disk"*
* http://cfconrad-vm.qa.suse.de/tests/4820 (create_hdd_minimal_base+sdk@s390x-kvm-sle12)
* http://cfconrad-vm.qa.suse.de/tests/4818 (memtest@svirt-hyperv)

backend/baseclass.pm
```diff
@@ -1212,28 +1234,82 @@ sub start_ssh_serial {
         $ssh->die_with_error("Unable to establish SSH channel for serial console");
     }
     $chan->blocking(0);
-    $chan->pty(1);
+    $chan->pty(1);    # TODO does this work?! Doc say it should be a terminal name.
```
Martchus: You're right. We should fix this (in another PR).

Martchus:
So it would be ready from your side?

cfconrad:
Yes from my side it is ready.

cfconrad:
Lets wait for some comments from @mnowaksuse @pevik , thx.

pevik:
Generally LGTM. Nit: commit message in `check_ssh_serial(): use read() instead of readline()` commit:
`os-autoinst.log` => it's `autoinst-log.txt` and `serial.log` => it's `serial0.txt` (just avoid confusing people :)).

I've run some tests (LTP on s390x, svirt-hyperv and ppc64le-spvm), let's see on Monday, whether they pass:

http://quasar.suse.cz/tests/3130 OK
http://quasar.suse.cz/tests/3131
http://quasar.suse.cz/tests/3132 OK
http://quasar.suse.cz/tests/3133
http://quasar.suse.cz/tests/3134 (ppc64le-spvm failed, but that's due not booting - maybe `/mnt` vs. `/mount` in PRG and NUE http://quasar.suse.cz/tests/3134#step/bootloader/25; maybe @dzedro would know more).

pevik:
I see you removed `svirt::get_ssh_output()`, even it's not that important, maybe it's worth of mentioning in commit message.

pevik:
It looks good to me, looks working, thus I'm for merging it.
Just two comments:
* 1st commit is broken as backend/spvm.pm is still using `get_ssh_output` from backend/svirt.pm, which has been removed. Can you please fix it? v(keep the function and remove it in 2nd commit or squash 1st and 2nd commit).
* was 4rd commit (`Make ssh connection persistent`) needed because changes in your previous commits? Then it should be at least noted in commit message what broke it
(it's a good practice not break any commit => helps when bisecting).

cfconrad:
I'm going to squash 1, 2 and 4 as they all belong together and are not separated. 
I had add the 4th commit only to show what I changed and make the review more easy. But it is a needed changed for this PR and without it, the 1st commit will not work properly.

dzedro:
> Generally LGTM. Nit: commit message in `check_ssh_serial(): use read() instead of readline()` commit:
> `os-autoinst.log` => it's `autoinst-log.txt` and `serial.log` => it's `serial0.txt` (just avoid confusing people :)).
> 
> I've run some tests (LTP on s390x, svirt-hyperv and ppc64le-spvm), let's see on Monday, whether they pass:
> 
> http://quasar.suse.cz/tests/3130 OK
> http://quasar.suse.cz/tests/3131 OK
> http://quasar.suse.cz/tests/3132 OK
> http://quasar.suse.cz/tests/3133
> http://quasar.suse.cz/tests/3134 (ppc64le-spvm failed, but that's due not booting - maybe `/mnt` vs. `/mount` in PRG and NUE http://quasar.suse.cz/tests/3134#step/bootloader/25; maybe @dzedro would know more).

better run it on osd http://open.qa/docs/#_triggering_tests_based_on_an_any_remote_git_refspec_or_open_github_pull_request
Prague PXE does not have same structure and mounts as NUE PXE, ask mcaj/infra

pevik:
@dzedro Thanks for a tip!

cfconrad:
@dzedro I don't get it, Isn't `CASEDIR` variable only for the tests repo like `os-autoinst-distri-opensuse`.

Martchus:
@cfconrad I guess you're right. So the suggestion to use `openqa-clone-custom-git-refspec` is not helpful to test this PR.

cfconrad:
Agree, I will add some tests related to the ssh utility functions.

cfconrad:
Added SSH usage tests. Sometime Codecov actually announced an increase of coverage :).

Also run several tests:

- proxyscc_upgrade_sles12sp4_allpatterns_hyperv@svirt-hyperv  http://cfconrad-vm.qa.suse.de/tests/5677
- install_ltp+sle+Server-DVD@s390x-kvm-sle12 http://cfconrad-vm.qa.suse.de/tests/5681
- memtest@svirt-hyperv http://cfconrad-vm.qa.suse.de/tests/5682
- jeos-main@svirt-hyperv http://cfconrad-vm.qa.suse.de/tests/5680 (Fail in apache)
- textmode_svirt@svirt-hyperv http://cfconrad-vm.qa.suse.de/tests/5679

From my point of view, none of these fail during this change.
Consider this PR as ready. WDYT?

backend/svirt.pm
```diff
@@ -106,81 +106,11 @@ sub do_stop_vm {
...
-}
-
-=head2
-See parameters and examples at C<run_cmd>.
-=cut
-sub run_ssh_cmd {
-    my ($self, $cmd, %args) = @_;
-
-    $args{hostname} //= get_required_var('VIRSH_HOSTNAME');
-    $args{password} //= get_var('VIRSH_PASSWORD');
```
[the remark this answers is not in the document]
cfconrad: If I'm not wrong it is here https://github.com/os-autoinst/os-autoinst/pull/1166/files#diff-447b3fde9d20bc30cc5c2ab06c1184deR210
And is in a test here https://github.com/os-autoinst/os-autoinst/pull/1166/files#diff-f71ac3517c371430265781487e96b828R158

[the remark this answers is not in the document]
pevik: @cfconrad strange, my comment has "14 hours ago", but I asked this question during initial version of this PR. I know you fixed it long time ago.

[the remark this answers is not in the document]
cfconrad: urgs that is very strange! Looks like a bug in github then! Maybe you had a review session open...
Consider this as done then.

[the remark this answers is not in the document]
pevik: @cfconrad sure, I consider this PR ready to be merged.

okurz:
see https://travis-ci.org/os-autoinst/os-autoinst/builds/571418444#L810 "../test-driver: line 95: ./24-spvm.t: Permission denied"

cfconrad:
thx for the hint!

okurz:
please rebase. You are welcome to ping me also over other channels to make sure you won't need another rebase :)

cfconrad:
VR  (still running):

- http://cfconrad-vm.qa.suse.de/t6250 - jeos-main_xenpv_image => passed
- http://cfconrad-vm.qa.suse.de/t6248 - install_ltp+sle+Server-DVD => passed
- http://cfconrad-vm.qa.suse.de/t6246 - proxyscc_upgrade_sles12sp4_allpatterns_hyperv => failed (looks like a needle problem)
- http://cfconrad-vm.qa.suse.de/t6247 - textmode_svirt => running
- http://cfconrad-vm.qa.suse.de/t6249 - memtest => scheduled

cfconrad:
Hm I think it is something with `Test::MockModule`
```
Can't locate object method "redefine" via package "Test::MockModule" at ./22-svirt.t line 112.
```
when I grep, this is the first time the method was used.
On my system "LEAP-15.1" I get this output:
```
> perl -MTest::MockModule -e 'print $Test::MockModule::VERSION .$/;'
0.13
```

Martchus:
@cfconrad I've answered in Rocket Chat. But where do you see the `Test::MockModule` error?

okurz:
using "redefine" is a good idea but can you try with "mock" instead?

cfconrad:
> @cfconrad I've answered in Rocket Chat. But where do you see the `Test::MockModule` error?

It is in the travis log at the end. As I now re triggered, I'm unable to point you ...

> using "redefine" is a good idea but can you try with "mock" instead?

Yes, after I got the version, I will use mock(). Thx.


If I'm not wrong, our travis container is `Ubuntu 14.04.5 LTS` which is quite old, maybe we should use LEAP 15.1 or so?

cfconrad:
Ok `leap15.1` isn't available on travis. And also xenial would show the same behavior:
```
xenial (16.04LTS) (perl): Perl module to override subroutines in a module for unit testing [universe]
0.11-1: all 
```
going to use `mock()`

cfconrad:
Just for the record, as I didn't found a package list for trusty:
```
Test::MockModule::VERSION 0.11
```

Martchus:
@cfconrad @okurz has already created a PR for that: https://github.com/os-autoinst/openQA/pull/2497

Note that is previously used Leap 42.3 and not Ubuntu. (At least the Docker container uses that and that's what matters.)

cfconrad:
Ok, I think I got it now, that we already using a docker image defined by https://github.com/os-autoinst/openQA/blob/master/docker/travis_test/Dockerfile . 

Thx for pointing me!

backend/baseclass.pm
```diff
@@ -1349,6 +1349,13 @@ sub stop_ssh_serial {
     return $self->{serial} = undef;
 }
 
+sub hide_password
+{
```
Martchus: For consistency, move the `{` to the line above.

backend/baseclass.pm
```diff
@@ -1349,6 +1349,13 @@ sub stop_ssh_serial {
...
+    my ($self, %args) = @_;
+    $args{password} = 'SECRET' if ($args{password});
+    return %args;
```
Martchus: So now we copy `%args` not only once when passing them to `log_call` but also when 2 more times when passing/returning to/from `hide_password`. That doesn't seem efficient. (Not sure whether copy elision is a thing in Perl.)

backend/baseclass.pm
cfconrad: sure thing, thx. Funny that tiny doesn't complain.

backend/baseclass.pm
cfconrad: What do you propose? I see the following:

- Don't log `%args` at all
- introduce `bmwqemu::log_call_no_secret($array_ref_with_secrets, %args)`
- Use hash ref in `hide_password()` to remove one copy
- copy in each function and pass this modified copy to `log_call()`
- Create argument string by my self (or with pp) and pass it to log_call

In general I  don't see it critical to copy a array with ~10 entries, if this isn't call in a loop. But maybe I'm wrong.

Martchus: I think we need at least one copy because `hide_password` should only affect what's being logged. So we can only get rid of one copy:

```
bmwqemu::log_call(%{$self->hide_password(%args)});
```

So at least `hide_password` would return a reference.

cfconrad:
From my side it is ready.

PR #1167, opened 2019-06-17
[WIP] Test::Warnings - fails even though I do not see warnings? e.g. 08-autotest.t

okurz:


okurz:
I don't know what happened here. I would have appreciated feedback but because there is non so far I will close this PR again to clean up.

PR #1169, opened 2019-06-21
Use default graphics on x86 to circumvent "cirrus" limitations

okurz:
"cirrus" was the default graphical adapter since 8 years (see e3ac9c27) or
longer, it is time to trust the qemu default option better now :)

This uses "std" on qemu >= 2.2 which we can assume to have on usual
systems nowadays. As we do not detect the qemu version we should simply
rely on the default on all versions.

This was already suggested e.g. in
https://bugzilla.suse.com/show_bug.cgi?id=1064041#c19

What about the alternatives?:

* *qxl*: Might provide better performance but is also more recent.
Still, we have problems with it in drivers, e.g. see
https://bugzilla.opensuse.org/show_bug.cgi?id=1057241 just recently
fixed and only for more recent products
* *virtio*: Demands a more recent qemu without more benefit for now

See https://progress.opensuse.org/issues/53339#note-2 for verification.

Further references:
* https://wiki.archlinux.org/index.php/QEMU#Graphic_card

Related progress issue: https://progress.opensuse.org/issues/53339

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 15;
+our $version = 16;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -637,10 +637,6 @@ sub start_qemu {
         sp('g', '1024x768');
         $use_usb_kbd = 1;
     }
-    else {
-        $vars->{QEMUVGA} ||= "cirrus";
-    }
-
     sp('vga', $vars->{QEMUVGA}) if $vars->{QEMUVGA};
 
     my @nicmac;


okurz:
Setting [WIP] as I will test the "std" graphics adapter explicitly for a subset of the Tumbleweed snapshot tests, see https://progress.opensuse.org/issues/53339#note-8

okurz:
Ready to be reviewed and merged. This should be safe for openSUSE and SLE as I set the graphics adapter "cirrus" explicitly now on all according machines on both o3 and osd after crosschecking that in general "std" seems to provide better results than "cirrus" and actually fixes some problems. After merge and deployment I will carefully test again and remove the override where possible.

@AdamWill any concerns regarding this PR?

Vogtinator:
+9001 from me, even though this PR is at least three years late. libvirt doesn't even offer the cirrus option anymore and upstream kwin does not support it either.

coolo:
damn. now we have a EVOTEOVERFLOW

okurz:
One european night has passed, I guess that should have given @AdamWill enough time as well as anyone else, so can we approve/merge?

AdamWill:
Hey, check your holiday calendars: yesterday was Canada Day, eh ;)

but no, no objections, in fact I've been setting it to std in our x86_64 machine definitions for years. FWIW, our virt people would like us to be using virtio, but I think last time I tried that I ran into https://bugzilla.redhat.com/show_bug.cgi?id=1403365 .

aaannz:
I have a problem with this change and using virtio-serial console. Unless I specify to use cirrus again I am unable to switch to virtio-serial-console during SLE15SP1 install. Did anyone observe similar behaviour? Base OS on which worker is running is also SLE15SP1 so QEMU version is 3.1.0.

PR #1170, opened 2019-06-27
Save RPM spec file in upstream repo same as for openQA

okurz:
https://build.opensuse.org/package/show/home:okurz:branches:devel:openQA:spec/os-autoinst
is the according branched package with corresponding changes for
submission.


okurz:
@coolo your opinion would be appreciated

coolo:
I don't mind - it has the usual pros and cons. but doing it differently for os-autoinst and openQA doesn't make sense, so +1 from me.

PR #1171, opened 2019-06-27
spec: Provide a "devel" subpackage pulling in all build+test dependencies

okurz:
This "devel" package can be used instead of explicit repetition in our
containers.

Tested in the following way:
* Building package locally using the spec file
* Checking runtime dependencies with `rpm -qpR $rpm_file`
* Building a container with Dockerfile:

```
FROM registry.opensuse.org/opensuse/leap:42.3
ARG minor=42.3
RUN zypper -n ar -f http://download.opensuse.org/repositories/devel:/openQA/openSUSE_Leap_${minor}/devel:openQA.repo && \
    zypper -n ar -f http://download.opensuse.org/repositories/devel:/openQA:/Leap:/${minor}/openSUSE_Leap_${minor}/devel:openQA:Leap:${minor}.repo && \
    zypper -n --no-gpg-checks ref
COPY os-autoinst-devel*.rpm /tmp/
RUN zypper --no-refresh -n --no-gpg-checks in /tmp/*.rpm
```

* Tagging as "os-autoinst/dev:latest" locally
* Running that container with

```
docker run -it --rm -v $(pwd):/opt/repo -v $(pwd)/docker/dev/os-autoinst-devel*.rpm:/tmp/os-autoinst-devel.x86_64.rpm os-autoinst/dev:latest sh -c 'cd /opt/repo && ./autogen && make && make check'
```

to install the package delta – if any – and check if we have all
dependencies for build.

The intention is to simplify
https://github.com/os-autoinst/openQA/blob/master/docker/travis_test/Dockerfile#L10
by not explicitly repeat the build dependencies but rely on the
dependencies generated by the spec file that already use it.

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -24,19 +24,8 @@ License:        GPL-2.0-or-later
 Group:          Development/Tools/Other
 Url:            https://github.com/os-autoinst/os-autoinst
 Source0:        %{name}-%{version}.tar.xz
-BuildRequires:  autoconf
-BuildRequires:  automake
-BuildRequires:  gcc-c++
-BuildRequires:  libtool
-BuildRequires:  opencv-devel > 3.0
-BuildRequires:  pkg-config
-BuildRequires:  perl(Module::CPANfile)
-BuildRequires:  perl(Perl::Tidy)
-BuildRequires:  perl(Test::Compile)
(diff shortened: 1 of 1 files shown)


okurz:
@Martchus for the test please see the description

@coolo your feedback would be appreciated

okurz:
no further feedback, have approval, will merge.

PR #1172, opened 2019-07-02
os-autoinst.spec: Fix vanished runtime requires

okurz:
Introduced in c168894d


okurz:
To fix the missing dependency in https://build.opensuse.org/package/show/devel:openQA/os-autoinst , see the results from os-autoinst-test. Sorry about that, I also don't know how that single line vanished :D

PR #1173, opened 2019-07-02
Save virsh output for later handling

SergioAtSUSE:
Based on the generated file, an automated decision can be made to restart libvirt.

- Related ticket: https://progress.opensuse.org/issues/45326
- Verification run: http://openqa.slindomansilla-vm.qa.suse.de/tests/1542

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -571,7 +571,7 @@ __END"
         $self->get_cmd_output('echo bios.bootDelay = \"10000\" >> /vmfs/volumes/datastore1/openQA/' . $self->name . '.vmx', {domain => 'sshVMwareServer'});
     }
 
-    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name);
+    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name . ' 2> >(tee /tmp/os-autoinst-' . $self->name . '-stderr.log >&2)');
     bmwqemu::diag("Dump actually used libvirt configuration file " . ($ret ? "(broken)" : "(working)"));
     $self->run_cmd("virsh $remote_vmm dumpxml " . $self->name);
     die "virsh start failed" if $ret;


SergioAtSUSE:
@coolgw, @coolo, @foursixnine, what do you think about this?

coolgw:
> @coolgw, @coolo, @foursixnine, what do you think about this?

I agree and please implement it, thanks a lot.
@coolo @foursixnine Could you help give comments on this? We need a try for this workround, this impact 390 migration test a lot. Thanks.

consoles/sshVirtsh.pm
```diff
@@ -571,7 +571,7 @@ __END"
         $self->get_cmd_output('echo bios.bootDelay = \"10000\" >> /vmfs/volumes/datastore1/openQA/' . $self->name . '.vmx', {domain => 'sshVMwareServer'});
     }
 
-    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name);
+    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name . ' 2> >(tee /tmp/' . $self->name '.stderr.log >&2)');
```
foursixnine: As discussed

```suggestion
    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name . ' 2> >(tee /tmp/os-autoinst-' . $self->name . '-stderr.log >&2)');
```

cfconrad:
Just an idea. Why not creating a `record_info()` and use this for that automated decision?

SergioAtSUSE:
@cfconrad, the idea of this change is to use it for an automated decision. But, there are some people agains the automatic was. So, first we need something where we can take manual or automatic decision.

Your proposal will be more complex, since record info will appear on the webui, where it will appear on a front-end for humans. And then you need ot match the job to the worker where is was run, then look for the /etc/openqa/workers.ini to look for the data of the remote SUT's host, where the libvirt is malfunctioning.

My change will save the file directly on the SUT's host, then, we could make a script that looks on each KVM host for that file and reboot if error is found.

cfconrad:
Ah you wanna restart libvirt :) I thought you are looking for a way to restart the job.
Ok got the idea, thx. 
I think it could work. 
I don't know if `/tmp` is the right place for this, but don't have a better one, as `/var/lib/openqa/*` doesn't exists on these hosts. And I would add `tee -a` so you get all failures and a quantity.

SergioAtSUSE:
@cfconrad, I don't want to count failures, I just want to know the last one. So, it is done without '-a' in purpose. If a machine is already failing allocating memory, any job will fail. There is no need filling a file with the same message, and there is no need to fill the file for any working job.

Also /tmp is chosen in purpose. If the machine gets rebooted, there is no need to restart libvirt, so the file may need to be also removed, so we don't restart a working libvirtd.

coolgw:
@SergioAtSUSE @foursixnine  why this still Draft mode? when this can be implement and can be try it on OSD? Is there any block part?

foursixnine:
> @SergioAtSUSE @foursixnine why this still Draft mode? when this can be implement and can be try it on OSD? Is there any block part?

@SergioAtSUSE was struggling a bit with having a working svirt backend-capable host. It's fixed now, but he's on Holidays for the time being, I guess I could take over at some future point.

coolgw:
> > @SergioAtSUSE @foursixnine why this still Draft mode? when this can be implement and can be try it on OSD? Is there any block part?
> 
> @SergioAtSUSE was struggling a bit with having a working svirt backend-capable host. It's fixed now, but he's on Holidays for the time being, I guess I could take over at some future point.

Thanks all for your quick feedback and effort, very glad to hear the progress of this ticket.

SergioAtSUSE:
@foursixnine, changes made as suggested.
Waiting for verification run.

foursixnine:
@SergioAtSUSE Remember to make it ready for review :)

consoles/sshVirtsh.pm
```diff
@@ -571,7 +571,7 @@ __END"
         $self->get_cmd_output('echo bios.bootDelay = \"10000\" >> /vmfs/volumes/datastore1/openQA/' . $self->name . '.vmx', {domain => 'sshVMwareServer'});
     }
 
-    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name);
+    $ret = $self->run_cmd("virsh $remote_vmm start " . $self->name . ' 2> >(tee /tmp/os-autoinst-' . $self->name . '-stderr.log >&2)');
```
okurz: you are saving logs to /tmp/ which is not controlled by the worker so potentially log files could pile up but you write to the same file per worker instance, should be ok.

The redirect looks a bit unusual. Is there a good reason to use this and not e.g. `… 2>&1 | tee …`?

SergioAtSUSE: > you are saving logs to /tmp/ which is not controlled by the worker so potentially log files could pile up but you write to the same file per worker instance, should be ok.

The idea is that the file is used a flag "need to restart libvirt". And a restart of the machine could also cleanup /tmp/.


> The redirect looks a bit unusual. Is there a good reason to use this and not e.g. `… 2>&1 | tee …`?

Yes, I want that the STDERR is still sent as STDERR, with your proposal, it will go to STDOUT. With my redirect, someone that executes isotovideo manually can still benefit from filtering STDERR only.

SergioAtSUSE: ...and I only want STDERR on that file, not STDOUT

SergioAtSUSE: ...and only for the last executed command. If it fails, it will still until libvirt is restarted.
and, if libvirt/machine is restarted, the next command will empty that file, to the flag should be interpreted as "KVM memory allocation ok"

okurz: > > you are saving logs to /tmp/ which is not controlled by the worker so potentially log files could pile up but you write to the same file per worker instance, should be ok.
> 
> The idea is that the file is used a flag "need to restart libvirt". And a restart of the machine could also cleanup /tmp/.

But when you restart the libvirt service this won't magically restart a machine or cleanup /tmp, what do you mean by that?

SergioAtSUSE: no, but the operator can,

- restart libvirt and empty/remove the file
- restart libvirt and restart the failing job (it will empty the file)
- restart the machine (and if tmp is configure to be empty, ok, if not, manual remove or manual job restart)

SergioAtSUSE: But, this is only the first step, my long-term goal is to automate it, so libvirt is automatically restarted (with proper cleanup) when no isotovideo is running (not possible to break a running job), etc.

okurz: sure, makes sense

PR #1174, opened 2019-07-05
VNC: reverse key order when sending key up events

AdamWill:
When map_and_send_key maps an input to multiple key presses -
e.g. a ':' requires a press of shift and a press of ';', or
when dealing with any input like "ctrl-t" - it gets an array,
sends a 'down' event for each key in the array, then sends an
'up' event for each key in the array. So for our ctrl-t example
it does this:

ctrl down
t down
ctrl up
t up

This is a bit odd and not how humans normally type. Humans
would normally hold the modifier key, press and release any
modified keys they want to press, then release the modifier.
I've run into one issue previously where this caused a problem,
which was ultimately fixed in qemu so that os-autoinst's unusual
key event order worked OK. Now I've run into another:

https://bugzilla.redhat.com/show_bug.cgi?id=1727388

All the details of that bug are still not worked out, but today
I *did* work out that changing this key press order avoids that
bug. By simply reversing the array for the key up events, so we
do this instead:

ctrl down
t down
t up
ctrl up

The bug stops happening. Since it's triggered two bugs now, and
as I said it does not map to how humans type, I think we should
really go ahead and change this.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -759,7 +759,7 @@ sub map_and_send_key {
     }
     usleep(2_000);
     if (!defined $down_flag || $down_flag == 0) {
-        for my $key (@events) {
+        for my $key (reverse @events) {
             $self->send_key_event_up($key);
         }
     }


consoles/VNC.pm
```diff
@@ -759,7 +759,7 @@ sub map_and_send_key {
     }
     usleep(2_000);
     if (!defined $down_flag || $down_flag == 0) {
-        for my $key (@events) {
+        for my $key (reverse @events) {
```
okurz: I guess a comment here would really be a good idea explaining the delicate impact of "reverse" :)

okurz:
Would be great if this also fixes other related problems like https://progress.opensuse.org/issues/46190

consoles/VNC.pm
coolo: git blame will help you if you're super curious

coolo:
> Would be great if this also fixes other related problems like https://progress.opensuse.org/issues/46190

I don't think so. What we see there are key events happening out of order. This is what I assume happens here as well - but possibly the meta keys up events serve as sync point somewhere?

Which makes one believe we can force keys to stay in order by pressing ctrl briefly :)

AdamWill:
yeah...if that consistently fails because it types characters that should be upper-case as lower-case, it *might* be something like this. But I'm not making any promises :P

consoles/VNC.pm
AdamWill: ...until 'tidy' hits the line, or something...

so yeah, I'm with @okurz (in fact I meant to add a comment originally but forgot, so thanks for the reminder!)

okurz:
@AdamWill do you think https://openqa.opensuse.org/tests/977825#step/ooffice/6 could be caused by this? I guess we we saw something similar in before, especially in libreoffice but WDYT?

bmwiedemann:
"Hello Worl!d" - the event sequence should have been
```
l down
l up
d down
d up
shift down
1 down
1 up
shift up
```

it would be strange if a reverse of the last two would cause such a change, but you never know.

AdamWill:
yeah, I agree with @bmwiedemann here, I'd be honestly surprised if this could cause that problem.

PR #1175, opened 2019-07-16
spec: Also require qemu-tools in -devel as needed by openQA

okurz:


okurz:
yes, you are right. Now I am confused by my own commit message. Maybe I meant something different. I will close this for now

PR #1176, opened 2019-07-17
sshVirtshSUT use ssh aware serial screen

cfconrad:
The method of using NET::SSH2->sock() to do the select and then read
on the NET::SSH2::Channel wasn't reliable.
It got stock from time to time: https://progress.opensuse.org/issues/54260

This approach is to use the polling approach of a none blocking socket.

Make the ssh socked none blocking. <- The code of serial_terminal actually expect it, but somehow it wasn't. This is already fixed in https://github.com/os-autoinst/os-autoinst/pull/1166

Progress Ticket: https://progress.opensuse.org/issues/54260

### DIFF
--- a/consoles/serial_screen.pm
+++ b/consoles/serial_screen.pm
@@ -26,10 +26,9 @@ use Carp 'croak';
 our $VERSION;
 
 sub new {
-    my ($class, $socket_fd, $select_fd) = @_;
+    my ($class, $socket_fd) = @_;
     my $self = bless {class => $class}, $class;
     $self->{socket_fd}    = $socket_fd;
-    $self->{select_fd}    = $select_fd;
     $self->{carry_buffer} = '';
     return $self;
 }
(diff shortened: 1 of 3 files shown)


cfconrad:
Consider this as working fix:
VR:
 * http://cfconrad-vm.qa.suse.de/tests/5060  [5061](http://cfconrad-vm.qa.suse.de/tests/5061) [5063](http://cfconrad-vm.qa.suse.de/tests/5063) [5064](http://cfconrad-vm.qa.suse.de/tests/5064) [5065](http://cfconrad-vm.qa.suse.de/tests/5065) [5077](http://cfconrad-vm.qa.suse.de/tests/5077) [5078](http://cfconrad-vm.qa.suse.de/tests/5078)

But I still get some failing tests([5070](http://cfconrad-vm.qa.suse.de/tests/5070) ). But it looks like an other issue on openqa or bug in system. There the LTP command just "hang" and we run in a "real" timeout, see http://cfconrad-vm.qa.suse.de/tests/5070#step/cgroup_fj_stress_freezer_2_2_each/10. Where I send SIGTERM to get the prompt again.

consoles/serial_screen.pm
```diff
@@ -152,6 +151,45 @@ sub normalise_pattern {
...
+Returns number of bytes read or undef on timeout. Note that 0 is a valid return code.
+If an failure occur the method croak.
+
+A undefined timeout will wait infinitive. A timeout of 0 just try to read once.
+=cut
+sub do_read
+{
+    my ($self, undef, %args) = @_;
+    my $buffer = '';
+    $args{timeout}  //= undef;    # wait till data is available
```
asmorodskyi: looks strange , it will be undef anyway . what the purpose of this line ?

consoles/serial_screen.pm
```diff
@@ -152,6 +151,45 @@ sub normalise_pattern {
...
+Attempts to read up to max_size bytes from the C<<$self->{socket_fd}>> into buffer.
+The method returns as soon as some data is available, even if the given size has not been reached.
+Returns number of bytes read or undef on timeout. Note that 0 is a valid return code.
+If an failure occur the method croak.
+
+A undefined timeout will wait infinitive. A timeout of 0 just try to read once.
+=cut
+sub do_read
+{
+    my ($self, undef, %args) = @_;
```
asmorodskyi: why you need this `undef` ?

consoles/serial_screen.pm
cfconrad: Yeah true, I added it to make the reader of that function aware that timeout is a valid parameter.
And to add the comment, that `undef` is treated as infinite waiting.

consoles/serial_screen.pm
cfconrad: The second parameter is used in "call by reference" mode. Like other implementations [sysread](https://perldoc.perl.org/functions/sysread.html) or [Net::SSH2::Channel::read](https://metacpan.org/pod/Net::SSH2::Channel#read-(-buffer,-max_size-%5B,-ext-%5D-)) do.
In the line 189 it looks like ` $_[1] = $buffer;` which does the trick. 
And I thought this `undef` makes the reader aware of, that here is something special with the second parameter.

okurz:
as you pointed this PR out to me in person … :)

coolo:
New file was forgotten in the Makefile - o3 is broken now

PR #1177, opened 2019-07-23
Fix missing 'consoles/serial_screen.pm' in Makefile.am

okurz:
See https://github.com/os-autoinst/os-autoinst/pull/1176#issuecomment-514074158


okurz:
*critsit* … as coolo would say

okurz:
wait, that was wrong :( serial_screen is now mentioned twice

PR #1178, opened 2019-07-23
Port away from the deprecated OpenCV 1 C API

Martchus:
These changes would be required to compile the video encoder against OpenCV 4. <s>Not sure about the pkg-config change because the pkg-config file is currently missing in the Tumbleweed package.</s> It works with the [opencv package from my home repo](https://build.opensuse.org/package/show/home:mkittler:branches:science/opencv) which I'm going to submit to Tumbleweed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,7 +31,20 @@ PERL_CFLAGS=`perl -MExtUtils::Embed -e ccopts`
 AC_MSG_RESULT([$PERL_CFLAGS])
 AC_SUBST([PERL_CFLAGS])
 
-PKG_CHECK_MODULES([OPENCV], [opencv >= 2.4])
+# ================================
+# Find dependencies via pkg-config
+# ================================
+# Use preferably the "opencv4", otherwise fall back to "opencv"
+PKG_CHECK_MODULES(
+	[OPENCV],
+	[opencv4 >= 4.0],
+	[AC_DEFINE([HAVE_OPENCV_4], [1], [Using opencv4 package])],
+	[PKG_CHECK_MODULES(
+		[OPENCV],
+		[opencv >= 2.4],
+		[AC_DEFINE([HAVE_OPENCV], [1], [Using opencv package])]
+	)]
+)
 PKG_CHECK_MODULES([THEORAENC], [theoraenc >= 1.1])
 PKG_CHECK_MODULES([FFTW], [fftw3])
 PKG_CHECK_MODULES([SNDFILE], [sndfile])
--- a/debugviewer/debugviewer.cpp
+++ b/debugviewer/debugviewer.cpp
@@ -18,7 +18,7 @@ int main(int argc, char** argv)
 
     Mat image(768, 1024, CV_8UC3, Scalar(0, 0, 0));
     while (1) {
-        Mat last = imread(argv[1], CV_LOAD_IMAGE_COLOR); // Read the file
+        Mat last = imread(argv[1], cv::IMREAD_COLOR); // Read the file
         if (last.data)
             image = last;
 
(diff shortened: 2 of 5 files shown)


StefanBruens:
Actually, this is a port to OpenCV 2, the API you are replacing is from OpenCV 1.x. It has been deprecated for long and removed with OpenCV 4.0.

Martchus:
Apparently the name "OpenCV" has 2 meanings: The "OpenCV" library/project and the "OpenCV" API. It seems that both use different versioning. Just considering the files contained by the openSUSE `opencv-devel` package for library/project version 4.1.0 it seems the latest API is version 2. So you're only criticizing the commit message which should rather contain the API version? Because otherwise I don't know to which other version I should possibly port the code to.

I've also checked with `readelf -d videoencoder` and now it is definitely using library version 4.1.

SR for OpenCV pkg-config file: https://build.opensuse.org/request/show/717945

StefanBruens:
The C++ API is the one _introduced_ with OpenCV 2, and still used with OpenCV 4.x. So the commit message should better read "Port away from the deprecated OpenCV 1 C API", see https://github.com/opencv/opencv/wiki/ChangeLog#version400

Also, there is no need to require OpenCV 4, the same code also works with OpenCV 2.x and 3.x, most distributions do not have OpenCV 4 yet (this includes SLE 15SP1 and Leap 15.1).

Martchus:
@StefanBruens 

Ok, so this is really just about the commit message. I'll change it as you suggest it.

> Also, there is no need to require OpenCV 4

I know but the pkg-config file provided by OpenCV 4.x is now called `opencv4`. This is likely done to reflect breaking API changes. (There were breaking API changes. Those likely don't affect us but I remember my VLC builds were failing.)

Note that this is just a draft. I actually plan to extend the build script to work with multiple OpenCV versions (see https://github.com/os-autoinst/os-autoinst/pull/1179#issuecomment-514236515).

Before continuing here I'd like to fix the missing pkg-config file in opencv-devel. Since you're also the maintainer of that package you could give me some feedback here: https://build.opensuse.org/request/show/717945

Martchus:
By the way, I gave it a local test run and it actually works. (The CI obviously fails at this point.)

okurz:
So according to https://github.com/os-autoinst/os-autoinst/pull/1178#issuecomment-514600348 we should not actually need opencv. Maybe you can focus your PR on that, "porting away from old, deprecated API" so that we have future-proof code but still work with opencv3 for the time being. In a subsequent PR we can then experiment with opencv4 support – at best while keeping opencv3 working – but also changing the CI tests to cover this.

Martchus:
Of course the idea is to support OpenCV 3 and OpenCV 4 at the same time. This should not be a problem - we just need remove all usages of the C API (already covered by this PR) and handle the renaming of the pkg-config file so OpenCV 4 is considered if available (this part needs amendment).

Martchus:
Note that I didn't continue because @StefanBruens did not accept my SR or gave me any feedback. But I can continue regardless of the availability of OpenCV 4 pkg-config support in Tumbleweed.

Martchus:
This should now work with both pkg-config packages (`opencv4` and `opencv`). Let's see whether the CI check passes.

Martchus:
I have tested whether the package still builds and it seems to be the case. So the log looks something like that:

```
[   89s] checking for opencv4 >= 4.0... no
[   89s] checking for opencv >= 2.4... yes
```

(Of course I also tested both cases locally.)

okurz:
would you like to try branch os-autoinst into an OBS project based on your git branch?

Martchus:
That is basically already how I tested it: https://build.opensuse.org/package/show/home:mkittler:branches:devel:openQA/os-autoinst

okurz:
Nice. https://build.opensuse.org/package/live_build_log/home:mkittler:branches:devel:openQA/os-autoinst/SLE_15/aarch64 shows a failure in osutils.t but this is very unlikely to be related to opencv

PR #1179, opened 2019-07-23
Fix build after opencv4 submission

okurz:
* spec: Force OBS to resolve choices on opencv-devel
* spec: Fix build for openSUSE Factory and backports after opencv4 submission

Needs the package opencv3 submitted to our backport repos. I can do that after PR approval just before merge.

See https://build.opensuse.org/package/show/home:okurz:branches:devel:openQA:fix:os-autoinst_opencv2/os-autoinst for build results

Related progress issue: https://progress.opensuse.org/issues/54356

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -24,7 +24,7 @@ License:        GPL-2.0-or-later
 Group:          Development/Tools/Other
 Url:            https://github.com/os-autoinst/os-autoinst
 Source0:        %{name}-%{version}.tar.xz
-%define         build_requires autoconf automake gcc-c++ libtool opencv-devel > 3.0, pkg-config perl(Module::CPANfile) perl(Perl::Tidy) perl(Test::Compile) pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc) make
+%define         build_requires autoconf automake gcc-c++ libtool pkgconfig(opencv) pkg-config perl(Module::CPANfile) perl(Perl::Tidy) perl(Test::Compile) pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc) make
 BuildRequires:  %build_requires
 # just for the test suite
 BuildRequires:  qemu-tools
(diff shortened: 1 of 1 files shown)


dist/rpm/os-autoinst.spec
```diff
@@ -24,7 +24,8 @@ License:        GPL-2.0-or-later
 Group:          Development/Tools/Other
 Url:            https://github.com/os-autoinst/os-autoinst
 Source0:        %{name}-%{version}.tar.xz
-%define         build_requires autoconf automake gcc-c++ libtool opencv-devel > 3.0, pkg-config perl(Module::CPANfile) perl(Perl::Tidy) perl(Test::Compile) pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc) make
+#!BuildIgnore: opencv3-devel
```
Martchus: What is the purpose of this line?

okurz: "spec: Force OBS to resolve choices on opencv-devel"

Martchus:
Why does this need the package opencv3? Shouldn't `pkgconfig(opencv)` ensure that either `opencv` or `opencv3` or basically any package providing the pkg-config file would be sufficient?

okurz:
@Martchus sorry I don't understand your question. Yes, `pkgconfig(opencv)` should be what just works which is either supplied by our old package `opencv` or the new backwards-compatibility package `opencv3`. This is exactly what this fix is about. So?

Martchus:
I'm only wondering why the package `opencv3` then needs to be submitted to our backport repos.

okurz:
We need `opencv3` in the backport repos because we use links to fctry for the opencv package which is *now* the opencv4 version without a proper pkg-config file so we need opencv3 in the backports repos along with opencv4 or delete the link and rely on the old (outdated) version from the base projects if there at all or provide both in parallel to be future proof for your support of opencv4 which we do :)

dist/rpm/os-autoinst.spec
Martchus: I think this code which just looks like an out-commented line deserves a comment within the file. Nobody likes to dig in the Git history for this.

okurz: done, updated

Martchus:
Ah, now it makes sense.

I guess I'll extend my draft for OpenCV 4 to be an option beside OpenCV 3 so we can switch whenever we like. So far I have fixed the package for OpenCV 4 to include the pkg-config file. Unfortunately the pkg-config file looks buggy so I'll do some testing and will submit it to Tumbleweed if it works.

PR #1181, opened 2019-07-24
crop.py: tweak imports to work with Python 3

AdamWill:
In Python 3 it's tkinter, not Tkinter. This works both ways.


AdamWill:
With just this change, I can at least open, modify and save a needle with Python 3. I didn't check extensively for anything else that might need changing for full Python 3 compat.

PR #1182, opened 2019-07-29
Fix missing data while reading from virtio_console

cfconrad:
During some test, we figured out that for unknown reason we miss some
data.
E.g. `cat large_file_140kb.txt` sporadically have some output gap somewhere
in the middle of the output.

Interesting is, that the logfile, which is created from qemu, contains
all data.

I checked `ss -x -a` on host and `cat /sys/kernel/debug/virtio-ports/*`
inside the VM. But I didn't found any hint of discarded packages.

### DIFF
--- a/consoles/serial_screen.pm
+++ b/consoles/serial_screen.pm
@@ -26,10 +26,12 @@ use Carp 'croak';
 our $VERSION;
 
 sub new {
-    my ($class, $socket_fd) = @_;
+    my ($class, $fd_read, $fd_write) = @_;
     my $self = bless {class => $class}, $class;
-    $self->{socket_fd}    = $socket_fd;
+    $self->{fd_read}      = $fd_read;
+    $self->{fd_write}     = $fd_write // $fd_read;
     $self->{carry_buffer} = '';
+
     return $self;
 }
 
@@ -89,7 +91,7 @@ and ETX is the same as pressing Ctrl-C on a terminal.
 =cut
 sub type_string {
     my ($self, $nargs) = @_;
-    my $fd = $self->{socket_fd};
+    my $fd = $self->{fd_write};
 
     bmwqemu::log_call(%$nargs);
 
@@ -155,7 +157,7 @@ sub normalise_pattern {
 
   my $num_read = do_read($buffer [, max_size => 2048][,timeout => undef]);
 
-Attempts to read up to max_size bytes from the C<<$self->{socket_fd}>> into buffer.
+Attempts to read up to max_size bytes from the C<<$self->{fd_read}>> into buffer.
 The method returns as soon as some data is available, even if the given size has not been reached.
 Returns number of bytes read or undef on timeout. Note that 0 is a valid return code.
 If an failure occur the method croak.
@@ -168,7 +170,7 @@ sub do_read
     my $buffer = '';
     $args{timeout}  //= undef;    # wait till data is available
     $args{max_size} //= 2048;
-    my $fd = $self->{socket_fd};
+    my $fd = $self->{fd_read};
 
     my $rin = '';
     vec($rin, fileno($fd), 1) = 1;
@@ -223,7 +225,7 @@ and { matched => 0, string => 'text from the terminal' } on failure.
 =cut
 sub read_until {
     my ($self, $pattern, $timeout) = @_[0 .. 2];
-    my $fd       = $self->{socket_fd};
+    my $fd       = $self->{fd_read};
     my %nargs    = @_[3 .. $#_];
     my $buflen   = $nargs{buffer_size} || 4096;
     my $overflow = $nargs{record_output} ? '' : undef;
@@ -311,7 +313,7 @@ sub peak {
 
     bmwqemu::log_call(%nargs);
   LOOP: {
-        $read = sysread($self->{socket_fd}, $buf, $buflen);
+        $read = sysread($self->{fd_read}, $buf, $buflen);
         last LOOP unless defined $read;
 
         $self->{carry_buffer} .= $buf;
--- a/consoles/virtio_terminal.pm
+++ b/consoles/virtio_terminal.pm
@@ -21,6 +21,7 @@ use autodie;
 
 use base 'consoles::console';
 
+use Mojo::File 'path';
 use Socket qw(SOCK_NONBLOCK PF_UNIX SOCK_STREAM sockaddr_un);
 use Errno qw(EAGAIN EWOULDBLOCK);
 use English -no_match_vars;
@@ -28,7 +29,8 @@ use Carp 'croak';
 use Scalar::Util 'blessed';
 use Cwd;
 use consoles::serial_screen ();
-use testapi 'check_var';
+use testapi qw(check_var get_var);
+use Fcntl;
 
 our $VERSION;
 
@@ -48,8 +50,8 @@ where it can start a tty on the virtual console. By default openSUSE and SLE
 automatically start agetty when the kernel finds the virtio console device, but
 another OS may require some additional configuration.
 
-It may also be possible to use a transport other than virtio. This code just
-requires a UNIX socket which inputs and outputs terminal ASCII/ANSI codes.
+It may also be possible to use a transport other than virtio. This code
+uses two pipes to communicate with virtio_consoles from qemu.
 
 =head1 SUBROUTINES/METHODS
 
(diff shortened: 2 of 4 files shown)


consoles/virtio_terminal.pm
```diff
@@ -59,9 +60,15 @@ sub new {
     my ($class, $testapi_console, $args) = @_;
     my $self = $class->SUPER::new($testapi_console, $args);
     $self->{socket_fd}      = 0;
-    $self->{socket_path}    = $self->{args}->{socked_path} // cwd() . '/virtio_console';
+    $self->{console_num}    = $self->{args}->{console_num} // 0;
     $self->{snapshots}      = {};
     $self->{preload_buffer} = '';
+
+    # W/A for backward compatibility
+    if (defined($self->{socked_path})) {
```
cfconrad: Gets obsolete with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8039

richiejp:
You could maybe just try increasing the buffer sizes on our end. Most buffers are 1Kb to 8Kb and you are trying to send 128Kb from a single command. QEMU might send all this in a single call to write depending on how quickly the SUT can get it out. Then we are just taking a few Kb at a time in perl and running regexs over it.

If changing to TCP stops the problem, then that is probably because you are slowing down the packet transmission and adding a receive buffer in the kernel. A Unix socket stream should be totally reliable. Maybe it would drop a packet if the send buffer is full... you can see the send buffer size at `/proc/sys/net/core/wmem_default` (see `man 7 socket`) and maybe try increasing it on a test machine. 

Possibly if QEMU gets EGAIN when writing it might drop some data instead of waiting which won't be reported by `ss`. Alternatively it could be the buffer handling on our end which won't be fully solved by moving to TCP either.

cfconrad:
@richiejp thx for your hints.

When I increase net.core.wmem_default, I don't face the problem anymore. But I'm not sure if this is a valid solution for us. As this has impact on all sockets.

richiejp:
I suppose we don't read fast enough then. You could try increasing the buffer size for `read_until` and `peak` as well (in the Perl code). Also maybe the ring buffer size.

cfconrad:
I increased the buffer size to 400k but the issue still exists. Also it never hits the limit.
http://cfconrad-vm.qa.suse.de/tests/5529/file/autoinst-log.txt look for "CLEMIX read"

richiejp:
So the problem is that we do not read the socket before QEMU fills the send buffer in the kernel.

We could create a separate process to read the socket faster or write to a file like the emulated serial port. However the first option is a lot of work and possibly just moves the problem instead of fixing it. The second is likely to make performance even worse which is the cause of this problem in the first place. If you switch to TCP, this will most likely solve the problem for `n` bytes, but not `n+m` where `m` is relatively small.

QEMU has a chardev type called `ringbuf`, I suspect this would solve the problem, but I'm not sure how to use it or if it would be practical to use it from Perl.

cfconrad:
I think this `ringbuf` is unidirectional. 
The `pipe` option looks to me like something we could give a try. WDYT?

Regarding a separate process to read the data from. Maybe we could just use `socat`. But I would like to avoid this.

cfconrad:
Thinking twice on it, if we don't increase the buffer, the only solution is to have a separate process for reading().
As we inject the command and then reading for output has always a gap in between.

richiejp:
Take a look at `man 7 pipe`. The max capacity appears to be much larger than for a socket, but the default capacity is only 65Kb, it might be possible to increase it from our side.

Looking at QEMU's src, it appears that the virtio console (and all the others AFAICT) do not wait if it gets EAGAIN during a write and there is no way to change that.

cfconrad:
Ok, I will give `pipe` a try.

cfconrad:
This is how it looks in action: http://cfconrad-vm.qa.suse.de/tests/5577 I also added `always_rollback => 1` and a `power_action('reboot');` to that test.

My current problem is that travis complains about `Fcntl::F_GETPIPE_SZ` looks like `Ubuntu 14.04.5 LTS` doens't have that.

Should I make it static or something like this:
```
sub F_GETPIPE_SZ
{
    return eval {no warnings "all"; Fcntl::F_GETPIPE_SZ;} || 1032;
}
```

consoles/virtio_terminal.pm
```diff
@@ -110,33 +108,38 @@ sub load_snapshot {
...
-Opens a socket to the host end of the virtio_console. This need to be
-corresponding to the console which was created by qemu-backend.
+Opens a the read and write pipe based on C<$pipe_prefix>.
 
-Returns the file descriptor for the open socket, otherwise it dies.
+Returns the read and write file descriptors for the open sockets,
+otherwise it dies.
 
 =cut
 sub open_socket {
```
richiejp: This should be changed to pipe. Maybe in the future we will allow connecting to a remote QEMU instance.

consoles/virtio_terminal.pm
```diff
@@ -110,33 +108,38 @@ sub load_snapshot {
...
-    $fd   = IO::Socket::INET->new(PeerAddr => '127.0.0.1', PeerPort => $port, Proto => 'tcp')
-      or die('Connection to virtio_terminal nr ' . $self->{console_num} . ' failed on port ' . $port);
+    bmwqemu::log_call(pipe_prefix => $self->{pipe_prefix});
+
+    sysopen(my $fd_w, $self->{pipe_prefix} . '.in', O_WRONLY)
+      or die "Can't open in pipe for writing $!";
+    sysopen(my $fd_r, $self->{pipe_prefix} . '.out', O_NONBLOCK | O_RDONLY)
+      or die "Can't open out pipe for reading $!";
+
+    my $newsize = get_var('VIRTIO_CONSOLE_PIPE_SZ', 1024 * 1024);
```
richiejp: You could read the max pipe size from the `/proc/sys/fs/pipe-max-size` and use that. On kernel's after 4.9 it might be set to that by default anyway. On older kernels or systems with unusual page sizes the limit might not be 1MB either. I don't know when a user would need to set a pipe size that is not the max for their system; they would need to have some huge max size for it to make sense.

You might already know this, but you may need to use `sysopen` and `sysread` on this file then parse the resulting string as an integer. The normal Perl functions might not work because these files don't implement `size` and stuff like that.

richiejp:
> ```
> sub F_GETPIPE_SZ
> {
>     return eval {no warnings "all"; Fcntl::F_GETPIPE_SZ;} || 1032;
> }
> ```

This is good, making it static is risky because it might have a different value on some architectures.

consoles/virtio_terminal.pm
cfconrad: Good idea, I will read out the default `$newsize` from that file.
Thx for the hints.

cfconrad: I'm not sure if I get the point regarding `sysopen/sysread`. I just tried a 
```
perl -e 'use Mojo::File "path"; print path("/proc/sys/fs/pipe-max-size")->slurp();'
```
and it worked for me.

cfconrad:
I consider this as READY.

I run a test and increasing the file size in step 1k steps:

**UNIX sockets:** 
 * fail with size of 11264 http://cfconrad-vm.qa.suse.de/tests/5594#step/virtio_console_long_output/1006
 * And then got weird problems with file size of 12288 http://cfconrad-vm.qa.suse.de/tests/5596#step/virtio_console_long_output/297

**FIFO PIPE**
  * Run good with file size 222208
  * Got weird problems with file size 223232 http://cfconrad-vm.qa.suse.de/tests/5581#step/virtio_console_long_output/357

So still it's not perfect, but I think we stabilized it with this patch.

consoles/serial_screen.pm
```diff
@@ -26,10 +26,16 @@ use Carp 'croak';
...
+    my ($class, $fd_read, $fd_write) = @_;
     my $self = bless {class => $class}, $class;
-    $self->{socket_fd}    = $socket_fd;
+    $self->{fd_read}      = $fd_read;
+    $self->{fd_write}     = $fd_write // $fd_read;
     $self->{carry_buffer} = '';
+
+    my $old_fh = select($fd_write);
+    $| = 1;
+    select($old_fh);
```
richiejp: I don't think we need this?

cfconrad: Your right from perldoc syswrite ` It bypasses buffered IO, so mixing this with reads (other than sysread))`

Going to remove it.

richiejp:
LGTM

cfconrad:
@Martchus @okurz WDYT?

richiejp:
@okurz The commit does contain this info?

cfconrad:
Ok I guess it was the title. I changed it to `Fix missing data while reading from virtio_console`.

okurz:
thx

okurz:
There is one benefit of having a ticket upfront: You could monitor the consequences of changes more easily :)

See https://progress.opensuse.org/issues/55412 . I guess we need to adjust the apparmor profiles in openQA. Do you think it's enough to whitelist `/proc/sys/fs/pipe-max-size` ?

PR #1184, opened 2019-08-01
Fix the readout of git hash as version when called from other directory

okurz:


okurz:
@foursixnine ?

foursixnine:
I just wonder, what was the motivation for this... But good catch?

okurz:
uhm, doesn't the subject line explain this? When calling the isotovideo of a local git checkout but from another directory, e.g. when starting an openQA worker with the `--isotovideo` parameter, previously the git command would not be executed in the correct folder.

foursixnine:
> uhm, doesn't the subject line explain this? 

It explain what's being fixed.

> 

When calling the isotovideo of a local git checkout but from another directory, e.g. when starting an openQA worker with the --isotovideo parameter, previously the git command would not be executed in the correct folder.

This is how I run openQA locally, but I've never paid attention to the actual version reported, so that could be it.

PR #1187, opened 2019-08-01
qemu: Fix warning about undefined value in match

okurz:


okurz:
I doubt we can easily cover this diff in tests although the full stack test should use qemu. However for this simple value definition handling I guess we don't need to be too strict.

PR #1188, opened 2019-08-02
Add missing 'use' for 'Data::Dumper'

okurz:
Dumper is used in a die message hence must be imported in before.


AdamWill:
shouldn't the dep also be added to the rpm spec, then? AFAICS there isn't one.

okurz:
@AdamWill good point, added https://github.com/os-autoinst/os-autoinst/pull/1192 . I did not (yet) add a check for cpanfile against the spec file. don't know what would be the best approach for that.

PR #1190, opened 2019-08-02
Add automatic lookup for UEFI_PFLASH_CODE/VARS to fix UEFI on Tumbleweed machines

okurz:
This allows to simply set UEFI=1 and neither BIOS nor UEFI_PFLASH which will
lookup corresponding "code" and "vars" firmware files instead of failing.

The firmware files used by Fedora and Debian were already referencing "code"
files as single pflash targets so we do the same for the openSUSE package
which by now in openSUSE Factory does not supply the single code+vars image
anymore.

Currently os-autoinst-distri-opensuse sets UEFI_PFLASH which is still
supported from os-autoinst but will not work if the worker is a more recent
openSUSE Tumbleweed. Simply not specifying the variable will fix that problem
as soon as this commit is in.

Tested locally with a vars.json file in three variants, with UEFI_PFLASH
resolving the single code+vars file as well as without UEFI_PFLASH and no
explicit path to neither BIOS nor UEFI_FPLASH_CODE so that the firmware file
is automatically looked up. The third variant is to explicitly specify either
code or vars file which still works.

Related progress issue: https://progress.opensuse.org/issues/55052

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -498,6 +498,12 @@ sub do_extract_assets {
 
 # baseclass virt method overwrite end
 
+sub find_ovmf {
+    foreach my $firmware (@bmwqemu::ovmf_locations) {
+        return $firmware if -e $firmware;
+    }
+}
+
 sub start_qemu {
     my $self = shift;
     my $vars = \%bmwqemu::vars;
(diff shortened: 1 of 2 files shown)


okurz:
@richiejp ?

backend/qemu.pm
```diff
@@ -540,22 +546,16 @@ sub start_qemu {
...
-            # We know this won't go well.
-            die "No UEFI firmware can be found! Please specify BIOS or UEFI_BIOS or install an appropriate package";
-        }
+    if ($vars->{UEFI_PFLASH}) {
+        $vars->{BIOS} //= find_ovmf =~ s/-code//r;
+    }
+    elsif ($vars->{UEFI}) {
+        $vars->{UEFI_PFLASH_CODE} //= find_ovmf;
+        $vars->{UEFI_PFLASH_VARS} //= $vars->{UEFI_PFLASH_CODE} =~ s/code/vars/r;
+        die "No UEFI firmware can be found! Please specify UEFI_PFLASH_CODE/UEFI_PFLASH_VARS or BIOS or UEFI_BIOS or install an appropriate package" unless $vars->{UEFI_PFLASH_CODE};
```
b10n1k: "or install an appropriate package" is kinda vague. i think it can be confusing. Can you elaborate a bit?

bmwqemu.pm
```diff
@@ -64,8 +64,8 @@ our $direct_output;
 # the kraxel.org nightly packages, third is Fedora's edk2-ovmf package,
 # fourth is Debian's ovmf package.
 our @ovmf_locations = (
-    '/usr/share/qemu/ovmf-x86_64-ms.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd',
-    '/usr/share/edk2/ovmf/OVMF_CODE.fd',  '/usr/share/OVMF/OVMF_CODE.fd'
+    '/usr/share/qemu/ovmf-x86_64-ms-code.bin', '/usr/share/edk2.git/ovmf-x64/OVMF_CODE-pure-efi.fd',
```
richiejp: It is possible that this file is present on other architectures as x86 can be emulated.

backend/qemu.pm
okurz: it's the same message as in before, I have not really changed the content. As package names differ a lot among different distributions I consider it not feasible to suggest a specific package.

bmwqemu.pm
okurz: yes, as was the case in before. I did not change this and also I don't think of it as an imminent problem. Did you just want to mention this? Do you see a problem? Do you have a suggestion?

richiejp: This was only used on x86 before, at least in this part of the code. Because you now use it to set the pflash code variable in general it could be used in more scenarios. I would suggest using the arch variable to generate the file name somehow.

okurz: my bad. I did not remember over the weekend that I removed the "if x86_64" part. Will change that after crosschecking other architectures.

okurz: Same as before the according code is only called for x86_64 now as I figured for other architectures the firmware can not be easily discovered, e.g. aarch64 would be `aavmf-aarch64-ms-code.bin` so not a simple s/x86_64/aarch64/

richiejp:
LGTM

okurz:
I will look into tests in a separate PR because this would be something to discuss first.

AdamWill:
ten points to anyone who can remember the entire sequence of how we got to where we are right now with all of this stuff...:P

okurz:
@AdamWill uhhh, … do you mean I overlooked something? I tried to dig deep in the git logs to understand how we got there but I have not experienced the full history personally :)

DrMullings:
Because of this change and the test suits not being adapted stuff is broken on  O3
eg https://openqa.suse.de/tests/3221776
@foursixnine

okurz:
@DrMullings the job you mentioned has hardcoded code+vars path but also UEFI_PFLASH. I did not know that this worked even in before but it shouldn't. Anyway, when you rebase your test code on latest master the variable `UEFI_PFLASH` should not be set anymore. Then this should work. I don't think further adjustements in test suites are necessary

DrMullings:
@okurz  UEFI_PFLASH is being set in the test suite, we just checked that

AdamWill:
@okurz I can't think of something specific, but we have changed this thing so damn often it's hard to be sure everything anyone does is gonna work (see @DrMullings issue for e.g.). "Just set `UEFI` and the code will figure out the file locations" is where we *started*, IIRC, waaaay back in time, then you could set `BIOS` to point to a single firmware file, then I think `UEFI_PFLASH` alone showed up and meant...something?...then `UEFI_PFLASH_CODE` and `UEFI_PFLASH_VARS`, and now this seems to bring back 'just set `UEFI` and we'll magically find stuff'. And unfortunately I don't think we put in enough tests to cover everything that worked.

I've already found weird gremlins in this stuff before, like at one point at least if you set a particular combination of variables the test would execute as a UEFI test, but the variable `UEFI` would not be truth-y, which had some odd consequences...

okurz:
yes, true. This motivated me to at least start a test for "backend/qemu.pm" with https://github.com/os-autoinst/os-autoinst/pull/1191 . If people like it then I hope we can extend that one

PR #1191, opened 2019-08-06
Add tests for backend/qemu.pm

okurz:
Needs new requirement "Test::MockObject" but we already provide this in
openQA.

### DIFF
--- a/t/18-backend-qemu.t
+++ b/t/18-backend-qemu.t
@@ -1,0 +1,48 @@
+#!/usr/bin/perl
+
+use 5.018;
+use strict;
+use warnings;
+
+use Test::More;
+use Test::MockModule;
+use Test::MockObject;
+use Test::Warnings;
+
+BEGIN {
+    unshift @INC, '..';
+}
+
+use backend::qemu;
+
+
+my $proc = Test::MockModule->new('OpenQA::Qemu::Proc');
+$proc->mock(exec_qemu            => undef);
+$proc->mock(connect_qmp          => undef);
+$proc->mock(init_blockdef_images => undef);
+ok(my $backend = backend::qemu->new(), 'backend can be created');
+# disable any graphics display in tests
+$bmwqemu::vars{QEMU_APPEND} = '-nographic';
+# as needed to start backend
+$bmwqemu::vars{VNC} = '1';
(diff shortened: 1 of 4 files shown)


t/18-backend-qemu.t
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/perl
+
+use 5.018;
+use strict;
+use warnings;
```
cfconrad: why not `use Mojo::Base -base` ?

Martchus: I would not use that base class just for enabling strict and warnings.

cfconrad: I thought it makes in general good default settings (feature, utf8 etc). Maybe not `Mojo::Base -base` but `Mojo::Base -strict`.
Just asking for best practice, as I currently thought it is preferred instead of using all the same `use` statements in each file.

okurz: not sure, I just used the same as in t/18-qemu.t

Martchus: I would merge it as-is. If we decide to change this we can do it later and consistently for all tests.

cfconrad:
When I run it, I get some 
```
Use of uninitialized value in string eq at ../backend/qemu.pm line 742.
```
I think the ARCH variable is missing. I think it's not worth to avoid it?!

okurz:
This is actually one of the reasons why I wanted to have a test :)
I have the same warnings and also a fix however I also would like to see the test fail first before.

Am 8. August 2019 12:05:04 MESZ schrieb cfconrad <notifications@github.com>:

t/18-backend-qemu.t
```diff
@@ -0,0 +1,48 @@
...
+
+use Test::More;
+use Test::MockModule;
+use Test::MockObject;
+use Test::Warnings;
```
okurz: @cfconrad I fixed the warnings with https://github.com/os-autoinst/os-autoinst/pull/1194 and check for no warnings with this line

t/18-backend-qemu.t
okurz: I see this as the necessary formal approval of this PR then :)

t/18-backend-qemu.t
cfconrad: thx for hint!

PR #1192, opened 2019-08-07
spec: Add missing dependency to 'Data::Dumper'

okurz:
See https://github.com/os-autoinst/os-autoinst/pull/1188#issuecomment-518847212


Martchus:
Note that this module is contained by `perl-base` anyways. But I guess it can not harm to add it explicitly.

AdamWill:
in Fedora it's not, but in Fedora we get an automatically-generated dependency anyway, so it works.

PR #1194, opened 2019-08-07
Fix superfluous blank line; Fix warnings

okurz:
* Fix blank line on process output
* Fix undefined warnings on undefined 'ARCH'

### DIFF
--- a/osutils.pm
+++ b/osutils.pm
@@ -107,6 +107,7 @@ sub _run {
     $p->wait_stop;
     close($p->$_ ? $p->$_ : ()) for qw(read_stream write_stream error_stream);
 
+    chomp($out) if $out;
     return $p->exit_status, $out;
 }
 
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -625,8 +625,9 @@ sub start_qemu {
     # misc
     my $arch_supports_boot_order = $vars->{UEFI} ? 0 : 1;    # UEFI/OVMF supports ",bootindex=N", but not "-boot order=X"
     my $use_usb_kbd;
+    my $arch = $vars->{ARCH} // '';
 
-    if ($vars->{ARCH} eq 'aarch64' || $vars->{ARCH} eq 'arm') {
+    if ($arch eq 'aarch64' || $arch eq 'arm') {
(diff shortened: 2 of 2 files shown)


osutils.pm
```diff
@@ -100,7 +100,7 @@ sub _run {
     $p->channels(0)->quirkiness(1)->internal_pipes(0)->separate_err(0)->start;
     $p->on(stop => sub {
             while (defined(my $line = $p->getline)) {
-                $out .= $line;
+                chomp($out .= $line);
```
cfconrad: Just curious, do we really need the output flatten to one line?

okurz: well, every "line" should just be that, one line, isn't it?

cfconrad: I think, output like
```
a
b
c
```
become
`abc`
instead of only stripping the last `\n`. If this is what you won't, ok.

Martchus: Not according to my test:

```
  DB<1> $foo = "foo\nbar\nfoo\n";

  DB<2> print($foo);
foo
bar
foo

  DB<3> print(chomp($foo));
1
  DB<4> print($foo);
foo
bar
foo
  DB<5> 
```

okurz: hm, because I also call chomp around the $out, could be. I should crosscheck this then.

cfconrad: I agree what you are tested is what I would expect. But I think we have a different case, more like this:
```
open(my $f, '<', '/etc/os-release') || die("ERROR on read");
my $out = "";
while (my $line = readline($f)) {
        chomp($out .= $line);
}
print $out;
close($f);
```

cfconrad: Maybe I have a misunderstanding from that `$p->getline` and it act different!?

Martchus: @cfconrad I get your point - my example is not like this code. Then the code looks wrong, indeed.

okurz: any suggestion what I should do instead?

cfconrad: You could `chomp($out);` before return.

okurz: yes, of course! Thanks :)

okurz:
ok, so let's merge this first so that the tests in https://github.com/os-autoinst/os-autoinst/pull/1191 don't show warnings anymore :)

PR #1196, opened 2019-08-08
Fix in-package test failure on git

okurz:
8d38d28 introduced a now fatal check if the version can be evaluated or
not. The package tests failed silently in before because git was not
available in the build context when it even should not be available
because we want to use a fixed version from the package within the
isotovideo application.

This commit moves the version replacement from %install to %prep to
adjust the version in the isotovideo source code before installing it
hence having an effect on both the installed file as well as the file
used by the tests in %check.

This fixes the problem observed in OBS with the error message:
"Can't exec "git": No such file or directory at
/home/abuild/rpmbuild/BUILD/os-autoinst-4.5.1565258100.2ea934eb/isotovideo
line 107."

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -82,6 +82,8 @@ This package contains openvswitch support for os-autoinst.
 %prep
 %setup -q
 sed -e 's,/bin/env python,/bin/python,' -i crop.py
+# Replace version number from git to what's reported by the package
+sed  -i 's/ my $thisversion = qx{git.*rev-parse HEAD}.*;/ my $thisversion = "%{version}";/' isotovideo
 
 %build
 mkdir -p m4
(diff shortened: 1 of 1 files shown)


okurz:
I don't consider it *that* bad. As soon as a new commit reaches master builds are triggered in devel:openQA. So assuming that everyone is subscribing to notifications in the build project one way or another a timely feedback is provided. What we can try of course is to build *one* package variant as part of each PR and leave it to OBS to build all other repos and architectures. WDYT?

kalikiana:
> What we can try of course is to build _one_ package variant as part of each PR and leave it to OBS to build all other repos and architectures. WDYT?

Yeah, I think that'd be great. I'm really interested in verifying the packaging here first of all, not building the world. As it is, even small changes to spec files potentially break all sorts of things. It would be nice knowing that a typo, or in this case misplacement of a command, fail a PR early.

PR #1199, opened 2019-08-14
Improve result display of validate_script_output

perlpunk:
See also https://progress.opensuse.org/issues/54548

Example code:

    package your::test::package;
    ...
    sub run {
        validate_script_output 'pwd', sub { m/some.*regex/ };
    }

Before the output showed this:

    # wait_serial expected: (whatever the output was)

    # Result:
    (whatever the output was)

"wait_serial" is wrong here, and it showed the actual output twice.

Now it shows:

    validate_script_output got:
    (whatever the output was)

    Check function (deparsed code):
    {
        package your::test::package;
        /some.*regex/;
    }

Also validate_script_output() can now take a simple regular expression qr/foo/
instead of a coderef.

Suggestions for improvement of the output welcome.
The deparsed code is generated with the B::Deparse module and usually looks a bit different than the original code.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -33,7 +33,8 @@ use OpenQA::Isotovideo::NeedleDownloader;
 use Digest::MD5 'md5_base64';
 use Carp qw(cluck croak);
 use MIME::Base64 'decode_base64';
-use Scalar::Util 'looks_like_number';
+use Scalar::Util qw(looks_like_number reftype);
+use B::Deparse;
 
 require bmwqemu;
 use constant OPENQA_LIBPATH => '/usr/share/openqa/lib';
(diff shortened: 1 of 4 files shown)


testapi.pm
```diff
@@ -33,7 +33,8 @@ use OpenQA::Isotovideo::NeedleDownloader;
 use Digest::MD5 'md5_base64';
 use Carp qw(cluck croak);
 use MIME::Base64 'decode_base64';
-use Scalar::Util 'looks_like_number';
+use Scalar::Util qw( looks_like_number reftype );
```
okurz: to stay in line with the existing style please delete the spaces after '(' and before ')'.

https://travis-ci.org/os-autoinst/os-autoinst/builds/571928539#L838 shows you other style issues

testapi.pm
```diff
@@ -1119,19 +1120,23 @@ sub get_test_data {
 =head2 validate_script_output
 
   validate_script_output($script, $code [, timeout => $timeout] [,quiet => $quiet])
+  validate_script_output($script, $regexp [, timeout => $timeout] [,quiet => $quiet])
```
okurz: Adding a second line with alternative syntax is uncommon for the test API and might be confusing. I suggest to specify the alternatives within a single line, e.g.

```perl
validate_script_output($script, $code | $regexp [, timeout => $timeout] [,quiet => $quiet])
```

perlpunk:
@okurz like discussed in https://progress.opensuse.org/issues/54548, should I remove the Deparse part?

okurz:
no I think how you did it with the stripping of the automatically added use statement it's fine.

perlpunk:
While writing the test, I noticed that existing tests do this:

    ok(!validate_script_output('script', sub { m/output/ }), 'validating output with default timeout');

But the sub `validate_script_output` isn't returning anything explicitly.
This is the last statement in the sub:

    if ($res eq 'fail') {
        croak "output not validating";
    }

So the false return value is relying on that last statement being evaluated to false.
If some day another statement is added, that evaluates to true, all the tests will break.

So I added an explicit `return 0` to make the actual behavior transparent.

perlpunk:
@okurz I guess I should also add B::Deparse to `cpanfile`?

perlpunk:
squashed and rebased to master

okurz:
yes, the added dependency to cpanfile is required but also please add to dist/rpm/os-autoinst.spec

perlpunk:
@okurz ok, I added it (rebased and squashed)

AdamWill:
This definitely looks like an improvement, thanks :+1:

perlpunk:
Rebased and resolved conflicts

okurz:
please check the coverage. codecov reports only 10% of diff hit and the overall project coverage reduced below our threshold. I will merge it nevertheless because the change looks fine however I would really appreciate you looking into it. Thx.

PR #1201, opened 2019-08-19
Fix unexpected error on xz in save_memory_dump()

cfconrad:
save_memory_dump() failed sporadically. This happen due to the call of
`system('xz ...')` and `autodie :all` is enabled.
Cause `xz` also return `!= 0` when warnings appear, which actually isn't an
error!

Adding `--no-warn` avoid this exit code and just return != 0 on error.

Also used `runcmd()` instead of `system()` which makes error hunting more
easy.

Tickets
 * https://progress.opensuse.org/issues/48671
 * https://progress.opensuse.org/issues/55595

Maybe related:
 * https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8202

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -338,7 +338,7 @@ sub save_memory_dump {
 
     if ($compress_method eq 'xz') {
         if (defined which('xz')) {
-            system(('xz', '-T', $compress_threads, "-v$compress_level", "ulogs/$filename"));
+            runcmd('xz', '--no-warn', '-T', $compress_threads, "-v$compress_level", "ulogs/$filename");
         }
         else {
             bmwqemu::fctwarn('xz not found; falling back to bzip2');
@@ -347,7 +347,7 @@ sub save_memory_dump {
     }
 
     if ($compress_method eq 'bzip2') {
-        system(('bzip2', "-v$compress_level", "ulogs/$filename"));
+        runcmd('bzip2', "-v$compress_level", "ulogs/$filename");
     }
 
     return;


backend/qemu.pm
```diff
@@ -338,7 +338,7 @@ sub save_memory_dump {
 
     if ($compress_method eq 'xz') {
         if (defined which('xz')) {
-            system(('xz', '-T', $compress_threads, "-v$compress_level", "ulogs/$filename"));
+            runcmd('xz', '--no-warn', '-T', $compress_threads, "-v$compress_level", "ulogs/$filename");
```
okurz: Interesting, my version of xz, openSUSE Leap 15.1, does not mention this option on `xz --help` but the manpage does.

cfconrad:
> Could you share how the ouput looks different now?
> 
> Or could you share a link to a test run in openQA how the output differs?

http://cfconrad-vm.qa.suse.de/tests/5968

Output is now:
```
[2019-08-19T14:23:18.536 CEST] [debug] running xz --no-warn -T 1 -v6 ulogs/save_memory_dump_01-vm-memory-dump
[2019-08-19T14:25:01.653 CEST] [debug] ulogs/save_memory_dump_01-vm-memory-dump: 49.8 MiB / 220.3 MiB = 0.226, 2.1 MiB/s, 1:43
```

Or just run a runcmd('false'); to see how it would look like if the command produce `!= 0`
```
[2019-08-19T14:23:18.536 CEST] [debug] running xz --no-warn -T 1 -v6 ulogs/save_memory_dump_01-vm-memory-dump
[2019-08-19T14:25:01.653 CEST] [debug] ulogs/save_memory_dump_01-vm-memory-dump: 49.8 MiB / 220.3 MiB = 0.226, 2.1 MiB/s, 1:43
```

PR #1202, opened 2019-08-20
Allow special value 'once' for PXEBOOT, to PXE boot...once

AdamWill:
For a PXE install test, we want to boot PXE just the *first*
time, then revert to default boot order (much like we boot
once=d by default for optical media installs). This allows the
special value 'once' for the PXEBOOT var to request this.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -792,7 +792,10 @@ sub start_qemu {
         }
 
         if ($arch_supports_boot_order) {
-            if ($vars->{PXEBOOT}) {
+            if ($vars->{PXEBOOT} // '' eq 'once') {
+                sp("boot", "once=n");
+            }
+            elsif ($vars->{PXEBOOT}) {
                 sp("boot", "n");
             }
             elsif ($vars->{BOOTFROM}) {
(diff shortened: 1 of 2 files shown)


backend/qemu.pm
```diff
@@ -792,7 +792,10 @@ sub start_qemu {
         }
 
         if ($arch_supports_boot_order) {
-            if ($vars->{PXEBOOT}) {
+            if ($vars->{PXEBOOT} eq 'once') {
```
okurz: Please see https://travis-ci.org/os-autoinst/os-autoinst/builds/574597957#L814 , you are now comparing against potentially undefined, better:

```perl
if (($vars->{PXEBOOT} // '') eq ...
```

AdamWill: oh yeah, good spot.

backend/qemu.pm
```diff
@@ -792,7 +792,10 @@ sub start_qemu {
         }
 
         if ($arch_supports_boot_order) {
-            if ($vars->{PXEBOOT}) {
+            if ($vars->{PXEBOOT} // '' eq 'once') {
```
okurz: I hope this works without parentheses …

AdamWill: D'oh...I didn't test it, didn't notice you had parentheses in yours. Let me go test in in a test script real quick.

...

well, seems fine according to this test script:

    #!/bin/perl

    use warnings;
    use strict;

    my $vars = {foo => 'foo'};
    print "True!\n" if ($vars->{bar} // '' eq 'once');
    print "True 2!\n" if ($vars->{foo} // '' eq 'foo');

it runs without complaints and prints 'True 2!'.

PR #1203, opened 2019-08-23
Add configuration option for number of NVMe queues

frankenmichl:
There are scenarios, where we should have control over the number of
queues a NVMe device has. The number of queues varies, and recent
versions of Qemu emulate 32 queues, for example.
In some scenarios, like the one depicted in [1], we need to set the
number of queues to a value smaller than the number of NUMA-nodes in the
system.
In order to allow to test this with reasonable ressource requirements,
instead of running a machine wit 40+ NUMA-nodes, it is more reasonable
to run a test with four nodes and two NVMe queues instead.
With these changes, it is possible to configure the NVMe queues by
setting the HDDNUMQUEUES_n parameter for the disk number n.

[1] https://progress.opensuse.org/issues/55502

A proof run can be found here: http://mmoese-vm.qa.suse.de/tests/28#

### DIFF
--- a/OpenQA/Qemu/BlockDevConf.pm
+++ b/OpenQA/Qemu/BlockDevConf.pm
@@ -96,7 +96,7 @@ sub del_overlay {
 }
 
 sub _push_new_drive_dev {
-    my ($self, $id, $drive, $model) = @_;
+    my ($self, $id, $drive, $model, $num_queues) = @_;
 
     die 'PFlash drives are not supported by DriveDevice, use PFlashDevice'
       if $model eq 'pflash';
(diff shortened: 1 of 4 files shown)


frankenmichl:
I did not even manage to find any tests for NVMe's. But anyway, code coverage is mainly impacted for files not even touched..

Martchus:
Ok. For me it is hard to approve this because I don't really know what this is about without digging into the topic myself. But don't worry about the code coverage report - it is sometimes inconsistent.

jlausuch:
LGTM

PR #1204, opened 2019-08-27
Add an error message to the script_run

b10n1k:
This PR adds a fail_message to the script_run command in the same way as it is used in the assert_script_run in testapi.


testapi.pm
```diff
@@ -951,7 +951,7 @@ Deprecated mode
 Run C<$cmd> (in the default implementation, by assuming the console prompt and typing
 the command). If C<$timeout> is greater than 0, wait for that length of time for
 execution to complete (otherwise, returns undef immediately). See C<distri->script_run>
-for default timeout.
+for default timeout. If C<$fail_message> is returned in the die message if specified.
```
Martchus: You mean if `C<$fail_message>` is specified it is included in the die message?

testapi.pm
```diff
@@ -953,6 +953,9 @@ the command). If C<$timeout> is greater than 0, wait for that length of time for
 execution to complete (otherwise, returns undef immediately). See C<distri->script_run>
 for default timeout.
 
+C<$output> can be used as a explanatory text that it will be displayed with the execution of
```
okurz: s/that it will be/that will be/

distribution.pm
```diff
@@ -139,13 +139,13 @@ sub script_run {
...
         if (testapi::is_serial_terminal) {
-            my $marker = " ; echo $str-\$?-";
+            my $marker = " ; echo $str-\$?- $args{output}";
             testapi::type_string($marker);
             testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
             testapi::type_string("\n");
         }
         else {
-            testapi::type_string " ; echo $str-\$?- > /dev/$testapi::serialdev\n";
+            testapi::type_string " ; echo $str-\$?- $args{output} > /dev/$testapi::serialdev\n";
```
okurz: as you add it twice please extract the generated string part into a variable. Actually `$marker` from L142 seems to be a good candidate already so probably you can simply do:

```perl
my $marker = " ; echo $str-\$?- $args{output}";
    …
}
else {
    testapi::type_string "$marker > /dev/$testapi::serialdev\n";
}
```

or just combine it all in `$str` already.

testapi.pm
```diff
@@ -969,9 +972,9 @@ sub script_run {
     my %args = compat_args(
         {
             timeout => undef,
+            output => '',
             quiet   => testapi::get_var('_QUIET_SCRIPT_CALLS')
         }, ['timeout'], @_);
-
```
okurz: unrelated change, please revert

t/03-testapi.t
```diff
@@ -291,7 +291,7 @@ subtest 'check_assert_screen' => sub {
         is_deeply($autotest::current_test->{details}, [
                 {
                     result     => 'unk',
-                    screenshot => 'basetest-11.png',
+                    screenshot => 'basetest-10.png',
```
okurz: why change that?

testapi.pm
```diff
@@ -953,6 +953,9 @@ the command). If C<$timeout> is greater than 0, wait for that length of time for
 execution to complete (otherwise, returns undef immediately). See C<distri->script_run>
 for default timeout.
 
+C<$output> can be used as a explanatory text that will be displayed with the execution of
```
Martchus: an explanatory

t/03-testapi.t
b10n1k: reverted  3669850

testapi.pm
b10n1k: reverted  3669850

b10n1k:
The 03-testapi.t is failing on the 'check quiet option on script runs' subtest but i cant understand why. if someone can help PM

Martchus:
Are you unable to run the test locally in general or does it only fail on the branch with your changes? What is the exact error message?

b10n1k:
@Martchus it fails on the branch. 

The errors from the local run are:

```
not ok 9 - Check default quiet argument
    #   Failed test 'Check default quiet argument'
    #   at ./03-testapi.t line 616.
    #          got: undef
    #     expected: '1'
Argument "output" isn't numeric in numeric gt (>) at ../distribution.pm line 139.
    not ok 10 - script_run with _QUIET_SCRIPT_CALLS=1
    #   Failed test 'script_run with _QUIET_SCRIPT_CALLS=1'
    #   at ./03-testapi.t line 620.
    #          got: undef
    #     expected: '0'
```

okurz:
@Martchus yes, I don't know who introduces these rather long timeouts. It's annoying, I agree. We can try to fix that again, e.g. with "mocking sleep". At best also with a test to not exceed runtime limits ;)

okurz:
To remind: Please cover the new functionality in tests

PR #1205, opened 2019-08-28
spec: Split build-, runtime and test requirements

okurz:
Follow-up to https://github.com/os-autoinst/openQA/pull/2161 applied for
os-autoinst.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -40,6 +40,7 @@ requires 'Net::SSH2';
 requires 'POSIX';
 requires 'Thread::Queue';
 requires 'Time::HiRes';
+requires 'Try::Tiny';
 requires 'XML::LibXML';
 requires 'base';
 requires 'constant';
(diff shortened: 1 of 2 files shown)


okurz:
Tested successfully in https://build.opensuse.org/package/show/home:okurz:branches:devel:openQA:os-autoinst:feature:spec_split_deps/os-autoinst

ggardet:
@okurz Because of this update, `qemu-img` is not installed when `os-autoinst` is installed, but it is required. This leads to https://progress.opensuse.org/issues/57014#note-3
```
[2019-10-11T07:43:12.675 EDT] [debug] Backend process died, backend errors are reported below in the following lines:
no kvm-img/qemu-img found
```

okurz:
ah, good to know. well, we have the ticket for this then, good! :)

PR #1206, opened 2019-08-29
docs: openSUSE package+repo improvements

okurz:
* docs: Mention devel package instead of hardcoded dependencies
* docs: Update openSUSE repos using , priorities, better names

### DIFF
--- a/INSTALL.asciidoc
+++ b/INSTALL.asciidoc
@@ -13,12 +13,12 @@ installation guide]
 git clone YOUR-WRITEABLE-COPY-OF/os-autoinst.git
 cd os-autoinst
 ------------------------------------------------
-3.  install the BuildRequires, as per the https://build.opensuse.org/package/show/devel:openQA/os-autoinst[spec file]
+3.  install the build requirements as visible in +dist/rpm/os-autoinst.spec+
+or if you have the openSUSE repositories added:
 +
------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


INSTALL.asciidoc
```diff
@@ -13,12 +13,18 @@ installation guide]
...
 +
------------------------------------------------------------------------------------------------------------------------------------
-sudo zypper in git-core make libtheora-devel opencv-devel fftw3-devel libsndfile-devel \
-  pkg-config libtool autoconf automake gcc-c++
------------------------------------------------------------------------------------------------------------------------------------
+----
+sudo zypper source-install --build-deps-only os-autoinst
+----
+
+or
```
Martchus: This interrupts the numbered list. Since we now have `os-autoinst-devel` you might just mention this one. (It seems to be better after all because it will pull future dependencies on updates.)

okurz: > It seems to be better after all because it will pull future dependencies on updates

Why is that? You mean `zypper source-install --build-deps-only os-autoinst` only reads information from the locally installed package or what?

Martchus: > Why is that? You mean zypper source-install --build-deps-only os-autoinst only reads information from the locally installed package or what?

I mean if you use `zypper source-install --build-deps-only` it will install all build dependencies present at this time. When we add further build dependencies they are not installed automatically when updating the system.

okurz: Right! That's a benefit of the devel package.

PR #1208, opened 2019-09-05
Update test_order.json if test schedule changes at runtime

mdoucha:
This pull request is part of os-autoinst/openQA#2302

After talking to @asmorodskyi via IRC, I've decided that the proper way to notify OpenQA server of test schedule change is by calling `write_test_order()` in `autotest::loadtest()`. The original idea was to call `write_test_order()` from the testcase that scheduled additional tests but that would expose internal implementation details to third-party code and os-autoinst developers might not notice that test schedule may change at runtime.

Both this PR and os-autoinst/openQA#2302 can be safely used with an unpatched version of the other package. The test modules scheduled at runtime won't show up in test results (just like now) but there'll be no errors or crashes.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -37,8 +37,7 @@ our %tests;        # scheduled or run tests
 our @testorder;    # for keeping them in order
 our $isotovideo;
 our $process;
-
-my $tests_running = 0;
+our $tests_running = 0;
 =head1 Introduction
 
 OS Autoinst decides which test modules to run based on a distribution specific
--- a/autotest.pm
+++ b/autotest.pm
@@ -37,6 +37,8 @@ our %tests;        # scheduled or run tests
 our @testorder;    # for keeping them in order
 our $isotovideo;
 our $process;
+
+my $tests_running = 0;
 =head1 Introduction
 
 OS Autoinst decides which test modules to run based on a distribution specific
@@ -145,6 +147,10 @@ sub loadtest {
 
     return unless $test->is_applicable;
     push @testorder, $test;
+
+    # Test schedule may change at runtime. Update test_order.json to notify
+    # the OpenQA server of the change.
+    write_test_order() if $tests_running;
     bmwqemu::diag("scheduling $test->{name} $script");
 }
 
@@ -219,6 +225,7 @@ sub load_snapshot {
 sub run_all {
     my $died      = 0;
     my $completed = 0;
+    $tests_running = 1;
     eval { $completed = autotest::runalltests(); };
     if ($@) {
         warn $@;
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -242,6 +242,48 @@ is(@{$autotest::tests{'tests-fatal'}}{@opts}, @{$autotest::tests{'tests-fatal' .
   && is(@{$autotest::tests{'tests-fatal' . $_}}{name}, 'fatal#' . $_)
   for 1 .. 10;
 
+# test scheduling new modules at job runtime
+subtest 'test scheduling test modules at test runtime' => sub {
+    $autotest::tests_running = 0;
+    @autotest::testorder     = ();
+    %autotest::tests         = ();
+
+    my %json_data;
+    my $json_filename = bmwqemu::result_dir . '/test_order.json';
+    my $testorder     = [
+        {
+            name     => 'scheduler',
+            category => 'tests',
+            flags    => {},
+            script   => 'tests/scheduler.pm'
+        },
+        {
+            name     => 'next',
+            category => 'tests',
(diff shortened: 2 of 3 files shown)


autotest.pm
```diff
@@ -145,6 +145,10 @@ sub loadtest {
 
     return unless $test->is_applicable;
     push @testorder, $test;
+
+    # Test schedule may change at runtime. Update test_order.json to notify
+    # the OpenQA server of the change.
+    write_test_order() if defined($process);
```
okurz: so I assume the impact is (only) that the file is written some more times?

mdoucha: Yes, and only if `loadtest()` gets called at test runtime. The tests scheduled before VM start will be written all at once in `runalltests()` as before.

okurz: I see, thx.

okurz:
checks have failed but it looks like not caused by your changes. I have retriggered the build on travis CI: https://travis-ci.org/os-autoinst/os-autoinst/builds/581164554?utm_source=github_status&utm_medium=notification

coolo:
This sounds very dangerous to me - the test order file is not a random file, but a communication vehicle to the worker.

And we have enough trouble with reading it realiably as it is

mdoucha:
> This sounds very dangerous to me - the test order file is not a random file, but a communication vehicle to the worker.
> 
> And we have enough trouble with reading it realiably as it is

For all jobs that have their entire test schedule available before VM start, this PR is a no-op. But I suppose I could add a check to os-autoinst/openQA#2302 whether `test_order.json` was loaded successfully and force a reload at next tick if not.

pevik:
@Martchus can you please review it?
@coolo Is there a way, how to allow this patch for these what need it and secure things for the rest?

coolo:
Yes, by giving the team time to look at it.

pevik:
> Yes, by giving the team time to look at it.

Sure, sorry for pushing for it.

mdoucha:
@okurz @Martchus Please review the code again. The original condition `write_test_order() if defined($process);` was unreliable because `$process` might remain undefined inside the subprocess due to call chaining: `$process = process(...)->...->start;`

autotest.pm
```diff
@@ -37,6 +37,7 @@ our %tests;        # scheduled or run tests
 our @testorder;    # for keeping them in order
 our $isotovideo;
 our $process;
+my $tests_running = 0;
 =head1 Introduction
```
okurz: please add a blank line in before

mdoucha:
> I can see that the coverage decreased now below the allowed threshold. Can you please try to add a test covering the added code? `t/08-autotest.t` is a good candidate to look into. If you think you really can not do that then adjust the coverage threshold in https://github.com/os-autoinst/os-autoinst/blob/master/codecov.yml#L16

I'd like to point out that the Codecov report above is utter garbage. I've added 3 lines of executable code. One line is executed fully. One line is executed partially (condition evaluates to false and the rest of the line is skipped). One line is not executed. That's the only coverage difference that Codecov should report.

Instead, Codecov reports huge loss of coverage in files that have nothing to do with this PR.

Anyway, I'll write a quick testcase for `load_test()` that'll cover the partially executed line. I'm not going to write a test for the whole subprocess entry point.

okurz:
> I'd like to point out that the Codecov report above is utter garbage

we do accept pull requests ;)

mdoucha:
According to Codecov, adding a unit test *decreased* coverage even more. ‾\\\_(ツ)\_/‾

asmorodskyi:
> According to Codecov, adding a unit test _decreased_ coverage even more. ‾\_(ツ)_/‾

@okurz , @coolo, @Martchus  - several times before , during reviews of my PRs I faced with situations when Codecov has failed status but no one actually bother because everybody aware that it is broken. My take that if something is always fail it should be turned off or fixed ...

okurz:
@asmorodskyi it does not always fail and I would also appreciate if we can fix it

asmorodskyi:
> @asmorodskyi it does not always fail and I would also appreciate if we can fix it

this is what I am saying - choosing between having test which is "not always fail" or as I stated "may fail when it actually suppose to pass " and not having test at all I am choosing - **not have test at all** . Because you can not trust its results anyway and you getting poisoning habit to ignore failures.

pevik:
@okurz would be great to assign somebody to try to fix it soon.

pevik:
Again, thanks for merging this. So we just need to wait this to be deployed before merging https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8329

okurz:
I agree. If you talk about the tools team then @coolo is our product owner and the go-to-person

PR #1210, opened 2019-09-10
Fix PXEBOOT once to actually check for 'once' value

aaannz:
Without parentheses this will return true every time PXEBOOT is set to true.


okurz:
I *knew* it ;)

AdamWill:
d'oh! sorry about that.

PR #1211, opened 2019-09-12
Use hda audio instead of deprecated ac97

asdil12:
Ticket: https://progress.opensuse.org/issues/55958#change-243110
Verification (on x86_64 - at least it doesn't break that): http://artemis.suse.de/tests/1644#step/aplay/21

As the default volume seems to be different for hda, we need to have some needles updated:

https://gitlab.suse.de/openqa/os-autoinst-needles-sles/merge_requests/1218/diffs


asdil12:
@okurz Fixed the typo

PR #1217, opened 2019-10-02
Allow loading needles form pool directory

Martchus:
See https://progress.opensuse.org/issues/56789


okurz:
s/form/from/

Martchus:
Yes, from os-autoinst's view the pool directory is just its current working directory. I also didn't change that.

However, the term "pool directory" is commonly used and I think it makes sense to use it in comments and variable names so it is easier to understand the meaning of the code inside the bigger picture of the whole architecture. In my opinion that is usually the hardest aspect of understanding the code. I also don't see a conflict with os-autoinst's independence because it is still clear that we're technically just using the current working directory here.

Note that this is also not the first place where the pool directory is mentioned within os-autoinst (the comment `# global assumption: cwd == pooldir` is just copied).

needle.pm
```diff
@@ -51,20 +52,29 @@ sub new {
...
-        die "Needle $jsonfile is not under project directory $bmwqemu::vars{PRJDIR}";
-    }
 
-    # $json->{file} contains path relative to $bmwqemu::vars{PRJDIR}
-    # $jsonfile contains absolute path within $bmwqemu::vars{PRJDIR}
+    # locate the needle file in one of the lookup directories ($bmwqemu::vars{PRJDIR} or the pool directory)
+    # - This code initializes $json->{file} so it contains the path relative to the lookup directory and re-assigns
+    #   $jsonfile to contain the absolute path.
+    # - The pool directory is considered to support loading custom needles with openqa-clone-custom-git-refspec.
+    my $pool_directory_checkout = (cwd . '/os-autoinst-needles-opensuse');    # global assumption: cwd == pooldir
```
coolo: I thought you gave an example when you mentioned `os-autoinst-needles-opensuse` - I don't think this repo name should be in our code.

Martchus: Makes sense. Then we need a variable to pass additional search locations.

coolo: So we? Can't this be a fixed name that the worker clones into?

Martchus:
I was under the impression that `openqa-clone-custom-git-refspec` would use a fixed, hard-coded subdirectory for needles (`os-autoinst-needles-opensuse`). However, it actually just sets `$bmwqemu::vars{NEEDLES_DIR}` to a repo URL. Then isotovideo itself clones it within the current working directory and sets `$bmwqemu::vars{NEEDLES_DIR}` to that repo path. So simply using the `NEEDLES_DIR` variable should do the trick. I'm still checking whether the directory is within the current working directory because the user might specify an arbitrary path as `NEEDLES_DIR`.

needle.pm
```diff
@@ -51,20 +52,32 @@ sub new {
...
 
-    # $json->{file} contains path relative to $bmwqemu::vars{PRJDIR}
-    # $jsonfile contains absolute path within $bmwqemu::vars{PRJDIR}
+    # locate the needle file in one of the lookup directories
+    # - This code initializes $json->{file} so it contains the path relative to the lookup directory and re-assigns
+    #   $jsonfile to contain the absolute path.
+    # - A custom needle repo checkout referenced by the $bmwqemu::vars{NEEDLES_DIR} variable is considered to support loading
+    #   custom needles with the openqa-clone-custom-git-refspec script. The custom needle repo must be within the current working
+    #   directory (usually the openQA worker's pool directory).
+    # - Otherwise, the needle will be looked up under $bmwqemu::vars{PRJDIR}.
```
okurz: please break lines

needle.pm
```diff
@@ -51,20 +52,32 @@ sub new {
...
-        die "Needle $jsonfile is not under project directory $bmwqemu::vars{PRJDIR}";
-    }
 
-    # $json->{file} contains path relative to $bmwqemu::vars{PRJDIR}
-    # $jsonfile contains absolute path within $bmwqemu::vars{PRJDIR}
+    # locate the needle file in one of the lookup directories
+    # - This code initializes $json->{file} so it contains the path relative to the lookup directory and re-assigns
+    #   $jsonfile to contain the absolute path.
+    # - A custom needle repo checkout referenced by the $bmwqemu::vars{NEEDLES_DIR} variable is considered to support loading
+    #   custom needles with the openqa-clone-custom-git-refspec script. The custom needle repo must be within the current working
```
okurz: Better stay with "os-autoinst speak" only, i.e. just reference NEEDLES_DIR, don't care about openQA tools, or just mention them as example, e.g. as you did one line below

needle.pm
```diff
@@ -51,20 +52,32 @@ sub new {
...
+    #   directory (usually the openQA worker's pool directory).
```
okurz: just to make sure you don't assume I still object, that's good wording, just mentioning openQA as an example

t/01-test_needle.t
```diff
@@ -398,6 +399,22 @@ is($needle->{file}, 'out-of-def-prj/test/data/other-desktop-dvd-20140904.json',
 eval { $needle = needle->new('out-of-prj/test/data/some-needle.json') };
 ok($@, 'died when accessing needle outside of prjdir');
 
+subtest 'needle->new accepts path relative to pool directory' => sub {
```
okurz: "relative to working directory"?

t/01-test_needle.t
```diff
@@ -398,6 +399,22 @@ is($needle->{file}, 'out-of-def-prj/test/data/other-desktop-dvd-20140904.json',
...
+    # prepare temporary pool directory
+    my $pool_dir = tempdir(CLEANUP => 1);
+    note("using temporary pool directory $pool_dir");
+    chdir($pool_dir);
```
okurz: again, no need to reference "pool" which has no meaning in this context

t/01-test_needle.t
```diff
@@ -398,6 +399,22 @@ is($needle->{file}, 'out-of-def-prj/test/data/other-desktop-dvd-20140904.json',
...
+    my $needles_dir = $bmwqemu::vars{NEEDLES_DIR} = "$pool_dir/os-autoinst-needles-opensuse";
```
okurz: no need to mention something "opensuse-specific"

PR #1218, opened 2019-10-04
Update to new perltidy version 20190915

perlpunk:


okurz:
Same applies as for https://github.com/os-autoinst/os-autoinst/pull/1163 . In general updating the tidy version is in most if not all cases very disruptive, hence costly, as all os-autoinst-distri-opensuse contributors also need to "get the message". We have the tidy version 20190915 in both openSUSE:Factory as well as https://build.opensuse.org/package/show/devel:openQA:Leap:15.1/perl-Perl-Tidy so I think this is the way to go however it would be good if we can ensure that related repos are compatible, e.g. openQA and os-autoinst-distri-opensuse. Do you know if source code adjusted to the new style would be compatible with the old tidy version?

perlpunk:
I would have to check with CPAN, because I don't have the old version anymore. That was the actual reason to update.

perlpunk:
Tested both os-autoinst and os-autoinst-distri-opensuse with Perl-Tidy-20190601, no changes

okurz:
> Tested both os-autoinst and os-autoinst-distri-opensuse with Perl-Tidy-20190601, no changes

So, does that mean we can update the style *first* and update the cpanfiles and dependencies *afterwards*?

perlpunk:
> So, does that mean we can update the style _first_ and update the cpanfiles and dependencies _afterwards_?

Yeah, it looks like. Shall I split my PR?

Martchus:
Likely the distribution should define its own tidy version independently (like openQA does). Having to think about every existing test distribution which might rely on the tidy version from this repo doesn't scale and seems needless.

okurz:
On Monday, 7 October 2019 11.13.35 CEST Martchus wrote:
> […] Having to think about every existing test distribution
> which might rely on the tidy version from this repo doesn't scale and seems
> needless.

Not sure what you mean. We within SUSE working in the "QA tools" team at least 
should have the obligation to provide as little disruptions to the downstream 
projects that other SUSE engineers rely on.

Martchus:
I mean this shouldn't be a disruptive change. Downstream should operate independently of the tidy version we use in os-autoinst.

okurz:
For now I think the best practice for a perl tidy version update is:

1. Ensure perl-Perl-Tidy is available in the used version in "usual repos", e.g. devel:openQA and subrepos as well as openSUSE:Factory (i.e. Tumbleweed)
1. Provide in parallel a PR updating the code and version dependency for openQA, os-autoinst, os-autoinst-distri-opensuse, os-autoinst-distri-example, os-autoinst-distri-openQA
1. Inform "usual contributors" with a mailing list post, e.g. on openqa@suse.de
1. Proactive information in current open PRs that a rebase and tidy call is necessary
1. Providing help and support in usual communication forums in case of questions and confusion by users, e.g. internal testing channel

So sounds a bit tedious but actually nothing that is tricky or hard to do. Can you commit to doing that?

okurz:
@mdoucha yes, the package in both Tumbleweed as well as in https://build.opensuse.org/package/show/devel:openQA:Leap:15.1/perl-Perl-Tidy is updated to 20190915 already. I will merge this PR and handle the followup in https://progress.opensuse.org/issues/58394

asmorodskyi:
> I mean this shouldn't be a disruptive change. Downstream should operate independently of the tidy version we use in os-autoinst.

Totally agree and I would follow up on this just after GMC . This would prevent thing like this https://travis-ci.org/os-autoinst/os-autoinst-distri-opensuse/jobs/600599853 from happening ... It's very nice that @okurz  prepared https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8735 but because no one actually merged on time his work does not help much :( also some downstream contributors time was wasted on figure out all this story and as far as I can tell it is not the first time we facing this problem and something telling me not the last time unless we will change the model ...

PR #1219, opened 2019-10-07
Add possibility to have a comment text in script_run (fixed)

okurz:
Re-added functionality reverted in 54352a1c with a fix for the
regression and a test covering the expected string appended to the
executed command.

Related progress issue: https://progress.opensuse.org/issues/55151

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -107,7 +107,7 @@ sub become_root {
 
 =head2 script_run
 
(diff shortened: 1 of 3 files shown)


okurz:
hm, I wonder why codecov reports that the patch is not covered, the unit test really should execute the covered code paths.

PR #1221, opened 2019-10-08
Fix error handling when passing message from ws client to isotovideo

Martchus:
* Mojo::Log::warning is actually Mojo::Log::warn
* The return inside the catch does not return from the actual function


Martchus:
Unfortunately it doesn't fix bigger symptoms. It only prevents Perl warnings/errors and shows the actual warning as it was always intended.

PR #1222, opened 2019-10-08
command server: Improve dead-lock prevention with isotovideo

Martchus:
* Handle the 'stop_processing_isotovideo_commands' command in all cases where a message from isotovideo is received.
* Consider command processing being stopped in all cases where the command server sends commands to isotovideo.

This might help with https://progress.opensuse.org/issues/57707. "Might" because I have not reproduced the issue so far.


okurz:
you can try the following to reproduce from the os-autoinst dir:

```
./isotovideo -d casedir=t/data/tests include_modules=foo
```

this should load a valid test distribution but skip all modules and hence could be about the fastest way to reproduce.

PR #1223, opened 2019-10-10
Exclude names containing % or @ from HashKeyQuotes check

cfconrad:
Following hash is valid but need quoted key:
my $h = { '%bla@foo' => 42 };

### DIFF
--- a/tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
+++ b/tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
@@ -22,10 +22,10 @@ sub violates {
     return if !is_hash_key($elem);
 
     my $c = $elem->content;
-    # special characters
-    return if $c =~ m/[- \/<>.=_:\\\$]/;
+    # Quotes allowed, if not matching following regex
+    return unless $c =~ m/^(["'])[a-zA-Z][0-9a-zA-Z]*\1$/;
 
-    my $desc = q{Hash key with quotes};
+    my $desc = qq{Hash key $c with quotes};
     my $expl = q{Avoid useless quotes};
     return $self->violation($desc, $expl, $elem);
 }


okurz:
typo "conatining"

cfconrad:
thx

tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
```diff
@@ -23,7 +23,7 @@ sub violates {
 
     my $c = $elem->content;
     # special characters
-    return if $c =~ m/[- \/<>.=_:\\\$]/;
+    return if $c =~ m/[-@% \/<>.=_:\\\$]/;
```
perlpunk: I would quote the `@` here because one can interpolate arrays in regexes, so for examle `m/[@array]/` would add all elements of `@array`

`m/[-\@% \/<>.=_:\\\$]/;`

perlpunk: I don't know what kind of keys we get here, but turning that blacklist check into a whitelist might be better, e.g. only allowing alpha-numeric

cfconrad: I see. I wasn't sure about it. Thx

cfconrad:
I would like to continue on the idea of making a white list. From https://perldoc.perl.org/perldata.html
```
The => operator is mostly just a more visually distinctive synonym for a comma, but it also arranges for its left-hand operand to be interpreted as a string if it's a bareword that would be a legal simple identifier.
```
But I didn't found what a "legal simple identifier" exactly mean. So I would come up with
```
/^(?!\d)\w+$/aa
```
from https://perldoc.perl.org/perldata.html#Identifier-parsing

@perlpunk WDYT?

cfconrad:
Pushed the white list approach as I find it more suitable.
I kept commit "Exclude names containing % or @ from HashKeyQuotes check" to see the difference. Going to drop it when we agreed.

One thing left, the `_` was special char previously but I see no need for it as Perl don't require it. But when we now include it, we get some warnings in our projects. So I will not do it without agreement :)

cfconrad:
@coolo ping you as you are the only one worked on this file. Thx for taking a look.

tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
```diff
@@ -22,10 +22,11 @@ sub violates {
     return if !is_hash_key($elem);
 
     my $c = $elem->content;
-    # special characters
-    return if $c =~ m/[- \/<>.=_:\\\$]/;
+    # Quotes allowed, if not matching following regex
+    # TODO: PERL allow _ in simple identifiers, so we should do it too
```
okurz: Please don't add TODOs in commits. Either you see it as important enough for a ticket or just leave it out

cfconrad: ok
https://progress.opensuse.org/issues/58100

schlad: Please don't use imperative form here.
And on top I disagree with you. TODOs are just fine from my pov.

cfconrad:
@pevik the exception is in correlation to https://perldoc.perl.org/perldata.html#Identifier-parsing which actually  allow _ in "simple identifiers", but this rule doesn't. And I just wanna make a pointer to this circumstance.

tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
okurz: Of course feel free to disagree. I am glad to receive your opinion regardless which direction it takes :) I am referring to what is commonly done in this project and what I and others sticked to in before.

cfconrad:
What are we waiting for?

okurz:
I don't know either :)

PR #1224, opened 2019-10-10
Revert "Allow loading needles from pool directory" to fix loading time regression

okurz:
Reverts os-autoinst/os-autoinst#1217 to fix https://progress.opensuse.org/issues/57962

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -20,7 +20,6 @@ use strict;
 use warnings;
 use autodie ':all';
 
-use Cwd 'cwd';
 use File::Find;
 use File::Spec;
 use Mojo::JSON 'decode_json';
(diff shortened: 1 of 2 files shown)


coolo:
this is purely based on guess work, right?

needle.pm
```diff
@@ -52,33 +51,20 @@ sub new {
...
-    # - This code initializes $json->{file} so it contains the path relative to the lookup
-    #   directory and re-assigns $jsonfile to contain the absolute path.
-    # - A custom needle repo checkout referenced by the $bmwqemu::vars{NEEDLES_DIR} variable
-    #   is considered to support loading custom needles (e.g. with the openqa-clone-custom-git-refspec
-    #   script). The custom needle repo must be within the current working directory (usually the
-    #   openQA worker's pool directory).
-    # - Otherwise, the needle will be looked up under $bmwqemu::vars{PRJDIR}.
-    my @lookup_dirs;
-    my $needles_dir = $bmwqemu::vars{NEEDLES_DIR};
-    push(@lookup_dirs, $needles_dir) if (defined $needles_dir && index($needles_dir, cwd) == 0);
```
coolo: this cwd call is what kills us. The whole if is pointless due to:

```
isotovideo:$bmwqemu::vars{NEEDLES_DIR} //= "$bmwqemu::vars{PRODUCTDIR}/needles";
```

As such we should move this cwd there. If NEEDLES_DIR is set but doesn't point to cwd, overwrite it.

okurz:
> this is purely based on guess work, right?

@coolo no, bisected and tested with crosscheck, see
https://progress.opensuse.org/issues/57962#note-1

needle.pm
Martchus: makes sense, I'll test this tomorrow

Martchus: When I move the check to that assignment one can re-assign that variable later to sneak out from the working directory and load needles from everywhere. (With `cwd` this wasn't possible because the working directory change within autotest is not synced to the backend.)

So I should do that check likely in `needle::init` where we also check that the needle directory exists. Then `cwd` would only be executed once per reload.

okurz: sounds useful. I was already surprised by that inefficiency when we looked at the code together yesterday :)

PR #1225, opened 2019-10-11
Exclude known flaky test 13-osutils in OBS check

okurz:


Martchus:
Why not just remove the file? Wouldn't this execute the POD test multiple times?

okurz:
> Why not just remove the file?

I don't now, just doing the same as in before. Will try to delete them already in a followup.

> Wouldn't this execute the POD test multiple times?

Yes

PR #1226, opened 2019-10-11
Allow loading needles from current working directory

Martchus:
See https://progress.opensuse.org/issues/56789 and particular commit messages-

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -22,9 +22,11 @@ use autodie ':all';
 
 use File::Find;
 use File::Spec;
+use Mojo::File;
 use Mojo::JSON 'decode_json';
 use Cpanel::JSON::XS ();
 use File::Basename;
+use Try::Tiny;
 require IPC::System::Simple;
 use OpenQA::Benchmark::Stopwatch;
 
@@ -67,14 +69,13 @@ sub new {
     # $jsonfile contains absolute path within $bmwqemu::vars{PRJDIR}
 
     if (!$json) {
-        local $/;
-        open(my $fh, '<', $jsonfile);
-        $json = decode_json(<$fh>);
-        close($fh);
-        if (!$json || $@) {
-            warn "broken json $jsonfile: $@";
-            return;
+        try {
+            $json = decode_json(Mojo::File->new($jsonfile)->slurp);
         }
+        catch {
+            warn "broken json $jsonfile: $_";
+        };
+        return undef unless $json;
     }
 
     $self->{tags}       = $json->{tags}       || [];
--- a/needle.pm
+++ b/needle.pm
@@ -20,6 +20,7 @@ use strict;
 use warnings;
 use autodie ':all';
 
+use Cwd 'cwd';
 use File::Find;
 use File::Spec;
 use Mojo::File;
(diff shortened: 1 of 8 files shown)


Martchus:
The tests are broken and it seems that using `$needle::needlesdir` globally in the way I'm doing it with this change doesn't work. Apparently needles are loaded from a process which is not the main backend and not the main isotovideo and in that process the `$needle::needlesdir` variable is not set. This architecture is really confusing. I thought only isotovideo (initially) and the backend (reload command) would load needles.

coolo:
hmm, why would autotest load needles?

Martchus:
I wasn't aware of that, too. It does not *load* needles in the sense that it scans for them. But it creates needle objects and therefore needs to know the path of the needle directory. An example for that instantiation is `testapi::_handle_found_needle`. It looks like the needle instances are used to record needle results.

I solved the problem by simply doing the initial needle scan (when the needle directory is initialized) before starting autotest. Then it should work like it did before.

coolo:
ok. and how is this handled in developer mode now? Which repo is the needle generated in and which is reloaded?

Martchus:
I've actually just run the developer mode fullstack test locally and it passes. So there's at least no regression in the case no `NEEDLES_DIR` is provided.

> Which repo is the needle generated in

This is done by the web UI as usually when saving a needle within the needle editor and completely independent of these changes.

> which is reloaded?

I guess there's a problem with that, indeed. The needles are downloaded into `$bmwqemu::vars{NEEDLES_DIR}` but that might be undefined and is not validated. It should actually use `$needle::needlesdir` which would point to the custom needle repository if one has been specified and otherwise to the default. So it would download needles to the custom needle directory and reload that one.

Martchus:
I took care about the developer mode's needle downloader (see commit message). Unfortunately we don't have an integration test for this. The unit test for the needle downloaded didn't fail because it actually initialized `NEEDLES_DIR` and the fullstack test didn't fail because it only tests "the local case" (worker and web UI are on the same machine as share the same needle directory making downloads not necessary).

Martchus:
> This custom needle repository is only checked out temporarily in the pool dir of an openQA worker and not configured for git push. That does not make much sense.

Apparently you think needles are saved by os-autoinst. That is not correct. Needles are saved by the web UI. Since this PR is about os-autoinst my statement was not about saving needles.

Note that for saving/pushing (which is again handled by the web UI) it makes no difference whether the developer mode is running or not. Needles are always either saved to a local directory or pushed to the configured Git repo.

I was talking about the needle downloader within os-autoinst. That downloader has only one purpose: It makes needles which have already been *persistently* saved by the web UI *instantly* available to the currently running job. So it is fine if these needles are only downloaded to a temporary directory.

> What is the "default needle dir"?

When I read the previous code correctly (which in my opinion makes it harder to figure out what's going on than the version in this PR) the default needle directory was `"$bmwqemu::vars{PRODUCTDIR}/needles"`. Now the default directory is still `"$bmwqemu::vars{PRODUCTDIR}/needles"`. Previously it was verified whether the needle dir is within `$bmwqemu::vars{PRJDIR}`. Now it can also be within the current working directory (of the isotovideo process at the time needles are initialized the first time).

> Is it still working to have a needle dir in the main directory of the test distribution as well as below the product dir if specified?

Are you talking about this fallback at the beginning of `isotovideo`? `$bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};`

That's still in place and should work as before. Note that there's also another default set by the worker: `$vars{PRODUCTDIR} //= OpenQA::Utils::productdir($vars{DISTRI}, $vars{VERSION}, $shared_cache);`

---

About tests:

> 1. no custom needle dir, no product dir, no caching
> 2. no custom needle dir, product dir, no caching

The developer mode fullstack test covers one of these. The `PRODUCTDIR` is set during the test run but it is just equal to `CASEDIR`. So it is likely covering `1.`. I already explained that the last two bullet points make no difference and this PR is not in the openQA repo so it can not even touch that part. It would make more sense to have a bullet point "The needle downloader works.". So far the needle downloader is only tested by a unit test which simply checks that it downloads the needles into the currently set needle directory (regardless how that needle directory has been deduced).

> 3. same as 2 but with caching

The fullstack test within openQA tests caching but not the needle downloader. Saving the needles is (again) done by the web UI and caching has no influence on it. I hope that answers your question where the needles are saved in that case.

> 4. custom needle dir, no caching
> 5. custom needle dir, caching

These are likely not covered by any automatic tests. And again, all of this has no influence on saving needles...

needle.pm
```diff
@@ -51,30 +54,37 @@ sub new {
...
-    # $json->{file} contains path relative to $bmwqemu::vars{PRJDIR}
-    # $jsonfile contains absolute path within $bmwqemu::vars{PRJDIR}
+    # locate the needle's JSON file in one of the allowed parent directories
+    # - This code initializes $json->{file} so it contains the path relative to the lookup
+    #   directory and re-assigns $jsonfile to contain the absolute path.
+    # - The needle must be within the needle directory ($needledir) or the project directory.
+    # - The needle directory might be within the current working directory to support loading custom
+    #   needles (e.g. with the openqa-clone-custom-git-refspec script).
+    my @allowed_parent_dirs = ($needledir, $bmwqemu::vars{PRJDIR});
+    for my $allowed_parrent_dir (@allowed_parent_dirs) {
```
okurz: yuk, I overlooked this typo :D

okurz: https://github.com/os-autoinst/os-autoinst/pull/1231

PR #1227, opened 2019-10-11
spec: Do not duplicate OBS-incompatible checks

okurz:
OBS test in https://build.opensuse.org/package/show/home:okurz:no_duplicate_checks/os-autoinst


okurz:
@coolo?

PR #1232, opened 2019-10-17
Allow consoles to persist over reset

czerw:
Fix poo#57329: PowerVM console is connected and disconnected during
test execution, but it means that not all console messages are
captured. Console should be connected all the time and we need to
skip reset for it.

Related PR in distri: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8718
Verification run: http://10.100.12.105/tests/3348

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -567,7 +567,7 @@ sub reset_consoles {
 
     # we iterate through all consoles
     for my $console (keys %{$testapi::distri->{consoles}}) {
-        #next if ($console eq 'x3270');
+        next if $self->console($console)->{args}->{persistent};
         $self->reset_console({testapi_console => $console});
     }
     return;
--- a/backend/spvm.pm
+++ b/backend/spvm.pm
@@ -46,9 +46,10 @@ sub do_start_vm {
         'novalink-ssh',
         'ssh-xterm',
         {
-            hostname => get_required_var('NOVALINK_HOSTNAME'),
-            password => get_required_var('NOVALINK_PASSWORD'),
-            username => get_var('NOVALINK_USERNAME', 'root')});
+            hostname   => get_required_var('NOVALINK_HOSTNAME'),
+            password   => get_required_var('NOVALINK_PASSWORD'),
+            username   => get_var('NOVALINK_USERNAME', 'root'),
+            persistent => 1});
     $ssh->backend($self);
 
     return {};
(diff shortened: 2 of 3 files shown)


backend/baseclass.pm
```diff
@@ -568,6 +568,7 @@ sub reset_consoles {
     # we iterate through all consoles
     for my $console (keys %{$testapi::distri->{consoles}}) {
         #next if ($console eq 'x3270');
```
okurz: I think this line is now obsolete and can be deleted. It's a very unobvious "TODO" for what you are actually doing in this PR ;)

backend/baseclass.pm
```diff
@@ -568,6 +568,7 @@ sub reset_consoles {
...
+        next if $self->console($console)->{args}->{skip_console_reset};
```
okurz: as we are already in the console what about naming the attribute just `skip_reset`? Or maybe it's better to describe the nature of the console, not what should be done with, like, `persistent`?

czerw: OK, `persistent` seems reasonable. Just thinking about documentation update. But existing http://open.qa/api/testapi/#_add_console is quite generic with `console parameters are console specific` without any other reference. Is it worth to mention that `persistent` is on of possible specific parameters?

czerw: Updated `testapi.pm` doc section.

backend/baseclass.pm
czerw: done

backend/baseclass.pm
okurz: Great that you found this place for documentation :) I did not ask you to add it as I also saw that `add_console` is just forwarding the hash of potential options

backend/baseclass.pm
```diff
@@ -567,7 +567,7 @@ sub reset_consoles {
 
     # we iterate through all consoles
     for my $console (keys %{$testapi::distri->{consoles}}) {
-        #next if ($console eq 'x3270');
+        next if $self->console($console)->{args}->{persistent};
```
okurz: IMHO this is the important part that should be emphasized in your git commit message subject line as well as the PR subject. You are not just adding the option, you are adding support for persistent consoles. How about instead of "Add persistent option to console specific parameters" you write "Allow consoles to persist over reset"?

Soulofdestiny:
LGTM

PR #1234, opened 2019-10-18
Remove PRJDIR, support loading needles from working directory

Martchus:
* Requires https://github.com/os-autoinst/openQA/pull/2410
* See https://progress.opensuse.org/issues/56789 and especially https://progress.opensuse.org/issues/56789#note-18 for the PRJDIR removal


autotest.pm
```diff
@@ -54,7 +54,10 @@ loadtest is called.
 sub find_script {
     my ($script) = @_;
     my $casedir = $bmwqemu::vars{CASEDIR};
-    my $script_override_path = join('/', $bmwqemu::vars{PRJDIR} // '', 'factory/other', $script);
+    my $script_override_path = join('/', 'factory/other', $script);
+    # FIXME: This was previously relying on PRJDIR. Find a better way to reference a custom script
+    #        provided as an asset. This feature is likely superseded by openqa-clone-custom-git-refspec
+    #        anyways.
```
okurz: There is still a valid use case. What I had in mind is the following: Imagine a discussion in a bug investigation when someone suggests "Could you please call these three lines of code in the openQA code to try out something or gather some data". Creating a fork of the test distribution and reference that in openqa-clone-custom-git-refspec is an additional hurdle that many are not willing to take. See https://github.com/os-autoinst/os-autoinst/pull/1019 and http://open.qa/docs/#_schedule_asset_nr_url for the original idea.

isotovideo
```diff
@@ -300,6 +296,10 @@ if ($bmwqemu::vars{_EXIT_AFTER_SCHEDULE}) {
 }
 testapi::init();
 
+# make sure the needles are initialized
+my $needles_dir = needle::init;
+$bmwqemu::vars{NEEDLES_GIT_HASH} = OpenQA::Isotovideo::Utils::calculate_git_hash($needles_dir);
```
okurz: looks like you can move the temporary variable just inside the `()`

autotest.pm
Martchus: I simply made it use `$bmwqemu::vars{ASSETDIR}` which is populated by the worker. That makes more sense than the previous code anyways because the logic for locating assets (and possibly caching them) is on the worker-side.

Martchus:
> In general I wonder if there is a strict need for supporting a needledir which are not subdirs of casedir.

I wouldn't add further restrictions unless it helps us to simplify the code. So far os-autoinst does not checkout custom needle repos within the casedir itself so we needed to adjust that.

okurz:
Could you state a bit more details about how you tested this?

Martchus:
I tested this by running some tests locally (also the automated fullstack tests). However, I've noticed that locating needles still doesn't work on the openQA-side with https://github.com/os-autoinst/openQA/pull/2410 so this isn't actually ready. Note that the PR for the openQA-side needs to be merged and deployed before this one.

okurz:
please update the API versions then and mark this PR here as [wip] to prevent in from being merged when it is depending on new openQA versions.

Martchus:
This change requires an addition to the API on the openQA-side but that API has no versioning. (We would only theoretically increase the API version when doing changes which are not backwards compatible but that doesn't help here.)

Martchus:
The web UI part should have been deployed in production so I'd say this is ready to be merged.

okurz:
Given the breakage we have seen lately in this area I recommend you test a little bit more, e.g. on our staging instance, with old and new openQA, caching, git checkout, nested needle dirs, etc. I guess it also makes sense to extend the automatic tests simulating these setups a bit better

Martchus:
The staging worker might still be used by @kraih. But I can tests these things locally, too.

kraih:
Yes, i just started deploying my latest changes to the staging machines.

Martchus:
Note that "old openQA" is not supported. Hence I've previously added the "not-yet" label. I resolved the compatibility problem only on the openQA-side which should be sufficient (the new openQA support both os-autoinst versions).

okurz:
@Martchus we recently did a partial revert in 7dfede39f , it's still compatible this way?

Martchus:
It should be. I'm going to do some local tests anyways and of course I'm testing against openQA with the partial revert.

PR #1235, opened 2019-10-18
Rely on QEMUMACHINE default in OFW

okurz:


asdil12:
I think we should go with this approach insted #1236 as it is more powerful.

PR #1237, opened 2019-10-20
spec: Fix missing, additional runtime requirements

okurz:
Missing requirements encountered when experimenting with os-autoinst
installations in clean container environments.


dist/rpm/os-autoinst.spec
```diff
@@ -35,7 +35,7 @@ Source0:        %{name}-%{version}.tar.xz
 %define test_requires %build_requires %requires perl(Perl::Tidy) perl(Test::Compile) perl(Test::Exception) perl(Test::Output) perl(Test::Fatal) perl(Test::Warnings) perl(Pod::Coverage) perl(Test::Pod) perl(Test::MockModule) perl(Test::MockObject) perl(Devel::Cover) perl(Test::Mock::Time) qemu-tools
 %define devel_requires %test_requires %requires_not_needed_in_tests
 BuildRequires:  %test_requires
-Requires:       %requires
+Requires:       %requires %requires_not_needed_in_tests
```
coolo: why would we require optipng? and git is optional/recommended

Martchus: I had the same thought. optipng is only executed on the worker so the dependencies belongs to the openQA-worker package.

And likely even qemu is optional considering it is not the only backend.

okurz: Good observation about optipng. I have removed optipng as it's properly covered in openQA worker. https://github.com/os-autoinst/os-autoinst/pull/1205#issuecomment-541054999 mentions qemu-img missing hence me trying to fix it. Should we move qemu into "Recommends:"?

About git, it is actually required to prevent the error

```
Can't exec "git": No such file or directory at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 29.
Use of uninitialized value $git_hash in scalar chomp at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 29
```

when trying to calculate the git hash from the loaded test or needle directory.

PR #1239, opened 2019-10-21
Fix svirt backend's 100 % CPU usage

Martchus:
See the particular commit messages and https://progress.opensuse.org/issues/33529.

---

So far I tested this only locally using svirt/QEMU (see https://github.com/Martchus/openQA-helper#testrun-svirt-backend-locally). I started a job and paused it when the SUT was running using the developer mode. Without this change isotovideo's backend process fully occupies one CPU core. With this change it drops to 1 % when idling. The warning about using `readline()` in async mode is gone, too.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1222,20 +1222,34 @@ sub start_ssh_serial {
 }
 
 sub check_ssh_serial {
-    my ($self, $fh) = @_;
+    my ($self, $fh, $write) = @_;
 
-    if ($self->{serial} && $self->{serial}->sock == $fh) {
-        my $chan = $self->{serial_chan};
-        my $line = <$chan>;
-        if (defined $line) {
-            print $line;
-            open(my $serial, '>>', $self->{serialfile});
-            print $serial $line;
-            close($serial);
-        }
+    my $ssh = $self->{serial};
+    return 0 unless $ssh;
+    my $ssh_socket = $ssh->sock;
+    return 0 unless $ssh_socket == $fh;
+
+    if ($write) {
+        bmwqemu::diag('SSH serial: setup error: socket has been wrongly selected for writing');
         return 1;
     }
-    return;
+
+    # read from SSH channel (receiving extended data channel as well via `$chan->ext_data('merge')`)
+    my $chan = $self->{serial_chan};
+    my $buffer;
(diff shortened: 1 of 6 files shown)


Martchus:
This somehow breaks `start_serial_grab`.

backend/baseclass.pm
```diff
@@ -1211,37 +1216,52 @@ sub start_ssh_serial {
...
+    return 0 unless $ssh;
+    my $ssh_socket = $ssh->sock;
+    return 0 unless $ssh_socket == $fh;
 
-    if ($self->{serial} && $self->{serial}->sock == $fh) {
-        my $chan = $self->{serial_chan};
-        my $line = <$chan>;
-        if (defined $line) {
-            print $line;
+    if ($write) {
```
kalikiana: This method fails whenever `$write` is true but doesn't actually check anything here... I gather this error should move to the callsite?

Martchus:
The commit 'Make start_serial_grab blocking' allows my local tests to come a little bit further (`$chan->exec($cmd)` no longer fails). However, I still couldn't conduct a successful test run locally with any of the different svirt variants. But without my changes I run into the same errors. So I simply don't know how to verify this change.

backend/baseclass.pm
Martchus: It checks for the existence of the SSH session and whether its socket matches `$fh`. In this case it returns `0` for "it wasn't that file descriptor". In the write case I made it return `1` because `$fd` matches and the caller shouldn't check anywhere else.

Martchus:
I now managed to run VMWare and s390x tests locally. They pass and the CPU usage is low and the warning `Calling Net::SSH2::Channel::readline in non-blocking mode is usually a programming error at /hdd/openqa-devel/repos/os-autoinst/backend/baseclass.pm line 1223.` no longer displayed.

okurz:
Can you please verify s390x z/VM as well as s390x kvm tests that include rebooting and shutdown where the ssh serial is started/stopped?

coolo:
I'm confident that this got enough testing to be deployed today

okurz:
too late for osd today but we can always go further with auto-deployment ;)

coolo:
I'd sneak in a manual grenache-1 deployment

PR #1240, opened 2019-10-22
Allow tidy to run only over local changes

foursixnine:
While running tidy doesn't take too long, when writing code it's often an annoyance
to run tidy and wait a long time until all files are checked, with this only the modified
files will be checked, deleted files are also seen, but tidy skips them as they don't exist
which is good enough for local development.

The main benefit from this would be @os-autoinst/tests-maintainer

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -10,6 +10,12 @@ if test "$1"  = '--check'; then
     check=1
 fi
 
+onlychanged=
+if test "$1"  = '--only-changed'; then
+    shift
+    onlychanged=1
+fi
+
 if ! which perltidy > /dev/null 2>&1; then
     echo "No perltidy found, install it first!"
     exit 1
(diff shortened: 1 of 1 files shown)


Martchus:
We have already the same in openQA but the option is called `--only-changed`. I would prefer to use `--only-changed` here as well for consistency. Besides, `--justlocal` is a bit meaningless as it misses the word "changes" (`--justlocalchanges`).

foursixnine:
> We have already the same in openQA but the option is called `--only-changed`. I would prefer to use `--only-changed` here as well for consistency. 

I'm with you here :)

foursixnine:
@Martchus Done, can you give it a second look? :) Also, fixed the globbing thing that was only taking the first file... now it takes all pm, and pl :)

Martchus:
It doesn't include tests. However, the globbing feature of `git status` seems quite limited so not sure how to extend the pattern.

foursixnine:
> However, the globbing feature of `git status` seems quite limited so not sure how to extend the pattern.

Not sure what you mean, but aside from `.t` files, it includes the same pattern that is in L47

Martchus:
I mean `.t` should be included but it is not possibly with the globbing provided by Git (at least I wouldn't know how). So I'd resort to using grep: https://github.com/os-autoinst/os-autoinst/pull/1241

PR #1243, opened 2019-10-25
Remove status API call

perlpunk:
The status API call is not used anymore since:
https://github.com/os-autoinst/os-autoinst/pull/1212

It is now writing a status file.

The corresponding PR in openQA to use the status file:
https://github.com/os-autoinst/openQA/pull/2327

Issue: https://progress.opensuse.org/issues/39845

Once openQA is deployed this can be merged.

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -232,7 +232,7 @@ sub isotovideo_command {
 
 sub isotovideo_get {
     my ($c) = @_;
-    return isotovideo_command($c, [qw(status version)]);
+    return isotovideo_command($c, [qw(version)]);
 }
 
 sub isotovideo_post {


okurz:
IMHO we have the corresponding openQA change deployed on both o3 as well as osd. Care to retest?

okurz:
*bump*

perlpunk:
I rebased to master.
There have been some issues in openQA where it wasn't sure if the status call removal was responsible for that. Are they resolved?

okurz:
good point. How about you test that :) Maybe an opportunity for you to get accustomed with the staging environment?

perlpunk:
> good point. How about you test that

I don't know. What exactly were the issues?

perlpunk:
Seems to me like all issues have been resolved.

perlpunk:
I ran all openQA tests with this, yes.

PR #1245, opened 2019-10-25
Avoid warning in comparison; num_queues might be undef

perlpunk:
Issue: https://progress.opensuse.org/issues/57719

There are warnings like that in `autoinst-log.txt`:
```
Use of uninitialized value in numeric ne (!=) at /usr/lib/os-autoinst/OpenQA/Qemu/DriveDevice.pm line 124.
```

### DIFF
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -121,7 +121,7 @@ sub gen_cmdline {
         # Configure bootindex only for first path
         $self->_push_ifdef(\@params, 'bootindex=',  $self->bootindex) if (!$path->id || $path->id eq 'path0');
         $self->_push_ifdef(\@params, 'serial=',     $self->serial);
-        $self->_push_ifdef(\@params, 'num_queues=', $self->num_queues) if ($self->num_queues != -1);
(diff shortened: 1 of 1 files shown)


OpenQA/Qemu/DriveDevice.pm
```diff
@@ -121,7 +121,7 @@ sub gen_cmdline {
         # Configure bootindex only for first path
         $self->_push_ifdef(\@params, 'bootindex=',  $self->bootindex) if (!$path->id || $path->id eq 'path0');
         $self->_push_ifdef(\@params, 'serial=',     $self->serial);
-        $self->_push_ifdef(\@params, 'num_queues=', $self->num_queues) if ($self->num_queues != -1);
+        $self->_push_ifdef(\@params, 'num_queues=', $self->num_queues) if ($self->num_queues and $self->num_queues != -1);
```
Martchus: Maybe `defined $self->num_queues` would be better for not changing the behavior. (The previous version would have added the param for 0.)

perlpunk: oh right!

perlpunk: I did it like that now:
`if (($self->num_queues // -1) != -1);`

force-pushed

PR #1246, opened 2019-10-29
Revert "Allow read_json() to return multiple results"

Martchus:
This reverts commit ec50b8785ff35be3c76f5323708a34ad64fd3d0e which causes problems in production: https://progress.opensuse.org/issues/58823

@perlpunk I've prepared this revert because the problem might not be trivial to fix but we likely want to have it soon working in production again.


perlpunk:
#1248 should fix the problem

PR #1248, opened 2019-10-29
myjsonrpc: Go back to incremental parsing

perlpunk:
The strategy to look for the next newline was too slow.

The time for `read_json()` increased to about 4-5 times with the changes introduced in #1230

https://progress.opensuse.org/issues/58823 "os-autoinst is too slow pressing F2 causing ARM tests to fail in "boot_to_desktop"

The `write_json` seemed to fail sometimes because the read took too long, resulting in test failures.


okurz:
Thanks for the quick merge and the added details in the commit message

PR #1249, opened 2019-10-30
Increase the read buffer for JSON RPC

coolo:
As we learned yesterday the JSON can be easily several megabytes, so reading them by pages doesn't really make sense as we just keep the CPU busy trying to parse JSON. So try to read all of it in one go

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -24,6 +24,7 @@ use Mojo::JSON;    # booleans
 use Cpanel::JSON::XS ();
 
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
+use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8000000;
 
 sub send_json {
     my ($to_fd, $cmd) = @_;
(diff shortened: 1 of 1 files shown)


okurz:
Without further testing I can't see if this is really good or not

coolo:
Not sure I asked

myjsonrpc.pm
```diff
@@ -120,7 +120,7 @@ sub read_json {
         }
 
         my $qbuffer;
-        my $bytes = sysread($socket, $qbuffer, 8000);
+        my $bytes = sysread($socket, $qbuffer, 8000000);
```
perlpunk: I would use a perl constant here. Maybe even modifiable via an environment variable, so testing different sizes is easier.

okurz:
> Not sure I asked

Well, I assume an open pull request means you implicitly ask the people with merge permissions to review.

coolo:
> > Not sure I asked
> 
> Well, I assume an open pull request means you implicitly ask the people with merge permissions to review.

Yeah. But I don't consider "I can't say" a review

okurz:
Well, I "*can* say" on many other PRs :) So this one is missing the information I would need to call it ok easily.

myjsonrpc.pm
coolo: done

okurz:
merged locally

PR #1250, opened 2019-10-30
script_output() - Add code comment for commit 72efc09b

cfconrad:
@okurz WDYT?

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -239,7 +239,10 @@ sub script_output {
         bmwqemu::log_call("Content of $script_path :\n \"$cat\" \n");
         testapi::type_string($cat . "\n");
         testapi::wait_serial("$cat", no_regex => 1, quiet => $args{quiet});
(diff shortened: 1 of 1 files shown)


distribution.pm
```diff
@@ -239,6 +239,9 @@ sub script_output {
         bmwqemu::log_call("Content of $script_path :\n \"$cat\" \n");
         testapi::type_string($cat . "\n");
         testapi::wait_serial("$cat", no_regex => 1, quiet => $args{quiet});
+        # Wait for input prompt of here tag before typing $script. This avoid
```
okurz: avoid**s**

cfconrad: thx

okurz:
Adding the comment unfortunately demands a style change now, see https://travis-ci.org/os-autoinst/os-autoinst/builds/604891540#L839

cfconrad:
thx for the pointer. Somehow I overseen it.

okurz:
nice, thanks!

PR #1252, opened 2019-11-05
myjsonrpc: Improve debugging

perlpunk:
output json_cmd_token to see if the message that was sent is also read

Related to https://progress.opensuse.org/issues/59100


perlpunk:
I will add a test for this

perlpunk:
Also removed a leftover line
`local $myjsonrpc::DEBUG_JSON = 1` from the previous implementation as a package var

PR #1253, opened 2019-11-05
Mojo::Reactor: stop/start reactor around reading

perlpunk:
If time passes after the read_json() then subsequent messages to
the socket are not triggering the io callback again.
This can be reproduced by adding a `sleep 3` after the read_json and running
full-stack.t in openQA.

stopping and starting the reactor at the start/end of the callback
fixes it.

Issue: https://progress.opensuse.org/issues/59100

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -324,6 +324,8 @@ sub run_daemon {
     # process json messages from isotovideo
     Mojo::IOLoop->singleton->reactor->io($isotovideo => sub {
             my ($reactor, $writable) = @_;
+            # Stop reactor as long as we are reading from the socket here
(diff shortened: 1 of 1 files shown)


perlpunk:
Not sure how I could add a test for this.

kraih:
I don't understand this change, stopping and restarting the reactor should do nothing here. If it in fact has an effect it might be hiding a real bug somewhere else.

kraih:
Oh wait, i missed something, the `$reactor->start` would be a blocking operation there. That changes the callstack completely and you are recursing. That can't be right.

PR #1256, opened 2019-11-07
Makefile: Check for spellcheck requirements

okurz:


okurz:
I would like to check if the spellcheck now fails in travis CI as it should as the spell check requirements are not installed. Then I can apply a fix

okurz:
fails in https://travis-ci.org/os-autoinst/os-autoinst/builds/608612482#L961 as spell is not available, that is good. https://github.com/os-autoinst/os-autoinst/pull/1257 for a fix, in a separate PR as we need the updated container image based on spec file changes first.

okurz:
test build in https://build.opensuse.org/package/show/home:okurz:fix:spellcheck/os-autoinst

okurz:
This needs an updated https://build.opensuse.org/package/show/devel:openQA/openqa_dev based on Leap 15.1 to provide spellcheck requirements as we probably want to run spellchecks in PR checks.

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1271#issuecomment-556009996 shows that the new image looks fine. Created https://github.com/os-autoinst/openQA/pull/2529 . We are still missing podspell though it seems.

PR #1257, opened 2019-11-07
spec: Add missing spellcheck requirements

okurz:


okurz:
I hoped you could set up the pull request builder in time :)

There you go: https://build.opensuse.org/project/show/home:okurz:fix:spellcheck2 (still building)

okurz:
this already shows the problem of missing deps for older products, e.g. SLE and such.

okurz:
Excluded the spellcheck packages for SLE where we do not have it: https://build.opensuse.org/package/show/home:okurz:fix:spellcheck2/os-autoinst . All green

PR #1258, opened 2019-11-07
Read all available bytes in check_ssh_serial

Martchus:
Maybe check_ssh_serial is not in any case invoked as long as there are bytes available to read from the socket. This change makes check_ssh_serial no longer rely on that and possibly fixes https://progress.opensuse.org/issues/59190.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -33,7 +33,7 @@ use bmwqemu;
 use IO::Select;
 require IPC::System::Simple;
 use myjsonrpc;
-use Net::SSH2;
+use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
 use OpenQA::Benchmark::Stopwatch;
 use MIME::Base64 'encode_base64';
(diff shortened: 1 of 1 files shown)


okurz:
A bit many "maybe's" ;) I retriggered https://travis-ci.org/os-autoinst/os-autoinst/builds/608857981?utm_source=github_status&utm_medium=notification to crosscheck if the coverage will be affected

Martchus:
Nice that it works. Hot-patching the production worker is of course also a way to test this.

By the way, as @coolo mentioned on Rocket Chat: The problem is that this code is using buffered IO so `IO::Select` won't work unless one reads everything. The previous code (before my change to reduce the CPU usage) just read one line and only worked because the method was called constantly regardless whether there were bytes available to read (hence the high CPU usage).

okurz:
We tested further internal jobs on a hotpatched production worker and we are confident that also other backends and test scenarios are not badly influenced.

PR #1259, opened 2019-11-08
send_key: limit rate with VNC_TYPING_LIMIT

ggardet:
- ticket: https://progress.opensuse.org/issues/55703

Update `send_key` to respect `VNC_TYPING_LIMIT`, which required update on `map_and_send_key`.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -20,6 +20,8 @@ use Try::Tiny;
 use Scalar::Util 'blessed';
 use OpenQA::Exceptions;
 
+use constant DEFAULT_KEY_PRESS_RELEASE_DELAY_US => 20_000; # default to 20 ms
+
 __PACKAGE__->mk_accessors(
     qw(hostname port username password socket name width height depth
       no_endian_conversion  _pixinfo _colourmap _framebuffer _rfb_version screen_on
@@ -723,7 +725,7 @@ sub init_ikvm_keymap {
 sub map_and_send_key {
     my ($self, $keys, $down_flag, $press_release_delay_us) = @_;
 
-    $press_release_delay_us //= 2_000;    # default to 2 ms
+    $press_release_delay_us //= DEFAULT_KEY_PRESS_RELEASE_DELAY_US;
 
     if ($self->ikvm) {
         $self->init_ikvm_keymap;
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -721,7 +721,9 @@ sub init_ikvm_keymap {
 
 
 sub map_and_send_key {
-    my ($self, $keys, $down_flag) = @_;
+    my ($self, $keys, $down_flag, $press_release_delay_us) = @_;
+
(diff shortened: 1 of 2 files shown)


consoles/VNC.pm
```diff
@@ -755,15 +757,15 @@ sub map_and_send_key {
     if (!defined $down_flag || $down_flag == 1) {
         for my $key (@events) {
             $self->send_key_event_down($key);
+            usleep($press_release_delay_us);
         }
     }
-    usleep(2_000);
```
coolo: danger mouse! you only sleep between combined keys and no longer between down and up of single keys.

And take the chance to raise the default delay - for all we know a little slower typing makes testing more stable at an acceptable price.

ggardet: So, you want to:
* send X keys down at a time
* wait a bit
* send X keys up at a time
?
Not sure it will really enhance things.

If we wait between each key, I think it should be more effective. But I cannot get why it could break. On my machine I can press `alt` wait 10 seconds, press `v` wait 10 seconds, release `v`, wait 10 seconds, release `alt` and all works as ecpected.

ggardet:
With this update, I tested successfully locally on aarch64 with Tumbleweed virtualization test where `virtman_view` usually fails because `alt-v` is not working as expected.

consoles/VNC.pm
coolo: No, I want to wait after every action -and one extra time before the key up actions.

ggardet: It is not a big deal to add this extra time, but not sure why it would be needed.

ggardet: Added it back now.

ggardet:
I think it is now ready to give it a try in production.

coolo:
the question is anyway if VNC needs to have a default or can't simply require the parameter

ggardet:
> Please don't duplicate the default - e.g. use a constant in VNC. And I repeat: increase the default, I'd say 20ms is still fast enough

Done.

consoles/VNC.pm
```diff
@@ -20,6 +20,8 @@ use Try::Tiny;
 use Scalar::Util 'blessed';
 use OpenQA::Exceptions;
 
+use constant default_key_press_release_delay_us => 20_000; # default to 20 ms
```
coolo: constants are normally all upper case

ggardet: Fixed.

ggardet:
Not sure why the test fail.

coolo:
yeah, if you follow the test running you'll see that it types slower than me now :)

I'll play with the values I guess

coolo:
Additionally the text is mismatching the md5 if typed slower

coolo:
See https://github.com/os-autoinst/os-autoinst/pull/1260 for my spin of the story

okurz:
#1260 merged. That should include all intended changes. Feel free to open another PR for further changes and thank you for adressing this problem!

PR #1260, opened 2019-11-09
Follow VNC_TYPING_LIMIT everywhere (including send_key)

coolo:
Also lower the default of it, systems are just not tested with that
fast typing speed and lead to unrealistic failures. Fast human typers
can reach 40-50 *words per minute*, which would translate to ~3 keys
per second (a word is averaged to be 5 keys in these type speed contexts).

So trying 50 keys per second is very likely to fail tests that wouldn't
be reproducible in real life. So let's lower this to ten times as fast
as human - and test scenarios that require super fast typing can
increase the VNC_TYPING_LIMIT

Co-Authored-By: Guillaume Gardet <guillaume.gardet@arm.com>

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -23,10 +23,14 @@ use feature 'say';
 use base 'consoles::network_console';
 
 use consoles::VNC;
-use Time::HiRes qw(usleep gettimeofday);
+use List::Util 'max';
+use Time::HiRes qw(usleep time);
 
 use Try::Tiny;
 use testapi 'get_var';
+
+# speed limit: 30 keys per second
+use constant VNC_TYPING_LIMIT_DEFAULT => 30;
 
 sub screen {
     my ($self) = @_;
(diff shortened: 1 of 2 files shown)


consoles/vnc_base.pm
```diff
@@ -130,32 +130,35 @@ sub type_string {
             "\e" => 'esc'
         };
         $letter = $charmap->{$letter} || $letter;
-        $self->{vnc}->map_and_send_key($letter);
-        $self->{backend}->run_capture_loop($seconds_per_keypress);
+        # 25% is spent hitting the key, 25% releasing it, 50% seaching the next key
```
okurz: seaching?

coolo: https://context.reverso.net/%C3%BCbersetzung/englisch-deutsch/seach

consoles/vnc_base.pm
```diff
@@ -23,11 +23,15 @@ use feature 'say';
...
 use consoles::VNC;
-use Time::HiRes qw(usleep gettimeofday);
+use List::Util 'max';
+use Time::HiRes qw(usleep time);
 
 use Try::Tiny;
 use testapi 'get_var';
 
+# speed limit: 30 keys per second
+use constant VNC_TYPING_LIMIT_DEFAULT => 30;
```
okurz: Would you mind putting your PR description comment rather in here to explain the value selection? And are you sure we want to loose the speed benefit when "mistyping" now seems to happen only if either modifier keys are involved or it's a clearly special product, e.g. the bootloader line.

consoles/vnc_base.pm
okurz: we "search" for the key? Is it a "keyboard from hell" that dynamically changes the layout after each key press? ;)

consoles/vnc_base.pm
coolo: I'm not so sure about this anymore. 30 keys a second is still very fast - and we hopefully see less mistypings on e.g. gnome terminals the more realistic our input is.

And I don't believe in code comments to explain numbers - they outdate super fast and the git history is so much more reliable.

coolo:
I don't like splitting the commit

okurz:
fine

consoles/vnc_base.pm
coolo: two finger key hunters reach up to 30wbm!

PR #1267, opened 2019-11-13
qemu: Fix MAKETESTSNAPSHOTS with test modules in nested dirs

okurz:
When MAKETESTSNAPSHOTS is used with test modules which reside in
multiple layers of directories, e.g. "openqa/tests/install" the
necessary directory structure would not be created and snapshot creation
would fail. By ensuring the prerequisite parent directories on creation
of the snapshot files this can be fixed.

Tested locally with

```
openqa-clone-job https://openqa.opensuse.org/tests/1080059 MAKETESTSNAPSHOTS=1
```

cloning a job from the test scenario
"openqa-Tumbleweed-dev-x86_64-Build:TW.3350-openqa_install+publish@64bit-2G"
based on the test distribution
https://github.com/os-autoinst/os-autoinst-distri-openQA/

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -451,7 +451,7 @@ sub save_snapshot {
 
     mkpath(VM_SNAPSHOTS_DIR);
     $self->_migrate_to_file(
-        filename         => VM_SNAPSHOTS_DIR . '/' . $snapshot->name,
+        filename         => File::Spec->catfile(VM_SNAPSHOTS_DIR, $snapshot->name),
         compress_level   => $vars->{QEMU_COMPRESS_LEVEL} || 6,
         compress_threads => $vars->{QEMU_COMPRESS_THREADS} // $vars->{QEMUCPUS},
         max_bandwidth    => $vars->{QEMU_MAX_BANDWIDTH});
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -22,6 +22,7 @@ use autodie ':all';
 
 use base 'backend::virt';
 
+use File::Basename 'dirname';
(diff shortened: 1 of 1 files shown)


backend/qemu.pm
```diff
@@ -186,8 +187,10 @@ sub open_file_and_send_fd_to_qemu {
     my ($self, $path, $fdname) = @_;
     my $rsp;
 
+    mkpath(dirname($path));
```
Martchus: As far as I know this won't throw an exception if it fails (in contrast to `File::Path::make_path` or `Mojo::File::make_path`). So either check the return value or use the one of the other functions.

backend/qemu.pm
```diff
@@ -186,8 +187,10 @@ sub open_file_and_send_fd_to_qemu {
...
     my $fd = POSIX::open($path, POSIX::O_CREAT() | POSIX::O_RDWR());
-    die "Failed to open $path: $!" unless (defined $fd);
+    use Carp;
```
Martchus: I'd rather add the `use` at the top of the file. It will be evaluated at compile time anyways.

backend/qemu.pm
okurz: it's covered by "autodie"

backend/qemu.pm
okurz: oops. That was only for temporary debugging. Removed the change.

backend/qemu.pm
Martchus: The fact that such an intrusive module exists seems insane but it is Perl after all so I'm not surprised.

PR #1268, opened 2019-11-15
Fix launching vncviewer within fullstack test

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/31501


okurz:
please check coverage, at least retrigger tests if you like and see if the number increases

okurz:
Please do not use contractions in git commit message subject lines

okurz:
I also tested locally how the shutdown procedure looks like and I can confirm that it's not worse than before – still a bit ugly and slow but ok-ish ;)

PR #1270, opened 2019-11-19
Allow unsetting 'fatal' test flag without snapshot support

Martchus:
This enables the workflow described in the following ticket:
https://progress.opensuse.org/issues/58166

However, it does *not* change the default behavior. So when tests
without snapshot support fail they are still stopped by default. Only
when `fatal => 0` is set via `test_flags` the job will continue. Of
course no rollback to previous snapshots is done.

Note that this also affects the QEMU backend when snapshots are not
available (e.g. `QEMU_DISABLE_SNAPSHOTS` has been set).

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 
 use Test::More;
-use Test::Output 'stderr_like';
+use Test::Output qw(stderr_like combined_from);
 use Test::Fatal;
 use Test::MockModule;
 use File::Basename ();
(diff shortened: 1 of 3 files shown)


Martchus:
Extended the unit tests.

t/08-autotest.t
```diff
@@ -200,30 +205,50 @@ is($completed, 1, 'unimportant test failure should complete');
...
+($died, $completed) = get_tests_done;
+is($died,      0, 'non-fatal serial failure test should not die');
+is($completed, 0, 'non-fatal serial failure test should not complete by default without snapshot support');
+@sent = [];
+
+$mock_basetest->mock(test_flags => sub { return {fatal => 0}; });
+stderr_like(sub { autotest::run_all }, qr/Snapshots are not supported/, 'run_all outputs status on stderr');
+($died, $completed) = get_tests_done;
+is($died,      0, 'non-fatal serial failure test should not die');
+is($completed, 1, 'non-fatal serial failure test should complete with {fatal => 0} without snapshot support');
```
okurz: would you mind to change the test messages to be unambiguous? Actually I would not even check for qr/Snapshots are not supported/ when this is not expected to change.

However I tried out just

```
autotest::run_all;
```

to see all output and I could see multiple occurences of "Loading a VM snapshot lastgood" which is odd when just before it reports "Snapshots are not supported" so I think we need to adapt the code to distinguish between "rollback module" and "loading snapshot".

Martchus: > when this is not expected to change

I thought the idea of tests is to check whether the code does what I expect it to do.

> so I think we need to adapt the code to distinguish between "rollback module" and "loading snapshot".

I'm afraid the entire test deserves an overhaul - it wasn't particularly easy to figure out how to extend it. And yes, the mocking might be inconsistent (in the sense that the present combination of variable values should not occur in practice). But as far as I see the test is still correct and verifies that the feature works.

okurz: Ok but I still assume "snapshots are not supported" in combination with "loading snapshot" is something that most likely your code changes triggered. How about that?

Martchus: Yes, because before we never ran without snapshot support. I've changed it to clean the state from previous runs so it doesn't think there's still a milestone saved. I still like the explicit check on the output to be sure the test mocking actually works as intended and also added a check that "Loading a VM snapshot" not occurs.

okurz: perfect!

PR #1271, opened 2019-11-20
t: Split tests in Makefile for easier diff handling

okurz:


okurz:
alright. So this shows that the new openSUSE Leap 15.1 based container image seems to be fine, e.g. https://travis-ci.org/os-autoinst/os-autoinst/builds/614512003#L662 shows "perl 5.26" so I assume it is not the old "perl 5.18" anymore ;)

PR #1272, opened 2019-11-20
Improve example in documentation for CASEDIR

Martchus:


okurz:
I suggest to close the PR again unless you want to specify the caveats instead of changing the URL.

Martchus:
Ok, I really thought it wouldn't because that's not an URL with scheme (and our code aborts then via `return undef unless $url->scheme;`). But the scheme is actually `git@github.com` here so it should generally work. Still, it would make more sense to have an example which doesn't lead into a pitfall.

perlpunk:
@okurz  Just a note: cloning via `git@` should work even if you have no "admin access" to the repo. It just requires your key, but you can clone any public repo with that.

perlpunk:
@Martchus 
I'd vote for having both http and git+ssh examples, and add a short explanation why only http will work on remote machines like osd/o3.

doc/backend_vars.asciidoc
```diff
@@ -6,7 +6,7 @@ Supported variables per backend
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
 Variable;Values allowed;Default value;Explanation
-CASEDIR;string;;Path to test distribution. Can be a git repository URL of a test distribution to checkout with an optional refspec into the current directory. Tries to follow the definition of https://docs.npmjs.com/files/package.json#git-urls-as-dependencies , for example `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git#feature/test`
+CASEDIR;string;;Path to test distribution. Can be a git repository URL of a test distribution to checkout with an optional refspec into the current directory. It tries to follow the definition of https://docs.npmjs.com/files/package.json#git-urls-as-dependencies (e.g. `https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#feature/test` or `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git`). Note that cloning via SSH is not possible if git is not configured accordingly on the worker host.
```
okurz: I suggest to avoid the double negation so instead of `is not possible if git is not configured accordingly` write `is only possible if git is configured accordingly`. Also "worker host" is openQA speak which should be avoided in os-autoinst content. Just end the sentence after "accordingly."

Martchus: > Also "worker host" is openQA speak which should be avoided in os-autoinst content.

But openQA nevertheless exists and different hosts might be involved. To make it clear that neither the openQA web UI host or the local host are relevant I changed it to "on the host where isotovideo is executed".

PR #1273, opened 2019-11-21
t: Prevent error in RPM build on empty line after backslash

okurz:


okurz:
hm, coverage decreased but there are still 31 test modules executed. I thought my change could have missed a file. Let me retrigger and see if the number changes.

okurz:
It's ok now. I will merge myself to fix the build of os-autoinst in OBS.

PR #1275, opened 2019-11-21
Make qemu boot menu optional to save 5s in every qemu test run

okurz:
See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/8972
for the removal of the unused boot menu handling code.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -838,20 +838,23 @@ sub start_qemu {
             sp('append', "dhcp && sanhook iscsi:$vars->{WORKER_HOSTNAME}::3260:1:$vars->{NBF}", no_quotes => 1);
         }
 
+        my @boot_args;
+        push @boot_args, ('menu=on,splash-time=' . $vars->{BOOT_MENU_TIMEOUT} // '5000') if $vars->{BOOT_MENU};
         if ($arch_supports_boot_order) {
             if (($vars->{PXEBOOT} // '') eq 'once') {
-                sp("boot", "once=n");
+                push @boot_args, 'once=n';
             }
             elsif ($vars->{PXEBOOT}) {
-                sp("boot", "n");
+                push @boot_args, 'n';
             }
             elsif ($vars->{BOOTFROM}) {
-                sp("boot", [qv "order=$vars->{BOOTFROM} menu=on splash-time=5000"]);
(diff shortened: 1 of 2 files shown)


backend/qemu.pm
```diff
@@ -838,20 +838,23 @@ sub start_qemu {
...
             elsif ($vars->{BOOTFROM}) {
-                sp("boot", [qv "order=$vars->{BOOTFROM} menu=on splash-time=5000"]);
+                push @boot_args, "order=$vars->{BOOTFROM}";
             }
             else {
-                sp("boot", [qw(once=d menu=on splash-time=5000)]);
+                push @boot_args, 'once=d';
             }
         }
+        sp('boot', join(',', @boot_args)) if @boot_args;
```
Martchus: The previous code just passes multiple args to `sp` without manually joining them before. I can't find the definition of `sp` (worst function name ever; `git grep 'sub sp'` returns no results) so I can't tell whether this is a problem or not.

It also seems that previously the combination `PXEBOOT` and having the boot menu wasn't possible. Now it is possible. I assume that's intended?

okurz: > The previous code just passes multiple args to `sp` without manually joining them before. I can't find the definition of `sp` (worst function name ever; `git grep 'sub sp'` returns no results) so I can't tell whether this is a problem or not.

see https://github.com/os-autoinst/os-autoinst/blob/master/backend/qemu.pm#L561 . It's part of the same method. I think it's a "function pointer" or something :)

> It also seems that previously the combination `PXEBOOT` and having the boot menu wasn't possible. Now it is possible. I assume that's intended?

Yes, this is intended because the boot menu would become optional anyway and leave the choice to the user if one thinks if it makes sense or not.

Martchus: > I think it's a "function pointer" or something :)

This is something weird, indeed. If it was part of the function I'd expected just `my $sp = sub {...` with a call syntax like `$sp->(...)`. So this is just a shortcut for `sub static_param`. It seems to distinguish exactly these cases so let's just say you thought about it and picked the right type of parameter passing here.

> leave the choice to the user

good

okurz:
Ran a test based on os-autoinst-distri-opensuse and it did not cause any problems

PR #1276, opened 2019-11-21
Enhance logs

okurz:
* t: Catch output in 18-backend-qemu.t
* Remove obsolete AUTO_INST after more than a year past c670720d
* Prevent duplicate "Saving QEMU state to qemu_state.json"
* Avoid the word "killing" to have nicer wording and discern it from POSIX "kill"

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1356,11 +1356,11 @@ sub hide_password {
 }
 
 # Send TERM signal to any child process
-sub _kill_children_processes {
+sub _stop_children_processes {
     my ($self) = @_;
     my $ret;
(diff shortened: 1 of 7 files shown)


okurz:
Yeah even if a method "stop" would do "term, wait, otherwise kill" I would not call it "kill" otherwise one might think it sends SIGKILL immediately which would be messy. You know, one of these beginner Linux tips: "I don't know, just call 'kill -9', always works for me!" ;)

okurz:
As I fixed the trivial style issue and no one commented anything else I assume I can assume your approval and merge as soon as tests have succeeded

okurz:
tests on travis CI are actually passed since at least 30 minutes

PR #1277, opened 2019-11-22
myjsonrpc: Fix unclean shutdown introduced in 449baece

okurz:
When we failed to write to the RPC queue we still not to terminate the
process as there is no better handler on any higher level.

This prevents problems as observed on openqa.opensuse.org when qemu
stops prematurely but the RPC queue continues for longer polluting the
log with misleading information, e.g.:

```
[2019-11-21T12:05:50.533 CET] [debug] QEMU: qemu-system-x86_64: -blockdev driver=qcow2,node-name=hd0-overlay3,file=hd0-overlay3-file,cache.no-flush=on: Could not open backing file: Could not open backing file: Could not open backing file: Could not open backing file: Could not open '/var/lib/openqa/pool/11/opensuse-Tumbleweed-x86_64-20191120-gnome-wayland@64bit_virtio.qcow2': No such file or directory
[2019-11-21T12:05:50.533 CET] [debug] sending magic and exit
[2019-11-21T12:05:50.534 CET] [debug] received magic close
[2019-11-21T12:05:50.534 CET] [debug] THERE IS NOTHING TO READ 15 4 3
[2019-11-21T12:05:50.534 CET] [debug] killing command server 5926 because test execution ended
[2019-11-21T12:05:50.534 CET] [debug] isotovideo: informing websocket clients before stopping command server: http://127.0.0.1:20113/gd9yYstnTlhgNmPO/broadcast
[2019-11-21T12:05:50.548 CET] [debug] backend process exited: 0
[2019-11-21T12:05:50.551 CET] [debug] commands process exited: 0
[2019-11-21T12:05:51.551 CET] [debug] done with command server
[2019-11-21T12:05:51.551 CET] [debug] killing autotest process 5937
[2019-11-21T12:05:51.551 CET] [debug] sysread failed:
Use of uninitialized value $command in string eq at /usr/lib/os-autoinst/autotest.pm line 217.
[2019-11-21T12:05:51.552 CET] [debug] sysread failed:
[2019-11-21T12:05:51.552 CET] [debug] sysread failed:
[2019-11-21T12:05:51.552 CET] [debug] ||| starting prep_pidgin tests/x11/pidgin/prep_pidgin.pm
[2019-11-21T12:05:51.552 CET] [debug] sysread failed:
[2019-11-21T12:05:51.552 CET] [debug] /var/lib/openqa/cache/openqa1-opensuse/tests/opensuse/tests/x11/pidgin/prep_pidgin.pm:27 called testapi::mouse_hide
[2019-11-21T12:05:51.552 CET] [debug] <<< testapi::mouse_hide(border_offset=1)
[2019-11-21T12:05:51.552 CET] [debug] sysread failed:
```

This is fixed here by exiting from the place where we encountered the
broken pipe and also not outputting "sysread failed" in this case so the
logs when qemu terminates midway looks like:

```
[2019-11-22T08:10:32.646 CET] [debug] WARNING: check_asserted_screen took 0.53 seconds for 32 candidate needles - make your needles more specific
[2019-11-22T08:10:32.647 CET] [debug] no match: 79.0s, best candidate: grub2-2-20131030 (0.00)
[2019-11-22T08:10:33.440 CET] [debug] no match: 78.0s, best candidate: grub2-2-20131030 (0.00)
[2019-11-22T08:10:33.768 CET] [debug] QEMU: qemu-system-x86_64: terminating on signal 15 from pid 17357 ()
[2019-11-22T08:10:33.794 CET] [info] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[2019-11-22T08:10:33.795 CET] [debug] sending magic and exit
[2019-11-22T08:10:33.795 CET] [debug] received magic close
[2019-11-22T08:10:33.796 CET] [debug] THERE IS NOTHING TO READ 17 4 3
[2019-11-22T08:10:33.796 CET] [debug] killing command server 17326 because test execution ended
[2019-11-22T08:10:33.796 CET] [debug] isotovideo: informing websocket clients before stopping command server: http://127.0.0.1:20013/3HbEX_vD1nTThFUd/broadcast
[2019-11-22T08:10:33.808 CET] [debug] backend process exited: 0
[2019-11-22T08:10:33.831 CET] [debug] Driver backend collected unknown process with pid 17333 and exit status: 0
[2019-11-22T08:10:34.783 CET] [debug] commands process exited: 0
[2019-11-22T08:10:35.784 CET] [debug] done with command server
[2019-11-22T08:10:35.784 CET] [debug] killing autotest process 17329
[2019-11-22T08:10:35.784 CET] [debug] myjsonrpc: remote end terminated connection, stopping
[2019-11-22T08:10:35.788 CET] [debug] [autotest] process exited: 0
```

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -1,4 +1,4 @@
-# Copyright © 2012-2016 SUSE LLC
+# Copyright © 2012-2019 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -53,9 +53,10 @@ sub send_json {
 
     my $wb = syswrite($to_fd, "$json");
     if (!$wb || !$wb == length($json)) {
-        confess "syswrite failed: $!" if DEBUG_JSON;
-        # remote end most likely terminated on request, silently ignore
-        return undef;
+        if (!DEBUG_JSON && $! =~ qr/Broken pipe/) {
+            die('myjsonrpc: remote end terminated connection, stopping');
+        }
+        confess "syswrite failed: $!";
     }
     return $cmdcopy{json_cmd_token};
 }
@@ -132,7 +133,10 @@ sub read_json {
 
         my $qbuffer;
         my $bytes = sysread($socket, $qbuffer, READ_BUFFER);
-        if (!$bytes) { bmwqemu::diag("sysread failed: $!"); return; }
+        if (!$bytes) {
+            bmwqemu::diag("sysread failed: $!") if DEBUG_JSON;
+            return;
+        }
         $cjx->incr_parse($qbuffer);
     }
 


myjsonrpc.pm
```diff
@@ -53,9 +54,11 @@ sub send_json {
 
     my $wb = syswrite($to_fd, "$json");
     if (!$wb || !$wb == length($json)) {
-        confess "syswrite failed: $!" if DEBUG_JSON;
-        # remote end most likely terminated on request, silently ignore
-        return undef;
+        if (!DEBUG_JSON && $! =~ qr/Broken pipe/) {
+            bmwqemu::diag("myjsonrpc: remote end terminated connection, stopping");
+            _exit(0);
```
coolo: careful here. if the backend is dead, we still don't want to kill isotovideo

okurz: yes but I don't think that happens. I did not paste the full log in the commit message. It actually ends with:

```
[2019-11-22T09:50:24.694 CET] [debug] done with autotest process
[2019-11-22T09:50:24.694 CET] [debug] isotovideo failed
[2019-11-22T09:50:24.695 CET] [debug] stopping backend process 25141
[2019-11-22T09:50:24.695 CET] [debug] done with backend process
25133: EXIT 1
```

so, are we good?

Martchus: Not sure, I needed to do some more own tests (e.g. killing the command server while the test is executed. Isotovideo should then still perform the usual cleanup and not just exit). Because this really looks like isotovideo would now just stop when a broken pipe is encountered which is definitely wrong. And even if that's not the case a comment stating why this code is not as dangerous as one would assume might be helpful.

coolo: I'm not saying it happens in your case. but myjsonrpc is used everywhere

coolo: even if it's not dangerous now - it might be used in different places where this exit then becomes dangerous. So returning undef and exiting from the non-dangerous caller is still preferred.

okurz: I see. I will try to do that. But please keep in mind that before 449baece it called "confess" in all cases so I would not see how _exit could be worse

Martchus: That is actually a good point. Maybe it works because everything is wrapped within the `if (!$wb || !$wb == length($json))` block. Not sure how to read that condition, though. The `!$wb == length($json)` part seems pointless because `$wb` is negated before the comparison. And that has changed with commit 449baece.

okurz: I think I have found a good testing procedure now:

* start `isotovideo -d`, e.g. with os-autoinst-distri-opensuse where the tests would go into an `assert_screen` waiting for a screen for a bit longer
* as soon as all processes are running and the test is waiting for needle matches, e.g. "no match: 79.0s, best candidate: grub2-2-20131030 (0.00)" kill one of the processes and look for proper shutdown in each case, e.g.
  * `pkill -f "qemu-system-x86_64"`
  * `pkill -f "commands"`
  * `pkill -f "autotest"`
  * `pkill -f "backend"`

It seems with my change the shutdown in case of `pkill -f "commands"` leaves subprocesses behind that only end themselves after the main process exited, i.e. there is a bash prompt already but then you see output.

Martchus: And that might be because the main process (= `isotovideo`) exits without doing the cleanup code which would take care of the other processes. Have you tested this without your change as well?

Besides, it would be good if the messed up condition could be fixed as well.

okurz: Yes, sure. I crosschecked "old" vs. "new". What "messed up conition"?

perlpunk: @okurz `if (!$wb || !$wb == length($json))`

Martchus: Yes, that's the condition I've mean (and pointed out in a previous comment).

okurz:
still haven't managed to have automatic tests covering this but I simplified it to have a simple "die" message equivalent to the "confess" we had earlier just with less noisy stack trace. I plan to introduce automatic tests, improve the condition as you mentioned and also replace the die with proper error propagation to the parent stack levels and terminate there.

PR #1278, opened 2019-11-22
Add powerVM over HMC backend

nicksinger:
Every command we currently use NovaLink for (starting/stopping LPARs and attach to their serial console) can also be achieved with the HMC command line. This is a preparation to support PPC machines (master-)managed by a Power HMC.

### DIFF
--- a/backend/pvm_hmc.pm
+++ b/backend/pvm_hmc.pm
@@ -1,0 +1,102 @@
+# Copyright © 2019 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package backend::pvm_hmc;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+
+use testapi qw(get_var get_required_var);
+
+# supporting the minimal command set of the HMC through a ssh tunnel
+
+sub new {
+    my $class = shift;
+    my $self  = $class->SUPER::new;
+    get_required_var('HMC_MACHINE_NAME');
+
+    return $self;
+}
+
+# only define the HMC console - we leave the actual
+# poweron to the test
+sub do_start_vm {
+    my ($self) = @_;
+
+    # truncate the serial file
+    open(my $sf, '>', $self->{serialfile});
+    close($sf);
+
+    my $ssh = $testapi::distri->add_console(
+        'powerhmc-ssh',
+        'ssh-xterm',
+        {
+            hostname   => get_required_var('HMC_HOSTNAME'),
+            password   => get_required_var('HMC_PASSWORD'),
+            username   => get_var('HMC_USERNAME', 'hscroot'),
+            persistent => 1});
+    $ssh->backend($self);
+
+    return {};
+}
+
+sub do_stop_vm {
+    my ($self) = @_;
+
+    $self->stop_serial_grab;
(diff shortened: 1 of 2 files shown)


backend/pvm_hmc.pm
```diff
@@ -0,0 +1,103 @@
...
+    $hostname ||= get_required_var('HMC_HOSTNAME');
+    $password ||= get_required_var('HMC_PASSWORD');
+    my $username = get_var('HMC_USERNAME', 'hscroot');
+
+    return $self->run_ssh_cmd($cmd, username => $username, password => $password, hostname => $hostname, keep_open => 0);
+}
+
+sub can_handle {
+    my ($self, $args) = @_;
+    return;
```
Martchus: Better use `return undef;` here. Otherwise this will return an empty list when called in list context.

backend/pvm_hmc.pm
```diff
@@ -0,0 +1,103 @@
...
+        return 1;
+    }
+    return $self->SUPER::check_socket($fh, $write);
+}
+
+sub stop_serial_grab {
+    my ($self) = @_;
+
+    $self->stop_ssh_serial;
+    return;
```
Martchus: Better use `return undef;` here. Otherwise this will return an empty list when called in list context.

backend/pvm_hmc.pm
```diff
@@ -0,0 +1,103 @@
...
+
+package backend::pvm_hmc;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+
+use testapi qw(get_var get_required_var check_var);
+use IO::Select;
```
Martchus: Seems like `IO::Select` is not used here.

backend/pvm_hmc.pm
```diff
@@ -0,0 +1,103 @@
...
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package backend::pvm_hmc;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+
+use testapi qw(get_var get_required_var check_var);
```
Martchus: `check_var` is not used here.

backend/pvm_hmc.pm
```diff
@@ -0,0 +1,103 @@
...
+sub can_handle {
+    my ($self, $args) = @_;
+    return;
+}
+
+sub is_shutdown {
+    my ($self) = @_;
+    my $lpar_id = get_required_var('LPAR_ID');
+    my $hmc_machine_name = get_required_var('HMC_MACHINE_NAME');
+    return $self->run_cmd("! lssyscfg -m ${hmc_machine_name} -r lpar --filter 'lpar_ids=${hmc_machine_name}' -F state | grep -i 'not activated' -q");
```
sqozz: `--filter 'lpar_ids=${hmc_machine_name}'` is a typo and should be `--filter 'lpar_ids=${lpar_id}'`

backend/pvm_hmc.pm
okurz: Also in general we should prefer to not use the external API anywhere inside. We can simply use `%bmwqemu::vars` directly

okurz:
s/archived/achieved/

Martchus:
@okurz I had the same thought but would not mix that kind of refactoring into that PR.

okurz:
Yes, we can do it independantly. I already drafted something in https://github.com/os-autoinst/os-autoinst/compare/master...okurz:enhance/dry_backend?expand=1

backend/pvm_hmc.pm
nicksinger: Leftover from copy&paste from the nova pvm back-end. Removed now :+1:

backend/pvm_hmc.pm
nicksinger: I'd leave this decision up to the follow-up PR which is supposed to unify these backends if you're fine with this? (I honestly have no clue what you mean with `%bmwqemu::vars`)

okurz: ok, I am only asking you for one thing then: Remind me of this after this PR here is done :)

okurz:
@nicksinger please mention the new file in Makefile.am as well. I guess we do not need more on top for now to be able to merge.

sqozz:
> @nicksinger please mention the new file in Makefile.am as well. I guess we do not need more on top for now to be able to merge.

Done and also ran a tidy so the tests can pass (hopefully) :+1:

PR #1279, opened 2019-11-22
Delete vim footer in files that still have it

okurz:
I like vim and use it myself but we do not need to
prescribe editor specific configs in files.


AdamWill:
I don't disagree with this, but FWIW, other editors can actually read and apply vim modelines fine. gedit does it for various things, including tab width and line length.

PR #1282, opened 2019-11-26
Add open_serial_console_via_ssh() test

cfconrad:
* Add test for open_serial_console_via_ssh() into 22-svirt.t
  * Small cleanup inside this function.

### DIFF
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -8,7 +8,8 @@ use Test::MockModule;
 use Test::MockObject;
 use Test::Warnings;
 use Test::Exception;
-use Test::Output 'stderr_like';
+use Test::Output;
+use Mojo::Log;
 use Scalar::Util 'refaddr';
 use XML::SemanticDiff;
 use backend::svirt;
@@ -16,6 +17,9 @@ use distribution;
 use Net::SSH2;
 use testapi qw(get_var get_required_var check_var set_var);
 use backend::svirt qw(SERIAL_CONSOLE_DEFAULT_PORT SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT);
+
+# Initialize bmwqemu loggers
+bmwqemu::init();
 
 set_var(WORKER_HOSTNAME => 'foo');
 set_var(VIRSH_HOSTNAME  => 'bar');
@@ -80,7 +84,6 @@ subtest 'XML config for VNC and serial console' => sub {
         ok('XML looks as expected');
     }
 };
-
 
 subtest 'SSH credentials' => sub {
 
@@ -152,9 +155,14 @@ subtest 'SSH usage in svirt' => sub {
 
     # check connection handling
     my $ssh1 = $svirt->new_ssh_connection();
-    my $ssh2 = $svirt->new_ssh_connection();
-    my $ssh3 = $svirt->new_ssh_connection(keep_open => 1);
-    my $ssh4 = $svirt->new_ssh_connection(keep_open => 1);
+    my ($ssh2, $ssh3, $ssh4);
+    my $exp_log        = qr/SSH connection to root\@bar established/;
+    my $default_logger = $bmwqemu::logger;
+    $bmwqemu::logger = Mojo::Log->new(level => 'debug');
+    stderr_like { $ssh2 = $svirt->new_ssh_connection(); } $exp_log, 'New SSH connection announced in logs';
+    stderr_like { $ssh3 = $svirt->new_ssh_connection(keep_open => 1); } $exp_log, 'New SSH connection announced in logs';
+    stderr_unlike { $ssh4 = $svirt->new_ssh_connection(keep_open => 1); } $exp_log, 'No new SSH connection announced, if it already exists';
+    $bmwqemu::logger = $default_logger;
     $ssh_expect_credentials->{username} = 'foo911';
     my $ssh5 = $svirt->new_ssh_connection(keep_open => 1, username => 'foo911');
     $ssh_expect_credentials->{username} = 'root';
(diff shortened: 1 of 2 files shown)


okurz:
well, the tests you added fail: https://travis-ci.org/os-autoinst/os-autoinst/builds/617277529#L979
What else would I need to say ;)

cfconrad:
Nothing :)
During writing this test, I removed all other subtests from this file. And I didn't realized that `VIRSH_VMM_FAMILY` is set in one of these to `hyperv`

t/22-svirt.t
```diff
@@ -203,4 +203,73 @@ subtest 'SSH usage in svirt' => sub {
...
+    $bmwqemu::vars{JOBTOKEN} = 'CHECK_DELETE_TOKEN';
+    $test_log_cnt = 11;
+    dies_ok(sub { $svirt->open_serial_console_via_ssh('NAME') }, "die() when log file wasn't created");
+
+    $test_log_cnt = 0;
+    $grep_return  = 0;
+    dies_ok(sub { $svirt->open_serial_console_via_ssh('NAME') }, 'die() when emulate CONSOLE_EXIT token in log file');
+
+    my $expected_serial_file = '/tmp/' . $svirt->SERIAL_TERMINAL_LOG_PATH . '.CHECK_DELETE_TOKEN';
+    is($_, $expected_serial_file, "Check if file get cleaned") for (@deleted_logs);
```
okurz: When trying this out manually I found that `@deleted_logs` includes the same file name twice. I stumbled over this as I tried to make the output string unique for each step with

```perl
-    is($_, $expected_serial_file, "Check if file get cleaned") for (@deleted_logs);
+    is($_, $expected_serial_file, "Check if file '$_' gets cleaned") for (@deleted_logs);
```

would be good if you can take a look into this and potentially include my change if you agree.

cfconrad: Yes, the name is there twice, that's why I made this `foreach` loop.
I think I'm going to remove this `foreach` and check `@deleted_logs` after each `open_serial_console_via_ssh()` call.

The thing is, that actually this function should be called only once, but there is
no check inside of it :/ (only a hint in perlpod https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L261). Maybe something we need to improve anyway, but I would do it not with this PR.

cfconrad:
I reduced output by mocking `bmwqemu::diag()` and `bmwqemu::log_call()`. Then I added one specific point to check for important log output with Test::Output.

I searched for a way to call `prove` to show only the TestMore result output, but I didn't found anything easy.

WDYT @okurz

t/22-svirt.t
```diff
@@ -21,10 +21,14 @@ BEGIN {
     unshift @INC, '..';
 }
 
+my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+# Silent bmwqemu::diag() and bmwqemu::log_call()
+$bmwqemu_mock->mock('diag');
+$bmwqemu_mock->mock('log_call');
```
okurz: interesting idea :)

okurz:
normal output from prove should be like the following:

```
$ (cd t && prove 05-pod.t)
05-pod.t .. ok   
All tests successful.
Files=1, Tests=1,  0 wallclock secs ( 0.02 usr  0.00 sys +  0.08 cusr  0.00 csys =  0.10 CPU)
Result: PASS
$ (cd t && prove -v 05-pod.t)
05-pod.t .. 
1..1
ok 1 - POD test for ./../testapi.pm
ok
All tests successful.
Files=1, Tests=1,  0 wallclock secs ( 0.02 usr  0.01 sys +  0.08 cusr  0.01 csys =  0.12 CPU)
Result: PASS
```

so only `..` between the start of the test module and the "ok". And a single line with result from every single test call when supplying the parameter "-v". Unfortunately many test modules in os-autoinst do not follow this yet so the console becomes a bit "polluted" so please don't get lost trying to fix all of it in one go :) If you want to show something from tests in verbose mode use `diag` (from Test::More, not bmwqemu). If you want to show something regardless if verbose or silent mode use `note`.

cfconrad:
Don't understand what do you propose. From my point of view the problem is that `Mojo::Log` by default logs to `STDERR`. Which we do, if nobody calls `bmwqemu::init()` with `$bmwqemu::direct_output == 0`.

This would be my take to use `bmwqemu:init()` to hide the output, but still checking the log messages from time to time. https://github.com/os-autoinst/os-autoinst/pull/1282/commits/9af73264e78703207400be64f9c16b34474ff123

t/22-svirt.t
```diff
@@ -21,6 +22,8 @@ BEGIN {
     unshift @INC, '..';
 }
 
+bmwqemu::init();
```
okurz: I am ok with this change to make the tests look nice but without any kind of programmatic check I doubt it will hold very long. The next contributor to tests will forget to add this in any new module and also it's not very clear what `bmwqemu::init()` does here.

cfconrad:
VR: http://cfconrad-vm.qa.suse.de/tests/6575

t/22-svirt.t
```diff
@@ -199,4 +208,73 @@ subtest 'SSH usage in svirt' => sub {
     is_deeply($svirt_console->get_cmd_output('echo -n "BLAFAFU"', {wantarray => 1}), ['BLAFAFU', ''], "sshVirtsh::get_cmd_output(wantarray => 1) ");
 };
 
+
+subtest open_serial_console_via_ssh => sub {
```
okurz: actually the first argument to "subtest" can (and should) be a human readable string, e.g.

```perl
subtest 'serial_console via SSH' => sub {
```

cfconrad: I changed it to `Method backend::svirt::open_serial_console_via_ssh()` as this section is only to check this specific method from backend::svirt.
wdyt?

t/22-svirt.t
```diff
@@ -199,4 +208,73 @@ subtest 'SSH usage in svirt' => sub {
     is_deeply($svirt_console->get_cmd_output('echo -n "BLAFAFU"', {wantarray => 1}), ['BLAFAFU', ''], "sshVirtsh::get_cmd_output(wantarray => 1) ");
 };
 
+
```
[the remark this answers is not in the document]
cfconrad: I also removed a other "additional blank line" so we have only one line between each subtest section.

t/22-svirt.t
okurz: sure

PR #1284, opened 2019-11-26
[WIP] Reduce test time; reduce shutdown time; cleanup log messages

okurz:
* Add test for shutdown time
* t: Cut 18-qemu-options.t runtime from 45s to 27s by allowing to abort qemu start early
* t: Cut 18-qemu-options.t runtime from 135s to 45s by configuring attempt interval
* myjsonrpc: Fix unclean shutdown introduced in 449baece

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -22,6 +22,7 @@ use bmwqemu ();
 use Errno;
 use Mojo::JSON;    # booleans
 use Cpanel::JSON::XS ();
+use POSIX '_exit';
 
 use constant DEBUG_JSON  => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8000000;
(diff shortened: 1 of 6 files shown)


myjsonrpc.pm
```diff
@@ -53,9 +54,11 @@ sub send_json {
 
     my $wb = syswrite($to_fd, "$json");
     if (!$wb || !$wb == length($json)) {
-        confess "syswrite failed: $!" if DEBUG_JSON;
-        # remote end most likely terminated on request, silently ignore
-        return undef;
+        if (!DEBUG_JSON && $! =~ qr/Broken pipe/) {
+            bmwqemu::diag("myjsonrpc: remote end terminated connection, stopping");
+            _exit(0);
```
Martchus: Is it possible to take that change out and keep it separate (see https://github.com/os-autoinst/os-autoinst/pull/1277#discussion_r349466656).

okurz: yes, sure. It will be handled in https://github.com/os-autoinst/os-autoinst/pull/1277#discussion_r349466656. I will mark the PR as [WIP] and I would appreciate a review of everything else included then

PR #1286, opened 2019-11-28
Allow relative "productdir" and "needledir" relative to "casedir"

okurz:
This might help with being more flexible regarding also custom git
checkouts in openQA as well as in the case of manual calls to isotovideo
not needing the user to provide productdir and/or needledir as absolute
paths when the general expectation is that they reside within casedir
anyway although this is not necessarily required.

Related progress issue: https://progress.opensuse.org/issues/56789

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -323,14 +323,7 @@ sub default_needles_dir {
 }
 
 sub init {
-    # validate that possibly user-provided NEEDLES_DIR is a path within the current working directory (usually the openQA worker's pool directory)
-    my $user_provided_needles_dir = $bmwqemu::vars{NEEDLES_DIR};
-    if (defined $user_provided_needles_dir) {
-        $user_provided_needles_dir = File::Spec->rel2abs($user_provided_needles_dir) unless File::Spec->file_name_is_absolute($user_provided_needles_dir);
-    }
-
-    # initialize/re-assign global $needledir
-    $needledir = ($user_provided_needles_dir // default_needles_dir);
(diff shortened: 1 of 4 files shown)


needle.pm
```diff
@@ -325,12 +325,10 @@ sub default_needles_dir {
 sub init {
     # validate that possibly user-provided NEEDLES_DIR is a path within the current working directory (usually the openQA worker's pool directory)
     my $user_provided_needles_dir = $bmwqemu::vars{NEEDLES_DIR};
-    if (defined $user_provided_needles_dir) {
-        $user_provided_needles_dir = File::Spec->rel2abs($user_provided_needles_dir) unless File::Spec->file_name_is_absolute($user_provided_needles_dir);
-    }
 
     # initialize/re-assign global $needledir
     $needledir = ($user_provided_needles_dir // default_needles_dir);
+    $needledir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needledir);
```
Martchus: We could write this as `$needledir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);` and remove all lines above.

okurz:
hm, we already allowed productdirs relative to the current working directory hence we could use "os-autoinst-distri-opensuse/products/opensuse" and such for custom git clones. I wonder if we really should try to fallback to relative to casedir as well. Maybe we just do it for needledir for a start?

needle.pm
okurz: done

PR #1287, opened 2019-11-29
ppmclibs: Fix warning "\1 better written as $1"

okurz:


okurz:
For something as trivial as this feel free to merge directly next time

PR #1288, opened 2019-11-29
Try to catch failures in "reset_consoles" like we do for activate

okurz:
Resetting a console might fail as well as activating or selecting one
depending on the backend implementation. For example the IPMI backend
needs to detach from a SOL session which might have terminated already
in before which we can catch and forward to the test scope.

This is untested.

Related progress issue: https://progress.opensuse.org/issues/60437

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -586,7 +586,13 @@ sub reset_consoles {
     # we iterate through all consoles
     for my $console (keys %{$testapi::distri->{consoles}}) {
         next if $self->console($console)->{args}->{persistent};
-        $self->reset_console({testapi_console => $console});
+        try {
+            local $SIG{__DIE__} = 'DEFAULT';
+            $self->reset_console({testapi_console => $console});
+        }
+        catch {
+            return {error => $_};
+        }
     }
     return;
 }
--- a/basetest.pm
+++ b/basetest.pm
@@ -652,20 +652,21 @@ sub standstill_detected {
 # state
 sub rollback_activated_consoles {
     my ($self) = @_;
+    my $ret;
     for my $console (@{$self->{activated_consoles}}) {
         # the backend will only reset its state, and call activate
         # the next time - the console itself might actually not be
         # able to activate a 2nd time, but that's up to the console class
-        autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
+        $ret = autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
+        last if $ret->{error};
     }
     $self->{activated_consoles} = [];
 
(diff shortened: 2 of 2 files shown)


okurz:
@coolo do you think this could work same as you did recently for "select_console"?

backend/baseclass.pm
```diff
@@ -586,7 +586,13 @@ sub reset_consoles {
     # we iterate through all consoles
     for my $console (keys %{$testapi::distri->{consoles}}) {
         next if $self->console($console)->{args}->{persistent};
-        $self->reset_console({testapi_console => $console});
+        try {
+            local $SIG{__DIE__} = 'DEFAULT';
+            $self->reset_console({testapi_console => $console});
+        }
+        catch {
+            return {error => $_};
```
Martchus: The `return` will return from the catch function. I think you want to return from the `sub reset_consoles` here? You could use `eval` to avoid this problem.

basetest.pm
```diff
@@ -652,20 +652,21 @@ sub standstill_detected {
...
 sub rollback_activated_consoles {
     my ($self) = @_;
+    my $ret;
     for my $console (@{$self->{activated_consoles}}) {
         # the backend will only reset its state, and call activate
         # the next time - the console itself might actually not be
         # able to activate a 2nd time, but that's up to the console class
-        autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
+        $ret = autotest::query_isotovideo('backend_reset_console', {testapi_console => $console});
+        last if $ret->{error};
```
Martchus: So there's no point in trying to reset the remaining consoles anymore? This looks like it should accumulate errors but continue. If `$autotest::last_milestone_console` is defined it will actually continue to do one more query but clears a previously occurred error which seems wrong, too.

So this function is inconsistent in deciding whether to abort on the first error or whether not and possibly suppresses errors.

backend/baseclass.pm
okurz: I would be open for pointers how to actually test this. Probably mock a dying reset console call?

basetest.pm
okurz: So far the reset consoles call will abort the test on error. I do not want to change this but just improve the error handling and notification. Also as the test would fail anyway I do not see a point trying to continue with the other consoles. About the specific problem of "sol deactivate" the problem could very well be to catch this in the specific backend and do whatever mitigation or repair is necessary, e.g. ignore exit code, retry, etc. But here, within the proxy call we just need to ensure the error is properly propagated

backend/baseclass.pm
Martchus: I guess I would go for the usual mocking and it it to `23-baseclass.t`.

PR #1289, opened 2019-11-29
Ensure we are also checking coverage in tests themselves

okurz:
We also want to know if there is dead code in our tests, e.g. by
accident. This also fixes the warning

```
^* matches null string many times in regex; marked by <-- HERE in m/^*
<-- HERE \.t|^data\/tests\/*|^fake\/tests\/*/ at
/usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover.pm
line 379.
```

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -1,5 +1,5 @@
 AM_MAKEFLAGS = \
-	PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db,-ignore,^.*\.t|^data\/tests\/|^fake\/tests\/" \
+	PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db,-ignore,^data\/tests\/*|^fake\/tests\/*" \
 	PERL5LIB="..:../ppmclibs:../ppmclibs/blib/arch/auto/tinycv:$$PERL5LIB"
 # Prevent error "blank line following trailing backslash" when deleting last
 # line in list in spec file to exclude tests


okurz:
@mkittler I think this should help with misleading warnings and should not be necessary but we can wait for coverage data to see if we are suddenly missing any.

Martchus:
I guess we should re-trigger the tests to find out whether the coverage drop is real.

okurz:
again the coverage is low. I don't understand what is happening here.

```
PERL5OPT="-MDevel::Cover=-db,./cover_db,-ignore,^data\/tests\/*|^fake\/tests\/*" perl -e 'print("report: " . Devel::Cover::report() . "\n") if Devel::Cover->can("report")'
```

shows at least the print line executed even though the report does not include anything but the `-MDevel::Cover` includes the coverage package. I would not need another `use Devel::Cover`, right?

However I always thought that tests themselves should not be excluded from coverage. We also want to know if there is dead code in our tests, e.g. by accident, right?

okurz:
For now I am repurposing the PR to add coverage analysis for test files as well, rest is reserved for another time. Analysis can go on in https://progress.opensuse.org/issues/42074 .

perlpunk:
Is there a way to get the `coverage.html` generated by Devel::Cover from codecov.io or Travis?
I have problems understanding the codecov.io results.

okurz:
Maybe the text based overview report in https://travis-ci.org/os-autoinst/os-autoinst/builds/618779656#L949 helps a bit. I don't think we have a html report when generating in codecov format.

perlpunk:
I don't know if it's related to the problem, but having a closer look at the Makefiles, I noticed that already `check-TESTS` is called with `PERL5OPT=-MDevel::Cover...`.
Actually only the test files should be run with that, but currently any perl script that is called during `check-TESTS` will generate coverage.
That could explain the weird messages from perltidy for example.

okurz:
So the only feedback I got is to not do this. I can live with that

perlpunk:
I can imagine having an additional make target that checks the coverage of the test files, but I also don't see the need to do this regularly.

PR #1290, opened 2019-11-29
Remove import of Devel::Cover in commands test

Martchus:
Since the coverage target invokes tests with `make PERL5OPT="-MDevel::Cover=...` this should not be necessary.


Martchus:
@okurz Just saw that you've already created a PR.

PR #1291, opened 2019-11-29
Increase calculated test coverage with "use Devel::Cover" in all tests

okurz:
It does not really make sense to me yet but including "use Devel::Cover"
in more tests raises coverage.

Tested with

```
mv t/cover_db{/,_before/} && ./configure && make && make coverage && mv t/cover_db{/,_all_devel_cover/} && firefox t/cover_db_{before,all_devel_cover}/coverage.html
```

I could see an overall coverage increase from 55.5 to 67.9

Certainly not all test files would really add to the coverage increase,
e.g. "00-compile-check-all.t" sounds unlikely to cause this but to be
sure unless we know better we might just as well add everywhere.

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -16,6 +16,7 @@
 
 use strict;
 use warnings;
+use Devel::Cover;
 use Test::Compile;
 use Test::Warnings;
 use Cwd;
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -2,6 +2,7 @@
 
 use strict;
 use warnings;
+use Devel::Cover;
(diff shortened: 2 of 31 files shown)


coolo:
please do not count tests as code to cover - this is just fake

okurz:
Hi coolo, what's up? Did you read my commit message? I don't know what more to say. Besides, the PR that only includes the first commit is https://github.com/os-autoinst/os-autoinst/pull/1289 . Probably you want to close this, too, then? Maybe the opinions of others would count as well?

okurz:
Opened https://github.com/os-autoinst/os-autoinst/pull/1292 instead with just the second commit.

PR #1292, opened 2019-11-30
[WIP] Increase calculated test coverage with "use Devel::Cover" in all tests

okurz:
It does not really make sense to me yet but including "use Devel::Cover"
in more tests raises coverage.

Tested with

```
mv t/cover_db{/,_before/} && ./configure && make && make coverage && mv t/cover_db{/,_all_devel_cover/} && firefox t/cover_db_{before,all_devel_cover}/coverage.html
```

I could see an overall coverage increase from 55.5 to 67.9

Certainly not all test files would really add to the coverage increase,
e.g. "00-compile-check-all.t" sounds unlikely to cause this but to be
sure unless we know better we might just as well add everywhere.

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -16,6 +16,7 @@
 
 use strict;
 use warnings;
+use Devel::Cover;
 use Test::Compile;
 use Test::Warnings;
 use Cwd;
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -2,6 +2,7 @@
 
 use strict;
 use warnings;
+use Devel::Cover;
 use Cwd 'abs_path';
 use Test::Exception;
 use Test::Output qw(combined_like stderr_like);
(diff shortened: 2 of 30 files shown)


coolo:
including devel:cover is just nonsense - and the coverage increase you see here is just the typical noise.
And the coverae increase you saw in your experiments are because you count the tests as covered code. More tests, more covered code -> nonsense.

okurz:
Can you please be less arrogant and offensive? I stated in the commit message myself that I do not see the sense in this and wanted to crosscheck different explanations. I did run the experiment locally with including tests both with "Devel::Cover" included and excluded and this made the difference, not the test code itself. Of course I can also run coverage analysis locally and not include tests in both scenarios.

okurz:
Judging from my experiments the coverage does not significantly increase. I understand now that the only test file where we need the include which I would like to explain with just adding a comment in https://github.com/os-autoinst/os-autoinst/pull/1297 . This is also explained in the discussion in https://github.com/os-autoinst/os-autoinst/pull/987 . I consider this PR here therefore obsolete. The comment from coolo to "remove the include of Devel::Cover in tests" I don't understand though

PR #1293, opened 2019-11-30
Simplify CI tests by relying on upstream openQA development image only

okurz:
Also as within openQA we want tests to rely on the dependencies
specified within cpanfile and spec and not install any missing
dependencies or updates during test runs


.travis.yml
```diff
@@ -2,8 +2,6 @@ sudo: required
 dist: trusty
 services:
   - docker
-before_install:
-  - docker build --no-cache docker/travis_test -t os-autoinst/travis_test:latest
 script:
   - CI_ENV=`bash <(curl -s https://codecov.io/env)` &&
-    docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_QEMU_START_S=20" $CI_ENV --rm -v $TRAVIS_BUILD_DIR:/opt/repo os-autoinst/travis_test:latest 'make && make check coverage-codecov VERBOSE=1'
+    docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_QEMU_START_S=20" $CI_ENV --rm --entrypoint '' -v $TRAVIS_BUILD_DIR:/opt registry.opensuse.org/devel/openqa/containers/openqa_dev sh -c 'cd /opt && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
```
perlpunk, line 7: This seems to have broken the coverage report.
Before the files had the prefix `/opt/run`, now it's only `/opt`

okurz: oh, could be. Good catch. I still wonder how codecov would cope with `/opt/run` then

PR #1295, opened 2019-11-30
WIP Tests & Coverage

perlpunk:
Run tests with `prove` from top directory.
This seems to report coverage of more files. The overall coverage is decreasing, but it might be closer to the truth than what we currently have ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ before_install:
   - docker build --no-cache docker/travis_test -t os-autoinst/travis_test:latest
 script:
   - CI_ENV=`bash <(curl -s https://codecov.io/env)` &&
-    docker run $CI_ENV --rm -v $TRAVIS_BUILD_DIR:/opt/repo os-autoinst/travis_test:latest 'make && make check coverage-codecov VERBOSE=1'
+    docker run $CI_ENV --rm -v $TRAVIS_BUILD_DIR:/opt/repo os-autoinst/travis_test:latest 'make && make check test-cover test-coverage-codecov VERBOSE=1'
--- a/Makefile.am
+++ b/Makefile.am
@@ -195,6 +195,20 @@ dist-hook:
 
 distclean-local: clean-local
 
+TEST_OPTS= \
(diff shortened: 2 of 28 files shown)


okurz:
In general I like it. I also wanted to make tests rely less on autotools and just prove.

perlpunk:
There's still something wrong with the reported coverage.
I can see code executed locally when adding debugging, but still Devel::Cover says it's not.

I can't even explain why those 27 new files appear in the coverage, and why they didn't before. Maybe it has to do with the current working directory changing one level up.
Will do more research.

okurz:
I guess we can finally see all the ugly warnings and output in the tests which are not hidden anymore :)

There seem to be some tests which are producing non-deterministic coverage, e.g. for "OpenQA/Benchmark/Stopwatch.pm" which commonly reports 71.6% stmt coverage or 100% as in your latest travis CI test run https://travis-ci.org/os-autoinst/os-autoinst/builds/619136255#L1501

Mainly files report higher coverage than what I am used to, but for example https://travis-ci.org/os-autoinst/os-autoinst/builds/619136255#L1507 shows just 41% for Utils.pm which I have seen with 100%. https://travis-ci.org/os-autoinst/os-autoinst/builds/619136255#L1520 shows 32% for autotest vs. 51% . The main application "isotovideo" is not mentioned at all. Seems your tests have lost this part which can probably also explain the other big losses. It looks like e.g. the fullstack test is executed but probably coverage from spawned processes not taken into account.

perlpunk:
`Coverage   41.44%   42.42%   +0.97%`
Now adding Devel::Cover back in 07-commands.t

perlpunk:
So we have over 5% more coverage (with keeping Devel::Cover in 07-commands.t), but still backend::driver has a loss of 67%.
Need to find out which test should be covering the missing code and compare master and PR in detail.

perlpunk:
Just removed Devel::Cover from 07-commands.t again and still got
`Coverage   41.44%   46.71%   +5.26%`
So it seems to be random.

PR #1297, opened 2019-12-01
t: Explain better the need for 'use Devel::Cover' in 07-commands.t

okurz:

### DIFF
--- a/t/07-commands.t
+++ b/t/07-commands.t
@@ -35,7 +35,7 @@ use Time::HiRes 'sleep';
 use Test::More;
 use Test::Warnings;
 use Test::Mojo;
-use Devel::Cover;
+use Devel::Cover;    # needed to collect coverage on spawned processes as well
 use Mojo::File qw(path tempfile tempdir);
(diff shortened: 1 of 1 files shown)


t/07-commands.t
```diff
@@ -35,7 +35,7 @@ use Time::HiRes 'sleep';
 use Test::More;
 use Test::Warnings;
 use Test::Mojo;
-use Devel::Cover;
+use Devel::Cover;    # needed to collect coverage on spawned processes as well
```
coolo: this is still a smell. The problem is more the _exit I suppose - because in openQA we put a lot of 
```Devel::Cover::report() if Devel::Cover->can('report');``` into the mix to have subprocesses reported

okurz: yeah, I have seen that and I agree it's a smell. I intend to let @perlpunk finish her part in https://github.com/os-autoinst/os-autoinst/pull/1295 before we can look into this further. Please also see my comments in https://progress.opensuse.org/issues/42074#note-2

PR #1298, opened 2019-12-02
SSH serial: Close connection on read error

cfconrad:
If SUT was triggered to do a reboot, it could happen that the SSH
connection was closed, but we still use the SSH->sock() within select().
This cause a loop of check_ssh_serial() calls, as we never handle the pending event.
More infos in https://progress.opensuse.org/issues/60416

VR: I run a lot of tests, as it only occur sporadically.
http://cfconrad-vm.qa.suse.de/t6476 .. http://cfconrad-vm.qa.suse.de/t6492

It happen 4 times out of 16:
http://cfconrad-vm.qa.suse.de/t6477
http://cfconrad-vm.qa.suse.de/t6483
http://cfconrad-vm.qa.suse.de/t6490
http://cfconrad-vm.qa.suse.de/t6491

search for
```
[debug] ssh serial: unable to read: transport read (error code: -43) - closing connection
```
in **autoinst-log.txt**

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1265,9 +1265,7 @@ sub check_ssh_serial {
     # read from SSH channel (receiving extended data channel as well via `$chan->ext_data('merge')`)
     my $chan = $self->{serial_chan};
     my $buffer;
-    my $bytes_read      = $chan->read($buffer, 4096);
-    my $could_read_once = defined $bytes_read;
-    while (defined $bytes_read) {
+    while (defined(my $bytes_read = $chan->read($buffer, 4096))) {
         return 1 unless $bytes_read > 0;
         print $buffer;
         open(my $serial, '>>', $self->{serialfile});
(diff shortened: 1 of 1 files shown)


backend/baseclass.pm
```diff
@@ -1341,9 +1340,10 @@ sub stop_ssh_serial {
 
     my $ssh = $self->{serial};
     return undef unless $ssh;
-
+    bmwqemu::diag("Close SSH serial connection with " . $ssh->hostname);
```
okurz: This is the "imperative" form and could confuse the reader to think that she would need to do anything. I guess you can state "Closing"? Also, as you are just concatenating the argument afterwards you can use single quotes.

cfconrad: done

okurz:
great, will merge. As always, as I assume you are still investigating further in this area so tests would be appreciated :)

cfconrad:
Is something missing for merging it? 
I will add the test with a separate PR, so we get this fix online.

okurz:
Oops, forgot to hit the button

PR #1301, opened 2019-12-03
myjsonrpc: Canonalize condition form

okurz:


myjsonrpc.pm
```diff
@@ -132,7 +133,10 @@ sub read_json {
 
         my $qbuffer;
         my $bytes = sysread($socket, $qbuffer, READ_BUFFER);
-        if (!$bytes) { bmwqemu::diag("sysread failed: $!"); return; }
+        if (!$bytes) {
+            bmwqemu::diag("sysread failed: $!") if DEBUG_JSON;
```
perlpunk: Do we really only want to output this if debugging is on?

okurz: yes, because we hit this situation in unclean shutdowns which are not at all related to the RPC communication part and therefore misleading the log reader

PR #1304, opened 2019-12-03
Add support for real systems without IPMI or KVM-IP

ggardet:
This is related to https://progress.opensuse.org/issues/60518 but it requires some work on `os-autoinst-distri-openSUSE` as well.

This PR enables SSH access when VNC is not selected on generalhw (`GENERAL_HW_VNC_IP` not defined), and add optional flash command support (`GENERAL_HW_FLASH_CMD` and `GENERAL_HW_FLASH_ARGS`).
You may need to adjust `SSH_XTERM_WAIT_SUT_ALIVE_TIMEOUT` to wait that SUT is live on network, default is `120 s`. Otherwise, `xtermvt` is started with an ssh command which fails due to `no route to host` and just hangs in this state.
By default, it uses a ssh serial. If you prefer to use the real serial output, you can set `GENERAL_HW_SOL_CMD` to a script which will grab the serial console. When available, this is better since you can check early boot traces, when a system is not booting properly until ssh.


okurz:
looks nice so far

consoles/sshXtermVt.pm
```diff
@@ -41,6 +42,19 @@ sub activate {
     my $sshcommand = $self->sshCommand($username, $hostname, $gui);
     my $serial     = $self->{args}->{serial};
 
+    # Wait that SUT is live on network (for generalhw/ssh)
+    my $p       = Net::Ping->new();
+    my $counter = get_var('SSH_XTERM_WAIT_SUT_ALIVE_TIMEOUT') // 120;
+    while ($counter > 0) {
+        if ($p->ping($ssh_args->{hostname})) {
+            last;
+        }
```
okurz: last if ...? :)

ggardet: Fixed! :)

backend/generalhw.pm
```diff
@@ -39,24 +39,33 @@ sub new {
...
+    $args = get_var('GENERAL_HW_FLASH_ARGS') if ($cmd eq 'GENERAL_HW_FLASH_CMD' and get_var('GENERAL_HW_FLASH_ARGS'));
+
     $cmd = get_required_var($cmd);
     $cmd = "$dir/" . basename($cmd);
     if (!-x $cmd) {
         die "CMD $cmd is not an executable";
     }
+
+    $cmd .= " $args" if $args;
+
```
okurz: can you just delete the blank lines, move the variable declaration onto the line of first use?

Or why not bring it all together:

```perl
my $args = get_var('GENERAL_HW_FLASH_ARGS') if ($cmd eq 'GENERAL_HW_FLASH_CMD' and get_var('GENERAL_HW_FLASH_ARGS'));
$cmd .= " $args" if $args;
```

backend/generalhw.pm
```diff
@@ -106,20 +115,35 @@ sub relogin_vnc {
 sub do_start_vm {
     my ($self) = @_;
 
+    $self->poweroff_host;
```
okurz: why poweroff on start?

consoles/sshXtermVt.pm
```diff
@@ -41,6 +42,17 @@ sub activate {
...
+    # Wait that SUT is live on network (for generalhw/ssh)
+    my $p       = Net::Ping->new();
+    my $counter = get_var('SSH_XTERM_WAIT_SUT_ALIVE_TIMEOUT') // 120;
+    while ($counter > 0) {
+        last if ($p->ping($ssh_args->{hostname}));
+        sleep(1);
+        $counter--;
+    }
+    $p->close();
+    bmwqemu::diag("$ssh_args->{hostname} does not seems to be alive. Continue anyway.\n") if ($counter == 0);
```
okurz: "Continue" sounds like imperative. How about "Continuing"?

backend/generalhw.pm
ggardet: I want to make sure the board is off before starting the flash process.

consoles/sshXtermVt.pm
ggardet: Or infinitive? ;) I will update it.

backend/generalhw.pm
ggardet: Fixed.

ggardet: Reverted, as `$cmd` is updated above and breaks comparison.

backend/generalhw.pm
okurz: ok, can you add this as a comment in source please?

ggardet: Done.

cpanfile
```diff
@@ -36,6 +36,7 @@ requires 'Mojo::UserAgent';
 requires 'Mojo::Log';
 requires 'Mojolicious::Lite';
 requires 'Net::DBus';
+requires 'Net::Ping';
```
okurz: please add perl-Net-Ping to the spec file in dist/rpm/os-autoinst.spec as well

ggardet: Fixed.

backend/generalhw.pm
```diff
@@ -44,19 +44,25 @@ sub get_cmd {
...
         die "CMD $cmd is not an executable";
     }
+    $cmd .= " $args" if $args;
+
     return $cmd;
 }
 
 sub run_cmd {
     my ($self, $cmd) = @_;
+    my @full_cmd = split / /, $self->get_cmd($cmd);
```
cfconrad: I would prefer here something like https://perldoc.perl.org/Text/ParseWords.html
e.g.
```suggestion
    my @full_cmd = quotewords('\s+', 0, $self->get_cmd($cmd));
```

backend/generalhw.pm
```diff
@@ -106,20 +112,35 @@ sub relogin_vnc {
 sub do_start_vm {
     my ($self) = @_;
 
+    $self->truncate_serial_file;
```
cfconrad: Don't we call `do_start_vm()` on reboot. And would this mean that each reboot removes everything before?

consoles/sshXtermVt.pm
```diff
@@ -35,6 +36,17 @@ sub activate {
     my $ssh_args        = $self->{args};
     my $gui             = $self->{args}->{gui};
 
+    # Wait that SUT is live on network (for generalhw/ssh)
+    my $p       = Net::Ping->new();
```
cfconrad: I had something similar in my mind. And I would prefer to probe ssh port directly like: https://github.com/cfconrad/os-autoinst/commit/4cc23ac953ea2cd83dfbd6e385f2d57b25f8c591

backend/generalhw.pm
ggardet: Not sure why it is best, but I can update it.

backend/generalhw.pm
ggardet: I do not think so. amt, ikvm, ipmi s390x, spvm and svirt backends have it: https://github.com/os-autoinst/os-autoinst/commit/63a2db6ba1617354ba062645a9e15143ed1d2850
@okurz ?

consoles/sshXtermVt.pm
ggardet: I am generally ok with your solution (not tested yet), even if I would prefer a bigger default wait value, but this must be configurable by a `get_var` as done here.
I can rebase my PR on top of your changes, once they will be merged.

cfconrad: I didn't created a PR for it yet. As I only saw it once during my local tests and so I wasn't sure if it is really needed. Feel free to grap it.

backend/generalhw.pm
cfconrad: I thought if the extra arguments contains spaces and need to be wrapped with `"`. But maybe there is just no need for it. It will not block me to give my ok!

ggardet: I uses an arg variable containing 2 args for my script, separated with white spaces and it does work fine.

consoles/sshXtermVt.pm
ggardet: Ok, so I grabbed your code and reworked it a bit.
Thanks!

backend/generalhw.pm
cfconrad: so if you don't expect any parameter will look like `param="foo bar"` then your split is ok.

Maybe a hint could be `/\s+/` so we don't add empty args to `IPC::Run::run`, when someone split the args with more then one `space`.

cfconrad:
LGTM, 
here is a run with this PR on x390 http://cfconrad-vm.qa.suse.de/tests/6580/file/autoinst-log.txt

backend/generalhw.pm
okurz: sorry, dunno

cfconrad: Ok, as this is used in other backends in the same way. Maybe my assumption is just wrong.

consoles/sshXtermVt.pm
```diff
@@ -62,6 +67,23 @@ sub activate {
...
+    $args{timeout} = 1 unless ($args{timeout} > 0);
+    my $endtime = time() + $args{timeout};
+    while (time() < $endtime) {
+        my $sock = IO::Socket::INET->new(PeerAddr => $hostname, PeerPort => $args{port}, Proto => 'tcp', Timeout => 1);
+        return 1 if (defined $sock);
+        sleep 1;
+    }
+    return 0;
+}
+
```
Soulofdestiny: hey - we already have a similiar check here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/opensusebasetest.pm#L674

you might want to unify those

ggardet: That one is not working in an sshXtermVt context since you need to check before starting Xterm. Otherwise XtermVT just hangs.
So, both should be kept.

okurz: ok but I would like to see how this looks like in real openQA tests, e.g. that this does not cause any hard to debug "incompletes"

ggardet: No `incompletes`:
```
[2019-12-10T17:18:06.510 CET] [debug] Connected to Xvnc - PID 26434
[2019-12-10T17:18:07.531 CET] [debug] Wait for SSH on host 192.168.0.15 (timeout: 120)
[2019-12-10T17:20:07.769 CET] [debug] 192.168.0.15 does not seems to have an active SSH server. Continuing anyway.

[2019-12-10T17:20:07.838 CET] [debug] led state 0 0 0 -261
[2019-12-10T17:20:07.852 CET] [debug] activate_console, console: root-console, type: ssh
[2019-12-10T17:20:07.853 CET] [debug] /var/lib/openqa/share/tests/opensuse/tests/jeos/prepare_firstboot.pm:34 called opensusebasetest::select_serial_terminal
[2019-12-10T17:20:07.853 CET] [debug] <<< testapi::assert_screen(mustmatch='password-prompt', timeout=30)
[2019-12-10T17:20:07.977 CET] [debug] no match: 59.9s, best candidate: password-prompt-20141126 (0.00)
[2019-12-10T17:20:08.860 CET] [debug] no change: 58.9s
```

And finally fail on test as usual.

PR #1305, opened 2019-12-03
Add new test/cover targets and run tests with prove

perlpunk:
Ideally tests should be run from the root directory, but several
tests and modules expect to be in the t directory, and also coverage
can't catch every file, so for now we cd into t/

See issue https://progress.opensuse.org/issues/42074


okurz:
se/severeal/several/

Makefile.am
```diff
@@ -213,6 +213,56 @@ clean-local:
 	-rm -rf t/cover_db/
 
 t/cover_db/: check
+COVER_OPTS= \
+	PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db,+inc,$(PWD),-select,$(PWD)/[^/]+.pm|$(PWD)/(OpenQA|consoles|backend)|$(PWD)/isotovideo,-ignore,/usr/bin/prove|t/data/|t/fake/|\.t|tools/lib/perlcritic"
```
okurz: this looks very dangerous as now we need to explicitly add new stuff to the list when we add new files or directories. Can we keep the old way or work with a blacklist instead?

Makefile.am
```diff
@@ -213,6 +213,56 @@ clean-local:
...
+	(cd t && $(TEST_OPTS) prove $(TEST_FILES) )
+
+test-cover:
+	(cd t && $(TEST_OPTS) $(COVER_OPTS) prove $(TEST_FILES) )
+	cover -summary cover_db
+
+test-cover-summary:
+	cover -summary cover_db
+
+test-cover-codecov: cover_db/
```
okurz: don't we also need to add `.PHONY`?

Makefile.am
```diff
@@ -213,6 +213,56 @@ clean-local:
...
+	cover -report codecov cover_db
 
 .PHONY: coverage-test
 coverage-test: t/cover_db/
```
okurz: now we have rules for `cover_db/` as well as `t/cover_db/`. Maybe this is now obsolete?

Makefile.am
perlpunk: I was trying a lot but I found no other way to collect all files.
But since in this PR I still fo `cd t` I can try with the old options.
*Ideally* we would have all modules in `lib/` and all scripts in `bin/` so we wouldn't have to struggle with this, but I guess moving all the modules into lib would be a bigger task.

Makefile.am
perlpunk: done

Makefile.am
perlpunk: I thought I'd keep it for comparison as long as I'm not sure the new target is working well.

Makefile.am
```diff
@@ -210,29 +210,85 @@ check-doc-testapi:
...
-	-rm -rf t/cover_db/
-
-t/cover_db/: check
-
-.PHONY: coverage-test
-coverage-test: t/cover_db/
+	-rm -rf cover_db/
+
+EMPTY =
+TEST_FILES = \
```
okurz: Please also adjust https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec#L91

perlpunk: Done.
I also changed the check to
`make check test VERBOSE=1`
The build will probably a bit faster as it doesn't use Devel::Cover anymore...

okurz: ok, can you test in OBS? If you haven't done this so far, I suggest the following:

* browse to https://build.opensuse.org/package/show/devel:openQA/os-autoinst
* click on "Branch package", click on "Accept"
* in the new project find the file "_service" and change it to point to …
  * your repo instead of "os-autoinst/os-autoinst"
  * your branch instead of "master"
* get a coffee, let it build and test :)

perlpunk: done, looks good: https://build.opensuse.org/package/show/home:tinita:branches:devel:openQA/os-autoinst

perlpunk:
added `cover_db` to `.gitignore`

t/Makefile.am
```diff
@@ -1,47 +1,9 @@
-UPDIR = $(dir $(PWD))
```
okurz: is the complete file even required now? I guess we can delete it

t/Makefile.am
```diff
@@ -1,47 +1,9 @@
...
-
-PATHRE = ^|$(UPDIR)|\.\./
-
 AM_MAKEFLAGS = \
-	PERL5OPT="-MDevel::Cover=-db,$(abs_builddir)/cover_db,-select,($(PATHRE))(OpenQA|backend|consoles|ppmclibs)/|($(PATHRE))isotovideo|($(PATHRE))[^/]+\.pm,-ignore,\.t|.*data/tests/|.*fake/tests/" \
 	PERL5LIB="$(PWD)/..:$(PWD)/../ppmclibs:$(PWD)/../ppmclibs/blib/arch/auto/tinycv:$$PERL5LIB"
 # Prevent error "blank line following trailing backslash" when deleting last
 # line in list in spec file to exclude tests
```
okurz: please move this comment into Makefile.am at the appropriate place that does the same

t/Makefile.am
perlpunk: ok, I deleted it

t/Makefile.am
perlpunk: done

t/Makefile.am
perlpunk: Now the build is faling with:
```
configure.ac:52: error: required file 't/Makefile.in' not found
Makefile.am: installing './depcomp'
autoreconf: automake failed with exit status: 1
```
https://travis-ci.org/os-autoinst/os-autoinst/builds/622425938#L525

perlpunk: added it back again, empty

PR #1306, opened 2019-12-03
Remove @INC modifications from tests

perlpunk:
Tests should not modify `@INC` themselves (and if it must, then via FindBin to
use absolute paths).
The paths should be provided via env vars or `prove -l` or `prove -I....`.
This makes it more flexible.

Also the paths were relative, so it was only possible to call the tests
from the t/ directory.

This is a part of a refactoring regarding this issue:
https://progress.opensuse.org/issues/42074

In `t/00-compile-check-all.t` the `use lib` is still necessary because
we are passing a relative path via PERL5LIB.

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -20,16 +20,12 @@ use Test::Compile;
 use Test::Warnings;
 use Cwd;
 
-BEGIN {
-    if (getcwd =~ /\/t$/) {
-        # not really using it, but we need the init to be called before the chdir
-        use FindBin;
-        chdir("..");
-    }
-    unshift @INC, ".";
-}
+use FindBin '$Bin';
+use lib "$Bin/..";
 
 use cv;
+
+chdir "$Bin/..";
 cv::init;
 
 my $test = Test::Compile->new();
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -16,7 +16,6 @@ eval 'use Test::More::Color';
 eval 'use Test::More::Color "foreground"';
 
 BEGIN {
-    unshift @INC, '..';
     $bmwqemu::vars{DISTRI}  = "unicorn";
     $bmwqemu::vars{CASEDIR} = "/var/lib/empty";
 }
--- a/t/02-test_ocr.t
+++ b/t/02-test_ocr.t
@@ -9,7 +9,6 @@ use File::Which 'which';
 use File::Basename;
 
 BEGIN {
-    unshift @INC, '..';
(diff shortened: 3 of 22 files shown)


okurz:
This seems same as for openQA we need to alter the path or use `prove -l` on every call, what is the benefit?

perlpunk:
> This seems same as for openQA we need to alter the path...

I don't understand this sentence.

We don't need to alter anything right now.
Currently the tests *need* to be run from the `t/` directory.
This is unusual for perl tests. Usually perl tests are run from the distribution root directory with just `prove -l`, which adds `lib/` to `@INC`.
A good test can be run from anywhere.
I also suspect that the changing PWD is a problem for Devel::Cover to select the right files.

Tests themselves should not put anything into `@INC`, and if really necessary, then it should not be a relative path.
I explained this already above.
The benefit is that we are more flexible.
In the end (after the refactoring) the tests can be run from anywhere.

We are already setting `PERL5LIB` for calling the tests, so where is the problem?
Your question is, what is the benefit, and I explained. Now my question is, what is the benefit of *keeping* `use lib '..'` everywhere if we are using PERL5LIB already? Not all tests have `use lib '..'` anyway, so why should we keep it in some tests?

perlpunk:
I created a followup PR #1307 where I modified the select option, because it seems to cover more files.
I also created PR #1308 which only modifies the select option to see if it makes a difference.
I tested branches in travis before creating PRs, now I would like to see codecov results.

perlpunk:
For some reason some tests were skipped:

https://travis-ci.org/os-autoinst/os-autoinst/builds/620272867#L689
```
PASS: 00-compile-check-all.t
PASS: 01-test_needle.t
PASS: 02-test_ocr.t
PASS: 03-testapi.t
PASS: 04-check_vars_docu.t
PASS: 05-pod.t
PASS: 06-pod-coverage.t
PASS: 07-commands.t
PASS: 08-autotest.t
PASS: 09-lockapi.t
PASS: 10-terminal.t
PASS: 10-test-image-conversion-benchmark.t
PASS: 11-image-ppm.t
PASS: 12-bmwqemu.t
PASS: 13-osutils.t
PASS: 18-qemu.t
PASS: 18-backend-qemu.t
PASS: 18-qemu-options.t
PASS: 19-isotovideo-command-processing.t
PASS: 20-openqa-benchmark-stopwatch-utils.t
PASS: 21-needle-downloader.t
PASS: 22-svirt.t
PASS: 23-baseclass.t
PASS: 24-myjsonrpc.t
PASS: 24-myjsonrpc-debug.t
PASS: 25-spvm.t
PASS: 99-full-stack.t```

That explains the low coverage

PR #1307, opened 2019-12-03
t: Adjust PERL5LIB and Devel::Cover options

perlpunk:
This removes the modifications of `@INC` from the test files.
Also it adjusts PERL5LIB to use absolute paths, and adds a select option for Devel::Cover.

This is a part of a refactoring regarding this issue:
https://progress.opensuse.org/issues/42074

Explanation:
Tests should not modify `@INC` themselves (and if it must, then via FindBin to
use absolute paths).
The paths should be provided via env vars or `prove -l` or `prove -I.....`
This makes it more flexible.

Also the paths were relative, so it was only possible to call the tests
from the t/ directory.

Usually code coverage is for unit tests.
You call the tests with `-Ilib` or `-Mblib` and unit tests run and a report gets created. Done.

We are calling tests from the `t/` directory, using `..` explicitly in tests, then in the tests we fork, and in the forked processes we do a `chdir` and another `use lib` with a relative path, and then we use other modules.
As far as I can see Devel::Cover is not able to collect all those modules with all those different paths unless we use the select option.

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -20,16 +20,12 @@ use Test::Compile;
 use Test::Warnings;
 use Cwd;
 
-BEGIN {
-    if (getcwd =~ /\/t$/) {
-        # not really using it, but we need the init to be called before the chdir
-        use FindBin;
-        chdir("..");
-    }
-    unshift @INC, ".";
-}
+use FindBin '$Bin';
+use lib "$Bin/..";
 
 use cv;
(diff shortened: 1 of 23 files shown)


perlpunk:
Test runs show that the coverage is broader (more files are in the report) and higher (+8.22%, +9.4%)

perlpunk:
rebased to master; getting another test run

PR #1309, opened 2019-12-03
t: Use Test::Strict like in openQA for an easier and broader 'compile' check

okurz:


t/00-compile-check-all.t
```diff
@@ -16,33 +16,21 @@
...
-@files = ('isotovideo', $test->all_pl_files("."));
-for my $file (@files) {
-    $test->ok($test->pl_file_compiles($file), "Compile test for $file");
-}
-$test->done_testing();
+$Test::Strict::TEST_SYNTAX   = 1;
+$Test::Strict::TEST_STRICT   = 1;
+$Test::Strict::TEST_WARNINGS = 1;
+$Test::Strict::TEST_SKIP     = [
+    't/data/tests/main.pm',
```
Martchus: Seems like it is executed within `t` so the relative path would just be `data/tests/main.pm`.

perlpunk: Yeah, the problem is that the `chdir "$Bin/..";` has been removed

okurz: done

PR #1310, opened 2019-12-04
Support VNC pseudo encoding 'last rect'

Martchus:
This should avoid problems with Xvnc, see
* https://github.com/TigerVNC/tigervnc/issues/914
* https://progress.opensuse.org/issues/60539

Tested locally with tigervnc/Xvnc 1.10.0. So far I've tested only the test 'sle-15-SP2-Online-ppc64le-Buildrwx788_os-autoinst-distri-opensuse_yast-btrfs_libstorage-ng@rwx788_os-autoinst-distri-opensuse_yast@ppc64le-spvm' (because that job showed the problem also on OSD). The tests no longer incompletes (and fails at some point due to an unrelated issue within the test itself.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -116,6 +116,11 @@ my @encodings = (
         name      => 'VNC_ENCODING_LED_STATE',
         supported => 1,
     },
+    {
+        num       => -224,
+        name      => 'VNC_ENCODING_LAST_RECT',
+        supported => 1,
+    },
 );
 
 sub login {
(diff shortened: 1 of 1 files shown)


consoles/VNC.pm
```diff
@@ -930,8 +935,8 @@ sub _receive_update {
 
     my $do_endian_conversion = $self->_do_endian_conversion;
 
-    foreach (1 .. $number_of_rectangles) {
-        $socket->read(my $data, 12) || die 'unexpected end of data';
+    foreach (my $i = 0; $i < $number_of_rectangles; ++$i) {
```
coolo: note that this only works because number_of_rectangles is unsigned  :)

Martchus: I know, it is 0xFFFF.

PR #1311, opened 2019-12-04
t: Enable 10-terminal.t to be run with prove

perlpunk:
Enabling test harness, for example with prove, requires the test numbers
to be subsequent and in order.

Because this test is running test functions in child processes, the
order is random, and prove will fail.

Usually, Test::SharedFork is used to create ordered test numbers.
But it does not work for this test. Maybe there is more fork magic
going on.

Therefor, we collect the test functions in the child processes via a
shared JSON file and run them when the childs finished.

To see the difference, run this command in master and with this PR:

    (cd t && prove -I$PWD/.. -v 10-terminal.t)

### DIFF
--- a/t/10-terminal.t
+++ b/t/10-terminal.t
@@ -16,14 +16,17 @@ use warnings;
 use Carp 'confess';
 use English -no_match_vars;
 use POSIX qw( :sys_wait_h sigprocmask sigsuspend mkfifo);
-use Fcntl;
+use Fcntl qw( :flock :seek );
 use Socket qw( PF_UNIX SOCK_STREAM sockaddr_un );
 use Time::HiRes 'usleep';
 use File::Temp 'tempfile';
 use Mojo::Log;
+use Mojo::JSON qw( encode_json decode_json );
 
 use Test::More;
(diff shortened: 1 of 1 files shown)


perlpunk:
> I'm only wondering whether we could have the same test coverage with a cheaper test setup as well.

What do you mean?
My suggestion is we switch to using `prove` in the main Makefile, but for that another PR is necessary.
I did the big PR moving everything to the main Makefile, but @okurz suggested to split the changes up, which I think is a good idea.

Martchus:
I mean that this test might be needlessly complicated with all the forking. But I've approved the PR because I'm not sure whether it is actually possible to simplify it and worth the effort.

okurz:
please rebase. By this we can also see if the coverage drop is reproducible

perlpunk:
rebased to master

perlpunk:
I guess the increase is very high now because it couldn't compare to the newest master (the last master build failed because some codecov problem, I restarted dit now)

perlpunk:
Ok, master build succeeded and codecov is uptodate, I will also restart this build now

PR #1313, opened 2019-12-05
ssh_screen: Fix initialize of base class

cfconrad:
The base class consoles::serial_screen changed in commit f9c71d0e. But this
class wasn't updated.

Ticket: https://progress.opensuse.org/issues/60725

I had this good run: http://cfconrad-vm.qa.suse.de/tests/6553

But currently I'm running in this error: http://cfconrad-vm.qa.suse.de/tests/6556#step/update_kernel/34
I'm not sure where it comes from.

### DIFF
--- a/consoles/serial_screen.pm
+++ b/consoles/serial_screen.pm
@@ -27,7 +27,13 @@ our $VERSION;
 
 sub new {
     my ($class, $fd_read, $fd_write) = @_;
-    my $self = bless {class => $class}, $class;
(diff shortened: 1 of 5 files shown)


cfconrad:
@okurz I removed the WIP to early. 
I think I found a solution to that constructor calling of the parent. Plz double check!

Also with this change, the tests would catch a regression we just had. Without it, the tests are more or less useless, or?

PR #1316, opened 2019-12-05
baseclass.t: Add tests for SSH serial methods.

cfconrad:
As mentioned here: https://github.com/os-autoinst/os-autoinst/pull/1298#issuecomment-560339247

### DIFF
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -5,8 +5,10 @@ use warnings;
 use Test::More;
 use Test::MockModule;
 use Test::MockObject;
+use Test::Output 'stdout_is';
 use Test::Warnings;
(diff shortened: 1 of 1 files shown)


t/23-baseclass.t
```diff
@@ -5,16 +5,18 @@ use warnings;
...
+use Mojo::File 'path';
 use Scalar::Util 'refaddr';
 use backend::baseclass;
 use POSIX 'tzset';
 
 # make the test time-zone neutral
 $ENV{TZ} = 'UTC';
 tzset;
-
+bmwqemu::init();
```
okurz: can you add a comment what this is needed for?

PR #1319, opened 2019-12-09
check_ssh_serial: Fix reading from channel

cfconrad:
Commit b2838e23 introduced a error on reading, as it called $chan-read()
twice per loop!

Tickets: https://progress.opensuse.org/issues/60815
VR: http://cfconrad-vm.qa.suse.de/tests/6572


cfconrad:
link PR https://github.com/os-autoinst/os-autoinst/pull/1298

cfconrad:
indeed!

PR #1320, opened 2019-12-10
Remove "use Devel::Cover" from 07-commands.t

perlpunk:
This is probably not needed anymore, as it might have been caused by
Devel::Cover not picking up all module paths by default, and it
was fixed by using the select option in PERL5OPT.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,5 @@ snd2png/test.pnm.md5
 t/*.log
 t/*.trs
 t/pool/*
-t/cover_db/*
 cover_db/*
 .*.swp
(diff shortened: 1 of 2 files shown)


coolo:
Can we just accept that coverage in isotovideo sucks and remove it from reports? :)

okurz:
does this mean we can also remove the according "Devel::Cover" magic we do in openQA tests?

perlpunk:
> does this mean we can also remove the according "Devel::Cover" magic we do in openQA tests?

which magic do you mean? the `Devel::Cover::report() ...` stuff is still needed I think

okurz:
Yeah that's what I meant

PR #1321, opened 2019-12-11
Clarify OFW variable purpose

asmorodskyi:
Current description of OFW variable is confusing - "QEMU Open Firmware is in use" . Because according to https://www.openfirmware.info/Open_Firmware supports x86 , ARM and Power . So person not familiar with actual situation may come up with false conclusions

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -78,7 +78,7 @@ NICTYPE_USER_OPTIONS;string;undef;Arbitrary options for NICTYPE
 NICVLAN;integer;undef;network (vlan) number to which the NIC should be connected, assigned by scheduler to jobs with NICTYPE != user
 NUMDISKS;integer;1;Number of disks to be created and attached to VM
 OFFLINE_SUT;boolean;0;Disable network for a VM
-OFW;boolean;0;QEMU Open Firmware is in use
+OFW;boolean;0;1 means that we running on PowerKVM
 QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64;boolean;undef;If set, for aarch64 systems use VGA as video adapter
 QEMU_DISABLE_SNAPSHOTS;boolean;undef;If set, disable snapshots in case the worker has slow disks to avoid save_vm calls failing due to timeouts (See https://bugzilla.suse.com/show_bug.cgi?id=1035453[bsc#1035453])
 PATHCNT;integer;2;Number of paths in MULTIPATH scenario


asmorodskyi:
@okurz  , @nicksinger  - WDYT ?

asmorodskyi:
@pevik , @czerw - FYI

doc/backend_vars.asciidoc
```diff
@@ -78,7 +78,7 @@ NICTYPE_USER_OPTIONS;string;undef;Arbitrary options for NICTYPE
 NICVLAN;integer;undef;network (vlan) number to which the NIC should be connected, assigned by scheduler to jobs with NICTYPE != user
 NUMDISKS;integer;1;Number of disks to be created and attached to VM
 OFFLINE_SUT;boolean;0;Disable network for a VM
-OFW;boolean;0;QEMU Open Firmware is in use
+OFW;boolean;0;1 means that we running on PowerKVM
```
okurz: Can you describe the variable in imperative mood what one can control with it?

pevik: I'd keep also original comment `QEMU Open Firmware is in use`, because that's still valid (the purpose of this variable).
BTW according to https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/8e30d7c02507d382eb70735aff5dc540c797e300 `OFW` was used to detect `ppc64 || ppc64le`. But this made things rather cryptic. I'd prefer to have some more meaningful variable `ppc64_all` or something.

asmorodskyi: > Can you describe the variable in imperative mood what one can control with it?

At least in my understanding this variable controls nothing . It is just indicator that you are running on PowerKVM .  It is like "Can you describe what you control with `ARCH='x86_64'` ? "

asmorodskyi: > I'd keep also original comment `QEMU Open Firmware is in use`, because that's still valid (the purpose of this variable).

What OFW stands for ? Obviously it stands for Openfirmware . What OFW means from perspective of os-autoinst / tests code obviously something different. And in my understanding this file should answer on second question not the first one .  

> BTW according to [os-autoinst/os-autoinst-distri-opensuse@8e30d7c](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/8e30d7c02507d382eb70735aff5dc540c797e300) `OFW` was used to detect `ppc64 || ppc64le`. But this made things rather cryptic. I'd prefer to have some more meaningful variable `ppc64_all` or something.

My personal opinion that this variable should be dropped . But this PR will not go so deep I just tried to clarify what we **currently** have in documentation

asmorodskyi:
after reading conversations on PR referenced by Petr , I think it worth to ask @SergioAtSUSE  how he would state definition of OFW variable **in this project**

doc/backend_vars.asciidoc
SergioAtSUSE: I agree with @pevik. But, I was asked by @coolo to not use a regexp to detect ppc* (performance) and instead abuse that setting to "detect" ppc. He mentioned that he would not have anything against when the times comes that OFW is not only use for PPC. So, the meaning of OFW is to detect OpenFirmware and has to continue to be so.
Maybe your case is enough to convince @coolo.

asmorodskyi: but it does not mentioned **anywhere** that OpenFirmware is used **only** in PowerKVM . Actually the opposite - everywhere I have seen statements that OpenFirmware supports a lot of different architectures , also I found even that exactly in case of PowerKVM default BIOS used is not OpenFirmware but Open Hack Ware which is just OFW-compatible.  Also we have PowerVM  which according to @nicksinger  using OpenFirmware too but all jobs which using spvm backend has no OFW variable defined . 
So can you please clarify for me what in terms of openQA OFW variable means ? Because I can not agree that it means openfirmware .

SergioAtSUSE: `OFW=1` should mean, and only mean, that the machine is using Openfirmware. I am against changing that definition.

The reality is that the setting is misused in the code, but this should not change the description of the setting.
I would approve a PR that replaces all the misused places of `check_var('OFW'...` to `check_var('ppc...`

czerw: I would say new function `is_powerkvm` (or similar) would be better than `check_var`. Anyway it looks like nice continuation of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/4477.

PR #1322, opened 2019-12-11
Escape parameters in log calls

perlpunk:
Otherwise the log will have linebreaks like this:

    [Thu Dec 14 04:10:45 2017] [12032:debug] <<< testapi::type_string(string='root
    ', max_interval=250, wait_screen_changes=0, wait_still_screen=0)

Also other unprintable characters can make the log unreadable ("\r", ...)

Issue: https://progress.opensuse.org/issues/29399


okurz:
please crosscheck the significant coverage drop. And I would appreciate an example or test.

perlpunk:
I added a test. Coverage was ok on my forked travis test and looks good here also.

PR #1323, opened 2019-12-13
Fix undefined needle dir in dynamic needle downloader with proper variable names

okurz:
It seems a subtle, unintended difference in variable names caused
problems within the openQA developer mode when one tries to reload the
needles that had just been created with the needle editor during the
same run. There was a warning

```
Use of uninitialized value $needle_dir in concatenation (.) or string at /usr/lib/os-autoinst/OpenQA/Isotovideo/NeedleDownloader.pm line 76.
```

which did not cause any more severe effects so the error could not be
seen earlier unfortunately.

Related progress issue: https://progress.opensuse.org/issues/61019

### DIFF
--- a/OpenQA/Isotovideo/NeedleDownloader.pm
+++ b/OpenQA/Isotovideo/NeedleDownloader.pm
@@ -72,8 +72,8 @@ sub _add_download {
 
     my $needle_name     = $needle->{name};
     my $latest_update   = $needle->{t_updated};
-    my $needle_dir      = $needle::needles_dir;
-    my $download_target = "$needle_dir/$needle_name.$extension";
+    my $needles_dir     = $needle::needles_dir;
+    my $download_target = "$needles_dir/$needle_name.$extension";
 
     if (my $target_stat = stat($download_target)) {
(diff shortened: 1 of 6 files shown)


okurz:
> I'm wondering why this is just a warning an not an error. Usually even Perl is better...

This is exactly still the challenge I have. I want to come up with a proper test for that but haven't learned how to do it properly yet.

okurz:
Please check again as I overlooked some instances of different variable name in other files. I went with the name "needles_dir" now to be also consistent with the test variable NEEDLES_DIR

perlpunk:
> I'm wondering why this is just a warning an not an error. Usually even Perl is better...

We could:

    use warnings "FATAL";

;-)

Martchus:
> Please check again as I overlooked some instances of different variable name in other files.

There are no more references to `$needle::needledir` or `$needledir` so it seems you've covered all places.

What are the disadvantages of `use warnings "FATAL";`?

okurz:
I will merge as is and see about further improvements later. I agree with making warnings fatal. For this we have https://progress.opensuse.org/issues/10704 :)

PR #1324, opened 2019-12-13
Use getter/setter for needles dir as safer approach rather than package variable

okurz:


perlpunk:
Now `$needles_dir` should be declared with `my` instead of `our` because you otherwise can still refer to it via the package name.

okurz:
> Now $needles_dir should be declared with my instead of our

yes, was already in the last update of the PR

PR #1325, opened 2019-12-13
Update to new perltidy version 20191203

okurz:


okurz:
good that you ask. I have created the according PR and I think we actually need os-autoinst updated first as I os-autoinst is checked out from master in the tests' travis CI test needing the updated cpanfile of os-autoinst.

PR #1330, opened 2019-12-20
Avoid qw() warning message in t/18-qemu.t

Amrysliu:
See: https://progress.opensuse.org/issues/61167


okurz:
@kraih @kalikiana @perlpunk WDYT? I am not sure if the new style makes it easier to read but it fixes the perl warning "Possible attempt to separate words with commas"

PR #1332, opened 2019-12-25
Fix default boot menu timeout in qemu backend

marmarek:
'5000' default should apply to the timeout only, not to the whole qemu option.


coolo:
thx

PR #1333, opened 2019-12-27
Record the spent time of each job module

Amrysliu:
See: https://progress.opensuse.org/issues/57419

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -341,6 +341,7 @@ sub runalltests {
     bmwqemu::diag "Snapshots are " . ($snapshots_supported ? '' : 'not ') . "supported";
 
     write_test_order();
+    my %modules_spent_time;
 
     for my $t (@testorder) {
         my $flags    = $t->test_flags();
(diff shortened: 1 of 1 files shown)


autotest.pm
```diff
@@ -407,6 +410,7 @@ sub runalltests {
             $t->save_test_result();
         }
     }
+    bmwqemu::save_json_file(\%modules_spent_time, "modules_spent_time.json") if (%modules_spent_time);
```
okurz: does this save a json result file an additional time? I assume the result file is already saved in another location. Would be good to not save an additional time when this is not necessary.

Amrysliu: yes, those spent times will be saved in a json file. Do you mean I should save the spent time in the `detail-xxxx.json` file?

okurz:
On Monday, 30 December 2019 03.01.34 CET LiuXiaojing wrote:
> Do you mean I should save the spent time in the `detail-xxxx.json` file?

yes, sounds reasonable.

autotest.pm
Martchus: Yes, but it is a little bit complicated, see https://github.com/os-autoinst/openQA/pull/2629#pullrequestreview-339219736.

Amrysliu:
it's not a good solution, so close it.

Martchus:
You don't have to close your PRs when a change is required. Just amend and force push your new version.

PR #1335, opened 2020-01-10
Show the execution time of job modules

Amrysliu:
In order to add the execution time of job modules in details-xxx.json
file, changing the top level of the files from ARRAY to HASH

Ticket: https://progress.opensuse.org/issues/57419

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -43,7 +43,7 @@ sub new {
     $category ||= 'unknown';
     my $self = {class => $class};
     $self->{lastscreenshot}         = undef;
-    $self->{details}                = [];
+    $self->{details}                = {};
     $self->{result}                 = undef;
     $self->{running}                = 0;
     $self->{category}               = $category;
@@ -186,7 +186,7 @@ sub record_screenmatch {
 
     $self->{result} ||= 'ok';
 
-    push @{$self->{details}}, $result;
+    push @{$self->{details}->{results}}, $result;
     return $result;
 }
 
@@ -253,14 +253,14 @@ sub record_screenfail {
 
     $self->{result} = $overall if $overall;
 
-    push @{$self->{details}}, $result;
+    push @{$self->{details}->{results}}, $result;
     return $result;
 }
 
 sub remove_last_result {
     my $self = shift;
     --$self->{test_count};
-    return pop @{$self->{details}};
+    return pop @{$self->{details}->{results}};
 }
 
 sub details {
@@ -360,7 +360,7 @@ sub runtest {
 
         $self->{result} = 'fail';
         # add a fail screenshot in case there is none
-        if (!@{$self->{details}} || ($self->{details}->[-1]->{result} || '') ne 'fail') {
+        if (!@{$self->{details}->{results}} || ($self->{details}->{results}->[-1]->{result} || '') ne 'fail') {
             $self->take_screenshot();
         }
         # show a text result with the die message unless the die was internally generated
@@ -387,8 +387,10 @@ sub runtest {
         $self->run_post_fail("test $name failed");
     }
 
+    my $execution_time = time - $starttime;
+    $self->{details}->{execution_time} = $execution_time;
     $self->done();
-    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), time - $starttime));
+    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), $execution_time));
     return;
 }
 
@@ -441,7 +443,7 @@ sub record_resultfile {
         result => $nargs{result},
         text   => $filename,
     };
-    push @{$self->{details}}, $detail;
+    push @{$self->{details}->{results}}, $detail;
     $self->write_resultfile($filename, $output);
 }
 
@@ -515,7 +517,7 @@ sub record_testresult {
 
     # add detail
     my $detail = {result => $result};
-    push(@{$self->{details}}, $detail);
+    push(@{$self->{details}->{results}}, $detail);
     ++$self->{test_count};
     return $detail;
 }
@@ -583,7 +585,7 @@ sub stop_audiocapture {
         result => 'unk',
     };
 
-    push @{$self->{details}}, $result;
+    push @{$self->{details}->{results}}, $result;
 
     return $result;
 }
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 19;
+our $version = 20;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -204,13 +204,13 @@ subtest 'send_key with wait_screen_change' => sub {
 
 is($autotest::current_test->{dents}, 0, 'no soft failures so far');
 stderr_like(\&record_soft_failure, qr/record_soft_failure\(reason=undef\)/, 'soft failure recorded in log');
-is($autotest::current_test->{dents},             1, 'one dent recorded');
-is(scalar @{$autotest::current_test->{details}}, 1, 'exactly one detail added recorded');
+is($autotest::current_test->{dents},                        1, 'one dent recorded');
+is(scalar @{$autotest::current_test->{details}->{results}}, 1, 'exactly one detail added recorded');
 
 stderr_like(sub { record_soft_failure('workaround for bug#1234') }, qr/record_soft_failure.*reason=.*workaround for bug#1234.*/, 'soft failure with reason');
-is($autotest::current_test->{dents},             2, 'one more dent recorded');
-is(scalar @{$autotest::current_test->{details}}, 2, 'exactly one more detail added recorded');
-my $details    = $autotest::current_test->{details}[-1];
+is($autotest::current_test->{dents},                        2, 'one more dent recorded');
+is(scalar @{$autotest::current_test->{details}->{results}}, 2, 'exactly one more detail added recorded');
+my $details    = $autotest::current_test->{details}->{results}[-1];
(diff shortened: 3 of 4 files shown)


basetest.pm
```diff
@@ -387,6 +387,7 @@ sub runtest {
         $self->run_post_fail("test $name failed");
     }
 
+    $self->{details}->{execution_time} = time - $starttime . ' s';
```
perlpunk: Just a minor thing: I would rather calculate the elapsed time once, and then use it here and two lines later.
So it is ensured it shows the exact same number of seconds.

basetest.pm
```diff
@@ -387,8 +387,10 @@ sub runtest {
         $self->run_post_fail("test $name failed");
     }
 
+    my $execution_time = time - $starttime;
+    $self->{details}->{execution_time} = $execution_time . ' s';
```
kalikiana: I feel like I'd expect the value to be the seconds as a number. Turning it into a string for display purposes belongs on the display side of things.

Amrysliu: fixed

kalikiana: Thanks!

Martchus:
That's why I added the 'not-yet' tag. Not sure about the API version, though. As far as I know `$OpenQA::Isotovideo::Interface::version` is about the test API but not related to the web UI. And increasing that version would in fact not change the web UI's behavior (so it would show a warning or something). So with the current versioning facilities it seems the best we can do is to wait a reasonable to merge this PR after the web UI PR has been merged.

Amrysliu:
@Martchus @okurz is there something else I need to do for this pull request? Could I remove the [WIP], and remain the `not-yet` tag that Martchus added.

Martchus:
I tested this last week locally and it didn't work. I'll test it again, maybe it was just a local configuration issue.

Martchus:
It really doesn't work. My web UI has your commit 544b4da9e62dc5168809fed091128c5d2abbcf0d.

But it seems that the openQA/worker-side is at fault. At least the `result-*.json` file produced by os-autoinst looks correct:

```
{
   "dents" : 0,
   "details" : {
      "execution_time" : 123,
      "results" : [
```

openQA is also able to parse the new structure but the `execution_time` seems to be discarded again at some point so the `details-*.json` file within the test results doesn't contain it anymore.

Martchus:
I was curious whether the rest works so I've just fixed the problems on the openQA-side (will submit a PR). Seems to work :-)

okurz:
IMHO a label is not visible enough to prevent premature merging. As long as the git commit message itself has "[WIP]" included I guess it is ok if the PR subject also includes this :)

I would be really careful changing the format and would like others to provide their feedback as well.

@AdamWill @coolo @aaannz @kalikiana @perlpunk @kraih can you also take a look please?

kraih:
@okurz Agreed, the little work in progress label has annoyed me quite a few times before. `[WIP]` in the PR title is better.

Amrysliu:
> I was curious whether the rest works so I've just fixed the problems on the openQA-side (see [os-autoinst/openQA#2680](https://github.com/os-autoinst/openQA/pull/2680)). The rest seems to work :-)

Thanks for your test and fixing. 
I did run a test case before I committed those two pull requests (openQA and os-autoinst). However, I did *not* run a test case after I modified the code while doing unit test. This is my fault. I will notice that next time. 
And thanks a lot for doing this test and fixing the bug.

AdamWill:
Uh. I don't *think* this would affect us at all, but I'm not sure, I'm not really familiar with this code. It seems like something that's mostly just passed from os-autoinst to openQA and not publicly really exposed? Is there some non-obvious implication here?

Martchus:
It means that you can not update os-autoinst without updating openQA, too. The required changes on the openQA-side were merged a few days ago (https://github.com/os-autoinst/openQA/pull/2654).

AdamWill:
OK, yeah, that makes sense. It's not a problem for us, though - in fact I generally update them together. I guess it would be nice if this was indicated in the RPM specs somehow, it's a good way to communicate it - maybe bump the minor version of both os-autoinst and openqa, and have openqa `Requires` the new os-autoinst minor, and os-autoinst `Conflicts` with any openqa below the new minor?

Martchus:
It is not only the openQA worker but also the web UI which needs to be updated first. Web UI and os-autoinst aren't necessarily installed on the same machine so version constraints on package level wouldn't help.

okurz:
IMHO we should still bump the version in OpenQA/Isotovideo/Interface.pm as it says in a comment "version of the test API and the API relevant to the worker". I guess this applies here :)

Amrysliu:
@Martchus @okurz  I have increased the version number and removed the WIP tag. Please have a review again. Thanks.

Amrysliu:
> I've already tested it locally at some point and it worked. The web UI side also has been updated for quite a while now. But there's still an improvement you can do on the web UI side: [os-autoinst/openQA#2654 (comment)](https://github.com/os-autoinst/openQA/pull/2654#issuecomment-575066593)

Thanks for your review. I will do this improvement.

okurz:
can you please also do at least part of what https://github.com/os-autoinst/os-autoinst/pull/1335#issuecomment-577175354 suggests:

> bump the minor version of both os-autoinst and openqa, and have openqa-worker Requires the new os-autoinst minor, and os-autoinst Conflicts with any openqa-common below the new minor,

So in the file dist/rpm/os-autoinst.spec set version from 4.6 to 4.7 .

okurz:
sorry, wrong button :)

okurz:
Waiting for https://github.com/os-autoinst/openQA/pull/2725

okurz:
@Martchus so the openQA PRs are in for long. I guess we can merge now. WDYT?

mdoucha:
FYI: This API change broke LTP tests. https://openqa.suse.de/tests/overview?distri=sle&version=12-SP4&build=4.12.14-474.1.gcb202a6&groupid=214

What's the point of munging `$test->{details}` instead of adding `$test->{execution_time}`? It'd be less work on both ends of OpenQA and it wouldn't break any existing QA code.

perlpunk:
> What's the point of munging `$test->{details}` instead of adding `$test->{execution_time}`?

The problem is that the data structure was an array before. Only by changing it to a hash, we were able to add a new item.

Martchus:
Yes, before it was just an array on the top-level. Nevertheless it is likely true that it would have been better to add the execution time at the top-level and not within `details`. In any case, the breaking change couldn't have been avoided. Can you point me to an exact example of a job which doesn't work anymore because of this? I'm asking because the web UI should still support the old format.

perlpunk:
> Nevertheless it is likely true that it would have been better to add the execution time at the top-level and not within `details`.

Oh, I actually wasn't aware of that. Yes, it should have been toplevel...

mdoucha:
> In any case, the breaking change couldn't have been avoided.

Yes, it could have. Here's an alternative patch that'd achieve the same result without making any backwards-incompatible changes to the API:

    diff --git a/basetest.pm b/basetest.pm
    index 6ae0df81..2856f8d1 100644
    --- a/basetest.pm
    +++ b/basetest.pm
    @@ -58,6 +58,7 @@ sub new {
         $self->{serial_failures}        = [];
         $self->{autoinst_failures}      = [];
         $self->{fatal_failure}          = 0;
    +    $self->{execution_time}         = 0;
         return bless $self, $class;
     }
     
    @@ -387,8 +388,9 @@ sub runtest {
             $self->run_post_fail("test $name failed");
         }
     
    +    $self->{execution_time} = time - $starttime;
         $self->done();
    -    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), time - $starttime));
    +    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), $self->{execution_time}));
         return;
     }
     
    @@ -396,9 +398,10 @@ sub save_test_result {
         my ($self) = @_;
     
         my $result = {
    -        details => $self->details(),
    -        result  => $self->result(),
    -        dents   => $self->{dents},
    +        details        => $self->details(),
    +        result         => $self->result(),
    +        dents          => $self->{dents},
    +        execution_time => $self->{execution_time},
         };
         $result->{extra_test_results} = $self->{extra_test_results} if $self->{extra_test_results};

> Can you point me to an exact example of a job which doesn't work anymore because of this? I'm asking because the web UI should still support the old format.

Here's a job group with 126 failed tests, all broken by this API change. Pick any one you like:
https://openqa.suse.de/tests/overview?distri=sle&version=12-SP4&build=4.12.14-474.1.gcb202a6&groupid=214

perlpunk:
Since showing the execution time is not a critical thing, we could actually fix that easily by putting it at the toplevel, right?

Martchus:
> Yes, it could have.

I was talking about the structure of the JSON document as a whole. Previously the root element was an array and that had to be changed to be a hash. But as I said, you're right that details could have kept as an array.

@Amrysliu Do you want to take care of this problem (as it is caused by your changes)? Otherwise me or @perlpunk can proceed.

> we could actually fix that easily by putting it at the toplevel, right?

I'd assume that would work.

okurz:
@Martchus @perlpunk maybe you have time today to do the change quickly so that we can have a fix ready sooner. But I fear we would need a corresponding change in openQA anyway. Should we better revert this and give corresponding os-autoinst+openQA changes more time? os-autoinst-distri-opensuse is important but os-autoinst+openQA compatibility should be more important for us than run_ltp relying on internal data formats of os-autoinst. A fix on test side should be faster

kraih:
Yea. I would revert and then go with the top-level solution.

Martchus:
I'm actually already over it because it would be good to fix this as soon as possible indeed.

So I've been drafting a PR to move the time to the top-level. However, that turned out to be mainly a revert of this PR so I've just been creating a revert now. We should merge it as soon as possible to and rework adding the execution time later.

By the way, this JSON document has in fact always been using a hash as root but only on the os-autoinst side. The openQA-side only takes the `details` producing files with an array as top-level element. So the openQA-side was actually the limitation here.

That means the openQA side needs to be improved as well. It should read the `execution_time` from the top-level and it would likely be the best if openQA would just store the entire result as provided by os-autoinst instead of storing just a part of it. That would make things less confusing.

Martchus:
To avoid confusion we should also document the file structure. This was an attempt to document it but only covered the array-ish openQA part: https://github.com/Martchus/openQA-helper/blob/master/result-format.md#structure-of-test-result-directory

PR #1337, opened 2020-01-11
qemu: Allow to configure maximum file migration time

okurz:
Related progress issue: https://progress.opensuse.org/issues/58013


doc/backend_vars.asciidoc
```diff
@@ -93,6 +93,7 @@ QEMUTHREADS;integer;0;Number of cpu threads used by VM
 QEMUVGA;see qemu -device ?;cirrus;VGA device to use with VM
 QEMU_COMPRESS_QCOW2;boolean;1;compress qcow2 images intended for upload
 QEMU_HUGE_PAGES_PATH;string;undef;Define a path to use huge pages (e.g. /dev/hugepages/)
+QEMU_MAX_MIGRATION_TIME;integer;240;Maximum time a migration to file may take for example for snapshot creation before being forcefully aborted.
```
Martchus: Maybe it is worth mentioning that this is measured in seconds.

okurz: absolutely. I forgot to mention the unit. Fixed that and will merge.

Martchus:
Looks like the change has been pushed directly on master.

okurz:
yes it was. Should have merged instead. However the change is in.

PR #1338, opened 2020-01-15
virtio_terminal: Don't force max PIPE_SZ

cfconrad:
Sometime (the reason why isn't known) the PIPE_SZ syscall fail.

[2020-01-14T03:06:11.813 CET] [debug] Backend process died, backend errors are reported below in the following lines:
Can't fcntl($fh, '1031', '1048576'): Operation not permitted at /usr/lib/os-autoinst/consoles/virtio_terminal.pm line 147

So we can be more gentle on setting new PIPE_SZ and retry with lower
values.

Ticket: https://progress.opensuse.org/issues/62108

### DIFF
--- a/t/10-virtio_terminal.t
+++ b/t/10-virtio_terminal.t
@@ -1,0 +1,115 @@
+#!/usr/bin/perl
+# Copyright © 2020 SUSE LLC
+
+use strict;
+use warnings;
+
+use Test::More;
+use Test::Exception;
+use Test::MockModule;
+use Test::Output;
+use POSIX 'mkfifo';
+
+use consoles::virtio_terminal;
+use testapi;
+use bmwqemu;
+
+
+sub prepare_pipes
+{
+    my ($socket_path) = @_;
+    my $pipe_in       = $socket_path . ".in";
+    my $pipe_out      = $socket_path . ".out";
+
+    for (($pipe_in, $pipe_out)) {
+        unlink $_ if (-e $_);
+        mkfifo($_, 0666) or die("Cannot create fifo pipe $_");
+    }
+
+    my $pid = fork || do {
+        my $running = 1;
+        $SIG{USR1} = sub { $running = 0; };
+        $SIG{ALRM} = sub {
+            die('Timeout for pipe other side helper');
+        };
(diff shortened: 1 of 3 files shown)


okurz:
still "WIP"?

cfconrad:
Sorry, I somehow lost track of this. 

I remember my last problem was to write a test with mocking the system function `fcntl()`, but I failed. Any Idea?

I will now continue and remove the WIP afterwards.

okurz:
I don't know. Maybe the section "Mocking + Export" in https://metacpan.org/pod/Test::MockModule#METHODS helps for mocking fcntl, similar to strftime?

cfconrad:
I tried hard but wasn't able to mock `fcntl()` reasonable.
Best solution I had was:
```
my $mock_fcntl     = sub { say("mocked fcntl: " . join(",", @_)); return undef; };

BEGIN {
    *CORE::GLOBAL::fcntl      = sub { $mock_fcntl->(@_); };
}
```
but when I added then `use Test::More;` I got compile errors, which I wasn't able to solve :(

So I worked around this...

t/10-virtio_terminal.t
```diff
@@ -0,0 +1,115 @@
...
+
+    testapi::set_var('VIRTIO_CONSOLE_PIPE_SZ', 5555);
+    $helper = prepare_pipes($socket_path);
+    $term   = consoles::virtio_terminal->new('unit-test-console', {socked_path => $socket_path});
+    stderr_like { $term->open_pipe() } qr/Set PIPE_SZ from 1024 to 5555/, 'Log mention new size';
+    cleanup_pipes($helper);
+    is($size, 5555, "PIPE_SZ is 5555 from VIRTIO_CONSOLE_PIPE_SZ");
+
+    $term = consoles::virtio_terminal->new('unit-test-console', {socked_path => $socket_path});
+    throws_ok { $term->open_pipe() } qr/No such file or directory/, "Throw execption if pipe doesn't exists";
```
okurz: typo in "execption" :)

cfconrad: thx done.

PR #1340, opened 2020-01-17
qemu: handle armv6/7 architectures properly

ggardet:


ggardet:
Not sure why `18-backend-qemu.t` fails since the default value remains the same. 
The legacy `armv5` arch is not supported by openSUSE, and currently only `armv7` is tested in openQA. Maybe `armv6` later.

backend/qemu.pm
```diff
@@ -710,6 +710,7 @@ sub start_qemu {
     my $arch_supports_boot_order = $vars->{UEFI} ? 0 : 1;    # UEFI/OVMF supports ",bootindex=N", but not "-boot order=X"
     my $use_usb_kbd;
     my $arch = $vars->{ARCH} // '';
+    $arch = 'arm' if ($vars->{ARCH} =~ /armv6|armv7/);
```
Martchus: The test likely fails because `$vars->{ARCH}` might be uninitialized here producing a warning. This likely should have been just `$arch`.

ggardet: Yes, I noticed it afterwards, I just pushed a fix.

PR #1341, opened 2020-01-23
Update tidy to 20200110

Martchus:


okurz:
Can you please also provide updates to our common test distributions? You know, same procedure as for every tidy update :)

Martchus:
I have already provided https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/9424 (the CI likely fails there because it relies on the PR to be merged first). Not sure what other distributions you're talking about (as far as I know the ones found under the os-autoinst organization don't use tidy).

okurz:
Yes, it's ok like this. Doesn't harm to adjust the style of the other repos though regardless if there is an automatic tidy check.

PR #1344, opened 2020-01-28
Allow to configure no drives for drive-less systems

okurz:


okurz:
yeah, configuring NUMDISKS=0 for backend qemu and checking the command line for no drives. Didn't check other backends however.

PR #1347, opened 2020-01-30
Add "assert_still_screen" test API function

okurz:
Related progress issue: https://progress.opensuse.org/issues/39116

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -59,7 +59,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
 
   upload_asset data_url check_shutdown assert_shutdown parse_junit_log parse_extra_log upload_logs
 
-  wait_screen_change assert_screen_change wait_still_screen wait_serial
+  wait_screen_change assert_screen_change wait_still_screen assert_still_screen wait_serial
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -714,6 +714,19 @@ sub wait_still_screen {
...
+
+  assert_still_screen([$args...])
+
+Run C<wait_still_screen> but C<die> if screen changed within timeout. Look
+into C<wait_still_screen> for details.
+
+=cut
+
+sub assert_still_screen(@) {
+    wait_still_screen(@_) or die 'assert_still_screen failed to detect a still screen';
```
Martchus: In other cases where we assert things we just record a result and set the overall test result to failed. Exceptions are usually thrown when there's a problem with the test code or backend itself and I assume they lead to an "incomplete" result. So have you tested this in a "fullstack-way" to see whether your change really produces the right *overall* behavior?

okurz: That's the same approach as for assert_screen_change. Any exception in test scope is handled properly to forward as test failure, but not when just calling the function in test directly

Martchus:
But tidy doesn't like it.

PR #1353, opened 2020-02-21
generalhw: Allow GENERAL_HW_SOL_CMD to get args with GENERAL_HW_SOL_ARGS

ggardet:


backend/generalhw.pm
```diff
@@ -46,6 +46,7 @@ sub get_cmd {
     }
 
     my $args = get_var('GENERAL_HW_FLASH_ARGS') if ($cmd eq 'GENERAL_HW_FLASH_CMD' and get_var('GENERAL_HW_FLASH_ARGS'));
+    $args = get_var('GENERAL_HW_SOL_ARGS') if ($cmd eq 'GENERAL_HW_SOL_CMD' and get_var('GENERAL_HW_SOL_ARGS'));
```
okurz: This is intended to overwrite GENERAL_HW_FLASH_ARGS, yes?

ggardet: There is no overwrite as it checks `$cmd`.

okurz: fine

PR #1355, opened 2020-02-25
Improve bwmqemu::log_call

perlpunk:
If called with only one parameter, don't try to log like key/vale pairs

Issue: https://progress.opensuse.org/issues/63808


okurz:
in commit message s/vale/value, rest fine

perlpunk:
@okurz fixed commit message

PR #1357, opened 2020-03-02
Fix serialdev configuration for pvm_hmc poo#64105 

michelmno:
I validated on my openQA instance in IBM intranet
for Tumbleweed and Leap15.2 ppc64le.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -117,7 +117,7 @@ Used for internal initialization, do not call from tests.
 =cut
 
 sub init {
-    if (get_var('OFW') || check_var('BACKEND', 's390x')) {
+    if (get_var('OFW') || get_var('BACKEND', '') =~ /s390x|pvm_hmc/) {
         $serialdev = "hvc0";
     }
     elsif (get_var('SERIALDEV')) {


testapi.pm
```diff
@@ -117,7 +117,8 @@ Used for internal initialization, do not call from tests.
 =cut
 
 sub init {
-    if (get_var('OFW') || check_var('BACKEND', 's390x')) {
+    if (get_var('OFW') || check_var('BACKEND', 's390x') ||
+        check_var('BACKEND', 'pvm_hmc')) {
```
okurz: Please combine to

```perl
if (get_var('OFW') || get_var('BACKEND', '') =~ /s390x|pvm_hmc/)) {
```

okurz:
thanks. Code looks ok but please change the git commit message subject line to explain the "why".

michelmno:
> thanks. Code looks ok but please change the git commit message subject line to explain the "why".

I did not changed the subject line, but explained the "why" in the commit body.
I did that because I  do not see a "why" in subject line of previous accepted commits.

Martchus:
I actually prefer to have the "what" in the subject line and the "why" in the body because so I'm fine with that. Usually the "why" is too much for the subject line anyways.

Martchus:
The question of "what" and "why" is independent of whether it is written from the user perspective. So one can write "what" this change does from the user perspective.

michelmno:
> @michelmno maybe others have not followed the approach. Nevertheless I would really much appreciate if you update the git commit message subject line to write from the users perspective. Using `git commit --amend` or `git rebase --interactive` this should be trivial to do and then I will merge your PR right away.

I just did the "why" in commit subject, even if I disagree.

Martchus:
I disagree as well. Now the subject line looks like a random statement on its own. However, if you would just prepend a "Fix " it would be better. It would describe *what* this commit changes from the user perspective. The *why* is given implicitly because "Fix" implies that the previous behavior is considered wrong. So @okurz should be fine.

okurz:
Yes, I am very sorry if I could not make myself clear. Please still state the git commit message subject line in imperative mood. "Fix" as a prefix is perfectly fine. Just imagine the following: As a user I read "Set serialdev to hvc0 for pvm_hmc backend" and I ask myself: Ok, why should I care? How does this impact or help me?. If you instead write "Fix serialdev configuration for pvm_hmc" then as a user I would say: Ok, if I want to use pvm_hmc I can probably benefit from this change.

PR #1358, opened 2020-03-05
Add support to checkout git refspec in existing working copies

okurz:
The test variable CASEDIR already provides the possibility to clone a
git repository and checkout a git repository. However this can be a
costly operation. When the provided CASEDIR is already a git working
copy we can instead simply checkout a commit within that working copy.

Related progress issue: https://progress.opensuse.org/issues/60272

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -17,10 +17,12 @@ package OpenQA::Isotovideo::Utils;
 use Mojo::Base -base;
 use Mojo::URL;
 
+use Exporter 'import';
 use File::Spec;
 use Cwd;
-use testapi 'diag';
-use bmwqemu;
+use bmwqemu 'diag';
+
+our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec);
 
 sub calculate_git_hash {
     my ($git_repo_dir) = @_;
@@ -69,6 +71,30 @@ sub checkout_git_repo_and_branch {
         diag "Skipping to clone '$clone_url'; $local_path already exists";
     }
     return $bmwqemu::vars{$dir_variable} = File::Spec->rel2abs($local_path);
+}
+
+=head2 checkout_git_refspec
+
+    checkout_git_refspec($dir, $refspec_variable);
+
+Takes a git working copy directory path and checks out a git refspec specified
+in a git hash test parameter if possible. Returns the determined git hash in
+any case, also if C<$refspec> was not specified or is not defined.
+
+Example:
+
+    checkout_git_refspec('/path/to/casedir', 'TEST_GIT_REFSPEC');
+
+=cut
+sub checkout_git_refspec {
+    my ($dir, $refspec_variable) = @_;
+    return undef unless defined $dir;
+    if (my $refspec = $bmwqemu::vars{$refspec_variable}) {
+        diag "Checking out local git refspec '$refspec' in '$dir'";
+        qx{env git -C $dir checkout -q $refspec};
+        die "Failed to checkout '$refspec' in '$dir'\n" unless $? == 0;
+    }
+    calculate_git_hash($dir);
 }
 
 1;
--- a/isotovideo
+++ b/isotovideo
@@ -84,7 +84,7 @@ use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 Getopt::Long::Configure("no_ignore_case");
 use OpenQA::Isotovideo::CommandHandler;
 use OpenQA::Isotovideo::Interface;
-use OpenQA::Isotovideo::Utils;
+use OpenQA::Isotovideo::Utils qw(checkout_git_repo_and_branch checkout_git_refspec);
 
 session->enable;
 session->enable_subreaper;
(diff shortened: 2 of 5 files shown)


okurz:
This works perfectly fine within the scope of os-autoinst but for openQA I am not so sure because we either have a shared test dir for all tests or we have a copy in the worker local cache but shared for all instances. We should try if we can clone locally from cache to pool using git and then checkout. Hm, but when we already specify CASEDIR with a repo we need to know the cache path. We do similar magic already for needle dirs within openqa-clone-custom-git-refspec. Maybe for the time being we can live with this approach as well, tweaking shell wrapper scripts.

Further thoughts?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -71,4 +74,28 @@ sub checkout_git_repo_and_branch {
...
+Example:
+
+    checkout_git_refspec('/path/to/casedir', 'TEST_GIT_REFSPEC');
+
+=cut
+sub checkout_git_refspec {
+    my ($dir, $refspec_variable) = @_;
+    return undef unless defined $dir;
+    if (my $refspec = $bmwqemu::vars{$refspec_variable}) {
+
```
Martchus: I'd remove that blank line.

isotovideo
```diff
@@ -302,7 +303,7 @@ if ($bmwqemu::vars{_EXIT_AFTER_SCHEDULE}) {
 testapi::init();
 
 # make sure the needles are initialized
-$bmwqemu::vars{NEEDLES_GIT_HASH} = OpenQA::Isotovideo::Utils::calculate_git_hash(needle::init);
+$bmwqemu::vars{NEEDLES_GIT_HASH} = checkout_git_refspec(needle::init => 'NEEDLES_GIT_REFSPEC');
```
Martchus: This initializes the needle directory and then switches the branch. It would make more sense to do this in the opposite order. To change the order you likely need to split `needle::init` into `needle::determine_needle_dir` and `needle::scan_needle_dir`.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -71,4 +74,28 @@ sub checkout_git_repo_and_branch {
...
+        diag "Checking out local git refspec '$refspec' in '$dir'";
+        qx{env git -C $dir checkout -q $refspec};
```
Martchus: Have you tested what happens when this Git command fails?

OpenQA/Isotovideo/Utils.pm
okurz: yeah, me too :)

isotovideo
okurz: I have a better idea. I just moved the part *within* needle::init. This also makes isotovideo look nicer and cleaner as we have `needle::init` just after `testapi::init` :)

OpenQA/Isotovideo/Utils.pm
okurz: Yes, for sure. I thought about having it ignore if the checkout fails until we have gathered some experience but that's not a good idea. I added a `die` now after the call if the command fails.

Martchus:
> This works perfectly fine within the scope of os-autoinst but for openQA I am not so sure [...]

I suppose this has the exact same limitations as setting using the `CASEDIR` variable to clone a custom Git repo. And then there's the additional problem that if you have multiple worker instances sharing the same test/needles directory this is obviously not going to work. I assume the use-case of this is local testing with one worker.

okurz:
Exactly, all this regarding shared tests is not anything new and nothing changed in this regard.

t/14-isotovideo.t
```diff
@@ -62,6 +64,13 @@ subtest 'isotovideo with custom git repo parameters specified' => sub {
     is_in_log('No scripts',        'the repo actually has no test definitions');
 };
 
+subtest 'isotovideo with git refspec specified' => sub {
+    chdir($pool_dir);
+    unlink('vars.json') if -e 'vars.json';
+    isotovideo(opts => "casedir=$data_dir/tests test_git_refspec=deadbeef _exit_after_schedule=1");
+    is_in_log("Checking.*local.*deadbeef", 'refspec in local git repository would be checked out');
```
kalikiana: What about the failure case where it dies?

okurz: Sorry, I don't understand you. Can you please be more specific?

okurz:
> This seems to lower coverage by 4%

I must admit I don't trust these numbers. Maybe you have a pointer what could be causing this? Just to be sure I retriggered tests on travis CI.

okurz:
See? Coverage increased again ;) By the way, in os-autoinst we record and check for not only statement coverage but also branch coverage, condition coverage and sub coverage. Also see https://travis-ci.org/os-autoinst/os-autoinst/builds/659835547 . I would not worry about this further right now. It should be pretty obvious that this PR should not effect something like consoles/sshIucvconn.pm

t/14-isotovideo.t
kalikiana: This only tests if git succeeds, no?

okurz: No, this actually fails as there is no "deadbeef" but it only looks if git was triggered correctly regardless if it fails or not.

okurz:
I'll go with approval by @Martchus , thx

PR #1359, opened 2020-03-06
lockapi: handle 400 which we could get from openQA API

asmorodskyi:
Currently in case mutex has wrong name you will get 400 error . This situation appear unexpected for os-autoinst

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -39,7 +39,7 @@ sub _try_lock {
         sleep 10;
     }
     bmwqemu::mydie "$type '$name': lock owner already finished" if $res == 410;
-    if ($res != 409) {
+    if ($res != 409 and $res != 400) {
         bmwqemu::fctwarn("Unknown return code $res for lock api");
     }
     return 0;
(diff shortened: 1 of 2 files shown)


okurz:
Can you please take a look if you can add a test, e.g. in t/09-lockapi.t . Also, 
the commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

Martchus:
Maybe:

```
Improve feedback on multi-machine API errors

Otherwise only the return code is logged and it is not clear which
parameters were wrong.
```

Beside a unit test you could also give it a regular test.

asmorodskyi:
After realizing  how deep is rabbit hole I think there is no point in this PR. Good intentions made by this PR will be ruined by copy/paste code in all upper level functions ( see response checks spread in all mutex and barrier related functions ). So to really change something in current behavior need to basically change half of lockapi file. I am not ready  for this ... 
Closing PR

Martchus:
I've created a ticket so we don't forget about this: https://progress.opensuse.org/issues/64322

PR #1361, opened 2020-03-09
Add execution time in the result file

Amrysliu:
Relate: https://progress.opensuse.org/issues/57419#note-8

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -58,6 +58,7 @@ sub new {
     $self->{serial_failures}        = [];
     $self->{autoinst_failures}      = [];
     $self->{fatal_failure}          = 0;
+    $self->{execution_time}         = 0;
     return bless $self, $class;
 }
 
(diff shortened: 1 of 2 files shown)


Martchus:
> we should invest more effort in describing the change

Maybe it is also time to finally document that format. (I've [already started](https://github.com/Martchus/openQA-helper/blob/master/result-format.md) with the results array.)

okurz:
Please add some more details about your testing procedure including openQA

Martchus:
Note that this change has unit tests. I just wanted to check whether everything plays well together so I did manual fullstack tests locally. Like I said, I've picked a random LTP test (https://openqa.opensuse.org/tests/1224670) which accesses the result data structure and cloned it locally. Unlike on o3 the job passed locally. I couldn't spot any unusual messages in the logs, too. I've been testing against openQA master before and after her other PR. In both cases there were no regressions which means we don't have to synchronize merging this with deploying the openQA change.

screenshot_20200406_154633

PR #1362, opened 2020-03-09
Add optional args to GENERAL_HW_POWERON_CMD and GENERAL_HW_POWEROFF_CMD

ggardet:
This is required for some power systems, such as those reachable by network where you need to pass IP/hostname.

### DIFF
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -45,10 +45,8 @@ sub get_cmd {
         die "GENERAL_HW_CMD_DIR is not pointing to a directory";
     }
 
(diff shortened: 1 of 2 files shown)


ggardet:
As args are optionals, I am not sure a hash would be that useful, as we need to make some checks. 
But I can try after lunch.

okurz:
s/optionnal/optional/

ggardet:
It should be fine now.

okurz:
Looks ok. I recommend you also add tests for the backend by mocking low-level calls. This way you would ensure further changes would not break expected behaviour of this backend which you seem to rely on more and more :)

PR #1365, opened 2020-03-10
Fix git hash not being able to compute when no git is available

okurz:
This fixes a regression introduced with 87856568 as we try to compute
the git hash of the needle checkout directory also within
t/01-test_needle.t as we moved the functionality to needle::init. As in
before we simply want to ignore any potential problem with the git
executable completely at this point. This also covers the executable not
being existant at all.


okurz:
This should help with the current problems we encountered in OBS package builds.

okurz:
Thanks a lot to @perlpunk for the actual debugging and help.

PR #1366, opened 2020-03-10
[onhold] Add a wrapper for Test::Output and Test::Warnings 

perlpunk:
We need to check for additional warnings manually.

Otherwise the warnings get swallowed by `Test::Output` and we only
see a failure at the end but not the actual warnings, which makes it hard
to debug.

Can be demonstrated by doing

    stderr_like { warn "hidden"; print STDERR "foo\n" } qr{foo};

Followup-To: https://github.com/os-autoinst/os-autoinst/pull/1365


okurz:
As in chat: Would be better to have a helper library instead of repeating similar boiler plate code in all tests.

Martchus:
The annoying thing about that: We need that helper library in os-autoinst *and* in openQA because tests of both projects use `std(err|out)_like`/`combined_like`. But I would simply copy that helper in openQA next time we have problems identifying warnings in one of the tests there and go away from plain `std(err|out)_like`/`combined_like` as needed.

okurz:
So a new module on CPAN Test::WarningsAndOutput? :)

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -0,0 +1,47 @@
+# Copyright (C) 2016-2020 SUSE LLC
```
Martchus: It should be only 2020.

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -0,0 +1,47 @@
...
+        Test::Output::stderr_like(sub { $code->() }, $re, $label);
+    });
+    if (@warnings) {
+        is(scalar @warnings, 0, 'Got no unexpected warnings in stderr_like call')
+          or Test::More::diag "Warnings: " . join '', @warnings;
+    }
+}
+
+sub combined_like {
+}
```
Martchus: Seems like this is still unimplemented. If we currently don't need/use it I would rather remove the empty definition.

perlpunk: I was just experimenting. I don't rely on the tests passing locally.

perlpunk:
> Have you thought about providing a CPAN module for this?

I would like to avoid creating another CPAN module right now that I have to maintain ;-)

I'm still thinking if this is the right way.
I think Test::Warnings should actually be able to provide the list of warnings on request. Looking into the code.

perlpunk:
I made a PR for Test::Warnings: https://github.com/karenetheridge/Test-Warnings/pull/10

okurz:
I would be fine to merge this and anytime later when we have a potentially updated Test::Warnings we can go back and use it but I wouldn't wait for that now.

perlpunk:
I renamed the module to OpenQA::Test::Warnings and added pod.

perlpunk:
Not needed anymore, https://github.com/karenetheridge/Test-Warnings/pull/10 has been merged

PR #1369, opened 2020-03-23
Feature/cmake

okurz:


snd2png/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+add_executable(${PROJECT_NAME} ${PROJECT_NAME}.cpp)
+find_package(OpenCV REQUIRED)
+include(FindPkgConfig)
+pkg_check_modules(FFTW3 fftw3)
+pkg_check_modules(SNDFILE sndfile)
+target_link_libraries(${PROJECT_NAME} ${FFTW3_LIBRARIES} ${OpenCV_LIBS} ${SNDFILE_LIBRARIES})
+
+enable_testing()
+add_test(NAME test_snd2png
+    COMMAND sh -c "./snd2png ${CMAKE_CURRENT_SOURCE_DIR}/aplay-captured.wav test.png && md5sum test.png > test.png.md5 && diff -u test.png.md5 ${CMAKE_CURRENT_SOURCE_DIR}/test.png.md5.original")
```
coolo: you can surely format this more nicely than that

ppmclibs/tinycv.xs
```diff
@@ -1,227 +0,0 @@
-#ifndef _GNU_SOURCE
```
coolo: where did this file end?

Martchus:
@coolo This is still *very* work in progress. E.g. the following points are missing:

* We haven't even tested whether the generated tinycv library works (maybe some flags are missing).
* It isn't clear as well whether `tinycv.xs` is a generated file or not.
* Cflags and includedirs are currently not taken from the pkg-config files of sndfile and fftw3
* Formatting should be improved; we could at least run https://github.com/cheshirekow/cmake_format one time over it before committing the changes (a package for that formatter is on my OBS home project).

coolo:
> * It isn't clear as well whether `tinycv.xs` is a generated file or not.
I can only recommend to study https://perldoc.perl.org/perlxs.html

ppmclibs/tinycv.xs
okurz: restored and did a different approach by still calling all the commands for ppmclibs as in before, just from cmake.

snd2png/CMakeLists.txt
okurz: done, extracted a test script which is called instead.

Martchus:
Seems like you have given up on building ppmclibs directly via CMake for now. I guess it is ok. At some point we or I can look into it again. This ugly Makefile generation within another build system (be it CMake or autotools) is actually the worst thing in the os-autoinst build script.

I suppose from the point of CMake the XS file can be just yet another C file it needs to compile. Likely one needs to set the `LANGUAGE` source file property manually due to the awkward extension (for a C file).

okurz:
yes, I still have a local branch "feature/cmake2" :) My original approach was anyway to go step by step. Based on our experiments we could even support both build systems in place in parallel.

Martchus:
I would actually prefer to keep both build systems in place for a while. I suppose that's the usual approach when migrating to a different build system.

Martchus:
As I recently had to debug more issues involving the C++ and XS parts so I'd like to continue with that idea a some point. The generation of the XS stuff is particularly bad. We should not nest build systems by invoking `perl Makefile.PL`. It would be easy to let CMake take care of everything. To answer my own previous question: The XS file is not generated the tool `xsubpp` is used to generate a C file from it. This file is compiled like the other "ppmclibs" units.

PR #1370, opened 2020-03-26
Check exit code when running qemu-img

perlpunk:
If the command failed, there will be no JSON to parse

Issue: https://progress.opensuse.org/issues/64854

PR #1372, opened 2020-03-26
Avoid misleading error message from qemu-img

perlpunk:
Issue: https://progress.opensuse.org/issues/64854

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -139,7 +139,13 @@ sub configure_controllers {
 sub get_img_json_field {
     my ($self, $path, $field) = @_;
     my $json = simple_run($self->qemu_img_bin, 'info', '--output=json', $path);
-    my $map  = decode_json($json);
+    # We can't check the exit code of qemu-img, because it sometimes returns
+    # 1 even for a successful command on ppc
+    # Did we get JSON or an error message?
+    my $map = eval { decode_json($json) };
+    if (my $err = $@) {
+        die "qemu-img command failed. Output: $json";
+    }
     die "No $field field in: " . Dumper($map) unless defined $map->{$field};
     return $map->{$field};
 }
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -5,7 +5,7 @@ use strict;
 use warnings;
 
 use Test::More;
-use Test::Warnings;
+use Test::Warnings 'warnings';
 use Mojo::File qw(tempfile path);
 use Carp 'cluck';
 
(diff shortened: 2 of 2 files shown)


okurz:
Let's try again :)

OpenQA/Qemu/Proc.pm
```diff
@@ -139,7 +139,13 @@ sub configure_controllers {
 sub get_img_json_field {
     my ($self, $path, $field) = @_;
     my $json = simple_run($self->qemu_img_bin, 'info', '--output=json', $path);
-    my $map  = decode_json($json);
+    # We can't check the exit code of qemu-img, because it sometimes returns
+    # 1 even for a successful command on ppc
+    # Did we get JSON or an error message?
+    my $map = eval { decode_json($json) };
+    if (my $err = $@) {
```
Martchus: The variable `$err` is unused.

OpenQA/Qemu/Proc.pm
```diff
@@ -139,7 +139,13 @@ sub configure_controllers {
 sub get_img_json_field {
     my ($self, $path, $field) = @_;
     my $json = simple_run($self->qemu_img_bin, 'info', '--output=json', $path);
-    my $map  = decode_json($json);
+    # We can't check the exit code of qemu-img, because it sometimes returns
+    # 1 even for a successful command on ppc
+    # Did we get JSON or an error message?
```
Martchus: I wouldn't ask questions in comments. What about:

```
# determine whether we've got JSON or an error message
# note: We can't check the exit code of qemu-img because it sometimes returns 1 even
#       on success under PowerPC.
```

okurz:
Somehow this PR managed to not address the original problem at all and I completely overlooked this. See https://github.com/os-autoinst/os-autoinst/pull/1376 for another approach by me.

@Martchus This also addresses both your concerns.

perlpunk:
> Somehow this PR managed to not address the original problem at all

Then the original problem was not explained enough in the issue ;-)

okurz:
sorry if https://progress.opensuse.org/issues/64854 wasn't clear enough. The idea was simply that there is no error about malformed JSON. This is what the acceptance criterion describes and actually you nailed it in https://github.com/os-autoinst/os-autoinst/pull/1370 and I have not realized you "brought back" that error line in this PR including a test covering this. I hope you like https://github.com/os-autoinst/os-autoinst/pull/1376

perlpunk:
I didn't bring back the error message. It will not be in the log, it's hidden by `eval`, but it's still counted as a warning in the test due to the `warnings` handler.

okurz:
And yet it is still visible in openQA tests, e.g. see https://openqa.opensuse.org/tests/1215921 as written in https://progress.opensuse.org/issues/64854#note-10

PR #1374, opened 2020-03-27
Create and remove tempfiles in every test

perlpunk:
and get rid of some gitignores.

Now every test can be run without relying on a previous test creating
a certain directory.

Issue: https://progress.opensuse.org/issues/64926

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -6,13 +6,18 @@ use warnings;
 
 use Test::More;
 use Test::Warnings 'warnings';
-use Mojo::File qw(tempfile path);
+use Mojo::File qw(tempfile tempdir path);
 use Carp 'cluck';
+use FindBin '$Bin';
 
 use OpenQA::Qemu::BlockDevConf;
 use OpenQA::Qemu::Proc;
 
 use constant TMPPATH => '/tmp/18-qemu.t/';
+
+my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
+chdir $dir;
+mkdir "$dir/testresults";
 
 $SIG{__DIE__} = sub { cluck(shift); };
 
@@ -90,7 +95,7 @@ my $proc;
     HDDMODEL  => 'virtio-blk',
     CDMODEL   => 'virtio-cd',
     HDDSIZEGB => 69,
-    HDD_1     => 'data/Core-7.2.iso',
+    HDD_1     => "$Bin/data/Core-7.2.iso",
     UEFI      => 1);
 $proc = OpenQA::Qemu::Proc->new()
   ->_static_params(['-foo'])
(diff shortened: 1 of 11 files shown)


perlpunk:
We could also get completely rid of the `t/pool` directory in git. Then also the cleanup will be easier, just `rmtree "$Bin/pool"`

t/18-qemu.t
```diff
@@ -6,14 +6,19 @@ use warnings;
 
 use Test::More;
 use Test::Warnings 'warnings';
-use Mojo::File qw(tempfile path);
+use Mojo::File qw(tempfile tempdir path);
```
Martchus: I can't see that `tempdir` is actually used. I'm also wondering why we can't make more use of it. It had the advantage that the source tree isn't polluted at all and there would be no potential problems when running tests concurrently. It looks like you've tested using `tempdir` but it didn't work for some reason.

perlpunk: Oh right. Yeah, that is a leftover. `tempdir` requires the dir to end with `XXX`, so it's not possible to use it with a specific directory.

Martchus: *That* is the reason? What prevents you from creating a directory `testresults` within the `tempdir`?

perlpunk: Simply because I don't know how to influence the location of the directory. If you look into `t/18-qemu.t` you will only see that I create that directory and at the end remove it. But the test code actually fills it.

Maybe it will just work if I do a `chdir` to the tempdir. I have to try that out.

perlpunk: No, `chdir` doesn't wor. Then it cannot find other data:
`Could not open 'data/Core-7.2.iso'`

perlpunk: It works if I specify `"$Bin/data/...` everywhere

Martchus: We can of course just go for what works for now because out of my head I also don't know how to influence that directory. Maybe it is possible via some environment variable or `$bmwqemu::vars`.

perlpunk: I'll try if I also can get it working with the other tests.
Being able to run tests in parallel would be cool.

perlpunk: @Martchus  I changed everything to use `tempdir` now

Martchus: Nice, so `chdir` works after all but causes some additional effort because other parts were relying on the previous working directory.

perlpunk:
squashed commits

perlpunk:
> Can we put that in END and put to the initial chdir?

Yeah, I can put it in an END block.
I don't understand the second part "put to the initial chdir".

okurz:
I mean to move the END block up to the corresponding code section where you call `chdir` but I wouldn't consider it mandatory :)

perlpunk:
@okurz the `chdir` can't be put in an END block because at that point File::Temp already tried to remove the tempdir, and we're still in that directory.

okurz:
I understand that this might help with running tests in parallel as well. Can you say if this as a non-negligible performance impact, e.g. tests running longer to setup the necessary data?

perlpunk:
It's only creating and removing directories, so I don't expect any visible performance impact.

PR #1375, opened 2020-03-27
Remove ConsistentQuoteLikeWords perlcritic policy

perlpunk:
Having

    use Foo 'bar';

and later adding a function requires changing the quoting, and the other way
round for removing a function.

For a proof of concept I changed it in `01-test_needle.t`


kraih:
Mojolicious will probably do the same upstream.

okurz:
if there is this specific rule "Perl::Critic::Policy::ConsistentQuoteLikeWords" and we added it I assume it has been added for a reason and there is a reason behind the policy. Have you found out what the reason could be?

perlpunk:
@okurz seems it was introduced by @coolo in 5b83714d54eb38e2458874b2523d50ddc8363704

okurz:
ok but why does Perl::Critic::Policy::ConsistentQuoteLikeWords exist? I guess you understand the perl ecosystem better than me :)

perlpunk:
This perlcritic rule is the one which makes me have to amend my PR most often.
And as far as I can see, this rule exists only in os-autoinst. Searching for "ConsistentQuoteLikeWords" gives me only os-autoinst results.

okurz:
what's with https://github.com/os-autoinst/openQA/blob/master/.perlcriticrc#L11 ?

perlpunk:
> what's with https://github.com/os-autoinst/openQA/blob/master/.perlcriticrc#L11 ?

It seems the rules must explicitly be included in the call, so this is ignored.
I guess *if* we get rid of that rule, we should also get rid of this line and the module itself.
Will do that tomorrow, just shutdown my laptop.

okurz:
@coolo your day or nay please?

kalikiana:
> oh, I understand now. Because of [#654 (comment)](https://github.com/os-autoinst/os-autoinst/pull/654#discussion_r90753569) … that is … because of me :D
> 
> So in light of this, I am sorry, I am _against_ this change to keep the consistency of single word imports with single ticks.

If the goal is consistency, can't we require the reverse? That is, never use things. And since the only argument here seems to be consistency, why not go for the option that would reduce time spent rewriting imports.

perlpunk:
I also removed the line in .perlcriticrc and ./tools/lib/perlcritic/Perl/Critic/Policy/ConsistentQuoteLikeWords.pm

okurz:
I am not sure now if style changes in os-autoinst also influence e.g. os-autoinst-distri-opensuse. In this case we should definitely only change style after involving the according contributors.

perlpunk:
> I am not sure now if style changes in os-autoinst also influence e.g. os-autoinst-distri-opensuse. In this case we should definitely only change style after involving the according contributors.

Yes, this is used here:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/Makefile#L113

It seems that if we remove that module here and/or in openQA, that would not work anymore.
How is os-autoinst-distri-opensuse using os-autoinst and/or openQA, that it can use a module that is used only for development?

foursixnine:
> I am not sure now if style changes in os-autoinst also influence e.g. os-autoinst-distri-opensuse. In this case we should definitely only change style after involving the according contributors.

most likely we'd need to do a search and replace, but shouldn't hurt too much

perlpunk:
> most likely we'd need to do a search and replace, but shouldn't hurt too much

Removing the policy will actually not require any changes.
We might think about introducing a new policy to enforce `qw()`, though.

PR #1377, opened 2020-03-31
Delete obsolete FIXME comments

okurz:
FIXME should not stay around in code for all infinity :)

### DIFF
--- a/crop.py
+++ b/crop.py
@@ -43,7 +43,7 @@ if filename.endswith('.png'):
     png = filename
     filename = filename[0:len(filename) - len(".png")] + '.json'
     needle = json.loads("""{
-        "tags": [ "FIXME" ],
+        "tags": [ "none" ],
         "area": [ { "height": 100, "width": 100,
         "xpos": 0, "ypos": 0, "type": "match" } ]
     }""")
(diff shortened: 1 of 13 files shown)


crop.py
```diff
@@ -43,7 +43,7 @@
     png = filename
     filename = filename[0:len(filename) - len(".png")] + '.json'
     needle = json.loads("""{
-        "tags": [ "FIXME" ],
+        "tags": [ "none" ],
```
Martchus: Considering the commit title "Delete obsolete FIXME *comments*" I would refrain from doing that change here.

okurz: Yes, good point. I will extract these changes into a separate commit stating that we want to exclude the special word from template strings

okurz: done with https://github.com/os-autoinst/os-autoinst/pull/1377/commits/1daacbbcba7b275b018cc4390e6ad858007f4642

okurz:
We lost Travis CI feedback but tests were actually run and passed: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/669873223

okurz:
Oh, there it is :)

PR #1380, opened 2020-04-02
[WIP]Adding a slow type parameter to assert_script_run

DrMullings:
We have the issue that commands are typed wrongly and want to avoid
tests failing because of this. The easiest workaround is to take some
time typing

This still needs documentation

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -107,7 +107,7 @@ sub become_root {
 
 =head2 script_run
 
-  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet])
+  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet] [,max_interval => $max_interval])
 
 Deprecated mode
 
@@ -116,6 +116,7 @@ Deprecated mode
 Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
 hashed command to serial line and wait for it in order to detect execution is finished.
 To avoid waiting, use I<$timeout> 0.
+In case you have mistyping problems lower I<$max_interval> which sets I<max_interval> in C<type_string>
 
 Use C<output> to add a description or a comment of the $cmd.
 
@@ -130,15 +131,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout      => $bmwqemu::default_timeout,
+            output       => '',
+            quiet        => undef,
+            max_interval => undef
         }, ['timeout'], @_);
 
     if (testapi::is_serial_terminal) {
         testapi::wait_serial($self->{serial_term_prompt}, no_regex => 1, quiet => $args{quiet});
     }
-    testapi::type_string "$cmd";
+    testapi::type_string("$cmd", max_interval => $args{max_interval});
     if ($args{timeout} > 0) {
         my $str    = testapi::hashed_string("SR" . $cmd . $args{timeout});
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -249,7 +249,16 @@ subtest 'script_run' => sub {
         'using two named arguments; fail message does not apply on timeout'
     );
     $fake_exit = 0;
-    $cmds      = [];
+
+    $cmds = [];    #resetting cmds for easier handling
+
+    # running assert_script_run with a non default max_interval, then checking if it passed
+    #  correctly to the underlying functions
+    is(assert_script_run('true', timeout => 2, max_interval => 1000), '1', 'assert_script_run timed out as expected');
+    like($cmds->[0]->{max_interval}, qr/1000/);
+
+    $cmds = [];
+
     is(script_run('true'), '0', 'script_run with no check of success, returns exit code');
     like($cmds->[1]->{text}, qr/; echo /);
     $cmds = [];
--- a/testapi.pm
+++ b/testapi.pm
@@ -943,7 +943,7 @@ sub _handle_script_run_ret {
 
 =head2 assert_script_run
 
-  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
+  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet] [,max_interval => $max_interval ]);
 
 Deprecated mode
 
@@ -960,6 +960,8 @@ For this to work correctly, it must return 0 if and only if C<$command> complete
 successfully. It must NOT return 0 if C<$command> times out. The default implementation
 should work on *nix operating systems with a configured serial device.>
 
+C<$max_interval> is mapped to C<max_interval> of C<type_string>
+
 =cut
 
 sub assert_script_run {
@@ -971,11 +973,12 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => undef
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, imeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
     _handle_script_run_ret($ret, $cmd, %args);
     return;
 }
@@ -1011,9 +1014,10 @@ sub script_run {
     my $cmd  = shift;
     my %args = compat_args(
         {
-            timeout => undef,
-            output  => '',
-            quiet   => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            timeout      => undef,
+            output       => '',
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => undef
         }, ['timeout'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);


DrMullings:
@SergioAtSUSE @foursixnine

distribution.pm
```diff
@@ -132,13 +132,14 @@ sub script_run {
         {
             timeout => $bmwqemu::default_timeout,
             output  => '',
-            quiet   => undef
+            quiet   => undef,
+            type_speed => testapi::get_var('VNC_TYPING_LIMIT')
```
foursixnine, line 136: Update the documentation accordingly

testapi.pm
```diff
@@ -971,11 +971,12 @@ sub assert_script_run {
...
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            type_speed => testapi::get_var('VNC_TYPING_LIMIT')
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, type_speed => $args{type_speed});
```
SergioAtSUSE, line 979: I think, that we should use the same name to be consistent.

```suggestion
    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
```

testapi.pm
```diff
@@ -971,11 +971,12 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            type_speed => testapi::get_var('VNC_TYPING_LIMIT')
```
SergioAtSUSE, line 975: This needs a default value:

```suggestion
            type_speed => testapi::get_var('VNC_TYPING_LIMIT', undef)
```

DrMullings: Should this not be set always?

testapi.pm
DrMullings: I would prefer a speaking name for the user, max_interval could be anything but type_speed is clear in what it does

distribution.pm
DrMullings: As soon as it is complete, when I now update the doc and need to implement changes this gets messy

testapi.pm
SergioAtSUSE: The default is set here: https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1408

distribution.pm
Martchus, line 136: I would use `bmwqemu` directly similar to the line above, e.g. `$bmwqemu::vars{VNC_TYPING_LIMIT}`.

DrMullings: Adapted it accordingly

DrMullings:
Code tidied up, but I don't understand the new error message

distribution.pm
```diff
@@ -130,15 +130,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout    => $bmwqemu::default_timeout,
+            output     => '',
+            quiet      => undef,
+            type_speed => $bmwqemu::vars{VNC_TYPING_LIMIT};
```
Martchus, line 136: The `;` should be `,`.

DrMullings: Is it even necessary on the last entry?

DrMullings: Apparently not :D

DrMullings:
Updated and tests are fine

distribution.pm
```diff
@@ -130,15 +130,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout    => $bmwqemu::default_timeout,
+            output     => '',
+            quiet      => undef,
+            type_speed => $bmwqemu::vars{VNC_TYPING_LIMIT}
```
SergioAtSUSE, line 136: - The default for **max_interval** is 250: https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1408
- The default for **VNC_TYPING_LIMIT_DEFAULT** is 30: https://github.com/os-autoinst/os-autoinst/blob/master/consoles/vnc_base.pm#L33
- [Doc](https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc) is confusing. Default of **VNC_TYPING_LIMIT** is 50?

In any case, this looks wrong. It is going to cause a slower typing than before each time `script_run` is used.

DrMullings: Should I hardcode it then to accept e.g. 'normal', 'slow' and 'fast'?

DrMullings:
> I guess the pod can be updated now for distribution and testapi

As long as the Typing Limit vs max_interval problem is not solved I would wait for the solution

distribution.pm
SergioAtSUSE:
```perl
type_speed => undef
```

Keep in mind that this change is for distribution, not only VNC based installations.
The function type_string for VNC is on: https://github.com/os-autoinst/os-autoinst/blob/master/consoles/vnc_base.pm#L98
And it already make use of
```perl
my $seconds_per_keypress = 1 / (get_var('VNC_TYPING_LIMIT', VNC_TYPING_LIMIT_DEFAULT) || 1);
```

DrMullings:
Added some documentation

distribution.pm
```diff
@@ -116,10 +116,11 @@ Deprecated mode
 Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
 hashed command to serial line and wait for it in order to detect execution is finished.
 To avoid waiting, use I<$timeout> 0.
+In case you have mistyping problems lower I<$type_speed> from its default value of 250
```
foursixnine, line 119:
```suggestion
 C<$type_speed> is 250 by default, lower this in case you have typing issues 
```

DrMullings: I tried to be consistent with the other parameters

DrMullings:
We still need a good way to verify these changes

DrMullings:
Adapted the naming and the documentation according to @okurz suggestions
Not sure about the type_password approach though

distribution.pm
```diff
@@ -107,7 +107,7 @@ sub become_root {
 
 =head2 script_run
 
-  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet])
+  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet] [, type_speed => $type_speed])
```
okurz, line 110: please use `max_interval` here as well

distribution.pm
```diff
@@ -116,10 +116,11 @@ Deprecated mode
 Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
 hashed command to serial line and wait for it in order to detect execution is finished.
 To avoid waiting, use I<$timeout> 0.
+In case you have mistyping problems lower I<$typing_speed> which sets I<max_interval> in C<type_string>
```
okurz, line 119: wrap lines

distribution.pm
```diff
@@ -116,10 +116,11 @@ Deprecated mode
...
 
 Use C<output> to add a description or a comment of the $cmd.
 
-Use C<quiet> to avoid recording serial_results.
+    Use C<quiet> to avoid recording serial_results.
```
okurz, line 123: unrelated change

distribution.pm
```diff
@@ -130,15 +131,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout      => $bmwqemu::default_timeout,
+            output       => '',
+            quiet        => undef,
+            typing_speed => undef
```
okurz, line 137: s/type_speed/max_interval/

testapi.pm
```diff
@@ -943,7 +943,7 @@ sub _handle_script_run_ret {
 
 =head2 assert_script_run
 
-  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
+  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet] [,type_speed => $type_speed ]);
```
okurz, line 946: s/type_speed/max_interval/

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            type_speed   => undef
```
okurz, line 977: s/type_speed/max_interval/

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
...
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, typing_speed => $args{type_speed});
```
okurz, line 981: s/typing_speed/max_interval/

DrMullings:
I am still strongly against a name that doesn't tell the function.
Wondering why tidy messes up again

okurz:
Yes, you have stated that point but as also stated in before: our test API should not break compatibility lightly hence my recommendation to use max_interval everywhere. What you can do is introduce a deprecation message for about a year and support both parameter names for the time being

DrMullings:
Changed the name despite my explicit wish to get this forward.
Currently working on a unit test for this change

okurz:
please phrase the git commit message in imperative mood

DrMullings:
If you have anything else please write it all in one message, as it is *really* annoying to get "change this, change that" feedback only bit by bit

distribution.pm
SergioAtSUSE: As mentioned by Oli, this should be split in a separate commit.

okurz:
Yes, I get that. I can also give it back that reading trivial changes once every two weeks can also be annoying especially as you have not attended other open comments since a month.

foursixnine:
> Yes, I get that. I can also give it back that reading trivial changes once every two weeks can also be annoying especially as you have not attended other open comments since a month.

Are we going to start playing `Tag it?`

DrMullings:
Giving in on a topic where I still have a completely different opinion feels not trivial for me.
That "short and demanding" style of communication in combination with forcing me to name parameters a way that I hate made me ignore that ticket for a month and coming back and then getting feedback to the commit message but not to the changes themselves and another sassy comment instead of a "please also take a look at the open topics" is really impacting my motivation.
Might be that I am over sensitive here and you're just being efficient, but it is in our both interest to get this finished asap and both of us being annoyed is not helping here.

t/03-testapi.t
```diff
@@ -231,6 +235,11 @@ subtest 'script_run' => sub {
 
     is(assert_script_run('true'), undef, 'nothing happens on success');
     $fake_exit = 1;
+
+    exception {assert_script_run ('false', timeout => 2, max_interval => 1000), 'false', "It failed"} or diag "The test failed";
+    diag explain $cmds;
+    BAIL_OUT("Testing is done");
+
```
DrMullings, line 242: t/03-testapi.t .. 
ok 1 - isotovideo version set (variable is considered part of test API)
# Subtest: type_string
[2020-05-19T15:07:03.000 CEST] [debug] <<< testapi::type_string(string="hallo", max_interval=250, wait_screen_changes=0, wait_still_screen=0, timeout=30, similarity_level=47)
[2020-05-19T15:07:03.000 CEST] [debug] backend_type_string
    1..0
    # No tests run!
not ok 2 - No tests run for subtest "type_string"

#   Failed test 'No tests run for subtest "type_string"'
#   at t/03-testapi.t line 166.
Not a HASH reference at /home/jora/Git/os-autoinst-distri-opensuse/os-autoinst/autotest.pm line 331.
ok 3 - no (unexpected) warnings (via END block)
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 3.
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/3 subtests 

Test Summary Report
-------------------
t/03-testapi.t (Wstat: 65280 Tests: 3 Failed: 1)
  Failed test:  2
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
Files=1, Tests=3,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.26 cusr  0.01 csys =  0.28 CPU)
Result: FAIL

shell returned 1

DrMullings: @Martchus could you please give a hint for the unit testing?

okurz: The syntax in L239 is broken. Take a close look at the arguments of `assert_script_run`. As first argument you pass `(…)` and then comma-separated after that `false` and "It failed". I recommend you take over the same style as the other calls in the same subtest, e.g. use

```perl
like(exception {assert_script_run 'false', timeout => 2, max_interval => 1000; }, qr/false/, 'foo');
```

Make sure you also test that the `max_interval` is properly passed through to the underlying command structure, e.g. use a similar format as further down in this test with `like($cmds->[1]->{text}, qr/; echo /)` but check for `max_interval` instead

t/03-testapi.t
```diff
@@ -85,6 +85,10 @@ sub fake_read_json {
     elsif ($cmd eq 'backend_mouse_set') {
         return {ret => {x => 100, y => 100}};
     }
+    elsif ($cmd eq 'backend_type_string') {
+        diag explain $cmd;
+        return  $cmds;
+    }
```
okurz, line 91: I assume you only added this for debugging? Should be removed again when you are done

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
...
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => undef
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
```
okurz, line 981: Can you please try if it also works to just pass on all additional parameters? Then you would not even need to define line L977 and also do not need to mention `max_interval` explicitly here. Similar as we do in the next line `_handle_script_run_ret($ret, $cmd, %args);`

distribution.pm
```diff
@@ -130,15 +131,16 @@ sub script_run {
...
+            output       => '',
+            quiet        => undef,
+            max_interval => undef
         }, ['timeout'], @_);
 
     if (testapi::is_serial_terminal) {
         testapi::wait_serial($self->{serial_term_prompt}, no_regex => 1, quiet => $args{quiet});
     }
-    testapi::type_string "$cmd";
+    testapi::type_string("$cmd", max_interval => $args{max_interval});
```
okurz, line 143: please also see my comment further down about passing `%args` unaltered. Maybe just `testapi::type_string $cmd, %args` is enough

distribution.pm
okurz: please attend this comment in your changes as well

distribution.pm
foursixnine: indeed

t/03-testapi.t
foursixnine: Actually we were wondering if we had to mock `backend_type_string` your comment further down, clarifies this!

t/03-testapi.t
foursixnine: *facepalms*, @DrMullings we need more coffee :P. 

@okurz with this, the changes should be properly tested right? as what's being passed to distribution is being tested through the test api itself (Or do I need more coffe here?)

okurz: Yes, I would say so. So if you add one call of `assert_script_run` with the right parameter and just check that the parameter is forwarded I consider this sufficient testing.

testapi.pm
DrMullings: I am pretty sure this would work, but do we really want to pass everything unattended?
Seems a bit risky to me

okurz: I don't see a problem with that as we only use selected parameters.

distribution.pm
DrMullings: I think we would mix up the parameters for script_run with the ones we want to parse further down? e.g. timeout should not be passed here if we handle it in the line afterwards

t/03-testapi.t
DrMullings: I hope I got the syntax correctly now

DrMullings:
Updated according to the suggestions, somehow the test file seems to be screwed up now?

foursixnine:
> Updated according to the suggestions, somehow the test file seems to be screwed up now?

Tidy :) https://travis-ci.org/github/os-autoinst/os-autoinst/builds/691682222#L690

DrMullings:
> > Updated according to the suggestions, somehow the test file seems to be screwed up now?
> 
> Tidy :) https://travis-ci.org/github/os-autoinst/os-autoinst/builds/691682222#L690

Somehow the shebang was copied to the very end of the file, I have no clue how that happened
Also stuff was touched that I never touched, most likely I will need to reset this file and add the two lines again

DrMullings:
Now it should be fine, I hope I caught everything

foursixnine:
> Now it should be fine, I hope I caught everything

Looks like not: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/691702877#L897

DrMullings:
Looks like something in the syntax is still wrong

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => undef
```
foursixnine, line 977: I think this is the line that mangles the input (https://travis-ci.org/github/os-autoinst/os-autoinst/builds/691702877#L852)

distribution.pm
```diff
@@ -130,15 +131,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout      => $bmwqemu::default_timeout,
+            output       => '',
+            quiet        => undef,
+            max_interval => undef
```
foursixnine, line 137: this one would mangle the input aswell

DrMullings: but we need those default values, don't we?

distribution.pm
DrMullings: If I understand the test output correctly we caused exactly that problem, as we pass quiet further down than it should be and therefore the default is wrong

distribution.pm
```diff
@@ -130,15 +131,16 @@ sub script_run {
     my ($self, $cmd) = splice(@_, 0, 2);
     my %args = testapi::compat_args(
         {
-            timeout => $bmwqemu::default_timeout,
-            output  => '',
-            quiet   => undef
+            timeout      => $bmwqemu::default_timeout,
+            output       => '',
+            quiet        => undef,
+            max_interval => 250
```
SergioAtSUSE, line 137: These values are the default if the arguments doesn't contain value.
ie. No argument passed when function was called.
https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L2164

By using 250 it is overriding `type_string` defaults.
https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1364

```suggestion
            max_interval => undef
```

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => 250
```
SergioAtSUSE, line 977:
```suggestion
            max_interval => undef
```

testapi.pm
```diff
@@ -1011,9 +1014,10 @@ sub script_run {
     my $cmd  = shift;
     my %args = compat_args(
         {
-            timeout => undef,
-            output  => '',
-            quiet   => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            timeout      => undef,
+            output       => '',
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => 250
```
SergioAtSUSE, line 1020:
```suggestion
            max_interval => undef
```

DrMullings: I know, that was just to be sure, now I am pretty sure that my test syntax is still not right

testapi.pm
```diff
@@ -971,11 +973,11 @@ sub assert_script_run {
             # not change default timeout.
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
```
SergioAtSUSE, line 976: I don't know what you are trying to do, but in the test you are passing `max_interval => 1000` and here I don't see `max_interval => undef` nor `max_interval => 250`

foursixnine: See line 980.

distribution.pm
foursixnine: @DrMullings actually @okurz is right, passing it should help, however the quet argument can be deleted from the hash before passing it

testapi.pm
SergioAtSUSE: I saw line 980 and still, `%args` will contain no key `max_interval`.

Try yourself:

```bash
user@host:~$ perl missing_parameter.pl 
a: fail! 
a: fail! 
a: fail! 
b: fail! 
b: fail! 
b: do things
```

```perl
#!/usr/bin/perl
# missing_parameter.pl

sub compat_args {
    my ($def_args, $fix_keys) = splice(@_, 0, 2); 
    my %ret;
    for my $key (@{$fix_keys}) {
        $ret{$key} = shift if (scalar(@_) >= 1 && (!defined($_[0]) || !grep(/^$_[0]$/, keys(%{$def_args}))));
    }   
    carp("Odd number of arguments") unless ((@_ % 2) == 0); 
    %ret = (%{$def_args}, %ret, @_);
    map { $ret{$_} //= $def_args->{$_} } keys(%{$def_args});
    return %ret;
}

sub assert_script_run_a {
  my $cmd = shift;
  my %args = compat_args(
    {   
      timeout      => 90, 
      fail_message => '', 
      quiet        => undef
    }, ['timeout', 'fail_message'], @_);

  if ($args{max_interval}) {
    print "a: do things \n";
  } else {
    print "a: fail! \n";
  }
}

sub assert_script_run_b {
  my $cmd = shift;
  my %args = compat_args(
    {   
      timeout      => 90, 
      fail_message => '', 
      quiet        => undef,
      max_interval => undef
    }, ['timeout', 'fail_message'], @_);

  if ($args{max_interval}) {
    print "b: do things \n";
  } else {
    print "b: fail! \n";
  }
}

assert_script_run_a('true');
assert_script_run_a('true', max_interval => undef);
assert_script_run_a('true', max_interval => 250);

assert_script_run_b('true');
assert_script_run_b('true', max_interval => undef);
assert_script_run_b('true', max_interval => 250);
```

DrMullings: if I set it to undef it will overwrite the default from type_string, if I set it to 250 we have duplicate code

SergioAtSUSE: The code I pasted prove that `undef` doesn't overwrite.
Please, try to execute the code yourself.

distribution.pm
```diff
@@ -107,7 +107,7 @@ sub become_root {
 
 =head2 script_run
 
-  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet])
+  script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet] [,max_interval => $max_interval])
```
DrMullings, line 110: @SergioAtSUSE might this line interfere and overwrite the default? I cant give all the arguments as some are not meant for type_string

SergioAtSUSE: Well this is just the comment.

I assume you mean line 143? Where you are passing a maybe empty (undef) `max_interval` key?

It doesn't, as the output of the pasted script demonstrate. `if ($args{max_interval})` checks is a value for that key exists, not if the key exists.

Or do you mean something else?

SergioAtSUSE: If you mean that you should only pass the parameters that are expected for type_string, I agree.
I am not sure how much benefit of "spare typing" vs "maintenance" we gain passing a hash instead of only the values we need.
Because removing the keys from `@args` may not be a good idea if some logic later needs to use them.

distribution.pm
SergioAtSUSE: @foursixnine, why? do you have an example that rejects the hypothesis supported by my pasted script?

distribution.pm
DrMullings: I meant passing an undef value explicitly, but if you tested it it should be fine.
Also I agree on not manipulating the hash just to pass the complete hash that contains the same as only the value we want to pass

SergioAtSUSE:
At the moment travis just complains about indentation: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/694615187#L695

distribution.pm
foursixnine: So, parameters in general are somehow being mangled, https://travis-ci.org/github/os-autoinst/os-autoinst/builds/694623878#L866 (but at least seems to be consistent now: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/694623878#L915)

SergioAtSUSE:
Well, it looks that passing a hash with undef values is not handled as expected in perl.

I added two hashes to my script:

```perl
my %args_a = (max_interval => undef); # <---
my %args_b = (max_interval => 250);   # <---
assert_script_run_a('true');
assert_script_run_a('true', max_interval => undef);
assert_script_run_a('true', %args_a); # <---
assert_script_run_a('true', %args_b); # <---
assert_script_run_a('true', max_interval => 250);

assert_script_run_b('true');
assert_script_run_b('true', max_interval => undef);
assert_script_run_a('true', %args_a); # <---
assert_script_run_a('true', %args_b); # <---
assert_script_run_b('true', max_interval => 250);
```

and the output:

```bash
user@host:~$ perl missing_parameter.pl
a: fail! 
a: fail! 
a: fail! 
a: fail! 
a: fail! 
b: fail! 
b: fail! 
a: fail! 
a: fail! 
b: do things
```

So, single parameters is better than hashes if they contain undef values.

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
...
+            max_interval => undef
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, %args);
```
SergioAtSUSE, line 981:
```suggestion
    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
```

DrMullings: I adopted that change

DrMullings:
For me it looks like we only need to fix the test syntax? I think we should go back to the like(exception) approach *if* we want to keep the inverse testing logic forcing a timeout.
We could also simply change to positive testing logic and increse the timeout

testapi.pm
```diff
@@ -971,11 +973,12 @@ sub assert_script_run {
...
             timeout      => 90,
             fail_message => '',
-            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS')
+            quiet        => testapi::get_var('_QUIET_SCRIPT_CALLS'),
+            max_interval => undef
         }, ['timeout', 'fail_message'], @_);
 
     bmwqemu::log_call(cmd => $cmd, %args);
-    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet});
+    my $ret = $distri->script_run($cmd, imeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
```
SergioAtSUSE, line 981:
```suggestion
    my $ret = $distri->script_run($cmd, timeout => $args{timeout}, quiet => $args{quiet}, max_interval => $args{max_interval});
```

DrMullings:
Lets not get this lost just because I lost focus.
All thats needed is a unit test

PR #1383, opened 2020-04-03
Write backend termination message to base_state.json

kalikiana:
See os-autoinst/openQA#2924 for the change in OpenQA

Related: https://progress.opensuse.org/issues/64857

### DIFF
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -9,6 +9,7 @@ use Test::Output 'stdout_is';
 use Test::Warnings;
 use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
 use Mojo::File 'path';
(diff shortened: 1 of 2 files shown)


t/23-baseclass.t
```diff
@@ -193,6 +194,15 @@ subtest 'SSH utilities' => sub {
 
         is($baseclass->check_ssh_serial(23), 0, 'Return 0 if SSH serial isn\'t connected');
     };
+
+    subtest Run => sub {
+        path('base_state.json')->remove;
```
okurz, line 199: Does this when the file does not exist?

kalikiana: Nothing happens if the file doesn't exist.

PR #1389, opened 2020-04-24
Dependencies: Remove Data::Dump, JSON, JSON::XS from spec

perlpunk:
They are not used anymore.

maybe we have to wait until https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10090 is merged


perlpunk:
I see that Data::Dump is used in https://github.com/os-autoinst/os-autoinst-distri-opensuse/ but it's not in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/os-autoinst-distri-opensuse-deps.spec
So maybe it should be added there.

Martchus:
> So maybe it should be added there.

Yes, it likely should be. But the list there seems to be auto-generated so let's ask the test distribution maintainers how they'd like to deal with it.

perlpunk:
Yeah, I'm already looking into that list. cpanfile is maintained manually, and the spec is generated from that. I'm currently preparing a PR to update the deps.

okurz:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10096 also merged

PR #1390, opened 2020-04-26
Speedup "integration" full stack tests

okurz:
The os-autoinst full stack test as well as openQA uses the full stack
test with the variable INTEGRATION_TESTS set to 1 to skip some
long-running test steps that are testing os-autoinst but are not
relevant to ensure the full stack of openQA and os-autoinst working
together. Over time the test module "boot" has grown to conduct much
more than what is relevant for booting the tested system. By moving out
all but the boot-relevant test code to dedicated test modules we can
save a lot of time in openQA full stack tests.

In my local environment this reduces the runtime of t/99-full-stack.t
from 4m57s to 3m27s without loosing any coverage as every test module is
still executed once.

The current openQA full stack test has multiple more runs of the full
stack test which can be significantly improved with this commit.

### DIFF
--- a/t/data/tests/main.pm
+++ b/t/data/tests/main.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2016-2017 SUSE LLC
+# Copyright (C) 2016-2020 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


okurz:
This decreases the runtime for just the *first* internal full stack runs of openQA from 2m20 to 48s. The complete runtime decreases from about 720s to 250s

PR #1395, opened 2020-05-04
Add optional fatal handling for qemu dbus calls for easier error analysis

okurz:
MM tests fail to access other machines in the network if tap device is
not present, there is a diag message reported but the test does not fail
at this point.

For example a test can show "Failed to connect to 10.0.2.2 port 20243:
No route to host" and fails. What is hidden in autoinst-log.txt:

```
[2020-05-04T08:30:16.586 CEST] [debug] Failed to run dbus command 'set_vlan' with arguments 'tap23 13' : 'tap23' is not connected to bridge 'br1'
```

Multi-machine tests relying on a tap device should fail hard if the tap
device is not present or can not be configured. With the added feature
switch QEMU_FATAL_DBUS_CALL this can be enabled. The original behaviour
had been introduced in 82fe22e2 . The reason for continuing despite the
error is unknown to me hence for now I am just adding a feature switch
for changed behaviour which we eventually might want to use as the only
new supported behaviour.

Related progress issue: https://progress.opensuse.org/issues/66376

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -130,21 +130,24 @@ sub stop_qemu {
     $self->_stop_children_processes;
 }
 
+sub _dbus_do_call {
+    my ($self, $fn, @args) = @_;
+    $self->{dbus}         ||= Net::DBus->system;
+    $self->{dbus_service} ||= $self->{dbus}->get_service("org.opensuse.os_autoinst.switch");
+    $self->{dbus_object}  ||= $self->{dbus_service}->get_object("/switch", "org.opensuse.os_autoinst.switch");
+    $self->{dbus_object}->$fn(@args);
+}
+
 sub _dbus_call {
     my $self = shift;
     my $fn   = shift;
     my @args = @_;
-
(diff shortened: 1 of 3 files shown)


okurz:
@schlad as discussed in https://progress.opensuse.org/issues/66376 this is my approach to make failed dbus calls fatal. I hope after deploying this change we can experiment with this, enable it in experimental runs which we schedule against openQA workers that do *not* have tap devices configured correctly and that should fail earlier and in a more obvious manner than currently.

schlad:
LGTM;
looks like a nice improvement on the backend side.
I agree, this seems like we should fail as early as possible.

But of course I still would wonder how could we prevent this: "shows that openqaworker-arm-2 and openqaworker-arm-3 are currently running incorrectly." After all this seems to be the root cause here. (thx for your analysis. Helps me understanding better where too look for problems).

okurz:
The misconfiguration will also be handled as part of https://progress.opensuse.org/issues/66376 but probably not with code changes on the side of os-autoinst. It is either something in the systemd service definitions of openQA or the controlling salt recipes.

PR #1400, opened 2020-05-12
Define travis CI command in simple helper script

okurz:
Inspired by https://github.com/os-autoinst/os-autoinst/pull/1398

### DIFF
--- a/tools/docker_run_ci
+++ b/tools/docker_run_ci
@@ -1,0 +1,5 @@
+#!/bin/bash -e
+CI_ENV=$(bash <(curl -s https://codecov.io/env))
+# shellcheck disable=SC2086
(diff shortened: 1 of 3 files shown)


tools/docker_run_ci
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+CI_ENV=`bash <(curl -s https://codecov.io/env)`
```
Martchus, line 2: `tools/docker_run_ci: 1: tools/docker_run_ci: Syntax error: "(" unexpected`

okurz:
The test changes should not have been included. I will correct this

perlpunk:
The test changes were still in the PR when you merged?

Besides that, I'm seeing build failures in OBS:
https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Factory/x86_64
```
[  213s] error: Installed (but unpackaged) file(s) found:
[  213s]    /usr/lib/os-autoinst/tools/docker_run_ci
```

perlpunk:
Created https://github.com/os-autoinst/os-autoinst/pull/1402

okurz:
The test changes were still in the PR but I moved them to an explicit commit. The build failures we couldn't see easily in before because we do not have these shiny automatic OBS tests for os-autoinst :( I never learned how they are set up and coolo is a bit silent about this topic.

PR #1401, opened 2020-05-13
Move Dockerfile for CI testing from openQA repo

perlpunk:

Also install patched qemu, see comment

### DIFF
--- a/docker/travis_test/Dockerfile
+++ b/docker/travis_test/Dockerfile
@@ -1,0 +1,157 @@
+#!BuildTag: openqa_dev
+FROM opensuse/leap:15.1
+
+# Define environment variable
+ENV NAME openQA test environment
+ENV LANG en_US.UTF-8
+
+RUN zypper ar -f -G 'http://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.1/openSUSE_Leap_15.1' devel_openqa
+
+RUN zypper in -y -C \
+       glibc-i18ndata \
+       glibc-locale \
+       automake \
+       curl \
+       fftw3-devel \
+       gcc \
+       gcc-c++ \
+       git \
+       gmp-devel \
+       gzip \
+       libexpat-devel \
+       libsndfile-devel \
+       libssh2-1 \
+       libssh2-devel \
+       libtheora-devel \
+       libtool \
+       libxml2-devel \
+       make \
+       opencv-devel \
+       patch \
+       postgresql-devel \
+       qemu \
+       qemu-tools \
+       qemu-kvm \
+       tar \
+       optipng \
+       python3-base \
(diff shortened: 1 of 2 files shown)


perlpunk:
> Hm, might be safer to go step by step and move over the unchanged file first before changing it. 

I tested the additional commands locally and in github actions though.

> You also need to adjust the use in OBS then, right?

Yes. Writing a comment in the issue right now.

perlpunk:
If we do the change in the next step (in a followup pull request) and it goes wrong, we have the same problems as we would have now.
Because we first need to merge it anyway to build the image.

okurz:
Well, you could test in a OBS repo checking out from your branch instead of openQA master

perlpunk:
> Well, you could test in a OBS repo checking out from your branch instead of openQA master

So you mean I should branch the OBS repo and use my branch URL in there?
And where will the resulting docker image be published?

perlpunk:
I removed the additional command now.
Is it now safe enough?

okurz:
> So you mean I should branch the OBS repo and use my branch URL in there?

yes

> And where will the resulting docker image be published?

All images are published on https://registry.opensuse.org/ . Simple example: https://build.opensuse.org/package/show/home:okurz:container/os-autoinst publishes `registry.opensuse.org/home/okurz/container/containers/tumbleweed:os-autoinst`

perlpunk:
Well, I did branch the package into https://build.opensuse.org/package/show/home:tinita:branches:devel:openQA/openqa_dev, but I don't see it in the registry

perlpunk:
Aha, I needed to remove the `<disable/>` in the meta of the parent repo. Now I can see it in the registry.

perlpunk:
Ok, I tested with my branched image: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/687086091
I added a `docker images` to show which image we actually have.

This way I will also be able to test when removing dependencies.

I'll replace the image now again with the official one.

perlpunk:
> Assuming the file is cleaned up in further commits to include only dependencies needed by os-autoinst.

Yes, I have already tested with a reduced list of deps (docker image shrinked from 1.57GB to 1.28GB), and it's working. But we should do this step by step.

perlpunk:
Are we really sure that the image is only used for os-autoinst testing? I cannot find any reference to it, but maybe I'm not searching all places.
Maybe it would make sense to use a new OBS package `os-autoinst:ci` for this? And we can later disable the other image (and if something/someone is using it we can enable it easily again)?

Martchus:
> Are we really sure that the image is only used for os-autoinst testing?

I'm not sure. I haven't dealt with the OBS/Docker setup and its use within our CIs so far.

PR #1403, opened 2020-05-14
Add back colors to log output if available

okurz:
* Define custom log color selection compatible with reverse video terminals
* Use colors in all log calls if colors are available
* Reduce code duplication in bmwqemu::init_logger
* Simplify logger init in bmwqemu
* t: Prevent errors on local, uncommited files in compile-check

Example output of colors in terminal windows. Black-on-white when Mojo::Log::Colored is not available (top-left), Black-on-white (bottom-left), White-on-black (right):

screenshot_os-autoinst_log_colors_white_and_black_background

Example output of colors in terminal windows with additional colors using https://github.com/os-autoinst/os-autoinst/pull/1403/commits/a65b8389a5755d45387393a889f7611aa7c98890 . Black-on-white when Mojo::Log::Colored is not available (top-left), Black-on-white (bottom-left), White-on-black (right):

screenshot_os-autoinst_log_colors_white_and_black_background_with_colors_for_modstart_and_fctres

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -69,6 +69,11 @@ our @ovmf_locations = (
 
 our %vars;
 
+sub logger {
+    $logger //= Mojo::Log->new(level => 'debug', format => \&log_format_callback);
+    return $logger;
+}
+
 sub load_vars {
     my $fn  = "vars.json";
     my $ret = {};
@@ -114,13 +119,8 @@ our $gocrbin = "/usr/bin/gocr";
 our $scriptdir;
 
 sub init_logger {
-    if ($direct_output) {
-        $logger = Mojo::Log->new(level => 'debug');
-    }
-    else {
-        $logger = Mojo::Log->new(level => 'debug', path => catfile(result_dir, 'autoinst-log.txt'));
-    }
-
+    my $log = Mojo::Log->new(level => 'debug');
+    $logger = $log->path(catfile(result_dir, 'autoinst-log.txt')) unless $direct_output;
     $logger->format(
         sub {
             my ($time, $level, @lines) = @_;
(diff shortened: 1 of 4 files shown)


bmwqemu.pm
```diff
@@ -42,6 +42,9 @@ our @EXPORT_OK = qw(diag);
 use backend::driver;
 require IPC::System::Simple;
 
+my $log_module = "Mojo::Log";
+eval 'use Mojo::Log::Colored; $log_module = "Mojo::Log::Colored"';
```
Martchus, line 46: Maybe some of the Perl people know something better?

bmwqemu.pm
```diff
@@ -70,6 +73,24 @@ our @ovmf_locations = (
 
 our %vars;
 
+sub result_dir { 'testresults' }
+
+sub logger {
```
Martchus, line 78: Now that this function exists, could `our $logger` be changed to `my $logger`?

okurz: Well, `our $logger` so far is only overwritten in tests by giving it another logger instead. If we would replace that with calls to the method then we would first initialize a logger which we then overwrite with a new object. Not that nice by design. Unless we want to allow dependency injection by allowing an optional `$logger` argument here to have something akin to:

```perl
sub logger {
    $logger = shift // $logger // $log_module->new(level => 'debug', format => \&log_format_callback…
```

bmwqemu.pm
kalikiana: Why the `eval`? Can't we just always use it?

okurz: Not when we see it as recommendation instead of requirement. The package is so far not even in openSUSE:Factory pending inclusion in [d:l:p](https://build.opensuse.org/request/show/805622) first. I am merely applying the same approach you used for Test::More::Color in openQA. Would you prefer an alternative that is not using Mojo::Log::Colored at all and just relying on Term::ANSIColor?

Martchus: Change 2. seems trivial enough to avoid introducing a dependency for that.

okurz: ok, done

PR #1404, opened 2020-05-15
Fix sporadic failures in t/17-basetest.t

okurz:
From https://build.opensuse.org/package/live_build_log/devel:openQA:tested/os-autoinst/openSUSE_Factory/x86_64

```
[  167s] [2020-05-14T19:27:59.004 UTC] [debug] ||| finished basetest foo at 2020-05-14 19:27:59 (3 s)
[  167s]
[  167s]     #   Failed test 'the execution time is correct'
[  167s]     #   at 17-basetest.t line 325.
[  167s]     #          got: '3'
[  167s]     #     expected: '2'
[  167s]     # Looks like you failed 1 test of 2.
[  167s]
[  167s] #   Failed test 'execute_time'
[  167s] #   at 17-basetest.t line 326.
[  167s] # Looks like you failed 1 test of 6.
[  167s] 17-basetest.t ..........................
```

A test relying on consistent timing behaviour can hardly be stable. A test
should not rely on exact runtime of internal test modules. The code "sleep 2"
does not ensure that we sleep 2 seconds but can take slightly longer. The test
could be changed to assert "2 or more". A better approach is if the test is
changed to not sleep at all. Also, the test module takes 2s just because of
this code, the rest is way faster.

This commit changes the mocked code to not sleep at all. This fixes the longer
runtime and should be more stable. Theoretically it can still be that the
recorded runtime is more than 0s but yielding to other threads/processes is
highly unlikely during the execution of runtest with no sleep execution in
between.

Related progress issue: https://progress.opensuse.org/issues/66886

### DIFF
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -319,10 +319,11 @@ subtest 'execute_time' => sub {
     my $mock_basetest  = Test::MockModule->new($basetest_class);
     my $test           = basetest->new('foo');
     is($test->{execution_time}, 0, 'the execution time is initiated correctly');
-    $mock_basetest->mock(run => sub { sleep 2; });
-    $mock_basetest->redefine(done => sub { });
+    $mock_basetest->mock(execution_time => 42);
+    $mock_basetest->mock(run            => undef);
+    $mock_basetest->redefine(done => undef);
     $test->runtest;
-    is($test->{execution_time}, 2, 'the execution time is correct');
+    is($test->{execution_time}, 42, 'the execution time is correct');
 };
 
 done_testing;
--- a/basetest.pm
+++ b/basetest.pm
@@ -337,6 +337,8 @@ sub run_post_fail {
     die $msg . "\n";
 }
 
+sub execution_time { time - shift }
+
 sub runtest {
     my ($self) = @_;
     my $starttime = time;
@@ -389,7 +391,7 @@ sub runtest {
     }
 
     $self->done();
-    $self->{execution_time} = time - $starttime;
+    $self->{execution_time} = execution_time($starttime);
     bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), $self->{execution_time}));
     return;
 }


okurz:
@Amrysliu this is changing the code you introduced to fix an instability. The side-effect is that it speeds up the test again. In general we should always aim for a runtime of less than 1s for unit tests.

t/17-basetest.t
```diff
@@ -319,10 +319,10 @@ subtest 'execute_time' => sub {
     my $mock_basetest  = Test::MockModule->new($basetest_class);
     my $test           = basetest->new('foo');
     is($test->{execution_time}, 0, 'the execution time is initiated correctly');
-    $mock_basetest->mock(run => sub { sleep 2; });
-    $mock_basetest->redefine(done => sub { });
+    $mock_basetest->mock(run => undef);
+    $mock_basetest->redefine(done => undef);
     $test->runtest;
-    is($test->{execution_time}, 2, 'the execution time is correct');
+    is($test->{execution_time}, 0, 'the execution time is correct');
```
perlpunk, line 325: What I don't like about this change is, that we now never test if the `execution_time` field is touched after the initialization.
I would prefer to test for `>= 2`

okurz: Sure, that would most likely work but not address that we waste two seconds just to test that two seconds have passed. Do you have another idea?

okurz: ok, fixed with mocking the calculation of execution time. Please check again.

perlpunk: Maybe we should log the time with milliseconds, at least if it is less than 1 second. Then we could `sleep 0.1`.

perlpunk: Just looked into basetest.pm. We are using `Time::HiRes`, but don't import a function. So we could use its `time` function and use `sprintf "%.1f"` if it is smaler than 1 sec.

perlpunk: Ok. Should we remove the `use Time::HiRes` also in this PR? It's not used.

okurz: @perlpunk I was not sure about the Time::HiRes part so I merged after two other proposals to first fix the sporadic issue first. I hope that is ok for you. I will propose a PR for the remove of Time::HiRes after crosschecking other users or files.

okurz: https://github.com/os-autoinst/os-autoinst/pull/1407 for my followup

PR #1405, opened 2020-05-15
Add conditional skipping of IPMI selftests

frankenmichl:
Some Machines have flaky or buggy IPMI implementations. It can happen,
that the selftest requests to the BMC fails. In some cases we want to
run the tests anyway, for example if we only have one machine of that
architecture and cannot easily replace this.
Simply set IPMI_SKIP_SELFTEST to 1 in order to skip the selftest. By
default, the selftest will be performed.

https://progress.opensuse.org/issues/58466 
and 
https://bugzilla.suse.com/show_bug.cgi?id=1171590
 
http://mmoese-vm.qa.suse.de/tests/96 without IPMI_SKIP_SELFTEST
http://mmoese-vm.qa.suse.de/tests/97 with IPMI_SKIP_SELFTEST=1

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -147,7 +147,7 @@ sub get_mc_status {
 
     $self->ipmitool("mc guid");
     $self->ipmitool("mc info");
-    $self->ipmitool("mc selftest");
+    $self->ipmitool("mc selftest") unless $bmwqemu::vars{IPMI_SKIP_SELFTEST};
 }
 
 sub do_mc_reset {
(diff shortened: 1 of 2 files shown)


backend/ipmi.pm
```diff
@@ -147,7 +147,9 @@ sub get_mc_status {
 
     $self->ipmitool("mc guid");
     $self->ipmitool("mc info");
-    $self->ipmitool("mc selftest");
+    if (!$bmwqemu::vars{IPMI_SKIP_SELFTEST}) {
+        $self->ipmitool("mc selftest");
+    }
```
okurz, line 152: please make that `$self->ipmitool("mc selftest") unless $bmwqemu::vars{IPMI_SKIP_SELFTEST}`

frankenmichl: Yeah sure.

PR #1406, opened 2020-05-15
Handle sound in QEMU via -audiodev as required for qemu 4.2+ - poo#66667

ggardet:
- Related ticket: https://progress.opensuse.org/issues/66667
- Runs:
  - gnome test with fixed qemu: https://openqa.opensuse.org/tests/1293436
  - gnome test with fixed qemu and `QEMU_AUDIODEV=AC97`: https://openqa.opensuse.org/t1293439
  - gnome test with SLE15SP1 host (qemu 3.1.1.1), still uses `-soundhw hda`: https://openqa.opensuse.org/tests/1268124


backend/qemu.pm
```diff
@@ -814,8 +823,22 @@ sub start_qemu {
     sp('only-migratable') if $self->can_handle({function => 'snapshots', no_warn => 1});
     sp('chardev', 'ringbuf,id=serial0,logfile=serial0,logappend=on');
     sp('serial',  'chardev:serial0');
-    my $soundhw = $vars->{QEMU_SOUNDHW} // 'hda';
-    sp('soundhw', $soundhw);
+
+    # poo#66667: Since qemu 4.2, -audiodev is required to record sounds, as we need an audiodev id for 'wavcapture'
+    if (version->declare($vars->{QEMU_VERSION}) lt version->declare(4.2)) {
```
Martchus, line 828: This is only the 2nd repetition of this condition but maybe it makes sense to have a separate function like `requires_audiodev` instead of hard-coding the version number on multiple places.

backend/qemu.pm
```diff
@@ -622,6 +624,13 @@ sub start_qemu {
     $self->{proc}->qemu_bin($qemubin);
     $self->{proc}->qemu_img_bin($qemuimg);
 
+    # Get qemu version
+    my $qemu_version = `$qemubin -version`;
+    $qemu_version =~ /([0-9]+([.][0-9]+)+)/;
+    $qemu_version = $1;
+    $vars->{QEMU_VERSION} = $qemu_version;
```
Martchus, line 631: I know that `bmwqemu::vars` are apparently the place to store everything in this project but wouldn't it make more sense to store the version within the backend object, e.g. `$self->{qemu_version}`?

Martchus:
I'm also wondering where `version->declare` comes from. It seems to be this CPAN module: https://metacpan.org/pod/version

It likely works because the dependency is pulled by another dependency but it would likely be a good idea to add the new dependency explicitly.

backend/qemu.pm
okurz: +1

looks like an abuse of test variables for persistent storage that IMHO does not even need persistent storage

backend/qemu.pm
```diff
@@ -76,10 +76,19 @@ sub _wrap_hmc {
         execute   => 'human-monitor-command',
         arguments => {'command-line' => $cmdline}};
 }
+
+# poo#66667: Since qemu 4.2, -audiodev is required to record sounds, as we need an audiodev id for 'wavcapture'
+sub requires_audiodev {
+    my ($self, $args) = @_;
```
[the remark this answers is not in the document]
ggardet: Done.

ggardet:
> Looks good besides the unused variable.

Fixed.

PR #1409, opened 2020-05-18
Distinguish test compilation errors from backend crashes

Martchus:
* Together with https://github.com/os-autoinst/openQA/pull/3096 the incompletion reasons are more specific about the component (tests vs. backend) when a compilation error occurs, e.g.
    * `tests died: unable to load tests/installation/bootloader.pm, check the log for the cause (e.g. syntax error)`
    * `backend died: Migrate to file failed`
* Runtime errors during the test execution are still treated as before (test result is failed and a failed step with the exception message is added).
* See particular commit messages and https://progress.opensuse.org/issues/64884

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -21,8 +21,9 @@ use Exporter 'import';
 use File::Spec;
 use Cwd;
 use bmwqemu 'diag';
+use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec);
+our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec load_test_schedule);
 
 sub calculate_git_hash {
     my ($git_repo_dir) = @_;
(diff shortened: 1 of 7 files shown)


bmwqemu.pm
```diff
@@ -26,10 +26,12 @@ use Fcntl ':flock';
 use POSIX;
 use Carp;
 use Mojo::JSON;    # booleans
+use Mojo::JSON qw(encode_json);
```
okurz, line 29: use both `use` lines?

isotovideo
```diff
@@ -284,22 +280,36 @@ if ($bmwqemu::vars{SCHEDULE}) {
...
 @INC = @oldINC;
 
 if ($bmwqemu::vars{_EXIT_AFTER_SCHEDULE}) {
     diag 'Early exit has been requested with _EXIT_AFTER_SCHEDULE. Only evaluating test schedule.';
     exit 0;
 }
+
+# start the command fork before we get into the backend, the command child
+# is not supposed to talk to the backend directly
+($cmd_srv_process, $cmd_srv_fd) = commands::start_server($cmd_srv_port = $bmwqemu::vars{QEMUPORT} + 1);
```
okurz, line 311: interesting idea to move this further down :)

OpenQA/Isotovideo/Utils.pm
```diff
@@ -119,4 +120,50 @@ sub checkout_git_refspec {
...
+sub load_test_schedule {
+    # add lib of the test distributions - but only for main.pm not to pollute
+    # further dependencies (the tests get it through autotest)
+    my @oldINC = @INC;
+    unshift @INC, $bmwqemu::vars{CASEDIR} . '/lib';
+    if ($bmwqemu::vars{SCHEDULE}) {
+        diag 'Enforced test schedule by \'SCHEDULE\' variable in action';
+        $bmwqemu::vars{INCLUDE_MODULES} = undef;
+        autotest::loadtest($_ . '.pm') foreach split(',', $bmwqemu::vars{SCHEDULE});
+        $bmwqemu::vars{INCLUDE_MODULES} = 'none';
```
Amrysliu, line 139: just a question, I am wondering why `vars{INCLUDE_MODULES}` is set twice?

okurz: don't worry. This code is just moved from https://github.com/os-autoinst/os-autoinst/pull/1409/files#diff-07616258592e792ab88f484012b38370L283 . The variable is evaluated within `autotest::loadtest` in there within `basetest::is_applicable`

PR #1411, opened 2020-05-19
Remove redundant newline; simplify some code

okurz:
* Simplify some code
* Remove redundant newline in diag message

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2015 SUSE LLC
+# Copyright © 2012-2020 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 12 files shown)


consoles/localXvnc.pm
```diff
@@ -38,9 +38,7 @@ sub callxterm {
     die('Missing "icewm"')        unless which('icewm');
     die('Missing "xterm"')        unless which('xterm');
     eval { system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\""); };
-    if (my $E = $@) {
-        die "cant' start xterm on $display (err: $! retval: $?)";
-    }
+    die "cant' start xterm on $display (err: $! retval: $?)" if $@;
```
perlpunk, line 41: I'm wondering when `$@` is actually set here (and why isn't it used in the error message then?)
Instead it should check `$?` I guess, or save the return value of `system`.

okurz: I would like to preserve functionality for now in this PR but will try to followup.

PR #1416, opened 2020-05-22
Dependencies: add tools/update-deps

perlpunk:

I took the script from openQA and added some parameters, so that it can be used for both repos.
Then we can remove it in openQA and use this one instead.

I tested it in openQA and it works.


perlpunk:
Still have to add a test for it

perlpunk:
Added a test

AdamWill:
why did you put `MakeMaker` in `main_requires`? Surely it should be in `build_requires`?

PR #1418, opened 2020-05-25
testapi: Add wait_still_screen option to send_key

kalikiana:
Fixes: [poo#65109](https://progress.opensuse.org/issues/65109)

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -549,6 +549,12 @@ subtest 'wait_still_screen & assert_still_screen' => sub {
         'assert_still_screen forwards arguments to wait_still_screen');
 };
 
+subtest 'send_key with wait_still_screen' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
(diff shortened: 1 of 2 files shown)


okurz:
change is ok. Instead of adding each option "max_interval/wait_still_screen/wait_screen_change" on each send_key/type_string/script_run/assert_script_run it could be more straightforward to allow passing all arguments and options to the underlying send_key calls. I do not see how this change would fix the whole issue https://progress.opensuse.org/issues/65109

kalikiana:
> change is ok. Instead of adding each option "max_interval/wait_still_screen/wait_screen_change" on each send_key/type_string/script_run/assert_script_run it could be more straightforward to allow passing all arguments and options to the underlying send_key calls. I do not see how this change would fix the whole issue https://progress.opensuse.org/issues/65109

Ack. In hindsight the `Fixes` is wrong here. This was meant to help advance the conversation. I like your suggestion.

PR #1419, opened 2020-05-25
Makefiles: set CXXFLAGS not CFLAGS or CPPFLAGS

AdamWill:
CFLAGS is for C compilation. CPPFLAGS is for the preprocessor.
CXXFLAGS is for C++ compilation, which is what we're actually
trying to apply the flags to in all three cases. CPPFLAGS happen
to be applied to the compile stage as well as the preprocessor
usually, so CPPFLAGS was working, but it wasn't correct. CFLAGS
is not applied to C++ compiles, so `videoencoder_CFLAGS` was
not being used at all.

This broke build on Fedora Rawhide, because /usr/include/opencv2
no longer exists there. There is /usr/include/opencv4/opencv2 and
this should be fine because we should get -I/usr/include/opencv4
from OPENCV_CFLAGS, but because of this bug, that wasn't set
correctly and the build failed. With this change it works.

Thanks to Orion Poplowski for reminding me of the CPPFLAGS vs.
CXXFLAGS distinction here.


AdamWill:
400 Bad Request

    {"meta": {"status": 400}, "error": {"reason": "Please provide the repository token to upload reports via `-t :repository-token`", "context": null}}

    make: *** [Makefile:1555: coverage-codecov] Error 255

    The command "tools/docker_run_ci" exited with 2.

doesn't look like my problem. :)

AdamWill:
Hum, in fact, as Orion Poplowski pointed out [here](https://bugzilla.redhat.com/show_bug.cgi?id=1839616#c3), should we change all three of these to be `CXXFLAGS`?

Martchus:
I've confused `CPP` with `CXX` when approving the PR. The flags for OpenCV need to be passed to the compiler so `CXXFLAGS` would make more sense. Likely the processor isn't invoked separately anyways.

It also seems wrong to hardcode flags like `-O2 -Wall -W` here. But I'd leave that for another PR. (Maybe we can finally switch to CMake at some point anyways.)

AdamWill:
OK, so this changes to `CXXFLAGS` in all three places.

I agree about the hardcoding - it seems bad but should be fixed separately...

PR #1420, opened 2020-05-27
Dependencies: Add code to update Dockerfile

perlpunk:
Issues:
* https://progress.opensuse.org/issues/56525
* https://progress.opensuse.org/issues/66325

Also
* Remove some more unnecessary dependencies.
* Add more perl modules explicitly which are in perl core

Verification run on a docker image in a branched OBS repo:
https://travis-ci.org/github/perlpunk/os-autoinst/builds/691653585


okurz:
this looks nice

PR #1421, opened 2020-05-28
Support using an external video encoder

Martchus:
* When an external video encoder is configured, the built-in videoencoder
  is only used for producing the last PNGs anymore (exactly like when
  NOVIDEO is used)
* This allows invoking an arbitrary binary by setting a certain test
  variable; merge the corresponding openQA change first to avoid this
* See changes in backend_vars.asciidoc for the configuration details
* See https://progress.opensuse.org/issues/67342


Martchus:
While this PR basically works I encountered one flaw: The external encoder is not shut down "appropriately". At least ffmpeg leaves us with a truncated Matroska file without index, e.g.:

```
parsing header of EBML element 0x18538067 "segment" at 36:
Data of EBML element seems to be truncated; unable to parse siblings of that element.
validating Matroska file index (cues):
No "Cues"-elements (index) found.
```

This should be fixed because otherwise seeking doesn't work nicely in most players.

Martchus:
These diagnostic messages are from my tag editor by the way. I've just opened an Ogg file produced by the built-in video encoder and apparently it is truncated as well:

```
parsing OGG bitstream header:
The OGG file is truncated.
```

For Ogg this is less problematic than for other format so we likely just didn't care/notice so far. (Or it is a bug in my tag parsing library. I don't know another tool to cross check because ffmpeg only complains about the keyframe issue.)

Martchus:
All jobs/videos I've tested last Friday were cancelled and therefore terminated by the worker. After a regular exit none of the videos is not truncated. (`ffmpeg` seems to catch `SIGINT` and `SIGTERM` and "finalizes" the file before existing.) So this problem needed to be addressed on worker-level.

doc/backend_vars.asciidoc
```diff
@@ -16,7 +16,10 @@ _EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation
...
 TEST_GIT_REFSPEC;string;;git refspec to checkout within `CASEDIR` when `CASEDIR` is a git working copy. By default, does not change the content of `CASEDIR`. Overrides the optional git refspec in `CASEDIR`. Can be used to explicitly select a git commit within an existing git working copy and also to skip unnecessary git network transfers when `CASEDIR` is already providing the right git working copy.
 NEEDLES_GIT_REFSPEC;string;;git refspec to checkout within `NEEDLES_DIR`. See `TEST_GIT_REFSPEC` for details.
-NOVIDEO;boolean;0;Do not encode video if set
+EXTERNAL_VIDEO_ENCODER_CMD;string;;Specifies the command line for invoking a custom video encoder. It which is supposed to accept a sequence of PPM images via stdin. The placeholder `%OUTPUT_FILE_NAME%` is replaced with the output file path. The output file path is appended if it is missing. Example: `ffmpeg -r 24 -f image2pipe -vcodec ppm -i - -c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 0`
```
okurz, line 19: "It which is supposed to accept" sounds ill-formed. Not sure what you want to say here, please rephrase

doc/backend_vars.asciidoc
```diff
@@ -16,7 +16,10 @@ _EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation
...
+EXTERNAL_VIDEO_ENCODER_OUTPUT_FILE_EXTENSION;string;webm;The extension of the output file when `EXTERNAL_VIDEO_ENCODER_CMD` is used.
+EXTERNAL_VIDEO_ENCODER_ADDITIONALLY;boolean;0;Whether `EXTERNAL_VIDEO_ENCODER_CMD` should only be invoked additionally to the built-in Theora encoder. (Configuring an external video encoder disables the built-in one by default.)
```
okurz, line 21: Can you state what is the consequence of only adding an additional video encoder: Does that mean two video files will be generated as part of uploadable data?

doc/backend_vars.asciidoc
```diff
@@ -16,7 +16,10 @@ _EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation
...
+NOVIDEO;boolean;0;Do not encode video if set; `EXTERNAL_VIDEO_ENCODER_` variables are ignored as well.
```
okurz, line 22: I wonder, how is this additional ";" rendered. Will it add another column to the complete table?

doc/backend_vars.asciidoc
Martchus: I wanted to break the long sentence into 2 smaller ones and forgot to delete the "which".

doc/backend_vars.asciidoc
Martchus: It means exactly that. Maybe I should also mention that this mode is intended for having a comparison.

doc/backend_vars.asciidoc
Martchus: No, it apparently screws up the table completely as the cell will be part of the next row. Good catch.

Martchus:
The documentation issues are fixed now (preview: https://github.com/Martchus/os-autoinst/blob/videoencoder-experimental/doc/backend_vars.asciidoc).

PR #1423, opened 2020-05-29
CI: Install possibly new dependencies

perlpunk:
This way we can add a new dependency and use it in the same PR.

Here is how it looks like if a new dependency was added:
https://travis-ci.org/github/perlpunk/os-autoinst/builds/692593416#L591-L594
And if no new dependencies are there:
https://travis-ci.org/github/perlpunk/os-autoinst/builds/692591735#L577-L578


tools/install-deps.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/sh
+
```
Martchus, line 2: This file is generated, right? Maybe it makes sense to add a comment stating that e.g. `Do not edit manually; generated by $tool_name`.

It would also make sense to mention that this file is used by the CI.

I'm also wondering whether this could be generated on the fly. Generally I'd avoid having generated files within Git.

perlpunk: An advantage is, that I can reuse the `install-deps.sh` script for the next task, installing refreshed dependencies from devel:languages:perl. I only need to pass a parameter from which repo to install, so I believe this will be the last file that I generate, making it 4 files in total (cpanfile, spec, Dockerfile, install-deps.sh).

okurz:
I am also concerned about storing the generated file within git. Maybe just configuration files are better than scripts. And as you already came up with a new format for dependencies then we should rather use this file.

```
zypper -n in $(yamlpp-load-dump --no-header dependencies.yaml | sed -n 's/  \([^%]*\): .*$/\1/p')
```

should work

perlpunk:
@okurz that'd clever ;-)
I think it could break easily if other things are added to the yaml file.

Insead I came up with this:
```
grep '^# AUTODEPS START' -A 999 docker/travis_test/Dockerfile \
  | grep '^# AUTODEPS END' -B 999 \
  | grep '^RUN ' -A 999 | grep -v '# AUTODEPS' | tail -n+2
```

perlpunk:
Or from dependencies.yaml:
```
perl -MYAML::PP -wE'
my $d = YAML::PP::Load(do { local $/; <> });
for (@{ $d->{targets}->{docker} }) {
 $_ = "${_}_requires" unless $_ eq "requires";
 push @d, keys %{$d->{$_}}
}; say for sort @d' < dependencies.yaml
```

perlpunk:
I'd vote for renaming `%requires` in the spec to `%main_requires` like in openQA.
Will save a bit of code.

perlpunk:
Created #1430 without a new generated file.

perlpunk:
Closing in favor of #1430

PR #1427, opened 2020-06-05
WIP Proof of Concept for using git-subrepo

perlpunk:
## How does [git-subrepo](https://github.com/ingydotnet/git-subrepo) work

The contents of the subrepo are directly stored in the repo. That means users don't need to have git-subrepo installed.
Only developers who want to update it.
This avoids some of the annoying things of git-submodule.

A good intro and comparison between subrepo, submodule and subtree:  https://github.com/ingydotnet/git-subrepo/blob/master/Intro.pod

It creates a `.gitrepo` file with necessary information about the subrepo:
```
% cat ext/test-openqa-subrepo.git/.gitrepo
; DO NOT EDIT (unless you know what you are doing)                                                                                                                                                                                                         
;
; This subdirectory is a git "subrepo", and this file is maintained by the
; git-subrepo command. See https://github.com/git-commands/git-subrepo#readme
;
[subrepo]
        remote = git@github.com:perlpunk/test-openqa-subrepo.git
        branch = devel
        commit = c35d3801c86eeef745a1d6cf49e191c6b74f37b6
        parent = 7ef1cfff732a147655ef2825ffa063e843f88e8b
        method = merge
        cmdver = 0.4.1
```

## Creating the subrepo

I used the branch `devel`:
```
git subrepo clone git@github.com:perlpunk/test-openqa-subrepo.git ext/test-openqa-subrepo.git -b devel
```

The command for creating the subrepo is automatically used as the commit message:

```
git subrepo clone --branch=devel git@github.com:perlpunk/test-openqa-subrepo.git ext/test-openqa-subrepo.git
```

## Pull in new changes

New changes from the subrepo can be pulled in via:

```
git subrepo pull ext/test-openqa-subrepo.git
```

It also has completion support, so you can do:
```
git subrepo pull <TAB>
```

## Replace script with link to script from subrepo

In the second commit I replaced `tools/update-deps` with a symlink to `ext/...`.

## Make changes to subrepo

It's possible to make changes in the subrepo contents, which I showed in the third commit.
I edited `ext/test-openqa-subrepo.git/tools/update-deps`, committed, and did:
`git subrepo push ext/test-openqa-subrepo.git`

This makes the change in the real subrepo and pushes.


okurz:
ok, interesting approach. I think I would prefer git submodules. What do you see as problematic with git submodules? Anyone using git has all that is necessary and is without any extra work except for a simple "--recursive" option on clone

perlpunk:
@okurz here is a good intro and comparison: https://github.com/ingydotnet/git-subrepo/blob/master/Intro.pod
You will find a section "Git submodules", listing all the pros and cons.

One point I already mentioned is, that users (or CI) cloning this repo do not need to explicitly fetch submodules. It just works.
It's also very easy to make temporary local changes without needing to push to the submodule.

One thing git-subrepo is not suitable for are large subrepos, because the content is stored directly as a copy.

Overall it's simply easier to maintain.

perlpunk:
Additionally, yesterday I did my first contribution to open-build-service, which is using submodules.

master was updated, so I rebased my branch, and suddenly git told me there is a modified file.
It turned out I had to `git submodule update`.
This is not necessary with subrepo.

Martchus:
I generally think sub *modules* are not as bad to work with as some people say and I'm not sure whether it is worth introducing an additional tool.

I'm also wondering what speaks against having these scripts directly within this repository and even install them and package them as part of the `-devel` package. Then e.g. openQA or any other project could easily use them. There would be no "extra complexity" required at all.

perlpunk:
@Martchus regarding "additional tool": subrepo is an additional tool just like submodule. Only that submodule is in git-core.
subrepo will soon be packaged in Ubuntu and Debian, too, it's currently prepared.

Regarding "extra complexity": It might seem so if you never used subrepo. But the usage is easier than submodule, once you used it for the first time.

If we made update-deps a dev tool that gets installed, that's also work that needs to be done, and for every change it requires updating and publishing os-autoinst first before we can use it elsewhere. To me, that's also "extra complexity". (Although I hope there won't be much changes necessary in the future, I certainly expect some changes before things are settling).

If we decide against subrepo, I would vote also against using submodules.

okurz:
The subrepo approach sounds like an easy compromise when one is torn if any dependency should be managed within the same repo or a separate one. That this is something recent and not available in openSUSE Leap is a downside though. Basically I am open to try any approach, not voting against anything :)

kalikiana:
I think this might be a lot of detail, obscuring the user story a little bit. And the already mentioned [comparison between subrepo, submodule and subtree](https://github.com/ingydotnet/git-subrepo/blob/master/Intro.pod) is missing a tl;dr; perhaps leaving you to wonder what it looks like in real life.

From what I understand, **subrepo** has one massive advantage in that most users really don't need to care about it. They will `clone` and `pull` like they've always done without realizing it's not just a *normal* git repo. So your `git clone ...` remains the same, before and after.
If you want to contribute, it boils down to variants of regular git commands:

    $ git subrepo clone git@github.com:my/thing foo -b mybranch
    $ git subrepo pull foo
    $ git subrepo push foo 

Conversely **submodule** requires that you *know* the steps needed to update it, which could be `git submodule update`, `git submodule init` or perhaps `git clone --recursive ...` if you have a more recent git version installed. Updates will also require `git submodule update` calls in tandem with updating the parent branch (or *super project*).

In comparison **subtree** shares the benefit of not requiring users to be ware of its existence. But the (commit history and the) developer experience is rather verbose:

    $ git subtree add --squash --prefix=foo git@github.com:my/thing mybranch
    # weeks go by…
    $ git subtree pull --squash --prefix=foo git@github.com:my/thing mybranch
    # time to push local subtree changes back upstream
    $ git subtree push --prefix=foo git@github.com:my/thing mybranch

> The subrepo approach sounds like an easy compromise when one is torn if any dependency should be managed within the same repo or a separate one. That this is something recent and not available in openSUSE Leap is a downside though. Basically I am open to try any approach, not voting against anything :)

There is no added dependency for users of it who don't attempt to contribute back to the repo.

PR #1429, opened 2020-06-05
Deps: rename %requires to %main_requires

perlpunk:
as all other targets are named like this. Makes it easier
to automate dependency processing

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -8,9 +8,9 @@
 #######################################################
 ---
 targets:
-  cpanfile: [        cover, devel,         requires,             test ]
-  spec:     [ build,        devel,         requires, spellcheck, test ]
-  docker:   [ build,        devel, docker, requires, spellcheck, test ]
+  cpanfile: [        cover, devel,         main,             test ]
+  spec:     [ build,        devel,         main, spellcheck, test ]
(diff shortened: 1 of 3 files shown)


perlpunk:
Travis doesn't report the build, but it passed:
https://travis-ci.org/github/os-autoinst/os-autoinst/builds/695052272

okurz:
hey, that was a bit fast. The name "main_requires" refers to "main subpackage within openQA aka. webui package". That does not make any sense for os-autoinst. Please change that back

perlpunk:
@okurz I created #1431 but I don't understand why this is a problem. Can you clarify?

okurz:
thanks. We don't need to necessarily rush changing it back but relying on the name "main_requires" is misleading and not scalable. As I stated "main" refers to the main sub-package of openQA so having your dependency management script depend on that name is not the best approach.

perlpunk:
> As I stated "main" refers to the main sub-package of openQA

I still don't understand it.

> having your dependency management script depend on that name is not the best approach

Before my change it was relying on the name `requires`. But since that is kept in the same repo, I don't see a problem with changing a name.
Does that mean we can never rename that definition?

okurz:
of course we are free to rename. Ok, nevermind. Just keep your changes as is.

PR #1430, opened 2020-06-05
CI: Install possibly new dependencies

perlpunk:
This way we can add a new dependency and use it in the same PR.
    
In the future `tools/tools.sh` can also be used within the Dockerfile,
so we don't have to generate the dependencies there.
    
For that OBS needs to be configured to extract dependencies.yaml and
tools/tools.sh first.

Followup-to: #1423

### DIFF
--- a/tools/docker_run_ci
+++ b/tools/docker_run_ci
@@ -8,5 +8,5 @@ docker images | grep opensuse
 
 docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_QEMU_START_S=20" \
   $CI_ENV --rm --entrypoint '' -v "$PWD":/opt $IMAGE \
-  sh -c 'cd /opt && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
+  sh -c 'cd /opt && ./tools/install-new-deps.sh && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
 
--- a/tools/install-new-deps.sh
+++ b/tools/install-new-deps.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+# This script is used for CI testing.
+# Used to install new dependencies.
+# If there are new dependencies, they won't be installed in the docker
+# container yet, so we just install all deps again.
+
+set -e
+
+OLDDEPS=/tmp/deps.txt
+NEWDEPS=/tmp/new-deps.txt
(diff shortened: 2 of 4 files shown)


perlpunk:
Again, travis does not report. Build passed:
https://travis-ci.org/github/os-autoinst/os-autoinst/builds/695101673

tools/install-deps.sh
```diff
@@ -0,0 +1,21 @@
...
+        push @d, grep {!/^%/} keys %{ $d->{$_ . "_requires"} }
+    };
+    say for sort @d' < dependencies.yaml
+}
+
+DEPS=($(getdeps))
+
+#echo "DEPS: ${DEPS[@]}"
+
+sudo zypper --no-refresh install -y -C ${DEPS[@]}
```
okurz, line 21: I consider the approach I suggested easier than the additional shell script with embedded procedural perl code (At least the double for-loop looks quite unusual. In that case I would go for a `say sort map { … } grep { … } …`)

perlpunk: The problem is, I think your solution could break:
https://github.com/os-autoinst/os-autoinst/pull/1423#issuecomment-639269454
if we ever add
```yaml
something else with:
  a key: and a value
```
Then we need to change the command anyway.

perlpunk: This is the shortest code (with some whitespace left) I came up with:
```
perl -MYAML::PP=Load -0wE'my $d = Load(<>);
say for sort grep {!/^%/} map { keys %{$d->{$_."_requires"}} }
@{$d->{targets}->{docker}}' < dependencies.yaml
```

okurz: I like that more :) The positive point about your solution is that it uses a proper YAML parser, the negative is still the "perl embedded in shell script, called by container script" but it's ok

perlpunk: I rearranged it a bit now.
There is `tools/tools.sh` which simply defines the function `getdeps_docker`.
It can be sourced from the script that installs dependencies.

We could also use it in Docker, do we don't have to generate the Dockerfile.
We only need to extract `dependencies.yaml` and `tools/tools.sh` in OBS. If we ever need to change/add something, we can simply edit `tools.sh`.

perlpunk:
There is actually a problem with calling `diff`: https://travis-ci.org/github/os-autoinst/os-autoinst/builds/696418082#L586
Need to figure out what happens

perlpunk:
rebased

PR #1431, opened 2020-06-05
Revert "Deps: rename %requires to %main_requires"

perlpunk:
Reverts os-autoinst/os-autoinst#1429

See comment https://github.com/os-autoinst/os-autoinst/pull/1429#issuecomment-639693218


okurz:
Thanks but I guess we can go without. Nevermind the confusion and sorry.

PR #1435, opened 2020-06-08
Move MakeMaker dependency to build_requires

AdamWill:
It is used at build time, not run time.


AdamWill:
@perlpunk

PR #1422, opened 2020-05-29
Capture isotovideo output in isotovideo test

kalikiana:
Note: The difference can seen with verbose output, i.e. `make testv TESTS=14-isotovideo.t`

PR #1436, opened 2020-06-08
Move git to %main_requires

AdamWill:
See:
https://github.com/os-autoinst/os-autoinst/pull/1422#issuecomment-640941787
since that change, it seems the tests fail if git is not present.
That means we need git at both runtime and test time, so we can
just put it in `%main_requires` and remove the concept of
`%requires_not_needed_in_tests` entirely as git was the last
thing in it.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -34,13 +34,11 @@ cover_requires:
 
 devel_requires:
   '%test_requires':
-  '%requires_not_needed_in_tests':
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
   perl(Perl::Tidy): '== 20200110'
 
 docker_requires:
-  git:
   pkgconfig(opencv):
   sudo:
   which:
@@ -77,6 +75,7 @@ test_requires:
   qemu-x86:
 
 main_requires:
+  git-core:
   perl(Cwd):
   perl(B::Deparse):
   perl(Carp):
--- a/docker/travis_test/Dockerfile
+++ b/docker/travis_test/Dockerfile
@@ -20,7 +20,7 @@ RUN zypper in -y -C \
        autoconf \
        automake \
        gcc-c++ \
-       git \
+       git-core \
        libtool \
        make \
        perl-base \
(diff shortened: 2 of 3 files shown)


AdamWill:
@kalikiana

docker/travis_test/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN zypper in -y -C \
        automake \
        gcc-c++ \
        git \
+       git \
```
Martchus, line 24: This is now duplicated. I suppose you have to remove `git` from `docker_requires` which apparently "inherits" from `test_requires` without de-duplication.

perlpunk:
There is this line in `dist/rpm/os-autoinst.spec` (ab0554c2a9f31f5f8d78604be993ff625563e844):
```
%define requires_not_needed_in_tests git-core
```
So for some reason it was explicitly excluded.

So maybe we should remove that definition?

docker/travis_test/Dockerfile
perlpunk: It doesn't inherit, but here is the definition of the docker dependencies:
https://github.com/os-autoinst/os-autoinst/blob/master/dependencies.yaml#L13

So yes, it should be removed from `docker_requires`.

okurz:
> So maybe we should remove that definition?

If build+test still works within OBS we can change that

kalikiana:
> @kalikiana

Thank you for catching it, and apologies for the disruption. The test was "pretending" to check this before and I didn't realize it wasn't a dep already.

AdamWill:
I noticed the duplication but figured it probably doesn't cause any practical problems, so why worry about it?

On `%requires_not_needed_in_tests` - yeah, that exists to express "things os-autoinst needs at runtime in normal use, but which don't need to be installed to run the test suite". Up until #1422 was merged that was the case for git, but since #1422 the tests fail (at least for me) if git isn't installed.

So thinking about it, that probably means that instead of adding it to `%test_requires` as well, we can simply add it to `%main_requires` and drop it from `%requires_not_needed_in_tests`.

One other thing this flags up: `git` vs. `git-core`. `git-core` is better for Fedora too (it gets you `/usr/bin/git` without some unnecessary fripperies), but I didn't know if we could put it in `dependencies.yaml`, I'm not sure if it's correct for...all other things that use that?

AdamWill:
Oh, forgot to mention - obviously an alternative change here would be to change the tests again so they pass if git isn't installed. But I don't know/don't recall whether tests passing without git installed was an explicit *goal* or *desire*, or just something that happened to be the case before.

perlpunk:
Regarding git vs. git-core - yeah, we should use `git-core` I think.

AdamWill:
OK, changed.

dependencies.yaml
```diff
@@ -34,13 +34,11 @@ cover_requires:
 
 devel_requires:
   '%test_requires':
-  '%requires_not_needed_in_tests':
```
okurz, line 37: why change that as well?

AdamWill: as the commit message says: "so we can...remove the concept of %requires_not_needed_in_tests entirely as git was the last thing in it."

okurz: I see. this key was only needed in spec and only for git. Fine

PR #1437, opened 2020-06-09
Add Pod::Html to build_requires

AdamWill:
It's needed to run pod2html as part of docs build. It used to
get pulled into package builds as a dependency of Perl::Tidy
which was in test_requires, but that got moved to devel_requires.


perlpunk:
hm, maybe we should put Perl::Tidy back to `test_requires`?

Martchus:
Why was it moved out from `test_requires` in the first place? Considering we have this tidy check it would make sense to move it back. But we could still list the dependency to `Pod::Html` explicitly.

perlpunk:
> Why was it moved out from `test_requires` in the first place?

Not sure. I think I thought since we do not do the tidy check in OBS, we don't need it there.
For CPAN modules, such tests are usually run as "Author tests", which are run by the author/developer/CI, but not when installing the module.

okurz:
Just to explain what was maybe not clear in the spec-file:

* *build_requires*: Everything that is needed to build the package (not runtime requirements and also not requirements to execute %check)
* *main_requires*: runtime requirements
* *requires_not_needed_in_tests*: runtime requirements that are *not* needed to execute %check from spec
* *devel_requires*: runtime requirements for the "devel" package that is to be used in all development environments, e.g. when a human developer wants to build, test, debug os-autoinst

That means: As long as %check in OBS does not fail then nothing else should be added to "test_requires"

AdamWill:
The RPM spec actually [explicitly disables the tidy checks](https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec#L135).

Obviously there's a slight tension with trying to consolidate the dependency statements in a single file, if we do slightly different things in the spec vs. in CI for instance.

Martchus:
Then let's keep it under *devel_requires*. Maybe we can add yet another category for CI runs later.

AdamWill:
rebased.

okurz:
So I am not sure now. As I stated, we don't need Pod::Html in the actual build requirements as per https://github.com/os-autoinst/os-autoinst/pull/1437#issuecomment-641376049 . I assume you saw a problem in the build for Fedora. Do you use the same spec file or just the cpanfile or what?

AdamWill:
Your comment doesn't say that? It just defines what things are.

We use a very similar spec file with the same %prep and %build, pretty much. I ran a build without this dep and it failed and that doesn't seem like a surprise, because building the docs clearly needs `pod2html` - see [here](https://github.com/os-autoinst/os-autoinst/blob/master/Makefile.am#L159). Are you saying you can do an RPM build without perl(Tidy) in the test requirements and it succeeds? Can you check the logs from that build and see if pod2html gets installed some other way or it's not run for some reason?

AdamWill:
https://koji.fedoraproject.org/koji/taskinfo?taskID=45559000 was my failed test build. Note this in [build.log](https://kojipkgs.fedoraproject.org//work/tasks/9000/45559000/build.log):

    pod2html basetest.pm > doc/basetest.html
      CXX      videoencoder-videoencoder.o
    pod2html testapi.pm > doc/testapi.html
    cd ppmclibs && \
    perl Makefile.PL OPTIMIZE="-O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection" CC="g++"
    /bin/sh: pod2html: command not found
    make[1]: *** [Makefile:1457: doc/basetest.html] Error 127
    make[1]: *** Waiting for unfinished jobs....
    /bin/sh: pod2html: command not found
    make[1]: *** [Makefile:1457: doc/testapi.html] Error 127

it's mixed up with other tasks as it's a parallel build, but those are the fatal errors.

okurz:
I checked https://build.opensuse.org/public/build/openSUSE:Factory/standard/x86_64/os-autoinst/_log and found that tidy *is* installed even though it's not specified as direct dependency. pod2html is called so specifying the dependency in a way as you did is correct.

perlpunk:
The issue is that Pod::Html is in perl core: http://perlpunks.de/corelist/mversion?module=Pod%3A%3AHtml
but probably not packaged in Fedora perl core?
That's why we don't see a problem on our side.

perlpunk:
Btw, how does Perl::Tidy depend on Pod::Html? Can't see anything in Makefile.PL, META.json or META.yml
https://metacpan.org/release/Perl-Tidy

PR #1438, opened 2020-06-09
Remove tools/preparepool

perlpunk:
Last update was 2013, and grepping for it in os-autoinst, openQA,
os-autoinst-distri-opensuse returned nothing.

This also allows us to completely delete the tools folder in the spec.


perlpunk:
I made #1440, so this PR removes the directory completely.

PR #1439, opened 2020-06-09
Allow video encoder to finalize the file

Martchus:
This has the same problem as https://github.com/os-autoinst/openQA/pull/3152, see https://progress.opensuse.org/issues/67342?issue_count=241&issue_position=1&next_issue_id=67144#note-10 but it is a start.


backend/driver.pm
```diff
@@ -98,7 +98,7 @@ sub start {
 
             $self->{backend}->run(fileno($process->channel_in), fileno($process->channel_out));
             _exit(0);
-    }, sleeptime_during_kill => .1)->blocking_stop(1)->separate_err(0)->subreaper(1)->start;
+    }, sleeptime_during_kill => 10)->kill_sleeptime(0)->blocking_stop(1)->separate_err(0)->subreaper(1)->start;
```
okurz, line 101: I think you found out by now that this is reverting a change I introduced to speedup the shutdown of processes. Your effect is also shown by tests when they time out.

backend/baseclass.pm
```diff
@@ -362,6 +365,33 @@ sub start_encoder {
...
+
+    # give the video encoder processes 30 seconds to finalize the video
+    # note: Sending SIGTERM only once because successive SIGTERM signals might be interpreted by the video encoder as
+    #       request to stop it forcefully.
+    no autodie qw(kill waitpid);
+    kill TERM => (keys %$video_encoder_pids);
+    bmwqemu::diag 'Waiting for video encoder to finalize the video';
+    for (my $timeout = 30; $timeout > 0; sleep($timeout -= 1)) {
+        for my $pid (keys %$video_encoder_pids) {
+            my $ret = waitpid($pid, WNOHANG);
```
okurz, line 382: That's now a rather low-level approach of what should be handled by the process management framework. Maybe we need to use an alternative to Mojo::IOLoop::ReadWriteProcess with just Mojo::IOLoop or also IPC::Run how I recently have done in openQA tests replacing manual fork calls

backend/driver.pm
Martchus: Yes, I've mentioned it also in the chat. The problem is that the stop function of Mojo::IOLoop::ReadWriteProcess simply makes no sense. You're either hammering the process with SIGTERM or wasting time.

backend/baseclass.pm
Martchus: I've been fighting against Mojo::IOLoop::ReadWriteProcess long enough so I though let's go back to the roots. Besides: Mojo::IOLoop is not used here in favor of our own `run_capture_loop`. Switching to Mojo::IOLoop in a consistent way would mean rewriting that and I don't think it is worth it. The process handling code within the backend is also generally low-level. We're using the pipes directly (within that `run_capture_loop`) and I don't want any framework to interfere with that.

foursixnine:
> This PR seems¹ to work now when used together with [os-autoinst/openQA#3152](https://github.com/os-autoinst/openQA/pull/3152). However, [mudler/Mojo-IOLoop-ReadWriteProcess#9](https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/pull/9) needs to be merged, released and added to our repositories first. I also still need to bump the required version in the spec file.
> 
> ¹ It still needs more testing.

Release is done, should be in the repos by now

okurz:
not yet. https://build.opensuse.org/package/show/openSUSE:Factory/perl-Mojo-IOLoop-ReadWriteProcess shows 0.27 is in Factory but Tumbleweed still shows 0.25 , see http://download.opensuse.org/tumbleweed/repo/oss/noarch/?P=*ReadWriteProcess*

PR #1442, opened 2020-06-10
spec: use 'check-doc' Makefile target, not 'check'

AdamWill:
The 'check' Makefile target basically runs code quality checks
then runs the 'check-doc' target (which runs spellchecks) as
long as CHECK_DOC isn't 0. In the RPM spec, we're calling 'make
check', but before that, we are hackily disabling all the code
quality checks, so all the 'check' target winds up doing is
calling 'check-doc' (except for SLE, where we set CHECK_DOC=0
because SLE doesn't have the necessary deps). I think it'd make
more sense to drop use of the 'check' target entirely, and just
call 'check-doc' directly when we want to do the spellchecks.

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -46,10 +46,8 @@ Source0:        %{name}-%{version}.tar.xz
 %if %{with spellcheck}
 # The following line is generated from dependencies.yaml
 %define spellcheck_requires aspell-en aspell-spell perl(Pod::Spell)
-%define make_check_args %{nil}
 %else
 %define spellcheck_requires %{nil}
-%define make_check_args CHECK_DOC=0
 %endif
 # The following line is generated from dependencies.yaml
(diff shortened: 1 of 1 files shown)


AdamWill:
So, what change are you requesting?

okurz:
Well, the extreme cases are: 1. Close PR without any changes or 2. overrule me :)

I would prefer if we still call `make check`. If you have a good idea how to do that but still keep a part of your idea then change accordingly. If not, maybe I have missed what other benefit your PR has?

AdamWill:
I mean, I guess the thing I'm wondering is, what do you envisage potentially being a thing that should be put in `make check` and not `make test`, but which should be run on package builds? What do you see as the distinction between `check` and `test` exactly?

okurz:
Good question. If it wouldn't be for autotools I would ditch "check" in favor of "test" completely

Martchus:
I suppose this PR is no longer required.

I've changed it to `%cmake_build check-pkg-build` which should behave similar to the change proposed here: The CMake build script will automatically conduct the spell checking via that target if `podspell/spell` are installed. (If that's not the case a status message is logged during the configuration stating that the tests are disabled.)

AdamWill:
"The CMake build script will automatically conduct the spell checking via that target if podspell/spell are installed. (If that's not the case a status message is logged during the configuration stating that the tests are disabled.)"

That seems reasonable to me. Thanks!

PR #1444, opened 2020-06-11
Fix test for undocumented backend variables

ilausuch:
https://progress.opensuse.org/issues/34303

Problem: When executing the test manually it shows some errors
but the exit code is 0

Solution: Force an exit with 1 error code.

Problem: Some vars don't have description. With this PR this generates the failure
of the test.

Solution: A new description has been added

Problem: When executing the test manually the method write_doc
shows some errors but exit code is 0. And some vars don't have
description or aren't on the file backend_vars.

Solution: Force an exit with 1 error code. And fix or add the
required variables and description.

Problem: There are redundancies in IKVM. All required variables
are included in IPMI declarations

Solution: Force to check IPMI as IKVM documented variables

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -38,17 +38,6 @@ AUTOINST_URL_HOSTNAME;string;;hostname or IP adress of host running the autoinst
 
 |====================
 
-.IKVM backend
-[grid="rows",format="csv"]
-[options="header",cols="^m,^m,^m,v",separator=";"]
-|====================
-Variable;Values allowed;Default value;Explanation
-IPMI_HOSTNAME;string;undef;Hostname/IP for IPMI interface
-IPMI_HW;string;supermicro;Hardware used for IPMI interface
-IPMI_PASSWORD;string;undef;Password for the IPMI interface
-IPMI_USER;string;undef;Username for the IPMI interface
-|====================
-
 .IPMI backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -38,17 +38,29 @@ AUTOINST_URL_HOSTNAME;string;;hostname or IP adress of host running the autoinst
 
 |====================
 
+.IKVM backend
+[grid="rows",format="csv"]
+[options="header",cols="^m,^m,^m,v",separator=";"]
+|====================
+Variable;Values allowed;Default value;Explanation
+IPMI_HOSTNAME;string;undef;Hostname/IP for IPMI interface
+IPMI_HW;string;supermicro;Hardware used for IPMI interface
+IPMI_PASSWORD;string;undef;Password for the IPMI interface
+IPMI_USER;string;undef;Username for the IPMI interface
+|====================
+
 .IPMI backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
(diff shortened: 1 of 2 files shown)


t/04-check_vars_docu.t
```diff
@@ -70,17 +73,22 @@ sub read_doc {
...
                 my ($var, $value, $default, $explanation) = $line =~ /^([^;]+);\s*([^;]*);\s*([^;]*);\s*(.*)$/;
+
                 next unless ($var);
                 $default = '' unless (defined $default);
                 $value   = '' unless (defined $value);
                 unless ($explanation) {
-                    say "still missing explanation for backend $backend variable $var";
+                    say "" if ($errors == 0);
+                    say color ('red') . "Error: still missing explanation for backend $backend variable $var";
+                    $errors++;
```
perlpunk, line 83: The current output is:
```
04-check_vars_docu.t ................... 
Dubious, test returned 1 (wstat 256, 0x100)
```
This usually means that a test was aborted unexpectedly.
I think it would be better to do
`fail "Error: still missing explanation for backend $backend variable $var"`

Also a note about `color()`: this will add the escape sequence for the color, but you never add the escape sequence for ending the color, so all output after that would be colored too.
It's better to use `colored()`.

okurz:
When we use functions from Test::More I think there would be no need for any additional handling of exit code nor colors but we can only merge a PR if tests are passed so you would need to fix the undocumented entries anyway.

t/04-check_vars_docu.t
kalikiana: I would also prefer the `fail` line suggested above, with no extra color. That would create failures in the test plan, which is what we want.

ilausuch: Done

ilausuch:
- Applied last review suggested changes
- Added description to the documentation vars that was failing

In the sub write_doc we have the same problem there are missing descriptions, and it is using say instead of fail. Should it be fixed too?

ilausuch:
Added the modifications for the write test method and added the necessary variables to backend_vars.asciidoc

ilausuch:
Rebased from master

doc/backend_vars.asciidoc
```diff
@@ -38,17 +38,32 @@ AUTOINST_URL_HOSTNAME;string;;hostname or IP adress of host running the autoinst
...
+IPMI_HOSTNAME;string;undef;Hostname/IP for IPMI interface
+IPMI_PASSWORD;string;undef;Password for the IPMI interface
+IPMI_USER;string;undef;Username for the IPMI interface
+IPMI_DO_NOT_POWER_OFF;boolean;undef;Don't power off the machine after test
 IPMI_BACKEND_MC_RESET;boolean;undef;Reset ipmi main board before test for sol console stability
+<<<<<<< HEAD
 IPMI_SKIP_SELFTEST;boolean;undef;Don't perform BMC selftest
+=======
+IPMI_HW;string;supermicro;Hardware used for IPMI interface
+>>>>>>> Fix documentation vars for write test
```
okurz, line 66: You have conflict markers here that need to be Resolved

ilausuch:
Fixed

okurz:
You only fixed the conflict markers, not the duplication as mentioned in https://github.com/os-autoinst/os-autoinst/pull/1444#pullrequestreview-430978813 right?

ilausuch:
yes, right. On my way

okurz:
Please fix typos in git commit message, e.g. s/documetation/documentation/ and s/Removed/Remove/

ilausuch:
Done

okurz:
sorry, typos still not fixed

PR #1446, opened 2020-06-15
WIP Create trigger-obs.yaml

perlpunk:
Testing a new workflow, as a simple push to the source repo doesn't seem to trigger the workflow run.


okurz:
We tried this approach but have not found a proper way where to store credentials accessible for CI only but also from forks.

PR #1447, opened 2020-06-17
Fix the issue could not find `ssh_VMwareServer` in ssh_credentials

Amrysliu:
when `$self->vmm_family eq 'vmware'`, the key defined in `ssh_credentials` is
`sshVMwareServer`, so the job failed because of could not find the key
`ssh_VMwareServer`.

The log of failed job showed that
```
Unknown ssh credentials domain ssh_VMwareServer at /usr/lib/os-autoinst/consoles/sshVirtsh.pm line 90.
```
See: http://openqa.suse.de/tests/4342402/file/autoinst-log.txt

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -348,7 +348,7 @@ sub add_disk {
               'fi;' .
               "vmkfstools -v1 -U $vmware_disk_path;" .
               "vmkfstools -v1 -c $size --diskformat thin $vmware_disk_path; sleep 10 ) 2>&1";
-            my $retval = $self->run_cmd($cmd, domain => 'ssh_VMwareServer');
+            my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer');
             die "Can't create VMware image $vmware_disk_path" if $retval;
         }
         else {


Amrysliu:
> I would be careful to just change this as I have to assume the old code was or is used somewhere and was or is working as is. Do you think this code was never used and never worked?

If I understand correctly this code never worked for several months.  we could see that since 2 months ago, the test `default_install_svirt@svirt-vmware65` failed in `bootloader_svirt` for the same reason. (http://openqa.suse.de/tests/4349060#next_previous).
@cfconrad Could you help to confirm that if this pr makes sense to the failed job?

cfconrad:
@Amrysliu very good catch, thx! Yes I confirm that this fix a regression introduced by me here https://github.com/os-autoinst/os-autoinst/pull/1166/files#diff-f2acac4570addb29b4ca9606f5de1e64L338

cfconrad:
Not good, I don't see a VR for svirt-vmware in https://github.com/os-autoinst/os-autoinst/pull/1166 :-( -- sorry I missed it back in the days.

Amrysliu:
> LGTM, thx!

@cfconrad Thanks for confirming that.

alice-suse:
Cool, thanks for the fix!

alice-suse:
> > I would be careful to just change this as I have to assume the old code was or is used somewhere and was or is working as is. Do you think this code was never used and never worked?
> 
> If I understand correctly this code never worked for several months. we could see that since 2 months ago, the test `default_install_svirt@svirt-vmware65` failed in `bootloader_svirt` for the same reason. (http://openqa.suse.de/tests/4349060#next_previous). And the tester told me that, last year this job would success, after the beginning of this year(2020), it failed.
> @cfconrad Could you help to confirm that if this pr makes sense to the failed job?

Yes, you are right. This code ever worked during sle12sp5. But fail since 15sp2.

alice-suse:
Look forward to this fix merged! @coolo

PR #1449, opened 2020-06-19
Add support for TPMv2 emulation via swtpm

AdamWill:
This adds support in os-autoinst for running qemu with a TPM
emulator device, using swtpm running on the worker host. See
https://www.qemu.org/docs/master/specs/tpm.html for general
background on this.

All this code does is run qemu with appropriate arguments if
the QEMUTPM setting is set. We use `/tmp/mytpmX/swtpm-sock` as
the socket path, where X is replaced by the worker's instance
number if QEMUTPM is set to 'instance', or by the QEMUTPM value
otherwise.

We explicitly do not handle running and shutting down swtpm
itself; I considered that design but thought it would get too
complex. Instead the admin is expected to handle that themselves,
by using a systemd unit or something similar.

I've tested this on x86_64 and it works. I haven't tested on
the other two arches yet as Fedora is moving data centres so
we're on reduced capacity, and I don't have aarch64 or ppc64le
worker hosts up yet.

### DIFF
--- a/t/18-qemu-options.t
+++ b/t/18-qemu-options.t
@@ -54,6 +54,7 @@ my $common_options = <<EOV;
    "ISO" : "$data_dir/Core-7.2.iso",
    "CDMODEL" : "ide-cd",
    "HDDMODEL" : "ide-drive",
+   "WORKER_INSTANCE": "3",
    "VERSION" : "1",
 EOV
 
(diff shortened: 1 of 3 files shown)


AdamWill:
Note, I think the aarch64 support requires qemu 5.0, which is very recent.

Here's the systemd unit I'm using for this (filename `openqa-swtpm@.service`):

    [Unit]
    Description=openQA swtpm service

    [Service]
    User=_openqa-worker
    ExecStartPre=-/usr/bin/rm -rf /tmp/mytpm%I
    ExecStartPre=/usr/bin/mkdir -p /tmp/mytpm%I
    ExecStart=/usr/bin/swtpm socket --tpm2 --tpmstate dir=/tmp/mytpm%I --ctrl type=unixio,path=/tmp/mytpm%I/swtpm-sock --log level=20
    ExecReload=/bin/true

    [Install]
    WantedBy=multi-user.target

The overall design is to designate one host per arch as the 'tpm' host, put it in the 'tpm' worker class and run as many instances of the systemd service as there are worker instances on that host, and set any jobs that use TPM emulation with that worker class.

AdamWill:
[This is my test job](https://openqa.fedoraproject.org/tests/618161) so far; the test just runs a single command that will only succeed if we have a working TPM.

AdamWill:
The ppc64le stuff I'm a little unsure of. I just cargo culted the args from https://www.qemu.org/docs/master/specs/tpm.html , but changed `scsi0` to `scsi9` as we already have a `scsi0` and may have a `scsi1`. I'm not sure if that's likely to cause any issues, and can't test until we have a ppc64le worker host up.

t/18-qemu-options.t
```diff
@@ -156,6 +157,71 @@ EOV
...
+    open(my $var, '>', 'vars.json');
+    print $var <<EOV;
+{
+$common_options
+   "QEMUTPM" : "instance"
+}
+EOV
+    close($var);
+    # call isotovideo with QEMUTPM=instance
+    system("perl $toplevel_dir/isotovideo -d qemu_disable_snapshots=1 2>&1 | tee autoinst-log.txt");
```
okurz, line 174: You know, you can just supply the additional qemutpm parameter as command line, if you prefer

AdamWill:
I was just following the style of the existing tests in that file. How do you suggest we test this without increasing the runtime?

okurz:
Well of course there can be runtime increase. I only ask to look out for *significant* increase. For example I see that the last build in master in https://travis-ci.org/github/os-autoinst/os-autoinst/builds/700671015#L1202 shows an overall runtime of 9m37s, travis CI run from your PR https://travis-ci.org/github/os-autoinst/os-autoinst/builds/700246411 shows 11m7s, that's 2 minutes and IMHO a significant increase but that can come from multiple sources. Locally for me t/18-qemu-options.t takes in master 16s. I now checked out your branch and there it takes 28s so more than a 200% runtime increase. I would really favor to be at a lower level again otherwise the test becomes again less useful. I think I already got it down to 6s and it's already increased again. Maybe just check some  a bit more level things in t/18-backend-qemu.t, optionally with mocking.

AdamWill:
right, sorry, I wasn't implying the request was absurd, I was just genuinely asking if you have an idea how to test this with *less* impact on the runtime :) i'll have a look at 18-qemu-options. t , thanks.

AdamWill:
OK, so I figured another option: when we *only* need to check the contents of the command line - not any output from qemu - we can set `QEMU_ONLY_EXEC`, which causes `start_qemu` to return immediately after running qemu and skip the stuff after that, which turns out to be taking most of the time. So this does that. On my system it results in the test taking 18 seconds rather than 26.

Another option here would be to move the option generation into `Qemu::Proc`, similar to `configure_blockdevs` and `configure_pflash`. That way of doing things is *much* easier to test. I'm not sure if there's any underlying logic to why those are done there but most other "turn variables into qemu args" stuff is done in `backend::qemu`, though. Is it just Historical Reasons, and if anyone had the time to break down that huge `start_qemu` into separate methods in `Qemu::Proc` it'd be welcomed?

okurz:
Nice. Thank you for improving the test runtime and also making it look cleaner.

Yes, `start_qemu` has grown a lot and that's just the natural evolution. `Qemu::Proc` was added by one person for a specific reason and that has not been done together with a complete backend revamp. So certainly it is just due to Historical Reasons that the startup code is not splitup further. Your changes just add to it and make it worse, that's how it is ;)

Anyone would be most welcome to refactor this.

AdamWill:
Thanks. I'll put refactoring it on that good old list of things to do when we get any free time :P

If the aarch64/ppc64le bits turn out to have issues when I have those worker hosts back, I'll send fixes.

ggardet:
@AdamWill I tried this on aarch64 with QEMU 5.0 but it seems to fail. 
Arguments are accepted by QEMU, but I get the following errors from EDK2:
```
PlatformPeim: TPM @ 0xC000000
Install PPI: B9A61AD0-2802-41F3-B513-9651CE6BD575
Install PPI: 7408D748-FC8C-4EE6-9288-C4BEC092A410
Loading PEIM 86D70125-BAA3-4296-A62F-602BEBBB9081
add-symbol-file /home/abuild/rpmbuild/BUILD/edk2-edk2-stable202005/Build/ArmVirtQemu-AARCH64/DEBUG_GCC5/AARCH64/MdeModulePkg/Core/DxeIplPeim/DxeIpl/DEBUG/DxeIpl.dll 0xBFEDF240
Loading PEIM at 0x000BFEDF160 EntryPoint=0x000BFEE4744 DxeIpl.efi
Install PPI: EE4E5898-3914-4259-9D6E-DC7BD79403CF
Install PPI: 1A36E4E7-FAB6-476A-8E75-695A0576FDD7
Install PPI: 0AE8CE5D-E448-4437-A8D7-EBF5F194F731
Loading PEIM BF7F2B0C-9F2F-4889-AB5C-12460022BE87
add-symbol-file /home/abuild/rpmbuild/BUILD/edk2-edk2-stable202005/Build/ArmVirtQemu-AARCH64/DEBUG_GCC5/AARCH64/OvmfPkg/Tcg/Tcg2Config/Tcg2ConfigPei/DEBUG/Tcg2ConfigPei.dll 0xBFEDC240
Loading PEIM at 0x000BFEDC160 EntryPoint=0x000BFEDDA78 Tcg2ConfigPei.efi
Tcg2ConfigPeimEntryPoint
Tcg2ConfigPeimEntryPoint: TPM2 detected
Install PPI: 7F4158D3-074D-456D-8CB2-01F9C8F79DAA
Loading PEIM A0C98B77-CBA5-4BB8-993B-4AF6CE33ECE4
add-symbol-file /home/abuild/rpmbuild/BUILD/edk2-edk2-stable202005/Build/ArmVirtQemu-AARCH64/DEBUG_GCC5/AARCH64/SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei/DEBUG/Tcg2Pei.dll 0xBFECF240
Loading PEIM at 0x000BFECF160 EntryPoint=0x000BFED88C4 Tcg2Pei.efi
TPM2: TPM_ST_RSP error - C4
TPM2 error! Build Hob
Install PPI: A030D115-54DD-447B-9064-F206883D7CCC
TPM2 error!
```

And Linux kernel does not find it apparently: 
```
[    3.234688] ima: No TPM chip found, activating TPM-bypass!
```

Maybe there is something missing in my setup?

AdamWill:
Did you have swtpm running with a socket at the expected path? Did you pass `--tpm2` to swtpm?

ggardet:
Started with the right user (`_openqa-worker`) it seems to be ok now. At least, EDK2 traces are fine. 
I still need to check in Linux kernel with a proper test, because kernel traces are not very verbose.

PR #1450, opened 2020-06-28
Fix the issue that does not show softfail when needle file with boo

Amrysliu:
See: https://progress.opensuse.org/issues/68413


Amrysliu:
Verify run: http://10.67.19.103/tests/142#step/welcome/3

Without this change, the result is shown like this: http://10.67.19.103/tests/140#

PR #1451, opened 2020-06-29
Workaround for cv::norm() returning NaN (poo#68474)

nadvornik:
https://progress.opensuse.org/issues/68474


nadvornik:
Updated as requested.

PR #1453, opened 2020-06-30
Add auto-merge same as in openQA based on single-approval

okurz:


.mergify.yml
```diff
@@ -0,0 +1,18 @@
+pull_request_rules:
+  - name: automatic merge when GitHub branch protection passes + explicit two approvals
+    conditions:
+      - "#approved-reviews-by>=1"
```
perlpunk, line 4: Only one required review?

okurz: yes. We shortly discussed this today already: As os-autoinst has less active reviewers we should not need more than one approval (for now).

PR #1455, opened 2020-07-02
deps: Make cpanfile targets configurable

perlpunk:
So far devel, test, cover were hardcoded in the script.

See also https://github.com/os-autoinst/openQA/pull/3173


perlpunk:
https://github.com/os-autoinst/openQA/pull/3232 was approved and merged. This PR here is doing the actual change, so if we don't merge this the change in openQA was useless ;-)

PR #1456, opened 2020-07-02
Avoid crashes due to signals being handled by OpenCV threads

Martchus:
* Create OpenCV's threads upfront
* Use the existing sigprocmask approach also within autotest (and not only the backend)
* Add a basic unit test
* See https://progress.opensuse.org/issues/53999

---

Still WIP because I could not actually reproduce the issue locally (despite the unit test). It is also not clear whether this change is actually sufficient (e.g. OpenCV might spawn more threads in other situations I'm currently not aware of).

Note that this does not cover the OpenBLAS threads observed in autotest crashes yet.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -27,6 +27,7 @@ use Socket;
 use IO::Handle;
 use POSIX '_exit';
 use cv;
+use signalblocker;
 use Scalar::Util 'blessed';
 use Mojo::IOLoop::ReadWriteProcess 'process';
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
@@ -265,8 +266,11 @@ sub start_process {
             $SIG{HUP}  = 'DEFAULT';
             $SIG{CHLD} = 'DEFAULT';
 
+            my $signal_blocker = signalblocker->new;
             cv::init;
             require tinycv;
+            tinycv::create_threads();
+            undef $signal_blocker;
 
             $0 = "$0: autotest";
             my $line = <$isotovideo>;
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -164,7 +164,15 @@ sub _dbus_call {
 
 sub do_stop_vm {
     my $self = shift;
-    $self->{proc}->save_state();
+
+    my $proc = $self->{proc};
+    if ($bmwqemu::vars{QEMU_WAIT_FINISH}) {
+        # wait until QEMU finishes on its own; used in t/18-qemu-options.t
+        if (my $qemu_pid = $proc->qemu_pid) {
+            waitpid $qemu_pid, 0;
+        }
+    }
+    $proc->save_state;
     $self->stop_qemu;
 }
 
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -998,7 +998,7 @@ sub start_qemu {
 
     $self->create_virtio_console_fifo();
     my $qemu_pipe = $self->{qemupipe} = $self->{proc}->exec_qemu();
-    return bmwqemu::fctinfo('Returning early as requested by QEMU_ONLY_EXEC.') if $vars->{QEMU_ONLY_EXEC};
+    return bmwqemu::fctinfo('Not connecting to QEMU as requested by QEMU_ONLY_EXEC') if $vars->{QEMU_ONLY_EXEC};
     $self->{qmpsocket} = $self->{proc}->connect_qmp();
     my $init = myjsonrpc::read_json($self->{qmpsocket});
     my $hash = $self->handle_qmp_command({execute => 'qmp_capabilities'});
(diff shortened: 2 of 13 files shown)


ppmclibs/tinycv_impl.cc
```diff
@@ -316,6 +316,24 @@ double getPSNR(const Mat& I1, const Mat& I2)
...
+ * - This function does a color space conversion to provoke the thread creation because there
+ *   seems to be no designated public API for that.
+ * - Apparently the image size must exceed a certain threshold to trigger multithreading. Hence
+ *   this function does not use an 1x1 px image.
+ */
+void create_opencv_threads()
+{
+    const Mat input = Mat::zeros(400, 400, CV_8UC3);
+    Mat output = Mat::zeros(400, 400, CV_8UC3);
+    cvtColor(input, output, cv::COLOR_BGR2GRAY);
```
Martchus, line 334: Seems like this is not good enough. It creates 7 additional threads on my machine with 8 cores but on a production-like machine with 32 cores no further threads are created. Not sure how to create the OpenCV threads upfront in a reliable way.

Martchus: Updated the PR. It seems `parallel_for_` is actually a public API. The new `create_opencv_threads` function now works reliably with any number of threads.

okurz:
looks like a good direction :)

ppmclibs/tinycv_impl.cc
kalikiana: Just following the code I appreciate the extensive documentation you added. Pretty neat trick to drive up the thread count like that.

autotest.pm
```diff
@@ -265,8 +266,11 @@ sub start_process {
             $SIG{HUP}  = 'DEFAULT';
             $SIG{CHLD} = 'DEFAULT';
 
+            my $signal_blocker = signalblocker->new;
```
kalikiana, line 269: Wasn't there a pattern to wrap this in a block instead of the explicit `undef`? Which would also ensure it's not used afterwards :thinking:

Martchus:
I've refactored `t/18-qemu-options.t` because it fails within the CI (not locally). I hope the refactored version will reveal what the problem is.

Martchus:
I could get rid of some of the errors in that test but still not the one which only happens in the CI and causes the tests to fail. It seems that within the CI QEMU can not be started or its output is at least not reaching the logfile.

The error `# Can't open 'serial0' for reading: 'No such file or directory' at (eval 187)[/usr/lib/perl5/5.26.1/Fatal.pm:1653] line 154` is also still present but I see it also locally where the test passes.

Martchus:
This problematic test really just relies on the outcome of a race condition when `QEMU_ONLY_EXEC` is used:

```
diff --git a/backend/qemu.pm b/backend/qemu.pm
index 9b7a84d3a..c85a23843 100644
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -164,6 +164,8 @@ sub _dbus_call {
 
 sub do_stop_vm {
     my $self = shift;
+    bmwqemu::diag("$$: do_stop_vm: " . ($bmwqemu::vars{QEMU_ONLY_EXEC} // '?'));
+    sleep 10 if $bmwqemu::vars{QEMU_ONLY_EXEC};
     $self->{proc}->save_state();
     $self->stop_qemu;
 }
```

The `QEMU_ONLY_EXEC` option is really problematic because establishing the connection to QEMU is usually the "synchronization point" to ensure that QEMU is running before starting any testing. Let's see whether I can improve the test setup. It has nothing to do anymore with the initial issue I want to fix but it is likely still a good idea to stabilize the tests.

backend/qemu.pm
```diff
@@ -164,7 +164,15 @@ sub _dbus_call {
 
 sub do_stop_vm {
     my $self = shift;
-    $self->{proc}->save_state();
+
+    my $proc = $self->{proc};
+    if ($bmwqemu::vars{QEMU_WAIT_FINISH}) {
+        # wait until QEMU finishes on its own; used in t/18-qemu-options.t
+        if (my $qemu_pid = $proc->qemu_pid) {
+            waitpid $qemu_pid, 0;
```
kalikiana, line 172: Just a thought... how about using this in production, too, to avoid stale qemu processes? Assuming that not waiting can be a cause of that.

Martchus: That makes no sense. If the shutdown is not triggered within the SUT (which we can not generally assume¹) QEMU would continue to run forever and we would wait for it forever. We must signal QEMU to stop. That is done via the stop function provided by ReadWriteProcess which will use `waitpid` internally. Here it is only useful beause QEMU is started via `-version` in that unit test and therefore expected to terminate itself.

¹ The simplest case is when we forcefully end the test execution in the middle after an error occurred.

Martchus:
I'd like to merge this but it would be nice to have at least one review.

Note that a lot of these changes actually just stabilize a unit test. The change to create threads upfront isn't actually that big of a deal and works reliable in my unit test and also when running real tests. The code within the new signalblocker module was actually there before. I've just moved it to that module, structured it to use the destructor and applied it to autotest as well (because it also loads the tinycv module). You can also start the unit test with a parameter to *not* use the signalblocker. Then it fails as expected. I have also tested it with various thread counts.

okurz:
I agree that overall this is a nice change. The only thing I dislike is the typo in the git commit message "Stabialize" – this will now be saved in history *forever* :D

okurz:
Introducing signalblocker.pm messed up openQA+os-autoinst completely, 100% incompletes on https://openqa.opensuse.org/tests/ as well as OBS failures. This is another instance of https://progress.opensuse.org/issues/54557 that should be fixed by https://github.com/os-autoinst/os-autoinst/pull/1464

PR #1457, opened 2020-07-03
Update Perl::Tidy to 20200619

Martchus:


okurz:
well, premature merge. We got that tested out then as well :D

okurz:
@Martchus next time if you want to prevent merge use either one of the terms "WIP", "Work in progress" or :construction: in the pull request title. Maybe the label "notready" also works.

Martchus:
I haven't thought about an immediate approval. Intuitively I didn't make it a draft or use "WIP" because I don't expect further work to be done. It would be good to know whether "notready" works.

Since people using Leap and our devel repo actually have this version and TW users can simply install that single package from `devel:languages:perl` so this is likely not a big deal.

okurz:
yes, sure. Should not be a big deal and hopefully also no deal with any next Tumbleweed snapshot. The perl tidy changes are always disruptive for other reasons still :(

PR #1459, opened 2020-07-03
Prevent calling cv::imwrite with an empty image

Martchus:
See https://progress.opensuse.org/issues/68554


Martchus:
I haven't tested because I can not reproduce the issue. I've just checked the code for calls to `cv::imwrite` which could possibly responsible for the issue.

I added a more detailed commit message.

PR #1462, opened 2020-07-06
Revert "Update Perl::Tidy to 20200619" as the tidy version is not even in openSUSE Tumbleweed yet

okurz:
Reverts os-autoinst/os-autoinst#1457


okurz:
If I understand https://travis-ci.org/github/os-autoinst/os-autoinst/builds/705397957#L800 correctly that means we can't go back to the old version because https://build.opensuse.org/project/show/devel:openQA:Leap:15.1 already has the new version

PR #1466, opened 2020-07-12
Add test for missing install specifications

okurz:
All openQA tests incomplete but neither package build nor unit tests
fail when a new file is added to os-autoinst without mentioning in
Makefile.am

See e.g. https://openqa.opensuse.org/tests/988325/file/autoinst-log.txt
failing with

```
Can't locate consoles/ssh_screen.pm in @INC (you may need to install the consoles::ssh_screen module) (@INC contains: …) at /usr/lib/os-autoinst/consoles/sshVirtshSUT.pm line 25.
BEGIN failed--compilation aborted at /usr/lib/os-autoinst/consoles/sshVirtshSUT.pm line 25.
```

caused by https://github.com/os-autoinst/os-autoinst/pull/1176

This commit adds a check that prevents a new os-autoinst package to be
installed if a new file was added but not referenced in the Makefile for
installation.

Example output:

```
for i in $((git ls-files "*.pm" 2>/dev/null || find -name "*.pm") | grep -v -E '(ppmclibs/|t/|tools/|backend/amt|backend/null|consoles/amtSol)'); do if ! grep -q $i Makefile.am; then echo "$i missing from install instructions in Makefile.am" && exit 1; fi; done
signalblocker.pm missing from install instructions in Makefile.am
```

Related progress issue: https://progress.opensuse.org/issues/54557

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -196,7 +196,10 @@ dist-hook:
 
 distclean-local: clean-local
 
-check-local:
+check-local-files-specified:
+	for i in $$((git ls-files "*.pm" 2>/dev/null || find -name "*.pm" -printf "%P\n") | grep -v -E '(ppmclibs/|t/|tools/|backend/amt|backend/null|consoles/amtSol)'); do if ! grep -q $$i Makefile.am; then echo "$$i missing from install instructions in Makefile.am" && exit 1; fi; done
+
+check-local: check-local-files-specified
 	$(srcdir)/tools/tidy --check
 	PERL5LIB=tools/lib/perlcritic:$$PERL5LIB perlcritic --gentle --include Perl::Critic::Policy::HashKeyQuote $(srcdir)
 	test x$(CHECK_DOC) = x0 || $(MAKE) check-doc


coolo:
I'm not a fan of that solution, that looks pretty hard to maintain.

How about make install DESTDIR=$tmp and then check if the files installed are all but a whitelist?

okurz:
I don't understand, how do you consider it hard to maintain? You state "check if the files installed are all but a whitelist". This sounds like you want to check if the installed files are the ones we intended to install. But the goal of the check is to ensure we do not *miss* to install any files as that causes all openQA tests to incomplete.

Regarding test-installing all files and checking them, something like:

```
tmpdir=$(mktemp -d)
make install DESTDIR=$tmpdir
installed_files=$(find $tmpdir -name '*.pm')
for i in $((git ls-files "*.pm" 2>/dev/null || find -name "*.pm" -printf "%P\n") | grep -v -E '(ppmclibs/|t/|tools/|backend/amt|backend/null|consoles/amtSol)'); do if ! echo $installed_files | grep -q $i; then echo "$i missing from installed files. Add to install instructions in Makefile.am."; fi; done
```

I consider this both inefficient due to the unnecessary install step as well as not as helpful as we should still point to Makefile.am where we need to reference the files.

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1466#event-3537389067 makes it look like mergify just merged this PR however I think this just means that the mentioned commit is part of https://github.com/os-autoinst/os-autoinst/pull/1467 which was merged after approval. However we can still discuss and I am also happy to change the approach if we find a better way.

coolo:
Parsing Makefiles is just not good - because you can still break make install even if you do mention the file.

Make install basically copies the structure from git somewhere else - so we can check what's in git, what's in $tmpdir and compare. No need to hardcode our expectations - outside of things we *expect to miss*.

Just a quick PoC:
```
find /tmp/fo/ -name \*.pm | sed -e 's,/tmp/fo/usr/local/lib/os-autoinst/,,'
backend/baseclass.pm
backend/console_proxy.pm
backend/driver.pm
backend/generalhw.pm
backend/ikvm.pm
backend/ipmi.pm
backend/qemu.pm
backend/s390x.pm
backend/svirt.pm
backend/pvm.pm
backend/spvm.pm
backend/virt.pm
consoles/console.pm
consoles/ipmiSol.pm
consoles/localXvnc.pm
consoles/remoteVnc.pm
consoles/s3270.pm
consoles/ssh_screen.pm
consoles/sshX3270.pm
consoles/sshVirtsh.pm
consoles/sshVirtshSUT.pm
consoles/sshIucvconn.pm
consoles/sshXtermIPMI.pm
consoles/sshXtermVt.pm
consoles/ttyConsole.pm
consoles/serial_screen.pm
consoles/virtio_terminal.pm
consoles/vnc_base.pm
consoles/network_console.pm
consoles/VNC.pm
OpenQA/Commands.pm
OpenQA/Exceptions.pm
OpenQA/Benchmark/Stopwatch.pm
OpenQA/Qemu/BlockDevConf.pm
OpenQA/Qemu/BlockDev.pm
OpenQA/Qemu/ControllerConf.pm
OpenQA/Qemu/DriveController.pm
OpenQA/Qemu/DriveDevice.pm
OpenQA/Qemu/DrivePath.pm
OpenQA/Qemu/MutParams.pm
OpenQA/Qemu/PFlashDevice.pm
OpenQA/Qemu/Proc.pm
OpenQA/Qemu/SnapshotConf.pm
OpenQA/Qemu/Snapshot.pm
OpenQA/Test/RunArgs.pm
OpenQA/Isotovideo/CommandHandler.pm
OpenQA/Isotovideo/NeedleDownloader.pm
OpenQA/Isotovideo/Interface.pm
OpenQA/Isotovideo/Utils.pm
autotest.pm
basetest.pm
bmwqemu.pm
testapi.pm
myjsonrpc.pm
mmapi.pm
lockapi.pm
commands.pm
distribution.pm
ocr.pm
needle.pm
osutils.pm
cv.pm
tools/lib/perlcritic/Perl/Critic/Policy/ConsistentQuoteLikeWords.pm
tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
/tmp/fo/usr/lib/perl5/site_perl/5.30.1/x86_64-linux-thread-multi/tinycv.pm
```

```
find . -name \*.pm | sed -e 's,^./,,' | sort -u > git
diff -u git /tmp/installed
coolo@f102#os-autoinst>diff -u git /tmp/installled 
--- git 2020-07-13 07:40:16.503314921 +0200
+++ /tmp/installled     2020-07-13 07:40:01.991340101 +0200
@@ -1,5 +1,4 @@
 autotest.pm
-backend/amt.pm
 backend/baseclass.pm
 backend/console_proxy.pm
 backend/driver.pm
@@ -15,7 +14,6 @@
 basetest.pm
 bmwqemu.pm
 commands.pm
-consoles/amtSol.pm
 consoles/console.pm
 consoles/ipmiSol.pm
 consoles/localXvnc.pm
@@ -61,27 +59,7 @@
 OpenQA/Qemu/Snapshot.pm
 OpenQA/Test/RunArgs.pm
 osutils.pm
-ppmclibs/blib/lib/tinycv.pm
-ppmclibs/tinycv.pm
-t/data/tests/bar/baz.pm
-t/data/tests/foo.pm
-t/data/tests/lib/testdistribution.pm
-t/data/tests/main.pm
-t/data/tests/tests/assert_screen_fail_test.pm
-t/data/tests/tests/boot.pm
-t/data/tests/tests/failing_module.pm
-t/data/tests/tests/modify_and_upload_file.pm
-t/data/tests/tests/reload_needles.pm
-t/data/tests/tests/select_console_fail_test.pm
-t/data/tests/tests/select_ssh_console_fail_test.pm
-t/data/tests/tests/shutdown.pm
-t/data/tests/tests/typing.pm
 testapi.pm
-t/fake/tests/fatal.pm
-t/fake/tests/ignore_failure.pm
-t/fake/tests/next.pm
-t/fake/tests/run_args.pm
-t/fake/tests/scheduler.pm
-t/fake/tests/start.pm
+/tmp/fo/usr/lib/perl5/site_perl/5.30.1/x86_64-linux-thread-multi/tinycv.pm
 tools/lib/perlcritic/Perl/Critic/Policy/ConsistentQuoteLikeWords.pm
 tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
```

(this is an older git checkout but we do see that I have a backend in my git that isn't installed).

okurz:
Right, looks comparable to what I suggested in https://github.com/os-autoinst/os-autoinst/pull/1466#issuecomment-657256977 . Created https://github.com/os-autoinst/os-autoinst/pull/1469

PR #1467, opened 2020-07-12
Add yamllint checks

okurz:


okurz:
hm, you mean call the install target and then try to list all installed files and compare the lists?

Martchus:
> hm, you mean call the install target and then try to list all installed files and compare the lists?

No. I mean call the install target and then set Perl search paths for the test execution so it only relies on the installed files. Tests would fail if a file is missing in the installation. This approach would not require any gobbing but of course  requires our unit tests to be sufficient.

PR #1471, opened 2020-07-16
Add amt+null backend to installed files

okurz:
* Install the "null" backend as well
* Add "amt" backend and console to installed files

### DIFF
--- a/tools/check-installed-files
+++ b/tools/check-installed-files
@@ -3,7 +3,7 @@ destdir="${destdir:-$(mktemp -d)}"
 make install DESTDIR="$destdir"
 installed_files=$(find "$destdir" -name '*.pm')
 all_source_files=$(git ls-files "*.pm" 2>/dev/null || find "$(dirname "$0")" -name "*.pm" -printf "%P\n")
-source_files=$(echo "$all_source_files" | grep -v -E '(ppmclibs/|t/|tools/|backend/null)')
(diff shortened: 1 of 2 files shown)


tools/check-installed-files
```diff
@@ -3,7 +3,7 @@ destdir="${destdir:-$(mktemp -d)}"
 make install DESTDIR="$destdir"
 installed_files=$(find "$destdir" -name '*.pm')
 all_source_files=$(git ls-files "*.pm" 2>/dev/null || find "$(dirname "$0")" -name "*.pm" -printf "%P\n")
-source_files=$(echo "$all_source_files" | grep -v -E '(ppmclibs/|t/|tools/|backend/amt|backend/null|consoles/amtSol)')
+source_files=$(echo "$all_source_files" | grep -v -E '(ppmclibs/|t/|tools/)')
```
kalikiana, line 6: So this even *tests* that we don't install these backends? :thinking:

Martchus: He's only added this test recently. (See his message in Rocket Chat.)

okurz: Yes, it tested that all files except this "blacklist" were installed as a consequence of the last major "all openQA instances product only incompletes" havoc that you might have missed :)

kalikiana: Sorry if I'm getting confused by the details, it just looks a bit odd that we intentionally don't install it, backed up by tests, only to ask why we don't install it ;-)

PR #1474, opened 2020-07-16
Slightly simplify code of all backends

okurz:


backend/driver.pm
```diff
@@ -156,14 +154,13 @@ sub _send_json {
...
-        return;
-    }
+    return undef if defined $rsp;
+    # this might have been closed by signal handler
+    no autodie 'close';
+    close($self->{backend_process}->channel_out);
+    $self->{backend_process}->channel_out(undef);
+    $self->{backend_process}->stop;
+    return;
     return $rsp->{rsp};
```
perlpunk, line 164: This line is now unreachable

okurz: that should explain the test failures :) Updated, thx.

backend/driver.pm
```diff
@@ -156,15 +154,12 @@ sub _send_json {
...
-        $self->{backend_process}->stop;
-        return;
-    }
-    return $rsp->{rsp};
+    return $rsp->{rsp} if defined $rsp;
+    # this might have been closed by signal handler
+    no autodie 'close';
+    close($self->{backend_process}->channel_out);
+    $self->{backend_process}->channel_out(undef);
+    $self->{backend_process}->stop;
```
perlpunk, line 162: I think this needs an explicit `return` now, as currently the return value of the `stop` method would be returned.

backend/generalhw.pm
```diff
@@ -161,14 +156,13 @@ sub start_serial_grab {
...
-        die "exec failed $!";
-    }
+    return unless $self->{serialpid} == 0;
+    setpgrp 0, 0;
+    open(my $serial, '>',  $self->{serialfile});
+    open(STDOUT,     ">&", $serial);
+    open(STDERR,     ">&", $serial);
+    exec($self->get_cmd('GENERAL_HW_SOL_CMD'));
+    die "exec failed $!";
     return;
```
Martchus, line 166: This `return` is now useless.

backend/generalhw.pm
```diff
@@ -161,14 +156,13 @@ sub start_serial_grab {
...
     $self->{serialpid} = fork();
-    if ($self->{serialpid} == 0) {
-        setpgrp 0, 0;
-        open(my $serial, '>',  $self->{serialfile});
-        open(STDOUT,     ">&", $serial);
-        open(STDERR,     ">&", $serial);
-        exec($self->get_cmd('GENERAL_HW_SOL_CMD'));
-        die "exec failed $!";
-    }
+    return unless $self->{serialpid} == 0;
```
coolo, line 159: unless == looks so python :)

okurz: uh, and what would look "more perl-ish"?

backend/generalhw.pm
```diff
@@ -24,7 +24,6 @@ use autodie ':all';
 
 use base 'backend::baseclass';
 
-use bmwqemu 'diag';
```
Amrysliu, line 27: removing this line affects the using `bmwqemu::diag()` in line 84?

okurz: good catch, thx.

okurz: fixed

Amrysliu:
LGTM

PR #1476, opened 2020-07-17
Use inclusive terminology "blocklist/passlist"

okurz:
Using `git ls-files | xargs sed -i -e 's/blacklist/blocklist/g' -e
's/whitelist/passlist/g'` following suggestions from
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=49decddd39e5f6132ccd7d9fdc3d7c470b0061bb


coolo:
the poor blocks :)

PR #1479, opened 2020-07-20
Allow running tests via CMake build script

Martchus:


cmake/test-targets.cmake
```diff
@@ -0,0 +1,72 @@
+cmake_minimum_required(VERSION 3.3.0)
+
+enable_testing()
+
+# add test for install target
```
okurz, line 5: these comments sound a bit like left-over notes from your TODO-list? :)

tools/invoke-tests
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash -e
+
+# process arguments
+if [[ $1 == --coverage ]]; then
```
okurz, line 4: As you already decided for a full-blown bash script I suggest to use getopt instead of the "manual" argument parsing, e.g. see https://github.com/os-autoinst/openQA/blob/master/script/configure-web-proxy#L14 for an example to follow. This can also help with the command line as used in Makefile.am as I consider `invoke-tests prove make $(abs_builddir)` not very clear as well

tools/invoke-tests
```diff
@@ -0,0 +1,68 @@
...
+if [[ $# -lt 3 ]]; then
+    echo "usage: $0 [--coverage] prove-tool-path make-tool-path build-directory [test-file...]"
+    exit
+fi
+prove_path=$1 make_path=$2 build_dir=$3
+shift 3
+test_files=("$@")
+
+# set Perl module include path and coverage options
+export PERL5LIB="$PWD:$PWD/ppmclibs:$PWD/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
```
okurz, line 17: I assume you want this to be either the root source dir in case of autotools-make and relative to the out-of-source cmake build dir but `$PWD` would only match by chance, right?
I guess it's a safer approach to check explicitly for CMAKE_CURRENT_BINARY_DIR or so and fallback to `$(dirname "$0")/..`

tools/invoke-tests
Martchus: Yes, I should make the argument parsing a bit more sophisticated.

cmake/test-targets.cmake
Martchus: Do you want me to remove the comment because you find it useless? I like it even just for the sake of consistency (every paragraph has a comment).

okurz: If you like to keep it then please rephrase to not be imperative, e.g. just "test for install target"

okurz:
See https://github.com/os-autoinst/os-autoinst/pull/1480 for my proposal to get rid of the hardcoded test list first. After that we could take a look what would still be needed from `tools/invoke-tests` and try if we can get rid of that script again.

tools/invoke-tests
okurz: I thought of CMAKE_CURRENT_BINARY_DIR because I assumed with cmake we would build ppmclibs below the build dir

okurz: I see, then it's alright

tools/invoke-tests
```diff
@@ -0,0 +1,51 @@
...
+# set variables for tests which need to invoke the make tool
+export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
+export OS_AUTOINST_MAKE_TOOL=$make_path
+
+# allow specifying the list of tests via TESTS in consistency with openQA; otherwise run all tests
+[[ ${#test_files[@]} -lt 1 ]] && test_files=($TESTS)
+[[ ${#test_files[@]} -lt 1 ]] && test_files=(-r)
+
+# invoke tests within the t directory
+cd t && "$prove_path" $PROVE_ARGS "${test_files[@]}"
```
okurz, line 51: IMHO that's overly complicated and we should be able to use pure POSIX-sh syntax and just define:

```shell
TESTS="$@"
TESTS="${TESTS:-"-r"}"
cd t && "$prove_path" $PROVE_ARGS $TESTS
```

wouldn't that work?

Martchus: It would work as long as there's no space within the path of a test file. Of course with my code it also only works for test files specified as CLI argument so I guess we can remove that and just never use spaces.

Martchus:
(I've restarted the Travis job after it succeeded but apparently could not report that.)

PR #1482, opened 2020-07-22
Use CMake build script within CI and RPM package builds

Martchus:
Related: [poo#68684](https://progress.opensuse.org/issues/68684)

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -32,7 +32,7 @@ Source0:        %{name}-%{version}.tar.xz
 %define opencv_require pkgconfig(opencv)
 %endif
 # The following line is generated from dependencies.yaml
-%define build_requires %opencv_require autoconf automake gcc-c++ libtool make perl(ExtUtils::Embed) perl(ExtUtils::MakeMaker) >= 7.12 perl(Module::CPANfile) perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
+%define build_requires %opencv_require cmake gcc-c++ ninja perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
 # The following line is generated from dependencies.yaml
 %define main_requires git-core perl(B::Deparse) perl(Carp) perl(Carp::Always) perl(Class::Accessor::Fast) perl(Config) perl(Cpanel::JSON::XS) perl(Crypt::DES) perl(Cwd) perl(Data::Dumper) perl(Digest::MD5) perl(DynaLoader) perl(English) perl(Errno) perl(Exception::Class) perl(Exporter) perl(ExtUtils::testlib) perl(Fcntl) perl(File::Basename) perl(File::Find) perl(File::Path) perl(File::Spec) perl(File::Temp) perl(File::Touch) perl(File::Which) perl(IO::Handle) perl(IO::Scalar) perl(IO::Select) perl(IO::Socket) perl(IO::Socket::INET) perl(IO::Socket::UNIX) perl(IPC::Open3) perl(IPC::Run::Debug) perl(IPC::System::Simple) perl(List::MoreUtils) perl(List::Util) perl(Mojo::IOLoop::ReadWriteProcess) >= 0.26 perl(Mojo::JSON) perl(Mojo::Log) perl(Mojo::URL) perl(Mojo::UserAgent) perl(Mojolicious) >= 8.42 perl(Mojolicious::Lite) perl(Net::DBus) perl(Net::IP) perl(Net::SNMP) perl(Net::SSH2) perl(POSIX) perl(Scalar::Util) perl(Socket) perl(Socket::MsgHdr) perl(Term::ANSIColor) perl(Thread::Queue) perl(Time::HiRes) perl(Try::Tiny) perl(XML::LibXML) perl(XML::SemanticDiff) perl(autodie) perl(base) perl(constant) perl(integer) perl(strict) perl(version) perl(warnings) perl-base
 # all requirements needed by the tests, do not require on this in the package
@@ -46,10 +46,8 @@ Source0:        %{name}-%{version}.tar.xz
 %if %{with spellcheck}
 # The following line is generated from dependencies.yaml
 %define spellcheck_requires aspell-en aspell-spell perl(Pod::Spell)
-%define make_check_args %{nil}
 %else
 %define spellcheck_requires %{nil}
-%define make_check_args CHECK_DOC=0
 %endif
 # The following line is generated from dependencies.yaml
 %define test_requires %build_requires %main_requires %spellcheck_requires perl(Benchmark) perl(Devel::Cover) perl(FindBin) perl(Pod::Coverage) perl(Test::Exception) perl(Test::Fatal) perl(Test::Mock::Time) perl(Test::MockModule) perl(Test::MockObject) perl(Test::Mojo) perl(Test::More) perl(Test::Output) perl(Test::Pod) perl(Test::Strict) perl(Test::Warnings) >= 0.029 perl(YAML::PP) procps python3-setuptools python3-yamllint qemu qemu-tools qemu-x86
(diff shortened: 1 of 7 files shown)


dist/rpm/os-autoinst.spec
```diff
@@ -130,14 +124,9 @@ ln -s ../sbin/service %{buildroot}%{_sbindir}/rcos-autoinst-openvswitch
...
 %check
-# disable code quality checks - not worth the time for package builds
-sed '/perlcritic/d' -i Makefile
-sed '/Perl::Critic/d' -i cpanfile
-sed '/tidy/d' -i Makefile
-
 # should work offline
 for p in $(cpanfile-dump); do rpm -q --whatprovides "perl($p)"; done
-make check test VERBOSE=1 %{make_check_args}
+make %{?_smp_mflags} VERBOSE=1 test-perl-testsuite test-installed-files test-doc
```
okurz, line 129: I would prefer to call a more generic target here to make sure we don't forget to cover any potential new target here and only exclude explicit sub-tests with a blocklist

tools/docker_run_ci
```diff
@@ -8,5 +8,5 @@ docker images | grep opensuse
 
 docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_ISOTOVIDEO_RUNTIME=20" \
   $CI_ENV --rm --entrypoint '' -v "$PWD":/opt $IMAGE \
-  sh -c 'cd /opt && ./tools/install-new-deps.sh && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
+  sh -c 'cd /opt && ./tools/install-new-deps.sh && cmake -DCMAKE_BUILD_TYPE=Release . && make -j4 check WITH_COVER_OPTIONS=1 && make coverage-codecov VERBOSE=1'
```
okurz, line 11: The `-j4` is hardcoded for something that can change. Also we should try to just have the need for one make target

dist/rpm/os-autoinst.spec
```diff
@@ -32,7 +32,7 @@ Source0:        %{name}-%{version}.tar.xz
 %define opencv_require pkgconfig(opencv)
 %endif
 # The following line is generated from dependencies.yaml
-%define build_requires %opencv_require autoconf automake gcc-c++ libtool make perl(ExtUtils::Embed) perl(ExtUtils::MakeMaker) >= 7.12 perl(Module::CPANfile) perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
+%define build_requires %opencv_require cmake gcc-c++ make perl(Module::CPANfile) perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
```
okurz, line 35: IMHO we should not even need make here anymore but instead ninja. Then the cmake macros further down automatically use ninja as well

dist/rpm/os-autoinst.spec
```diff
@@ -107,15 +105,11 @@ for i in 07-commands 13-osutils 14-isotovideo 18-qemu-options 18-backend-qemu 99
 done
 
 %build
-mkdir -p m4
-autoreconf -f -i
-%configure --docdir=%{_docdir}/%{name}
-make INSTALLDIRS=vendor %{?_smp_mflags}
+%cmake -DOS_AUTOINST_DOC_DIR:STRING=%{_docdir}/%{name}
```
okurz, line 108: Why is the special dir variable needed?

tools/docker_run_ci
Martchus: Not sure whether nproc is available within that context but I can try.

It was even using three make targets before so I already consider this an improvement. Of course it would be nice having to use only `coverage-codecov`. However, that target only runs the tests it needs to run in order to compute the coverage and not e.g. the spell checking or install tests. (Of course I could add a target to combine check and coverage-codecov but I don't think that makes it easier to understand what we're doing within the CI.)

dist/rpm/os-autoinst.spec
Martchus: I guess I could actually call ctest here directly (it allows specifying an exclude regex).

dist/rpm/os-autoinst.spec
Martchus: I don't know what `%{_docdir}` might possibly resolve to. In our builds for x86_64 TW it would resolve to the same what the build system would use by default anyways. However, this was explicitly specified before and I saw no reason to change that considering I don't know all possible values of `%{_docdir}` and I don't want to break things. (We can still improve this later.)

dist/rpm/os-autoinst.spec
Martchus: I could also use ninja but I haven't noticed any remarkable benefit when testing with it so far. Hence I've just stuck to what works so far. (We can improve this later. As a first step I would be happy if the build would succeed at all and refrain from additional experiments.)

okurz: Just use ninja, ok? :)

https://build.opensuse.org/package/live_build_log/devel:openQA:TestGithub:OPR-1482/os-autoinst/openSUSE_Tumbleweed/x86_64 shows some completely different problem right now:

```
[   85s] + rpm -q --whatprovides 'perl(Perl::Critic)'
[   85s] no package provides perl(Perl::Critic)
```

Martchus: If you insist. I've already seen the other problems. Any ideas regarding them?

dist/rpm/os-autoinst.spec
Martchus: I've added a special target so we don't have to bother with ctest-syntax at this level. This target internally uses an exclude-pattern.

dist/rpm/os-autoinst.spec
Martchus: I've been switching to ninja. The other problems have been resolved by removing this check which likely doesn't make much sense anymore.

dist/rpm/os-autoinst.spec
```diff
@@ -107,15 +105,12 @@ for i in 07-commands 13-osutils 14-isotovideo 18-qemu-options 18-backend-qemu 99
...
-make INSTALLDIRS=vendor %{?_smp_mflags}
+%define __builder ninja
+%cmake -DOS_AUTOINST_DOC_DIR:STRING=%{_docdir}/%{name}
+%cmake_build
 
 %install
-%make_install INSTALLDIRS=vendor
-# only internal stuff
-rm -r %{buildroot}/usr/lib/os-autoinst/tools/
+%cmake_install install-openvswitch
```
okurz, line 113: Can you explain that in an in-file comment why not just `%cmake_install`?

tools/docker_run_ci
Martchus: I've switched to ninja which executes a reasonable number of jobs in parallel by default.

tools/docker_run_ci
```diff
@@ -6,7 +6,11 @@ IMAGE=registry.opensuse.org/devel/openqa/containers/os-autoinst_dev
 docker pull $IMAGE
 docker images | grep opensuse
 
-docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_ISOTOVIDEO_RUNTIME=20" \
-  $CI_ENV --rm --entrypoint '' -v "$PWD":/opt $IMAGE \
-  sh -c 'cd /opt && ./tools/install-new-deps.sh && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
-
+echo docker run --env "QEMU_QMP_CONNECT_ATTEMPTS=10" --env "EXPECTED_ISOTOVIDEO_RUNTIME=20" \
```
okurz, line 9: Echo just for testing?

Martchus: Good catch. It was just for testing.

dist/rpm/os-autoinst.spec
Martchus: Shouldn't it be clear that we want to install additional, openvswitch-related files here? (I've excluded these files from the regular install target because they are apparently not considered generally useful. At least I suppose so from the fact that we also ship them in the separate `os-autoinst-openvswitch` package.)

okurz: I don't understand. We did not need to explicitly need to mention this special target in before. What happens if we just do `%cmake_install`?

Martchus: That macro adds the target `install` to the make/ninja invocation. That target will install everything which is so far contained by the `os-autoinst` package. The target `install-openvswitch` will only install everything which is so far contained by the `os-autoinst-openvswitch` package. Since we're creating both packages here at once I'm using `install-openvswitch` here (in addition to `install` which comes from the macro).

tools/docker_run_ci
```diff
@@ -7,6 +7,10 @@ docker pull $IMAGE
...
-  $CI_ENV --rm --entrypoint '' -v "$PWD":/opt $IMAGE \
-  sh -c 'cd /opt && ./tools/install-new-deps.sh && ./autogen.sh && make && make check coverage-codecov VERBOSE=1'
-
+  $CI_ENV --rm --entrypoint '' -v "$PWD":/opt $IMAGE sh -ec '
+cd /opt
+./tools/install-new-deps.sh
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Release .
+export WITH_COVER_OPTIONS=1
+ninja -v check
+ninja -v coverage-codecov'
```
kalikiana, line 16: Thank you for making this multi-line and more readable!

Martchus:
This broke the package, see https://github.com/os-autoinst/os-autoinst/pull/1484 for the fix.

PR #1483, opened 2020-07-23
Avoid updating last_good if there's no possible user of it

Vogtinator:
Taking a snapshot is really expensive, so it should be avoided.
If there are two milestone tests and the latter is fatal, a failure causes
the entire run to fail instead of rolling back, so last_good won't be used.
The last test does not need to update last_good either.

Before: http://10.160.67.86/tests/740 -> 24:00 minutes
After: http://10.160.67.86/tests/739 -> 22:32 minutes

In theory, more patterns could be optimized by just going through the test list until the next milestone and only making a snapshot if there's a non-fatal or always_rollback one, but that would break if some test calls `loadtest`, which is apparently supported...

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -346,7 +346,8 @@ sub runalltests {
 
     write_test_order();
 
-    for my $t (@testorder) {
+    for my $testindex (0 .. $#testorder) {
+        my $t        = $testorder[$testindex];
         my $flags    = $t->test_flags();
         my $fullname = $t->{fullname};
 
@@ -404,7 +405,13 @@ sub runalltests {
                 load_snapshot('lastgood');
                 $last_milestone->rollback_activated_consoles();
             }
-            if ($snapshots_supported && ($flags->{milestone} || $bmwqemu::vars{TESTDEBUG})) {
+            my $makesnapshot = $bmwqemu::vars{TESTDEBUG};
+            # Only make a snapshot if there is a next test and it's not a fatal milestone
+            if ($testindex ne $#testorder) {
+                my $nexttestflags = $testorder[$testindex + 1]->test_flags();
+                $makesnapshot ||= $flags->{milestone} && !($nexttestflags->{milestone} && $nexttestflags->{fatal});
+            }
+            if ($snapshots_supported && $makesnapshot) {
                 make_snapshot('lastgood');
                 $last_milestone         = $t;
                 $last_milestone_console = $selected_console;
--- a/autotest.pm
+++ b/autotest.pm
@@ -361,54 +361,54 @@ sub runalltests {
             }
             $vmloaded = 1;
         }
-        if ($vmloaded) {
-            my $name = $t->{name};
-            bmwqemu::modstart "starting $name $t->{script}";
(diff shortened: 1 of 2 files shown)


Vogtinator:
> This is great. Would be cool if you could add some corresponding test code in t/*autotest.t accordingly which shouldn't be too hard as this is rather "pure logic" :)

Trying, but I need some help.
I have to schedule tests with different flags, which is currently not done anywhere.
I'd like to do `loadtest "fatal"; loadtest "fatal_milestone"; loadtest "next"`, but that requires that I can reset the list of tests before and after. How can I do that?

> Also please no contractions in the git commit message subject

Will fix.

Vogtinator:
I improvised and test the case where all tests are fatal milestones now, might be enough?

autotest.pm
```diff
@@ -361,54 +362,60 @@ sub runalltests {
...
-            bmwqemu::diag "skipping $fullname";
-            $t->skip_if_not_running();
-            $t->save_test_result();
+            if (!$flags->{no_rollback} && $last_milestone && $flags->{always_rollback}) {
+                load_snapshot('lastgood');
+                $last_milestone->rollback_activated_consoles();
+            }
+            my $makesnapshot = $bmwqemu::vars{TESTDEBUG};
+            # Only make a snapshot if there is a next test and it's not a fatal milestone
+            if ($testindex ne $#testorder) {
```
okurz, line 410: http://10.160.67.86/tests/723/file/autoinst-log.txt says

```
Can't call method "test_flags" on an undefined value at /usr/lib/os-autoinst/autotest.pm line 410.
```

Not sure if that's really the same line "410" here but I have not seen this error in before so I assume your changes caused this. That should be fixed.

Vogtinator: It's fixed after I changed the condition, but I didn't update the link to the verification runs.

Vogtinator: I don't have one for after the autotest addition yet, will add once they've finished.

Vogtinator: Done

okurz:
Seems like we have a regression from this change: https://progress.opensuse.org/issues/69432 . I will prepare a revert and then hopefully we can bring back this change and have that problem fixed as well

autotest.pm
```diff
@@ -346,7 +346,8 @@ sub runalltests {
 
     write_test_order();
 
-    for my $t (@testorder) {
+    for my $testindex (0 .. $#testorder) {
+        my $t        = $testorder[$testindex];
```
mdoucha, line 350: Here's the change that broke run-time scheduling. If new items get added to `@testorder`, the original version will cycle through them while the index sequence will ignore them.

richiejp:
Just FYI, related: https://progress.opensuse.org/issues/40538,

okurz:
@richiejp are you sure https://progress.opensuse.org/issues/40538 "Reset/Clear guest RAM when it reboots in QEMU to reduce RAM snapshot size" is related?

richiejp:
@okurz They are both related to snapshot performance.

PR #1484, opened 2020-07-24
CMake: Install isotovideo as executable

Martchus:
* Covers os-autoinst-openvswitch as well
* Since these scripts are not an executable target they are apparently
  not installed with execution permissions by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,7 +165,12 @@ add_custom_target(generate-install-versions ALL
 )
 
 # install non-C++ files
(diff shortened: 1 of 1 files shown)


Martchus:
This problem could have been observed by our test package builds:

```
[   90s] + isotovideo --help
[   90s] /var/tmp/rpm-tmp.0G3eZM: line 34: /usr/bin/isotovideo: Permission denied
```

That's actually where I've discovered it. Unfortunately our "OBS Package Build" check doesn't include those packages.

PR #1486, opened 2020-07-25
os-autoinst-openvswitch: Fix spurious network startup race-conditions (2nd)

okurz:
Similar as in 05a44404 we need to wait further until also an IP is ready
on the device as the original problem is still present. It looks like
just waiting for the bridge device is not enough to ensure an IP is
configured so we should wait for that explicitly as well.

Tested manually on openqaworker12 with

```
$ ./os-autoinst-openvswitch
Waiting for bridge 'br0' to be created and configured...
Waiting for bridge 'br0' to be created and configured...
…
```

and

```
env OS_AUTOINST_USE_BRIDGE=br1 ./os-autoinst-openvswitch
```

Related progress issue: https://progress.opensuse.org/issues/59300

### DIFF
--- a/os-autoinst-openvswitch
+++ b/os-autoinst-openvswitch
@@ -33,10 +33,14 @@ sub init_switch {
     }
     system('ovs-vsctl', 'br-exists', $self->{BRIDGE});
 
-    my $bridge_conf = `ip addr show $self->{BRIDGE}`;
-
-    $self->{MAC} = $1 if $bridge_conf =~ /ether\s+(([0-9a-f]{2}:){5}[0-9a-f]{2})\s/;
-    $self->{IP}  = $1 if $bridge_conf =~ /inet\s+(([0-9]+.){3}[0-9]+\/[0-9]+)\s/;
+    for (my $timeout = 60; $timeout > 0; --$timeout) {
+        my $bridge_conf = `ip addr show $self->{BRIDGE}`;
+        $self->{MAC} = $1 if $bridge_conf =~ /ether\s+(([0-9a-f]{2}:){5}[0-9a-f]{2})\s/;
+        $self->{IP}  = $1 if $bridge_conf =~ /inet\s+(([0-9]+.){3}[0-9]+\/[0-9]+)\s/;
+        last if $self->{IP};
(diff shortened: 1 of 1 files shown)


os-autoinst-openvswitch
```diff
@@ -33,10 +33,17 @@ sub init_switch {
...
-
-    $self->{MAC} = $1 if $bridge_conf =~ /ether\s+(([0-9a-f]{2}:){5}[0-9a-f]{2})\s/;
-    $self->{IP}  = $1 if $bridge_conf =~ /inet\s+(([0-9]+.){3}[0-9]+\/[0-9]+)\s/;
+    my $timeout = 60;
+    do {
+        my $bridge_conf = `ip addr show $self->{BRIDGE}`;
+
+        $self->{MAC} = $1 if $bridge_conf =~ /ether\s+(([0-9a-f]{2}:){5}[0-9a-f]{2})\s/;
+        $self->{IP}  = $1 if $bridge_conf =~ /inet\s+(([0-9]+.){3}[0-9]+\/[0-9]+)\s/;
+        last if $self->{IP};
```
Amrysliu, line 42: I am not sure how to break this loop. I wrote a familiar Perl script file using `do {  last } until ()`, and when I execute it, report `Can't "last" outside a loop block at ... line`. @perlpunk could you have a look at it? I am not very familiar with `do { } until `. Thanks.

okurz: you are right. Can't say how I missed this as I was sure I tested this code :D

perlpunk: I actually completely avoid `do ... until` in perl.
But I have to admit that I don't understand why the `last` gives a warning, but it still seems to do the trick.

Doubling the `{}` will remove the warning:
```
do {{
  ...
  last if ...;
}} until ...;
```
But i'd still suggest we rather use a while loop here.

okurz: Well, used an explicit for-loop which I hope you are also ok with :)

okurz:
I tested manually as visible in the description above. An automatic test is always possible but not always worth the effort to do when just introducing little things that only really matter in special systems where one needs to verify anyway. We could have a system level test maybe based on   VM or containers as well as unit tests for functions that we extract and test individually together with mocking

PR #1487, opened 2020-07-27
Improve robustness of IPMI

grisu48:
The IMPI backend occasionally fails due to self-test/reboot of the IMPI controller itself.
This commit adds the functionality of retrying failing IPMI commands to address those issues and increase the robustness of the IMPI backend.

Related issue: https://progress.opensuse.org/issues/57875

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -40,17 +40,27 @@ sub ipmi_cmdline {
 }
 
 sub ipmitool {
-    my ($self, $cmd) = @_;
+    my ($self, $cmd, %args) = @_;
+    $args{tries} //= 1;
 
     my @cmd = $self->ipmi_cmdline();
     push(@cmd, split(/ /, $cmd));
 
     my ($stdin, $stdout, $stderr, $ret);
-    $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
+    # Mitigate occasional failures of impitool due to self-tests/self-reboots of
+    # the IPMI controller by a simple sleep/retry mechanism.
+    my @tries = (1 .. $args{tries});
+    for (@tries) {
+        $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
+        if ($ret == 0) {
+            $self->dell_sleep;
+            last;
+        } else {
+            sleep 4;
+        }
+    }
     chomp $stdout;
     chomp $stderr;
-
-    $self->dell_sleep;
 
     die join(' ', @cmd) . ": $stderr" unless ($ret);
     bmwqemu::diag("IPMI: $stdout");
@@ -67,12 +77,12 @@ sub dell_sleep {
 sub restart_host {
     my ($self) = @_;
 
-    my $stdout = $self->ipmitool('chassis power status');
+    my $stdout = $self->ipmitool('chassis power status', tries => 3);
     if ($stdout !~ m/is off/) {
         $self->ipmitool("chassis power off");
         while (1) {
             sleep(3);
-            my $stdout = $self->ipmitool('chassis power status');
+            my $stdout = $self->ipmitool('chassis power status', tries => 3);
             last if $stdout =~ m/is off/;
             $self->ipmitool('chassis power off');
         }
@@ -81,7 +91,7 @@ sub restart_host {
     $self->ipmitool("chassis power on");
     while (1) {
         sleep(3);
-        my $ret = $self->ipmitool('chassis power status');
+        my $ret = $self->ipmitool('chassis power status', tries => 3);
         last if $ret =~ m/is on/;
         $self->ipmitool('chassis power on');
     }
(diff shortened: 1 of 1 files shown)


okurz:
I guess you mean IPMI not IMPI :)

backend/ipmi.pm
```diff
@@ -46,11 +46,18 @@ sub ipmitool {
     push(@cmd, split(/ /, $cmd));
 
     my ($stdin, $stdout, $stderr, $ret);
-    $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
-    chomp $stdout;
-    chomp $stderr;
-
-    $self->dell_sleep;
+    # Prevent failure due to occational self-tests/self-reboots of the IPMI controller
```
okurz, line 49: typo "occational" here as well

backend/ipmi.pm
```diff
@@ -46,12 +46,19 @@ sub ipmitool {
     push(@cmd, split(/ /, $cmd));
 
     my ($stdin, $stdout, $stderr, $ret);
-    $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
+    # Prevent failure due to occational self-tests/self-reboots of the IPMI controller
+    for (my $tries = 2; $tries >= 0; $tries--) {
```
okurz, line 50: IMHO this can be written in simpler perl-style: `for (0 .. $tries)` as you do not use the variable within the loop. However I think there should be a debug output line reporting about the retry.

grisu48:
> Can you add more details under which circumstances this happens? There had been tons of IPMI tests and I would be careful to say that a retry is needed. Also, I am not sure a retry is actually safe to do for all commands.

https://progress.opensuse.org/issues/57875 has more details: `ipmitool` fails, if the IMPI controller is performing some self-test or rebooting while querying the `status` info.

    # ipmitool -I lanplus chassis power status; echo $?
    Chassis Power is off
    0
    # ipmitool -I lanplus chassis power status; echo $?
    Unable to get Chassis Power Status

It's still work in progress, indeed in the end we need to retry only on subcommands that are safe and not on commands like reboot ecc.

grisu48:
The number of retries is now a new argument in the `ipmitool` routine. The default is 1, thus no retries. Between retries there is a 4 second delay.

I've also updated the `chassis power status` requests to use this new feature (3 times).
This is according to the poo and the only `ipmitool` command, where retrying should be safe.

backend/ipmi.pm
```diff
@@ -40,18 +40,28 @@ sub ipmi_cmdline {
 }
 
 sub ipmitool {
-    my ($self, $cmd) = @_;
+    my ($self, $cmd, $tries) = @_;
+    $tries //= 1;
```
okurz, line 44: As the meaning of the third argument is not quite clear when it is just a number I suggest to pass a hash value instead, e.g. do:

```perl
my ($self, $cmd, %args) = @_;
$args{tries} //= 1;
...
for (1 .. $args{tries}) { ...
```

and call like

```perl
my $stdout = $self->ipmitool('chassis power status', tries => 3);
```

okurz:
The progress ticket is a good reference. I suggest to also mention it in the git commit message for further reference in the future.

A further suggestion: As the ticket https://progress.opensuse.org/issues/57875 is already 11 months old the situation might have changed. Have you actually checked the stability and occurence of problems that you want to cover with retries? Might be that the situation changed and either no or less retries are needed or even more by now or maybe longer back-off time between retries.

backend/ipmi.pm
grisu48: Very good point :+1:

grisu48:
> A further suggestion: As the ticket https://progress.opensuse.org/issues/57875 is already 11 months old the situation might have changed. Have you actually checked the stability and occurence of problems that you want to cover with retries? Might be that the situation changed and either no or less retries are needed or even more by now or maybe longer back-off time between retries.

We saw similar issues as well in our latest virtualization runs, so I guess it's safe to assume the issue is still present.

Given the complexity of an adequate test case and the rather defensive/mild changes of this commit I would have a look in the upcoming runs, if this already solves most of our problems. This of course might or might not solve the original issue coming from Jan - But without more information it's difficult to accurately assess the problem.

Anyhow, I think this is a good start to increase the robustness of the ipmi backend and we can do the detail configuration (like longer or configurable back-off times, number of retries ecc.) when we have problem at hand on which we can test. This commit gives us the ability for fine-tuning in the meantime.

okurz:
Yes, I agree with all. You ended up with a clean, configurable and safe implementation, appreciated

grisu48:
Thanks for the input Oliver, much appreciated :+1:

baierjan:
Just to clarify, the hardware we use is still the same (even after 11 months) and the firmware will never be updated, so yes the problem still persists. The retry on status is in my opinion an elegant solution.

okurz:
I know, my question was aiming for other software related changes that might have changed the behavior

PR #1488, opened 2020-07-28
Fix SLE12 packages using autotools, re-establish compatibility with OpenCV 3

Martchus:
See https://progress.opensuse.org/issues/68684#note-9

---

I've also updated the package link for `perl-Test-Warnings`: `osc linkpac openSUSE:Factory perl-Test-Warnings devel:openQA:SLE-12`

I've tested `osc build SLE_12_SP4 x86_64 os-autoinst.spec` and `osc build openSUSE_Tumbleweed x86_64 os-autoinst.spec` locally. (SLE_12_SP4 is the oldest distro we still build packages for in our devel project.)

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -32,7 +32,11 @@ Source0:        %{name}-%{version}.tar.xz
 %define opencv_require pkgconfig(opencv)
 %endif
 # The following line is generated from dependencies.yaml
-%define build_requires %opencv_require cmake gcc-c++ ninja perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
+%define build_base_requires %opencv_require gcc-c++ perl(Pod::Html) pkg-config pkgconfig(fftw3) pkgconfig(libpng) pkgconfig(sndfile) pkgconfig(theoraenc)
+# The following line is generated from dependencies.yaml
+%define build_legacy_requires %build_base_requires autoconf automake libtool make perl(ExtUtils::Embed) perl(ExtUtils::MakeMaker) >= 6.66 perl(Module::CPANfile)
+# The following line is generated from dependencies.yaml
+%define build_requires %build_base_requires cmake ninja
 # The following line is generated from dependencies.yaml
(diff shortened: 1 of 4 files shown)


dist/rpm/os-autoinst.spec
```diff
@@ -46,14 +50,24 @@ Source0:        %{name}-%{version}.tar.xz
...
+%define test_base_requires %main_requires perl(Benchmark) perl(Devel::Cover) perl(FindBin) perl(Pod::Coverage) perl(Test::Exception) perl(Test::Fatal) perl(Test::Mock::Time) perl(Test::MockModule) perl(Test::MockObject) perl(Test::Mojo) perl(Test::More) perl(Test::Output) perl(Test::Pod) perl(Test::Strict) perl(Test::Warnings) >= 0.029 procps python3-setuptools qemu qemu-tools qemu-x86
+# The following line is generated from dependencies.yaml
+%define test_legacy_requires %build_legacy_requires %test_base_requires
+# The following line is generated from dependencies.yaml
+%define test_requires %build_requires %spellcheck_requires %test_base_requires perl(YAML::PP) python3-yamllint
 # The following line is generated from dependencies.yaml
 %define devel_requires %test_requires perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy)
+%if 0%{?suse_version} == 1315
```
okurz, line 66: where did you get this exact-match check from? I guess it should be `sle_version` according to https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto

t/27-make-update-deps.t
```diff
@@ -16,12 +16,18 @@
...
 use FindBin '$Bin';
 
 if (not -e "$Bin/../.git") {
-    pass("Skipping all tests, not in a git repository");
+    pass('Skipping all tests, not in a git repository');
+    done_testing;
+    exit;
+}
+if (!can_load(modules => {'YAML::PP' => undef})) {
+    pass('Skipping all tests, YAML::PP is not installed');
```
okurz, line 30: Not sure, why should it be a good idea to pretend that everything was fine here?

dist/rpm/os-autoinst.spec
Martchus: I've got it from https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto

t/27-make-update-deps.t
Martchus: Yes, I better just remove the file.

PR #1493, opened 2020-07-29
Provide more incomplete reasons

Martchus:
* See particular commit messages and https://progress.opensuse.org/issues/63718
* Still leaves some cases uncovered, e.g. when the command server, the backend or autotest crash

Only tested manually so far by terminating/killing certain processes.

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -389,21 +389,37 @@ sub gen_runfile {
 }
 
 sub exec_qemu {
-    my $self   = shift;
+    my ($self) = @_;
+
     my @params = $self->gen_cmdline();
     session->enable;
-    bmwqemu::diag("starting: " . join(" ", @params));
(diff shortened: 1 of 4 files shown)


okurz:
Looks good so far but could you please test especially https://github.com/os-autoinst/os-autoinst/pull/1493/commits/b494e695467472ab5d58d5848a4ff8df75d68fcc using openQA with at best one "passed" openQA job and one "failed" openQA job

Martchus:
I've already tested it locally with one "passed" and one "failed" openQA job. But to be sure it actually works I better also run a test which does an explicit shutdown and a test which does a reboot.

Martchus:
I've executed `create_hdd_textmode` locally which has `reboot_after_installation` and `shutdown` and it still worked. I've also noticed that we start QEMU with `-no-shutdown` and according to the documentation that prevents QEMU to exit on guest shutdown.

PR #1496, opened 2020-07-31
Fix warning "Subroutine OpenQA::Isotovideo::Utils::diag redefined"

okurz:
In test distributions including testapi a perl warning is triggered
about the redefinition of diag. This can be prevented by specifying
explicit module method calls instead of including the sub into the
current scope of OpenQA::Isotovideo::Utils.

Related progress issue: https://progress.opensuse.org/issues/69484

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -20,7 +20,7 @@ use Mojo::URL;
 use Exporter 'import';
 use File::Spec;
 use Cwd;
-use bmwqemu 'diag';
+use bmwqemu ();
 use Try::Tiny;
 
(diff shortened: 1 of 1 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -20,7 +20,7 @@ use Mojo::URL;
 use Exporter 'import';
 use File::Spec;
 use Cwd;
-use bmwqemu 'diag';
+use bmwqemu;
```
perlpunk, line 23: maybe it would be better to explicitly import nothing:
```
use bmwqemu ();
```
since the module exports two subs by default, which we don't use.

okurz: Thank you. This is also making it more obvious what we want to ensure here. I was thinking about a comment in front but `use bmwqemu ();` is better :)

PR #1497, opened 2020-07-31
Handle any Perl warnings within test distribution code as fatal by default

okurz:
To ensure no regressions or problems are introduced without being
explicitly informed this change will introduce fatal handling of any
Perl warning within test distribution code. This behaviour can be
disabled with the test variable 'NO_FATAL_WARNINGS'.

Related progress issue: https://progress.opensuse.org/issues/10704

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -256,6 +256,7 @@ checkout_git_repo_and_branch('NEEDLES_DIR');
 
 bmwqemu::ensure_valid_vars();
 
+$SIG{__WARN__} = sub { die "warning treated fatal (disable with 'NO_FATAL_WARNINGS'): @_" } unless $bmwqemu::vars{NO_FATAL_WARNINGS};
 # as we are about to load the test modules checkout the specified git refspec,
 # if specified, or simply store the git hash that has been used. If it is not a
 # git repo fail silently, i.e. store an empty variable
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -21,6 +21,7 @@ EXTERNAL_VIDEO_ENCODER_OUTPUT_FILE_EXTENSION;string;webm;The extension of the ou
 EXTERNAL_VIDEO_ENCODER_ADDITIONALLY;boolean;0;Whether `EXTERNAL_VIDEO_ENCODER_CMD` should only be invoked additionally to the built-in Theora encoder. This means two videos will be created and uploaded which can be useful for comparison. (Configuring an external video encoder disables the built-in one by default.)
 NOVIDEO;boolean;0;Whether the creation of the video should be disabled and also any `EXTERNAL_VIDEO_ENCODER_` variables be ignored.
 NO_DEBUG_IO;boolean;0;Disable the I/O debug output in case of needle comparison times longer than expected
+NO_FATAL_WARNINGS;boolean;0;Disable the fatal handling of Perl warnings
 OSUTILS_WAIT_ATTEMPT_INTERVAL;float;1;The interval in seconds between "attempts" in osutils, e.g. used for connections to qemu qmp backend
 SCREENSHOTINTERVAL;float;0.5;The interval in seconds at which screenshots are taken internally
 SSH_CONNECT_RETRY;integer;5;Maximum retries to connect to SSH based console targets
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -41,32 +41,6 @@ sub is_in_log {
     is(system("grep -q \"$regex\" autoinst-log.txt"), 0, $msg);
 }
 
-subtest 'error handling when loading test schedule' => sub {
-    my $base_state = path(bmwqemu::STATE_FILE);
-    subtest 'no schedule at all' => sub {
-        $base_state->remove;
-        $bmwqemu::vars{CASEDIR} = $bmwqemu::vars{PRODUCTDIR} = $dir;
-        throws_ok { load_test_schedule } qr/'SCHEDULE' not set and/, 'error logged';
-        my $state = decode_json($base_state->slurp);
-        if (is(ref $state, 'HASH', 'state file contains object')) {
-            is($state->{component}, 'tests', 'state file contains component message');
-            like($state->{msg}, qr/unable to load main\.pm/, 'state file contains error message');
-        }
-    };
-    subtest 'unable to load test module' => sub {
-        $base_state->remove;
-        my $module = 'foo/bar';
-        $bmwqemu::vars{SCHEDULE} = $module;
-        combined_like { throws_ok {
-                load_test_schedule } qr/Can't locate $module\.pm/, 'error logged' } qr/error on $module\.pm: Can't locate $module\.pm/, 'debug message logged';
-        my $state = decode_json($base_state->slurp);
-        if (is(ref $state, 'HASH', 'state file contains object')) {
-            is($state->{component}, 'tests', 'state file contains component message');
-            like($state->{msg}, qr/unable to load foo\/bar\.pm/, 'state file contains error message');
-        }
-    };
-};
-
 subtest 'standalone isotovideo without vars.json file and only command line parameters' => sub {
     chdir($pool_dir);
     unlink('vars.json') if -e 'vars.json';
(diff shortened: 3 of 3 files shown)


okurz:
CI tests will fail before https://github.com/os-autoinst/os-autoinst/pull/1496 is included which exactly shows how the feature is supposed to work.

okurz:
@os-autoinst/tests-maintainer I would like to explicitly invite for your feedback here as well.

perlpunk:
Can you explain why you use a custom signal handler and not
```
use warnings FATAL => "all";
```
for example?

With the signal handler, also intentional warnings like `warn "..."` will be fatal now.

Vogtinator:
+1 on the idea, but AFAICT the CI doesn't currently make sure that the current test code is compatible with this.
Would it be possible to e.g. scan logs on o3 and osd for missed warnings?

okurz:
> Can you explain why you use a custom signal handler and not
> 
> ```
> use warnings FATAL => "all";
> ```
> 
> for example?

Well, `use warnings FATAL => "all";` is lexically scoped so it will not cover the test distribution code which is running in a different process IIUC, right?

> With the signal handler, also intentional warnings like `warn "..."` will be fatal now.

Yes, I know. I have not yet seen intentional good uses of these when running isotovideo or openQA code. Maybe you have a different idea how to cover both cases of also covering warnings in test distribution code?

> +1 on the idea, but AFAICT the CI doesn't currently make sure that the current test code is compatible with this.
> Would it be possible to e.g. scan logs on o3 and osd for missed warnings?

I did and have created https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10764 for warnings I have found in openSUSE and SLE tests. For other perl warnings I see no feasible option then to just let openQA tests fail and handle individual problems case by case as running *all* test cases against this PR I see as unfeasible.

okurz:
I rebased the code now after https://github.com/os-autoinst/os-autoinst/pull/1496 was merged fixing CI tests.

Vogtinator:
> > +1 on the idea, but AFAICT the CI doesn't currently make sure that the current test code is compatible with this.
> > Would it be possible to e.g. scan logs on o3 and osd for missed warnings?
> 
> I did and have created [os-autoinst/os-autoinst-distri-opensuse#10764](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10764) for warnings I have found in openSUSE and SLE tests. For other perl warnings I see no feasible option then to just let openQA tests fail and handle individual problems case by case as running _all_ test cases against this PR I see as unfeasible.

Well, scanning the logs of the last month or so would at least cover most of the relevant cases.

perlpunk:
To make sure tests won't suddenly fail, the signal handler could just issue another warning that we can easily grep for.

perlpunk:
https://travis-ci.org/github/os-autoinst/os-autoinst/builds/713672681#L1288
Interesting that this issues a warning, because IO::Handle does not `use warnings`.

rwx788:
> To make sure tests won't suddenly fail, the signal handler could just issue another warning that we can easily grep for.

+1 here, I would totally support idea of failing CI in case some warnings occur, but as in the test runs we might get warnings because we call some method from perl package in deprecated way and test will start failing without explicit code change. And I agree that this should be fixed and not ignored, but we should still be able to get results for the build without setting `NO_FATAL_WARNINGS`.

okurz:
> To make sure tests won't suddenly fail, the signal handler could just issue another warning that we can easily grep for.

I don't understand: When perl warnings are already ignored by test reviewers, why should we just add one more "another warning" from a signal handler? Surely we could make that a little bit more complicated to just collect warnings and fail the test after execution in case of any warnings but still execute to the end. Would this be preferrable?

rwx788:
> > To make sure tests won't suddenly fail, the signal handler could just issue another warning that we can easily grep for.
> 
> I don't understand: When perl warnings are already ignored by test reviewers, why should we just add one more "another warning" from a signal handler? Surely we could make that a little bit more complicated to just collect warnings and fail the test after execution in case of any warnings but still execute to the end. Would this be preferrable?

They are not reviewed every day, because if the test worked, why would we expect anyone to look into the logs. By having better way to track them, we could regularly catch such warnings and act on those. This another warning would unify the way how to look for them. We definitely can come up with better way to get informed about warnings, but builds status have more business values and these are warnings in the end. Do we already have a way to look for strings in the os-autoinst logs?

perlpunk:
> Surely we could make that a little bit more complicated to just collect warnings and fail the test after execution in case of any warnings but still execute to the end. Would this be preferrable?

Something like that, yeah. At least don't make them fatal immediately. remember that a global handler also influences perl core and CPAN modules.
My suggestion was to be defensive in a first step and only report the warnings in a way that we can easily grep for.
If we don't find warnings reported that way for a month, then the code could be modified to make the tests fail.
But don't die in the signal handler directly.

asmorodskyi:
What if we will do this other way around ? Let's introduce variable `FATAL_WARNING` when you set it to the job all warnings became fatal ?

okurz:
> What if we will do this other way around ? Let's introduce variable `FATAL_WARNING` when you set it to the job all warnings became fatal ?

Yes, I wanted to avoid this as apparently the interest in fixing warnings was not high so far. I doubt it is easier to convince many to use this flag and find a good place to set it instead of just fixing variables that we find and workaround the places with NO_FATAL_WARNINGS where we can not find them.

isotovideo
```diff
@@ -256,6 +256,7 @@ checkout_git_repo_and_branch('NEEDLES_DIR');
 
 bmwqemu::ensure_valid_vars();
 
+$SIG{__WARN__} = sub { die "warning treated fatal (disable with 'NO_FATAL_WARNINGS'): @_" } unless $bmwqemu::vars{NO_FATAL_WARNINGS};
```
Martchus, line 259: It would be good to pass this as reason to openQA using `bmwqemu::serialize_state`.

I'd also like to note that this will not only affect the test code but also all code executed within isotovideo and all subsequently forked sub processes such as the command server. However, the commit message states that this change would only affect warnings within the test distribution code. So I would rather unset this after test modules have been loaded and explicitly enable it within the autotest process.

Martchus (changes requested):

isotovideo
okurz: do you think it would work if our die handler in general would do the state serialization?

Martchus: Now it actually already does that for the backend and isotovideo processes.

okurz: so is your comment already resolved then?

Martchus: Not really. It looks like this would also affect subsequently started process and not only test distribution code. So the impact of this change might be quite big. So I would rather unset this after test modules have been loaded and explicitly enable it within the autotest process.

okurz: hm, and if I strip the "within test distribution code" from the commit message to make it apply to all executed code, would it be fine? :)

Martchus: Maybe. If all our code was perfect and wouldn't produce any warnings it would definitely be fine. However, I'm not sure how "warning free" the code of os-autoinst (and of course any 3rdparty Perl module we use) in reality is. It might be that you provoke problems here which so far haven't been any. But if you want to generate some incompletes for your auto-review script to work with...

schlad: https://lore.kernel.org/linux-block/CAHk-=wiFwUUEpCQ2FDWxBFb3XSxu=+yODHGfLPncvTQa9BeP9w@mail.gmail.com/t/ 

:-)

okurz: Proving my point. We want to prevent any application to continue on any warning. Otherwise angry Linus will yell at us! And I guess nobody of us wants for that to happen ;)

Martchus: Wasn't he talking about warnings which can be detected at compile time? This type of warnings can be easily detected an addressed. We should have already a similar check.

This PR is affecting *runtime* warnings which are much harder to detect and fix in advance. Hence it is hard to say how much trouble we'll gain by merging this PR. In the best case it won't be notable change. I'd still consider it good if we find a few problems this way in our code although it means we're producing incomplete jobs until a fix is deployed. I'm just afraid that too many problems will come to the surface at once so we will have a hard time to fix these at once and explaining our users why there are now so many incomplete tests.

If you want to compare with the kernel: I suppose the kernel is actually very preservative when it comes to producing fatal errors (kernel panic) as it rarely makes sense to stop the entire system.

kraih: Treating all runtime warnings as exceptions will definitely go wrong at some point. All it takes is a deprecation in one of the modules we depend on and nothing would work anymore.

PR #1498, opened 2020-08-04
Tweak CMake build script, prevent running tests twice within CI

Martchus:
* See for the whole CMake story: https://progress.opensuse.org/issues/68684

### DIFF
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -15,25 +15,31 @@ options:
 
 # parse arguments
 prove_path=prove make_path=make build_directory=$source_directory
-opts=$(getopt --options h --longoptions help,coverage,prove-tool:,make-tool:,build-directory: --name "$(basename "$0")" -- "$@") || usage
+opts=$(getopt \
+    --options h \
+    --longoptions help,coverage,skip-if-cover-db-exists,prove-tool:,make-tool:,build-directory: \
+    --name "$(basename "$0")" -- "$@") || usage
 eval set -- "$opts"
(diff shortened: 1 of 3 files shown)


Martchus:
The CI runtime is back again at ~10 minutes (from ~20 minutes).

tools/invoke-tests
```diff
@@ -15,25 +15,31 @@ options:
...
+    --make-tool               ) make_path=$2;              shift 2 ;;
+    --build-directory         ) build_directory=$2;        shift 2 ;;
+    --                        ) shift; break ;;
+    *                         ) break ;;
   esac
 done
 [[ "$@" ]] && TESTS="$@"
 
+db="$build_directory/cover_db"
+[[ $SKIP_IF_COVER_DB_EXISTS ]] && [[ -d $db ]] && exit 0
```
okurz, line 38: This looks like you work around borked build dependency in the build system within a shell script. Just looks quite wrong but as I don't have a better idea right now I will still not reject

Martchus: The tracking within the build system basically works. That's why I didn't notice the issue locally. However, it seems that CMake always wants to invoke the command which produces a certain output (another command depends on) at least once. That seems to be even the case if that output has already been created elsewhere. Maybe I'm doing something wrong here or I'm missing an option but I couldn't find it.

Martchus:
I've been merging this now because doubling the CI runtime is quite annoying and this PR fixes it. If we can find a more elegant solution that would be nice of course.

PR #1500, opened 2020-08-05
Suppress 'unable to inform websocket…' error when command server exited

Martchus:
When the command server has already exited we would not do that http
request anyways. This change improves the case  when the command server
exits while the http request is sent (and therefore the http request
fails). Then it makes more sense to suppress the error because the
command server has likely just been stopped by the worker or the user.
A message like `commands process exited: 0` is still logged in any case.

---

Still a draft because I need to do some testing.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -171,22 +171,13 @@ sub stop_commands {
         my $timeout = 15;
         my $ua      = Mojo::UserAgent->new(request_timeout => $timeout);
         my $tx      = $ua->post($url, json => {stopping_test_execution => $reason});
-        try {
-            my $res = $tx->result;
-            if ($res->code != 200) {
-                my $error_message = $res->to_string;
-                diag('isotovideo: unable to inform websocket clients about stopping command server: ' . $error_message);
-                if ($error_message =~ qr/.*timeout.*/i) {
-                    diag("isotovideo: The command server could be under heavy load and the timeout of $timeout seconds has been exceeded. This message is most likely not be related to the MAX_JOB_TIME of the current job possibly being exceeded.");
-                }
-                else {
-                    diag('The command server might have already been stopped by the worker after the user has aborted the job or the job timeout has been exceeded.');
-                }
-            }
+        if (my $err = $tx->error) {
+            my $msg = $err->{code} ? "$err->{code} response: $err->{message}" : "Connection error: $err->{message}";
+            $msg .= "\nisotovideo: The command server could be under heavy load and the timeout of $timeout"
+              . ' seconds has been exceeded. This message is not be related to the MAX_JOB_TIME'
+              . ' of the current job possibly being exceeded.' if $err =~ qr/.*timeout.*/i;
+            diag('isotovideo: unable to inform websocket clients about stopping command server: ' . $msg) if $cmd_srv_process->is_running;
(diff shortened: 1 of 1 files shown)


Martchus:
It sort of works but it is still not race-free. Here the command server already refuses the connection although it hasn't completely terminated yet:

```
[0m[37m[2020-08-05T18:42:02.766 CEST] [debug] isotovideo received signal TERM
[33m[0m[2020-08-05T18:42:02.766 CEST] [info] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[0m[37m[2020-08-05T18:42:02.766 CEST] [debug] terminating command server 7058 because test execution ended
[0m[37m[2020-08-05T18:42:02.766 CEST] [debug] isotovideo: informing websocket clients before stopping command server: http://127.0.0.1:20013/dKbxtEMgLaD6kpkw/broadcast
[0m[37m[2020-08-05T18:42:02.767 CEST] [debug] autotest received signal TERM, saving results of current test before exiting
[0m[37m[2020-08-05T18:42:02.771 CEST] [debug] isotovideo: unable to inform websocket clients about stopping command server: Connection error: Connection refused
[0m[37m[2020-08-05T18:42:02.772 CEST] [debug] commands process exited: 0
```

isotovideo
```diff
@@ -171,22 +171,13 @@ sub stop_commands {
...
-                else {
-                    diag('The command server might have already been stopped by the worker after the user has aborted the job or the job timeout has been exceeded.');
-                }
-            }
+        if (my $err = $tx->error) {
+            my $msg = $err->{code} ? "$err->{code} response: $err->{message}" : "Connection error: $err->{message}";
+            $msg .= "\nisotovideo: The command server could be under heavy load and the timeout of $timeout"
+              . ' seconds has been exceeded. This message is not be related to the MAX_JOB_TIME'
+              . ' of the current job possibly being exceeded.' if $err =~ qr/.*timeout.*/i;
+            diag('isotovideo: unable to inform websocket clients about stopping command server: ' . $msg) if $cmd_srv_process->is_running;
```
okurz, line 179: the check `if $cmd_srv_process->is_running` is probably not enough to prevent all race conditions here, right? I was more thinking of a try…catch that would just ignore the error in case when we are already shutting down. But maybe we can adjust `$cmd_srv_process->is_running` to not return true in case of already shutting down?

Martchus: But how would isotovideo know whether the command server is currently shutting down?
But this seems all too complicated anyways. Maybe the command server can trigger this in its own signal handler. That would be way more elegant. I tried that also when implementing the developer mode initially. Maybe it is worth a second shot.

okurz: you mean just relying on signals rather than sending RPC?

Martchus:
https://github.com/os-autoinst/os-autoinst/pull/1501 has been merged instead

isotovideo
Martchus: The command server is terminated sooner or later anyways. Either it received SIGTERM/SIGINT out of our control or we stop it ourselves a few lines below. If the command server would send these web socket messages itself from its own signal handler we would not have to deal with that at all in isotovideo.

PR #1503, opened 2020-08-10
Close video encoder pipes and wait for it to finalize the video before exiting

Martchus:
See https://progress.opensuse.org/issues/68923#note-9

---

I've tested it manually. So far I haven't noticed any truncated videos anymore. (Not when the test exists normally and also not when stopped by the worker.)

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -62,6 +62,7 @@ sub new {
     $self->{serial_offset}                     = 0;
     $self->{video_frame_data}                  = [];
     $self->{video_frame_number}                = 0;
+    $self->{video_encoders}                    = {};
     $self->{external_video_encoder_image_data} = [];
     $self->{min_image_similarity}              = 10000;
     $self->{min_video_similarity}              = 10000;
@@ -325,6 +326,15 @@ sub check_select_rate {
     return 0;
 }
 
+sub _invoke_video_encoder {
+    my ($self, $pipe_name, $display_name, @cmd) = @_;
+
+    my $pid  = open($self->{$pipe_name}, '|-', @cmd);
+    my $pipe = $self->{$pipe_name};
+    $self->{video_encoders}->{$pid} = {name => $display_name, pipe => $pipe};
+    $pipe->blocking(0);
+}
+
 sub _start_external_video_encoder_if_configured {
     my ($self) = @_;
 
@@ -336,8 +346,7 @@ sub _start_external_video_encoder_if_configured {
     $cmd .= " '$output_file_path'" unless $cmd =~ s/%OUTPUT_FILE_NAME%/$output_file_path/;
 
     bmwqemu::diag "Launching external video encoder: $cmd";
-    open($self->{external_video_encoder_cmd_pipe}, '|-', $cmd);
-    $self->{external_video_encoder_cmd_pipe}->blocking(0);
+    $self->_invoke_video_encoder(external_video_encoder_cmd_pipe => 'external video encoder', $cmd);
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


backend/baseclass.pm
```diff
@@ -361,6 +369,33 @@ sub start_encoder {
...
+    my $video_encoders = delete $self->{video_encoders};
+    return undef unless defined $video_encoders && keys %$video_encoders;
+
+    # give the video encoder processes 30 seconds to finalize the video
+    # note: Closing the pipe should cause the video encoder to terminate. Not sending SIGTERM/SIGINT because the signal might be
+    #       already sent by the worker or shell and ffmpeg will not continue finalizing the video after receiving a 2nd exit signal.
+    no autodie qw(close waitpid);
+    close $video_encoders->{$_}->{pipe} for keys %$video_encoders;
+    bmwqemu::diag 'Waiting for video encoder to finalize the video';
+    for (my $timeout = 30, my $interval = 1; $timeout > 0; sleep $interval, $timeout -= $interval) {
```
okurz, line 384: Does the sleep mean we wait needlessly at least one second in every case? The for loop with a sleep isn't convincing me yet :)

Martchus: No - it only sleeps when at least one video encoder process is still running. (Although it is not used often in Perl, this is a plain-old for loop and it behaves identical to the for loop in C/C++. The for loop has the advantage that all loop variables can be defined within the loop scope.)

By the way, I didn't simply make it two consecutive non-blocking `waitpid` calls because this way we couldn't implement a timeout. However, there's further cleanup code which is supposed to run so I'd refrain from blocking here forever.

Martchus: There are actually ways to enforce a timeout more efficiently: https://stackoverflow.com/questions/282176/waitpid-equivalent-with-timeout  
However, I think it isn't worth the effort (at this point).

okurz: ok, I would prefer if you can show that the case of shutdown does not increase over what is actually necessary. Could you please provide some data from experiments to back this up?

Martchus: Yes. I've started a few tests which are very short and hence the video is also only ~2 minutes. Then the finalizing seems to be so fast that all the waiting code is basically a no-op, see the timestamps in the following logs:

```
[0m[37m[2020-08-11T13:55:40.423 CEST] [debug] isotovideo done
[0m[37m[2020-08-11T13:55:40.423 CEST] [debug] QEMU status is not 'shutdown', it is 'running'
[0m[37m[2020-08-11T13:55:40.423 CEST] [debug] BACKEND SHUTDOWN 
[0m[33m[2020-08-11T13:55:40.424 CEST] [info] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[0m[37m[2020-08-11T13:55:42.468 CEST] [debug] flushing frames
[0mframe=  248 fps=1.8 q=0.0 Lsize=     431kB time=00:00:10.29 bitrate= 343.4kbits/s speed=0.0734x    
video:429kB audio:0kB subtitle:0kB other streams:0kB global headers:0kB muxing overhead: 0.504507%
[37m[2020-08-11T13:55:47.411 CEST] [debug] Waiting for video encoder to finalize the video
[0m[37m[2020-08-11T13:55:47.412 CEST] [debug] The external video encoder (pid 19794) terminated
[0m[37m[2020-08-11T13:55:47.412 CEST] [debug] The built-in video encoder (pid 19795) terminated
```

```
[0m[37m[2020-08-11T14:02:59.982 CEST] [debug] isotovideo done
[0m[37m[2020-08-11T14:02:59.983 CEST] [debug] QEMU status is not 'shutdown', it is 'running'
[0m[37m[2020-08-11T14:02:59.983 CEST] [debug] BACKEND SHUTDOWN 
[0m[33m[2020-08-11T14:02:59.984 CEST] [info] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[0m[37m[2020-08-11T14:03:02.024 CEST] [debug] flushing frames
[0mframe=  250 fps=1.8 q=0.0 Lsize=     531kB time=00:00:10.37 bitrate= 419.3kbits/s speed=0.0741x    
video:529kB audio:0kB subtitle:0kB other streams:0kB global headers:0kB muxing overhead: 0.413080%
[37m[2020-08-11T14:03:05.479 CEST] [debug] Waiting for video encoder to finalize the video
[0m[37m[2020-08-11T14:03:05.479 CEST] [debug] The built-in video encoder (pid 21404) terminated
[0m[37m[2020-08-11T14:03:05.479 CEST] [debug] The external video encoder (pid 21403) terminated
```

okurz: I see a shutdown time of 6-7s in the both cases you referenced which I consider bad. Do you know what caused it? I was now used to 1-2s after my last termination improvements. But the time between the later messages with just 1ms is negligible of close

Martchus: I had a look at the code again to find that out. It seems like my new code now duplicates the existing "flushing frames" handling for the internal video encoder. I somehow assumed we had no special handling for the old video encoder.

I better streamline the handling for the internal video encoder with the new code. I can also investigate why the "flushing frames" part takes so long.

Martchus: I now added the commit "Pass remaining frames to external video encoder as well". But it doesn't change the problem that it takes quite a while to send the remaining data to the internal video encoder.

backend/baseclass.pm
```diff
@@ -361,6 +369,61 @@ sub start_encoder {
...
+                }
+                elsif ($fh == $external_pipe) {
+                    $self->_write_buffered_data_to_file_handle('External encoder', $video_data_for_external_encoder, $fh);
+                    $select->remove($fh) unless @$video_data_for_external_encoder;
+                }
+            }
+        }
+    }
+    catch {
+        bmwqemu::diag "Unable pass remaining frames to video encoder: $_";
```
okurz, line 403: "unable *to* pass"?

backend/baseclass.pm
```diff
@@ -361,6 +369,61 @@ sub start_encoder {
...
+    };
+
+    # give the video encoder processes time to finalize the video
+    # note: Closing the pipe should cause the video encoder to terminate. Not sending SIGTERM/SIGINT because the signal might be
+    #       already sent by the worker or shell and ffmpeg will not continue finalizing the video after receiving a 2nd exit signal.
+    no autodie qw(close waitpid);
+    close $video_encoders->{$_}->{pipe} for keys %$video_encoders;
+    bmwqemu::diag 'Waiting for video encoder to finalize the video';
+    for (my $interval = 0.5; $timeout > 0; sleep($interval), $timeout -= $interval) {
```
okurz, line 412: I would prefer a lower interval, e.g. 0.1

PR #1504, opened 2020-08-11
spec: Exclude unstable test 28-signalblocker.t

okurz:
Related progress issue: https://progress.opensuse.org/issues/69820


Martchus:
I'm actually about to create a PR for stabilizing the test. (Apparently the thread count can be reduced at the end again. I suppose I should just check whether at least one thread is still running.) So I would directly merge my PR.

PR #1506, opened 2020-08-14
Improve error handling of D-Bus calls to contain full context in any case

Martchus:
* Don't duplicate `_dbus_do_call` call for different paths of error
  handling
* Especially the function name is useful because it corresponds to one of
  the functions in the `os-autoinst-openvswitch` script.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -144,22 +144,20 @@ sub _dbus_do_call {
 sub _dbus_call {
     my ($self, $fn, @args) = @_;
     my ($rt, $message);
-    if (!$bmwqemu::vars{QEMU_NON_FATAL_DBUS_CALL}) {
-        ($rt, $message) = $self->_dbus_do_call($fn, @args);
-        die $message unless $rt == 0;
-    }
     eval {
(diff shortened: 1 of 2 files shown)


Martchus:
By the way, this came up when looking into an incomplete job with `Reason: backend died: 'tap74' is not connected to bridge 'br1' at /usr/lib/os-autoinst/backend/qemu.pm line 149.`.

okurz:
Not sure I get it. Could you please provide examples of how the error messages look different from examples now?

Martchus:
The example from my previous comment would look like this: `Reason: backend died: Open vSwitch command 'set_vlan' with arguments '???' failed: 'tap74' is not connected to bridge 'br1' at /usr/lib/os-autoinst/backend/qemu.pm line ???.`

Not sure about the exact arguments in this case and `set_vlan` is also just a guess.

Note that before the introduction of `QEMU_NON_FATAL_DBUS_CALL` the context information was written to the log. `QEMU_NON_FATAL_DBUS_CALL` made the error more apparent but lost the context information.

PR #1507, opened 2020-08-17
Improve error handling when running commands in generalhw backend

Martchus:
* Remove explicit check for the command being executable because it does
  not help in any case anyways, e.g. https://progress.opensuse.org/issues/70135
* Rather catch the exception from IPC::Run, add some context information
  and re-throw it

---

I'm wondering whether there are unit tests for this code. If not, maybe it is time to create a test for this backend - at least for the command code.


perlpunk:
OBS build should be fixed after #1509 has been merged

PR #1508, opened 2020-08-17
Add `mouse_drag` subroutine to support click and drag in tests

lruzicka:
1. This PR adds the `mouse_drag` subroutine that enables the mouse and drag
support for OpenQA tests. To use the subroutine, it is possible to either
provide a "starting" and "ending" needle from which the starting and finishing
points will be calculated (using the centres of the needle tag), or
coordinates might be provided directly. If both needles and coordinates are
provided at the same time, then the coordinates precede over the needles.
Combinations are possible, too, which means that one of the points can
be passed as a needle and the other as coordinates.

2. In order not to repete code, the part responsible for calculating the
center of needle was removed from the `click_lastmatch` subroutine and moved
into a standalone non-exported subroutine called `calculate_clickpoint` to be
also used by the `mouse_drag` subroutine, as well as by the original
`click_lastmatch`.

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -683,4 +683,32 @@ subtest 'autoinst_url' => sub {
     is(autoinst_url('foo'), 'http://localhost:1/foo', 'we can configure the hostname that autoinst_url returns');
 };
 
+subtest '_calculate_clickpoint' => sub {
+    my %fake_needle = (
+        area => [{x => 10, y => 10, w => 20, h => 30}],
+    );
+    my %fake_needle_area = (x    => 100, y    => 100, w => 50, h => 40);
+    my %fake_click_point = (xpos => 20,  ypos => 10);
+
+    # Everything is provided.
+    my ($x, $y) = testapi::_calculate_clickpoint(\%fake_needle, \%fake_needle_area, \%fake_click_point);
+    is($x, 120);
+    is($y, 110);
+
+    # Everything is provided but the click point is 'center'
+    ($x, $y) = testapi::_calculate_clickpoint(\%fake_needle, \%fake_needle_area, "center");
+    is($x, 125);
+    is($y, 120);
+
+    # Just the area is provided and no click point.
+    ($x, $y) = testapi::_calculate_clickpoint(\%fake_needle, \%fake_needle_area);
+    is($x, 125);
+    is($y, 120);
+
+    # Just the needle is provided and no area and click point.
+    ($x, $y) = testapi::_calculate_clickpoint(\%fake_needle);
+    is($x, 20);
+    is($y, 25);
+};
+
 done_testing;
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -711,4 +711,117 @@ subtest '_calculate_clickpoint' => sub {
     is($y, 25);
 };
 
+subtest 'mouse_drag' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+    my @area         = ({x => 100, y => 100, w => 20, h => 20});
+    $mock_testapi->redefine(assert_screen => {area => \@area});
+
+    my ($startx, $starty) = (0,   0);
+    my ($endx,   $endy)   = (200, 200);
+    my $button = "left";
+    $cmds = [];
+    # Startpoint from a needle. Endpoint coordinates.
+    mouse_drag(startpoint => 'foo', endx => $endx, endy => $endy, button => $button, timeout => 30);
+    is_deeply($cmds, [
+            {
+                cmd => 'backend_mouse_set',
+                x   => 110,
+                y   => 110
+            },
+            {
+                bstate => 1,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+            {
+                cmd => 'backend_mouse_set',
+                x   => 200,
+                y   => 200
+            },
+            {
+                bstate => 0,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+    ], 'mouse drag (startpoint defined by a needle)') or diag explain $cmds;
+
+    # Startpoint from coordinates, endpoint from a needle.
+    $cmds = [];
+    mouse_drag(endpoint => 'foo', startx => $startx, starty => $starty, button => $button, timeout => 30);
+    is_deeply($cmds, [
+            {
+                cmd => 'backend_mouse_set',
+                x   => 0,
+                y   => 0
+            },
+            {
+                bstate => 1,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+            {
+                cmd => 'backend_mouse_set',
+                x   => 110,
+                y   => 110
+            },
+            {
+                bstate => 0,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+    ], 'mouse drag (endpoint defined by a needle)') or diag explain $cmds;
+
+    # Using coordinates only.
+    $cmds = [];
+    mouse_drag(endx => $endx, endy => $endy, startx => $startx, starty => $starty, button => $button, timeout => 30);
+    is_deeply($cmds, [
+            {
+                cmd => 'backend_mouse_set',
+                x   => 0,
+                y   => 0
+            },
+            {
+                bstate => 1,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+            {
+                cmd => 'backend_mouse_set',
+                x   => 200,
+                y   => 200
+            },
+            {
+                bstate => 0,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+    ], 'mouse drag (start and endpoints defined by coordinates)') or diag explain $cmds;
+
+    # Both needle and coordinates provided for startpoint (coordinates should win).
+    $cmds = [];
+    mouse_drag(startpoint => "foo", endx => $endx, endy => $endy, startx => $startx, starty => $starty, button => $button, timeout => 30);
+    is_deeply($cmds, [
+            {
+                cmd => 'backend_mouse_set',
+                x   => 0,
+                y   => 0
+            },
+            {
+                bstate => 1,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+            {
+                cmd => 'backend_mouse_set',
+                x   => 200,
+                y   => 200
+            },
+            {
+                bstate => 0,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+    ], 'mouse drag (redundant definiton by a needle)') or diag explain $cmds;
+};
+
 done_testing;
--- a/testapi.pm
+++ b/testapi.pm
@@ -1615,7 +1615,6 @@ sub mouse_drag {
     }
     # Get the button variable. If no button has been provided, assume the "left" button.
     my $button = $args{button} // "left";
-    bmwqemu::log_call("mouse dragged", button => $button);
 
     # Now, perform the actual mouse drag. Navigate to the startpoint location,
     # press and hold the mouse button, then navigate to the endpoint location
@@ -1624,6 +1623,7 @@ sub mouse_drag {
     query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
     mouse_set($endx, $endy);
     query_isotovideo('backend_mouse_button', {button => $button, bstate => 0});
+    bmwqemu::log_call(message => "Mouse dragged from $startx,$starty to $endx, $endy", button => $button);
 }
 
 =head1 multi console support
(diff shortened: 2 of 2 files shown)


testapi.pm
```diff
@@ -108,6 +108,31 @@ os-autoinst is used in the openQA project.
...
 =head1 internal
 
+=head2 calculate_clickpoint
+
+This is used to support several subroutines dealing with mouse clicks, do not call
+this from the tests.
+
+=cut    
+
+sub calculate_clickpoint {
```
Martchus, line 118: Other internal functions within this module already use the convention to start the name with `_` so I suggest to stick with that convention.

testapi.pm
```diff
@@ -108,6 +108,31 @@ os-autoinst is used in the openQA project.
...
+    # This is a non-exported helper function to calculate the clickpoints to limit repetition
+    # and redundancy of code.
```
Martchus, line 120: I'd merge this comment with the description of the function within the POD comment.

testapi.pm
```diff
@@ -527,20 +552,17 @@ sub click_lastmatch {
         last;
     }
 
+    my $x;
+    my $y;
```
Martchus, line 556: I guess declaring them as `my ($x, $y);` would look better.

testapi.pm
```diff
@@ -1549,6 +1571,72 @@ sub mouse_hide(;$) {
...
+will be used to set up the locations and the needle location will be overridden.
+
+=cut
+
+sub mouse_drag {
+    my %args = @_;
+    my $startx =
+      my $starty =
+      my $endx =
+      my $endy = 0;
```
Martchus, line 1592: My Perl parser strikes here.

testapi.pm
```diff
@@ -1549,6 +1571,72 @@ sub mouse_hide(;$) {
...
+        $endx = $args{endx};
+        $endy = $args{endy};
+    }
+    elsif (defined $args{endpoint}) {
+        my $endmatch = $args{endpoint};
+        my $end_matched_needle = assert_screen($endmatch, $args{timeout});
+        ($endx, $endy) = calculate_clickpoint($end_matched_needle);
+    }
+    else {
+        die "The ending point of the drag was not correctly provided. Either provide the 'endx' and 'endy' coordinates, or a needle marking the starting point.";
```
Martchus, line 1622: s/a needle marking the starting point/a needle marking the end point/

testapi.pm
```diff
@@ -1549,6 +1571,72 @@ sub mouse_hide(;$) {
...
+    }
+    # Get the button variable. If no button has been provided, assume the "left" button.
+    my $button = $args{button} //= "left";
```
Martchus, line 1625: Is the assignment to `$args{button}` intended? I'd removed it because it looks like it isn't used in subsequent code anyways.

testapi.pm
```diff
@@ -1549,6 +1571,72 @@ sub mouse_hide(;$) {
...
+    # Now, perform the actual mouse drag. Navigate to the startpoint location,
+    # press and hold the mouse button, then navigate to the endpoint location
+    # and release the mouse button.
+    mouse_set($startx, $starty);
+    query_isotovideo('backend_mouse_button', {button => $button, bstate => 1});
+    mouse_set($endx, $endy);
+    query_isotovideo('backend_mouse_button', {button => $button, bstate => 0});
+
+    # Sleep for a while to leave a trace on the SUT video.
+    sleep("5");
```
Martchus, line 1637: Sleeping isn't nice so I would at least make the time configurable. And is it really just for the video? Because then it is questionable whether leaving a trace in the video justifies slowing down the test by 5 seconds. (Test are already slow enough.)

lruzicka: Ok, the sleeping time is totally unnecessary here, cause anybody could put the sleep in their tests, if they wanted. I will delete it.

testapi.pm
lruzicka: Ok, this is my very first PR, so I did not know, what exactly is requested. I will try to fix what I am able to.

testapi.pm
lruzicka: This was required by the Travis test (number 4) which was choking if there was not such an indentation.

testapi.pm
lruzicka: I think it is used for the `query_isotovideo` calls on lines 1631 and 1633 at least.

testapi.pm
Martchus: Don't refer to tests by their "numbers". I assume you refer to the tidy check. The problem here is that the lines are missing semicolons and doing an initialization like this is quite uncommon in Perl. I would rather write it as `my ($startx, $starty, $endx, $endy);`. (It looks like the default-initialization with 0 is not required here.)

testapi.pm
Martchus: But you're just using `$button` there. Note that I'm talking about the assignment of `$args{button}`. I suggest to replace `//=` with just `//`.

Martchus:
Looks like all issues I've mentioned have been addressed but tests are still missing. Not that the failing **./18-qemu.t** in the OBS build is unrelated. (It fails on my own PR as well and I'll have to look into it.)

testapi.pm
```diff
@@ -527,20 +552,16 @@ sub click_lastmatch {
...
-        $relevant_area = $last_matched_needle->{area}->[-1];
+    if (!$relevant_area and (!$relative_click_point || $relative_click_point eq "center")) {
+        ($x, $y) = _calculate_clickpoint($last_matched_needle);
     }
-    if (!$relative_click_point || $relative_click_point eq 'center') {
-        $relative_click_point = {
-            xpos => $relevant_area->{w} / 2,
-            ypos => $relevant_area->{h} / 2,
-        };
+    else {
```
kalikiana, line 560: `$relevant_area` can still be `undef` or `"center"` here. Maybe these should be separate conditions?

Martchus: That's likely also the reason why the test fails.

lruzicka: Ok, I will make the conditions separate, thanks.

testapi.pm
```diff
@@ -522,25 +552,13 @@ sub click_lastmatch {
...
-            ypos => $relevant_area->{h} / 2,
-        };
-    }
-
-    # calculate absolute click position and click
-    my $x = int($relevant_area->{x} + $relative_click_point->{xpos});
-    my $y = int($relevant_area->{y} + $relative_click_point->{ypos});
+    my ($x, $y);
+    # Calculate the absolute click point.
+    ($x, $y) = _calculate_clickpoint($last_matched_needle, $relevant_area, $relative_click_point);
```
Martchus, line 561:
```suggestion
    # Calculate the absolute click point
    my ($x, $y) = _calculate_clickpoint($last_matched_needle, $relevant_area, $relative_click_point);
```

Martchus (commented):
The code looks good now beside one style issue. I've restarted the CI because it failed to upload coverage data. But even without the report is is clear that the PR would decrease test coverage because not tests for the new test API function are provided. Having a test for the `mouse_drag` function would be desirable.

testapi.pm
lruzicka: Sure! Thanks :+1:

lruzicka:
I am happy that the PR now passes the required test. I will try to provide the tests for the `mouse_drag` function as you request and also for the `_calculate_clickpoint`. Just let me figure it out. :smiley_cat:

t/03-testapi.t
```diff
@@ -683,4 +683,19 @@ subtest 'autoinst_url' => sub {
     is(autoinst_url('foo'), 'http://localhost:1/foo', 'we can configure the hostname that autoinst_url returns');
 };
 
+subtest '_calculate_clickpoint' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+    my @areas        = ({x => 10, y => 10, w => 20, h => 30});
+    $mock_testapi->redefine(assert_screen => {area => \@areas});
+
+    my %fake_needle_area = {x => 100, y => 100, w => 50, h => 40};
+    my %fake_click_point = {xpos => 20, y => 10};
```
Martchus, line 692: Regular braces should be used when initializing a hash.

t/03-testapi.t
```diff
@@ -683,4 +683,19 @@ subtest 'autoinst_url' => sub {
...
+
+    $cmds = [];
+    # Everything is provided.
+    my ($x, $y) = testapi::_calculate_clickpoint('foo', %fake_needle_area, %fake_click_point));
```
Martchus, line 696: Looks like there's one `)` too much. Besides, you should likely pass the hashes as reference, e.g. `… \%fake_needle_area, \%fake_click_point …`.

t/03-testapi.t
```diff
@@ -683,4 +683,19 @@ subtest 'autoinst_url' => sub {
...
+    ok($x == 120, 'xpoint correct');
+    ok($y == 110, 'ypoint correct');
```
Martchus, line 698: Please use `is` here (for better output when the check fails).

Martchus (changes requested):

t/03-testapi.t
lruzicka: What is the difference between this suggestion and the original one? What does "passing as reference" means? Sorry, told you I am not quite experienced in Perl :D

AdamWill: in perl, when e.g. calling a subroutine with an arg that's an array or hash, you can either send *the actual array or hash itself*, or you can send a *reference* to it. A reference is a scalar which refers to a hash or an array (or other things - you can have a coderef, for instance, a reference to a block of code).

You usually don't have to worry about this in Python because Python takes care of it for you and doesn't give you any choices.

https://perldoc.perl.org/perlreftut.html is a good starting point on this stuff. The example @Martchus gave is the way to create a reference to an existing hash: if you have a hash `%foo`, then `\%foo` creates a reference to that hash. You can pass an anonymous reference when calling a subroutine - e.g. `mysub(\%foo);` calls `mysub` with an anonymous reference to `%foo` as the arg - or you can assign it to a scalar: `my $fooref = \%foo;` and then do stuff with the scalar, e.g. `mysub($foo);`.

If you want a subroutine to take arrays or hashes as args it's almost always best to use references, because passing actual arrays and hashes as subroutine args doesn't really work the way you might expect. Especially if the sub takes more than one arg.

AdamWill: if you look at `_calculate_checkpoint` itself, you can see it's expecting to be passed scalars:

    sub _calculate_clickpoint {
        my ($needle_to_use, $needle_area, $click_point) = @_;

and the rest of the code actually shows that `$needle_area` and `$click_point` are (mostly) expected to be hash refs, once you understand references.

    if (!$click_point || $click_point eq 'center') {
        $click_point = {
            xpos => $needle_area->{w} / 2,
            ypos => $needle_area->{h} / 2,
        };
    }

First off, see that use of `->`? That is the convenient way to get a value out of a hash *reference*. It's not how you'd get the value out of a hash. If you had a *hash* `%needle_area` you'd use `$needle_area{w}`.

You can also see that here we're handling the case where `$click_point` is not passed at all or is passed not as a hash ref but as the special string value 'center', and what we're actually doing in those cases is *redefining `$click_point` as a hash ref*. Note we use curly braces to define `$click_point` here: that is a way to create a hash ref directly, as opposed to creating a named hash then creating a reference to it. If you were just creating a hash you'd do `%click_point = (<stuff>)` - you use *regular* braces. Doing `$click_point = {<stuff>}`, with curly braces, creates a new hash *ref*.

lruzicka: Thank you very much, Adam, for this explanation.

AdamWill:
I think the problem you're having with the test failures is because you're doing `ok(mouse_drag(blahblah));`

Per [the docs](https://perldoc.perl.org/Test/More.html#I'm-ok%2c-you're-not-ok.), `ok`'s behaviour is: evaluate whatever expression it's given, and pass if the result is truth-y. Fail if the result is not truth-y.

Why's it failing? Because `mouse_drag` *doesn't explicitly return anything*, which means it actually returns undef. Which is false-y. There's nothing wrong with a sub that returns `undef`, but it does mean you can't test it with `ok`. Instead you probably should just run it, and then check the results were what you expected - i.e. drop the use of `ok` entirely and just rely on the check of `$cmds` afterwards to make sure the `mouse_drag` call did what it was supposed to.

Martchus:
> Because mouse_drag doesn't explicitly return anything, which means it actually returns undef.

Technically it returns the return value of the last statement but that appears to be undef and is nothing we want to check here.

AdamWill:
right, sorry, I elided that a bit :)

t/03-testapi.t
```diff
@@ -683,4 +683,145 @@ subtest 'autoinst_url' => sub {
...
+                cmd => 'backend_mouse_set',
+                x   => 200,
+                y   => 200
+            },
+            {
+                bstate => 0,
+                button => 'left',
+                cmd    => 'backend_mouse_button'
+            },
+    ], 'mouse drag (using coordinates only) was succesful') or diag explain $cmds;
```
Martchus, line 824: I guess successful is written differently but I'd actually remove it completely because that's kind of implied for a test description.

lruzicka: All right.

lruzicka:
This might be ready to merge. Thanks for all your support.

Martchus (approved):

PR #1512, opened 2020-08-20
Use subrepo for update-deps script

perlpunk:
Please see the README for info and instructions:
https://github.com/os-autoinst/os-autoinst-common/blob/master/README.md

This PR creates a subrepo at `external/os-autoinst-common/`
`tools/update-deps` is replaced with a symlink to `external/os-autoinst-common/tools/update-deps`.

See also https://github.com/os-autoinst/openQA/pull/3323 (merged)

Issue: https://progress.opensuse.org/issues/70654


Martchus:
Looks generally good but the first commit message title should be simplified. I also think that the directory name `ext` doesn't really speak for itself. Maybe `external` which would at lest not be ambiguous? Or simply `subrepos`? Or just omit that extra directory level and have a directory `common` directly on the top-level?

perlpunk:
I moved the directory to top: `os-autoinst-common/`

perlpunk:
For some reason the `.gitrepo` file still says the branch is `dev`. Will look into it

okurz (changes requested):
I am reverting my approval because I liked your previous directory name selection better. Same as Martchus I suggest `external/` as top folder to state the origin of the code better than just "os-autoinst-common".

perlpunk:
@okurz then we have a conflict.
@kraih suggested `os-autoinst-common` in the top folder.
Not sure how to proceed if we can't resolve this.
Who is going to decide?

perlpunk:
@kalikiana @okurz @kraih @Martchus please comment only in https://github.com/os-autoinst/openQA/pull/3323

perlpunk:
https://github.com/os-autoinst/openQA/pull/3323 has been merged, and I updated this with the `external` directory.

okurz (approved):

PR #1518, opened 2020-08-26
Refactor tools/tidy

perlpunk:
So that it can be made equal to openQA/tools/tidy

Issue: https://progress.opensuse.org/issues/70384

Tested it also in os-autoinst-distri-opensuse

PR for openQA: https://github.com/os-autoinst/openQA/pull/3334

In the future we could put this into a common git-subrepo, so that updating it in the three repos gets easier.


perlpunk:
https://github.com/os-autoinst/openQA/pull/3334 has been merged

Martchus (approved):

PR #1519, opened 2020-08-26
Disable python3-yamllint when building for SLE < 15 SP2

Martchus:
The package is not available under SLE < 15 SP2 and unlike for SP2 the
build in our devel projects fails.

CMake will just disable the YAML syntax check (emitting a status
message).

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -8,9 +8,9 @@
 #######################################################
 ---
 targets:
-  cpanfile: [                                  cover, devel,         main,             lint, test_base, test              ]
-  spec:     [ build_base, build_legacy, build,        devel,         main, spellcheck,       test_base, test, test_legacy ]
-  docker:   [ build_base, build,                      devel, docker, main, spellcheck, lint, test_base, test              ]
+  cpanfile: [                                  cover, devel,         main,             lint,           test_base, test              ]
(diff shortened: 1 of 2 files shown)


dependencies.yaml
```diff
@@ -8,14 +8,15 @@
 #######################################################
 ---
 targets:
-  cpanfile: [                                  cover, devel,         main,             lint, test_base, test              ]
-  spec:     [ build_base, build_legacy, build,        devel,         main, spellcheck,       test_base, test, test_legacy ]
-  docker:   [ build_base, build,                      devel, docker, main, spellcheck, lint, test_base, test              ]
+  cpanfile: [                                  cover, devel,         main,             lint, yamllint, test_base, test              ]
```
perlpunk, line 11: I think it's not needed in the cpanfile, because this only lists perl modules.

dependencies.yaml
```diff
@@ -8,14 +8,15 @@
...
+  spec:     [ build_base, build_legacy, build,        devel,         main, spellcheck,       yamllint, test_base, test, test_legacy ]
+  docker:   [ build_base, build,                      devel, docker, main, spellcheck, lint, yamllint, test_base, test              ]
   cpanfile-targets:
     # target: cpanfile target type (default main)
     devel: devel
     cover: cover
     test_base: test
+    yamllint: test
```
perlpunk, line 19: see above comment

perlpunk (approved):

PR #1521, opened 2020-08-28
Show only one newline between serial output lines

asdil12:
This should result in a nicer output in the openQA webui

See https://github.com/os-autoinst/openQA/pull/3340 and https://progress.opensuse.org/issues/70648


asdil12:
@Martchus

Martchus (approved):

PR #1523, opened 2020-09-02
Fix error handling for QMP connection when invoking QEMU twice

Martchus:
Follow-up of https://github.com/os-autoinst/os-autoinst/pull/1520 to fix
tests like https://openqa.opensuse.org/tests/1381351 where QEMU is
invoked multiple times.

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -397,6 +397,7 @@ sub exec_qemu {
     session->enable_subreaper;
 
     my $process = $self->_process;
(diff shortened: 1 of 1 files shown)


okurz:
LGTM, thanks for super-fast reaction. Did you do your local testrun yet?

Vogtinator:
LGTM, but suggestion for a refactoring: Have a variable `qemu_running`  instead

Martchus:
It is actually supposed to be "private". I'll merge the PR when my local test succeeds.

PR #1524, opened 2020-09-03
Fix tools/tidy

perlpunk:
In certain cases, `file --mime-type` doesn't return perl, and so the
grep would actually fail, and abort the script.

This was the case on travis ci for os-autoinst-distri-opensuse.

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -58,7 +58,7 @@ fi
 find-files() {
     local files=()
     [[ -d script ]] && files+=(script/*)
-    files=($(file --mime-type * "${files[@]}" | grep text/x-perl | awk -F':' '{ print $1 }'))
+    files=($(file --mime-type * "${files[@]}" | (grep text/x-perl || true) | awk -F':' '{ print $1 }'))
     files+=('**.p[ml]' '**.t')
     if $only_changed; then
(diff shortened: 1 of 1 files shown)


tools/tidy
```diff
@@ -56,6 +56,7 @@ if [ "$perltidy_version_found" != "$perltidy_version_expected" ]; then
 fi
 
 find-files() {
+    set +o pipefail # grep might fail
     local files=()
     [[ -d script ]] && files+=(script/*)
     files=($(file --mime-type * "${files[@]}" | grep text/x-perl | awk -F':' '{ print $1 }'))
```
nicksinger, line 62: Instead of completely disabling pipefails I'd rather add a fallback to the command which can fail:
```
files=($(file --mime-type * "${files[@]}" | (grep text/x-perl || true) | awk -F':' '{ print $1 }'))
```

perlpunk: done

tools/tidy
```diff
@@ -56,6 +56,7 @@ if [ "$perltidy_version_found" != "$perltidy_version_expected" ]; then
 fi
 
 find-files() {
```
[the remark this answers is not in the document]
perlpunk: I used @nicksinger 's suggestion instead

Martchus (approved):

PR #1525, opened 2020-09-04
[ARM64 ThunderX for openQA] Keep ssh connection alive to display long-time run test result

waynechen55:
* **Ssh** console may halt and can not display returned result from long-time run test. It is easy to address this by adding -o TCPKeepAlive=yes -o ServerAliveCountMax=480 -o ServerAliveInterval=60 to ssh command. This will help keep ssh console alive to the maximum of ServerAliveCountMax x ServerAliveInterval seconds from client's perspective.
* **Without** these options, although the ssh console looks alive, it will not display returned test result after long-time run test. So the test will halt and keep waiting for timeout of the whole test suite.
* **This** issue currently only affects ARM64 ThunderX machine
* **Verification runs:**
  * [Test result is not returned and displayed on screen without this modification](http://10.67.133.40/tests/431)
  * [Test result is returned and displayed on screen with this modification](http://10.67.133.40/tests/441)

### DIFF
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -74,10 +74,17 @@ sub screen {
 }
 
 # helper function
+# Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds
+# even without receiving any messsage back from the server, and this will not affect normal
+# ssh disconnect and console switching. Ssh console may not display returned result of
+# long-time run test without these options. TCPKeepAlive ensures that if network goes down
+# or the remote host dies, machines will be properly noticed
 sub sshCommand {
     my ($self, $username, $host, $gui) = @_;
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
+    my $server_alive_count_max = get_var('_SSH_SERVER_ALIVE_COUNT_MAX', 480);
+    my $server_alive_interval  = get_var('_SSH_SERVER_ALIVE_INTERVAL',  60);
+    my $sshopts = "-o TCPKeepAlive=yes -o ServerAliveCountMax=$server_alive_count_max -o ServerAliveInterval=$server_alive_interval -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
 
     if ($gui) {
         $sshopts = "-X $sshopts";
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -38,6 +38,15 @@ AUTOINST_URL_HOSTNAME;string;;hostname or IP adress of host running the autoinst
 
 |====================
 
+.SSH backend
+[grid="rows",format="csv"]
+[options="header",cols="^m,^m,^m,v",separator=";"]
+|====================
+Variable;Values allowed;Default value;Explanation
+_SSH_SERVER_ALIVE_COUNT_MAX;integer;480;Sets the number of server alive messages which may be sent without receiving any messages back from the server. If this threshold is reached while server alive messages are being sent, ssh will disconnect from the server, terminating the session.  The server alive mechanism is valuable when the client or server depend on knowing when a connection has become inactive.
+_SSH_SERVER_ALIVE_INTERVAL;integer;60;Sets a timeout interval in seconds after which if no data has been received from the server, client will send a message through the encrypted channel to request a response from the server.
+|====================
+
 .IPMI backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]


consoles/console.pm
```diff
@@ -74,10 +74,15 @@ sub screen {
 }
 
 # helper function
+# Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds 
+# without receiving any messsage back from the server, and this will not affect normal
+# ssh disconnect and console switching. Ssh console may not display returned result of 
+# long-time run test with these options. TCPKeepAlive ensures that if network goes down
```
Martchus, line 80: You mean without these options?

waynechen55: 1. Updated comments 2. ssh connection will be deems as lost during long-time run test on some machines. So no result can be returned after test finishes.

okurz (changes requested):
please care for spelling in the git commit message, e.g. "diaply" and "adddress". Also please make the git commit message subject line shorter

waynechen55:
> please care for spelling in the git commit message, e.g. "diaply" and "adddress". Also please make the git commit message subject line shorter

Done

consoles/console.pm
```diff
@@ -74,10 +74,15 @@ sub screen {
 }
 
 # helper function
+# Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds
```
okurz, line 77: delete whitespace at end of line, see https://travis-ci.org/github/os-autoinst/os-autoinst/builds/724766444#L842

consoles/console.pm
```diff
@@ -74,10 +74,15 @@ sub screen {
...
+# Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds 
+# even without receiving any messsage back from the server, and this will not affect normal
+# ssh disconnect and console switching. Ssh console may not display returned result of 
+# long-time run test without these options. TCPKeepAlive ensures that if network goes down 
+# or the remote host dies, machines will be properly noticed
 sub sshCommand {
     my ($self, $username, $host, $gui) = @_;
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
+    my $sshopts = "-o TCPKeepAlive=yes -o ServerAliveCountMax=480 -o ServerAliveInterval=60 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
```
okurz, line 85: better extract the values you provide into variables and just define default values to prevent having "magic numbers" here. E.g.

```perl
my $server_alive_count_max = get_var('_SSH_SERVER_ALIVE_COUNT_MAX', 480);
my $server_alive_interval = get_var('_SSH_SERVER_ALIVE_INTERVAL', 60);
my $sshopts = "-o TCPKeepAlive=yes -o ServerAliveCountMax=$server_alive_count_max -o ServerAliveInterval=$server_alive_interval -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
```

and describe the variables in doc/backend_vars.asciidoc

I realized the method "sshCommand" is only used within two specific console files and should not be defined in the generic base level file "consoles/console.pm". I will handle this in a separate PR. This will cause a conflict with your code changes but I can handle that after merging your PR then.

okurz (changes requested):

consoles/console.pm
waynechen55: Why there is preceding underscore to these variables ?

okurz: This is commonly done for test parameters that are not to be used within the test distribution itself. You can also leave out the leading underscore if you prefer

waynechen55: Done. underscore is kept.

okurz:
you still need to attend https://github.com/os-autoinst/os-autoinst/pull/1525/files#r484664251 to fix tests

consoles/console.pm
waynechen55: Done

waynechen55:
> you still need to attend https://github.com/os-autoinst/os-autoinst/pull/1525/files#r484664251 to fix tests

Do you mean travis-ci test results ?

okurz:
Yes. See https://travis-ci.org/github/os-autoinst/os-autoinst/builds/725773367#L1287 where travis CI reports that "test-local-tidy" failed. If you call `tools/tidy` locally then this will tidy the format. You probably know this command from os-autoinst-distri-opensuse.

waynechen55:
> Yes. See https://travis-ci.org/github/os-autoinst/os-autoinst/builds/725773367#L1287 where travis CI reports that "test-local-tidy" failed. If you call `tools/tidy` locally then this will tidy the format. You probably know this command from os-autoinst-distri-opensuse.

I think that is because "Wrong version of perltidy. Found '20190601', expected '20200822'". Any idea ?

okurz:
I assume you see this message locally? Then you need to upgrade tidy. But actually the output in travis CI is "3: Wrong version of perltidy. Found '20200907', expected '20200822'" so there is an even newer version available. We will need to bump the version in os-autoinst. Preparing a PR.

okurz:
We need https://github.com/os-autoinst/os-autoinst/pull/1533 first

waynechen55:
> I assume you see this message locally? Then you need to upgrade tidy. But actually the output in travis CI is "3: Wrong version of perltidy. Found '20200907', expected '20200822'" so there is an even newer version available. We will need to bump the version in os-autoinst. Preparing a PR.

See both locally and also here

waynechen55:
> We need #1533 first

Resolved. But travis-ci still has other problems.

okurz (approved):

PR #1526, opened 2020-09-04
add missing column separators

phil-hands:
The HDDNUMQUEUES_$i line was missing some columns, which when fixed revealed the extra semi-colon in the QEMUTPM line


okurz (approved):
Good fix, thank you

Martchus (approved):

PR #1527, opened 2020-09-07
 Make command server utilize asset cache instead of only relying on ASSETDIR

Martchus:
* Serve assets from working/pool directory and not just from ASSETDIR
* No longer only rely on ASSETDIR being mounted
* See https://progress.opensuse.org/issues/70723


Amrysliu (commented):
LGTM

okurz (approved):

PR #1529, opened 2020-09-08
Move console helper function "sshCommand" to child class where needed

okurz:


okurz:
should wait for #1525 first

consoles/localXvnc.pm
```diff
@@ -27,6 +27,18 @@ require IPC::System::Simple;
...
+sub sshCommand {
+    my ($self, $username, $host, $gui) = @_;
+
+    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
+
+    if ($gui) {
+        $sshopts = "-X $sshopts";
+    }
+
+    return "ssh $sshopts; read";
```
Martchus, line 39: Moving the function makes sense. When moving it anyways you can also get rid of the blank lines.

okurz: done. To go even further see #1531

Martchus (approved):

PR #1530, opened 2020-09-08
ci: Adjust path in codecov.yml

perlpunk:
Issue: https://progress.opensuse.org/issues/71095

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,4 +17,4 @@ coverage:
       default:
         branches: null
 fixes:
-  - "/opt/run/::"
+  - "/opt/::"


Martchus (approved):
Seems like it worked, e.g. https://codecov.io/gh/os-autoinst/os-autoinst/src/2a2ad9c0de5f78f1a2a41878425f0d42d469c604/isotovideo

I've also noticed that the HTML which locally rendered has a similar problem. Is that still the case? (If not we can still merge this PR of course.)

okurz (approved):

perlpunk:
What do I need to do to generate the cover report locally?
`make coverage-html` doesn't work, neither in the top directory nor in build/
```
make: *** No rule to make target 'coverage-html'.  Stop.
```

okurz:
According to `cmake/test-targets.cmake` the target is now called just `coverage` because @Martchus likes to change stuff for the sake of change I guess ;)

perlpunk:
I am getting local test failures, there is no report generated.

Martchus:
@okurz Previously there was the `coverage` and the `coverage-html` target which were both creating the HTML report. I wanted to keep things simple and therefore did not port the `converage-html` alias. I also did not port `test-cover-summary` because I don't think it makes sense to add a target for all the possibly usefol invocations of `cover`.

@perlpunk A complete test run shouldn't be required. It should be sufficient to run a single test just to get any data into the coverage database (see https://github.com/os-autoinst/os-autoinst#cmake).

okurz (approved):
if you don't mind let's go step by step. This fix is already consistent and self-contained so we can merge and do the experiments as part of the ticket. This might still need further changes along the road which touch this code path again though but that is ok to me.

Martchus:
I've already approved the PR so I obviously don't mind. I've nevertheless checked whether it fixes the local report and it actually does (although `/opt` is not used at all locally).

PR #1532, opened 2020-09-09
Suggest using the check target (instead of test target) in the README

Martchus:
Only then the following sentence is actually true. Unfortunately the `test` target itself can not easily be adjusted because it is a special target provided by CMake itself.

At least it is consistent with the Autotools check target this way.


kalikiana (approved):
Nice catch! Sadly a design flaw of CMake.

PR #1533, opened 2020-09-10
Update perltidy to 20200907

okurz:


perlpunk (approved):

okurz:
The last published Tumbleweed snapshot has the updated version already as well, see https://software.opensuse.org/package/perl-Perl-Tidy

PR #1536, opened 2020-09-18
Include "t/" in coverage analysis as well to prevent never called tests

okurz:
Last time this was proposed in https://github.com/os-autoinst/os-autoinst/pull/1289

Other previous approaches:
* https://github.com/os-autoinst/os-autoinst/pull/1291
* https://github.com/os-autoinst/os-autoinst/pull/1292
* https://github.com/os-autoinst/os-autoinst/pull/1537
* https://github.com/os-autoinst/os-autoinst/pull/1591
* https://github.com/os-autoinst/os-autoinst/pull/1624

why am I even trying?

### DIFF
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -58,7 +58,5 @@ if [[ $TESTS = *[!\ ]* ]]; then
     "$prove_path" $PROVE_ARGS $TESTS
 else
     # shellcheck disable=SC2086
-    cd t && "$prove_path" $PROVE_ARGS -r
-    # shellcheck disable=SC2086
-    cd ../xt && "$prove_path" $PROVE_ARGS -r
+    "$prove_path" $PROVE_ARGS -r t/ xt/
 fi
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -43,7 +43,7 @@ db="$build_directory/cover_db"
 export PERL5LIB="$source_directory:$source_directory/ppmclibs:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
 if [[ $WITH_COVER_OPTIONS ]]; then
     path_regex="^|$source_directory/|\\.\\./"
-    select="($path_regex)(OpenQA|backend|consoles|ppmclibs)/|($path_regex)isotovideo|($path_regex)[^/]+\.pm,-ignore,\.t|data/tests/|fake/tests/|/CheckGitStatus.pm|$prove_path"
+    select="($path_regex)(OpenQA|backend|consoles|ppmclibs|t)/|($path_regex)isotovideo|($path_regex)[^/]+\.(pm|t),-ignore,/CheckGitStatus.pm|$prove_path"
     export PERL5OPT="-I$source_directory/t/lib -MOpenQA::Test::CheckGitStatus -MDevel::Cover=-db,$db,-select,$select,-coverage,statement"
 fi
 


okurz:
fails with:

```
[2/3] cd /opt && /opt/tools/invoke-tests --coverage --skip-if-cover-db-exists --prove-tool /usr/bin/prove --make-tool /usr/bin/ninja --build-directory /opt
[3/3] cd /opt && /usr/bin/cover -report codecov /opt/cover_db
FAILED: CMakeFiles/coverage-codecov 
cd /opt && /usr/bin/cover -report codecov /opt/cover_db
Reading database from /opt/cover_db
Devel::Cover: Warning: can't open 28-signalblocker.t for MD5 digest: No such file or directory
Devel::Cover: Warning: can't open 09-lockapi.t for MD5 digest: No such file or directory
Devel::Cover: Warning: can't open 10-terminal.t for MD5 digest: No such file or directory
Devel::Cover: Warning: can't open 11-image-ppm.t for MD5 digest: No such file or directory
Devel::Cover: Warning: can't open 16-send_with_fd.t for MD5 digest: No such file or directory
Devel::Cover: Warning: can't open 17-basetest.t for MD5 digest: No such file or directory
```

anyone an idea why?

perlpunk:
Add it to `($path_regex)(OpenQA|backend|consoles|ppmclibs)`?
Probably just `($path_regex)(OpenQA|backend|consoles|ppmclibs|t)`

perlpunk:
and maybe the tests have to be called with their absolute name, due to the unusual non-perl setup we have here.

Martchus:
> the unusual non-perlish setup we have here

Maybe it is finally time to execute tests within the top-level directory. I suppose most tests should already work that way anyways.

okurz:
> Maybe it is finally time to execute tests within the top-level directory. I suppose most tests should already work that way anyways.

-> https://github.com/os-autoinst/os-autoinst/pull/1537

okurz:
despite all checks being passed there are many warnings in travis CI output and the actual *.t files are not listed in coverage.

okurz:
@Mergifyio rebase

okurz:
https://github.com/os-autoinst/os-autoinst/runs/4485273873?check_suite_focus=true#step:3:890 says "Cannot read 00-compile-check-all.t: No such file or directory at /usr/lib/perl5/site_perl/5.26.1/Devel/Cover/Report/Codecovbash.pm line 76.". I don't know right now why tests are listed from top-level as if they would not be in the subfolder "t/"

perlpunk:
> I don't know right now why tests are listed from top-level as if they would not be in the subfolder "t/"

Because we `cd` into the test folder:
https://github.com/os-autoinst/os-autoinst/pull/1536/files#diff-926440f28c20c95a9687a624a95204e9034bfb0276ef9383597c896c5673dfc1R60-R61

I once tried to make the tests work without having to do that `cd`, but then I think some code coverage got lost.

It would be really good if we could put all modules under the lib/ dir and all scripts under scripts/ or bin/, so that we have all files under some directory and not in the root folder.
I believe the current setup makes things like coverage setup harder.

okurz:
I just *created* https://github.com/os-autoinst/os-autoinst/pull/1536/commits/575f1633a64451ea2c44e35e3521280c1749eb7e and then that code looked so familiar … I already did the same already in another local branch which I eventually gave up as I could not get it to work :sob:

okurz:
ok, all looks cool now. But still I would prefer to go step by step and do https://github.com/os-autoinst/os-autoinst/pull/1883 first separately

okurz:
> > I don't know right now why tests are listed from top-level as if they would not be in the subfolder "t/"
> 
> Because we `cd` into the test folder: https://github.com/os-autoinst/os-autoinst/pull/1536/files#diff-926440f28c20c95a9687a624a95204e9034bfb0276ef9383597c896c5673dfc1R60-R61
> 
> I once tried to make the tests work without having to do that `cd`, but then I think some code coverage got lost.
> 
> It would be really good if we could put all modules under the lib/ dir and all scripts under scripts/ or bin/, so that we have all files under some directory and not in the root folder. I believe the current setup makes things like coverage setup harder.

So regardless of any move to lib/ and scripts/ I see the problem as follows: Whenever you put coverage into a coverage DB you should ensure that you are doing it from the same directory, so that Devel::Cover can match and locate the files in the coverage DB. Meaning that chdir in tests is harmful. We either try to run everything from the root source folder or everything from another folder, previously from within t/ but of course we could also run everything from the build folder or a temporary directory. I am trying yet again to include t/ in coverage but without changing where we call tests from, just trying to get `-select` and `-ignore` right for Devel::Cover because I think I might slowy understand how that works :)

okurz:
obsoleted by https://github.com/os-autoinst/os-autoinst/pull/1886

PR #1538, opened 2020-09-21
Fix "sshCommand" after moving to consoles/localXvnc

okurz:
Fixes the problem introduced in 148c594c

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -27,11 +27,23 @@ require IPC::System::Simple;
 use Socket;
 use File::Which;
 
+# helper function
+# Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds
+# even without receiving any messsage back from the server, and this will not affect normal
+# ssh disconnect and console switching. Ssh console may not display returned result of
+# long-time run test without these options. TCPKeepAlive ensures that if network goes down
+# or the remote host dies, machines will be properly noticed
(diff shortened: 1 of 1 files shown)


Martchus (approved):
Looks like this PR restores #1525 after it has been reverted by merging #1529.

consoles/localXvnc.pm
```diff
@@ -27,11 +27,23 @@ require IPC::System::Simple;
...
+# even without receiving any messsage back from the server, and this will not affect normal
+# ssh disconnect and console switching. Ssh console may not display returned result of
+# long-time run test without these options. TCPKeepAlive ensures that if network goes down
+# or the remote host dies, machines will be properly noticed
 sub sshCommand {
     my ($self, $username, $host, $gui) = @_;
 
-    my $sshopts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PubkeyAuthentication=no $username\@$host";
-    $sshopts = "-X $sshopts" if $gui;
+    my $server_alive_count_max = get_var('_SSH_SERVER_ALIVE_COUNT_MAX', 480);
```
mdoucha, line 39: `get_var()` is not defined in this module. You need to add `use testapi;`. All our s390 workers are currently broken because of that.

mdoucha: https://openqa.suse.de/tests/4723195

PR #1539, opened 2020-09-21
.gitignore: Ignore files created by current state of tests

okurz:


Martchus (approved):

perlpunk (approved):

perlpunk (commented):
I think it would be better to actually delete such files at the end of the tests.
Leftover files can influence other tests, so I believe every test should start with a clean environment.

okurz:
@perlpunk I agree with that. We already tried that in the past and we can do so again.

perlpunk:
I just ran all tests and could only see two leftover files:
* `doc/backend_vars.asciidoc.newvars`
* `t/serial0` (created by `29-backend-generalhw.t`, should be easy to fix)

Can you share your workflow where you got those leftover files you added to .gitignore?

Maybe we should also add a `git status` in `.travis.yml` after the tests (and ideally at some point check that it does not have any leftover files).

okurz:
hm, maybe the files differ when running checks and tests based on autotools or cmake

PR #1542, opened 2020-09-23
Fix tty key definition after initialization

okurz:
This fixes a regression introduced in c2ba84c4 when tests update the
"tty" of a console after initialization. By introducing a method to read
the console key combination we can evaluate the configuration
dynamically.

Verification run: http://lord.arch.suse.de/tests/2793

### DIFF
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -50,11 +50,9 @@ sub new {
 
 sub init {
     my ($self) = @_;
-    return unless $self->{args}->{tty};
     # Special keys like Ctrl-Alt-Fx are not passed to the VM by xfreerdp.
     # That means switch from graphical to console is not possible on Hyper-V.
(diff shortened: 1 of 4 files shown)


consoles/console.pm
```diff
@@ -108,4 +106,10 @@ sub set_tty {
     # that this method is called in the right process
 }
 
+sub console_key {
+    my ($self) = @_;
+    return unless $self->{console_hotkey} && $self->{args}->{tty};
```
kalikiana, line 111:
```suggestion
    return undef unless $self->{console_hotkey} && $self->{args}->{tty};
```

We should always return an explicit value

okurz: fixed

Vogtinator:
Example for a failure caused by this: https://openqa.opensuse.org/tests/1400997

Martchus (commented):
I suppose this works now but it would make sense to test whether e.g. the job @Vogtinator mentioned now passes.

okurz:
Started a verification run: http://lord.arch.suse.de/tests/2793 (running)

PR #1543, opened 2020-09-24
Support FORCE_PUBLISH_HDD_ when the job fails

Amrysliu:
For now when jobs fail, and the failed module with flag 'fatal =>1',
(so do some other specify scenarios)the FORCE_PUBLISH_HDD does not make sense.
We could get this conclusion by this line code:
https://github.com/os-autoinst/os-autoinst/blob/master/autotest.pm#L394

```
if ($t->{fatal_failure} || $flags->{fatal} || (!exists $flags->{fatal} && !$snapshots_supported) || $bmwqemu::vars{TESTDEBUG}) {
                bmwqemu::stop_vm();
                return 0;
}
```
When `isotovideo` deals with the `FORCE_PUBLISH_HDD_` in `handle_generated_assets`,
if the `$command_handler->test_completed` is 0, it does nothing. So when
jobs fail, and the function `runalltests` in `autotest.pm` returns 0, the asset defined
by `FORCE_PUBLISH_HDD` will not be published.

Verify run: http://10.67.19.103/tests/342#downloads
(this job has a setting `FORCE_PUBLISH_HDD_1=test.qcow2`)

Related: https://progress.opensuse.org/issues/65004

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -143,7 +143,7 @@ sub handle_generated_assets {
     my ($command_handler, $clean_shutdown) = @_;
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $command_handler->test_completed && ($bmwqemu::vars{BACKEND} eq 'qemu');
+    return unless $bmwqemu::vars{BACKEND} eq 'qemu';
     my @toextract;
     my $nd = $bmwqemu::vars{NUMDISKS};
     if ($command_handler->test_completed) {
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -13,7 +13,8 @@ use Mojo::File qw(tempdir path);
 use Mojo::JSON qw(decode_json);
 use Mojo::Util qw(scope_guard);
 use FindBin '$Bin';
-use OpenQA::Isotovideo::Utils qw(load_test_schedule);
+use OpenQA::Isotovideo::Utils qw(load_test_schedule handle_generated_assets);
+use OpenQA::Isotovideo::CommandHandler;
 
 my $dir          = tempdir("/tmp/$FindBin::Script-XXXX");
 my $toplevel_dir = abs_path(dirname(__FILE__) . '/..');
@@ -124,6 +125,36 @@ subtest 'upload assets on demand even in failed jobs' => sub {
     like $log, qr/qemu-img.*foo.qcow2/, 'requested image is published even though the job failed';
     ok(-e $pool_dir . '/assets_public/foo.qcow2', 'published image exists');
     ok(!-e $pool_dir . '/base_state.json',        'no fatal error recorded');
+};
+
+# mock backend/driver
+{
+    package FakeBackendDriver;
+    sub new {
+        my ($class, $name) = @_;
+        my $self = bless({class => $class}, $class);
+        require "backend/$name.pm";
+        $self->{backend} = "backend::$name"->new();
+        return $self;
+    }
+    sub extract_assets {
+        my $self = shift;
+        $self->{backend}->do_extract_assets(@_);
+    }
+}
+
+subtest 'upload the asset even in an incomplete job' => sub {
+    my $command_handler = OpenQA::Isotovideo::CommandHandler->new();
+    $bmwqemu::vars{BACKEND}             = 'qemu';
+    $bmwqemu::vars{NUMDISKS}            = 1;
+    $bmwqemu::vars{FORCE_PUBLISH_HDD_1} = 'force_publish_test.qcow2';
+    $bmwqemu::vars{PUBLISH_HDD_1}       = 'publish_test.qcow2';
+    $command_handler->test_completed(0);
+    $bmwqemu::backend = FakeBackendDriver->new('qemu');
+    my $return_code = handle_generated_assets($command_handler, 1);
+    is $return_code, 0, 'The asset was uploaded success';
+    ok(-e $pool_dir . '/assets_public/force_publish_test.qcow2', 'test.qcow2 image exists');
+    ok(!-e $pool_dir . '/assets_public/publish_test.qcow2',      'the asset defined by PUBLISH_HDD_X would not be generated in an incomplete job');
 };
 
 subtest 'error handling when loading test schedule' => sub {


Martchus (approved):
Makes sense; otherwise the subsequent if block `if ($command_handler->test_completed) {` seems useless anyways.

okurz (changes requested):
I checked t/14-isotovideo.t and the runtime increased from 12s to 19s with your changes. It is good that you add a test but I consider the runtime increase too high to be easily acceptable.

t/14-isotovideo.t
```diff
@@ -152,6 +152,20 @@ subtest 'upload assets on demand even in failed jobs' => sub {
...
+    path('vars.json')->remove         if -e 'vars.json';
+    my $module = 'tests/failing_module_with_fatal';
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests schedule=$module force_publish_hdd_1=foo2.qcow2 qemu_no_kvm=1 arch=i386 backend=qemu qemu=i386", exit_code => 0);
+    } qr/scheduling failing_module_with_fatal $module\.pm/, 'module scheduled';
+    like $log, qr/qemu-img.*foo2.qcow2/, 'requested image is published even though the job failed and the module has "fatal =>1"';
+    ok(-e $pool_dir . '/assets_public/foo2.qcow2', 'published image exists');
+    ok(!-e $pool_dir . '/base_state.json',         'no fatal error recorded');
+
+};
```
okurz, line 167: this is quite some duplication

Amrysliu:
> I checked t/14-isotovideo.t and the runtime increased from 12s to 19s with your changes. It is good that you add a test but I consider the runtime increase too high to be easily acceptable.

I remove the test I added,  so the runtime will not increase. I changed the test module code in order to cover this scenario.

okurz:
Well, your previous test was nice as it explicitly tested the feature, it just took too long. Maybe we can optimize that

Amrysliu:
> Well, your previous test was nice as it explicitly tested the feature, it just took too long. Maybe we can optimize that

I modified the test module to had the `{fatal => 1}` flag, it also tested the feature. If it is not enough, I could add the previous test back.

okurz:
The thing is: That one module is "fatal" is not obvious why the test is like that to anyone reading the test code or changing it later. So an explicit test is better. Maybe if you add back your original idea and mock all expensive functions it is more efficient.

Amrysliu:
> The thing is: That one module is "fatal" is not obvious why the test is like that to anyone reading the test code or changing it later. So an explicit test is better. Maybe if you add back your original idea and mock all expensive functions it is more efficient.

I have a try, the expensive functions is 'attempt' in `osutils.pm` and `stop_vm` in `bmwqemu.pm`. I also wrote the mock in this subtest, but the mock does work in my test,  the original functions are still called. Do you have any suggestions?

Martchus:
The `t/14-isotovideo.t` test is invoking isotovideo as a new sub process so the usual mocking doesn't work. I would say one needed to refactor the code (move it out of isotovideo) and then just call the moved-out function from the test.

> That one module is "fatal" is not obvious why the test is like that to anyone reading the test code or changing it later.

Some comments would also help to make that clear.

okurz:
> The `t/14-isotovideo.t` test is invoking isotovideo as a new sub process so the usual mocking doesn't work. I would say one needed to refactor the code (move it out of isotovideo) and then just call the moved-out function from the test.

Yes, I think this is a good idea. Worth a try :)

okurz:
I think https://github.com/os-autoinst/os-autoinst/pull/1545 can help you

Amrysliu:
> > The `t/14-isotovideo.t` test is invoking isotovideo as a new sub process so the usual mocking doesn't work. I would say one needed to refactor the code (move it out of isotovideo) and then just call the moved-out function from the test.
> 
> Yes, I think this is a good idea. Worth a try :)

If a test module does not have the flag `{fatal => 1}`, will the job be considered passed? I am not sure because I have not been aware of it. But according to the code, `if ($command_handler->test_completed)`, the PUBLISH_HDD will be generated, we could see that the job is considered passed. So I think the test module `failing_module` has no `{fatal => 1}` is not reasonable in this scenario.

okurz:
> If a test module does not have the flag `{fatal => 1}`, will the job be considered passed? I am not sure because I have not been aware of it.

http://open.qa/docs/#_how_to_write_tests is the complete description. fatal does not impact the overall result but only changes the behaviour if other modules should be run or not at all.

> But according to the code, `if ($command_handler->test_completed)`, the PUBLISH_HDD will be generated, we could see that the job is considered passed. So I think the test module `failing_module` has no `{fatal => 1}` is not reasonable in this scenario.

As explained above, *no* `{fatal => 1}` does *not* mean that the job would be considered passed.

t/14-isotovideo.t
```diff
@@ -152,6 +153,36 @@ subtest 'upload assets on demand even in failed jobs' => sub {
...
+        $self->{backend} = "backend::$name"->new();
+        return $self;
+    }
+    sub extract_assets {
+        my $self = shift;
+        $self->{backend}->do_extract_assets(@_);
+    }
+}
+
+subtest "upload assets even in an incomplete job " => sub {
```
okurz, line 172: please use single ticks in strings unless you want to evaluate perl variables. Also there is an superfluous space after "job".

okurz (changes requested):
Very nice new test. As the functionality you want to cover does not exist in isotovideo itself anymore I suggest to create a test module specific to the file `OpenQA/Isotovideo/Utils.pm`, e.g. `t/19-isotovideo-utils.t`, next to `t/19-isotovideo-command-processing.t`. Should be feasible to move the test code there. If this is too complicated then leave it in place where you have it.

Martchus (approved):

Amrysliu:
> Very nice new test. As the functionality you want to cover does not exist in isotovideo itself anymore I suggest to create a test module specific to the file `OpenQA/Isotovideo/Utils.pm`, e.g. `t/19-isotovideo-utils.t`, next to `t/19-isotovideo-command-processing.t`. Should be feasible to move the test code there. If this is too complicated then leave it in place where you have it.

Because `14-isotovideo.t` also includes some other tests about the Isotovideo/Utils.pm, such as `load_test_schedule`, if you don't mind, I prefer to do not that in this pr.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -143,7 +143,7 @@ sub handle_generated_assets {
     my ($command_handler, $clean_shutdown) = @_;
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $command_handler->test_completed && ($bmwqemu::vars{BACKEND} eq 'qemu');
+    return if ($bmwqemu::vars{BACKEND} ne 'qemu');
```
okurz, line 146: I would prefer if you keep the unless, i.e. `return unless $bmwqemu::vars{BACKEND} eq 'qemu';`

t/14-isotovideo.t
```diff
@@ -126,6 +127,36 @@ subtest 'upload assets on demand even in failed jobs' => sub {
...
+        $self->{backend} = "backend::$name"->new();
+        return $self;
+    }
+    sub extract_assets {
+        my $self = shift;
+        $self->{backend}->do_extract_assets(@_);
+    }
+}
+
+subtest 'upload the asset even in an test incomplete job' => sub {
```
okurz, line 146: "even in an test incomplete job"? Better: "even in an incomplete job", just strip the "test"

OpenQA/Isotovideo/Utils.pm
Amrysliu: fixed

okurz (approved):

PR #1544, opened 2020-09-24
t: Prevent .git files to interfer with local 00-compile-check-all runs

okurz:
As we look up all files which look like perl files in the root source
dir and all subdirectories also for example files within .git/rr-cache
are picked up. As these files either have the extension .pl, .pm, .t or 
the perl shebang but include patch conflict markers Test::Strict will
unhelpful record failed result. This can be avoided by just listing all
version controlled files which is also faster than manually traversing
the directory tree.

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -1,5 +1,5 @@
 #!/usr/bin/perl
-# Copyright (C) 2015-2019 SUSE LLC
+# Copyright (C) 2015-2020 SUSE LLC
 #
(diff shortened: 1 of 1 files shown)


perlpunk:
Can you describe the problem?
btw, as far as I understand it, this would be completely avoidable if we had all our perl modules under lib (and scripts under bin). Sadly this is not easy to fix.

okurz:
updated the git commit message to further describe the problem.

Martchus (approved):

PR #1545, opened 2020-09-25
Extract 'handle_generated_assets' function from isotovideo for easier testing

okurz:


OpenQA/Isotovideo/Utils.pm
```diff
@@ -131,6 +132,57 @@ sub checkout_git_refspec {
...
 
+=head2 handle_generated_assets
+
+Handles the assets generated by the test depending on status and test
+configuration variables.
+
+=cut
+
+sub handle_generated_assets {
+    my ($command_handler, $return_code, $clean_shutdown) = @_;
```
Martchus, line 143: The `$return_code` should be passed as scalar ref (and the assignments should change the value referred by `$return_code`) so the remaining code in `isotovideo` has the updated value.

Martchus (changes requested):

OpenQA/Isotovideo/Utils.pm
Martchus: So you would call it like `handle_generated_assets($command_handler, \$return_code, $clean_shutdown);` and do the assignments like `$$return_code = 1`.

okurz: changed to not take the variable at all but return a value which is used within the calling scope together with a check if we should call the function at all.

Martchus (changes requested):

OpenQA/Isotovideo/Utils.pm
```diff
@@ -131,6 +132,57 @@ sub checkout_git_refspec {
...
+        my $name = $bmwqemu::vars{"FORCE_PUBLISH_HDD_$i"} || next;
+        diag "Requested to force the publication of '$name'";
+        push @toextract, _store_asset($i, $name, 'assets_public');
+    }
+    for my $asset (@toextract) {
+        local $@;
+        eval { $bmwqemu::backend->extract_assets($asset); };
+        if ($@) {
+            bmwqemu::serialize_state(component => 'backend', msg => "unable to extract assets: $@", log => 1);
+            return 1;
```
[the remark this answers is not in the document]
okurz: yes, you are right. Changed that to save the value into a local variable and return that instead at the end of the function body.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -131,6 +132,57 @@ sub checkout_git_refspec {
...
+        }
+        if ($bmwqemu::vars{UEFI} && $bmwqemu::vars{PUBLISH_PFLASH_VARS}) {
+            push(@toextract, {pflash_vars => 1,
+                    name   => $bmwqemu::vars{PUBLISH_PFLASH_VARS},
+                    dir    => 'assets_public',
+                    format => 'qcow2'});
+        }
+        if (@toextract && !$clean_shutdown) {
+            bmwqemu::serialize_state(component => 'isotovideo', msg => 'unable to handle generated assets: machine not shut down when uploading disks', log => 1);
+            return 1;
```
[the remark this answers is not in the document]
okurz: in before it was

```perl
$return_code = 1;
return;
```

so I consider this equivalent

Martchus (approved):

PR #1546, opened 2020-09-28
Improve shutdown

Martchus:
* See particular commit messages and https://progress.opensuse.org/issues/71110
* I came up with these changes when investigating the os-autoinst shutdown code but the real improvement is actually done in https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/pull/13.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -282,7 +282,13 @@ sub start_process {
             bmwqemu::load_vars();
 
             run_all;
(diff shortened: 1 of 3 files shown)


okurz (commented):
I don't see the improvement except the log call. I suggest you split that part out, the part about the options we can argue about in another PR :)

okurz (approved):
Actually, nevermind, it's fine.

PR #1548, opened 2020-10-02
Prevent all warnings and common style issues with "strictures"

okurz:
With "strictures" we can replace "strict" and "warnings", make most
warnings fatal but only within development environments. This also
replaces what "Test::Warnings" offered so far.

Changed with command:

```
git grep -l 'use strict;' | xargs sed -i -e 's/use strict;/use strictures;/' -e '/use warnings;/d'; git grep -l 'use Mojo::Base -base;' | xargs grep -L 'use strict' | xargs sed -i -e 's/use Mojo::Base -base;/use strictures;\n&/'; git grep -l 'use Test::Warnings' | xargs sed -i -e '/use Test::Warnings/d'
```

Related progress issue: https://progress.opensuse.org/issues/10704


okurz:
I wonder if anyone has a good idea how to get rid of https://github.com/os-autoinst/os-autoinst/pull/1548/files#diff-6e44a11b903ef2cd087e0f9a3c1bff9aR54 and alike

autotest.pm
```diff
@@ -52,6 +51,7 @@ loadtest is called.
 
 =cut
 
+no warnings 'redefine';    # unclear why this is needed
```
kalikiana, line 54: I think that's the `use bmwqemu 'diag';` in `Utils.pm`. There's one `diag` call that isn't namespaced.

kalikiana: See #1549 for a fix

okurz:
https://travis-ci.org/github/os-autoinst/os-autoinst/builds/733313494#L1011 still fails with `# Subroutine find_script redefined at /opt/autotest.pm line 54.` This is triggered by https://github.com/os-autoinst/os-autoinst/blob/master/bmwqemu.pm#L241 with `use autotest '$current_test';` .

I assume the restricted imported does not help here because basically we have a circular dependency and that is problematic in general. I wonder if anyone has a good suggestion how to replace importing a variable from another package.

perlpunk:
> This is triggered by https://github.com/os-autoinst/os-autoinst/blob/master/bmwqemu.pm#L241 with `use autotest '$current_test';` .

It doesn't solve the problem, but importing `$current_test` seems useless, if the code is actually using the fully qualified package name `$autotest::current_test`.
I guess we can just remove the import and remove it from `@autotest::EXPORT_OK`, since it is not used anyhere.

And yes, circular dependencies should be avoided and resolved.

perlpunk:
thinking...

A first step could be to factor out subroutines from bmwqemu and autotest that don't depend on each other into a third module, like `bmwqemu::diag` and the following log functions. Let `bmwqemu::diag` just be a wrapper and use the new function throughout our code.

Then replacing
```
use autotest '$current_test';
sub current_test {
    return $autotest::current_test;
}
```
with
```
sub current_test {
    require autotest;
    return $autotest::current_test;
}
```
would get rid of the warning, I guess.
(which can be tested easily with `perl -I. -c bmwqemu.pm`).

In general we should avoid providing global vars to other modules.

Years ago I have been refactoring code with a hundred global variables, but this was easy because the codebase was no used by a third party.

okurz:
thanks for the hints. I will look into this later as this shouldn't be urgent.

PR #1549, opened 2020-10-06
Consistently use diag in the bmwqemu namespace

kalikiana:
To avoid *redefining* diag from testapi.

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -20,7 +20,7 @@ use Mojo::URL;
 use Exporter 'import';
 use File::Spec;
 use Cwd;
-use bmwqemu 'diag';
+use bmwqemu;
 use Try::Tiny;
 
 our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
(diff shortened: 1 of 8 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -20,7 +20,7 @@ use Mojo::URL;
 use Exporter 'import';
 use File::Spec;
 use Cwd;
-use bmwqemu 'diag';
+use bmwqemu;
```
perlpunk, line 23: Since bmwqemu has two default exports, it might be better to do
`use bwmqemu ();` in order to import nothing at all.

kalikiana: Ah, because `fileContent` and `save_vars` are exported otherwise even if they're unused, good point!

kalikiana: We never seem to make use of those actually. Maybe they should just not be exported?

perlpunk: good idea. I also cannot see it used implicitly in os-autoinst-distri-opensuse

perlpunk (approved):

PR #1553, opened 2020-10-22
Use OpenQA::Test::TimeLimit in some tests, especially ones invoking QEMU

Martchus:
See
* https://progress.opensuse.org/issues/71110
* https://progress.opensuse.org/issues/67039


perlpunk:
Just a note: do not rebase or squash commits generated by `git subrepo`, as the commit sha in the `.gitrepo` file points to the parent commit, and later operations will fail if they can't find the commit. (this can be manually fixed though by editing the `.gitrepo` file though).
If the parent commit stays the same, it shouldn't be a problem.

okurz (approved):

kalikiana (approved):

PR #1558, opened 2020-11-06
Add python binding/import/wrapping code for testapi

okurz:
This is based on bmwiedemann's work, see for example the initial idea in
https://github.com/os-autoinst/os-autoinst/pull/364 .
The idea for "python bindings" was recently brought up as well in
See https://trello.com/c/GSXjk2wZ/35-python-bindings-for-openqa

With this test modules can written in python code importing the testapi
module from python. All perl objects not within the testapi can still be
accessed using the dynamic python module "perl".

Detailed changes:

* 08-autotest.t: Add test for python test modules
* t: Add test for python test modules using perl testapi
* Add python testapi interface

This is a resubmit of
https://github.com/os-autoinst/os-autoinst/pull/1089 updated to the
current state of the project but keeping the original idea. By now
different approaches have been followed to improve the maintainability
of complex test plans. However the concerns in
https://github.com/os-autoinst/os-autoinst/pull/1089 have not been in
addressed to change the design.

Related progress issue: https://progress.opensuse.org/issues/72868

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -67,7 +67,7 @@ Source0:        %{name}-%{version}.tar.xz
 # The following line is generated from dependencies.yaml
 %define test_version_only_requires perl(Mojo::IOLoop::ReadWriteProcess) >= 0.28
 # The following line is generated from dependencies.yaml
-%define test_requires %build_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(YAML::PP)
+%define test_requires %build_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(Inline::Python) perl(YAML::PP)
 # The following line is generated from dependencies.yaml
 %define devel_requires %test_requires perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy)
 BuildRequires:  %test_requires %test_version_only_requires
@@ -76,6 +76,8 @@ Recommends:     tesseract-ocr
 Recommends:     /usr/bin/xkbcomp /usr/bin/Xvnc dumponlyconsole
 Recommends:     qemu >= 2.0.0
 Recommends:     /usr/bin/qemu-img
+# Optional dependency for Python test API support
+Recommends:     perl(Inline::Python)
 Requires(pre):  %{_bindir}/getent
 Requires(pre):  %{_sbindir}/useradd
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/autotest.pm
+++ b/autotest.pm
@@ -131,7 +131,7 @@ sub loadtest {
     eval $code;
     if ($@) {
         my $msg = "error on $script: $@";
-        bmwqemu::diag($msg);
+        bmwqemu::fctwarn($msg);
         bmwqemu::serialize_state(component => 'tests', msg => "unable to load $script, check the log for the cause (e.g. syntax error)");
         die $msg;
     }
--- a/autotest.pm
+++ b/autotest.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2020 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -116,13 +116,14 @@ sub loadtest {
         $code .= "require '$script_path';";
     }
     elsif ($script =~ m/\.py$/) {
-        # import the test API and all methods from the test API by default
-        # into python context
+        # Adding the include path of os-autoinst into python context
+        my $inc = File::Basename::dirname(__FILE__);
         $code .= "
             use base 'basetest';
-            use Inline Python => <<'END_OF_PYTHON_CODE';\n";
-        $code .= `cat $casedir/$script`;
-        $code .= "\nEND_OF_PYTHON_CODE\n";
+            use Mojo::File 'path';
+            use Inline Python => 'import sys; sys.path.append(\"$inc\")';
+            use Inline Python => path('$casedir/$script')->slurp;
+            ";
     }
     else {
         die "impossible codepath";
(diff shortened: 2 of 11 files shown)


foursixnine:
Just 2c: https://github.com/os-autoinst/os-autoinst/pull/1089#issuecomment-459260356 and https://github.com/os-autoinst/os-autoinst/pull/1089#issuecomment-459264773

I also believe that it's still a bad idea to have mixed code.

okurz:
as python2 is getting old – and might explain the OBS failures – I created a submission to updated perl-Inline-PYthon to python3: https://build.opensuse.org/request/show/848418

okurz:
I failed to spot the error in https://travis-ci.org/github/os-autoinst/os-autoinst/builds/745502531 but I think if exit code returned is "104" then this might come from zypper where 104 means "ZYPPER_EXIT_INF_CAP_NOT_FOUND" means "Returned by the install and the remove command in case any of the arguments does not match any of the available (or installed) package names or other capabilities." says the man page. I assume this means that one of the added dependencies is not found as a package. Because tools/docker_run_ci calls 
tools/install-new-deps.sh I assume this means that we did not specify a dependency which we should? Or maybe need to specify that still in one file within the openQA repo?

kalikiana:
Seems like all attempts to execute tests fail like this now

    6:     # [2021-01-04T10:05:19.541 UTC] [debug] error on tests/pre_boot.py: Error -- py_eval raised an exception at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Inline/Python.pm line 177.

which I'm guessing means, since it fails in [py_eval](https://metacpan.org/pod/distribution/Inline-Python/Python.pod#py_eval()), an exception was thrown *after* importing the Python testapi (since otherwise that'd be the error message). But not the actual error on stderr coming from a Python exception.

kalikiana:
Still seeing this in the latest result 🤔

         # error on tests/pre_boot.py: Error -- py_eval raised an exception at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Inline/Python.pm line 177.

okurz:
yes, I would welcome help on this issue :)

kalikiana:
> yes, I would welcome help on this issue :)

What did you try to address it so far? I can't tell from the last commit. It looks exactly the same when I was investigating *py_eval*.

okurz:
Yes, nothing changed. Months ago the tests were still fine AFAIR but now we have this problem.

dist/rpm/os-autoinst.spec
```diff
@@ -36,7 +36,7 @@ Source0:        %{name}-%{version}.tar.xz
...
...fine main_requires git-core perl(B::Deparse) perl(Carp) perl(Carp::Always) perl(Class::Accessor::Fast) perl(Config) perl(Cpanel::JSON::XS) perl(Crypt::DES) perl(Cwd) perl(Data::Dumper) perl(Digest::MD5) perl(DynaLoader) perl(English) perl(Errno) perl(Exception::Class) perl(Exporter) perl(ExtUtils::testlib) perl(Fcntl) perl(File::Basename) perl(File::Find) perl(File::Path) perl(File::Spec) perl(File::Temp) perl(File::Touch) perl(File::Which) perl(IO::Handle) perl(IO::Scalar) perl(IO::Select) perl(IO::Socket) perl(IO::Socket::INET) perl(IO::Socket::UNIX) perl(IPC::Open3) perl(IPC::Run::Debug) perl(IPC::System::Simple) perl(Inline::Python) perl(List::MoreUtils) perl(List::Util) perl(Mojo::IOLoop::ReadWriteProcess) >= 0.26 perl(Mojo::JSON) perl(Mojo::Log) perl(Mojo::URL) perl(Mojo::UserAgent) perl(Mojolicious) >= 8.42 perl(Mojolicious::Lite) perl(Net::DBus) perl(Net::IP) perl(Net::SNMP) perl(Net::SSH2) perl(POSIX) perl(Scalar::Util) perl(Socket) perl(Socket::MsgHdr) perl(Term::ANSIColor) perl(Thread::Queue) perl(Time::HiRes) perl(Try::Tiny) perl(XML::LibXML) perl(XML::SemanticDiff) perl(autodie) perl(base) perl(constant) perl(integer) perl(strict) perl(version) perl(warnings) perl-base
```
Martchus, line 39: I think Python should only be an optional dependency.

Martchus (commented):
I'd approve it as an experimental feature we make no promises to ours users about.

dist/rpm/os-autoinst.spec
okurz: done

Martchus (approved):

okurz:
https://build.opensuse.org/package/live_build_log/devel:openQA:TestGithub:OPR-1558/os-autoinst/openSUSE_Factory/x86_64 shows that actually all packages on OBS have a success build status. Likely the OBS status reporting back is not working due to the PR being a bit older :)

Martchus (approved):

AdamWill:
Wow, this is...sure an idea! You guys did read the [description of the perl module in Pypi](https://pypi.org/project/perl/), right? "Note: This is very silly and probably shouldn’t go anywhere near production code."?

I'm probably not gonna ship `testapi.py` in the Fedora package at this point, it just seems a bit too out there...

BTW, er, are the tests really running Python inline in perl that then imports perl in Python? Has anyone called the Pimp My Ride people yet?!

Martchus:
We do not use the module from Pypi you're referencing.

AdamWill:
What `perl` module is line 18 of `testapi.py` supposed to import, then?

Martchus:
I suppose perl-Inline-Python uses XS to use make use of Python's underlying C library and when doing that it somehow makes the `perl` "module" available within the Python context to expose functions from Perl. I don't think this is a regular Python module you need to care about in term of dependencies.

okurz:
> BTW, er, are the tests really running Python inline in perl that then imports perl in Python? Has anyone called the Pimp My Ride people yet?!

yes

> edit2: oh, I see it's apparently a part of perl-Inline-Python? well, alright then. this still seems...goofy. (also I don't think our perl-Inline-Python package ships it...)

yes

> I don't think this is a regular Python module you need to care about in terms of dependencies.

correct

PR #1560, opened 2020-11-12
Remove GNU Autotools-based build system

Martchus:
The CMake-based build system should work well enough, is documented and already used within the CI and for package builds.

See https://progress.opensuse.org/issues/68684

@os-autoinst/tools-team, @AdamWill


okurz:
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages. Please also reference the corresponding progress ticket.

Martchus:
Is Fate still a thing? If not you should update your message template.

Is it sufficient to add the sentence from the PR description? It should be pretty clear that the main motivation is to avoid maintaining 2 build systems at the same time. Otherwise this PR fixes nothing. If one was annoyed by the flaws of the old build system they could already have used the new one.

okurz:
> Is Fate still a thing?

yes, in read-only :)

Please add the content from the PR description without the github mention and the progress ticket and also what this fixes for us.

kalikiana (approved):
Looks sound. I might've left the explicit mention of building with CMake but no strong opinion.

AdamWill:
No objections, I switched our package builds to CMake on the same schedule you folks did.

okurz:
should have selected "requesting change" explicitly :D

PR #1561, opened 2020-11-13
Add convenience Makefile after we removed autotools

okurz:


Martchus (changes requested):
I personally don't care much, but would actually tend to refrain from such a change:

1. This mixes `make` and `ninja` which can be confusing. I would rather add a simple `./configure` script which seems to be the more common approach (e.g. taken by the Qt project after their CMake migration).
2. This adds another level to have "targets" on. So after this change we would always need to clarify which kind of target is meant.
3. People might wrongly add targets to this Makefile instead to the actual build system. If such a change is done nevertheless, we'd end up again with the build-system within a build-system situation which adds additional complexity and potentially leads to problems.
4. The presence of a Makefile prevents doing an in-source tree build with the Makefile generator. That's nothing I would recommended but it still feels bad if something you'd generally expect to work with any CMake project is prevented in our case. Some distributions might also still use exactly that way to build, e.g. [I've recently learned](https://github.com/Martchus/syncthingtray/issues/80) that Fedora only switched to out-of-source-tree builds in version 33.
5. You're only invoking the `all` target (of the actual build system¹) here. The symlink creation which is likely required for developing is not part of that target. It would be better to call the `symlinks` target instead.

¹ which I now have to state explicitly thanks to the 2nd point.

---

So better just create a configure script which invokes CMake with some default parameters.

okurz:
> * This mixes `make` and `ninja` which can be confusing. I would rather add a simple `./configure` script which seems to be the more common approach (e.g. taken by the Qt project after their CMake migration).

A configure script would be a two step process. With the Makefile one can call just `make` and everything is done automatically, as described in the README file. If I write a configure-script then wouldn't I be "mixing shell and ninja"? We also have a Makefile for openQA

> * This adds another level to have "targets" on. So after this change we would always need to clarify which kind of target is meant.
> 
> * People might wrongly add targets to this Makefile instead to the actual build system. If such a change is done nevertheless, we'd end up again with the build-system within a build-system situation which adds additional complexity and potentially leads to problems.

This can happen just as well as anyone could add a shell script. We have a review process for that. If you like I can add a disclaimer to the Makefile

> * The presence of a Makefile prevents doing an in-source tree build with the Makefile generator. That's nothing I would recommended but it still feels bad if something you'd generally expect to work with any CMake project is prevented in our case. Some distributions might also still use exactly that way to build, e.g. [I've recently learned](https://github.com/Martchus/syncthingtray/issues/80) that Fedora only switched to out-of-source-tree builds in version 33.

It does not prevent in-source builds. The Makefile would obviously not be needed in this case as you would need to call cmake manually anyway and it would be overwritten by the version you want to use. Nothing prevented.

> * You're only invoking the `all` target (of the actual build system¹) here. The symlink creation which is likely required for developing is not part of that target. It would be better to call the `symlinks` target instead.

First, I am not only invoking the `all` target. There is `%:` which means you can call `make foo` and that would be equivalent to `(cd build && ninja foo)`. Also if `symlinks` is the target you think should be default then the existing rules are already not following a good convention which is to have "all". However, this can be easily changed and should be no reason against having a Makefile in general.

I still want this :)

Martchus:
> We have a review process for that.

From the point of a contributor it is preferable to have one clear place to add a target instead of finding out from a review comment to one has just used the wrong place.

> If you like I can add a disclaimer to the Makefile

Please do that. We should also make it clear within in the build instructions of the README that this Makefile is purely for convenience and e.g. when creating a Linux package¹ or when using an IDE with CMake integration or when any kind of customization is required one should treat os-autoinst as a regular CMake project.

> This can happen just as well as anyone could add a shell script.

A shell script is not a build system target, not sure what you mean.

> The Makefile would obviously not be needed in this case as you would need to call cmake manually anyway and it would be overwritten by the version you want to use.

Ok, I guess that will work.

> First, I am not only invoking the all target. There is %: which means you can call make foo and that would be equivalent to (cd build && ninja foo)

Then it is ok as well.

> Also if symlinks is the target you think should be default then the existing rules are already not following a good convention which is to have "all".

I made `symlinks` a different target because this target is always affecting the source directory which we shouldn't do by default. One might have multiple differently configured builds at the same time and which one to "activate" (via the `symlinks` target) should be an explicit decision. Besides, this target is really only required for local development.

---

If you add a note in the README like I've mentioned I'll be fine with it.

---

¹ for consistent packaging with all build flags introduced via e.g. RPM macros

okurz:
Updated together with mkittler as discussed

Martchus (approved):

PR #1567, opened 2020-11-15
docker: Add Dockerfile for isotovideo+qemu-x86

okurz:
I had
https://build.opensuse.org/package/view_file/home:okurz:container/isotovideo/Dockerfile
since about a year. This commit pushes that upstream to os-autoinst and
adapts to work in more general CI pipelines in a limited way, e.g. also
github actions where no nested virt and hence no KVM is available. With
this commit we can publish the container and image as an easier entry
point for users and also to be used within CI pipelines of other
projects.

The same Dockerfile configuration was used
in
https://build.opensuse.org/package/view_file/home:okurz:container/isotovideo-qemu-x86/Dockerfile?expand=1
which in turn was used in 
https://github.com/os-autoinst/os-autoinst-distri-example/pull/4
to show how isotovideo can be run as part of a CI pipeline using github
actions.

Related progress issue: https://progress.opensuse.org/issues/77905

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -78,7 +78,7 @@ Issues are tracked on https://progress.opensuse.org/projects/openqav3/.
 For an overview of the architecture, see link:doc/architecture.md[doc/architecture.md].
 
 Rules for commits
-^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~
 
 * Every commit is checked by https://travis-ci.org/travis[Travis CI] as soon as
 you create a pull request but you *should* run the os-autoinst tests locally. Checkout
@@ -98,7 +98,7 @@ Build instructions
 ------------------
 
 Installing dependencies
-^^^^^^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~~~~~~
 
 On openSUSE one can install the package
 link:https://build.opensuse.org/project/show/devel:openQA[`os-autoinst-devel`] which
@@ -109,7 +109,7 @@ within that file are specific to openSUSE but can be easily transferred to other
 distributions.)
 
 Conducting the build
-^^^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~~~
 
 Create a build directory outside of the source directory. The following commands need
 to be invoked within that directory.
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -28,12 +28,22 @@ link:http://open.qa/docs/#_development_version_repository[openQA documentation].
 
 For building os-autoinst manually checkout the build instructions below.
 
-To execute an instance of os-autoinst one needs to create a file named
-`vars.json`. It stores the values of the different variables that will configure
-the behavior of the test execution.
+The main executable `isotovideo` can read test parameters from the command
+line or read test parameters from a file named `vars.json`. This file stores
+the values of the different variables that will configure the behavior of the
+test execution.
 
-There are some variables used by os-autoinst itself and other that are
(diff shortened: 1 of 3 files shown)


okurz:
still testing in https://github.com/os-autoinst/os-autoinst-distri-example/pull/4

README.asciidoc
```diff
@@ -28,12 +28,23 @@ link:http://open.qa/docs/#_development_version_repository[openQA documentation].
 
 For building os-autoinst manually checkout the build instructions below.
 
-To execute an instance of os-autoinst one needs to create a file named
-`vars.json`. It stores the values of the different variables that will configure
-the behavior of the test execution.
+The main executable is `isotovideo` can read test parameters from the command
```
Martchus, line 31:
```suggestion
The main executable `isotovideo` can read test parameters from the command
```

README.asciidoc
```diff
@@ -28,12 +28,23 @@ link:http://open.qa/docs/#_development_version_repository[openQA documentation].
...
+line or read test parameters from a file named `vars.json`. This file stores
+the values of the different variables that will configure the behavior of the
+test execution.
 
-There are some variables used by os-autoinst itself and other that are
-used by the tests. A minimal `vars.json` file can be:
+
```
Martchus, line 36:
```suggestion
```

Martchus (changes requested):

README.asciidoc
okurz: please check again if I did what you suggested. The diff showed something for me which I could not map to the current state of text I had locally. Maybe I forgot to push an update. Please check again.

docker/isotovideo/Dockerfile.qemu-x86
```diff
@@ -0,0 +1,5 @@
+#!BuildTag: isotovideo:qemu-x86
+
+FROM opensuse/tumbleweed
+RUN zypper -n in os-autoinst qemu-x86 qemu-tools
```
ilausuch, line 4: Could you please clarify why the qemu-x86 qemu-tools packages has to be added?

okurz: yes. os-autoinst supports multiple backends. qemu is just one of them so this is why I built multiple variants in https://build.opensuse.org/project/show/home:okurz:container, e.g. isotovideo-qemu, isotovideo-qemu-kvm, isotovideo-qemu-x86, see the build tag and the filename. Do you think we can better specify that specific variant elsewhere or differently? We could add specific subpackages to https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec#L92 and install these.

ilausuch: Thanks for the explanation. Could it be possible to do it architecture independent as we did on https://github.com/os-autoinst/openQA/blob/6a2b837473d592e55b6c49e7a0a18a9d53270b59/docker/worker/Dockerfile#L12?

okurz: I see the current PR as ready to be merged but based on your comment I have also provided https://github.com/os-autoinst/os-autoinst/pull/1573 now with an alternative approach. They don't conflict and we can also change Dockerfile.qemu-x86 later if the other packages are provided.

ilausuch: Thanks, I think I'll approve both :+1:

ilausuch (approved):

okurz:
@Martchus can you approve?

Martchus (approved):

PR #1568, opened 2020-11-17
t/22-svirt: Fix sporadic failures of `run_ssh_cmd(keep_open=>0)`

cfconrad:
Ticked: poo#69346
The assumption is, that the NET:SSH2 module calls DESTROY of one of the
created objects and then reuse the same memory (see _new implementation
in SSH2.xs [1]).
The "fix" is, do not use `_new` to create the class anymore. Just
blessing a HASH. And create a unique random string to identify each
instance. This identifier is used for the key in `$ssh_obj_data`.

This change is actually the minimal impact solution I found. Because we do
not use the internal `_new()` method of NET:SSH2 anymore, but getting a blessed
HASH, we could improve this code later and drop `$ssh_obj_data` completly.

[1] https://metacpan.org/source/RKITOVER/Net-SSH2-0.71/SSH2.xs#L902

### DIFF
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -6,7 +6,6 @@ use Test::MockObject;
 use Test::Warnings ':report_warnings';
 use Test::Output;
 use Mojo::Log;
-use Scalar::Util 'refaddr';
 use XML::SemanticDiff;
 use backend::svirt;
 use distribution;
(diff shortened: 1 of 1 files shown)


t/22-svirt.t
```diff
@@ -106,6 +105,16 @@ subtest 'SSH usage in svirt' => sub {
     my $ssh_expect_credentials = {username => 'root', password => 'password'};
     my $ssh_obj_data           = {};                                             # used to store Net::SSH2 fake data per object
     my $net_ssh2               = Test::MockModule->new('Net::SSH2');
+    $net_ssh2->redefine(new => sub {
+            my ($class, %opts) = @_;
+            my $self = bless {}, $class;
+            my $id   = $self->{my_custom_id} = bmwqemu::random_string(32);
+            die("Identifier not uniqe") if (exists($ssh_obj_data->{$id}));
```
Martchus, line 112:
```suggestion
            die 'Identifier not unique' if exists $ssh_obj_data->{$id};
```
(style change and typo fix)

t/22-svirt.t
```diff
@@ -106,6 +105,16 @@ subtest 'SSH usage in svirt' => sub {
...
+            $ssh_obj_data->{$id} = {};
+            return $self;
+    });
+    sub refaddr { return shift->{my_custom_id}; }
+    $net_ssh2->mock(DESTROY => sub { });
```
Martchus, line 117:
```suggestion
    $net_ssh2->noop('DESTROY');
```

Martchus (changes requested):
Looks good besides the mentioned nitpicks.

t/22-svirt.t
cfconrad: thx

t/22-svirt.t
cfconrad: thx

Martchus (approved):

okurz (approved):

Martchus:
I can not access OBS right now (from WiFi at SUSE office) but the failure is likely unrelated (like last time). So I suppose we can force-merge.

PR #1570, opened 2020-11-23
Slightly simplify needle.pm with condensed lines

okurz:


needle.pm
```diff
@@ -354,24 +337,18 @@ sub tags {
...
-    }
+    return $goods || [] unless $goods && @wanted;
     my @results;
 
     # now check that it contains all the other tags too
   NEEDLE: for my $n (@$goods) {
-        for my $t (@wanted) {
-            next NEEDLE if (!$n->has_tag($t));
-        }
+        next NEEDLE if (!$n->has_tag($_)) for (@wanted);
```
Martchus, line 345: I'm surprised that Perl can parse this. Are you sure this works? If it does I'd still remove the braces.

okurz: who ever stated that this would work? :D https://travis-ci.org/github/os-autoinst/os-autoinst/builds/745513140 tells "no"

Martchus: At the time I was writing the CI was still pending but I assumed you would have at least run the needles test locally.

okurz: no, sorry :blush:

needle.pm
```diff
@@ -121,21 +121,13 @@ sub new {
...
     my $png = $self->{png} || $self->{name} . ".png";
 
     $self->{png} = File::Spec->catdir(dirname($jsonfile), $png);
-
-    if (!-s $self->{png}) {
-        warn "Can't find $self->{png}";
-        return;
-    }
-
+    warn "Can't find $self->{png}" && return unless -s $self->{png};
```
Martchus, line 130: Where does `warn` come from? I just wanted to cross-check whether it actually returns a truthy value all the time. Otherwise I'd refrain from depending on the return value of the function.

needle.pm
```diff
@@ -336,10 +322,7 @@ sub init {
     find({no_chdir => 1, wanted => \&wanted_, follow => 1}, $needles_dir);
     bmwqemu::diag(sprintf("loaded %d needles", scalar keys %needles));
 
-    if ($cleanuphandler) {
-        &$cleanuphandler();
-    }
-
+    &$cleanuphandler() if $cleanuphandler;
```
Martchus, line 325: If you're already simplifying:

```suggestion
    $cleanuphandler->() if $cleanuphandler;
```

needle.pm
okurz: It's Perl's core "warn" and it always returns a truthy value

Martchus (approved):

coolo:
Simplify?

Martchus:
I approved because the complex abstractions are just a very intuitive syntax in this case. However, I generally agree with the comment. Saving lines does not make the code generally easier to understand.

okurz:
Sure. I also don't think it's that important. It was just a side-commit as part of a bigger PR that had/has other problems and I wanted to trigger some CI checks with coverage analysis to check if codecov in general still works considering the current problems we have within the scope of openQA

needle.pm
perlpunk: This is actually wrong.
`warn "foo" && return` will eval the arguments for `warn` and after that call `warn`, but the `return` will simply return from the sub and the `warn` will be forgotten.
I noticed that while debugging #2603 and wondering why it didn't print the warning.

Can we be more careful with such "improvements"? Especially since this change will automatically make the line covered, so we don't see that this is actually not covered.

needle.pm
```diff
@@ -121,21 +121,13 @@ sub new {
     }
 
     # one match is mandatory
-    unless ($gotmatch) {
-        warn "$jsonfile missing match area\n";
-        return;
-    }
+    warn "$jsonfile missing match area\n" && return unless $gotmatch;
```
perlpunk, line 124: Same problem as below https://github.com/os-autoinst/os-autoinst/pull/1570/files#r1906157649

perlpunk: * #2613

PR #1571, opened 2020-11-24
Fix "Use of uninitialized value in numeric eq" on connection errors in mmapi

Martchus:
* Show the connection error instead
* See https://progress.opensuse.org/issues/80264

---

There seem to be no tests for this code so I'll either create some or resort to manual testing for now.

### DIFF
--- a/mmapi.pm
+++ b/mmapi.pm
@@ -1,4 +1,4 @@
-# Copyright (c) 2015 SUSE LLC
+# Copyright (c) 2015-2020 SUSE LLC
 #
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
+1 for doing this but changing a lot without good tests is risky. I suggest with our better experience and knowledge nowadays it should be feasible for you to create good unit tests using mocking

Martchus:
That's what I'm currently doing. Setting up a fake server is pretty easy with Mojolicious :-)

okurz (approved):
awesome!

PR #1572, opened 2020-11-25
Improve mmapi error logging and tests

Martchus:
Another small improvement in the context of https://progress.opensuse.org/issues/80264


okurz (approved):
I trust that with Test::Output you ensured there is no unhandled output

Martchus:
As you mention it, actually the entire log of the fake server is unhanded. I added another commit to make it use `note` so the log only shows up when starting prove with `-v`.

PR #1573, opened 2020-11-26
spec: Add qemu specific convenience sub-packages

okurz:
Related progress issue: https://progress.opensuse.org/issues/77905

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -107,6 +107,28 @@ Requires:       os-autoinst
 %description openvswitch
 This package contains openvswitch support for os-autoinst.
 
+%ifarch x86_64
+%package qemu-kvm
+Summary:        Convenience package providing os-autoinst+qemu-kvm
+Group:          Development/Tools/Other
+Requires:       os-autoinst
+Requires:       qemu-kvm
+Requires:       qemu-tools
+
+%description qemu-kvm
+
+%package qemu-x86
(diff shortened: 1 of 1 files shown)


ilausuch (approved):

kalikiana (approved):
I'm not sure why these should be useful, but I'm gonna guess pulling these in by default has both political and technological problems, so why not

okurz:
> pulling these in by default has both political and technological problems

What do you mean by that? Do you mean that os-autoinst should pull in qemu-x86 by default? This package is not available on all architectures. Also it is only necessary if one wants to run the qemu backend, e.g. not needed when running something like s390x z/VM or IPMI bare metal.

Martchus (approved):

okurz:
https://build.opensuse.org/package/live_build_log/devel:openQA:TestGithub:OPR-1573/os-autoinst/openSUSE_Factory/x86_64 shows the correctly built packages. The travis CI tests are still pending but also not necessary as they are not impacted by the spec file

kalikiana:
> > pulling these in by default has both political and technological problems
> 
> What do you mean by that? Do you mean that os-autoinst should pull in qemu-x86 by default? This package is not available on all architectures. Also it is only necessary if one wants to run the qemu backend, e.g. not needed when running something like s390x z/VM or IPMI bare metal.

That's exactly what I meant. Doing the right thing is impossible in practice.

okurz:
Well, I still see it as possibility that we can provide some architecture-agnostic helper package in qemu itself

PR #1574, opened 2020-11-26
Restore compatibility of mmapi::api_call

Martchus:
* Restore return value of mmapi::api_call which has changed in
  3655e0c1383457c49e8b4669c37624b23fb1c240.
* See https://progress.opensuse.org/issues/80412

### DIFF
--- a/mmapi.pm
+++ b/mmapi.pm
@@ -68,7 +68,7 @@ sub set_app {
     $ua->server->app($app = shift);
 }
 
-sub api_call {
+sub api_call_2 {
     my ($method, $action, $params, $expected_codes) = @_;
     _init                                       unless $ua;
     bmwqemu::mydie('Missing mandatory options') unless $method && $action && $ua;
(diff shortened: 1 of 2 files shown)


mmapi.pm
```diff
@@ -92,6 +92,8 @@ sub api_call {
     return ($res, $tx);
 }
 
+sub api_call { api_call_2(@_)->res }
```
okurz, line 95: please add a method description here and above.

okurz (changes requested):
Please add method descriptions

Also I suggest you roll this change out temporarily on a machine, e.g. staging and run real openQA tests, e.g. clone https://openqa.opensuse.org/tests/1484464 and maybe also a real multi-machine test

mmapi.pm
kalikiana: Ah, nice, I was just about to ask what the docs now state 👍

kalikiana (approved):

okurz (approved):

Martchus:
I've cloned the job with the failing module https://openqa.opensuse.org/tests/1484451#step/yast2_nfs_server/60 and it works now locally.

Martchus:
I would actually like to merge this before we deploy the broken version again. However, Travis didn't even enqueue the job. Hence I'm running the testsuite locally now and will merge the PR manually if it passes.

PR #1576, opened 2020-11-26
Add github actions CI definitions

okurz:
Related progress issue: https://progress.opensuse.org/issues/66721

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,5 +12,6 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
+          file: ./cover_db/codecov.json
           fail_ci_if_error: true
           verbose: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,11 @@
+---
+name: ci
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Static checks, unit- and integration tests
+        run: tools/docker_run_ci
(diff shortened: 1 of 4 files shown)


Martchus:
Unfortunately it fails. Maybe https://progress.opensuse.org/issues/66721#Patch-qemu is required?

perlpunk:
The required patch is already (commented out) in the dockerfile

okurz:
This is interesting:

```
podman run --rm -it --entrypoint '' registry.opensuse.org/devel/openqa/containers/os-autoinst_dev cat /etc/os-release
```

shows openSUSE Leap 15.1, I expected openSUSE Leap 15.2

Seems like we never properly updated the container image we use. I recorded that in
https://progress.opensuse.org/issues/80962
and will try to switch to Tumbleweed in https://github.com/os-autoinst/os-autoinst/pull/1586 but here I will first try a smaller step with going to 15.2 explicitly

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1590#issuecomment-742596504 is a proof that codecov reports are actually uploaded by GHA

Martchus (commented):
Looks good… I suppose. The quoting of `'$target'` looks a bit odd at the first look. A comment explaining the difference between codecov and codecovbash and why we sometimes need the latter would make sense.

okurz:
> The quoting of `'$target'` looks a bit odd at the first look.

I know. That's because of nested quoting. Maybe I can do better with `\"`

> A comment explaining the difference between codecov and codecovbash and why we sometimes need the latter would make sense.

Sure. I wonder where to put it, maybe in the "COMMENT" in cmake/test-targets.cmake?

Updated:

* Replaced single ticket quotes within `"` with `\"`
* Added comments explaining codecov and codecovbash in cmake/test-targets.cmake
* Added command line parsing to tools/docker_run_ci

Martchus (approved):

PR #1577, opened 2020-11-30
Handle PRODUCTDIR logic in here

Amrysliu:
This pull request is a part of https://github.com/os-autoinst/openQA/pull/3604.
Because PRODUCTDIR is a child directory of CASEDIR, we simply find
`main.pm` throught out CASEDIR.
When users trigger a job using github path as CASEDIR, the CASEDIR does
not include needles, so we need to find needles in `DEFAULT_CASEDIR`.

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -198,6 +198,7 @@ sub load_test_schedule {
     unshift @INC, $bmwqemu::vars{CASEDIR} . '/lib';
     if ($bmwqemu::vars{SCHEDULE}) {
         bmwqemu::diag 'Enforced test schedule by \'SCHEDULE\' variable in action';
+        unshift @INC, '.' unless File::Spec->file_name_is_absolute($bmwqemu::vars{CASEDIR});
         $bmwqemu::vars{INCLUDE_MODULES} = undef;
         autotest::loadtest($_ . '.pm') foreach split(',', $bmwqemu::vars{SCHEDULE});
         $bmwqemu::vars{INCLUDE_MODULES} = 'none';
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -7,6 +7,7 @@ Supported variables per backend
 |====================
 Variable;Values allowed;Default value;Explanation
 CASEDIR;string;;Path to test distribution. Can be a git repository URL of a test distribution to checkout with an optional refspec or also git hash into the current directory. It tries to follow the definition of https://docs.npmjs.com/files/package.json#git-urls-as-dependencies (e.g. `https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#feature/test` or `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git`). Note that cloning via SSH is only possible if git is configured accordingly on the host where isotovideo is executed. Can be combined with `TEST_GIT_REFSPEC` and `NEEDLES_GIT_REFSPEC`.
+DEFCASEDIR;string;;A default value of CASEDIR. When CASEDIR is a git repository URL of a test distribution, this value will be the path to test distribution. In this scenario it's used to find the correct needles directory.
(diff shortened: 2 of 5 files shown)


okurz (changes requested):
copying my suggestions from https://github.com/os-autoinst/os-autoinst/pull/1569#issuecomment-734130398 which still apply:

So what I am considering: As stated in https://github.com/os-autoinst/os-autoinst/pull/1569#discussion_r529016541 this PR is trying to fix a real problem based on an already suboptimal design of the interface between os-autoinst and openQA. Hence I created https://progress.opensuse.org/issues/80372 to solve that. But I am tending to accept this PR as a "temporary workaround" – which by nature tend to stay forever – but at least we should do the following:

* Document the variable as a "workaround" in doc/backend_vars.asciidoc
* Add the explanation from https://github.com/os-autoinst/os-autoinst/pull/1569#discussion_r529016541 in the code as comment
* Update the git commit message to describe the situation, the problem and the proposed solution

@Amrysliu can you do that?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -197,7 +197,7 @@ sub load_test_schedule {
     my @oldINC = @INC;
     unshift @INC, $bmwqemu::vars{CASEDIR} . '/lib';
     if ($bmwqemu::vars{SCHEDULE}) {
-        bmwqemu::diag 'Enforced test schedule by \'SCHEDULE\' variable in action';
```
okurz, line 200: why have you removed this message? I would prefer it to stay

isotovideo
```diff
@@ -242,7 +242,7 @@ checkout_git_repo_and_branch('CASEDIR');
 #
 # This allows further structuring the test distribution collections with
 # multiple distributions or flavors in one repository.
-$bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{CASEDIR};
+$bmwqemu::vars{PRODUCTDIR} ||= $bmwqemu::vars{DISTRI} ? $bmwqemu::vars{CASEDIR} . '/products/' . $bmwqemu::vars{DISTRI} : $bmwqemu::vars{CASEDIR};
```
okurz, line 245: I doubt this is correct. Having the variable "DISTRI" defined does not mean that we use a product dir

okurz (changes requested):
Please see my comment https://progress.opensuse.org/issues/80372#note-5 why I think it should be possible that we do not need to change anything in os-autoinst.

Another problem with changing os-autoinst to support a new openQA feature is that we would need to break backward compatibility of os-autoinst+openQA. If this is necessary at all we should be careful to introduce new support into os-autoinst, then wait some time (e.g. weeks) and only then have openQA need the newer os-autoinst. Not strictly necessary because we always deploy together but would be nice.

doc/backend_vars.asciidoc
```diff
@@ -7,6 +7,7 @@ Supported variables per backend
 |====================
 Variable;Values allowed;Default value;Explanation
 CASEDIR;string;;Path to test distribution. Can be a git repository URL of a test distribution to checkout with an optional refspec or also git hash into the current directory. It tries to follow the definition of https://docs.npmjs.com/files/package.json#git-urls-as-dependencies (e.g. `https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#feature/test` or `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git`). Note that cloning via SSH is only possible if git is configured accordingly on the host where isotovideo is executed. Can be combined with `TEST_GIT_REFSPEC` and `NEEDLES_GIT_REFSPEC`.
+DEFCASEDIR;string;;A default value of CASEDIR. When CASEDIR is a git repository URL of a test distribution, this value will be the path to test distribution. In this scenario it's used to find the correct needles directory.
```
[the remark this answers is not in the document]
Amrysliu: I haven't tested this scenario, I guess it does not work. I double if the current code does work for it without specifying NEEDLES_DIR. Because I do not change the needles_dir logic here, just search for more possible needles_dir.

Amrysliu:
See #1606

PR #1583, opened 2020-12-09
Increase sleep time after mc cold reset in case host not ready

jerrytt:
We have some hardware need more time to be ready after mc reset cold.
1 openqa ipmi backend  connecting at a wrong status of mc may cause timeout issue
2 it works well at second connecting by ipmitool manually

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -168,7 +168,7 @@ sub do_mc_reset {
         else {
             bmwqemu::diag("IPMI mc reset success!");
             # wait seconds until mc reset really sent to board
-            sleep 10;
+            sleep 120;
             bmwqemu::diag("sleep 10 ends, will do ping");
             # check until  mc reset is done and ipmi recovered
             my $count    = 0;


Martchus (changes requested):
Wouldn't it make more sense to increase the number of tries within the subsequent loop instead? Otherwise we would always wait 2 minutes (instead of just 10 seconds).

jerrytt:
> Wouldn't it make more sense to increase the number of tries within the subsequent loop instead? Otherwise we would always wait 2 minutes (instead of just 10 seconds).

problem is ping check just mean network is on.
In this case , we are facing ping check  pass , but failed to get output from sol.

backend/ipmi.pm
```diff
@@ -168,7 +168,7 @@ sub do_mc_reset {
         else {
             bmwqemu::diag("IPMI mc reset success!");
             # wait seconds until mc reset really sent to board
-            sleep 10;
+            sleep 120;
```
okurz, line 171: You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted.

We are waiting because some condition changes some time ... keyword(s) is/are *some time*. It is important to know the exact condition that the code should wait for and check for that instead of delaying program execution for an arbitrary amount of time. For openQA tests in general just try to do what you as a human user would also do. You would not look on your watch waiting for exactly 100s before you look at the screen again, right? :)

okurz (changes requested):

backend/ipmi.pm
jerrytt: Thanks @okurz .
That 's true ,I agree sleep just hide delay the problem. and this situation looks does not have a good solution.
before have better solution ,  I will keep this setting in bejing local server so that not block out testing.

okurz: If you say ping is the problem maybe you want to call more than one ping? E.g. `ping -c 3`?

Maybe it would help to make all parameters configurable and you can set your own selection? See https://github.com/os-autoinst/os-autoinst/pull/1584 for my proposal

okurz:
Please see https://github.com/os-autoinst/os-autoinst/pull/1584 for an alternative proposal

okurz:
@jerrytt we have merged https://github.com/os-autoinst/os-autoinst/pull/1584 . Feel free to provide an alternative proposal based on that or just close the PR if it's enough for you to tweak the config parameters locally. I hope this helps.

backend/ipmi.pm
jerrytt: ping is not problem,  what does ping succeed mean ?  it means  network function is work. 
but not mean redirect  serial console works right ?
the issue is we have not api from bios that mean redirect serial console function is ok to do the work.

okurz: Sure. I thought maybe the problem is that ping returns true even before MC reset is fully finished and the call to ping with multiple counts would ensure that the interface is fully up. But this is of course no guarantee, as you rightly explained. Please see my proposal in https://github.com/os-autoinst/os-autoinst/pull/1584 first. What you can also do is look into the parameters that ipmitool offers, e.g. there are some timing parameters. Other than this check for what is really expected e.g. instead of only ping check the expected ipmi commands

PR #1584, opened 2020-12-09
backend: ipmi: Make "mc reset" parameters configurable

okurz:
In some cases the connection to the IPMI interfaces is even less stable
than expected. With this change we are making all common parameters
configurable so that the test or instance admin can configure more and
be more graceful on connection problems.

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -95,6 +95,7 @@ sub restart_host {
         last if $ret =~ m/is on/;
         $self->ipmitool('chassis power on');
     }
+    1;
 }
 
 sub do_start_vm {
@@ -159,7 +160,7 @@ sub do_mc_reset {
     local $SIG{__DIE__} = {};
 
     # mc reset cmd should return immediately, try maximum 5 times to ensure cmd executed
-    my $max_tries = 5;
+    my $max_tries = $bmwqemu::vars{IPMI_MC_RESET_MAX_TRIES} // 5;
     for (1 .. $max_tries) {
         eval { $self->ipmitool("mc reset cold"); };
         if ($@) {
@@ -168,7 +169,7 @@ sub do_mc_reset {
         else {
             bmwqemu::diag("IPMI mc reset success!");
             # wait some time until mc reset really sent to board
-            $bmwqemu::vars{IPMI_MC_RESET_SLEEP_TIME_S} // 10;
+            sleep $bmwqemu::vars{IPMI_MC_RESET_SLEEP_TIME_S} // 10;
             bmwqemu::diag("sleep ends, will do ping");
             # check until  mc reset is done and ipmi recovered
             my $count      = 0;
@@ -195,5 +196,3 @@ sub do_mc_reset {
 
     die "IPMI mc reset failure after $max_tries tries! Exit...";
 }
-
-1;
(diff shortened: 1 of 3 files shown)


backend/ipmi.pm
```diff
@@ -167,18 +167,20 @@ sub do_mc_reset {
...
+            $bmwqemu::vars{IPMI_MC_RESET_SLEEP_TIME_S} // 10;
+            bmwqemu::diag("sleep ends, will do ping");
             # check until  mc reset is done and ipmi recovered
-            my $count    = 0;
-            my $timeout  = 60;
-            my $ping_cmd = "ping -c1 " . $bmwqemu::vars{IPMI_HOSTNAME};
+            my $count      = 0;
+            my $timeout    = $bmwqemu::vars{IPMI_MC_RESET_TIMEOUT}    // 60;
+            my $ping_count = $bmwqemu::vars{IPMI_MC_RESET_PING_COUNT} // 1;
+            my $ping_cmd   = "ping -c$ping_count " . $bmwqemu::vars{IPMI_HOSTNAME};
```
Martchus, line 177: Not really related to your change, but it would make sense to quote the host name:
```suggestion
            my $ping_cmd   = "ping -c$ping_count '$bmwqemu::vars{IPMI_HOSTNAME}'";
```

Martchus (commented):
I couldn't find any obvious mistakes. Nevertheless it would be great if the change was tested somehow.

backend/ipmi.pm
okurz: done

okurz (commented):
ok, followed your suggestions and also did tests.

okurz:
Updated: There was one typo in the doc which failed t/04-check_vars_docu.t . Fixed and verified locally. We should not need to wait for travis CI if it takes very long.

backend/ipmi.pm
```diff
@@ -95,6 +95,7 @@ sub restart_host {
         last if $ret =~ m/is on/;
         $self->ipmitool('chassis power on');
     }
+    1;
```
Martchus, line 98: Is this just a leftover?

Martchus (commented):
Looks good, if the `1;` is just a leftover and removed, consider it approved.

backend/ipmi.pm
okurz: Actually I thought every good perl module should end with a success status but I could remove that without a failing test so I keep that change out of this PR. No problem.

Martchus: But this is the middle of a function, not the end of the module :-)

backend/ipmi.pm
```diff
@@ -193,5 +196,3 @@ sub do_mc_reset {
 
     die "IPMI mc reset failure after $max_tries tries! Exit...";
 }
-
-1;
```
Martchus, line 197: Likely this shouldn't have been removed.

backend/ipmi.pm
okurz: oh, wow. Well I have removed it because I misread your comment. That was stupid of me.

okurz: https://github.com/os-autoinst/os-autoinst/pull/1587

By the way, because obviously we should not be blocked by hours-long waiting time for travis CI I continued with https://github.com/os-autoinst/os-autoinst/pull/1576

PR #1594, opened 2020-12-11
localXvnc: Simplify error handling

okurz:
See https://github.com/os-autoinst/os-autoinst/pull/1411/files/94d698b595b050ece94caaf22c69de78ff242e04#diff-0d5bcd4a55204a29679ead9f0e0c43aa

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -16,6 +16,8 @@ use File::Path 'mkpath';
 use File::Which;
 use Time::Seconds;
 
+our $xterm_vt = 'xterm-console';
+
 # helper function
 # Keep ssh session for the maximum of ServerAliveCountMax x ServerAliveInterval seconds
 # even without receiving any message back from the server, and this will not affect normal
@@ -33,8 +35,8 @@ sub sshCommand ($self, $username, $host, $gui = undef) {
 sub callxterm ($self, $command, $window_name) {
     my $display = $self->{DISPLAY};
     $command = "TERM=xterm $command";
-    my $xterm_vt_cmd = which "xterm-console";
-    die "Missing 'xterm-console'" unless $xterm_vt_cmd;
+    my $xterm_vt_cmd = which $xterm_vt;
+    die "Missing '$xterm_vt'" unless $xterm_vt_cmd;
     die('Missing "Xvnc"') unless which('Xvnc');
     die('Missing "icewm"') unless which('icewm');
     die('Missing "xterm"') unless which('xterm');
@@ -42,8 +44,7 @@ sub callxterm ($self, $command, $window_name) {
         mkpath 'ulogs';
         $command = "script -f ulogs/hardware-console-log.txt -c \"$command\"";
     }
-    eval { system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\""); };
-    die "cant' start xterm on $display (err: $! retval: $?)" if $@;
+    system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\"");
 }
 
 sub fullscreen ($self, $args) {
--- a/t/33-consoles-localxvnc.t
+++ b/t/33-consoles-localxvnc.t
@@ -1,0 +1,27 @@
+#!/usr/bin/perl
+
+use Test::Most;
+use Test::Warnings qw(:all :report_warnings);
+use Test::Output;
+
+use consoles::localXvnc;
+
+$consoles::localXvnc::xterm_vt = 'true';
+ok my $c = consoles::localXvnc->new(), 'object can be instanciated';
+like $c->sshCommand('my_user', 'my_host'), qr/ssh.*my_user\@my_host/, 'sshCommand returns valid command string';
+$c->{DISPLAY} = ':42';
+$consoles::localXvnc::xterm_vt = 'xterm-console';
+#my $ret;
+#my ($stdout, $stderr) = output_from { $ret = $c->callxterm('true', 'my_window_name') };
+#print "stdout: $stdout\n";
+#print "stderr: $stderr\n";
+# TODO this fails being unable to find the error string but instead states
+# that "xterm PID is …" is found. Why does the output differ depending on how
+# I call the stuff?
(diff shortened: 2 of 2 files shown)


Martchus (commented):
Codecov says the changed line isn't tested. Maybe we should add tests at least for the error case.

okurz:
hm, isn't `$@` set by eval? But ok, we can try to come up with more tests.

Martchus:
> hm, isn't $@ set by eval?

Likely the thinking was that system doesn't throw an exception if the command just exists with a non-zero exit code and one would usually check `$?` for that. We're using autodie here so I'm not sure how it behaves in this case.

okurz:
oh, I understand now. Of course we would not except an error from system I guess. Well, ok, let's see if I can come up with a test for that and we might find out that we can just remove that :)

consoles/localXvnc.pm
```diff
@@ -55,7 +55,7 @@ sub callxterm {
     die('Missing "icewm"')        unless which('icewm');
     die('Missing "xterm"')        unless which('xterm');
     eval { system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\""); };
-    die "cant' start xterm on $display (err: $! retval: $?)" if $@;
+    die "cant' start xterm on $display (err: $! retval: $?, $@)" if $@;
```
Martchus, line 58: You could also fix the

okurz:
I now got rid of the die-line completely as I think we do not need it at all. Tested manually within a perl debug session. For example specifying an invalid display number like ":23523" causing xterm to throw the error "xterm: Xt error: Can't open display: :26236" which however does not trigger `system` to throw an exception so the behaviour in this regard is unchanged.

Martchus (changes requested):
This now lacks proper error handling. As mentioned in previous comments, the check for `$@` was not that useful but `$!` should be checked.

Note that xterm's output is not sufficient, e.g. there can still be other errors:

```
  DB<7> system('xterm2 foo bar');
  DB<8> die "cant' start xterm on $display (err: $! retval: $?)" if $!;
cant' start xterm on  (err: No such file or directory retval: -1) at (eval 14)[/usr/lib/perl5/5.32.1/perl5db.pl:738] line 2.
```

(Not sure whether we should actually die here, though.)

okurz:
> This now lacks proper error handling. As mentioned in previous comments, the check for `$@` was not that useful but `$!` should be checked.
> 
> Note that xterm's output is not sufficient, e.g. there can still be other errors:
> 
> ```
>   DB<7> system('xterm2 foo bar');
>   DB<8> die "cant' start xterm on $display (err: $! retval: $?)" if $!;
> cant' start xterm on  (err: No such file or directory retval: -1) at (eval 14)[/usr/lib/perl5/5.32.1/perl5db.pl:738] line 2.
> ```
> 
> (Not sure whether we should actually die here, though.)

If you find a better error example then maybe we would still need the error handling but we already ensure that the command is callable in the lines in before with `which`. This would already die if the command either points to an invalid application or if it's not actually executable.

Martchus:
I still think the correct way is to handle the error, e.g. the file might be removed or no longer be executable after the `which` call.

I also found that catching en exception is required after all, e.g. to handle the case when the binary is corrupted:
```
  DB<12> system('xterm2 foo bar');
/usr/bin/xterm2: Zeile 1: foo: Kommando nicht gefunden.
```

or if a shared library is missing:

```
  DB<13> system('xterm2 foo bar');
xterm2: error while loading shared libraries: libc++utilities-git.so.5: cannot open shared object file: No such file or directory
```

okurz:
> I still think the correct way is to handle the error

effectively we still "handle the error", just not explicitly in this step. Right now it's broken because

```
eval { system('xterm2 foo bar') };
die "error caught" if $@;
```

likely never triggers the "die", as @perlpunk correctly noted in https://github.com/os-autoinst/os-autoinst/pull/1411/files/94d698b595b050ece94caaf22c69de78ff242e04#diff-0d5bcd4a55204a29679ead9f0e0c43aa . I am simply removing the effectively "dead code". Do we want to change the behaviour? Then I would just replace `if $@` with `if $?`, how about that?

Martchus:
> I am simply removing the effectively "dead code".

As stated in my last comment, this is not "dead code". It runs if the binary has been corrupted, a library is missing or possibly for further reasons like exceeding the max. number of processes.

The `$?` is a different case. All cases it would catch *might* be covered by the `which`-check - although it might not cover all and the file might be removed after the `which`-check. I consider the `$?` less important as the cases when it can happen despite the `which`-check are a bit contrived (but still possible).

okurz:
> > I am simply removing the effectively "dead code".
> 
> As stated in my last comment, this is not "dead code". It runs if the binary has been corrupted, a library is missing or possibly for further reasons like exceeding the max. number of processes.

I don't think we understand each other. If the binary has been corrupted then `system` would return with a non-zero, i.e. "false" result, not die so the `if $@` would not trigger, isn't it?

Martchus:
In my tests with the Perl debugger it dies if one runs `use autodie;` before.

okurz:
> In my tests with the Perl debugger it dies if one runs `use autodie;` before.

ah, right! I missed that part in my manual debugging session. Embarrasing considering I introduced "autodie" ;)

Martchus (changes requested):
I guess ignoring the `$?` error cases is ok considering they are a bit contrived. Simply letting the exception pass through should be ok as well.

I'm only requesting changes because the test contains many out-commented lines.

t/33-consoles-localxvnc.t
```diff
@@ -0,0 +1,27 @@
...
+$consoles::localXvnc::xterm_vt = 'true';
+ok my $c = consoles::localXvnc->new(), 'object can be instanciated';
+like $c->sshCommand('my_user', 'my_host'), qr/ssh.*my_user\@my_host/, 'sshCommand returns valid command string';
+$c->{DISPLAY} = ':42';
+$consoles::localXvnc::xterm_vt = 'xterm-console';
+#my $ret;
+#my ($stdout, $stderr) = output_from { $ret = $c->callxterm('true', 'my_window_name') };
+#print "stdout: $stdout\n";
+#print "stderr: $stderr\n";
+# TODO this fails being unable to find the error string but instead states
```
okurz, line 18: Someone has an idea here?

okurz:
I give up

PR #1596, opened 2020-12-11
Separate developer-centric tests into xt/ for faster coverage analysis runs

okurz:
* Makefile: Fix call of make without argument on already existing build dir
* Separate developer-centric tests into "xt/" for faster coverage analysis runs


Martchus (approved):

perlpunk (approved):

okurz:
hm, would have expected a codecov report by now. https://github.com/os-autoinst/os-autoinst/pull/1596/checks?check_run_id=1538703235 looks like it uploads a report successfully. Maybe the feedback into github never worked in GHA.

PR #1599, opened 2020-12-14
Add generic serial terminal over SSH

mdoucha:
Quite suprisingly, we still don't have a real serial terminal over SSH. Just a lot of VNC-over-SSH tunnels. Having a real serial terminal would significantly speed up tests on baremetal, s390 and SPVM.

This is an implementation of generic text-based SSH serial terminal that'll work with any SUT which runs an SSH server accessible (via network) from the worker instance. One minor limitation is that commands longer than ~2000 characters will not be echoed by SSH in full which will cause `script_run()` errors.

Note that this serial terminal could be modified to execute commands directly without sending special termination markers around but that would require redesign of the `script_run()` logic in `distribution.pm`. This would also eliminate the command length limitation mentioned above.

### DIFF
--- a/consoles/sshSerial.pm
+++ b/consoles/sshSerial.pm
@@ -1,0 +1,83 @@
+# Copyright © 2020 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+#
+#
+# Simple serial terminal over SSH
+
+package consoles::sshSerial;
+
+use strict;
+use warnings;
+
+use base 'consoles::console';
+
+use consoles::ssh_screen;
+
+sub new {
+    my ($class, $testapi_console, $args) = @_;
+
+    return $class->SUPER::new($testapi_console, $args);
+}
+
+sub screen { shift->{screen} }
+
+sub disable {
+    my ($self) = @_;
+
+    return unless $self->{ssh};
+    bmwqemu::diag("Closing SSH connection with " . $self->{ssh}->hostname);
+    $self->{ssh}->disconnect;
+    $self->{ssh} = $self->{screen} = undef;
+    return;
+}
+
+sub activate {
+    my ($self) = @_;
+    my $hostname = $self->{args}->{hostname} || die('we need a hostname to ssh to');
+    my $password = $self->{args}->{password} // $testapi::password;
+    my $username = $self->{args}->{user}     // 'root';
+    my $pty_cols = $self->{args}->{pty_cols} // 2048;
+
+    bmwqemu::diag("Connecting SSH serial console for $username\@$hostname");
+
+    my $ssh = $self->backend->new_ssh_connection(
+        hostname => $hostname,
+        password => $password,
+        username => $username
+    );
+    my $chan = $ssh->channel()
+      or $ssh->die_with_error('Cannot open SSH channel');
+
+
+    # Enable echo, no ANSI color codes, $pty_cols character line width
+    # (Sending commands longer than line width will break read-back check)
+    $chan->pty('dumb', {echo => 1}, $pty_cols)
+      or $ssh->die_with_error('PTY request failed');
+    $chan->ext_data('merge');
+    $chan->shell or $ssh->die_with_error('Failed to start remote shell');
+    $chan->blocking(0);
+
+    $self->{screen} = consoles::ssh_screen->new(
+        ssh_connection => $ssh,
+        ssh_channel    => $chan,
+        logfile        => $self->{args}->{logfile} // "serial_terminal.txt"
+    );
+    $self->{ssh} = $ssh;
+    return;
+}
+
+sub is_serial_terminal { 1 }
+
+1;
--- a/consoles/ssh_screen.pm
+++ b/consoles/ssh_screen.pm
@@ -28,6 +28,11 @@ sub new {
     croak('Missing parameter ssh_connection') unless $self->ssh_connection;
     croak('Missing parameter ssh_channel')    unless $self->ssh_channel;
 
+    if ($self->{logfile}) {
+        open($self->{loghandle}, ">>", $self->{logfile})
+          or croak('Cannot open logfile ' . $self->{logfile});
+    }
+
     return $self->SUPER::new($self->ssh_channel);
 }
 
@@ -44,6 +49,7 @@ sub do_read
         my $read = $self->ssh_channel->read($buffer, $args{max_size});
         if (defined($read)) {
             $_[1] = $buffer;
+            print {$self->{loghandle}} $buffer if $self->{loghandle};
             return $read;
         }
 
--- a/consoles/ssh_screen.pm
+++ b/consoles/ssh_screen.pm
@@ -17,9 +17,12 @@ package consoles::ssh_screen;
 
 use Mojo::Base 'consoles::serial_screen';
 use Carp 'croak';
+use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
 
 has ssh_connection => undef;
 has ssh_channel    => undef;
+
+use constant TYPE_STRING_TIMEOUT => 60;
 
 sub new {
     my $class = shift;
@@ -58,4 +61,44 @@ sub do_read
     }
     return undef;
 }
+
+sub type_string {
+    my ($self, $nargs) = @_;
+
+    bmwqemu::log_call(%$nargs);
+
+    my $text           = $nargs->{text};
+    my $terminate_with = $nargs->{terminate_with} // '';
+    my $written        = 0;
+    my $stime          = consoles::serial_screen::thetime();
+
+    $text .= "\cC" if ($terminate_with eq 'ETX');
+
+    while ($written < length($text)) {
+        my $elapsed = consoles::serial_screen::elapsed($stime);
+
+        croak("type_screen(): Timed out after $elapsed seconds.")
+          if ($elapsed > TYPE_STRING_TIMEOUT);
+
+        my $chunk = $self->ssh_channel->write(substr($text, $written));
+
+        if (!defined($chunk)) {
+            my ($errcode, $errname, $errstr) = $self->ssh_connection->error;
+
+            croak "Lost SSH connection to SUT: $errstr"
+              if $errcode != LIBSSH2_ERROR_EAGAIN;
+            select(undef, undef, undef, 0.1);
+        } elsif ($chunk < 0) {
+            # Old Net::SSH2 error signaling
+            croak "Lost SSH connection to SUT"
+              if $chunk != LIBSSH2_ERROR_EAGAIN;
+            select(undef, undef, undef, 0.1);
+        } else {
+            $written += $chunk;
+        }
+    }
+
+    $self->ssh_channel->send_eof if ($terminate_with eq 'EOT');
+}
+
 1;
--- a/t/31-sshSerial.t
+++ b/t/31-sshSerial.t
@@ -1,0 +1,191 @@
+#!/usr/bin/perl
+
+use Test::Most;
+use FindBin '$Bin';
+use lib "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '20';
+use Test::MockObject;
+use Test::MockModule;
+use Test::Warnings ':report_warnings';
+use Test::Output;
+use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
+
+use consoles::sshSerial;
+
+my $eagain = [
+    LIBSSH2_ERROR_EAGAIN,
+    'LIBSSH2_ERROR_EAGAIN',
+    'Operation would block'
+];
+
+my $mock_backend = Test::MockObject->new();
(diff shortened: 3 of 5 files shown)


consoles/sshSerial.pm
```diff
@@ -0,0 +1,82 @@
+# Copyright © 2018-2019 SUSE LLC
```
okurz, line 1: Please update

consoles/sshSerial.pm
```diff
@@ -0,0 +1,82 @@
...
+    return;
+}
+
+sub activate {
+    my ($self) = @_;
+    my $hostname = $self->{args}->{hostname} || die('we need a hostname to ssh to');
+    my $password = $self->{args}->{password} // $testapi::password;
+    my $username = $self->{args}->{user}     // 'root';
+
+    bmwqemu::diag(sprintf("Connecting SSH serial console for %s@%s",
```
okurz, line 54: The sprintf should be easy to avoid:

```suggestion
    bmwqemu::diag("Connecting SSH serial console for $username@$hostname");
```

consoles/ssh_screen.pm
```diff
@@ -52,4 +59,41 @@ sub do_read
...
+            croak "Lost SSH connection to SUT"
+              if $chunk != LIBSSH2_ERROR_EAGAIN;
+            select(undef, undef, undef, 0.1);
+        } else {
+            $written += $chunk;
+        }
+    }
+
+    if ($terminate_with eq 'EOT') {
+        $self->ssh_channel->send_eof;
```
okurz, line 95: I suggest to reduce some whitespace here, e.g.


```perl
$self->ssh_channel->send_eof if $terminate eq 'EOT';
```

consoles/sshSerial.pm
```diff
@@ -0,0 +1,82 @@
...
+        password => $password,
+        username => $username
+    );
+    my $chan = $ssh->channel();
+
+    $chan->blocking(0);
+
+    # Enable echo, no ANSI color codes, 2048 character line width
+    # (Sending commands longer than line width will break read-back check)
+    $chan->pty('dumb', {echo => 1}, 2048);
```
okurz, line 68: This looks like a magic number that might be useful to make configurable. Or is this bound to ssh internals which we maybe can read out from there instead of hard coding here?

okurz (changes requested):
Great to see your contribution and you were quick! :)
Can you please look into adding according test code? before we grow to rely on this new console implementation we should have at least complete statement coverage (with mocking low-level code) to reduce the risk of regressions when we change related code in the future

consoles/ssh_screen.pm
```diff
@@ -52,4 +59,41 @@ sub do_read
     }
     return undef;
 }
+
+sub type_string
+{
```
Martchus, line 64:
```suggestion
sub type_string {
```

consoles/ssh_screen.pm
```diff
@@ -52,4 +59,41 @@ sub do_read
...
+
+        if (!defined($chunk)) {
+            my ($errcode, $errname, $errstr) = $self->ssh_connection->error;
+
+            croak "Lost SSH connection to SUT: $errstr"
+              if $errcode != LIBSSH2_ERROR_EAGAIN;
+            select(undef, undef, undef, 0.1);
+        } elsif ($chunk < 0) {
+            # Old Net::SSH2 error signaling
+            croak "Lost SSH connection to SUT"
```
Martchus, line 86: Are we actually connecting to the SUT here? This console seems to be rather generic and also from the PR title I would assume that it is intended for generic use. Note that at least when using the svrit backend there seems to be more work to do for connecting to the actual SUT (and not just the host running the VM), e.g. calling `virsh console` (see `start_serial_grab` in `svirt.pm`).

Martchus (changes requested):

consoles/ssh_screen.pm
mdoucha: Yes, this serial terminal is intended to connect directly to the SUT, unlike the existing sshVirtshSUT console which connects to the hypervisor instead and attaches to a virtual serial device.

This terminal is generic in the sense that it'll work with any backend as long as the SUT runs SSH daemon on a port accessible from the worker.

Here's an experimental run with generic SSH terminal on ZKVM: https://openqa.suse.de/tests/5175184

Martchus: Ah, ok - you're simply starting an SSH daemon within the SUT. That makes sense.

consoles/sshSerial.pm
mdoucha: Fixed.

consoles/sshSerial.pm
mdoucha: Fixed.

consoles/sshSerial.pm
mdoucha: Added `SSH_PTY_WIDTH` constant to configure this. It's not something internal to SSH, it's a setting needed to prevent the remote side from chopping off the beginning of echoed strings that are too long for the emulated terminal.

consoles/sshSerial.pm
```diff
@@ -0,0 +1,82 @@
...
+
+    bmwqemu::diag(sprintf("Connecting SSH serial console for %s@%s",
+            $username, $hostname));
+
+    my $ssh = $self->backend->new_ssh_connection(
+        hostname => $hostname,
+        password => $password,
+        username => $username
+    );
+    my $chan = $ssh->channel();
```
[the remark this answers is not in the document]
mdoucha: Fixed.

consoles/sshSerial.pm
```diff
@@ -0,0 +1,82 @@
...
+use base 'consoles::console';
+
+use consoles::ssh_screen;
+
+sub new {
+    my ($class, $testapi_console, $args) = @_;
+
+    my $self = $class->SUPER::new($testapi_console, $args);
+
+    return $self;
```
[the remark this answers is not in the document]
mdoucha: Fixed.

consoles/ssh_screen.pm
mdoucha: Fixed.

consoles/ssh_screen.pm
mdoucha: Fixed.

mdoucha:
I've fixed the code style issues and added a 60 second timeout to `type_string()` just to be safe. I'm still working on unit tests.

consoles/sshSerial.pm
```diff
@@ -0,0 +1,84 @@
...
+package consoles::sshSerial;
+
+use strict;
+use warnings;
+
+use base 'consoles::console';
+
+use consoles::ssh_screen;
+
+use constant SSH_PTY_WIDTH => 2048;
```
Martchus, line 28: Maybe @okurz had more something like
```suggestion
use constant SSH_PTY_WIDTH => $ENV{OS_AUTOINST_SSH_PTY_WIDTH} // 2048;
```
in mind?

mdoucha: Using `%ENV` to configure this is probably a bad idea. If anything, I could make it configurable through `add_console()` extra params. But otherwise most serial consoles already use PTY width hardcoded to 2048 cols:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/97fb6cf314f6e9c337b98056265eff460d3c36e6/lib/serial_terminal.pm#L124

mdoucha:
BTW, perltidy version check in GitHub CI is failing:

    3: Test command: /opt/tools/tidy "--check"
    3: Test timeout computed to be: 10000000
    3: Wrong version of perltidy. Found '20201207', expected '20201202'.Consider '--force' but results might not be consistent.
    3/6 Test #3: test-local-tidy ..................***Failed    0.20 sec

Martchus:
I'll create a PR to update the tidy version.

okurz (commented):
you can now rebase to get an updated tidy version

mdoucha:
Rebased, unit test added.

Martchus (commented):
The code looks generally good. I'm wondering whether the missing test coverage is just a problem of collecting the coverage or whether there's still so much uncovered code (despite the unit test).

mdoucha:
Sorry, I made a typo in the unit test filename and the file didn't get executed. Fixed now.

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+$mock_ssh->{error} = undef;
+
+$mock_channel->{write_limits} = [];
+$mock_channel->{write_buffer} = '';
+$mock_channel->{read_queue}   = [];
+$mock_channel->{blocking}     = 1;
+
+$mock_channel->mock(pty => sub {
+        return 1;
+});
```
okurz, line 33: you can simplify to

```suggestion
$mock_channel->mock(pty => sub { 1 });
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+
+$mock_channel->mock(ext_data => sub {
+        return 1;
+});
+
+$mock_channel->mock(shell => sub {
+        return 1;
+});
```
okurz, line 41: Alternative

```suggestion
$mock_channel->set_true('pty', 'ext_data', 'shell');
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+
+$mock_channel->mock(blocking => sub {
+        my $self = shift;
+        my $arg  = shift;
+
+        $self->{blocking} = $arg if defined($arg);
+        return $self->{blocking};
+});
```
okurz, line 49:
```suggestion
$mock_channel->mock(blocking => sub {
        my ($self, $arg) = @_;
        $self->{blocking} = $arg if defined($arg);
});
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+        return length($data);
+});
+
+$mock_channel->mock(send_eof => sub {
+        return 1;
+});
+
+$mock_ssh->mock(channel => sub {
+        return $mock_channel;
+});
```
okurz, line 90:
```suggestion
$mock_ssh->mock(channel => sub { $mock_channel });
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        $data = substr($data, 0, $limit) if defined($limit);
+        $self->{write_buffer} .= $data;
+        return length($data);
+});
+
+$mock_channel->mock(send_eof => sub {
+        return 1;
+});
```
okurz, line 86:
```suggestion
$mock_channel->mock(send_eof => sub { 1 });
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+$mock_ssh->mock(channel => sub {
+        return $mock_channel;
+});
+
+$mock_ssh->mock(blocking => sub {
+        my $self = shift;
+        my $arg  = shift;
+
+        return $mock_channel->blocking($arg);
+});
```
okurz, line 97:
```suggestion
$mock_ssh->mock(blocking => sub {
        my ($self, $arg) = @_;
        $mock_channel->blocking($arg);
});
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+
+$mock_ssh->mock(die_with_error => sub {
+        my ($self, $arg);
+
+        die $arg;
+});
+
+$mock_backend->mock(new_ssh_connection => sub {
+        return $mock_ssh;
+});
```
okurz, line 115:
```suggestion
$mock_backend->mock(new_ssh_connection => sub { $mock_ssh });
```

t/31-sshSerial.t
```diff
@@ -0,0 +1,204 @@
...
+    $mock_channel->{blocking} = 1;
+
+    my $console = consoles::sshSerial->new(undef, {hostname => 'localhost'});
+    $console->backend($mock_backend);
+    $console->activate();
+    my $screen = $console->screen();
+    my $data;
+    my $ret;
+
+    ok(!$mock_channel->{blocking});
```
okurz, line 136: please add messages as last argument in all Test::More calls, e.g.

```suggestion
    ok !$mock_channel->{blocking}, 'non-blocking by default';
```

okurz (changes requested):
Please ensure that by default the test does not show any output that is intermixed with the status line, see https://github.com/os-autoinst/os-autoinst/pull/1599/checks?check_run_id=1571630704#step:3:592 and compare to the status line of every other test module.

you can just mock the "diag" and "fctinfo" functions from bmwqemu in the test:

```perl
use Test::MockModule;
…
my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
# silence some log output for cleaner tests
$bmwqemu_mock->noop('diag', 'fctinfo');
```

As alternative use "Test::Output" (which the test already imports) to check the output against actual expectations where the log message is important.

t/31-sshSerial.t
mdoucha: Fixed.

t/31-sshSerial.t
mdoucha: I prefer the first version.

t/31-sshSerial.t
mdoucha: Fixed.

t/31-sshSerial.t
mdoucha: Fixed argument declaration. But dropping the `return` statement will break `ssh_screen::do_read()`.

t/31-sshSerial.t
mdoucha: Fixed.

t/31-sshSerial.t
mdoucha: Fixed.

t/31-sshSerial.t
mdoucha: Fixed.

t/31-sshSerial.t
mdoucha: Fixed. But with a different message.

okurz (approved):

Martchus (approved):

PR #1604, opened 2020-12-16
Optimize hadolint call with passing project dir

okurz:
By passing the root folder of the project into the container we can both
act on all Dockerfiles at once to save time as well as read a proper
hadolint rules file from the project directory instead of needing to
repeat the same ignored rules in all Dockerfiles.

Related progress issue: https://progress.opensuse.org/issues/81118


Martchus:
In the output of "ci / static-check-containers" I don't see much feedback on the processed/checked container files. Not sure whether this worked.

okurz:
https://github.com/os-autoinst/os-autoinst/runs/2131988255?check_suite_focus=true#step:3:2 is the result, just the tools script with exit code 0 as everything passed. Same as in the former way the check would have shown up the policy violations in case of problems.

Martchus:
So we don't really know that it checked anything from the log (which we could tell from a "3 of 3 tests passed"). If you think it is nevertheless good enough, feel free to merge it.

okurz:
yeah I wanted to show you a URL pointing to a previous CI check that showed failures but couldn't find any history.

PR #1605, opened 2020-12-16
t: Bump timeout in 10-virtio_terminal.t due to fail in OBS

okurz:


Martchus (approved):

Martchus:
Looks like it needed to be rebased against the tidy update.

PR #1611, opened 2021-01-06
Print the version number without serializing the state

kalikiana:
See: [poo#75232](https://progress.opensuse.org/issues/75232)

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -5,6 +5,7 @@ use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '20';
 use autodie ':all';
+use IPC::System::Simple qw(system capture);
 use Test::Output;
 use File::Basename;
 use File::Path qw(remove_tree rmtree);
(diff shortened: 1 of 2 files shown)


okurz:
As visible in https://github.com/os-autoinst/os-autoinst/pull/1611/checks?check_run_id=1656848742#step:3:572 the test already takes 48s:


```
6: [14:27:21] ./14-isotovideo.t ........................ ok    48792 ms ( 0.02 usr  0.00 sys + 45.07 cusr  2.20 csys = 47.29 CPU)
```

can you please check that you have not introduced a significant increase of runtime?

kalikiana:
> can you please check that you have not introduced a significant increase of runtime?

The new subtest adds around `500ms` which is basically the time it takes for `isotovideo` to execute anything.

But apparently the test in general could be faster, so I optimized the log handling a bit.

t/14-isotovideo.t
```diff
@@ -22,27 +23,28 @@ my $pool_dir     = "$dir/pool";
...
+    my @cmd = ($^X, "$toplevel_dir/isotovideo", '-d', $args{default_opts}, split(' ', $args{opts}));
+    note("Starting isotovideo with: @cmd");
+    capture [$args{exit_code}], @cmd;
 }
 
+subtest 'get the version number' => sub {
+    # Make sure we're in a folder we can't write to, no base_state.json should be created here
+    chdir('/');
+    combined_like { system $^X, "$toplevel_dir/isotovideo", '--version' } qr/Current version is.+\[interface v20\]/, 'version printed';
+    ok(!-e bmwqemu::STATE_FILE, 'no state file was written');
```
Amrysliu, line 41: I do not understand this test code clearly. is it used to prove that we don't write any message to `STATE_FILE ` file?  I guess we also need to check $fatal_error will not be filed with messages when running `--version`

Amrysliu (commented):
The changes of `$SIG{__DLE__}` looks good to me. Please check the CI check, it fails at 14-isotovide.t

t/14-isotovideo.t
kalikiana: Yes. If `$fatal_error` is set, the file is written. So I'm checking that it's not written. Which means there's no error.

Amrysliu (approved):
LGTM

Martchus:
Thanks, this was really annoying. Considering my local worker log it seems like it worked :+1:

PR #1613, opened 2021-01-14
Use Mojolicious::Routes::Route::any instead of deprecated …::route

Martchus:
Otherwise
```
Mojolicious::Routes::Route::route is DEPRECATED in favor of Mojolicious::Routes::Route::any at …/commands.pm line 253.
```
shows up in the log.

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -5,7 +5,7 @@ use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '20';
 use autodie ':all';
-use IPC::System::Simple qw(system capture);
+use IPC::System::Simple qw(system);
 use Test::Output;
 use File::Basename;
 use File::Path qw(remove_tree rmtree);
@@ -30,8 +30,13 @@ sub isotovideo {
     $args{opts}         //= '';
     $args{exit_code}    //= 1;
     my @cmd = ($^X, "$toplevel_dir/isotovideo", '-d', $args{default_opts}, split(' ', $args{opts}));
-    note("Starting isotovideo with: @cmd");
-    capture [$args{exit_code}], @cmd;
+    note "Starting isotovideo with: @cmd";
+    my $output = qx(@cmd);
+    my $res    = $?;
+    return fail 'failed to execute isotovideo: ' . $!         if $res == -1;    # uncoverable statement
+    return fail 'isotovideo died with signal ' . ($res & 127) if $res & 127;    # uncoverable statement
+    is $res >> 8, $args{exit_code}, 'isotovideo exit code';
+    return $output;
 }
 
 subtest 'get the version number' => sub {
--- a/commands.pm
+++ b/commands.pm
@@ -250,7 +250,7 @@ sub run_daemon {
 
     my $r = app->routes;
     $r->namespaces(['OpenQA']);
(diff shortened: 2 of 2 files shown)


Martchus:
By the way, I've tested this manually and features depending on that routing still work.

perlpunk (approved):

Martchus:
Not sure why the tests fail:

```
6: #   Failed test 'No tests run for subtest "load test success when casedir and productdir are relative path"'
6: #   at ./14-isotovideo.t line 200.
6: "/usr/bin/perl" failed to start: "No child processes" at ./14-isotovideo.t line 34.
```

The failure seems unrelated and I can not reproduce it locally. Unfortunately the log archive doesn't contain more verbose output.

okurz:
#1614 only failed in tidy but not in the same test as here where we have:

```
6:     # No tests run!
6: 
6: #   Failed test 'No tests run for subtest "load test success when casedir and productdir are relative path"'
6: #   at ./14-isotovideo.t line 200.
6: "/usr/bin/perl" failed to start: "No child processes" at ./14-isotovideo.t line 34.
6: # Tests were run but no plan was declared and done_testing() was not seen.
6: # Looks like your test exited with 255 just after 10.
6: [09:58:13] ./14-isotovideo.t ........................ 
6: Dubious, test returned 255 (wstat 65280, 0xff00)
6: Failed 1/10 subtests 
```

Martchus:
Seems like it is just a matter of luck. After pushing the tidy changes (which should not affect anything) I get the same error on #1614:

```
 6: Test Summary Report
6: -------------------
6: ./14-isotovideo.t                      (Wstat: 65280 Tests: 10 Failed: 1)
6:   Failed test:  10
6:   Non-zero exit status: 255
6:   Parse errors: No plan found in TAP output
6: Files=36, Tests=924, 583 wallclock secs ( 0.74 usr  0.08 sys + 429.06 cusr 26.54 csys = 456.42 CPU)
6: Result: FAIL
6/6 Test #6: test-perl-testsuite ..............***Failed  583.28 sec
```

Martchus:
It seems I was just very lucky when running the test locally. It actually fails there as well most of the time. I hope my fix makes the CI pass as well. See the commit message for more details.

t/14-isotovideo.t
```diff
@@ -30,8 +30,13 @@ sub isotovideo {
     $args{opts}         //= '';
     $args{exit_code}    //= 1;
     my @cmd = ($^X, "$toplevel_dir/isotovideo", '-d', $args{default_opts}, split(' ', $args{opts}));
-    note("Starting isotovideo with: @cmd");
-    capture [$args{exit_code}], @cmd;
+    note "Starting isotovideo with: @cmd";
+    my $output = qx(@cmd);
```
kalikiana, line 34: Any idea why `qx` would be more *reliable* than `capture`? I couldn't even repro this with your branch.

Martchus: Looks like a bug in `IPC::Run::Simple` which provides `capture`. However, I'm not really motivated to investigate bugs in `IPC::Run::Simple`, especially if using built-in functions does not even require much more code.

Martchus: And about the reproducibility: I couldn't reproduce it locally as well until today where it actually almost always failed (untils switching to qx). So I'm not sure what "variable" influences whether the bug is triggered or not.

PR #1615, opened 2021-01-19
tools/tidy: Use the perltidy version from the calling repo

Vogtinator:
tools/tidy is linked into various other repos as well, which use it for running
perltidy on the contained code. Those repos have their own cpanfile which may
specify a different version of Perl::Tidy, which should get used instead.
This is especially important if os-autoinst is always pulled from master, as
changes in its cpanfile otherwise affect other repos as well.

If the repo tools/tidy is run against does not specify a version of Perl::Tidy,
it falls back to the old behaviour.

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -49,9 +49,14 @@ if ! command -v perltidy > /dev/null 2>&1; then
 fi
 
 # cpan file is in top directory
-dir="$(dirname $(readlink -f $0))/.."
-perltidy_version_found=$(perltidy -version | sed -n '1s/^.*perltidy, v\([0-9]*\)\s*$/\1/p')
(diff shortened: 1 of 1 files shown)


Vogtinator:
@okurz With this, both repos can specify their version of Perl::Tidy independently. It seems to work fine in my tests: An old state of `os-autoinst-distri-opensuse` (old `Perl::Tidy`) with this branch of `os-autoinst` (new `Perl::Tidy`) passes checks.

Martchus (approved):
Approving since you've tested it manually. I suppose that's how it should have been working in the first place because the "quickly create another PR in the other repo" approach is kind of annoying.

okurz (commented):
+1

PR #1616, opened 2021-01-20
VNC: Improve error message on key mapping

kalikiana:
A failed assert_script_run with unicode - lookalike characters leads to a message like `No map for 'Ã'`. The error should clarify the intention and refer to all keys to provide more context. So the error should make more sense as a "reason" w/o reading os-autoinst source code.

~@ilausuch might have a ticket reference shortly~ [poo#88061](https://progress.opensuse.org/issues/88061)

The offending line was `assert_script_run("podman run -d -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa –network-alias=db --name db postgres", timeout => 600);` in the particular case. Look really hard if you don't see the problem with this command.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -738,7 +738,7 @@ sub map_and_send_key {
             next;
         }
         else {
-            die "No map for '$key'";
+            die "Key '$key' not found in map ($keys)";
         }
     }
 


okurz (changes requested):
yes, people hitting wrong keys can uncover interesting issues :) The lines you changed seem to be not covered by tests with coverage analysis. So while you are at it could you look into extending according tests? I think I can ask you as you are part of the SUSE QE Tools team, maybe I would not for a drive-by contributor :) Also, could you please provide an example how the error message would like like in the new case (in the commit message please)

consoles/VNC.pm
```diff
@@ -738,7 +738,7 @@ sub map_and_send_key {
             next;
         }
         else {
-            die "No map for '$key'";
+            die "Key '$key' not found in map ($keys)";
```
Martchus, line 741:
```suggestion
            die "No mapping for key '$key' available (full key sequence: $keys)";
```

Martchus (changes requested):
> Also, could you please provide an example how the error message would like like

I was going to ask the same because it might be misleading. At least judging by the code I suppose `$keys` does not contain the key mapping but is simply the full "key sequence". However, the error message makes it sound like this info would be the map because the brace follows the word "map" directly. Note that printing the full key map is likely a bit too verbose and maybe also not the best idea.

consoles/VNC.pm
kalikiana: There's no documentation of expected input or errors so first of all I tried to add context to message on what *is* expected regardless of input. I like your improved version!

kalikiana: To answer the question, `$keys` could be `"echo test"` or one out of minus/tab/res/backspace/esc.

PR #1618, opened 2021-01-20
tools: Update path to Tumbleweed based new dev image

okurz:
The OBS project changed the path under which we can get the new
development image for CI jobs.

Related progress issue: https://progress.opensuse.org/issues/80962


okurz:
seems like project gets stuck. Maybe I need to let it run into some hour-long timeout to realize what's going on

Martchus (approved):
Looks like it works now :-)

PR #1619, opened 2021-02-08
Handle '&' terminator in "script_run"; Add new test API function "enter_cmd"

okurz:
See individual commits

Related progress issue: https://progress.opensuse.org/issues/88452

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -1,4 +1,4 @@
-# Copyright © 2018-2019 SUSE LLC
+# Copyright © 2018-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 20;
+our $version = 21;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 21;
+our $version = 22;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/distribution.pm
+++ b/distribution.pm
@@ -114,9 +114,11 @@ Deprecated mode
 
   script_run($program, [$timeout])
 
-Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
-hashed command to serial line and wait for it in order to detect execution is finished.
-To avoid waiting, use I<$timeout> 0.
+Run I<$cmd> (by assuming the console prompt and typing the command). After
+that, echo hashed command to serial line and wait for it in order to detect
+execution is finished. To avoid waiting, use I<$timeout> 0. The C<script_run>
+command string must not be terminated with '&' otherwise an exception is
+thrown.
 
 Use C<output> to add a description or a comment of the $cmd.
 
@@ -141,9 +143,10 @@ sub script_run {
     }
     testapi::type_string "$cmd";
     if ($args{timeout} > 0) {
-        my $terminator = ((substr $cmd, -1) eq '&') ? '' : ';';
-        my $str        = testapi::hashed_string("SR" . $cmd . $args{timeout});
-        my $marker     = "$terminator echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
+        die "Terminator '&' found in script_run call. script_run can not check script success. Consider using 'type_string'."
+          if $cmd =~ qr/(?<!\\)&$/;
+        my $str    = testapi::hashed_string("SR" . $cmd . $args{timeout});
+        my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
             testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
--- a/distribution.pm
+++ b/distribution.pm
@@ -141,8 +141,9 @@ sub script_run {
     }
     testapi::type_string "$cmd";
     if ($args{timeout} > 0) {
-        my $str    = testapi::hashed_string("SR" . $cmd . $args{timeout});
-        my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
+        my $terminator = ((substr $cmd, -1) eq '&') ? '' : ';';
+        my $str        = testapi::hashed_string("SR" . $cmd . $args{timeout});
+        my $marker     = "$terminator echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
             testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
--- a/testapi.pm
+++ b/testapi.pm
@@ -44,6 +44,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   get_var get_required_var check_var set_var get_var_array check_var_array autoinst_url
 
   send_key send_key_until_needlematch type_string type_password
+  enter_cmd
   hold_key release_key
 
   assert_screen check_screen assert_and_dclick save_screenshot
@@ -87,6 +88,7 @@ sub send_key;
 sub check_screen;
 sub type_string;
 sub type_password;
+sub enter_cmd;
 
 
 =head1 introduction
@@ -1467,7 +1469,7 @@ A convenience wrapper around C<type_string>, which doesn't log the string.
 
 Uses C<$testapi::password> if no string is given.
 
-You can pass same optional parameters as for C<type_string> function.
+You can pass the same optional parameters as for C<type_string> function.
 
 =cut
 
@@ -1475,6 +1477,22 @@ sub type_password {
     my ($string, %args) = @_;
     $string //= $password;
     type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
+}
+
+=head2 enter_cmd
+
+  enter_cmd($string [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, secret => 1 ]
+  [, timeout => <num>] [, similarity_level => <num>] );
+
+A convenience wrapper around C<type_string>, that adds a linefeed to execute a
+command within a command line prompt.
+
+You can pass the same optional parameters as for C<type_string> function.
+
+=cut
+
+sub enter_cmd {
+    type_string shift, lf => 1, @_;
 }
 
 =head1 mouse support
--- a/testapi.pm
+++ b/testapi.pm
@@ -1383,7 +1383,7 @@ sub send_key_until_needlematch {
 =head2 type_string
 
   type_string($string [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, secret => 1 ]
-  [, timeout => <num>] [, similarity_level => <num>] );
+  [, timeout => <num>] [, similarity_level => <num>] [, lf => 1 ]);
 
 send a string of characters, mapping them to appropriate key names as necessary
 
@@ -1404,6 +1404,9 @@ C<similarity_level> can be passed as argument for wrapped C<wait_still_screen> c
 
 C<secret (bool)> suppresses logging of the actual string typed.
 
+C<lf (bool)> finishes the string with an additional line feed, for example to
+enter a command line.
+
 =cut
 
 sub type_string {
@@ -1417,6 +1420,7 @@ sub type_string {
         %args = @_;
     }
     my $log = $args{secret} ? 'SECRET STRING' : $string;
+    $string .= "\n" if $args{lf};
 
     if (is_serial_terminal) {
         bmwqemu::log_call(text => $log, %args);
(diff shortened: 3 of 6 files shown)


distribution.pm
```diff
@@ -141,8 +141,9 @@ sub script_run {
     }
     testapi::type_string "$cmd";
     if ($args{timeout} > 0) {
-        my $str    = testapi::hashed_string("SR" . $cmd . $args{timeout});
-        my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
+        my $terminator = ((substr $cmd, -1) eq '&') ? '' : ';';
```
Martchus, line 144: 1. This breaks if the command is `foo & ` (note the space in the end which could also be a tab or similar).
2. This also breaks if the `&` is escaped, e.g. in `cat foo\&` where the `&` is part of the argument.
3. This breaks if the command ends with a comment although this was broken before of course.

Maybe there are more problems with this. Considering that having code which modifies a shell script is messy anyways I suppose this is kind of ok, though as it does not make things worse.

distribution.pm
```diff
@@ -141,8 +141,9 @@ sub script_run {
...
+        my $str        = testapi::hashed_string("SR" . $cmd . $args{timeout});
```
Martchus, line 145:
```suggestion
        my $str        = testapi::hashed_string("SR${cmd}$args{timeout}");
```

coolo:
I can't say that's a valid use of assert_script_run - the script isn't run and it's result isn't asserted.

okurz:
> I can't say that's a valid use of assert_script_run - the script isn't run and it's result isn't asserted.

yes, that is what I wonder myself if `type_string` is not just enough as I suggested in the ticket.

grisu48 (approved):
Thanks Oliver! Great work :rocket: 

I would argue that it is still valid for `script_run`. If the user decides to use `&` then it's known what it does. One could argue about `assert_script_run` though. Would `type_string` work with a serial terminal?

coolo:
Sure. a serial terminal is still a terminal.

grisu48:
Perhaps we focus on `script_run` and leave `assert_script_run` aside for a moment. I would still argue that the following should be possible to do

    script_run('wget ... &');
    script_run('wget ... &');
    script_run('wait');

Without this PR this would result in a syntax error in openQA, which should IMHO not happen.
The whole idea was to make the life of new openQA devs a bit easier by preventing those syntax errors, which might be confusing in the beginning.

coolo:
But script_run is defined to return the exit code of said script. So I can see the value of using script_run for the clearness of the code, but then we turn script_run into a bash interpreter. Because all you demand here is type_string and an enter. But script_run is actually (serial terminal or not) extending the command by a lot to make sure it exited and we return the exit code.

So if we want to support the testers trying the above, we should actually catch this as an error and report it properly.

okurz:
> So if we want to support the testers trying the above, we should actually catch this as an error and report it properly.

Yes, that is the alternative I thought about. So how about just detecting the `&` in any script_run calls and terminate with a clear message and suggesting to use `type_string` instead?

grisu48:
> But script_run is defined to return the exit code of said script. So I can see the value of using script_run for the clearness of the code, but then we turn script_run into a bash interpreter. Because all you demand here is type_string and an enter. But script_run is actually (serial terminal or not) extending the command by a lot to make sure it exited and we return the exit code.
> 
> So if we want to support the testers trying the above, we should actually catch this as an error and report it properly.

I get your point. From the perspective of openQA it makes more sense to use `type_string` in this case. However from the perspective of a openQA dev it is not clear, why a background job should not work with `script_run` or `assert_script_run`.
While it is true, that in such a case the extended functionality of those subroutines is not needed, this is how most people write test cases. For someone how has a deep knowledge about the openQA internals this might sound silly.
But from the superficial view that most of the devs have, it appears to make a simple task of running a background job unnecessarily complicated. If all that I want to do is to run simple bash command in the background, why do I have to use a different subroutine than what is used in 99% of the time?

https://github.com/os-autoinst/os-autoinst/pull/1619/files#r572146171 illustrates that there are unintended side-effects that need to be considered. This would make the suggestion of detecting the `&` and print a warning the better solution.
But if this would happen solemnly for the purpose of "this is not how script_run is intended to be used" it would feel like solving the issue only half way, right?

okurz:
> [#1619 (files)](https://github.com/os-autoinst/os-autoinst/pull/1619/files#r572146171) illustrates that there are unintended side-effects that need to be considered. This would make the suggestion of detecting the `&` and print a warning the better solution.
> But if this would happen solemnly for the purpose of "this is not how script_run is intended to be used" it would feel like solving the issue only half way, right?

The problem is that if we allow more we invite for the next test writers to expect even more and more until it becomes very confusing.

If you say that you never used `type_string` then maybe `script_run` already allows too much?

There is also another thing. `script_run $cmd, 0;` effectively also does the same as `type_string "$cmd\n";` so, how about we provide the explicit error message to `script_run` pointing to that parameter and explaining in the doc-strings?

coolo:
I personally find the need to construct the final line feed with type_string a little problematic. Perhaps we ease that pain by allowing `final_lf => true` there?

coolo:
```
coolo@e138#os-autoinst-distri-opensuse>git grep type_string tests | wc -l
943
coolo@e138#os-autoinst-distri-opensuse>git grep type_string tests | grep '\n"' | wc -l
554
```

More than half the cases would benefit - we might even make it default :)

Martchus:
> we might even make it default

But that would be a breaking change and the current behavior might be relied upon in other test distributions.

Maybe it would be cleaner to add `type_cmd` instead of introducing another parameter?

coolo:
> > we might even make it default
> 
> But that would be a breaking change and the current behavior might be relied upon in other test distributions.
I wasn't serious.

> 
> Maybe it would be cleaner to add `type_cmd` instead of introducing another parameter?
I like that!

perlpunk:
> Maybe it would be cleaner to add `type_cmd` instead of introducing another parameter?

I was also thinking about that. I thought `type_line` or `enter_string`

coolo:
`enter_string` is the most clear suggestion here. It's not just typing the string, it's also entering. Quite a nuance, but I think documentation can fix that

okurz:
I think this is a bad idea. Add was proven by @grisu48 test writers already don't even know so the basic test API functions. A additional parameter on a known function might work as test writers might look into the doc if something does not work as expected

coolo:
if they don't know the function, they can still go back to type_string with a \n - but the documentation of the functions can easily reference each other instead of talking about paramaters. *some* test writers might actually read documentation :)

But the more likely case is that they copy & paste other tests, so if we'd introduce enter_string, I think it's crucial that we do mass replacements in the tests asap :)

grisu48:
> But the more likely case is that they copy & paste other tests, so if we'd introduce enter_string, I think it's crucial that we do mass replacements in the tests asap :)

I'm really sorry, but I have to quote the German poem [The Sorcerer's Apprentice](https://en.wikipedia.org/wiki/The_Sorcerer%27s_Apprentice) here :slightly_smiling_face: 

> Ach, da kommt der Meister!
> Herr, die Not ist groß!
>Die ich rief, die Geister
> werd ich nun nicht los.

Jokes aside, @coolo suggestion has some truth in it. We do copy&paste a lot or work with custom snippets, so giving good examples would be something that is much appreciated.

okurz:
Updated:
* When detecting a terminator the execution is stopped with an error pointing to "type_string"
* Added optional parameter "lf" to "type_string" which adds a line-feed, can be used like `type_string …, lf => 1`
* Added convenience function "enter_cmd" doing the same. After accepting this PR we can mass-replace uses in os-autoinst-distri-opensuse to use that function. (I considered "enter_string" highly ambiguous as "enter" could be understood as "just entering characters" which does not imply at all a newline. Also IMHO it's too similar to "type_string" so does not explain the difference. Of course I have explained the new option and function in doc strings which get extracted into the API documentation which we also render on http://open.qa/api/testapi/

distribution.pm
```diff
@@ -114,9 +114,11 @@ Deprecated mode
 
   script_run($program, [$timeout])
 
-Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
-hashed command to serial line and wait for it in order to detect execution is finished.
-To avoid waiting, use I<$timeout> 0.
+Run I<$cmd> (by assuming the console prompt and typing the command). After
+that, echo hashed command to serial line and wait for it in order to detect
+execution is finished. To avoid waiting, use I<$timeout> 0. If a terminator
+is found, for example '&', C<script_run> tries to produce a (bash) valid
```
Martchus, line 120: The `(bash)` doesn't make much sense at this position of the sentence.

distribution.pm
```diff
@@ -114,9 +114,11 @@ Deprecated mode
...
+command line but warns about this discouraged use.
```
Martchus, line 121: Judging by the code it looks more like it would thrown an exception (instead of just printing a warning).

OpenQA/Isotovideo/Interface.pm
```diff
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 20;
+our $version = 22;
```
Martchus, line 25: Why not just incrementing the version by one?

testapi.pm
```diff
@@ -1473,6 +1479,23 @@ sub type_password {
...
 
+=head2 enter_cmd
+
+  enter_cmd($string [, max_interval => <num> ] [, wait_screen_changes => <num> ] [, wait_still_screen => <num> ] [, secret => 1 ]
+  [, timeout => <num>] [, similarity_level => <num>] );
+
+A convenience wrapper around C<type_string>, that adds a linefeed to execute a
+command within a command line prompt.
+
+You can pass the same optional parameters as for C<type_string> function.
```
Martchus, line 1490: Not really important but I'd write it as:
```suggestion
Supports the same optional parameters as the C<type_string> function.
```

It would make sense to remove the parameter list here (to avoid them from becoming out-of-sync).

testapi.pm
```diff
@@ -1473,6 +1479,23 @@ sub type_password {
...
+
+=cut
+
+sub enter_cmd {
+    my ($string, %args) = @_;
+    type_string $string, lf => 1, %args;
```
Martchus, line 1496: I'd avoid creating an intermediate hash when the only goal is forwarding the arguments:
```suggestion
    type_string shift, lf => 1, @_;
```

Martchus (changes requested):

OpenQA/Isotovideo/Interface.pm
okurz: That's done in two individual commits, one for the changed handling and optional command line parameter, one for the additional function

testapi.pm
okurz: You mean remove the parameter list from the summary string? I know it's tricky to maintain but I would say we can not do better with the very verbose and manual way of handling POD. I don't want to change the existing testapi.pm all over to implement that suggestion consistently

testapi.pm
okurz: done

distribution.pm
okurz: done

distribution.pm
okurz: done

distribution.pm
```diff
@@ -141,8 +143,10 @@ sub script_run {
     }
     testapi::type_string "$cmd";
     if ($args{timeout} > 0) {
+        my $terminator = ((substr $cmd, -1) eq '&') ? '' : ';';
+        die "Terminator '&' found in script_run call. script_run can not check script success. Consider using 'type_string'." unless $terminator;
```
grisu48, line 147: At first I though we should make this a softfailure to have a smooth grace period. But this should anyways always result in a test failure, so it should be good. Only making this remark to point out, that we do not believe this has any destructive impact on existing test cases.

grisu48 (approved):
LGTM! :rocket: 

I think the next important step is to have meaningful examples so that those API functions get used in the correct way.

okurz:
> I think the next important step is to have meaningful examples so that those API functions get used in the correct way.

Yes, after this PR would be accepted I would look into converting some (or all) relevant calls in os-autoinst-distri-opensuse

distribution.pm
okurz: Yes, I believe this has not any destructive impact on existing test cases. The definition of "softfailure" from http://open.qa/docs/#_jobs is "At least one known, non-critical issue has been found." but within os-autoinst-distri-opensuse we enforce softfailure's to link to existing issue tracker entries, e.g. bug reports in bugzilla, which we would not have here.

okurz:
@Martchus I see that you have resolved the discussions but not yet approved the PR. Do you see other changes pending? I have asked others for feedback hence added the "not-ready" so your approval would not automatically trigger a merge.

distribution.pm
```diff
@@ -114,9 +114,11 @@ Deprecated mode
...
   script_run($program, [$timeout])
 
-Run I<$cmd> (by assuming the console prompt and typing the command). After that, echo
-hashed command to serial line and wait for it in order to detect execution is finished.
-To avoid waiting, use I<$timeout> 0.
+Run I<$cmd> (by assuming the console prompt and typing the command). After
+that, echo hashed command to serial line and wait for it in order to detect
+execution is finished. To avoid waiting, use I<$timeout> 0. If a terminator
+is found, for example '&', C<script_run> tries to produce a valid bash
+command line but raises an error about the not allowed use.
```
Martchus, line 121: This still seems a bit weird. The script tries to produce a valid command line and *at the same time* raises an error about it. Judging by the code the function actually only raises an error. I would just write "The command must not be terminated with '&' (or an exception is thrown).".

distribution.pm
Martchus: The detection of the termination still breaks in these three cases. I would at least fix case 2. because that would be a valid use which should not be prevented. If that's too much work I'm wondering whether it makes sense to have this kind of validation at all.

Martchus (changes requested):

distribution.pm
okurz: Yes, you are right. Fixed

okurz:
@Martchus Fixed case 2 of https://github.com/os-autoinst/os-autoinst/pull/1619#discussion_r572146171 . negative look-behind regex to the rescue :)

distribution.pm
```diff
@@ -141,8 +143,10 @@ sub script_run {
     }
     testapi::type_string "$cmd";
     if ($args{timeout} > 0) {
+        my $terminator = ((substr $cmd, -1) =~ qr/(?<!\\)&/) ? '' : ';';
```
Martchus, line 146: I don't think the `(?<!\\)` will have any effect combined with `substr $cmd, -1`. Did you test it?

Martchus (changes requested):

distribution.pm
okurz: oh dammit, I tested the regex, yes. But not the substr of course.

okurz: Fixed

Martchus (approved):

PR #1621, opened 2021-02-10
14-isovideo.t: create 'foo' branch in test git repo

AdamWill:
The test uses this branch, but doesn't create it when setting up
the repo. Somehow this seems to work most of the time - I'm not
sure how - but fails on package builds in Fedora infrastructure
(it works doing a package build locally in mock, oddly). Creating
the branch when setting up the repo seems to avoid the problem.

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -71,7 +71,7 @@ subtest 'isotovideo with custom git repo parameters specified' => sub {
     $base_state->remove       if -e $base_state;
     path('vars.json')->remove if -e 'vars.json';
     path('repo.git')->make_path;
-    my $git_init_output = qx{git init -q --bare repo.git 2>&1};
+    my $git_init_output = qx{git init -q --bare -b foo repo.git 2>&1};
     is($?, 0, 'initialized test repo') or diag explain $git_init_output;
     # Ensure the checkout folder does not exist so that git clone tries to
     # create a new checkout on every test run


AdamWill:
I'm not 100% sure this is 'right' - the test may actually be expecting something to create the branch along the way somehow - but it seems to be needed to make the test pass in Fedora infra builds, so I'm gonna have to patch it anyway, so I figured I'd submit it...

AdamWill:
Huh. So doing it *this* way fails in gtihub actions. Why, though? That seems odd.

okurz:
Might depend on the order of test module execution or in case you exclude some tests on Fedora package build?

perlpunk:
I'm wondering why the command output is not shown.
But I'm guessing the git version is too old.
I get
```
error: unknown switch `b'`
```
with git 2.26.2

perlpunk:
I think the point of the test is that the branch `foo` is *not* there. So I'm wondering how the test is failing on your side.

AdamWill:
Well I was wondering about that, but it doesn't really seem clear. It can't just be about test order execution because I'm building the *same source RPM* either locally or in Koji, and in Koji it fails but locally it succeeds.

This is what gets logged in a failed Koji build:

    2: [2021-02-10T00:01:03.458 UTC] [debug] Cloning git URL 'file:///tmp/14-isotovideo.t-L8PD/pool/repo.git' to use as test distribution
    2: sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
    2: [2021-02-10T00:01:03.469 UTC] [debug] Cloning into 'repo'...
    2:  /bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
    2:  fatal: Remote branch foo not found in upstream origin
    2: 
    2: Unable to clone Git repository 'file:///tmp/14-isotovideo.t-L8PD/pool/repo.git#foo' specified via CASEDIR (see log for details) at /builddir/build/BUILD/os-autoinst-2e2b378af50bdac175287945a8fb70ec0ecd0f35/OpenQA/Isotovideo/Utils.pm line 69.
    2: Can't locate object method "combined_from" via package "1467454: EXIT 1
    2: " (perhaps you forgot to load "1467454: EXIT 1
    2: "?) at ./14-isotovideo.t line 79.
    2: # Tests were run but no plan was declared and done_testing() was not seen.
    2: # Looks like your test exited with 1 just after 4.

which looks like we're in `checkout_git_repo_and_branch`. Hmm - I also see `--depth` in that code, which is interesting, because it reminds me of something. There is/was a similar disjoint between Koji build and local mock build before, it also involved git, and I think it had something to do with the filesystem arrangement. Ooh, yeah, there it is, in our `openqa.spec`:

    # GIT_CEILING_DIRECTORIES here avoids a case where git error handling
    # can differ when you run the build in mock and cause 16-utils-runcmd
    # to fail
    make test GIT_CEILING_DIRECTORIES="/" CHECKSTYLE=0 PROVE_ARGS='-l -r' TEST_PG_PATH=%{buildroot}/DB

I wonder if setting `GIT_CEILING_DIRECTORIES` would help here too...I'll have to see.

perlpunk:
Ah!
```
Can't locate object method "combined_from" via package "1467454: EXIT 1
```
I just tested with Test::Output 1.032 locally and get the same error message.
The problem is that the new version uses some new form of exporting, and that doesn't seem to work.
Looking into it...

perlpunk:
See #1622

AdamWill:
I think that's only a problem with *formatting the error message* though, right? If we fix that the test will still fail, it'll just fail in a more readable way?

AdamWill:
Adding `export GIT_CEILING_DIRECTORIES="/"` to the spec file did seem to fix it for me, BTW.

perlpunk:
> I think that's only a problem with _formatting the error message_ though, right? If we fix that the test will still fail, it'll just fail in a more readable way?

No, the `fatal: Remote branch foo not found in upstream origin` is expected. The test is explicitly testing something which will fail. Only the `combined_from` call failed because it is not exported any more, so the function could not be found.

AdamWill:
ooo, I see. I was missing that, d'oh. That does make some sense, then, because the other case was a similar thing, IIRC: we get a different error message with the koji filesystem layout if `GIT_CEILING_DIRECTORIES` isn't set, one the code isn't expecting.

perlpunk:
#1622 was merged, could you test it?

AdamWill:
Yeah, can do...

PR #1622, opened 2021-02-10
t: Explicitly import Test::Output functions

perlpunk:
In Test::Output version 1.032 certain functions don't seem to be exported by
default anymore.
Since explicit imports are preferrable most of the times, we can
just fix it with that.

The error message:

    Can't locate object method "combined_from" via package "12466: EXIT 1
    " (perhaps you forgot to load "12466: EXIT 1
    "?) at t/14-isotovideo.t line 79.

See also #1621

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -6,7 +6,7 @@ use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '20';
 use autodie ':all';
 use IPC::System::Simple qw(system);
-use Test::Output;
+use Test::Output qw(combined_like combined_from);
 use File::Basename;
 use File::Path qw(remove_tree rmtree);
 use Cwd 'abs_path';


Martchus (approved):

perlpunk:
Created an issue https://github.com/briandfoy/test-output/issues/8

AdamWill:
Yeah, this seems good in my test. A .src.rpm with this patch and *not* with my patch or `export GIT_CEILING_DIRECTORIES="/"` built successfully in Koji.

briandfoy:
Can you tell me which versions of Perl and Sub::Exporter you have when you see this? Looking at the Test::Output code (which I didn't write), I see that the exporting stuff hasn't changed in 15 years. I do see the issue, but I don't know why it's happening (and Sub::Exporter is way too much magic for this).

AdamWill:
oof, you actually just made me realize my test was invalid, kinda; i changed two factors. bad me. The build where we saw this failure I built on Rawhide; subsequent builds where I declared it was fixed by various things were all on f33. So that's a confounding factor I didn't account for.

I'll go check that the patch actually fixes the problem on Rawhide first...

AdamWill:
...yeah, it does. a rawhide scratch build of current git master with no patch and no export succeeded.

So, in Rawhide, we have perl-5.32.1-471.fc34 and perl-Sub-Exporter-0.987-25.fc34 .

briandfoy:
So, what had happened was.... basically I screwed up the release in 2017 by doing it from an unmerged experimental branch. The very recently release came from master without the experiment, and that's what caused your problem. I've fix that situation and pushed Test::Output 1.033, which should act the same as 1.031. Explicit imports are still good though.

The experiment was to reduce the dependency footprint of the module. Old versions used Sub::Exporter, which is a lot of accidental complexity in the wrong place (basically, lots of support for currying functions). Since the experiment has been out in the wild for years without anyone complaining means that's now the behavior. 

Again, total screw up on my part, and now I have to go fix my release tool to refuse to work if it isn't on master.

okurz:
It's awesome to see how tiny little things breaking in tests leading to this conclusions and nice collaboration :) 👍 Thank you all

PR #1623, opened 2021-02-18
Improve scoping of '$current_test' variable

okurz:
This should help with introducing a stricter programming style check.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -69,7 +69,7 @@ sub find_script {
 
 =head2 loadtest
 
-	loadtest(<string>, [ name => <string>, run_args => <OpenQA::Test::RunArgs> ]);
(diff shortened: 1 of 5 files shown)


autotest.pm
```diff
@@ -295,33 +295,6 @@ sub start_process {
     return ($process, $child);
 }
 
-
-# TODO: define use case and reintegrate
-sub prestart_hook {
```
Martchus, line 300: Aren't these functions strictly a part of the API? It looks like they are not used by our test distribution and likely also not elsewhere, though.

okurz: I think we only consider testapi.pm and *maybe* distribution.pm as part of the official, public API that should use deprecation before removal

Martchus: right

Martchus (approved):
Tidy doesn't like it but otherwise it should be fine.

PR #1624, opened 2021-02-23
Ensure we are also checking coverage in tests themselves

okurz:
We also want to know if there is dead code in our tests, e.g. by
accident.


okurz:
fails in https://github.com/os-autoinst/os-autoinst/pull/1624/checks?check_run_id=1960298358#step:3:799 with

```
Cannot read 00-compile-check-all.t: No such file or directory at /usr/lib/perl5/site_perl/5.26.1/Devel/Cover/Report/Codecovbash.pm line 76.
```

Similar problem as I had in https://github.com/os-autoinst/os-autoinst/pull/1591

PR #1625, opened 2021-02-23
cmake: Delete working directory setting where it is already default

okurz:


Martchus (commented):
I would rather keep it explicit so one knows just from reading the code in which directory the test is executed (without knowing CMake's defaults and looking into the documentation). So please keep the code self-explaining.

okurz:
But we don't spell out all other defaults. This is why it's *default*. I don't think anyone should change cmake settings without reading the documentation about cmake :) I know that cmake is very explicit and it's not a bad choice but I would not spell out defaults here. I know, this is just one line but I opened the PR to trigger your reaction before reviewing other places as well :)

Martchus:
> But we don't spell out all other defaults.

This doesn't explain why this one isn't useful. We're also not overriding a customization point here.

> I don't think anyone should change cmake settings without reading the documentation about cmake :)

This is also about understanding what's going on. Maybe one wants to run certain commands manually. And in this case it is a bit involved: The documentation says it runs in the build directory corresponding to the current source directory. So now one has to search for all occurrences where the module has been included (which could be multiple) to find the source directory of the inclusions. Then one has to read the documentation of `include` to find out how this kind of inclusion changes the source directory to deduce the corresponding build directory.

> I know, this is just one line but I opened the PR to trigger your reaction before reviewing other places as well :)

That's good because I prefer to write the code explicit rather than implicit. I know you hate verbose code very much (otherwise you wouldn't open all these "condensed rewrite PRs") but I prefer code to by read-write and not just read-only. This is important especially when it comes to build system code which is often read by people who are not that familiar with it.

okurz:
Ok, I see. That makes sense

PR #1627, opened 2021-02-25
Clarify where the path come from when init needles failed

Amrysliu:
see: https://github.com/os-autoinst/openQA/pull/3712#issuecomment-784966964

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -311,9 +311,13 @@ sub wanted_ {
 sub default_needles_dir { "$bmwqemu::vars{PRODUCTDIR}/needles" }
 
 sub init {
-    $needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
-    $needles_dir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needles_dir) unless -d $needles_dir;
-    die "needles_dir not found: $needles_dir (check vars.json?)" unless -d $needles_dir;
+    my $init_needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
+    $needles_dir = $init_needles_dir;
+    unless (-d $needles_dir) {
+        die "Can't init needles from $needles_dir" if (File::Spec->file_name_is_absolute($needles_dir));
+        $needles_dir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needles_dir);
+        die "Can't init needles from $init_needles_dir; If one doesn't specify NEEDLES_DIR, the needles will be loaded from \$PRODUCTDIR/needles firstly or $needles_dir (\$CASEDIR + $init_needles_dir), check vars.json" unless -d $needles_dir;
+    }
     $bmwqemu::vars{NEEDLES_GIT_HASH} = checkout_git_refspec($needles_dir => 'NEEDLES_GIT_REFSPEC');
 
     %needles = ();
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -441,4 +441,14 @@ subtest 'workaround property' => sub {
     is($no_workaround_needle->get_property_value("glossy"),             undef, "glossy property is a string, has no value");
 };
 
+subtest 'clarify error message when needles directory does not exist' => sub {
+    $bmwqemu::vars{CASEDIR}     = '/tmp/foo';
+    $bmwqemu::vars{PRODUCTDIR}  = '/tmp/boo/products/boo';
+    $bmwqemu::vars{NEEDLES_DIR} = undef;
+    throws_ok { needle::init } qr/Can't init needles from \/tmp\/boo\/products\/boo\/needles at.*/, 'do not combine CASEDIR when the default needles directory is an absolute path';
+
+    $bmwqemu::vars{PRODUCTDIR} = 'boo/products/boo';
+    throws_ok { needle::init } qr/Can't init needles from boo\/products\/boo\/needles;.*\/tmp\/foo\/boo\/products\/boo\/needles/, 'combine CASEDIR when the default needles directory is a relative path';
+};
+
 done_testing();


okurz (changes requested):
Can you provide a test or at least example output from manual tests?

Amrysliu:
> Can you provide a test or at least example output from manual tests?

Here is a manual test: http://10.67.19.103/tests/487#

kalikiana:
> > Can you provide a test or at least example output from manual tests?
> 
> Here is a manual test: http://10.67.19.103/tests/487#

Screenshot from 2021-02-25 10-02-10

I like it. This is actually pretty clear.

kalikiana (approved):

Amrysliu:
I did a little improvement according @kalikiana suggestions. base on: https://progress.opensuse.org/issues/88482
Verify run: http://10.67.19.103/tests/490 (default needles_dir is an absolute path)
                 http://10.67.19.103/tests/488 (default needles_dir is a relative path)

needle.pm
```diff
@@ -311,9 +311,13 @@ sub wanted_ {
...
 sub init {
-    $needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
-    $needles_dir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needles_dir) unless -d $needles_dir;
-    die "needles_dir not found: $needles_dir (check vars.json?)" unless -d $needles_dir;
+    my $default_needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
+    $needles_dir = $default_needles_dir;
+    unless (-d $needles_dir) {
+        die "Can't init needles from $needles_dir" if (File::Spec->file_name_is_absolute($needles_dir));
+        $needles_dir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needles_dir);
+        die "Can't init needles from $default_needles_dir (If user doesn't specify NEEDLES_DIR, the needles will be inited from \$PRODUCTDIR/needles firstly) or $needles_dir (\$CASEDIR + $default_needles_dir) (check vars.json?)" unless -d $needles_dir;
```
Martchus, line 319:
```suggestion
        die "Can't init needles from $default_needles_dir; if one doesn't specify NEEDLES_DIR, the needles will be loaded from \$PRODUCTDIR/needles firstly or $needles_dir (\$CASEDIR + $default_needles_dir), check vars.json" unless -d $needles_dir;
```

But this still exceeds the limit for the reason. Maybe we should just put a reference to some documentation here.

okurz: Or point to the log output with an explicit error message.

kalikiana: You can drop the mention of `vars.json` so long as the relevant variables are mentioned. Pointing to logs or files here is mostly confusing if you don't know what you're looking for e.g. it won't tell you that a magic folder is expected but missing or in the wrong place.

Amrysliu: I modified the message according @Martchus comments. 
IMHO the whole error message is shown in autoinst-log.txt, users can see the details in it even the reason is not complete. 
I keep that `vars.json` to reminder users could get the CASEDIR or PRODUCTDIR or NEEDLES_DIR from it. So I prefer to keep that message.

okurz: Yeah, I think that is ok. Could you please add a test covering the line?

needle.pm
```diff
@@ -311,9 +311,13 @@ sub wanted_ {
 sub default_needles_dir { "$bmwqemu::vars{PRODUCTDIR}/needles" }
 
 sub init {
-    $needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
-    $needles_dir = File::Spec->catdir($bmwqemu::vars{CASEDIR}, $needles_dir) unless -d $needles_dir;
-    die "needles_dir not found: $needles_dir (check vars.json?)" unless -d $needles_dir;
+    my $default_needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
+    $needles_dir = $default_needles_dir;
```
okurz, line 315: I would prefer to use a different variable name than `$default_needles_dir` because you also read the `NEEDLES_DIR` variable

needle.pm
Amrysliu: added

needle.pm
Amrysliu: Do you have any suggestion? this parameter only be used in `die` message.

okurz: Ok, sorry that I didn't provide a suggestion. How about `$needles_dir_from_settings` or `$init_needles_dir`?

t/01-test_needle.t
```diff
@@ -441,4 +441,16 @@ subtest 'workaround property' => sub {
     is($no_workaround_needle->get_property_value("glossy"),             undef, "glossy property is a string, has no value");
 };
 
+subtest 'clarify error message when needles directory does not exist' => sub {
+    $bmwqemu::vars{CASEDIR}     = '/tmp/foo';
+    $bmwqemu::vars{PRODUCTDIR}  = '/tmp/boo/products/boo';
+    $bmwqemu::vars{NEEDLES_DIR} = undef;
+    eval { needle::init };
+    like($@, qr/Can't init needles from \/tmp\/boo\/products\/boo\/needles at.*/, 'do not combine CASEDIR when the default needles directory is an absolute path');
```
okurz, line 449: You can use https://metacpan.org/pod/Test::Exception here which is already imported

Amrysliu: Thanks. fixed it.

needle.pm
Amrysliu: Thanks for your ideas. I prefer `$init_needles_dir`.

okurz (approved):
I checked the output on https://github.com/os-autoinst/os-autoinst/pull/1627/checks?check_run_id=1986435562 and all looks good.

PR #1632, opened 2021-03-11
Add background_script_run() testapi function

mdoucha:
This is a followup to PR #1619. `background_script_run()` is a more convenient way to run background processes than `enter_cmd()`. It'll automatically return the PID of your background process so that you can get its exit code later by calling `wait $pid`.

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -22,7 +22,7 @@ use warnings;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 22;
+our $version = 23;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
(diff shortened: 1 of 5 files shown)


grisu48 (approved):
LGTM :rocket: 

I especially like the approach of returning the PID, which allows interaction with the job (wait, kill, ecc.).

Martchus (commented):
I like it. Maybe it makes sense to bump the API version. Not sure whether the test failures on OBS are related.

mdoucha:
> I like it. Maybe it makes sense to bump the API version. Not sure whether the test failures on OBS are related.

Hmm, it does appear to be related to the two extra test cases in `script_run` subtest. If I remove the 3 added lines, the failures go away. But I have no idea why.

mdoucha:
OK, so just adding two more test cases creates more result boxes which threw off the result numbering in `check_assert_screen` subtest... Eww, that subtest is brittle as hell and needs to be better initialized. But that's not my problem.

okurz:
Please bump the test API version by one. The only concern I have so far is that we are introducing more *nix-specifics into the test API which make it too specific and also more complicated. You have not introduced that problem, sure, but adding to it.

mdoucha:
Version bumped.

Ideally, these system-specific details should be implemented in console backend drivers. But that would require rewriting the console subsystem from scratch.

kalikiana (approved):

PR #1633, opened 2021-03-12
Fix for select_patterns - make needle matching less forgiving

coolo:
For years we could ignore that the needle matching was just *very* colour blind - and could workaround it.

Stefan and I discussed several mechanisms how to fix this in a better way, so there might come even more in the future. But for now I would like to lift the differenianted shades of gray for 16 to 256.

Give the old snake some refreshment (https://youtu.be/vGpH4SQ9rQo?t=587)

CC @StefanBruens


Martchus (approved):
I've been following your messages in the chat in I suppose this change makes sense. Obviously it might cause some re-needling effort so I've been adding the "not-ready" label to prevent mergify from merging it automatically. Maybe it makes more sense to merge this on Monday?

It would also make sense to make the matching behavior more configurable because such changes aren't always a clear win.

coolo:
the maching behaviour is configurable - you can lower the match rate to those needles where it's not too important to be less than 95%

coolo:
The xterm one is fun - 15.2 finds a different slot than Tumbleweed (which I used to develop the test)

Martchus:
I know that the overall threshold is configurable. The idea was just to make also details like how we account colors configurable. But it is just an idea.

coolo:
gitlab actions find even a 3rd spot :)

coolo:
> I know that the overall threshold is configurable. The idea was just to make also details like how we account colors configurable. But it is just an idea.

Yeah, we have an issue somewhere where virtualization team complains that we treat an all dark blue screen as all black - for (non colour blind) humans it's hard to understand what's going on :)

We have to see how the rollout of this change works out, but I guess by now people prefer reneedling over lowering the match rate. I would bet a Kasten Bier that less than 8.3% of our SLE needles have a lowered match rate.

okurz:
> We have to see how the rollout of this change works out, but I guess by now people prefer reneedling over lowering the match rate. I would bet a Kasten Bier that less than 8.3% of our SLE needles have a lowered match rate.

I won't take the bet because I think the same. This is unfortunate. Maybe we need to make "reneedling" harder … or make changing the match rate easier :)

okurz (approved):

coolo:
Should it worry me that 15.2 caluclates slightly different? At least github and TW agree now

Martchus:
TW (and therefore the CI which is using TW) use OpenCV 4 and Leap (at least our packages) use OpenCV 3. Maybe that makes the difference. If that's true we could simply package OpenCV 4 for Leap to get rid of the difference if it is important.

coolo:
I added a round now

StefanBruens:
> TW (and therefore the CI which is using TW) use OpenCV 4 and Leap (at least our packages) use OpenCV 3. Maybe that makes the difference. If that's true we could simply package OpenCV 4 for Leap to get rid of the difference if it is important.

I somewhat doubt this is a change between OCV4 and OCV3 - the template_match code is unchanged for years. There may be changes in the underlying basic operations, but that should not effect the result much, if at all.

@coolo Where can these different results?

t/01-test_needle.t
```diff
@@ -225,15 +225,44 @@ ok(!defined $res, "the contrast is just too different");
...
+    'we find the xterm though'
+);
 
 $img1   = tinycv::read($data_dir . "pkcon-proceed-prompt-20141205.test.png");
 $needle = needle->new("pkcon-proceed-prompt-20141205.json");
-$res    = $img1->search($needle, 0, 0.7);
+($res, $cand) = $img1->search($needle, 0, 0.7);
+
+ok(!defined $res, "the prompt is the same to the human eye, but it differs in shades of gray");
+# 15.2 calculcates 92.9 and TW 92.8 - different opencv versions, different compilers...
```
okurz, line 249: The 15.2 is a bit ambiguous. Could you please rephrase? Either "openSUSE Leap 15.2" or just different OS versions.

ppmclibs/tinycv_impl.cc
```diff
@@ -90,12 +90,7 @@ double enhancedMSE(const Mat& _I1, const Mat& _I2)
         const uchar* I2_data = I2.ptr<const uchar>(j);
 
         for (int i = 0; i < I1.cols; i++) {
-            // reduce the colours to 16 before checking the diff
-            if (abs(I1_data[i] - I2_data[i]) < 16)
-                continue; // += 0
-            double t1 = round(I1_data[i] / 16.);
-            double t2 = round(I2_data[i] / 16.);
-            double diff = (t1 - t2) * 16;
+            double diff = I1_data[i] - I2_data[i];
```
StefanBruens, line 93: This calculation was always a bit strange ...
- For small differences (abs < 16), the error is set to 0
- Otherwise, the inputs are quantized and then the difference is calculated. In some cases this reduces the error (when both are rounded towards the other, e.g. 39 and 9 -> 32-16=16), in some cases the exaggerates the error (41 and 7 -> 48-0=48).

AFAIK one/the motivation for the quantization was to suppress JPEG artifacts, introduced by some VNC/RFB connections with some BMCs.

StefanBruens: With the changed calculation, this is just a plain (normalized) sum-of-squared-differences, i.e. the exact same value which is returned by template_match.

I.e. you can scrap this function altogether and just use `double mse = sse / (width * height)` @ L260.

coolo: I tried to simplify the code but ran into some overflow issues that I didn't want to continue debugging - so let's leave that to another PR.

coolo: You can see here an example of a change that was ok before and is now marked as 92% (a bit unexpected to the human eye I would say):



(animated gif with before and after frame)

StefanBruens: ~49% of the pixels have a color difference of 0, ~49% of 1..15, and the remaining 1.7% have a difference of 16..19.

Algorithmically, the changed result is well expected.

coolo: I'm not saying the algorithm is faulty - I'm trying to explain the motiviation of the old one :)

coolo: Basically if we could get an easy math for perceived difference, we could win. I tend to believe the answer lies in the HSV model

StefanBruens: The "V" in "HSV" is more or less the same as the current gray channel.

For a distance metric, HSV is quite bad - e.g. dark yellow and dark blue have the maximum distance of 180 degress, although both are perceived as "almost black". The problem is not YUV, but taking only the lightness.

A good metric:
1. suppresses large error values due to small changes, e.g. for a large monochrome background being slightly darker/lighter
2. takes into account shifts in color. I.e. also us YU channels

The old code is fairly good at (1.), but fails for (2.). The proposed change no longer supports (1.), and does not help for (2.).

coolo: It helps those cases where the shift in colour is light. But basically we're fighting a general colour blindless and so far we got pretty convincing if we defended's openQA colour blindness with it being a problem to (some) humans too.

If you upload the needle in question to https://www.color-blindness.com/coblis-color-blindness-simulator/ you'll get 
image

Now tell me: is the pattern selected or not?

coolo: this form of colour blindness affects 1 out of 30.000 people, so it's not a strong argument. The more frequent Deuteranopia is simulated as
image

which leave more contrast in the checkbox.

StefanBruens: I have no vision impairments, and in my opinion the theming for the checkboxes is just horrible. Any impairments (color weakness/blindness, far/shotsighted, ...) just make it worse.

coolo: I do not disagree.

My concern is just testers creating a needle come and say "how can this be 100% when it's clearly changing from grey to green" and so I prefer if we make such problems discoverable.

okurz:
@os-autoinst/tests-maintainer your help would be appreciated to create and submit needles for openSUSE+SLE based on this os-autoinst version where required *before* we merge this PR so that when we accept and roll out this change at best on test fails due changes in needle matching. Otherwise we would need to accept this change without the needles in a ready state. I expect some days before we do this.

ppmclibs/tinycv_impl.cc
Martchus: @StefanBruens But are you fine if we'd merge it? The new behavior might not be a good metric for measuring the perceived difference but it helps with the use-case we currently see in production.

czerw (approved):
LGTM

okurz:
Asked more common contributors of os-autoinst-distri-opensuse and they have confirmed that they are fine with the change in general and accept a slight breakage in case of need to update needles. Before accepting I would appreciate a full run of a common openSUSE openQA test for verification

SergioAtSUSE:
> Asked more common contributors of os-autoinst-distri-opensuse and they have confirmed that they are fine with the change in general and accept a slight breakage in case of need to update needles. Before accepting I would appreciate a full run of a common openSUSE openQA test for verification

These two are the most needle-problematic in functional job group:

- https://openqa.suse.de/tests/5686286#live
- https://openqa.suse.de/tests/5686287#live

SergioAtSUSE:
> > Asked more common contributors of os-autoinst-distri-opensuse and they have confirmed that they are fine with the change in general and accept a slight breakage in case of need to update needles. Before accepting I would appreciate a full run of a common openSUSE openQA test for verification
> 
> These two are the most needle-problematic in functional job group:
> 
>     * https://openqa.suse.de/tests/5686286#live
> 
>     * https://openqa.suse.de/tests/5686287#live

oh, of course those verification are not valid since this is os-autoinst, and not os-autoinst-distri-opensuse...

lemon-suse:
For migration test, I have run two jobs to verify this:

Regression test with all modules and extensions: http://10.161.8.165/tests/10#step/boot_to_desktop/49
SLED migration test: http://10.161.8.165/tests/9#step/boot_to_desktop/45

Both failed very early, for tag ' inst-bootmenu-boot-harddisk'. It means almost all migration test will be blocked here.

coolo:
Well, noone said it would work without new needles. Your needle is comparing blue with green background - and as we reduced the colourblindness, this obviously fails now.

lemon-suse:
> Well, noone said it would work without new needles. Your needle is comparing blue with green background - and as we reduced the colourblindness, this obviously fails now.

Ok, I understand it Just update the failed needle and it pass this screen now, I guess there are still some needles need to re-create in the following flow. http://10.161.8.165/tests/11#step/boot_to_desktop/2 .

okurz:
@lemon-suse and the idea is that after creating the needles locally you can provide a pull request to have these needles in the main repos before we deploy the changes from this PR into production

Also my idea is that https://github.com/os-autoinst/openQA/pull/3804 can help to ease testing of this and other os-autoinst versions in the future

lemon-suse:
> @lemon-suse and the idea is that after creating the needles locally you can provide a pull request to have these needles in the main repos before we deploy the changes from this PR into production
> 
> Also my idea is that [os-autoinst/openQA#3804](https://github.com/os-autoinst/openQA/pull/3804) can help to ease testing of this and other os-autoinst versions in the future

Sure, I'm working on this on my local environment, I will file a MR when finished and please help to review and merge. Thanks.

lemon-suse:
> > @lemon-suse and the idea is that after creating the needles locally you can provide a pull request to have these needles in the main repos before we deploy the changes from this PR into production
> > Also my idea is that [os-autoinst/openQA#3804](https://github.com/os-autoinst/openQA/pull/3804) can help to ease testing of this and other os-autoinst versions in the future
> 
> Sure, I'm working on this on my local environment, I will file a MR when finished and please help to review and merge. Thanks.

Hi Oliver and Coolo, please help to review and merge the MR https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1525 .
There 20 needles included in the MR but shouldn't include all needles affected by the PR since only have three x86 jobs tested on local environment. I think there will be some needles need to re-needle after this PR merged for the different of ARCH and more test coverage, that is expected. So would you merge the PR during RC phase? The concern is still the workload for needles. How about to do this after RC phase?

Martchus:
The SR request looks good. What about needles for openSUSE?

coolo:
We will do them as needed. The 17 needles that lemon did are just a start either :(

Martchus:
This PR has somehow been forgotten. Now some time passed and I'm wondering whether it is still needed. I also like to avoid merging it now (shortly before multiple team members are on vacation).

foursixnine:
> This PR has somehow been forgotten. Now some time passed and I'm wondering whether it is still needed. I also like to avoid merging it now (shortly before multiple team members are on vacation).

I guess it's a good time to merge this, giving a heads up to @os-autoinst/tests-maintainer since Leap and SLES have been already released.

foursixnine (approved):

okurz:
https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1525 was still open. I merged it. The argument of "now is a good time to merge because Leap and SLES have been released" would be true if people would run tests soon after we merge this PR :)

marita22:
Ok, nice, has anybody thought about maintenance?

okurz:
yes, we thought about maintenance from the beginning. Relevant reviewers have been pinged in https://github.com/os-autoinst/os-autoinst/pull/1633#issuecomment-799293002 and elsewhere at well 3 months ago. Also at least @pevik and @mimi1vx gave their "thumbs up" to https://github.com/os-autoinst/os-autoinst/pull/1633#issuecomment-855816330 which included another ping to the "tests maintainers". So I finally merged the pull request after that.

Vogtinator:
This got merged and reneedling is almost done (some maintenance tests are not complete yet).

AFAICT, this is a good opportunity to do some needle cleanup again. I suggest to wait until it's done and then delete all needles which haven't been used for a month (or so).

foursixnine:
@Vogtinator I would say a little bit more than that, since we have also the xmas needles and so on?

Vogtinator:
There aren't actually that many xmas needles (looks the same across all distros and only affects gfxboot)

coolo:
the christmas needles are reason why we have a filter for needles that were actually looked at in that time period. Because the christmas needles shouldn't be looked at in June.

PR #1635, opened 2021-03-25
Further enhancements to #1558

okurz:


autotest.pm
```diff
@@ -112,11 +112,28 @@ sub loadtest {
...
+        # Adding the include path of os-autoinst into python context
+        my $inc = File::Basename::dirname(__FILE__);
+        $code .= "
+            use base 'basetest';
+            use Mojo::File 'path';
+            use Inline Python => 'import sys; sys.path.append(\"$inc\")';
+	    #use testapi;
+	    #use mmapi;
+	    #use Inline Python => 'for i in dir(perl.testapi): locals()[i] = getattr(perl.testapi, i)';
+	    #use Inline Python => 'for i in dir(perl): for j in i: locals()[j] = getattr(i, j)';
```
Martchus, line 129: I suppose these are still a leftover.

t/data/tests/tests/pre_boot.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2019 SUSE LLC
```
Martchus, line 1: This PR has really lying around for a while now; you could update the year. The same counts for the other occurrences.

t/data/tests/tests/pre_boot.py
```diff
@@ -0,0 +1,28 @@
...
+
+import sys
+print(sys.path)
+from testapi import *
+
+def run(self):
+    send_key('esc')
+    if not check_screen('should_not_match', 0):
+        return
+    raise Exception('Should not reach here')
```
Martchus, line 24: Out of curiosity: What would happen if we'd actually thrown an exception from Python? Is the error shown appropriately?

okurz: yes, the perl code will invoke the customized die-handler

t/data/tests/tests/pre_boot.py
okurz: done

autotest.pm

Martchus:
I only haven't approved because it looks like #1558 should be merged first.

okurz:
correct. I created this separate PR as it shows with CI checks that everything is ok whereas in https://github.com/os-autoinst/os-autoinst/pull/1558 we have custom OBS checks due to container image builds in that temporary OBS project.

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1558 is now merged. I rebased to fixed the conflict.

Martchus (approved):

PR #1636, opened 2021-03-25
Various minor enhancements

okurz:
* t: Fix deprecation warning about 'ide-drive'
* Also accept test modules with file extension in SCHEDULE
* Use proper logging calls instead of bare print
* t: Simplify log parsing in 99-full-stack.t


backend/driver.pm
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2009-2013 Bernhard M. Wiedemann
+#~ Copyright © 2009-2013 Bernhard M. Wiedemann
```
Martchus, line 1: This change looks unintended but you could update the year in the line below.

doc/backend_vars.asciidoc
```diff
@@ -11,7 +11,7 @@ PRODUCTDIR;string;;Path to optional "product directory" which includes the test
 NEEDLES_DIR;string;;Path to needles subdirectory to use, defaults to "needles" within `PRODUCTDIR`. Can be a git repository URL, comparable to `CASEDIR`
 INCLUDE_MODULES;string;;comma separated names or fullnames of test modules to be included while excluding all that do not match, e.g. "boot,mod1"
 EXCLUDE_MODULES;string;;comma separated names or fullnames of test modules to exclude. Can be combined with INCLUDE_MODULES and has precedence, e.g. to additionally exclude modules based on an include-list
-SCHEDULE;string;;comma separated list of relative paths to test modules within CASEDIR without the implicit file extension '.pm' to be scheduled instead of evaluating a schedule from the test distributions main.pm file, e.g. "boot,console/mod1"
+SCHEDULE;string;;comma separated list of relative paths to test modules within CASEDIR to be scheduled instead of evaluating a schedule from the test distributions main.pm file, e.g. "boot,console/mod1". Test modules can be specified without the implicit file extension '.pm'.
```
Martchus, line 14: Shouldn't this "can" be a "must"?

Martchus (changes requested):

backend/driver.pm
okurz: done

doc/backend_vars.asciidoc
okurz: Actually both works. This is the reason I even revisited the doc string

Martchus (approved):

PR #1637, opened 2021-04-07
Prevent Inline::Python failing in downstream projects

okurz:
E.g. see
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12238/checks?check_run_id=2286115726#step:5:32
failing with

```
! Configure failed for Inline-Python-0.56. See /github/home/.cpanm/work/1617787806.1116/build.log for details.
```

I would not have expected that os-autoinst-distri-opensuse CI checks try
to install that package as was listed as only a "test" requirement in
cpanfile#L75 and the actual error why it fails to be pulled in is not
obvious. We should look into both errors but to avoid this for now this
commit is moving the dependency into the dependency list which is really
only used within the local test environment of os-autoinst.

Related progress issue: https://progress.opensuse.org/issues/90761

### DIFF
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -70,7 +70,6 @@ RUN zypper in -y -C \
        'perl(IPC::Open3)' \
        'perl(IPC::Run::Debug)' \
        'perl(IPC::System::Simple)' \
-       'perl(Inline::Python)' \
        'perl(List::MoreUtils)' \
        'perl(List::Util)' \
        'perl(Mojo::IOLoop::ReadWriteProcess)' \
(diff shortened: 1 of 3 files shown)


container/os-autoinst_dev/Dockerfile
```diff
@@ -70,7 +70,6 @@ RUN zypper in -y -C \
        'perl(IPC::Open3)' \
        'perl(IPC::Run::Debug)' \
        'perl(IPC::System::Simple)' \
-       'perl(Inline::Python)' \
```
Martchus, line 73: Don't we need it here to execute the tests?

okurz:
As visible in https://github.com/os-autoinst/os-autoinst/pull/1637/checks?check_run_id=2286251783 I got it wrong. But hopefully with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12243 this PR should not be necessary anymore at all.

PR #1640, opened 2021-04-13
signalblocker: Also block SIGCHLD

Vogtinator:
The perl-unaware threads created by OpenCV also crash when they encounter a
SIGCHLD signal, which happens by using perl's "system" function for instance.

Related ticket: https://progress.opensuse.org/issues/59926

### DIFF
--- a/signalblocker.pm
+++ b/signalblocker.pm
@@ -1,4 +1,4 @@
-# Copyright © 2020 SUSE LLC
+# Copyright © 2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


Vogtinator:
CC @AdamWill @coolo due to #1032. With this, OCR needles no longer crash `isotovideo` here.

Martchus (approved):
Thanks. When introducing the signal blocker and `tinycv::create_threads` I was already wondering why it still didn't fix all crashes. Maybe this now finally covers everything.

AdamWill:
Thanks for the heads up!

AdamWill:
Huh, fun sequel: I'm now seeing crashes that seem to be caused by a SIGFPE. Don't know what's causing the SIGFPE.

Vogtinator:
> Huh, fun sequel: I'm now seeing crashes that seem to be caused by a SIGFPE. Don't know what's causing the SIGFPE.

SIGFPE is very strange. That's for FPU exceptions and like SEGV it's delivered to the thread that caused it (or some other process sent it, but that's even stranger). If you have a way to reproduce that, it should be possible to find the cause by getting a backtrace. I guess something does a div / 0 which isn't supposed to happen.

AdamWill:
Hum, well that may be not what's happening, but it's difficult to tell. I've got a lot of these crashes. The one I was looking at there is from an x86_64 host and the traceback shows:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fbdc4ce68d0 in Perl_csighandler3 (sig=<optimized out>, sip=<optimized out>, uap=<optimized out>) at /usr/src/debug/perl-5.38.2-506.fc40.x86_64/mg.c:1599
1599	           sig == SIGFPE ||
```
note we're in the line of the perl signal handler where it checks if the signal is SIGFPE, apparently. And `sig` is optimized out, which is irritating.

But on another host, I have at least two that look like this:
```
#0  0x0000ffffa47056f4 in Perl_csighandler3 (sig=15, sip=<optimized out>, uap=<optimized out>) at /usr/src/debug/perl-5.38.2-506.fc40.aarch64/mg.c:1601
1601	           (PL_signals & PERL_SIGNALS_UNSAFE_FLAG))
[Current thread is 1 (Thread 0xfffe93a0f180 (LWP 2365970))]
```
note now we're a couple of lines lower, and signal is 15, which is SIGTERM. But if it's SIGTERM, why didn't this blocker work? Mmmmf.

AdamWill:
actually, that difference seems to be consistent between the two hosts I'm looking at, which is rather interesting. the x86_64 host consistently shows sig as optimized out, and line 1599 (the SIGFPE line). the aarch64 host consistently shows sig as 15 and line 1601.

Vogtinator:
The full backtrace for the SIGTERM case on x86_64 would be interesting, as that shows which kind of thread it is. You could also try to get its signal masks.

PR #1642, opened 2021-04-13
baseclass: Recover broken SSH connection with keep_open=>1

cfconrad:
If the remote side close the connection but we have `keep_open=>1` we
try to establish a new connection.

Ticked: https://progress.opensuse.org/issues/81878
VR: http://cfconrad-vm.qa.suse.de/t8107

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -40,6 +40,7 @@ use List::Util 'min';
 use List::MoreUtils 'uniq';
 use Scalar::Util 'looks_like_number';
 use Mojo::File 'path';
+use OpenQA::Exceptions;
 
 # should be a singleton - and only useful in backend process
 our $backend;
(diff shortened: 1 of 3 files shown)


Martchus:
Looks like baseclass tests fail now. Otherwise it looks good.

cfconrad:
thx @Martchus for pointing me.
I also realized that `22-svirt.pm` and `23-baseclass.t` is using almost the same SSH tests.
I would like to have not the code twice. I will rethink these both tests once more.

Do we have some "lib" files only for tests?

Martchus:
In openQA we have https://github.com/os-autoinst/openQA/tree/master/t/lib so it would make sense to follow the same pattern here. Note that if both tests are really just redundant (testing the same code on the same level) it would also make sense to get rid of one of the redundant tests completely.

cfconrad:
Ready from my side. I split it in several commit's to make review more easy, but feel free to squash them on merge or request it.

okurz (commented):
The splitted commits are good :)

+1

Martchus (approved):

PR #1644, opened 2021-04-15
Fix return value of _dbus_do_call

AdamWill:
This was relying on perl's default return behaviour before
(return the result of the final line of the function, if it's an
expression). But now that's wrong, as the last line is something
else. We need to be explicit about what we're returning (which
would generally be a good thing to do anyway).

Related progress issue: https://progress.opensuse.org/issues/91163

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -141,8 +141,9 @@ sub _dbus_do_call {
     my $bus         = Net::DBus->system(private => 1);
     my $bus_service = $bus->get_service("org.opensuse.os_autoinst.switch");
     my $bus_object  = $bus_service->get_object("/switch", "org.opensuse.os_autoinst.switch");
-    $bus_object->$fn(@args);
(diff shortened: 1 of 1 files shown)


AdamWill:
Unfortunately it doesn't look easy to make the tests test this, because they never actually test a *working* case of `_dbus_do_call`, they only test various different ways it goes wrong.

AdamWill:
I tested this in production (well, staging...) by cloning a job with `OVS_DEBUG=1`, and indeed hit the exact failure from https://progress.opensuse.org/issues/91163 . I then applied this change to the system's `qemu.pm`, tried the job again, and this time it's running and it correctly logged the debug output. So this does look good. Very sorry for the inconvenience.

okurz:
Can you tidy it please?

okurz (approved):

okurz:
tidied

PR #1646, opened 2021-04-16
Update qemu argument used to disable floppy drive for qemu 6.0

AdamWill:
The way we were doing this before no longer works with qemu 6.0.
Markus Armbruster suggests using -nodefaults, but that's a bigger
gun and we'd have to check we aren't relying on the default
devices in any way before using it. This was his second choice
suggestion "If you'd prefer not to". It does seem to work (with
both older and newer qemu). See:

https://bugs.launchpad.net/bugs/1923663

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -857,7 +857,7 @@ sub start_qemu {
     {
         # Remove floppy drive device on architectures
         unless ($arch eq 'aarch64' || $arch eq 'arm' || $vars->{QEMU_NO_FDC_SET}) {
-            sp('global', 'isa-fdc.driveA=');
+            sp('global', 'isa-fdc.fdtypeA=none');
         }
 
         sp('m',       $vars->{QEMURAM})     if $vars->{QEMURAM};


AdamWill:
I don't see what's not tidy. I have a newer tidy here than openQA wants - 20210402 vs. 20210111 - but it finds nothing to correct.

okurz:
The CI check also failed in 14-isotovideo.t though

AdamWill:
oh, hmm, yeah, didn't see that. Would be nice if there was, uh, any useful info at all. I guess I'll have to fiddle with it tomorrow. My scratch builds for Fedora 33, 34 and 35 worked, and that test should've run in those, so it's a bit odd.

Martchus:
Created https://github.com/os-autoinst/os-autoinst/pull/1647 for the tidy error.

AdamWill:
Well, uh, I rebased it and the tests passed this time? I mean, if it was actually broken, you'd have expected the qemu tests to fails as well as / instead of the one that did, right?

Martchus (approved):

okurz (approved):
well, we will see :)

AdamWill:
I'll wait to get tagged on a POO issue about how your data center is now on fire because of this or something...:P

okurz:
what if your change killed progress.opensuse.org and we can not even assign a ticket to you because of that? ;)

PR #1652, opened 2021-04-21
Make ipmi SOL console persistent

czerw:
Fix poo#60437: Persistent console is already implemented for PowerVM
and it works without issue. Ipmi SOL console needs the same
approach. Internal openQA console for SOL should be alive all the time
during the test without stopping and starting it again.

Progress: https://progress.opensuse.org/issues/60437

Verifications without persistent console:
* http://black-bit.suse.cz/tests/135
* http://black-bit.suse.cz/tests/127
* http://black-bit.suse.cz/tests/128

Log contains errors during reconnection to console, test usually fail at some point (soon or later)

Verification with persistent console:
* http://black-bit.suse.cz/tests/138
System survived 32 reboots (just reached max job life time).

More verifications:
system installation: http://black-bit.suse.cz/tests/139
kdump:  http://black-bit.suse.cz/tests/140#step/kdump/47  (console is all the time available after crash, the rest of the test is not ready yet)

VT testing: http://black-bit.suse.cz/tests/143#
VT testing 2nd run: http://black-bit.suse.cz/tests/144#

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -112,7 +112,7 @@ sub do_start_vm {
     $self->get_mc_status;
     $self->restart_host unless $bmwqemu::vars{IPMI_DO_NOT_RESTART_HOST};
     $self->truncate_serial_file;
-    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm');
+    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', {persistent => 1});
     $sol->backend($self);
     return {};
 }


okurz (changes requested):
s/Ipmo/IPMI/
s/strating/starting/

pdostal (approved):

czerw:
@alice-suse @Julie-CAO Could you please recommend some bare metal job from VT testing for verification?

Julie-CAO:
> @alice-suse @Julie-CAO Could you please recommend some bare metal job from VT testing for verification?

Hi Petr,
the ipmi sol console is lost after typing 'reboot' in this test
https://openqa.nue.suse.com/tests/5840869#step/reboot_and_wait_up_normal/8

 if it is just the case the PR fixes, the test is suitable. We met this ipmi problem on the DELL AMD machine which was added to OSD one month ago, ipmi sol console always lost after the second reboot.  we have to remove the machine from our product testing pool, and create a particular test for further investigation. You can use this test to verify. One of our SUT will be saved if the ipmi is fixed. Let me know if I can help.

czerw:
> You can use this test to verify. One of our SUT will be saved if the ipmi is fixed. Let me know if I can help.
@Julie-CAO Verification passed, link is in description.

czerw:
@frankenmichl @pevik @mdoucha just fyi

okurz (changes requested):
Please link a ticket in the git commit message as well.

Awaiting the verification of VT tests.

EDIT: ok, I see that verification jobs are ok so far

czerw:
> Please link a ticket in the git commit message as well.

Done, not sure why I didn't (cause I always do) :)

Martchus (commented):
Looks generally good. Does tidy require spreading this over 3 lines?

czerw:
> Looks generally good. Does tidy require spreading this over 3 lines?

Nope, just my style, fixed.

Julie-CAO:
> > You can use this test to verify. One of our SUT will be saved if the ipmi is fixed. Let me know if I can help.
> > @Julie-CAO Verification passed, link is in description.

Amazing, two verification runs passed. Can we have more runs to confirm it is fixed?  such as 5 runs, as the test did not fail every time.

okurz (approved):

czerw:
> Amazing, two verification runs passed. Can we have more runs to confirm it is fixed? such as 5 runs, as the test did not fail every time.

I run your job 5 times in total and it was fine all the time, see http://black-bit.suse.cz/tests/147#next_previous. You can try to put your job into production schedule when is this change deployed on OSD.

Julie-CAO:
> > Amazing, two verification runs passed. Can we have more runs to confirm it is fixed? such as 5 runs, as the test did not fail every time.
> 
> I run your job 5 times in total and it was fine all the time, see http://black-bit.suse.cz/tests/147#next_previous. You can try to put your job into production schedule when is this change deployed on OSD.

Cool! You save one machine for us! and maybe more machines in the future :+1:

alice-suse:
@czerw Sorry for the late reply. I was on FTO yesterday. Thanks for the efforts to fix  poo#60437. However I would **suggest to make the sol console persistent option configurable via /etc/openqa/workers.ini,** which is needed for 
different ipmi vendor machines(having different born ipmi stability) and different test purpose it serves (hours long jobs or minutes short jobs -- the longer job, the less stable sol becomes, the higher chance needed for non-persistent console).

I know that you have provided VT validation tests based on the job that julie provided. Thanks for it. But the SUT that it's configured to is a new machine that we recently get, which has definitely different ipmi behavior compared to our other ipmi machines, just told from the fact that the same automation code running on it meets ipmi issues while all other machines run well. Thus vice versus, when the validations on it run well , it can not promise that this change will be safe for other machines. Likely not, but it will be seen from next build new jobs.

Let me provide more background info here for ipmi console to explain why. Virtualization team has been using SUTs configured with ipmi backend for several years and we've undergone the big changes of this ipmi console usage from the very beginning(sol console only -> root-ssh console mainly + sol as less as possible -> proactive sol console reconnection  around reboot for higher stability + necessary sol reconnection when sol is detected problematic). The main reasons for above changes are:
- ipmi console itself(underlying software provided by hardware vendors, nothing related to openqa) is unstable, especially when used for long time(hours). It is actually not designed for automated testing purpose, instead designed for administration purpose which just requires usage from time to time. So it does not perform correctly and stablely for long time automation tests, eg several hours which VT jobs are. So that's why sol console is not configured to be persistent .
- different SUT vendors ipmi technology is different and performance during long test time job is different, some better while some worse. So it is better to make persistent option configurable in worker config.
- when sol console becomes unstable, no matter due to long time usage or unluckily short time usage, try to reconnect console can survive with high chance(not always either). So this is why we even proactively reset it before using it around reboot during test.

czerw:
@alice-suse Thanks for feedback, what i see from experience, the main problem when using the SOL console are around console reset, either in internal openQA console level or ipmi level:
*  proactive sol console reconnection around reboot - this is not helping anything, only masking and delaying the issue, for example 2x `reset_consoles` is not solution https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a1d88102f2d865153f87aaa902168ae31de066bf/tests/virt_autotest/login_console.pm#L48
*  necessary sol reconnection when sol is detected problematic - this  is not changed and it should stay as it is

alice-suse:
> @alice-suse Thanks for feedback, what i see from experience, the main problem when using the SOL console are around console reset, either in internal openQA console level or ipmi level:
> 
> * proactive sol console reconnection around reboot - this is not helping anything,

I understand how you feel it this way. With your test code, ipmi SUT setup, and the specific ipmi SUT, you've been troubled by sol console reconnection for long time. But trust me, VT jobs on all osd ipmi SUTs except the recent new purchased one have been working well on sol console reconnection.

> only masking and delaying the issue, 

Unluckily, what we need is to mask or workaround the unstable ipmi console(not related to openqa layer, but ipmi console itself provided by HW vendor), so that test can continue. Our test purpose is not to test ipmi console(HW software level), but to test suse products.
However if the issue you mentioned is openqa backend level(ipmi backend and sol console), or openqa test code level, NO, this is definitely not what we want to mask or delay by resetting console.

>for example 2x `reset_consoles` is not solution https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a1d88102f2d865153f87aaa902168ae31de066bf/tests/virt_autotest/login_console.pm#L48

Agree. To be honest, I was not aware of two lines of reset_consoles or who did that. Definitely this is what I meant to do.

> * necessary sol reconnection when sol is detected problematic - this  is not changed and it should stay as it is

Yep.

PR #1654, opened 2021-04-22
Make persistent SOL console configurable

czerw:
PR https://github.com/os-autoinst/os-autoinst/pull/1652#issuecomment-824505667 made IPMI SOL console persistent to increase stability, there was raised concern: *suggest to make the sol console persistent option configurable* .  Configuration can be done at test suite level, machine defintion,  wrokers.ini file...

Fix poo#60437: It is possible to use IPMI_SOL_PERSISTENT_CONSOLE
variable to configure SOL console persistency. It is disabled by default.

Progress: https://progress.opensuse.org/issues/60437

Verification runs:
IPMI_SOL_PERSISTENT_CONSOLE enabled:
VT test: http://black-bit.suse.cz/tests/158
Installation: http://black-bit.suse.cz/tests/159
Kdump:  http://black-bit.suse.cz/tests/160

Special brute force reset test for variable:
persistency enabled: http://black-bit.suse.cz/tests/169#step/ipmi_reset/30 ok
persistency disabled:  http://black-bit.suse.cz/tests/168#step/ipmi_reset/8  fail

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -112,7 +112,7 @@ sub do_start_vm {
     $self->get_mc_status;
     $self->restart_host unless $bmwqemu::vars{IPMI_DO_NOT_RESTART_HOST};
     $self->truncate_serial_file;
-    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', {persistent => 1});
+    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', {persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 0});
     $sol->backend($self);
     return {};
 }
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -65,6 +65,7 @@ IPMI_MC_RESET_SLEEP_TIME_S;integer;10;Time to sleep after sending mc reset comma
 IPMI_MC_RESET_TIMEOUT;integer;60;Counts to try to reach IPMI interface after mc reset
 IPMI_MC_RESET_PING_COUNT;integer;1;Ping counts that must be successful after mc reset
 IPMI_MC_RESET_IPMI_TRIES;integer;3;Maximum number of IPMI command tries that are conducted after mc reset
+IPMI_SOL_PERSISTENT_CONSOLE;boolean;0;Make SOL console persistent and don't reset it, disabled by default
 |====================
 
 .QEMU backend


czerw:
@okurz @Martchus please check, it it is done in properly or it should be handled by other way

czerw:
Not sure about default value, honor previous behavior or use new one?

alice-suse:
LGTM. @Julie-CAO FYI. Once VT jobs turn unstable on sol console level on existing osd ipmi machines with persistent sol, we may need to switch off it on  job group level. But likely the new machine that we got recently may need this on in worker config, depending on how it turns out. I am not sure which config overwrites the other, better to try.

alice-suse:
> Not sure about default value, honor previous behavior or use new one?

If I can suggest, I would suggest the previous behavior as default since not all people are aware of the persistent sol console and long time sol console connection(for hours long jobs) is highly likely to become problematic.

czerw:
> > Not sure about default value, honor previous behavior or use new one?
> 
> If I can suggest, I would suggest the previous behavior as default since not all people are aware of the persistent sol console and long time sol console connection(for hours long jobs) is highly likely to become problematic.

OK, disabled by default.

okurz (approved):
I consider a feature switch a good idea on general. The conservative approach is to introduce a change in behavior with a default-off switch. After a stabilization time the default can be turned around and eventually if deemed stable the feature switch can be removed again with the behaviour resembling always-on.

However I don't understand why certain machines would need a non-persistent IPMI console. To me, same as I think you, @czerw , it does not make sense to disconnect an IPMI connection when just the SUT reboots. That was the original idea. So my goal would still be to always have a persistent IPMI console. Of course, an IPMI connection can fail, potentially at any time, and a reconnection and proper handling of disconnects is a good idea. It would be better though that we reduce the difference between what QE VT team is doing and what other teams or doing so a non-persistent console connection should only be a very rare and at best temporary exception. If we give test maintainers too much flexibility then bad choices are made, e.g. also see the overuse of VNC_TYPING_LIMIT

Martchus (approved):

czerw:
Verifications are finished.

alice-suse:
> However I don't understand why certain machines would need a non-persistent IPMI console. To me, same as I think you, @czerw , it does not make sense to disconnect an IPMI connection when just the SUT reboots. 

Let me explain why. For most tests running on ipmi SUTs, not only VT jobs, during the whole test time(even hours long), sol console is only used just on booting/installation or rebooting related steps, namely boot_from_pxe and other reboot related steps. When os is up and ssh is up, root-ssh console which is based on ssh connection is used always, which takes most of the test time. So to have higher chance for a workable sol console, it is better to freshly restart a new connection rather than using a long time existing one. That's why before reboot(before we uses sol console), we try to disconnect and reconnect ipmi sol connection. It is proactive selection, rather than a necessity.


>Of course, an IPMI connection can fail, potentially at any time, and a reconnection and proper handling of disconnects is a good idea. 

It is not easy to distinguish problematic console from other failures(eg no proper needle in repo and sol does not react to get timely console picture back). So not easy to do precise console reconnection from problematic console. Besides the points to meet problematic console is too much and the cost is high to add such blocks wherever applicable. So current solution is chosen that reconnecting it before using sol console.

>It would be better though that we reduce the difference between what QE VT team is doing and what other teams or doing so a non-persistent console connection should only be a very rare and at best temporary exception. If we give test maintainers too much flexibility then bad choices are made, e.g. also see the overuse of VNC_TYPING_LIMIT

Overuse is not good. Glad that ipmi users are not too much. But it really is necessary to have this persistent option configurable according to SUT ipmi  conditions mainly. 

Thanks again for the PR support and discussion here.

PR #1655, opened 2021-04-23
dist: Install testapi.py along with other files for easier inclusion

okurz:
Tested manually on a machine which only has the packaged version of os-autoinst available


Martchus (changes requested):
You need to install the file via CMake as well. (Just extend `MISC_FILES` or make a new variable and add it to the loop with the `install` command.)

Martchus (approved):

PR #1656, opened 2021-04-23
Enable smbd usage to upload assets from worker pool

mloviska:
Especially for WSL testing we can upload WSL image from worker pool
using qemu's samba feature.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -874,6 +874,7 @@ sub start_qemu {
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
+                $nictype_user_options .= ",smb=${\(dirname($basedir))}" if ($vars->{QEMU_ENABLE_SMBD});
                 sp('netdev', [qv "user id=qanet$i$nictype_user_options"]);
             }
             elsif ($vars->{NICTYPE} eq "tap") {
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -140,6 +140,7 @@ QEMU_NON_FATAL_DBUS_CALL;boolean;0;Ignore failed dbus calls and ignore instead o
 QEMU_MAX_BANDWIDTH;integer;INT_MAX;Limits the transfer rate during a snapshot.
 QEMU_DUMP_COMPRESS_METHOD;string;xz;The compression to use during a memory dump. Can be set to xz, bzip2 or internal (QEMU's internal compression, not compatible with crash or gdb). If xz is set, but not available, it will fallback to bzip2. Also see QEMU_COMPRESSION_LEVEL.
 QEMU_APPEND;string;;Append parameters on qemu command line. The first item will have '-' prepended to it.
+QEMU_ENABLE_SMBD;boolean;0;Enable QEMU's built-in samba service for user network. Exported worker's pool will be accessible on `\\10.0.2.4\qemu` share.
 VIRTIO_CONSOLE;boolean;1;Enable/disable virtio console. (@see `-device virtconsole` qemu option)
 VIRTIO_CONSOLE_NUM;integer;1;Number of virtio consoles.
 QEMU_BALLOON_TARGET;integer;undef;The target guest RAM usage before a snapshot is taken. It is intended to speed up snapshots by forcing the guest to drop various caches. Setting this enables the virtio-balloon device which requires a kernel with a virtio-balloon driver. Setting this far below the RAM required by the guest will probably cause the guest to panic or deadlock. However it should be able to cope with it being set slightly below what is needed.


backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
+                if ($vars->{QEMU_ENABLE_SMBD}) {
+                    my $smbd_bin = find_bin('/usr/sbin/', 'smbd');
+                    $smbd_bin                             or die "No server to provide SMB/CIFS services to SUT";
+                    (qx($smbd_bin --version) =~ /(\d.*)/) or die "smbd version was not matched";
```
Martchus, line 880: The error message leaves the question which version was expected. However, this check doesn't really validate whether the version matches some expected version (a digit followed by anything counts as version) which is also a bit weird.

backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
+                if ($vars->{QEMU_ENABLE_SMBD}) {
+                    my $smbd_bin = find_bin('/usr/sbin/', 'smbd');
+                    $smbd_bin                             or die "No server to provide SMB/CIFS services to SUT";
```
Martchus, line 879: Please use single-quotes for strings not using interpolation.

backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
...
+                    (qx($smbd_bin --version) =~ /(\d.*)/) or die "smbd version was not matched";
+                    bmwqemu::diag "Worker is using smbd of version: $1";
```
Martchus, line 881: The use of "Worker" is a little bit confusing here as we usually mean the openQA worker in this context. I'd rather log "smbd version found on worker host: ".

Martchus (changes requested):
I'd also find it a bit weird that the path determined here isn't passed anywhere. So technically the code does not really make sure QEMU uses a specific smbd version but writes a log message which could be interpreted this way. And is the error message QEMU would print if the version doesn't match or smbd is not installed at all really so bad that we need to do all this checking here as well?

backend/qemu.pm
mloviska: Ah, I just want to print the version of samba in case of issues, so we know what version has been used. On the second thought `die`ing is the best idea. :) Thanks for the point!

okurz:
s/assests/assets/

backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
+                if ($vars->{QEMU_ENABLE_SMBD}) {
+                    my $smbd_bin = find_bin('/usr/sbin/', 'smbd');
+                    $smbd_bin                             or die 'No server to provide SMB/CIFS services to SUT';
+                    (qx($smbd_bin --version) =~ /(\d.*)/) or warn 'smbd version regex was not matched';
+                    bmwqemu::diag "smbd version found on worker host: $1";
```
okurz, line 881: why is the version important? Do you rely on certain features?

backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq "user") {
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
+                if ($vars->{QEMU_ENABLE_SMBD}) {
+                    my $smbd_bin = find_bin('/usr/sbin/', 'smbd');
```
okurz, line 878: Any particular reason why you use `find_bin` from the os-autoinst module osutils? I don't think this should be necesary here. AFAIU `find_bin` in particular is useful when we want to lookup binaries in multiple alternative paths but here AFAICS you can rely on "smbd" from PATH, isn't it?

backend/qemu.pm
```diff
@@ -874,6 +874,13 @@ sub start_qemu {
...
+                    $smbd_bin                             or die 'No server to provide SMB/CIFS services to SUT';
+                    (qx($smbd_bin --version) =~ /(\d.*)/) or warn 'smbd version regex was not matched';
+                    bmwqemu::diag "smbd version found on worker host: $1";
+                    $nictype_user_options .= ",smb=${\(dirname($basedir))}";
+                }
```
okurz, line 883: indention gets a bit long. Could you please extract the code into a function and then simply call like

```perl
enable_qemu_smbd if $vars->{QEMU_ENABLE_SMBD};
```

okurz (changes requested):

mloviska:
> I'd also find it a bit weird that the path determined here isn't passed anywhere. So technically the code does not really make sure QEMU uses a specific smbd version but writes a log message which could be interpreted this way. And is the error message QEMU would print if the version doesn't match or smbd is not installed at all really so bad that we need to do all this checking here as well?

I am not really sure whether I can choose what smbd binary to execute, the documentation is very succinct. However, if there is no samba server present on the host, which we want to use, qemu will fail to create a share that some tests rely on.

backend/qemu.pm
mloviska: In the past there has been a reported incompatibility.
https://unix.stackexchange.com/questions/165554/shared-folder-between-qemu-windows-guest-and-linux-host

mloviska:
> I'd also find it a bit weird that the path determined here isn't passed anywhere. So technically the code does not really make sure QEMU uses a specific smbd version but writes a log message which could be interpreted this way. And is the error message QEMU would print if the version doesn't match or smbd is not installed at all really so bad that we need to do all this checking here as well?

I have just removed `samba` package and here is the error that I get.

```
[0m[37m[2021-04-26T10:45:47.877 CEST] [debug] QEMU: QEMU emulator version 5.2.0 (openSUSE Tumbleweed)
[0m[37m[2021-04-26T10:45:47.877 CEST] [debug] QEMU: Copyright (c) 2003-2020 Fabrice Bellard and the QEMU Project developers
[0m[37m[2021-04-26T10:45:47.877 CEST] [debug] QEMU: qemu-system-x86_64: Could not find '/usr/sbin/smbd', please install it
[0m[37m[2021-04-26T10:45:47.878 CEST] [debug] sending magic and exit
```

backend/qemu.pm
okurz: I see. To simplify I think we can ignore that assuming that we have newer versions of qemu. The referenced issue was referring to needing at least qemu >= 2.2.0 so maybe instead of smbd check qemu and die instead of warn+diag? I always favor aborting all operations in case of unforeseen cases. If we find out we need to gracefully handle that we can still change it later on. But all that is uncritical when the feature is only enabled anyway based on the feature switch you introduce

okurz (changes requested):
I consider `Could not find '/usr/sbin/smbd', please install it` good enough so I don't see a need to crosscheck further within our code. After all the openQA job result would still fail the same, right?

Martchus (approved):
Looks like you've been removing the checks now and QEMU's own error message is good enough, indeed.

mloviska:
> Looks like you've been removing the checks now and QEMU's own error message is good enough, indeed.

Yup, just kept the bare minimum as the creation of VM dies on its own with the QEMU's error message. :)

okurz (approved):

PR #1657, opened 2021-04-24
stop appending the basename to a set log_name

phil-hands:
At present, if one calls upload_logs thus:

  upload_logs( '/var/log/syslog', log_name => 'SysLog.txt' );

the filename that shows up in the webUI is 'SysLog.txt-syslog' whereas
it should be just 'SysLog.txt' -- this tiny patch fixes that.

BTW one could fix this by simply removing the parentheses, but I prefer to move them as it makes sure that the intent is clear.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -2230,7 +2230,7 @@ sub upload_logs {
     }
     bmwqemu::log_call(file => $file, failok => $failok, timeout => $timeout, %args);
     my $basename = basename($file);
-    my $upname   = ($args{log_name} || $autotest::current_test->{name}) . '-' . $basename;
+    my $upname   = $args{log_name} || ($autotest::current_test->{name} . '-' . $basename);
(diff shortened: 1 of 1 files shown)


okurz:
I agree with that change. But as it is a non-backward-compatible change although of minor impact I am trying to bring in some power users to review as well

@os-autoinst/tests-maintainer WDYT?

alice-suse:
Nice change for users.

coolo:
We don't seem to use log_name from within `lib` folder, so we seem to be unaffected by this. unless of course someone relies on this while fetching logs, but I would say we can ignore that.
The only function is `save_remote_upload_y2logs` in the tests, which seems unused since 2019

okurz (approved):

rwx788 (commented):
PR has been merged, but still +1 to the change ;)

okurz:
thx

PR #1660, opened 2021-04-27
console_proxy: Allow ARRAY as return type

cfconrad:
This allow to use a perl `ARRAY` as return type within that
`backend_proxy_console_call`.

I currently know only one place where we actually run in that trap the lately changed function `sshVirtsh::run_cmd()` https://github.com/os-autoinst/os-autoinst/blob/master/consoles/sshVirtsh.pm#L634

But from my point of view it would be nice to have a complete transparent interface. On the other hand, we actually avoid `wantarray` where ever we can, so maybe that need to be made different. But for now, I don't see any other.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -855,7 +855,7 @@ sub proxy_console_call {
         # Move the decision to actually die to the server side instead.
         # For this ignore backend::baseclass::die_handler.
         local $SIG{__DIE__} = 'DEFAULT';
-        $wrapped_result->{result} = $console->$function(@$args);
+        $wrapped_result->{result} = $wrapped_call->{wantarray} ? [$console->$function(@$args)] : $console->$function(@$args);
     };
     $wrapped_result->{exception} = join("\n", bmwqemu::pp($wrapped_call), $@) if $@;
     return $wrapped_result;
--- a/backend/console_proxy.pm
+++ b/backend/console_proxy.pm
@@ -57,6 +57,7 @@ sub AUTOLOAD {
 			    console => $self->{console},
 			    function => $function,
 			    args => $args,
+			    wantarray => wantarray,
 			   };
         #>>>
 
@@ -72,7 +73,7 @@ sub AUTOLOAD {
         # get more screenshots from consoles, especially from x3270 on s390
         $autotest::current_test->take_screenshot;
 
-        return $wrapped_retval->{result};
+        return wantarray ? @{$wrapped_retval->{result}} : $wrapped_retval->{result};
     };
 
     goto &$AUTOLOAD;
--- a/backend/console_proxy.pm
+++ b/backend/console_proxy.pm
@@ -50,16 +50,14 @@ sub AUTOLOAD {
     # allow symbolic references
     no strict 'refs';
     *$AUTOLOAD = sub {
-        my $self = shift;
-        my $args = \@_;
-	#<<< perltidy, this _is_ tidy...
-	my $wrapped_call = {
-			    console => $self->{console},
-			    function => $function,
-			    args => $args,
-			    wantarray => wantarray,
-			   };
-        #>>>
+        my $self         = shift;
+        my $args         = \@_;
+        my $wrapped_call = {
+            console   => $self->{console},
+            function  => $function,
+            args      => $args,
+            wantarray => wantarray,
+        };
 
         bmwqemu::log_call(wrapped_call => $wrapped_call);
         my $wrapped_retval = autotest::query_isotovideo('backend_proxy_console_call', $wrapped_call);
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -576,30 +576,7 @@ subtest 'wait_still_screen & assert_still_screen' => sub {
         'assert_still_screen forwards arguments to wait_still_screen');
 };
 
-subtest 'set console tty and other args' => sub {
-    # ensure previously emitted commands are cleared
-    $cmds = ();
-
-    console->set_tty(4);
-    console->set_args(tty => 5, foo => 'bar');
-
-    is_deeply(
-        $cmds,
-        [
-            {
-                cmd      => 'backend_proxy_console_call',
-                console  => 'a-console',
-                function => 'set_tty',
-                args     => [4],
-            },
-            {
-                cmd      => 'backend_proxy_console_call',
-                console  => 'a-console',
-                function => 'set_args',
-                args     => ['tty', 5, 'foo', 'bar'],
-            },
-        ]);
-
+subtest 'test console::console argument settings' => sub {
     # test console's methods manually since promoting the commands is mocked in this test
     my $console = consoles::console->new('dummy-console', {tty => 3});
     is($console->{args}->{tty}, 3);
--- a/t/32-console_proxy.t
+++ b/t/32-console_proxy.t
@@ -1,0 +1,153 @@
+#!/usr/bin/perl
+
+use Test::Most;
+use Test::MockObject;
+use Test::MockModule;
+use Test::Warnings qw(:all :report_warnings);
+
+use backend::baseclass;
+use bmwqemu;
+use testapi;
+use basetest;
+use distribution;
+
+my $mock_bmwqemu   = Test::MockModule->new('bmwqemu');
+my $mock_basetest  = Test::MockModule->new('basetest');
+my $mock_baseclass = Test::MockModule->new('backend::baseclass');
+my $mock_jsonrpc   = Test::MockModule->new('myjsonrpc');
+
+my $console_check_args = [];
+my %console_const_hash = (x => 'a', y => 'b', z => 'c');
+my $console            = Test::MockObject->new();
+$console->mock('ret_array',           sub { my @array = qw(a b c d ); });
+$console->mock('ret_array_empty',     sub { my @a; });
+$console->mock('ret_array_ref',       sub { [qw(a b c d)]; });
+$console->mock('ret_array_ref_empty', sub { []; });
+$console->mock('ret_hash',            sub { %console_const_hash; });
+$console->mock('ret_hash_empty',      sub { my %empty_hash; });
+$console->mock('ret_hash_ref',        sub { {x => 'a', y => 'b', z => 'c'}; });
+$console->mock('ret_hash_ref_empty',  sub { {} });
+$console->mock('ret_scalar',          sub { "a" });
+$console->mock('ret_undef',           sub { return undef; });
+$console->mock('ret_list',            sub { qw(a b c d); });
+$console->mock('ret_list_empty',      sub { return; });
+$console->mock('ret_die',             sub { die("!!Urgs!!"); });
+$console->mock('check_args', sub {
+        my ($self, @args) = @_;
+        is_deeply(\@args, $console_check_args, 'Got expected (' . join(',', @args) . ') arguments');
+});
+
+$mock_bmwqemu->noop('log_call');
+$mock_basetest->noop('_result_add_screenshot');
+$mock_baseclass->redefine('console', $console);
+
+my $baseclass = backend::baseclass->new();
+testapi::set_distribution(distribution->new());
+$autotest::current_test = basetest->new();
+
+my $jsonrpc_cmds    = [];
+my $jsonrpc_results = [];
+$mock_jsonrpc->redefine(
+    send_json => sub { push(@$jsonrpc_cmds, $_[1]); },
+    read_json => sub {
+        my $cmd = $jsonrpc_cmds->[-1]->{cmd};
+        if ($cmd eq 'backend_proxy_console_call') {
+            push @$jsonrpc_results, $baseclass->proxy_console_call($jsonrpc_cmds->[-1]);
+            return {ret => $jsonrpc_results->[-1]};
+        }
+        elsif ($cmd eq 'backend_select_console') {
+            return {ret => {activated => 0}};
+        }
+
+        die("$cmd not handled in this test");
+    });
+
+subtest 'Verify fake console return values in scalar context' => sub {
+    is_deeply(scalar($console->ret_array()),           4,                              'ARRAY');
+    is_deeply(scalar($console->ret_array_empty()),     0,                              'Empty ARRAY');
+    is_deeply(scalar($console->ret_array_ref()),       ['a', 'b', 'c', 'd'],           'ARRAY-REF');
+    is_deeply(scalar($console->ret_array_ref_empty()), [],                             'Empty ARRAY-REF');
+    is_deeply(scalar($console->ret_hash()),            3,                              'HASH');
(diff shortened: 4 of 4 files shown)


cfconrad:
This is how I played with this topic:
```perl
#!/usr/bin/perl

use Mojo::Base -strict;
use Test::Most;

sub func_array {
    return qw(a b c d);
}

sub func_scalar {
    return "a";
}

sub func_hash {
    return (x => 'a', y => 'b', z => 'c');
}

sub func_hash_ref {
    return { x => 'a', y => 'b', z => 'c' };
}

sub func_array_ref {
    return [qw(a b c d)];
}

sub get_result{
    my $func = shift;
    my $ret = [$func->()];
    return wantarray ? @{$ret} : $ret->[-1];
}


subtest 'foo' => sub {

    for my $func (qw(func_array func_scalar func_hash func_hash_ref func_array_ref)){
        my @array_got = get_result(\&{$func});
        my @array_exp = &{\&{$func}}();
        is_deeply(\@array_got, \@array_exp, "$func() in array context");

        my $scalar_got = get_result(\&{$func});
        my $scalar_exp = &{\&{$func}}();
        is_deeply($scalar_got, $scalar_exp, "$func() in scalar context");

    }
};

done_testing;
```

Martchus:
I'm adding the "not-ready" label to prevent mergify from merging the PR as it is tagged as RFC.

I would personally refrain from using `wantarray`. The return type of a function should not depend on the context it is being called. I also don't really see the benefit.

Besides, returning an array is usually also slower than returning an array reference. (Has Perl RVO?)

cfconrad:
Not sure if `speed` is a argument on something which goes over a socked and is `serialized/de-serialized` from json.
I think it is just a perlish call usage or not. Something like:
```
my (undef, undef, $need_this, $need_that) = func();
```
against
```
my (undef, undef, $need_this, $need_that) = @{func()};
```
or 
```
my $ret = func();
my $need_this = $ret->[2];
my $need_that = $ret->[3];
```

b10n1k:
> I'm adding the "not-ready" label to prevent mergify from merging the PR as it is tagged as RFC.
> 
> I would personally refrain from using `wantarray`. The return type of a function should not depend on the context it is being called. I also don't really see the benefit.
> 
> Besides, returning an array is usually also slower than returning an array reference. (Has Perl RVO?)

@Martchus if i can add something here, even if @cfconrad made it pretty clear IMO. Without this there is strange behavior for some callbacks. For instance https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12332/files#diff-47a0a7f15c1bd3166c2ca1c8c3f4f0c7f7beb1b1b5c4d349e4da8cab17e9ae5eR235. This will call `return $self->backend->run_ssh_cmd($cmd, %credentials, %args);` which in turn it reaches this [return](https://github.com/os-autoinst/os-autoinst/blob/master/backend/baseclass.pm#L1383) inside the `run_ssh_cmd`. then(when you expect an array) the expected output is not in the same type. it is a scalar. So it doesnt make sense. I was about to open a bug today. But this PR looks to solve this issue.

okurz (changes requested):
The idea in general looks valid to me

Please add corresponding test coverage

kraih:
It is best practice these days to avoid `wantarray` completely in Perl programming. It has caused way too many CVEs.

cfconrad:
Not sure how to proceed now.
Plz lets make a decision, before I continue write tests for it.

The problem I have is, that we do not allow ARRAY as return value when using the `console_proxy`.
If someone have a other solution than the one in that PR, I would like to see it.

The other possibility is, that we change the specific functions to return a ARRAY_REF.

Martchus:
I would vote against wantarray and prefer an explicit parameter to influence the return type. If it is possible (does not break too much existing code) you can also always return an array or an array ref. Just don't make it dependent on the calling context.

kraih:
My personal preference would be differently named functions for different return values, but a parameter is fine too.

cfconrad:
Ok the function ([run_cmd()](https://github.com/os-autoinst/os-autoinst/blob/master/consoles/sshVirtsh.pm#L630)) has already the additional parameter `wantarray => (0|1)` and this is **not** the problem I try to solve with this PR.

The problem comes from the https://github.com/os-autoinst/os-autoinst/blob/master/backend/console_proxy.pm and how we store the result in the `$wrapped_result`, see https://github.com/os-autoinst/os-autoinst/blob/master/backend/baseclass.pm#L858.

Here we store it **always** in scalar context. So a ARRAY which is returned get just lost!

And yes, we could walk over the arguments in `baseclass::proxy_console_call`, search for `wantarray` and use the next value as the boolean to get if it return array or not. But from my point of view this is more ugly then using that native `wantarray` feature here.

Maybe there is just a other way to find out, if a function returned a array or a scalar!
So I'm looking for a `magic_function()` which provide me the following:
```perl
my $rv = magic_function($console->$function(@$args));
if ($rv->{is_array}) {
    # we need to de reference the result on the other side of console_proxy !
}
```

perlpunk:
> Maybe there is just a other way to find out, if a function returned a array or a scalar!
> So I'm looking for a `magic_function()` which provide me the following:
> 
> ```perl
> my $rv = magic_function($console->$function(@$args));
> if ($rv->{is_array}) {
>     # we need to de reference the result on the other side of console_proxy !
> }
> ```

Unfortunately that's not possible in perl.
perl subroutines always return a list of scalars, never an array or a hash.
A scalar can be a string, number, undef, but it also can be an array (or hash) reference.

If you meant "array reference" instead, then there is a function: `ref`
```
say ref [];
say ref {};
say ref "string"
__END__
ARRAY
HASH

```
So you could use `if (ref $var eq 'ARRAY') { ... }`.

cfconrad:
**Thx Tina**, 
but I was talking about the list of scalar case. And now I start to wonder, where is the difference between a list of scalar and a array?

And this is the solution https://stackoverflow.com/questions/34685788/convert-array-to-list

in my above example it would be
```
sub get_result{
    my $func = shift;
    my $ret = [$func->()];
    return @{$ret}[0..$#{$ret}];
}
```

cfconrad:
Looks like this also doesn't work. There is a unreasonable situation when returning a empty ARRAY instead of an empty LIST in scalar context see this perltest:
```perl
#!/usr/bin/perl

use Mojo::Base -strict;
use Test::Most;

sub func_array_empty {
    my @a = qw();
    return @a ;
}

sub func_list_empty {
    return;
}

subtest 'foo' => sub {
    is(scalar(func_array_empty()), 0 , "[SCALAR] Empty array is 0");
    is(scalar(func_list_empty()), undef , "[SCALAR] Empty list is undef");

    is_deeply([func_array_empty()], [] , "[ARRAY] Empty array is []");
    is_deeply([func_list_empty()], [] , "[ARRAY] Empty list is []");
};

done_testing;
```

So from my side, the only way to solve it completely is `wantarray`. Or we force a empty LIST also to `0`. For me it is still better then forcing SCALAR always.

kraih:
Using a bare`return;` is also considered bad practice btw., since it behaves differently based on context (just like `wantarray`).

cfconrad:
Might be and others say that using explicit `return undef;` is bad, as it is true in `list` context.
Anyway.. thats perl.

I would like to allow using lists in return value. And I'm fine, if we do not "allow" context specific `return;` and overwrite such results just with `0` in scalar context (in perl it is `undef`) and empty list in list context.

If we agree on this, then I would continue and write tests for this implementation.

kraih:
For a time people did indeed say that `return undef;` was bad (because of the Perl Best Practices) book. But that rule was deemed a mistake in retrospect and `return undef;` is once again considered good style. 😉

cfconrad:
So finally I failed to find a "good" solution without using `wantarray`.

I think this `console_proxy` should work transparent as good as possible for the caller. And for this, we need to call the real function in the same context as the proxy function was called. 
To achieve this, I now forward the value of `wantarray` to `backend::baseclass` and call the real console function in the same context.

Attached latest version of my sandbox to play with it:
```perl
#!/usr/bin/perl

use Mojo::Base -strict;
use Test::Most;
use Cpanel::JSON::XS ();

sub return_array {
    my @a = qw(a b c d);
    return @a;
}

sub return_array_empty {
    my @a = qw();
    return @a;
}

sub return_scalar {
    my $s = "a";
    return $s;
}

sub return_undef {
    return undef;
}

my %h = ( x => 'a', y => 'b', z => 'c' );

sub return_hash {
    return %h;
}

sub return_hash_ref {
    return { x => 'a', y => 'b', z => 'c' };
}

sub return_hash_ref_empty {
    return {};
}

sub return_array_ref {
    return [qw(a b c d)];
}

sub return_array_ref_empty {
    return [];
}

sub return_list {
    return qw(a b c d);
}

sub return_list_empty {
    return;
}

sub console_proxy {
    my $func = shift;
    my $ret;

    #---
    $ret = wantarray ? [ $func->() ] : $func->();

    # emulate sending via JSON
    my $cjx  = Cpanel::JSON::XS->new->canonical->utf8->convert_blessed();
    my $msg  = { result => $ret };
    my $json = $cjx->encode($msg);

    # emulate receiving via JSON
    $msg = $cjx->decode($json);
    $ret = $msg->{result};

    #---
    return wantarray ? @{$ret} : $ret;
}

subtest 'Emulate console_proxy' => sub {

    for my $func (
        qw(return_array return_array_empty return_scalar return_undef return_hash return_hash_ref return_hash_ref_empty return_array_ref return_array_ref_empty return_list return_list_empty)
      )
    {

        my @array_got = console_proxy( \&{$func} );
        my @array_exp = &{ \&{$func} }();
        is_deeply( \@array_got, \@array_exp, "$func() in list context" );

        my $scalar_got = console_proxy( \&{$func} );
        my $scalar_exp = &{ \&{$func} }();
        is_deeply( $scalar_got, $scalar_exp, "$func() in scalar context" );
    }
};

done_testing;
```

backend/baseclass.pm
```diff
@@ -855,7 +855,11 @@ sub proxy_console_call {
         # Move the decision to actually die to the server side instead.
         # For this ignore backend::baseclass::die_handler.
         local $SIG{__DIE__} = 'DEFAULT';
-        $wrapped_result->{result} = $console->$function(@$args);
+        if ($wrapped_call->{wantarray}) {
+            $wrapped_result->{result} = [$console->$function(@$args)];
```
okurz, line 859: as codecov is reporting that body of the if is now not covered by tests. Please add according test coverage.

perlpunk: yes, we should have a test that actually uses the list context.

Martchus:
That sounds more convincing. So I suppose it would be ok if we merged it. It would be nice to extend the test as the changed code doesn't seem to be fully covered.

cfconrad:
Sounds good, I will start writing tests!

t/03-testapi.t
```diff
@@ -576,30 +578,144 @@ subtest 'wait_still_screen & assert_still_screen' => sub {
...
+    });
 
-    is_deeply(
-        $cmds,
-        [
+    my $baseclass      = backend::baseclass->new();
+    my $mock_baseclass = Test::MockModule->new('backend::baseclass');
+    $mock_baseclass->redefine('console', sub {
+            return $console;
+    });
```
Martchus, line 637:
```suggestion
    $mock_baseclass->redefine(console => sub { $console; });
```

Maybe some of the other redefines can also be written in a more compact way.

t/03-testapi.t
```diff
@@ -576,30 +578,144 @@ subtest 'wait_still_screen & assert_still_screen' => sub {
         'assert_still_screen forwards arguments to wait_still_screen');
 };
 
-subtest 'set console tty and other args' => sub {
-    # ensure previously emitted commands are cleared
-    $cmds = ();
+subtest 'testapi::console() => backend::console_proxy => backend::baseclass::proxy_console_call()' => sub {
+    use backend::baseclass;
```
Martchus, line 582: I'd move that to the top.

Martchus (commented):
I suppose the tests look good. I'd only use a more compact style.

t/03-testapi.t
cfconrad: done

t/03-testapi.t
cfconrad: done

backend/baseclass.pm
cfconrad: done

Martchus (approved):
I'm keeping the "not-ready" label so more people can review it (before mergify automatically merges it).

perlpunk (approved):

backend/baseclass.pm
```diff
@@ -855,7 +855,11 @@ sub proxy_console_call {
...
+        } else {
+            $wrapped_result->{result} = $console->$function(@$args);
+        }
```
okurz, line 862: Can you make that a single assignment call with the ternary operator?

```suggestion
        $wrapped_result->{result} = $wrapped_call->{wantarray} ? [$console->$function(@$args)] : $console->$function(@$args);
```

backend/console_proxy.pm
```diff
@@ -50,15 +50,14 @@ sub AUTOLOAD {
     # allow symbolic references
     no strict 'refs';
     *$AUTOLOAD = sub {
-        my $self = shift;
-        my $args = \@_;
-	#<<< perltidy, this _is_ tidy...
```
okurz, line 55: I appreciate that you could get rid of this line … finally :D

t/03-testapi.t
```diff
@@ -9,11 +9,14 @@ use Test::Mock::Time;
...
 use Test::Output qw(combined_like stderr_like);
 use Test::Fatal;
+use Test::MockObject;
+use Test::Exception;
 use Test::Warnings qw(:all :report_warnings);
 use Scalar::Util 'looks_like_number';
 
 use OpenQA::Isotovideo::Interface;
 use consoles::console;
+use backend::baseclass;
```
okurz, line 19: As this test module t/03-testapi.t is already a bit big I suggest to move out this use and all corresponding tests into a different or its own test module so that we try to keep the "high level of test API only" in this test module

okurz (changes requested):

t/03-testapi.t
cfconrad: Ok, I'm going for it.

backend/baseclass.pm
cfconrad: sure, I took the long version, because the line get so long and thought it is more readable. But I don't have hard feelings.

t/03-testapi.t
cfconrad: What is the trick, that the test is automatically loaded/run? I remember that I previously added it in some `Makefile`.

cfconrad: And `done` from my side.

perlpunk: It's not necessary anymore to explicitly add test files

okurz (approved):

PR #1661, opened 2021-04-27
Fixes return type of run_cmd when wantarray

b10n1k:
draft

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -631,7 +631,14 @@ sub run_cmd {
     my ($self, $cmd, %args) = @_;
     my %credentials = $self->get_ssh_credentials($args{domain});
     delete $args{domain};
-    return $self->backend->run_ssh_cmd($cmd, %credentials, %args);
+    my @response;
+    if ($args{wantarray}) {
+        @response = $self->backend->run_ssh_cmd($cmd,  %credentials, %args);
+        return @response;
+    } else {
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -631,7 +631,14 @@ sub run_cmd {
...
     delete $args{domain};
-    return $self->backend->run_ssh_cmd($cmd, %credentials, %args);
+    my @response;
+    if ($args{wantarray}) {
+        @response = $self->backend->run_ssh_cmd($cmd,  %credentials, %args);
+        return @response;
+    } else {
+        $ret = $self->backend->run_ssh_cmd($cmd, %credentials, %args);
+        return $ret
+    }
```
cfconrad, line 641: I don't see, that your change is solving the problem. 
I would go with:

```suggestion
    if ($args{wantarray}) {
        return [$self->backend->run_ssh_cmd($cmd,  %credentials, %args)];
    } else {
        return $self->backend->run_ssh_cmd($cmd, %credentials, %args);
    }
```

okurz: then please use the ternary operator directly:

```suggestion
    return $args{wantarray} ? @{$self->backend->run_ssh_cmd($cmd,  %credentials, %args)} : $self->backend->run_ssh_cmd($cmd, %credentials, %args);
```

okurz:
Please add corresponding unit tests

b10n1k:
(much better permanent solution)[https://github.com/os-autoinst/os-autoinst/pull/1660]

PR #1662, opened 2021-05-03
qemu: Allow to configure an additional serial device for test-internal use

okurz:
Hotpatched on openqaworker7 with

```
curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/1662.patch | patch -p1 --directory=/usr/lib/os-autoinst
```

* Triggered a test without any additional parameters

```
openqa-clone-job --skip-chained-deps --within-instance https://openqa.opensuse.org/tests/2296755 TESTS=os-autoinst_gh_1662 BUILD=okurz-os-autoinst_gh_1662 SCHEDULE=tests/boot/boot_to_desktop,tests/console/ncurses WORKER_CLASS=openqaworker7
```

Created job #2298224: opensuse-Tumbleweed-DVD-x86_64-Build20220413-extra_tests_misc@64bit -> https://openqa.opensuse.org/t2298224 (passed, still default serial device serial0)

* Adding custom serial device config

```
openqa-clone-job --skip-chained-deps --within-instance https://openqa.opensuse.org/tests/2296755 TESTS=os-autoinst_gh_1662 BUILD=okurz-os-autoinst_gh_1662 SCHEDULE=tests/boot/boot_to_desktop,tests/console/ncurses WORKER_CLASS=openqaworker7 QEMU_SERIAL=chardev:serial42
```

Created job #2298234: opensuse-Tumbleweed-DVD-x86_64-Build20220413-extra_tests_misc@64bit -> https://openqa.opensuse.org/t2298234

Related progress issue: https://progress.opensuse.org/issues/91965

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -153,6 +153,7 @@ QEMU_NO_KVM;boolean;0;Don't use KVM acceleration.
 QEMU_NO_TABLET;boolean;0;Don't use USB tablet.
 QEMU_VIRTIO_RNG;boolean;1;Enable virtio random number generator
 QEMU_NUMA;boolean;0;Enable NUMA simulation, requires QEMUCPUS to be greater than one
+QEMU_SERIAL;string;;Add an additional qemu serial device configuration, e.g. "chardev:serial42"
 QEMU_SMBIOS;see qemu -smbios ?;undef;pass this value to qemu -smbios
 QEMU_SOUNDHW;see qemu -soundhw ?;had;pass this value to qemu -soundhw (for qemu < 4.2)
 QEMU_AUDIODEV;see qemu -device ?;intel-hda;Audio device to use with audiodev to qemu -device (for qemu >= 4.2)
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -767,8 +767,13 @@ sub start_qemu ($self) {
     $self->{proc}->init_blockdev_images();
 
     sp('only-migratable') if $self->can_handle({function => 'snapshots', no_warn => 1});
-    sp('chardev', 'ringbuf,id=serial0,logfile=serial0,logappend=on');
-    sp('serial', 'chardev:serial0');
+    my $serial_id = 0;
+    if ($vars->{QEMU_SERIAL}) {
+        sp('serial', $vars->{QEMU_SERIAL});
+        $serial_id++;
+    }
+    sp('chardev', "ringbuf,id=serial$serial_id,logfile=serial$serial_id,logappend=on");
+    sp('serial', "chardev:serial$serial_id");
 
     if ($self->requires_audiodev) {
         my $audiodev = $vars->{QEMU_AUDIODEV} // 'intel-hda';


okurz:
@dcermak maybe this helps with the use case of relying on the first serial device to control the SUT within tests? If you are interested I would be very happy if you could try this out and give some feedback.

dcermak:
@okurz I'd love to give this a try, but I honestly don't really understand how I would use this? I guess I would set `QEMU_SERIAL` to e.g. `1`, but then what?

okurz:
no, you need to add qemu parameters into `QEMU_SERIAL`. For that please see the documentation of qemu itself or copy-paste qemu parameters from a working libvirt example that you mentioned as reference. Then whatever additional device you configure should be the *first* device within the SUT, i.e. the one that shows the login prompt that you want to interact with. And then you can try to interact with that device from the outside, e.g. write to the device file

okurz:
@dcermak I am still open to explore this idea further myself if you could give me a hint about an appliance that offers a getty on a serial terminal by default so that eventually I could test myself.

dcermak:
Thanks for this PR @okurz and apologies for not coming back to this earlier. I have taken a look at this, but I'm afraid that this is not really useful for my case. I would need a way how I could access the qemu serial console on a arbitrary worker supporting qemu, and not only on one that I have setup in some special way.

okurz:
Well, a test variable like `QEMU_SERIAL` could be specified in any scope so for example it could be set globally for an instance or on a special worker.

can you confirm if the feature works for you for an individual, local isotovideo run before we go further?

dcermak:
> Well, a test variable like `QEMU_SERIAL` could be specified in any scope so for example it could be set globally for an instance or on a special worker.
> 
> can you confirm if the feature works for you for an individual, local isotovideo run before we go further?

I honestly don't know what to specify here :-( I'd really just like to be able to read the serial console. Writing would be great to have but reading is sufficient, even after the fact. Atm I am writing tests for kiwi images, which from time to time don't show anything useful over vnc but provide lots of useful output via the serial console. But I cannot access that, when I run the test on o3. Having the serial log available after the test (failed) would be a net win already for me.

okurz:
Because you said you don't know what to "specify", let me try to rephrase my earlier comment. Maybe that helps:

> Well, a test variable like `QEMU_SERIAL` could be specified in any scope so for example it could be set globally for an instance or on a special worker.

So if this PR would be verified to work and would be deployed then a test variable `QEMU_SERIAL` could be set on multiple different places, whatever suits best. For example if there is only a single job where you want that, e.g. for manual investigation, you could set the test variable on job trigger time, e.g. `openqa-cli api -X post jobs TEST=my_test … QEMU_SERIAL=chardev:serial0` and you should have *two* logfiles resulting, the one you explicitly selected "serial0.txt" and one automatically created, in that case with the id incremented by one, "serial1.txt". If that second serial port makes sense for a broader context, e.g. an openQA "machine definition" then it could be put there, let's imagine next to the default on o3 "64bit" there could be a "64bit-with_two_serial_ports" which has the same test variable. The difference is merely that every test that is scheduled with this machine definition inherits the second serial port qemu setting so no need to manually maintain the same setting (and duplicating) on all test suites or jobs.

But, before we merge this I would prefer if we – or rather you – can verify that this helps, e.g. by running isotovideo built based on my branch and try out if it works in your context, for the example you mentioned.

doc/backend_vars.asciidoc
```diff
@@ -153,6 +153,7 @@ QEMU_NO_KVM;boolean;0;Don't use KVM acceleration.
 QEMU_NO_TABLET;boolean;0;Don't use USB tablet.
 QEMU_VIRTIO_RNG;boolean;1;Enable virtio random number generator
 QEMU_NUMA;boolean;0;Enable NUMA simulation, requires QEMUCPUS to be greater than one
+QEMU_SERIAL;string;;Add an additional qemu serial device configuration. Example string "chardev:serial42"
```
Martchus, line 156:
```suggestion
QEMU_SERIAL;string;;Add an additional qemu serial device configuration, e.g. "chardev:serial42"
```

okurz: done

okurz:
@dcermak can you say what qemu command line you would like to end up with so that I try to generate that with os-autoinst?

The default is something like

```
qemu-system-x86_64 -chardev ringbuf,id=serial1,logfile=serial1,logappend=on -serial chardev:serial1
```

PR #1663, opened 2021-05-03
Fix `eject_cd` command by using `id` and `force` parameters

Martchus:
* The `device` parameter is deprecated and it seems not possible to refer
  to the correct CD device with it (always got `DeviceNotFound` error). It
  seems to work with the `id` parameter and apparently `cd0-device` is the
  correct ID for the device providing the default ISO.
* The force parameter is enabled to make sure it works in any situation.
  The conversion to Mojo::JSON->true/false is required because QEMU is
  picky on the type.
* Tested by sending the command via the developer console. As expected, the
  machine showed the screen that no bootable device has been found (after
  resetting the system via the QMP command `system_reset`).
* See https://progress.opensuse.org/issues/91971
* See https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#qapidoc-966

---

Note that the extra commit "Add function to QEMU backend which allows executing any QMP command" adds a function which I've been using to play around with the QMP interface. It is likely very useful for testing other QMP-related behavior so I'd like to keep it and added a unit test for it.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -109,7 +109,10 @@ sub power {
 
 sub eject_cd {
     my ($self, $args) = @_;
-    $self->handle_qmp_command({execute => 'eject', arguments => {device => ($args->{device} // 'cd0')}});
+    $self->handle_qmp_command({execute => 'eject', arguments => {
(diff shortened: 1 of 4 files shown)


okurz (approved):
Just perfect :)

PR #1665, opened 2021-05-06
Serial failure detection: Support 'info' type

BillAnastasiadis:
Checks and subsequent logic added for the new fail type `info`, added in `known_bugs.pm`.

- Related ticket: https://progress.opensuse.org/issues/46502
- Related pr: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12491

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -705,14 +705,17 @@ sub parse_serial_output_qemu {
             my $type    = $regexp_table->{type};
 
             # Input parameters validation
-            die "Wrong type defined for serial failure. Only 'soft' or 'hard' allowed. Got: $type" if $type !~ /^soft|hard|fatal$/;
-            die "Message not defined for serial failure for the pattern: '$regexp', type: $type"   if !defined $message;
+            die "Wrong type defined for serial failure. Only 'info', 'soft', 'hard' or 'fatal' allowed. Got: $type" if $type !~ /^info|soft|hard|fatal$/;
+            die "Message not defined for serial failure for the pattern: '$regexp', type: $type"                    if !defined $message;
 
             # If you want to match a simple string please be sure that you create it with quotemeta
             if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
                 $regexp_matched{$regexp} = 1;
                 my $fail_type = 'softfail';
-                if ($type =~ 'hard|fatal') {
+                if ($type eq 'info') {
+                    $fail_type = 'ok';
+                }
+                elsif ($type =~ 'hard|fatal') {
                     $die                   = 1;
                     $fail_type             = 'fail';
                     $self->{fatal_failure} = $type eq 'fatal';
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -82,6 +82,13 @@ subtest parse_serial_output => sub {
     is($message,            undef, 'test details do not have extra message');
 
     $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'ok',                                                       'test result set to ok');
+    is($message,            'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
+
+    $basetest->{serial_failures} = [
         {type => 'soft', message => 'SimplePattern', pattern => qr/Serial/},
     ];
 


basetest.pm
```diff
@@ -705,20 +705,24 @@ sub parse_serial_output_qemu {
...
 
             # If you want to match a simple string please be sure that you create it with quotemeta
             if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
                 $regexp_matched{$regexp} = 1;
-                my $fail_type = 'softfail';
-                if ($type =~ 'hard|fatal') {
-                    $die                   = 1;
-                    $fail_type             = 'fail';
-                    $self->{fatal_failure} = $type eq 'fatal';
+                if ($type =~ 'info') {
```
Martchus, line 714:
```suggestion
                if ($type eq 'info') {
```

basetest.pm
```diff
@@ -705,20 +705,24 @@ sub parse_serial_output_qemu {
             my $type    = $regexp_table->{type};
 
             # Input parameters validation
-            die "Wrong type defined for serial failure. Only 'soft' or 'hard' allowed. Got: $type" if $type !~ /^soft|hard|fatal$/;
+            die "Wrong type defined for serial failure. Only 'soft' or 'hard' allowed. Got: $type" if $type !~ /^info|soft|hard|fatal$/;
```
Martchus, line 708: I know that the "Only …" part already misses `fatal` but it would still make sense to update it.

Martchus (commented):
The commit message (and not just the PR description) should have more details and the ticket reference. It would also be good if the commit title states that this is about serial failure detection, e.g. `serial failure detection: Support 'info' type`.

It would be nice to extend the unit tests.

okurz (changes requested):
As mkittler stated, please add tests

basetest.pm
BillAnastasiadis: Thanks for the suggestions! Implemented the changes, will add unit tests.

basetest.pm
BillAnastasiadis: Nice catch. Changed

basetest.pm
```diff
@@ -705,20 +705,24 @@ sub parse_serial_output_qemu {
...
+                    testapi::record_info("$type", "$message");
+                } else {
+                    my $fail_type = 'softfail';
+                    if ($type =~ 'hard|fatal') {
+                        $die                   = 1;
+                        $fail_type             = 'fail';
+                        $self->{fatal_failure} = $type eq 'fatal';
+                    }
+                    $self->record_resultfile($message, $message . " - Serial error: $line", result => $fail_type);
+                    $self->{result} = $fail_type;
```
okurz, line 724: you can still use `$self->record_resultfile` here so no need to make it that different and hence can avoid the higher complexity of multiple levels of if.

```suggestion
                my $fail_type = 'softfail';
                if ($type eq 'info') {
                    $fail_type = 'ok';
                }
                elsif ($type =~ 'hard|fatal') {
                    $die                   = 1;
                    $fail_type             = 'fail';
                    $self->{fatal_failure} = $type eq 'fatal';
                }
                $self->record_resultfile($message, $message . " - Serial error: $line", result => $fail_type);
                $self->{result} = $fail_type;
```

okurz (changes requested):

basetest.pm
okurz: To answer your question where "record_info" is to be used: You don't need to call it directly. Sorry if the ticket was phrased not entirely ambiguously in that regard. Actually I think it's a good practice to not call any public test API internally in other os-autoinst modules (following "design by contract" principles).

basetest.pm
```diff
@@ -705,17 +705,20 @@ sub parse_serial_output_qemu {
...
                 my $fail_type = 'softfail';
-                if ($type =~ 'hard|fatal') {
-                    $die                   = 1;
-                    $fail_type             = 'fail';
-                    $self->{fatal_failure} = $type eq 'fatal';
+                if ($type eq 'info') {
+                    $fail_type = 'info';
+                }
+                elsif ($type =~ 'hard|fatal') {
+                        $die                   = 1;
```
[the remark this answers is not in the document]
BillAnastasiadis: Done

basetest.pm
```diff
@@ -705,17 +705,20 @@ sub parse_serial_output_qemu {
...
             # If you want to match a simple string please be sure that you create it with quotemeta
             if (!exists $regexp_matched{$regexp} and $line =~ /$regexp/) {
                 $regexp_matched{$regexp} = 1;
                 my $fail_type = 'softfail';
-                if ($type =~ 'hard|fatal') {
-                    $die                   = 1;
-                    $fail_type             = 'fail';
-                    $self->{fatal_failure} = $type eq 'fatal';
+                if ($type eq 'info') {
+                    $fail_type = 'info';
```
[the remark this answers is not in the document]
BillAnastasiadis: Done.

okurz (approved):

PR #1666, opened 2021-05-08
t: Reduce runtime of 31-sshSerial by mocking time based methods

okurz:


kalikiana (approved):
Would be awesome if you could add some context!

But I'll approve in any case. Nice tweak

okurz:
> Would be awesome if you could add some context!

not sure what context you could be interested in. Basically the only git commit subject line explains the benefit of the change and how it's done. The only other context that I can think of is that I was reviewing another pull request (don't remember which one exactly) and found that the test module spends 2 seconds sleeping so I tried to fix that. Overall we do not save a lot of time when other test modules like the qemu ones and full stack run for multiple minutes but I wanted to see if I can get rid of the sleeps because I consider that the best programming style in general and to be fair I should be able to achieve that myself and not only suggest it to others in code review :)

PR #1668, opened 2021-05-12
Replace raw prints with the correct logs messages in qemu

ilausuch:
Previously the autoinst-log.txt shows the raw line:
Start CPU

With this commit it shows:
[2021-05-12T12:30:12.407 CEST] [debug] Start CPU

https://progress.opensuse.org/issues/91527

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -1080,10 +1080,10 @@ sub start_qemu {
     }
 
     if ($bmwqemu::vars{DELAYED_START}) {
-        print "DELAYED_START set, not starting CPU, waiting for resume_vm()\n";
+        bmwqemu::diag("DELAYED_START set, not starting CPU, waiting for resume_vm()");
     }
     else {
-        print "Start CPU\n";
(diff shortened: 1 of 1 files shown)


perlpunk (changes requested):
It doesn't compile, and also the base from where you branched seems to be rather old.

But apart from that, wouldn't it be better to use the functions from `bmwqemu.pm` instead of copying the code from there?

ilausuch:
> But apart from that, wouldn't it be better to use the functions from `bmwqemu.pm` instead of copying the code from there?

Done, I was working in an old version

perlpunk (changes requested):
Sorry for not mentioning it explicitly; when I wrote my first comment I was preparing for a meeting:
I think you can use `bmwqemu::diag()` directly.

ilausuch:
Ok, something new I know now

perlpunk (approved):

Martchus (approved):

kalikiana (approved):

PR #1669, opened 2021-05-12
Indent the new lines in bmwqem::log_call

ilausuch:
Previously the new lines appear without indentation. But for a
better reading and parsing the new lines that belongs to a
log entry are indented

```
[2021-05-12T16:34:22.474 CEST] [debug] <<< testapi::assert_screen(mustmatch=[
  "partitioning-edit-proposal-button",
  "before-role-selection",
  "inst-instmode",
  "online-repos"
], timeout=120)
```

This is the new format:

```
[2021-05-12T16:34:22.474 CEST] [debug] <<< testapi::assert_screen(mustmatch=[
    "partitioning-edit-proposal-button",
    "before-role-selection",
    "inst-instmode",
    "online-repos"
  ], timeout=120)
```

https://progress.opensuse.org/issues/91527

PR #1670, opened 2021-05-13
Indent the new lines in logs using the log format

ilausuch:
Previously the new lines appear without indentation. But for a
better reading and parsing the new lines that belongs to a
log entry are indented

```
[2021-05-12T16:34:22.474 CEST] [debug] <<< testapi::assert_screen(mustmatch=[
  "partitioning-edit-proposal-button",
  "before-role-selection",
  "inst-instmode",
  "online-repos"
], timeout=120)
```

This is the new format:

```
[2021-05-12T16:34:22.474 CEST] [debug] <<< testapi::assert_screen(mustmatch=[
    "partitioning-edit-proposal-button",
    "before-role-selection",
    "inst-instmode",
    "online-repos"
  ], timeout=120)
```

http://progress.opensuse.org/issues/91527

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -191,10 +191,16 @@ our $backend;
 # util and helper functions
 
 sub log_format_callback {
-    my ($time, $level, @lines) = @_;
+    my ($time, $level, @items) = @_;
+
+    my $lines = join("\n", @items, '');
+
+    # ensure indentation for multi-line output
+    $lines =~ s/(?<!\A)^/  /gm;
+
     # Unfortunately $time doesn't have the precision we want. So we need to use Time::HiRes
     $time = gettimeofday;
-    return sprintf(strftime("[%FT%T.%%03d %Z] [$level] ", localtime($time)), 1000 * ($time - int($time))) . join("\n", @lines, '');
+    return sprintf(strftime("[%FT%T.%%03d %Z] [$level] ", localtime($time)), 1000 * ($time - int($time))) . $lines;
 }
 
 sub diag {
--- a/t/12-bmwqemu.t
+++ b/t/12-bmwqemu.t
@@ -64,6 +64,12 @@ subtest 'log_call' => sub {
         bmwqemu::log_call("bar\tbaz\rboo\n");
     }
     stderr_like(\&log_call_test_single, qr{\Q<<< main::log_call_test_single("bar\tbaz\rboo\n")}, 'log_call escapes special characters');
+
+    sub log_call_indent {
+        my $lines = ["a", ["b"]];
+        bmwqemu::log_call(test => $lines);
+    }
+    stderr_like(\&log_call_indent, qr{\Q<<< main::log_call_indent(test=[\E\n\Q    "a",\E\n\Q    [\E\n\Q      "b"\E\n\Q    ]\E\n\Q  ])}, 'log_call auto indentation');
 };
 
 subtest 'update_line_number' => sub {


ilausuch:
This is a better solution than https://github.com/os-autoinst/os-autoinst/pull/1669

ilausuch:
There are 3 alternatives for the regular expression

`my $a="hello\nworld\n with one space\n  with two spaces\n";`

Option 1
`$a =~ s/\n(?=\S)/\n  /g;`

```
hello
  world
 with one space
  with two spaces
```

Option 2

`$a =~ s/(?<!\A)^/  /gm;`

```
hello
  world
   with one space
    with two spaces
```

Option 3

`$a =~ s/(?<!\A)^ */  /gm;`

``hello
  world
  with one space
  with two spaces
```

I like the 3th option

**Comment by Martchus** [comment]
I don't get the same results. Looks like there's a white-space missing in the the ` with two spaces` part of your initial assignment of `$a`.

**Comment by ilausuch** [comment]
> There are 3 alternatives for the regular expression
> I like the 3th option

Please give me your oppinions

**Comment by ilausuch** [comment]
I uploaded a output autoinst-log at https://progress.opensuse.org/issues/91527#note-10

**Comment by perlpunk** [comment]
I suggested the first two options, but in retrospective I think we shouldn't alter the number of spaces, just increase by two spaces for every line. There might be reasons for existing spaces, e.g. a YAML document for example :)
So I would vote for option 2, but with 2 spaces.

File: bmwqemu.pm [hunk 1]
```diff
@@ -192,6 +192,9 @@ our $backend;
 
 sub log_format_callback {
     my ($time, $level, @lines) = @_;
+    foreach ( @lines ) {
+        $_ =~ s/(?<!\A)^ */  /gm;
```
perlpunk, line 196: the `$_ =~ ` part is default here, so it can be removed

perlpunk: And maybe `@lines` could be renamed to something like `@items`, because `@lines` is misleading. each "line" can contain multiple lines.

ilausuch: Ah, ok thanks

ilausuch:
> I suggested the first two options, but in retrospective I think we shouldn't alter the number of spaces, just increase by two spaces for every line. There might be reasons for existing spaces, e.g. a YAML document :)
> So I would vote for option 2, but with 2 spaces.

@okurz you voted for the 3th one in the RC chat, but I think the @perlpunk are convincing arguments for me.

okurz:
I agree, I have not understood that detail. Option 2 is better

bmwqemu.pm
```diff
@@ -192,6 +192,9 @@ our $backend;
 
 sub log_format_callback {
     my ($time, $level, @lines) = @_;
+    foreach ( @lines ) {
+        s/(?<!\A)^/  /gm;
+    }
```
okurz, line 197:
```suggestion
    # ensure indentation for multi-line output
    s/(?<!\A)^/  /gm foreach @lines;
```

okurz (changes requested):

ilausuch:
This last commit contains the option 2 and the optimization proposed by @perlpunk
Thanks for give me your opinion

bmwqemu.pm
perlpunk: `s/foreach/for/` :)

perlpunk (approved):

Martchus (commented):
Wouldn't this affect the first line as well? It would be good to have a test.

perlpunk (changes requested):
Oh right @Martchus , a test would be good.

perlpunk:
> Wouldn't this affect the first line as well?

The `@lines` array can contain multiline items. That's why I suggested to rename the array to `@items` for example.

perlpunk:
Also the items of the `@lines` array are joined together, so it can happen that the first line of `$lines[1]` is not indented too.
Probably we would need to join the items together first and then indent. Or split the individual lines and then have a simpler `s///` operation on each line.

perlpunk:
Examples:
```
bmwqemu::diag("Results: " . Dumper($results))
bmwqemu::diag("Results:", Dumper($results))
bmwqemu::diag("Results:", $yaml);
```

okurz (approved):

Martchus:
I don't like that examples like the ones mentioned by @perlpunk now become further intended:

```
  DB<7> bmwqemu::diag('Result: '. Dumper({foo => 1, bar => 2}));
[2021-05-17T12:43:28.052 CEST] [debug] Result: $VAR1 = {
            'bar' => 2,
            'foo' => 1
          };
```

Although the difference isn't actually that big:

```
  DB<6> bmwqemu::diag('Result: '. Dumper({foo => 1, bar => 2}));
[2021-05-17T12:46:07.553 CEST] [debug] Result: $VAR1 = {
          'bar' => 2,
          'foo' => 1
        };
```

However, it makes C&P more difficult as one has to remove the additional spaces again. Not sure whether we actually log any YAML documents but in this case the changed indentation level also seems more disturbing than useful.

I'm wondering whether this is a big improvement anyways. We have the timestamp-tags which already act as a visual anchor to the next log message.

By the way, I currently find the ANSI escape codes the most annoying part of the logs. Not because coloring is a bad idea but because the web UI cannot display them in a good way.

It would also be great to be able to filter out debug messages (e.g. on the web UI side) but then some of the current debug messages should be info messages.

ilausuch:
I commit a change where the lines are join first, and then the indentation

ilausuch:
> However, it makes C&P more difficult as one has to remove the additional spaces again. Not sure whether we actually log any YAML documents but in this case the changed indentation level also seems more disturbing than useful.

Ok, we need a consensus with that. I agree with @perlpunk

> By the way, I currently find the ANSI escape codes the most annoying part of the logs. Not because coloring is a bad idea but because the web UI cannot display them in a good way.

For this case I think is the client reader problem and then it could be solved on the UI. For instance if I open the log with lnav it shows that without the escape codes

image

okurz:
> By the way, I currently find the ANSI escape codes the most annoying part of the logs. Not because coloring is a bad idea but because the web UI cannot display them in a good way.

we do not have a "log viewer", we only provide a download and browsers can choose to show logs verbatim. Compare to gitlab in https://gitlab.suse.de/openqa/auto-review/-/jobs/345765/raw showing the raw log with control codes from both code that comes from us as well as gitlab internal code doing the same in comparison to a nice log viewer https://gitlab.suse.de/openqa/auto-review/-/jobs/345765 that shows color for both parts again, e.g. nice colorful autoinst-log.txt

> It would also be great to be able to filter out debug messages (e.g. on the web UI side) but then some of the current debug messages should be info messages.

yes, would be nice as well but completely out of scope for a PR like this.

As also you brought up the "would be nice …" ideas I created https://progress.opensuse.org/issues/92746 to make the ideas which we actually had already noted down in our tickets already more prominent. I hope we can use the "drive" to be able to cover that soon as well then :)

Martchus:
A log viewer like GitLab's would be nice/convenient. (Of course I can also just use e.g. `curl https://openqa.opensuse.org/tests/1745552/file/autoinst-log.txt` in any terminal.)

By the way, I've only mentioned the other ideas because that's where I would have been started to improve the log. Of course it is out of scope for this PR. If others think the indentation is a plus from them we can add it. I personally just don't see a big benefit considering the mentioned downsides.

okurz (approved):

perlpunk (approved):

PR #1671, opened 2021-05-13
Handle a NEEDLES_DIR that contain a trailing / correctly

radolin:
Currently if the NEEDLES_DIR parameter is supplied with a trailing slash it causes the first character of the filename to be stripped and json needle file is not found.

### DIFF
--- a/needle.pm
+++ b/needle.pm
@@ -64,7 +64,7 @@ sub new {
     # - $jsonfile is re-assigned to contain the absolute path the the JSON file.
     # - The needle must be within the needle directory.
     if (index($jsonfile, $needles_dir) == 0) {
-        $self->{file} = substr($jsonfile, length($needles_dir) + 1);
+        $self->{file} = File::Spec->abs2rel($jsonfile, $needles_dir);
     }
     elsif (-f File::Spec->catfile($needles_dir, $jsonfile)) {
         # json file path already relative


okurz:
Hi @radolin , welcom to this project :fireworks: :partying_face: . Thanks for your contribution. Looks like a sensible approach and would make code even easier to read. I like it :) Let's see what our CI tests will tell. Please correct the git commit subject line by replacing "than" with "that" and delete the dot at the end of line

Martchus (changes requested):
This looks wrong. The existing code makes this a relative path within the needles directory. The new code removes all directories leaving only the file name. Maybe just run `File::Spec->canonpath( $path )` on both paths (`$jsonfile` and `$needles_dir`) before the `if`.

radolin:
Thanks for the warm welcome @okurz and thanks for the review @Martchus! Please have a look at my second attempt, that should keep identical behavior like the previous version.

okurz:
I suggest you extend t/01-test_needle.t accordingly to ensure the code does what you need

Martchus (commented):
This should work now for your case but as @okurz mentions it would be great to extend tests (covering a sub directory and a trailing `/`). Of course it is still broken if a path contains multiple consecutive slashes (and the other doesn't) due to the `index()`-check. Hence I've suggested to run `File::Spec->canonpath()` on both paths before.

Martchus:
@radolin If we improve this we should improve it in a way so it doesn't break on the next best case like having two consecutive slashes so it would be great to update the PR accordingly. Or does my suggestion to use `File::Spec->canonpath()` has any downsides I'm not seeing?

kraih:
Have to agree with @Martchus, `canonpath()` seems reasonable here.

kalikiana:
@radolin Are you planning to update this PR? If you need help, feel free to ask 😃

radolin:
@kalikiana I'm not familiar with the Perl testing framework and will need to catch up on that. I would update it when I have some more free time, but have other priorities now. If you want to take this please feel free.

okurz:
@kraih @Martchus some of you wanting to give it a try, add tests and use `canonpath` or maybe something from `Mojo::File`?

PR #1672, opened 2021-05-14
Find out if qemu is killed because of out-of-memory

Amrysliu:
Related: https://progress.opensuse.org/issues/90974

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -1,4 +1,4 @@
-# Copyright © 2018-2020 SUSE LLC
+# Copyright © 2018-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -405,8 +405,11 @@ sub exec_qemu {
     $process->on(
         collected => sub {
             $self->{_qemu_terminated} = 1;
-            bmwqemu::serialize_state(component => 'backend', msg => 'QEMU exited unexpectedly, see log for details')
-              unless $self->{_stopping};
+            unless ($self->{_stopping}) {
+                my $msg = 'QEMU exited unexpectedly, see log for details';
+                $msg = 'QEMU was killed due to the system being out of memory' if ($self->check_qemu_oom == 0);
+                bmwqemu::serialize_state(component => 'backend', msg => $msg);
+            }
         });
     $process->code(sub {
             $SIG{__DIE__} = undef;    # overwrite the default - just exit
@@ -429,6 +432,8 @@ sub stop_qemu {
 }
 
 sub qemu_pid { shift->_process->process_id }
+
+sub check_qemu_oom { system("$bmwqemu::scriptdir/check_qemu_oom " . shift->qemu_pid); }    # uncoverable statement
 
 =head3 connect_qmp
 
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -11,6 +11,9 @@ use Test::Warnings qw(warnings :report_warnings);
 use Mojo::File qw(tempfile tempdir path);
 use Carp 'cluck';
 use Mojo::Util qw(scope_guard);
+use Mojo::JSON 'decode_json';
+use Test::MockModule;
+use bmwqemu;
 
 use OpenQA::Qemu::BlockDevConf;
 use OpenQA::Qemu::Proc;
@@ -100,11 +103,7 @@ my $proc;
     HDDSIZEGB => 69,
     HDD_1     => "$Bin/data/Core-7.2.iso",
     UEFI      => 1);
-$proc = OpenQA::Qemu::Proc->new()
-  ->_static_params(['-foo'])
-  ->qemu_bin('qemu-kvm')
-  ->qemu_img_bin('qemu-img')
-  ->configure_blockdevs('disk', 'raid', \%vars);
+$proc  = qemu_proc('-foo', \%vars);
 @gcmdl = $proc->gen_cmdline();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu command line for single existing UEFI disk using vars');
 
@@ -132,12 +131,7 @@ is_deeply(\@gcmdl, \@cmdl, 'Generate qemu-img command line for single existing U
     SCSICONTROLLER => 'virtio-scsi-device',
     MULTIPATH      => 1,
     PATHCNT        => 2);
-$proc = OpenQA::Qemu::Proc->new()
-  ->_static_params(['-foo'])
-  ->qemu_bin('qemu-kvm')
-  ->qemu_img_bin('qemu-img')
-  ->configure_controllers(\%vars)
-  ->configure_blockdevs('disk', 'raid', \%vars);
+$proc = qemu_proc('-foo', \%vars);
 
 @gcmdl = $proc->gen_cmdline();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu command line for new drives on multipath');
@@ -177,12 +171,7 @@ ok $proc->init_blockdev_images(), 'init_blockdev_images passes';
     ISO            => "$Bin/data/Core-7.2.iso",
     HDDSIZEGB      => 10,
     SCSICONTROLLER => 'virtio-scsi-device');
-$proc = OpenQA::Qemu::Proc->new()
-  ->_static_params(['-static-args'])
-  ->qemu_bin('qemu-kvm')
-  ->qemu_img_bin('qemu-img')
-  ->configure_controllers(\%vars)
-  ->configure_blockdevs('disk', 'raid', \%vars);
+$proc  = qemu_proc('-static-args', \%vars);
 @gcmdl = $proc->gen_cmdline();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu command line for new drive and cdrom using vars');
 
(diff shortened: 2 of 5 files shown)


Amrysliu:
Blocked by: https://github.com/os-autoinst/openQA/pull/3903

OpenQA/Qemu/Proc.pm
```diff
@@ -405,8 +405,12 @@ sub exec_qemu {
     $process->on(
         collected => sub {
             $self->{_qemu_terminated} = 1;
-            bmwqemu::serialize_state(component => 'backend', msg => 'QEMU exited unexpectedly, see log for details')
-              unless $self->{_stopping};
+            unless ($self->{_stopping}) {
+                my $msg      = 'QEMU exited unexpectedly, see log for details';
+                my $qemu_pid = $self->qemu_pid;
+                $msg = 'QEMU was killed because of out of memory' if (system("dmesg | grep \"Out of memory: Killed process $qemu_pid\"") == 0);
```
okurz, line 411:
```suggestion
                $msg = 'QEMU was killed because of out of memory' if qx{dmesg} =~ qr/Out of memory: Killed process $qemu_pid/;
```

This should make https://github.com/os-autoinst/openQA/pull/3903 easier by only needing to allow dmesg, not needing grep

okurz (changes requested):

OpenQA/Qemu/Proc.pm
Martchus: Or you just use `index()` as we don't use any actual regex matching here.

okurz: Nice idea. But that sounds like less clear code and a premature optimization attempt and I don't even know if index is faster than the regex comparison. We only do that in limited cases and on the openQA worker anyway

Martchus: I've searched about it before and the index function is faster than regex. I'm also wondering why `index()` would be less clear. One wouldn't have to worry about escaping which actually sounds like a plus.

okurz: ok then, regex match as well as index are fine for me

Amrysliu: Thanks, used `index`.

Amrysliu:
@Martchus @okurz Could you give me some suggestions about the test code?
I tried to use `use Test::Mock::Cmd 'qx' => sub { return "[4585996.643455] Out of memory: Killed process 1234" };` to mock the `qx`. But it doesn't work in Proc.qemu.  And ideas? Thanks a lot.

perlpunk:
> @Martchus @okurz Could you give me some suggestions about the test code?
> I tried to use `use Test::Mock::Cmd 'qx' => sub { return "[4585996.643455] Out of memory: Killed process 1234" };` to mock the `qx`. But it doesn't work in Proc.qemu. And ideas? Thanks a lot.

It should work by defining the mock before loading the packages that actually call `qx`.

An alternative approach could be to put an executable shell script `dmesg` into our repo (under `t/`) and manipulate `$ENV{PATH}` in the script.

okurz:
Mocking should work

> An alternative approach could be to put an executable shell script `dmesg` into our repo (under `t/`, or just create a temp file during the test) and manipulate `$ENV{PATH}` in the test.

Before going that route I suggest to put the `qx{dmesg}` call into a single line sub definition that you can mock directly

```perl
sub dmesg { qx{dmesg} }
```

Martchus (approved):

OpenQA/Qemu/Proc.pm
```diff
@@ -405,8 +405,12 @@ sub exec_qemu {
     $process->on(
         collected => sub {
             $self->{_qemu_terminated} = 1;
-            bmwqemu::serialize_state(component => 'backend', msg => 'QEMU exited unexpectedly, see log for details')
-              unless $self->{_stopping};
+            unless ($self->{_stopping}) {
+                my $msg      = 'QEMU exited unexpectedly, see log for details';
+                my $qemu_pid = $self->qemu_pid;
+                $msg = 'QEMU was killed because of out of memory' if (index($self->dmesg, "Out of memory: Killed process $qemu_pid") != -1);
```
okurz, line 411:
```suggestion
                $msg = 'QEMU was killed due to the system being out of memory' if (index($self->dmesg, "Out of memory: Killed process $qemu_pid") != -1);
```

OpenQA/Qemu/Proc.pm
```diff
@@ -430,6 +434,8 @@ sub stop_qemu {
 
 sub qemu_pid { shift->_process->process_id }
 
+sub dmesg { qx{dmesg} }
```
okurz, line 437: hm, I hoped that the sub definition would be coverage enough :) Can you add

```suggestion
sub dmesg { qx{dmesg} }  # uncoverable statement
```

Please check if that is enough to ensure a complete coverage of your changes

t/18-qemu.t
```diff
@@ -402,6 +404,22 @@ subtest 'relative assets' => sub {
...
+
+subtest 'qemu was killed because of out of memory' => sub {
+    my $mock_proc = Test::MockModule->new('OpenQA::Qemu::Proc');
+    $mock_proc->redefine(qemu_pid => sub { return '1234' });
+    $mock_proc->redefine(dmesg    => sub { return '[4585996.643455] Out of memory: Killed process 1234' });
+    $proc = OpenQA::Qemu::Proc->new()
+      ->_static_params(['-foo'])
+      ->qemu_bin('qemu-kvm')
+      ->qemu_img_bin('qemu-img')
+      ->configure_blockdevs('disk', 'raid', \%vars);
```
okurz, line 417: As this is repeated from code above I suggest to extract that, e.g.

```suggestion
sub qemu_proc {
    OpenQA::Qemu::Proc->new()
      ->_static_params(['-foo'])
      ->qemu_bin('qemu-kvm')
      ->qemu_img_bin('qemu-img')
      ->configure_blockdevs('disk', 'raid', \%vars);
}
…
    $proc = qemu_proc;
```

okurz (changes requested):

OpenQA/Qemu/Proc.pm
Amrysliu: fixed

OpenQA/Qemu/Proc.pm
Amrysliu: Added

okurz (changes requested):
This looks all good but unfortunately we learned something new in the according openQA PR. So instead of the `dmesg` function better create a separate script which you mock in tests and install as part of the package so that we can only give the apparmor permission to call "dmesg" from that script.

Amrysliu:
> This looks all good but unfortunately we learned something new in the according openQA PR. So instead of the `dmesg` function better create a separate script which you mock in tests and install as part of the package so that we can only give the apparmor permission to call "dmesg" from that script.

Ok. I will do that :)
Could you help me confirm some questions:
1. modify `usr.share.openqa.script.worker` by adding such as:
    ```
       profile /usr/lib/os-autoinst/check_qemu_oom {
           #include <abstractions/base>
           /usr/bin/dmesg rix,
       }
    ```
2. add a script on `os-autoinst` repo to get the `dmesg` result. But where I should put it after the rpm installed? `/usr/bin/` or `/usr/lib/os-autoinst`?

okurz:
> > This looks all good but unfortunately we learned something new in the according openQA PR. So instead of the `dmesg` function better create a separate script which you mock in tests and install as part of the package so that we can only give the apparmor permission to call "dmesg" from that script.
> 
> Ok. I will do that :)
> Could you help me confirm some questions:
> 
>     1. modify `usr.share.openqa.script.worker` by adding such as:
>        ```
>           profile /usr/lib/os-autoinst/check_qemu_oom {
>               #include <abstractions/base>
>               /usr/bin/dmesg rix,
>           }
>        ```

yes, this looks correct.

>     2. add a script on `os-autoinst` repo to get the `dmesg` result. But where I should put it after the rpm installed? `/usr/bin/` or `/usr/lib/os-autoinst`?

`/usr/lib/os-autoinst`

t/18-qemu.t
```diff
@@ -391,17 +374,22 @@ subtest DriveDevice => sub {
...
+    $proc = qemu_proc('-foo', \%vars);
     my @gcmdl = $proc->blockdev_conf->gen_qemu_img_cmdlines();
     @cmdl = map { [qw(create -f qcow2 -b), "$dir/Core-7.2.iso", "raid/$_-overlay0", 11116544] } qw(hd0 cd0 cd1);
     push @cmdl, ["create", "-f", "qcow2", "-b", "$Bin/data/uefi-code.bin", "raid/pflash-code-overlay0", 1966080];
     push @cmdl, ["create", "-f", "qcow2", "-b", "$dir/Core-7.2.iso",       "raid/pflash-vars-overlay0", 11116544];
     is_deeply(\@gcmdl, \@cmdl, 'find the asset real path');
+};
+
+subtest 'qemu was killed due to the system being out of memory' => sub {
+    $bmwqemu::scriptdir = "$Bin/fake/tests";
```
okurz, line 386: This is an interesting approach but it feels like bit of an overhead of having a separate fake script. Also feels like cheating on the code coverage infrastructure. How about keeping the approach of having a single-line sub that can be mocked? In that single-line sub include the call to the external script and mark the sub with `# uncoverable statement`. Maybe better if that works: Mock `system`, e.g. `*CORE::system` or something

okurz (changes requested):

t/18-qemu.t
Amrysliu: I tried Mock system in test code, but it didn't work.  So I changed that to a single-line sub.

okurz (approved):

Martchus:
My other review comment (https://github.com/os-autoinst/openQA/pull/3903#issuecomment-849427476) hasn't been answered so I'm adding the not-ready label to be on the safe side we didn't overlook something.

Martchus (approved):

Amrysliu:
> My other review comment ([os-autoinst/openQA#3903 (comment)](https://github.com/os-autoinst/openQA/pull/3903#issuecomment-849427476)) hasn't been answered so I'm adding the not-ready label to be on the safe side we didn't overlook something.

https://github.com/os-autoinst/openQA/pull/3903#issuecomment-851943899

Martchus:
Ok, feel free to remove the label then.

PR #1674, opened 2021-05-18
Fix test execution - prevent duplicate running of tests in the wrong directory

BillAnastasiadis:
Fixed script running perl tests in the `t` and `xt` directories. When the user gives specific tests to be run, they need to provide the directory of the test as well. Now each test is only run once and doesn't fail trying to execute in both directories.

- Related issue: https://progress.opensuse.org/issues/92497

### DIFF
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -52,5 +52,10 @@ export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
 export OS_AUTOINST_MAKE_TOOL=$make_path
 
 # invoke tests within the t and xt directory
-cd t && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
-cd ../xt && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
+if [[ $TESTS = *[!\ ]* ]]
+then
+	"$prove_path" $PROVE_ARGS $TESTS
+else
+	cd t && "$prove_path" $PROVE_ARGS -r
+	cd ../xt && "$prove_path" $PROVE_ARGS -r
+fi
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -256,8 +256,11 @@ make test-perl-testsuite
 
 Run individual tests by specifying them explicitly:
 ----
-make test-perl-testsuite TESTS="15-logging.t 28-signalblocker.t"
+make test-perl-testsuite TESTS="t/15-logging.t t/28-signalblocker.t"
 ----
+
+Notice that the user needs to include the test directory for each test (either t for normal or 
+xt for developer-centric tests) when specifying individual tests.
 
 Add additional arguments to the `prove` invocation, e.g. enable verbose output:
 ----


okurz (changes requested):
Woah. I would appreciate a solution which is less verbose and hopefully also without the need to iterate over individual arguments. I assume also something like "echo" is for debugging only? I recommend to instead execute a shell script with the argument "-x" to show each argument before execution

BillAnastasiadis:
> Woah. I would appreciate a solution which is less verbose and hopefully also without the need to iterate over individual arguments. I assume also something like "echo" is for debugging only? I recommend to instead execute a shell script with the argument "-x" to show each argument before execution

Fixed

tools/invoke-tests
```diff
@@ -51,6 +51,24 @@ fi
...
 
-# invoke tests within the t and xt directory
-cd t && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
-cd ../xt && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
+# prepare test arguments for the t directory
+cd t
+argt=""
+for word in ${TESTS:-'-r'}
+do
+    if [[ $(ls | grep -x $word) ]]; then argt="${argt}$word "; fi
```
Martchus, line 59: Calling `ls | grep …` in a loop seems bad performance-wise. The code also seems wrong because `argt` is only set but never used.

Maybe it makes more sense if the caller specifies the subdirectory as well, e.g. `TESTS='xt/foo.t t/bar.t'`. This would at least ease the manual splitting a little bit. Or we simply require two different invocations of this script and the correct directory needs to be passed as CLI argument (maybe defaulting to `t`). The required changes on the CMake-side shouldn't be too challenging (at least better than this).

Martchus (changes requested):

perlpunk:
I think our long term goal was to call the tests from the main directory (which would include the directory name), as this is the usual workflow for perl tests anyway.
But we are still having problems with that. I was able to get the tests running, but the coverage was lower.

So I would vote for including the directory name, as this is what we would finally want to have anyway.

tools/invoke-tests
BillAnastasiadis: Nice catch, corrected. This pr is complete now, at least as far as the initial implementation is concerned, so I'm removing the WIP.

As for the suggestions: the first one is trivial. I considered that the exact original intended behavior a requirement, thus the current implementation . The second one would not let the user run tests in both directories in a single run, should they want to, and wouldn't be my first choice.

If you agree, I have the implementation of your first suggestion ready. The change should also be reflected in the documentation. I'll be waiting for the ok to push it.

BillAnastasiadis: Nice catch, corrected. This pr is complete now, at least as far as the initial implementation is concerned.

As for the suggestions: the first one is trivial. I considered that the exact original intended behavior a requirement, thus the current implementation . The second one would not let the user run tests in both directories in a single run, should they want to, and wouldn't be my first choice.

Since @perlpunk agrees as well, I will implement your first suggestion.

Martchus: > I considered that the exact original intended behavior a requirement, thus the current implementation .

I only created this script to avoid redundancy between the autotools-based and CMake-based build script when porting to CMake. It is not a requirement that this script behaves like before as long as the instructions in the README for executing tests still work.

> The change should also be reflected in the documentation.

I consider this file an implementation detail and would rather encourage people to use the build system targets or prove directly.

Makefile
```diff
@@ -3,7 +3,7 @@
 
 .PHONY: all
 all: build/CMakeCache.txt
-	ninja -C build/ symlinks
+ 	ninja -C build/ symlinks
```
okurz, line 6: unrelated change

okurz (changes requested):

Makefile
BillAnastasiadis: Thanks, fixed.

tools/invoke-tests
BillAnastasiadis: I implemented the first solution, with a workaround for the coverage statistics.

perlpunk (approved):

tools/invoke-tests
```diff
@@ -52,5 +52,10 @@ export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
 export OS_AUTOINST_MAKE_TOOL=$make_path
 
 # invoke tests within the t and xt directory
-cd t && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
-cd ../xt && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
+if [ -z $TESTS ]
```
okurz, line 55: `-z` is like a negative check so we have double negation. I suggest to turn that around and use `-n` and use the current "else"-path first

tools/invoke-tests
```diff
@@ -52,5 +52,10 @@ export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
...
+then
+        cd t && "$prove_path" $PROVE_ARGS -r
+        cd ../xt && "$prove_path" $PROVE_ARGS -r
+else
+        "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
```
okurz, line 60: here you shouldn't need to evaluate TESTS anymore, right? So better

```perl
"$prove_path" $PROVE_ARGS $TESTS
```

tools/invoke-tests
BillAnastasiadis: Thank you for the suggestion, I could in the same way not use -`z` or `-n` at all ( `[ $TESTS]` or `[[  $TESTS ]]` would work fine). However, may I suggest `[[ $TESTS = *[!\ ]* ]]`? Unlike all other options, it will behave correctly even in the case the user provides an empty or whitespace $TESTS variable.

tools/invoke-tests
```diff
@@ -52,5 +52,10 @@ export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
 export OS_AUTOINST_MAKE_TOOL=$make_path
 
 # invoke tests within the t and xt directory
-cd t && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
-cd ../xt && "$prove_path" $PROVE_ARGS ${TESTS:-'-r'}
+if [[ $TESTS = *[!\ ]* ]]
+then
+	"$prove_path" $PROVE_ARGS $TESTS
```
Martchus, line 57: Don't mix tabs and spaces. This file uses 4 spaces for an indentation level so far so let's keep it this way.

Martchus (changes requested):
And please update the examples within `README.md`.

tools/invoke-tests
BillAnastasiadis: I noticed that too, fixed

BillAnastasiadis:
> And please update the examples within `README.md`.

Done!

Martchus (approved):

okurz (approved):

tools/invoke-tests
BillAnastasiadis: Sorry about that! wrong `.vimrc` :sweat:
Thank you for taking care of that @okurz

PR #1675, opened 2021-05-18
virtio_console: Store activated state on (save_|load_)snapshot()

cfconrad:
The activated state is used by test modules to discover if a login
prompt is expected or not. This state need to be recovered when loading
a snapshot.

Thanks to @mloviska  we found this solution together as a it address the same issue like https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/11405 .

Ticked: 
Bug's: 
  * https://bugzilla.suse.com/show_bug.cgi?id=1184917#c5
VR's:
  * http://cfconrad-vm.qa.suse.de/tests/8140 (without that change: http://cfconrad-vm.qa.suse.de/tests/8137)

### DIFF
--- a/consoles/virtio_terminal.pm
+++ b/consoles/virtio_terminal.pm
@@ -87,21 +87,35 @@ sub disable {
 sub save_snapshot {
     my ($self, $name) = @_;
 
-    if (defined($self->{screen})) {
-        $self->{snapshots}->{$name} = $self->{screen}->peak();
-    } else {
-        $self->{snapshots}->{$name} = '';
-    }
+    $self->set_snapshot($name, 'activated', $self->{activated});
+    $self->set_snapshot($name, 'buffer',    $self->{screen} ? $self->{screen}->peak() : $self->{preload_buffer});
 }
 
 sub load_snapshot {
     my ($self, $name) = @_;
 
+    $self->{activated} = $self->get_snapshot($name, 'activated') // 0;
+    my $buffer = $self->get_snapshot($name, 'buffer') // '';
     if (defined($self->{screen})) {
-        $self->{screen}->{carry_buffer} = $self->{snapshots}->{$name};
+        $self->{screen}->{carry_buffer} = $buffer;
     } else {
-        $self->{preload_buffer} = $self->{snapshots}->{$name};
+        $self->{preload_buffer} = $buffer;
     }
+}
+
+sub get_snapshot {
+    my ($self, $name, $key) = @_;
+    return undef unless defined($name);
+
+    my $snapshot = $self->{snapshots}->{$name};
+    return (defined($key) && $snapshot) ? $snapshot->{$key} : $snapshot;
+}
+
+sub set_snapshot {
+    my ($self, $name, $key, $value) = @_;
+    return undef if (!defined($name) || !defined($key));
+
+    $self->{snapshots}->{$name}->{$key} = $value;
 }
 
 =head2 F_GETPIPE_SZ
--- a/t/10-virtio_terminal.t
+++ b/t/10-virtio_terminal.t
@@ -2,6 +2,7 @@
 # Copyright © 2020 SUSE LLC
 
 use Test::Most;
+use Test::Warnings qw(:all :report_warnings);
 
 # OpenQA::Test::TimeLimit not used as `prepare_pipes` defines an ALRM handler
 # internally already
@@ -16,9 +17,9 @@ use bmwqemu;
 
 sub prepare_pipes
 {
-    my ($socket_path) = @_;
-    my $pipe_in       = $socket_path . ".in";
-    my $pipe_out      = $socket_path . ".out";
+    my ($socket_path, $write_buffer) = @_;
+    my $pipe_in  = $socket_path . ".in";
+    my $pipe_out = $socket_path . ".out";
 
     for (($pipe_in, $pipe_out)) {
         unlink $_ if (-e $_);
@@ -37,6 +38,8 @@ sub prepare_pipes
           or die "Can't open in pipe for writing $!";
         open(my $fd_w, ">", $pipe_out)
           or die "Can't open out pipe for reading $!";
+
+        syswrite($fd_w, $write_buffer) if ($write_buffer);
 
         sysread($fd_r, my $buf, 1024) while ($running);
         exit 0;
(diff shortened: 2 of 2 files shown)


cfconrad:
Set to WIP as tests are needed, but the idea is to start with VR's already.

t/10-virtio_terminal.t
```diff
@@ -111,5 +114,73 @@ subtest "Test open_pipe() error condition" => sub {
...
+    $term->load_snapshot('snap3');
+    is_deeply($term->screen()->peak(), 'foo', '[snap3] carry over buffer sucessful loaded');
+    is($term->{activated}, 0, '[snap3] console is not activated');
+
+    cleanup_pipes($helper);
+
+};
+
+subtest "virtio_terminal snapshot without screen" => sub {
+    $bmwqemu::logger = Mojo::Log->new(level => 'error'); # hide debug messages withhin this test
```
Martchus, line 163:
```suggestion
    $bmwqemu::logger = Mojo::Log->new(level => 'error'); # hide debug messages within this test
```

t/10-virtio_terminal.t
```diff
@@ -111,5 +114,73 @@ subtest "Test open_pipe() error condition" => sub {
...
+
+    my $socket_path    = './virtio_console_open_test';
+    my $helper = prepare_pipes($socket_path, 'This should not be readed');
```
Martchus, line 166:
```suggestion
    my $helper = prepare_pipes($socket_path, 'This should not be read');
```

t/10-virtio_terminal.t
```diff
@@ -111,5 +114,73 @@ subtest "Test open_pipe() error condition" => sub {
...
+    my $term   = consoles::virtio_terminal->new('unit-test-console', {socked_path => $socket_path});
+    $term->{preload_buffer} = 'test';
+
+    $term->save_snapshot('snap1');
+    is($term->snapshot('snap1', 'buffer'), 'test', 'virtio_terminal preload_buffer stored');
+    is($term->snapshot('snap1', 'activated'), 0, 'console snapshot is not activated');
+
+    $term->select();
+    $term->screen()->read_until('This should not be readed', 60);
```
Martchus, line 175:
```suggestion
    $term->screen()->read_until('This should not be read', 60);
```

Martchus (changes requested):

Martchus:
But I guess the change is generally good.

consoles/console.pm
```diff
@@ -85,6 +86,29 @@ sub select {
...
+
+sub snapshot {
+    my ($self, $name, $key, $value) = @_;
+    return undef unless defined($name);
+
+    if (defined($key) && defined($value)) {
+        $self->{snapshots}->{$name} //= {};
+        $self->{snapshots}->{$name}->{$key} = $value;
+    } elsif (defined($key)) {
+        return ($self->{snapshots}->{$name}) ? $self->{snapshots}->{$name}->{$key} : undef;
```
mloviska, line 107: Shouldn't we check for existence of `$name` in snapshots in the elsif branch?

consoles/console.pm
```diff
@@ -85,6 +86,29 @@ sub select {
...
+    my ($self, $name) = @_;
+    $self->{activated} = $self->snapshot($name, 'activated') // 0;
+}
+
+sub snapshot {
+    my ($self, $name, $key, $value) = @_;
+    return undef unless defined($name);
+
+    if (defined($key) && defined($value)) {
+        $self->{snapshots}->{$name} //= {};
```
okurz, line 104: you should actually not need that, see https://perlmaven.com/autovivification . Just try to access hash entries without specifying the intermediate layers

consoles/console.pm
cfconrad: its in line 101, right?

mloviska: That is only whether the argument is `undef`, but the key might still be missing in snapshots sub hash.

consoles/console.pm
cfconrad: Thx for the hint, wasn't aware of autovivification!

consoles/console.pm
cfconrad: Sorry I misunderstood you. I check `$self->{snapshots}->{$name}` in that ternary operator.

mloviska: ok, so explicit test is not needed. https://perldoc.perl.org/functions/exists#exists-EXPR  thanks

consoles/console.pm
```diff
@@ -85,6 +86,29 @@ sub select {
...
+    my ($self, $name) = @_;
+    $self->snapshot($name, 'activated', $self->{activated});
+}
+
+sub load_snapshot {
+    my ($self, $name) = @_;
+    $self->{activated} = $self->snapshot($name, 'activated') // 0;
+}
+
+sub snapshot {
```
Vogtinator, line 99: It's not really obvious what this function actually does based on the name and there's not a comment either.

The functionality is mostly split into `get` and `set` paths anyway,  so what about having separate functions for those?

cfconrad: Sure I can split it. I had this `has` style in my mind like: https://metacpan.org/pod/Class::Accessor#SYNOPSIS
Actually I'm not sure what is the best practice in perl for `getter/setter` but I thought it is it. Do we still wanna have it splitted?

Also I'm not sure why we use `Class::Accessor` here and not `Mojo::Base`... but I think this is a topic for a other PR.

consoles/virtio_terminal.pm
```diff
@@ -86,21 +85,20 @@ sub disable {
...
     my ($self, $name) = @_;
+    $self->SUPER::load_snapshot($name);
 
+    my $buffer = $self->snapshot($name, 'buffer') // '';
     if (defined($self->{screen})) {
-        $self->{screen}->{carry_buffer} = $self->{snapshots}->{$name};
+        $self->{screen}->{carry_buffer} = $buffer;
     } else {
-        $self->{preload_buffer} = $self->{snapshots}->{$name};
+        $self->{preload_buffer} = $buffer;
```
okurz, line 101: also see this report about missing coverage of changed lines

okurz (changes requested):

consoles/console.pm
Vogtinator: The issue is that it's not just a plain getter/setter for an attribute, but is actually passed two parameters as well. That makes it look much more like a regular function call than getter or setter.

So IMO the code would be shorter and easier to understand by having separate methods.

And a step further, the "make a function if you do it more than once" rule can be applied in reverse: Both setter and getter are only called once, so it could even be inlined completey.

consoles/virtio_terminal.pm
cfconrad: I get `unauthorized` if I click on https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/1675 (the coverage picture), do I do something wrong? Or could I just get read access?

consoles/console.pm
cfconrad: ack, I will change that.

consoles/virtio_terminal.pm
Martchus: For me it works but one has to login on codecov (via GitHub) first.

cfconrad: hm ok, now it is working for me as well. Maybe I wasn't logged in...

cfconrad:
I'm a bit puzzled about the effect of this change, because I thought from the beginning that each console should store it's `activate` state per snapshot. If I now look at the dependencies I'm getting doubts.

Who is affected by this change:
### Backends:
  * qemu (call of `backend::baseclass::load_console_snapshots()`)
 
### Consoles
  * consoles/console.pm
    * consoles/sshVirtshSUT.pm
    * consoles/amtSol.pm
    * consoles/ipmiSol.pm
    * consoles/ttyConsole.pm
    * consoles/sshSerial.pm
    * consoles/virtio_terminal.pm
    * consoles/network_console.pm
      * consoles/sshIucvconn.pm
      * consoles/vnc_base.pm

At least the `sshSerial` looks like a candidate who might not work with this change. As I think a `load_snapshot()` will break the current ssh-connection and it isn't recovered because `activate()` will not be called again, or?

I just guess that most of the console implementations are not aware of snapshots at all.

So maybe it is better to not store the `activate` state always but doing it only in `virtio_terminal` for now.

WDYT?

cfconrad:
> I'm a bit puzzled about the effect of this change, because I thought from the beginning that each console should store it's `activate` state per snapshot. If I now look at the dependencies I'm getting doubts.
> <snip>

I changed it to `virtio_terminal` only! So from my side this PR is done.

Just for reference, this is the old way of doing it https://github.com/os-autoinst/os-autoinst/commit/3e07477198b342bb90f0fc0c6cd15da841ee253b

VR with latest version:
 * http://cfconrad-vm.qa.suse.de/tests/8191

consoles/virtio_terminal.pm
```diff
@@ -87,23 +87,37 @@ sub disable {
...
     }
 }
 
+sub get_snapshot {
```
Vogtinator, line 106: When `$key` is `undef`, `get_snapshot($name)` returns the entire snapshot, but `set_snapshot($name)` is a noop. Shouldn't be an issue in practice, but might be a trap.

Vogtinator:
There is some weirdness in the VR, but I don't think it's caused by this VR:

```
[2021-05-28T23:55:09.818 CEST] [debug] <<< testapi::select_console(testapi_console="x11")
[37m[2021-05-28T23:55:10.226 CEST] [debug] activate_console, console: x11, type: 
[0m[37m[2021-05-28T23:55:10.226 CEST] [debug] activate_console called with generic type, no action
```

The `x11` console has no type set?

Martchus (approved):

cfconrad:
> The x11 console has no type set?

Looks like, didn't checked the code but a other random job: https://openqa.suse.de/tests/5990438/file/autoinst-log.txt

```
[37m[2021-05-11T00:22:06.711 CEST] [debug] activate_console, console: x11, type: 
[0m[37m[2021-05-11T00:22:06.711 CEST] [debug] activate_console called with generic type, no action

File: consoles/virtio_terminal.pm (hunk 8 above)
Comment (cfconrad) [reply consoles/virtio_terminal.pm:106]: I thought it might be good to get the full snapshot for what ever reason...

Anyway, if we think these setter/getters are over engineering I can also remove them. The reason why I kept it was:
* if we will implement snapshot in other consoles, it could help to move it back to `consoles::console`
* I didn't need to touch the tests again (no strong reason, I know :)

Comment (Vogtinator) [reply consoles/virtio_terminal.pm:106]: IMO the getters and setters are fine, but IMO it should just assume the key is always defined.

**Review by okurz — APPROVED** [review]

**Comment by okurz** [comment]
@cfconrad your change causes failures within OBS checks. please check https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Tumbleweed/x86_64

```
[  513s] 3: [33m[0m[33m[0m[33m[0m
[  513s] 3:     #   Failed test '[snap1] virtio_terminal stored all available data'
[  513s] 3:     #   at ./10-virtio_terminal.t line 136.
[  513s] 3:     #          got: ''
[  513s] 3:     #     expected: 'Test data foo'
[  513s] 3: 
[  513s] 3:     #   Failed test '[snap1] Snapshots data verified'
[  513s] 3:     #   at ./10-virtio_terminal.t line 139.
[  513s] 3:     #     Structures begin differing at:
[  513s] 3:     #          $got->{buffer} = ''
[  513s] 3:     #     $expected->{buffer} = 'Test data foo'
[  513s] 3: [33m[0m[33m[0m[33m[0m[33m[0m[33m[0m
[  513s] 3:     #   Failed test '[snap1] carry over buffer sucessful loaded'
[  513s] 3:     #   at ./10-virtio_terminal.t line 153.
[  513s] 3:     #          got: ''
[  513s] 3:     #     expected: 'Test data foo'
[  514s] 3: [33m[0m    # Looks like you failed 3 tests of 22.
[  514s] 3: 
[  514s] 3: #   Failed test 'Test snapshot handling'
[  514s] 3: #   at ./10-virtio_terminal.t line 169.
[  514s] 3: # Looks like you failed 1 test of 3.
[  514s] 3: [04:17:15] ./10-virtio_terminal.t ................... 
[  514s] 3: Dubious, test returned 1 (wstat 256, 0x100)
[  514s] 3: Failed 1/3 subtests 
```

PR #1676, opened 2021-05-19
Doc fix in backend vars

radolin:
Fix messed up table formatting and add note about GENERAL_HW_CMD_DIR variable ignored by openQA.

### DIFF
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -229,7 +229,6 @@ ISO;string;undef;isos from nfs mount on VIO side to VMLibrary
 VIOISO;string;undef;Virtual Optical Media ISO
 VNC;integer;undef;VNC port
 WORKER_ID;string;undef;osauto id
-
 |====================
 
 .PVM_HMC backend
@@ -238,6 +237,7 @@ WORKER_ID;string;undef;osauto id
 |====================
 Variable;Values allowed;Default value;Explanation
 HMC_USERNAME;string;;Username for HMC
+|====================
 
 .GENERAL_HW backend
 [grid="rows",format="csv"]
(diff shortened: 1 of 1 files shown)


Martchus (approved):

doc/backend_vars.asciidoc
```diff
@@ -238,14 +237,15 @@ WORKER_ID;string;undef;osauto id
...
+|====================
 
 .GENERAL_HW backend
 [grid="rows",format="csv"]
 [options="header",cols="^m,^m,^m,v",separator=";"]
 |====================
 Variable;Values allowed;Default value;Explanation
 GENERAL_HW_VNC_IP;string;;Hostname of the gadget's network. If not set, SSH consoles will be used
-GENERAL_HW_CMD_DIR;string;;Directory with allowed CMD scripts
+GENERAL_HW_CMD_DIR;string;;Directory with allowed CMD scripts. Note: This variable should be set in the workers.ini file, otherwise it will be ignored by openQA.
```
okurz, line 248: Interesting. Can you provide more information about that? There is the line https://github.com/os-autoinst/openQA/blob/931b86d3f088ff0f0da0cd3d5b3e8c435b872015/lib/OpenQA/Worker/Job.pm#L238

This was introduced in c200f277c

@colo can you explain why you did that? The comment says "block the job from having dangerous settings (isotovideo specific though) it needs to come from worker_settings" but I don't understand why that specific variable needs to be handled in that special manner

okurz (changes requested):

doc/backend_vars.asciidoc
coolo: Because I wanted to ensure that the worker has the scripts in a place where they belong before enabling it for tests. Just as the SUT_IP should come from workers settings and not from webui settings. 

But the overall mentality about these things changed in the time since I'd say. Nowadays it seems okay to run random code on the worker - in doubt through podman.

okurz: exactly, all is lost :)

Ok but then the change is fine this way and only the external video encoder does something similar so we can accept the PR as is

okurz (approved):

PR #1679, opened 2021-05-26
Use black to check python code style

mimi1vx:
https://progress.opensuse.org/issues/92665

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -8,9 +8,9 @@
 #######################################################
 ---
 targets:
-  cpanfile: [                    cover, devel,         main,             lint,           test_base, test, test_version_only ]
-  spec:     [ build_base, build,        devel,         main, spellcheck,       yamllint, test_base, test, test_version_only ]
-  docker:   [ build_base, build,        devel, docker, main, spellcheck, lint, yamllint, test_base, test                    ]
+  cpanfile: [                    cover, devel,         main,             lint,           test_base, test, test_version_only               ]
+  spec:     [ build_base, build,        devel,         main, spellcheck,       yamllint, test_base, test, test_version_only, python_style ]
+  docker:   [ build_base, build,        devel, docker, main, spellcheck, lint, yamllint, test_base, test,                    python_style      ]
   cpanfile-targets:
     # target: cpanfile target type (default main)
     devel: devel
@@ -41,6 +41,7 @@ cover_requires:
 
 devel_requires:
   '%test_requires':
+  '%python_style_requires':
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
   perl(Perl::Tidy): '== 20210402'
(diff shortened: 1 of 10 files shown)


dist/rpm/os-autoinst.spec
```diff
@@ -62,12 +62,23 @@ Source0:        %{name}-%{version}.tar.xz
...
+%else
+%define python_style_requires %{nil}
+%endif
 # The following line is generated from dependencies.yaml
 %define test_base_requires %main_requires cpio perl(Benchmark) perl(Devel::Cover) perl(FindBin) perl(Pod::Coverage) perl(Test::Fatal) perl(Test::Mock::Time) perl(Test::MockModule) perl(Test::MockObject) perl(Test::Mojo) perl(Test::Most) perl(Test::Output) perl(Test::Pod) perl(Test::Strict) perl(Test::Warnings) >= 0.029 procps python3-setuptools qemu qemu-tools qemu-x86
 # The following line is generated from dependencies.yaml
 %define test_version_only_requires perl(Mojo::IOLoop::ReadWriteProcess) >= 0.28
 # The following line is generated from dependencies.yaml
-%define test_requires %build_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(Inline::Python) perl(YAML::PP)
+%define test_requires %build_requires %python_style_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(Inline::Python) perl(YAML::PP)
```
Martchus, line 81: I don't think we need to run style checks during the package build. Better add this only to `devel_requires` like it is done for `perl(Perl::Tidy)`.

Martchus (changes requested):
Looks good apart from the point about the spec file.

dist/rpm/os-autoinst.spec
mimi1vx: tru, thanks

okurz (approved):

kalikiana (approved):

dependencies.yaml
```diff
@@ -61,6 +62,9 @@ lint_requires:
   perl(Perl::Critic::Policy):
   perl(Perl::Critic::Utils):
 
+python_style_requires:
```
Martchus, line 65: Note that adding this simply to `devel_requires` would have the same effect but would be simpler.

perlpunk: It's probably not necessary at all because `python_style_requires` is defined dynamically in the spec here:
`%if %{with black}`

perlpunk: oh wait, sorry, to appear in the Dockerfile, this *is* actually needed

mimi1vx: @Martchus  `black` is unfortunately only packed in Tumbleweed so needs a small difference from standard `devel_requires`

Martchus (approved):

PR #1681, opened 2021-05-27
Improve documentation of CASEDIR/PRODUCTDIR/NEEDLES_DIR

Martchus:
* Improve wording
* State what relative paths are relative to
* State whether variable is mandatory or defaults value


doc/backend_vars.asciidoc
```diff
@@ -6,9 +6,9 @@ Supported variables per backend
...
-NEEDLES_DIR;string;;Path to needles subdirectory to use, defaults to "needles" within `PRODUCTDIR`. Can be a git repository URL, comparable to `CASEDIR`
+CASEDIR;string;;Absolute or relative path to test distribution. Specifying the variable is mandatory. Can also be a Git repository URL of a test distribution to checkout with an optional refspec or Git commit hash into the current directory. It tries to follow the definition of https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies (e.g. `https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#feature/test` or `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git`). Note that cloning via SSH is only possible if Git is configured accordingly on the host where isotovideo is executed. Can be combined with `TEST_GIT_REFSPEC` and `NEEDLES_GIT_REFSPEC`. Relative paths are considered relative to the working directory of `isotovideo`.
+PRODUCTDIR;string;;Absolute or relative path the "product directory" which includes the test schedule entry point `main.pm` as well as a `needles` subdirectory with the needles to load. Defaults to `CASEDIR`. Relative paths are considered relative to `CASEDIR`.
```
okurz, line 10: I think the whole path is still "optional" as AFAIK it will not be sweet in vars.json if not needed.

Also it must say "path *to* the"

Amrysliu: `CASEDIR` is always needed, right?

Martchus: You're talking about the `RODUCTDIR`, right? I left out the optional because I thought it was obvious when there's a default.

okurz: yes, `PRODUCTDIR`. I prefer to keep the "optional" if you are ok with that

Martchus: changed

okurz (approved):

PR #1683, opened 2021-05-28
Add shell style checks based on ShellCheck

okurz:


cmake/test-targets.cmake
```diff
@@ -37,6 +37,17 @@ add_test(
     WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
 )
 
+find_program(SHELLCHECK_PATH shellcheck)
+if (SHELLCHECK_PATH)
+    add_test(
+        NAME test-local-shellcheck
+        COMMAND shellcheck -x $(file --mime-type script/* t/* container/worker/*.sh tools/* | sed -n 's/^\(.*\):.*text\/x-shellscript.*$/\\1/p')
```
Martchus, line 44: SHELLCHECK_PATH should be used there.

tools/check-shellcheck
```diff
@@ -0,0 +1 @@
+shellcheck -x $(file --mime-type tools/* | sed -n 's/^\(.*\):.*text\/x-shellscript.*$/\1/p')
```
Martchus, line 1: You've been adding the command directly withing CMake so this looks like a leftover.

Martchus (changes requested):

tools/check-yaml-syntax
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh -e
 # fall back to find if there is no git, e.g. in package builds
-"${1:-yamllint}" --strict $(git ls-files "*.yml" "*.yaml" 2>/dev/null || find -name '*.y*ml')
+"${1:-yamllint}" --strict "$(git ls-files "*.yml" "*.yaml" 2>/dev/null || find . -name '*.y*ml')"
```
BillAnastasiadis, line 3: Is quoting the command substitution the intended behavior?  Looking at the ci error

`No such file or directory: '.github/stale.yml\n.github/workflows/ci.yml\n.github/workflows/ci_extended.yml\n.hadolint.yaml\n.mergify.yml\ncodecov.yml\ndependencies.yaml'`

it seems that the command needs the splitted `$()` output and not the single string `"$()"` one.

cmake/test-targets.cmake
okurz: I needed to change that now similar as we do for the python and yaml style check by passing the SHELLCHECK_PATH as first argument to the check script

tools/check-shellcheck
okurz: Actually I first tried to get the quoting right in CMake but quicky gave up after my nightmares from 15 years ago about CMake shell script quoting bubbled up again :D So now I am calling this script from cmake

tools/check-yaml-syntax
okurz: yes, of course. I fixed that and changed accordingly same as I did for the python style check that was added in the meantime

Martchus (approved):

PR #1688, opened 2021-06-14
Correct user -> username in sshSerial::activate

dcermak:
The activate subroutine of sshSerial was using the hash key 'user' to obtain the
username, but the sub get_ssh_credentials uses 'username' in all other places
instead.


Martchus (approved):
Looks like we're using `username` everywhere else indeed.

PR #1689, opened 2021-06-15
Allow to specify the port for a ssh connection

dcermak:
Currently ssh connections always use port 22. But for vagrant + virtualbox we
must be able to change this to a different value, as this backend uses ports
2200 and higher.
For that we extend `get_ssh_credentials` to also allow to return a `port` hash
key which is then forwarded to `ssh->connect()`.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1244,13 +1244,14 @@ sub new_ssh_connection {
     my %credentials = $self->get_ssh_credentials;
     $args{$_} //= $credentials{$_} foreach (keys(%credentials));
     $args{username} ||= 'root';
+    $args{port}     ||= 22;
     $args{keep_open} //= 0;
     my $connection_key;
 
     # e.g. using hyperv_intermediate host which is running Windows need to keep the connection.
     # Otherwise a mount point doesn't exists within the next command.
     if ($args{keep_open}) {
-        $connection_key = join(',', map { $_ . "=" . $args{$_} } qw(hostname username));
+        $connection_key = join(',', map { $_ . "=" . $args{$_} } qw(hostname username port));
         my $con = $self->{ssh_connections}->{$connection_key};
         if (defined($con)) {
             # Check if we still can create channels on that connection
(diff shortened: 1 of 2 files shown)


backend/baseclass.pm
```diff
@@ -1270,8 +1271,10 @@ sub new_ssh_connection {
 
     # Retry multiple times, in case of the guest is not running yet
     my $counter = $bmwqemu::vars{SSH_CONNECT_RETRY} // 5;
+    my $con_pretty = "$args{username}\@$args{hostname}";
+    $con_pretty = $con_pretty . ":$args{port}" unless $args{port} == 22;
```
Martchus, line 1275:
```suggestion
    $con_pretty = "$con_pretty:$args{port}" unless $args{port} == 22;
```

Martchus (commented):
Looks like tidy doesn't like it.

Martchus (approved):

dcermak:
> Looks like tidy doesn't like it.

I'll try to fix that and will also add some tests.

dcermak:
Turns out I actually had a bug in there with `keep_alive => 1`: you wouldn't get a new connection when connecting to the same host at different ports…

backend/baseclass.pm
```diff
@@ -1269,9 +1270,11 @@ sub new_ssh_connection {
     my $ssh = Net::SSH2->new;
 
     # Retry multiple times, in case of the guest is not running yet
-    my $counter = $bmwqemu::vars{SSH_CONNECT_RETRY} // 5;
+    my $counter    = $bmwqemu::vars{SSH_CONNECT_RETRY} // 5;
+    my $con_pretty = "$args{username}\@$args{hostname}";
+    $con_pretty = "$con_pretty:$args{port}" unless $args{port} == 22;
```
okurz, line 1275:
```suggestion
    $con_pretty .= ":$args{port}" unless $args{port} == 22;
```

okurz (changes requested):

PR #1690, opened 2021-06-15
Add vagrant backend

dcermak:
This is a draft/WIP for a vagrant backend for os-autoinst.

So far it is not documented at all and only some rudimentary manual verification has been performed. I have been able send simple commands to the SUT, but needle matching is not working atm, albeit I am not sure if it even makes sense for vagrant (at least initially).

This currently also includes the changes from #1689, as these are needed for the virtualbox vagrant provider.


backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+    my $vars  = \%bmwqemu::vars;
+    $self->{vagrant_cwd}  = File::Temp->newdir();
+    $self->{box_name}     = get_required_var('VAGRANT_BOX');
+    $self->{boot_timeout} = get_var('VAGRANT_BOOT_TIMEOUT') // 60;
+    $self->{provider}     = get_required_var('VAGRANT_PROVIDER');
+
+    $self->{vagrantfile} = "$self->{vagrant_cwd}/Vagrantfile";
+    bmwqemu::diag("Writing Vagrantfile to $self->{vagrantfile}");
+
+    open(my $fh, '>', "$self->{vagrant_cwd}/Vagrantfile") or die $!;
```
Martchus, line 42: It would be better to use `Mojo::File::spurt` here (or actually later when printing) because you want to throw an exception in error cases anyways.

backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    print($fh $vagrant_file_header . $vagrant_file_provider . $vagrant_file_tail) or die $!;
+    close($fh)                                                                    or die $!;
+
+    return $self;
+}
+
+sub run_vagrant_command {
+    my ($self, $args) = @_;
```
Martchus, line 78:
```suggestion
sub run_vagrant_command ($self, $args) {
```

Martchus (commented):
When adding new code you could actually make use of signatures for the argument handling. (Add `use Mojo::Base -signatures;` at the top and change it like the suggestion in the inline comment.)

A cannot say much about the actual behavior of the code because I don't know much about vagrant.

backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package backend::vagrant;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+use bmwqemu;
+use testapi qw(get_required_var get_var);
```
okurz, line 23: I suggest to avoid use of the external facing API within backends. I tried to remove all existing uses already in other backends but have not achieved that yet. Please instead of something like `get_var` use `$bmwqemu::$vars{…}`

backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+use File::Temp;
+use File::chdir;
+use IPC::Run;
+
+sub new {
+    my $class = shift;
+    my $self  = $class->SUPER::new;
+    my $vars  = \%bmwqemu::vars;
+    $self->{vagrant_cwd}  = File::Temp->newdir();
+    $self->{box_name}     = get_required_var('VAGRANT_BOX');
```
okurz, line 35: as above.

```suggestion
    $self->{box_name}     = $bmwqemu::$vars{VAGRANT_BOX} || die 'Need variable \'VAGRANT_BOX\'';
```

backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+END
+    } elsif ($self->{provider} eq 'libvirt') {
+        $vagrant_file_provider = <<END;
+    config.vm.provider :libvirt do |libvirt|
+    libvirt.cpus = $vars->{QEMUCPUS}
+    libvirt.memory = $vars->{QEMURAM}
+  end
+END
+    } else {
+        # FIXME: log an error or die?
```
okurz, line 68: yes, just call `die` :)

backend/vagrant.pm
```diff
@@ -0,0 +1,190 @@
...
+
+    chomp $res->{stdout};
+    return $res->{stdout};
+}
+
+sub can_handle {
+    # FIXME: what should this function actually do?
+    my ($self, $args) = @_;
+    return;
+}
```
okurz, line 173:
```suggestion
sub can_handle {}
```

okurz (changes requested):

backend/vagrant.pm
dcermak: What shall this function actually do? I really have no clue what the backend should indicate that it can handle.

backend/vagrant.pm
```diff
@@ -32,7 +32,7 @@ sub new {
     my $self  = $class->SUPER::new;
     my $vars  = \%bmwqemu::vars;
     $self->{vagrant_cwd}  = File::Temp->newdir();
-    $self->{box_name}     = get_required_var('VAGRANT_BOX');
+    $self->{box_name}     = $bmwqemu::$vars{VAGRANT_BOX} || die 'Need variable \'VAGRANT_BOX\'';
     $self->{boot_timeout} = get_var('VAGRANT_BOOT_TIMEOUT') // 60;
```
okurz, line 36: Sorry if my previous comment was not clear. Because what I meant is to apply the suggestion of avoiding `get_var` and `get_required_var` or just *any* use of testapi.pm in the backends

okurz (changes requested):

backend/vagrant.pm
dcermak: How would I go about replacing `$testapi::distri->add_console` with `bmwqemu`?

okurz: I don't know. Keep it then

backend/vagrant.pm
```diff
@@ -0,0 +1,186 @@
...
+    push(@vagrant_cmd, ("--machine-readable")) unless $args->{not_machine_readable};
+    if (defined $args->{extra_args}) {
+        push(@vagrant_cmd, @{$args->{extra_args}});
+    }
+    my $timeout = $args->{timeout} // $self->{boot_timeout};
+
+    bmwqemu::diag("Invoking vagrant command: @vagrant_cmd");
+    {
+        chdir($self->{vagrant_cwd}) or die $!;
+        my $handle = IPC::Run::start([@vagrant_cmd], \$stdin, \$stdout, \$stderr, IPC::Run::timeout($timeout));
```
Martchus, line 93:
```suggestion
        my $handle = IPC::Run::start(\@vagrant_cmd, \$stdin, \$stdout, \$stderr, IPC::Run::timeout($timeout));
```

dependencies.yaml
```diff
@@ -123,6 +123,7 @@ main_requires:
   perl(ExtUtils::testlib):
   perl(Fcntl):
   perl(File::Basename):
+  perl(File::chdir):
```
Martchus, line 126: The specfile should be updated as well. You can run the `update-deps` target for that.

backend/vagrant.pm
```diff
@@ -0,0 +1,186 @@
...
+package backend::vagrant;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+use bmwqemu;
+use testapi ();
+
+use Mojo::Base -signatures;
```
Martchus, line 25: Looks like you've enabled signatures but don't make use of them. It would be nice to make use of them.

backend/vagrant.pm
```diff
@@ -0,0 +1,186 @@
...
+    my $res  = $self->run_vagrant_command({cmd => "halt"});
+    die "Failed to execute vagrant halt" unless $res->{retval} == 0;
+}
+
+sub run_cmd {
+    my ($self, $cmd) = @_;
+    my @full_cmd = split(/ /, $cmd);
+
+    my @args = ("--");
+    push(@args, @full_cmd);
```
Martchus, line 162:
```suggestion
    my @args = ('--', split(/ /, $cmd));
```

dependencies.yaml
dcermak: Should be fixed now

backend/vagrant.pm
dcermak: Oh, I thought this would do $magic by itself, sorry… I'll try to fix that.

dcermak: Like this?

Martchus: Yes, looks good now :-)

okurz (changes requested):
Especially for new code I would prefer to have 100% statement coverage from the beginning.

dcermak:
Oliver Kurz ***@***.***> writes:

> @okurz requested changes on this pull request.
>
> Especially for new code I would prefer to have 100% statement coverage from the beginning.

Getting there, unfortunately this requires some changes in my mock code
to actually verify `do_start_vm` properly.

dcermak:
I have converted this back to a draft because I still need to figure out how to handle boxes that do not come from vagrant cloud.

okurz:
please make sure to have no unhandled debug output showing up from tests, see https://github.com/os-autoinst/os-autoinst/pull/1690/checks?check_run_id=2885138500#step:3:600

dcermak:
> please make sure to have no unhandled debug output showing up from tests, see https://github.com/os-autoinst/os-autoinst/pull/1690/checks?check_run_id=2885138500#step:3:600

Hm, how should I achieve that? Use `like` for anything that could write to the log or remove this logging?

Martchus:
You could mock the logging away or mock it to use test helpers like `note`.

okurz:
or set a different log level like in https://github.com/os-autoinst/os-autoinst/blob/master/t/10-virtio_terminal.t#L118

For mocking: https://github.com/os-autoinst/os-autoinst/blob/master/t/31-sshSerial.t#L42

dcermak:
I hope that this could work :crossed_fingers:

okurz (approved):

backend/vagrant.pm
```diff
@@ -0,0 +1,244 @@
...
+package backend::vagrant;
+
+use strict;
+use warnings;
+
+use base 'backend::virt';
+use bmwqemu;
+use testapi ();
+
+use Mojo::Base -signatures;
```
okurz, line 25:
```suggestion
use base 'backend::virt';
use bmwqemu;
use testapi ();

use Mojo::Base -strict, -signatures;
```

backend/vagrant.pm
```diff
@@ -0,0 +1,244 @@
...
+
+        opendir(my $dh, $asset_dir) or die "Could not opendir $asset_dir: $!";
+        foreach ("iso", "iso/fixed", "hdd", "hdd/fixed", "other", "other/fixed") {
+            my $subdir = abs_path("$asset_dir/$_");
+            next unless (defined($subdir) and (-d $subdir));
+
+            my $box_path = abs_path("$subdir/$box_name");
+            if (-e $box_path) {
+                $box_abs_path = $box_path;
+            }
```
okurz, line 55:
```suggestion
            my $box_path = abs_path("$subdir/$box_name");
            $box_abs_path = $box_path if -e $box_path;
```

backend/vagrant.pm
```diff
@@ -0,0 +1,244 @@
...
+}
+
+sub run_vagrant_command ($self, $args) {
+    my ($retval, $stdin, $stdout, $stderr);
+
+    my @vagrant_cmd = ("vagrant", $args->{cmd});
+    push(@vagrant_cmd, ("--machine-readable")) unless $args->{not_machine_readable};
+    if (defined $args->{extra_args}) {
+        push(@vagrant_cmd, @{$args->{extra_args}});
+    }
```
okurz, line 120:
```suggestion
    push @vagrant_cmd, @{$args->{extra_args}} if defined $args->{extra_args};
```

backend/vagrant.pm
```diff
@@ -0,0 +1,244 @@
...
+    my $vars = \%bmwqemu::vars;
+
+    $self->{vagrant_cwd} = File::Temp->newdir();
+    $self->{up_timeout}  = $vars->{VAGRANT_UP_TIMEOUT} // 300;
+    $self->{provider}    = $vars->{VAGRANT_PROVIDER}   // die 'Need variable \'VAGRANT_PROVIDER\'';
+    $self->{box_name}    = $vars->{VAGRANT_BOX} || die 'Need variable \'VAGRANT_BOX\'';
+    $self->{box_url}     = $vars->{VAGRANT_BOX_URL};
+
+    if (substr($self->{box_name}, 0, 1) eq '/') {
+        my $asset_dir    = $vars->{ASSETDIR} // '/var/lib/openqa/share/factory/';
```
okurz, line 43: I would prefer to not introduce openQA-isms into backends. Can you make that a required variable with the PREFIX `VAGRANT_`?

okurz (changes requested):

backend/vagrant.pm
dcermak: Do you mean `ASSETDIR`? I don't necessarily disagree with you, but I imagine that that will result in pretty bad UX, because now as a test writer you'll have to specify `ASSETDIR` twice or know the default.

okurz: Why twice?

Martchus: Well, this breaks in the general case, e.g. one might use `OPENQA_BASEDIR` to use some other prefix than `/var/lib`. Besides, in productions workers assets are cached and we don't want to rely on the `/var/lib/openqa/share` mount.

okurz: You could try the same as for HDD and ISO where by default we look in the current directory, I think

dcermak: > Why twice?

Let's assume your worker uses a different `ASSETDIR`, then you'd have to set `ASSETDIR` and `OPENQA_ASSETDIR` to the same value. That's what I meant by twice.


> Well, this breaks in the general case, e.g. one might use `OPENQA_BASEDIR` to use some other prefix than `/var/lib`. Besides, in productions workers assets are cached and we don't want to rely on the `/var/lib/openqa/share` mount.

Oh bummer. How can I get the cached assets then? Look into cwd as @okurz suggested?

okurz: well, which assets are we actually talking about? And how do they end up in any folder? who or which component would copy them there at which time?

dcermak: I'm talking about vagrant boxes that are pushed to e.g. o3 from OBS staging project. At the moment they are available as `ASSET_LIBVIRT` and `ASSET_VIRTUALBOX` (see e.g. https://openqa.opensuse.org/tests/1803591#settings)

dcermak: @okurz I've tried to implement something like you suggested during today's QE tools meeting. Did you have something like that in mind?

okurz (changes requested):
One trivial grammar typo error. Other than this I would really appreciate if you can bring the code coverage for the newly introduced file backend/vagrant.pm up to 100%. As you are nearly there and already learned how to do the mocking that should be really easy for you to achieve :)

doc/backends.md
```diff
@@ -47,3 +47,48 @@ Otherwise there is no configuration required. The test will configure libvirt on
...
+Alternatively to downloading boxes directly from the internet, the vagrant
+backend can use vagrant boxes that are available in the directories defined by
+the variable `VAGRANT_ASSETDIR`. The backend will look for a file with the name
+provided in `VAGRANT_BOX` if it starts with a `/`. If the box is not found, then
+the backend will report an error.
+
+
+### Worker configuration
+
+The worker must have `vagrant` installed and it needs the correct permissions to
```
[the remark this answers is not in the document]
dcermak: I've changed the formulation a bit.

t/33-vagrant.t
```diff
@@ -0,0 +1,511 @@
...
+        $ret->set_true('backend');
+        return $ret;
+});
+$testapi::distri = distribution->new;
+
+my $check_socket_ret            = undef;
+my $check_socket_fh_expected    = undef;
+my $check_socket_write_expected = undef;
+my $mock_virt                   = Test::MockModule->new("backend::virt", no_auto => 1);
+$mock_virt->redefine('check_socket', sub {
```
okurz, line 54: backend::virt::check_socket does not exist. Did you forget to commit something?

t/33-vagrant.t
```diff
@@ -0,0 +1,511 @@
+#!/usr/bin/perl
+
+use Cwd qw(abs_path);
+use Test::Most;
+use File::chdir;
+use FindBin '$Bin';
+use lib "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '20';
```
okurz, line 8: for me gladly the test module only takes a second to execute so I suggest to lower this. We use 5 as the lowest value in most test modules.

dcermak:
Uh, the CI now fails because of a wrong perlcritic version. What's going on there?

t/33-vagrant.t
dcermak: Ups, no, I just did not check where `check_socket` is defined.

doc/backends.md
```diff
@@ -47,3 +47,49 @@ Otherwise there is no configuration required. The test will configure libvirt on
...
 of the running virtual machine.
+
+
+## vagrant
+
+The vagrant backend leverages [vagrant](https://www.vagrantup.com/) to launch
+VMs and use them as the SUT, providing a ssh connection to the SUT via `vagrant
+ssh`.
+
+The selects the box from the variable `VAGRANT_BOX` and the provider from
```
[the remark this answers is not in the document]
dcermak: Indeed, thanks for spotting that!

t/33-vagrant.t
```diff
@@ -0,0 +1,511 @@
...
+
+my $expected_console_name        = undef;
+my $expected_console_type        = undef;
+my %expected_console_credentials = ();
+my $distri                       = Test::MockModule->new('distribution');
+$distri->redefine('add_console', sub {
+        my ($self, $name, $type, $creds) = @_;
+
+        if (defined($expected_console_name)) {
+            is($name, $expected_console_name, 'Check the name of the console');
```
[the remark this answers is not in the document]
dcermak: Could you point me to an existing example what you envision such a function to look like?

okurz:
We need https://github.com/os-autoinst/os-autoinst/pull/1703 for the tidy-update first

t/33-vagrant.t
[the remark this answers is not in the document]
okurz: Well, as I said, how about a `die`-call? So

```suggestion
            die "Invalid name of console '$name'. Expected '$expected_console_name'" unless $name eq $expected_console_name;
```

t/33-vagrant.t
```diff
@@ -0,0 +1,522 @@
...
+        my ($self, $name, $type, $creds) = @_;
+
+        if (defined($expected_console_name)) {
+            die "Invalid console name '$name'. Expected '$expected_console_name'" unless $name eq $expected_console_name;
+        }
+        if (defined($expected_console_type)) {
+            die "Invalid console type '$type'. Expected '$expected_console_type'" unless $type eq $expected_console_type;
+        }
+        if (%expected_console_credentials) {
+            is_deeply($creds, \%expected_console_credentials, 'Check the name of the console');
```
okurz, line 42: here also I would have avoided the use of a Test::More function but nevermind

okurz (approved):

t/33-vagrant.t
dcermak: Unfortunately I don't really know how to perform a deep comparison of hashes without relying on `is_deeply`.

t/33-vagrant.t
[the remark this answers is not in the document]
perlpunk: The test framework allows to set the level, so that the correct line will be reported:
https://metacpan.org/pod/Test::Builder#Test-style
```perl
sub helper {
    local $Test::Builder::Level = $Test::Builder::Level + 1;
    ok($x);
}
```

t/33-vagrant.t
perlpunk: See my comment above for setting the test level

dcermak: Thanks! I've applied this to the current test code.

doc/backends.md
```diff
@@ -47,3 +47,49 @@ Otherwise there is no configuration required. The test will configure libvirt on
 own. By default, the openSUSE test distribution uses the domain name `openQA-SUT-1`.
 So you can for instance use `virsh dumpxml openQA-SUT-1` to investigate the configuration
 of the running virtual machine.
+
+
+## vagrant
```
Martchus, line 52:
```suggestion
## Vagrant
```
Same counts for the other occurrences; as it is a name it should start with a capital letter. (Unless it is the literal command of course.)

t/33-vagrant.t
```diff
@@ -0,0 +1,524 @@
...
+
+        local $Test::Builder::Level = $Test::Builder::Level + 1;
+
+        if (defined($expected_console_name)) {
+            die "Invalid console name '$name'. Expected '$expected_console_name'" unless $name eq $expected_console_name;
+        }
+        if (defined($expected_console_type)) {
+            die "Invalid console type '$type'. Expected '$expected_console_type'" unless $type eq $expected_console_type;
+        }
+        if (%expected_console_credentials) {
```
Martchus, line 43:
```suggestion
        if (keys %expected_console_credentials) {
```

t/33-vagrant.t
```diff
@@ -0,0 +1,524 @@
...
+my $mock_run  = Test::MockModule->new("IPC::Run");
+my $mock_file = Test::MockModule->new("File::chdir");
+
+my $expected_spurt_str = undef;
+my $mock_path          = Test::MockObject->new();
+$mock_path->mock('make_path', sub {
+        my ($self, %opts) = @_;
+        return $self;
+});
+$mock_path->mock('child', sub {
```
Martchus, line 77: With signatures code like this would possibly fit into one line.

Martchus (changes requested):

t/33-vagrant.t
dcermak: I've replaced all the old style `sub`s with signatures also in `33-vagrant.t`.

t/33-vagrant.t
```diff
@@ -0,0 +1,512 @@
...
+    close(FH);
+
+    $backend_vars->{VAGRANT_ASSETDIR} = $tmpdir;
+    $backend_vars->{VAGRANT_PROVIDER} = 'libvirt';
+    $backend_vars->{VAGRANT_BOX}      = $box_name;
+    $backend_vars->{QEMUCPUS}         = 1;
+    $backend_vars->{QEMURAM}          = 1024;
+
+    # my $abs_path_to_box = abs_path($box_file);
+
```
Martchus, line 228: These lines should be removed.

t/33-vagrant.t
```diff
@@ -0,0 +1,512 @@
...
+    $backend_vars->{VAGRANT_BOX} = "foobar";
+    $backend_vars->{ASSETDIR}    = undef;
+};
+
+subtest 'Vagrantfile with a box file existing in VAGRANT_ASSETDIR' => sub {
+    my $box_name = "/Tumbleweed.box";
+    my $box_file = "$tmpdir$box_name";
+    open(FH, ">$box_file");
+    print FH '';
+    close(FH);
```
Martchus, line 219: Looks like you're creating an empty file here. Using `Mojo::File::touch` would be nicer.

Martchus (changes requested):

t/33-vagrant.t
dcermak: They are gone.

t/33-vagrant.t
dcermak: It doesn't look as pretty, since Mojo::file is being mocked, but it's now just one line.

Martchus (approved):

t/33-vagrant.t
Martchus: Thanks, it is still better than the previous code.

PR #1692, opened 2021-06-16
ci: Workaround ci_extended container setup gid issue

asdil12:
https://progress.opensuse.org/issues/93880

This is only for testing and is to be removed before merge:

```diff
-on: push
+on: [push, pull_request]
```


.github/workflows/ci_extended.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: ci extended
 # yamllint disable-line rule:truthy
-on: push
+on: [push, pull_request]
```
okurz, line 4: I assume you do that just for testing. Please revert before merging

okurz (changes requested):

.github/workflows/ci_extended.yml
asdil12: yes - as mentioned in the first comment.

container/os-autoinst_dev/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM opensuse/tumbleweed
 ENV NAME openQA test environment
 ENV LANG en_US.UTF-8
 
-RUN zypper ar -f -G 'http://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed' devel_openqa
+RUN zypper ar -f -G 'http://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed' devel_openqa && zypper -n in system-user-qemu && zypper clean
```
okurz, line 8: Please add a comment here to explain the workaround

asdil12: done

Martchus (approved):

okurz (approved):

PR #1695, opened 2021-06-19
Fix QEMU_COMPRESS_QCOW2 missing implementation introduced with c670720d

okurz:
The commit c670720d kept the initial reading of the parameter
QEMU_COMPRESS_QCOW2 and documentation in place but the actual
implementation was removed completely so the parameter had no effect
when specified.

### DIFF
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -139,8 +139,14 @@ sub gen_qemu_img_cmdlines {
 
 sub gen_qemu_img_convert {
     my ($self, $img_dir, $name) = @_;
-
-    return ['convert', '-c', '-O', QEMU_IMAGE_FORMAT, $self->drive->file, "$img_dir/$name"];
+    my @cmd = qw(convert);
+    # By compressing we are making the images self contained, i.e. they are
+    # portable by not requiring backing files referencing the openQA instance.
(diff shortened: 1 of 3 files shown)


Martchus (commented):
The commit message could state what the regression actually is.

marmarek:
Thanks, I'll give it a spin and report back, likely tomorrow or on Monday.

okurz:
@Martchus 

> The commit message could state what the regression actually is.

done. First I thought it's obvious but I guess I can make that more clear. Updated the subject line as well as details.

Martchus (approved):

foursixnine:
Just to get a bit more familiar with the qemu backend's state after the rewrite, I did the same but took a different approach, precisely avoiding global variables from bmwqemu:

https://github.com/foursixnine/os-autoinst/commit/eb201b1db212bafebd0e670101b38a7620eb3a87

marmarek:
It works :)

okurz:
@marmarek good that it helps you 

@foursixnine I like your approach. Would be cleaner in the long term to use function parameters like you did and only read test variables on the highest level

okurz:
@foursixnine maybe you want to try to rebase your approach and we can take it

foursixnine:
Let me open the PR at last and I'll rebase in the course of the day

PR #1696, opened 2021-06-19
Update all files to use perl "signatures"

okurz:
Using

```
sed -i "s/use warnings;/&\nuse feature 'signatures';\nno warnings 'experimental::signature';/" $(git ls-files)
sed -i '/^sub / {N;s/ {\n.*my [(]\?\([^)=]*\).*$/(\1) {/}' $(git ls-files)
tools/tidy --only-changed
```

and some manual fixes.

https://progress.opensuse.org/issues/99663

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -1060,3 +1060,7 @@ subtest 'upload_asset' => sub {
 };
 
 done_testing;
+
+END {
+    unlink 'vars.json';
+}
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -92,6 +92,9 @@ sub fake_read_json ($fd) {
     elsif ($cmd eq 'backend_wait_screen_change' || $cmd eq 'backend_wait_still_screen') {
         return {ret => {sim => $fake_similarity, elapsed => 5, timed_out => $fake_timeout}};
     }
+    elsif ($cmd eq 'backend_start_audiocapture') {
+        return {ret => 0};
+    }
     else {
         note "mock method not implemented \$cmd: $cmd\n";
     }
@@ -1046,16 +1049,30 @@ like(exception { ensure_installed }, qr/implement.*for your distri/, 'ensure_ins
 lives_ok { hold_key('ret') } 'hold_key can be called';
 lives_ok { release_key('ret') } 'release_key can be called';
 lives_ok { reset_consoles } 'reset_consoles can be called';
+
+subtest 'assert/check recorded sound' => sub {
+    $cmds = [];
+    lives_ok { start_audiocapture } 'start_audiocapture can be called';
+    like $cmds->[0]->{filename}, qr/captured\.wav/, 'audiocapture started with expected args' or diag explain $cmds;
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->noop('_snd2png');
+    $mock_basetest->noop('verify_sound_image');
+    ok assert_recorded_sound('foo'), 'assert_recorded_sound can be called';
+    ok check_recorded_sound('foo'), 'check_recorded_sound can be called';
+};
+
 lives_ok { power('on') } 'power can be called';
 lives_ok { save_memory_dump } 'save_memory_dump can be called';
 like(exception { save_storage_drives }, qr/should be called.*post_fail_hook/, 'save_storage_drives should be called special');
 lives_ok { freeze_vm } 'freeze_vm can be called';
 lives_ok { resume_vm } 'resume_vm can be called';
 
-subtest 'upload_asset' => sub {
+subtest 'upload_asset and parse_junit_log' => sub {
     my $mock_testapi = Test::MockModule->new('testapi');
     $mock_testapi->noop('assert_script_run');
     ok upload_asset('foo'), 'upload_asset can be called';
+    $mock_testapi->redefine(upload_logs => sub { die 'foo' });
+    like(exception { parse_junit_log('foo') }, qr/foo/, 'parse_junit_log calls upload_logs');
 };
 
 done_testing;
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -347,7 +347,6 @@ subtest 'script_run' => sub {
 
 sub assert_script_sudo_test ($waittime, $is_serial_terminal) {
     my $mock_testapi = Test::MockModule->new('testapi');
-    $mock_testapi->redefine(_handle_found_needle => sub { return $_[0] });
     $mock_testapi->noop(qw(send_key enter_cmd));
     my $script_sudo = '';
     $mock_testapi->redefine(hashed_string => 'XXX');
--- a/testapi.pm
+++ b/testapi.pm
@@ -245,8 +245,7 @@ sub force_soft_failure ($reason) {
     $autotest::current_test->record_soft_failure_result($reason, force_status => 1);
 }
 
-sub _handle_found_needle {
-    my ($foundneedle, $rsp, $tags) = @_;
+sub _handle_found_needle ($foundneedle, $rsp, $tags) {
     # convert the needle back to an object
     $foundneedle->{needle} = needle->new($foundneedle->{needle});
     my $img = tinycv::from_ppm(decode_base64($rsp->{image}));
@@ -1339,9 +1338,7 @@ Throws C<FailedNeedle> exception if needle is not matched until C<$counter> is 0
 
 =cut
 
-sub send_key_until_needlematch {
-    my ($tag, $key, $counter, $timeout) = @_;
-
+sub send_key_until_needlematch ($tag, $key, $counter = undef, $timeout = undef) {
     $counter //= 20;
     $timeout //= 1;
 
@@ -1768,22 +1765,20 @@ I<Only supported by qemu backend.>
 
 =cut
 
-sub start_audiocapture {
+sub start_audiocapture () {
     my $fn = $autotest::current_test->capture_filename;
     my $filename = join('/', bmwqemu::result_dir(), $fn);
     bmwqemu::log_call(filename => $filename);
     return query_isotovideo('backend_start_audiocapture', {filename => $filename});
 }
 
-sub _check_or_assert_sound {
-    my ($mustmatch, $check) = @_;
+sub _snd2png ($wavfile, $imgpath) { system("snd2png $wavfile $imgpath") }    # uncoverable statement
 
+sub _check_or_assert_sound ($mustmatch, $check = undef) {
     my $result = $autotest::current_test->stop_audiocapture();
     my $wavfile = join('/', bmwqemu::result_dir(), $result->{audio});
-    system("snd2png $wavfile $result->{audio}.png");
-
     my $imgpath = "$result->{audio}.png";
-
+    _snd2png($wavfile, $result);
     return $autotest::current_test->verify_sound_image($imgpath, $mustmatch, $check);
 }
 
@@ -1798,8 +1793,7 @@ I<Only supported by QEMU backend.>
 
 =cut
 
-sub assert_recorded_sound {
-    my ($mustmatch) = @_;
+sub assert_recorded_sound ($mustmatch) {
     return _check_or_assert_sound $mustmatch;
 }
 
@@ -1814,8 +1808,7 @@ I<Only supported by QEMU backend.>
 
 =cut
 
-sub check_recorded_sound {
-    my ($mustmatch) = @_;
+sub check_recorded_sound ($mustmatch) {
     return _check_or_assert_sound $mustmatch, 1;
 }
 
@@ -2017,7 +2010,7 @@ file is then parsed as jUnit format and extra test results are created from it.
 =cut
 
 # XXX: To keep until tests are adapted
-sub parse_junit_log { return parse_extra_log('JUnit', shift) }
+sub parse_junit_log ($path) { return parse_extra_log('JUnit', $path) }
 
 =head2 parse_extra_log
 
@@ -2033,9 +2026,7 @@ file is then parsed as the format supplied, that can be understood by OpenQA::Pa
 
 =cut
 
-sub parse_extra_log {
-    my ($format, $file) = @_;
-
+sub parse_extra_log ($format, $file) {
     $file = upload_logs($file);
     my @tests;
 
--- a/testapi.pm
+++ b/testapi.pm
@@ -1142,8 +1142,7 @@ This will return content of the file located in data/autoyast/autoinst.xml
 
 =cut
 
-sub get_test_data {
-    my ($path) = @_;
+sub get_test_data ($path) {
     defined $bmwqemu::vars{CASEDIR} or die 'Need variable CASEDIR';
     $path = "$bmwqemu::vars{CASEDIR}/data/$path";
     bmwqemu::log_call(path => $path);
@@ -1239,9 +1238,7 @@ I<The implementation is distribution specific and not always available.>
 
 =cut
 
-sub become_root {
-    return $distri->become_root;
-}
+sub become_root () { $distri->become_root }
 
 =head2 ensure_installed
 
@@ -1253,9 +1250,7 @@ I<The implementation is distribution specific and not always available.>
 
 =cut
 
-sub ensure_installed {
-    return $distri->ensure_installed(@_);
-}
+sub ensure_installed (@args) { $distri->ensure_installed(@args) }
 
 =head2 hashed_string
 
@@ -1314,8 +1309,7 @@ Hold one C<$key> until release it
 
 =cut
 
-sub hold_key {
-    my ($key) = @_;
+sub hold_key ($key) {
     bmwqemu::log_call(key => $key);
     query_isotovideo('backend_hold_key', {key => $key});
 }
@@ -1328,8 +1322,7 @@ Release one C<$key> which is kept holding
 
 =cut
 
-sub release_key {
-    my $key = shift;
+sub release_key ($key) {
     bmwqemu::log_call(key => $key);
     query_isotovideo('backend_release_key', {key => $key});
 }
@@ -1746,7 +1739,7 @@ if you did something to the system that affects the console (e.g. trigger reboot
 
 =cut
 
-sub reset_consoles {
+sub reset_consoles () {
     query_isotovideo('backend_reset_consoles');
     return;
 }
@@ -1836,10 +1829,7 @@ Trigger backend specific power action, can be C<'on'>, C<'off'>, C<'acpi'> or C<
 
 =cut
 
-sub power {
-
-    # params: (on), off, acpi, reset
-    my ($action) = @_;
+sub power ($action) {
     bmwqemu::log_call(action => $action);
     query_isotovideo('backend_power', {action => $action});
 }
@@ -1936,8 +1926,7 @@ I<Currently only qemu backend is supported.>
 
 =cut
 
-sub save_memory_dump {
-    my %nargs = @_;
+sub save_memory_dump (%nargs) {
     $nargs{filename} ||= $autotest::current_test->{name};
 
     bmwqemu::log_call(%nargs);
@@ -1959,8 +1948,8 @@ I<Currently only qemu backend is supported.>
 
 =cut
 
-sub save_storage_drives {
-    my $filename ||= $autotest::current_test->{name};
+sub save_storage_drives ($filename = undef) {
+    $filename //= $autotest::current_test->{name};
     die "save_storage_drives should be called within a post_fail_hook" unless ((caller(1))[3]) =~ /post_fail_hook/;
 
     bmwqemu::log_call();
@@ -1990,7 +1979,7 @@ I<Currently only qemu backend is supported.>
 
 =cut
 
-sub freeze_vm {
+sub freeze_vm () {
     # While it might be a good idea to allow the user to stop the vm within a test
     # we're not encouraging them to do that outside a post_fail_hook or at any point
     # in the test code.
@@ -2011,7 +2000,7 @@ I<Currently only qemu backend is supported.>
 
 =cut
 
-sub resume_vm {
+sub resume_vm () {
     bmwqemu::log_call();
     query_isotovideo('backend_cont_vm');
 }
@@ -2220,9 +2209,7 @@ C<$nocheck> parameter:
 
 =cut
 
-sub upload_asset {
-    my ($file, $public, $nocheck) = @_;
-
+sub upload_asset ($file, $public = undef, $nocheck = undef) {
     if (get_var('OFFLINE_SUT')) {
         record_info('upload skipped', "Skipped uploading asset '$file' as we are offline");
         return;
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 36;
+our $version = 37;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 35;
+our $version = 36;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/xt/01-style.t
+++ b/xt/01-style.t
@@ -13,7 +13,7 @@ ok system(qq{git grep -I -l '[#/ ]*SPDX-License-Identifier ' ':!COPYING' ':!exte
 is qx{git grep -I -L '^use Test::Most' t/**.t}, '', 'All tests use Test::Most';
 is qx{git grep -I -L '^use Test::Warnings' t/**.t}, '', 'All tests use Test::Warnings';
 is qx{git grep -I -l '^use testapi' backend/ consoles/}, '', 'No backend or console files use external facing testapi';
-is qx{git grep -l -e '^sub \\S\\+ [^(]\\+' --and --not -e 'sub [(\{]' --and --not -e 'sub \\S\\+(' --and --not -e 'sub \\S\\+;' --and --not -e '# no:style:signatures' ':!testapi.pm' ':!external/'}, '', 'All files use sub signatures everywhere (nameless and in-place definitions still allowed)';
+is qx{git grep -l -e '^sub \\S\\+ [^(]\\+' --and --not -e 'sub [(\{]' --and --not -e 'sub \\S\\+(' --and --not -e 'sub \\S\\+;' --and --not -e '# no:style:signatures' ':!external/'}, '', 'All files use sub signatures everywhere (nameless and in-place definitions still allowed)';
 is qx{git grep -L '^#!.*perl' t/**.t}, '', 'All test files have shebang';
 is qx{git ls-files -s t/**.t | grep -v ^1007}, '', 'All test modules are executable';
 is qx{git grep -l '^use POSIX;'}, '', 'Use of bare POSIX import is discouraged, see https://perldoc.perl.org/POSIX';


okurz:
@os-autoinst/tools-team what are your preferences

1.
```
```
2.
```
use feature qw(signatures);
no warnings qw(experimental::signatures);
```
3. something different

Martchus:
Not sure. If we don't already use `Mojo::Base` in a module it would make sense to use 2. instead. However, the module is likely loaded at some point anyways (because some module actually uses it) so it is likely not making a real difference.

perlpunk:
For completenes::
```
use experimental qw(signatures);
```

kraih:
[This](https://github.com/mojolicious/mojo/blob/6deb28ad70a5acd865acbf9bbca61532a1ee39ba/lib/Mojo/Base.pm#L124) is the code `Mojo::Base` uses currently. This is an intended use for `Mojo::Base` so we could just keep using that and avoid the extra boilerplate. If we do decide to use our own version, then perhaps we should just make it reusable across all our modules with an `OpenQA::Base` module?

kraih:
Maybe worth mentioning that `use Mojo::Base -strict, -signatures;` is equivalent with this currently:
```
use strict; 
use warnings; 
use utf8; 
use feature ':5.16'; 
use mro;
use experimental qw(signatures postderef);
```

kraih:
Regarding the `sed` command, you will need to check for uses of prototypes. Since those are present in some openQA modules, and will break when signatures are activated. And we cannot switch to the `:prototype` attribute, since that doesn't work the same in all versions of Perl. So the best course of action would probably to just remove uses of prototypes if possible, and if not move the prototype code into dedicated modules where signatures are not activated at all.

Martchus:
I know but hopefully the compile check shows a problem when prototypes were used. At least in the cases I've encountered it ended up being a compile error. Note that it is possible to use *only* the `:prototype` attribute. (Otherwise this change wasn't a good idea after all: https://github.com/os-autoinst/openQA/commit/01d8c8795ae1984718f9a95d8d824447a173fe4c#diff-150605f30628d39e6c9a7316cd78f4c876fe3be95dee857b9b395100b5d5ea8cR636)

kraih:
> Note that it is possible to use _only_ the `:prototype` attribute. (Otherwise this change wasn't a good idea after all: [os-autoinst/openQA@01d8c87#diff-150605f30628d39e6c9a7316cd78f4c876fe3be95dee857b9b395100b5d5ea8cR636](https://github.com/os-autoinst/openQA/commit/01d8c8795ae1984718f9a95d8d824447a173fe4c#diff-150605f30628d39e6c9a7316cd78f4c876fe3be95dee857b9b395100b5d5ea8cR636))

You're right. Only using `:prototype` with `@_` should be fine.

foursixnine:
> So I understand there is no strong preference for one or the other way? I prefer `use Mojo::Base -strict, -signatures;` because we can then remove some more use-statements like for "warnings" and "strict" and as long as the project relies on mojo anyway I don't think we have a problem using that.

I'd go for `use Mojo::Base -strict, -signatures;` too...  it's less things to change in the future on one end, and we get nice shiny 'say $thing', which is two characters less typing consuming than print for debugging '(no \n needed!)';

OpenQA/Benchmark/Stopwatch.pm
```diff
@@ -2,6 +2,8 @@ package OpenQA::Benchmark::Stopwatch;
 
 use strict;
 use warnings;
+use feature 'signatures';
+no warnings 'experimental::signatures';
```
foursixnine, line 6: You forgot this one? :)

okurz: no, my sed expressions actually do not use the Mojo way so the PR draft is a mix of both to show the alternatives to reviewers. I will go with Mojo::Base now and cover all cases

OpenQA/Benchmark/Stopwatch.pm
```diff
@@ -236,9 +236,7 @@ sub as_data ($self) {
     return \%data;
 }
 
-sub time {
-    &{$_[0]{_time}};
-}
+sub time ($self) { $self->{_time} }
```
perlpunk, line 239: This needs to be called as a subroutine, I suggest
```suggestion
sub time ($self) { $self->{_time}->() }
```
alternatively
```suggestion
sub time ($self) { &{$self->{_time}} }
```

testapi.pm
```diff
@@ -430,8 +416,7 @@ is hit. Default timeout is 30s.
 
 =cut
 
-sub assert_screen {
-    my ($mustmatch) = shift;
+sub assert_screen ($mustmatch, @) {
     my $timeout;
     $timeout = shift if (@_ % 2);
```
perlpunk, line 421: `@_` and `shift` is still used

testapi.pm
```diff
@@ -458,8 +443,7 @@ Returns matched needle or C<undef> if timeout is hit. Default timeout is 0s.
 
 =cut
 
-sub check_screen {
-    my ($mustmatch) = shift;
+sub check_screen ($mustmatch, @) {
     my $timeout;
     $timeout = shift if (@_ % 2);
```
perlpunk, line 448: same here

testapi.pm
```diff
@@ -681,8 +660,7 @@ Default timeout is 30s, default stilltime is 7s.
 
 =cut
 
-sub wait_still_screen {
-    my $stilltime = looks_like_number($_[0]) ? shift : 7;
+sub wait_still_screen ($stilltime, @) {
     my $timeout = (@_ % 2) ? shift : $bmwqemu::default_timeout;
```
perlpunk, line 664: `@_` and `shift`

testapi.pm
```diff
@@ -974,8 +943,7 @@ should work on *nix operating systems with a configured serial device.>
 
 =cut
 
-sub assert_script_run {
-    my $cmd = shift;
+sub assert_script_run ($cmd, @) {
```
perlpunk, line 946: `@_` is used below

testapi.pm
```diff
@@ -1022,8 +990,7 @@ device C<$serialdev>.
 
 =cut
 
-sub script_run {
-    my $cmd = shift;
+sub script_run ($cmd, @) {
```
perlpunk, line 993: `@_` is used below

testapi.pm
```diff
@@ -1119,10 +1084,7 @@ I<The implementation is distribution specific and not always available.>
 
 =cut
 
-sub script_sudo {
-    my $name = shift;
-    my $wait = shift // 2;
-
+sub script_sudo ($name, $wait = 2) {
```
perlpunk, line 1087:
```suggestion
sub script_sudo ($name, $wait = 2) {
  $wait //= 2;
```

testapi.pm
```diff
@@ -1153,8 +1115,7 @@ and can be tweaked by setting the C<$wait> positional parameter.
 
 =cut
 
-sub script_output {
-    my $script = shift;
+sub script_output ($script, @) {
```
perlpunk, line 1118: `@_` is used below

testapi.pm
```diff
@@ -1325,10 +1276,7 @@ MD5 algorithm and taking the first characters.
 
 =cut
 
-sub hashed_string {
-    my ($string, $count) = @_;
-    $count //= 5;
-
+sub hashed_string ($string, $count = 5) {
```
perlpunk, line 1279:
```suggestion
sub hashed_string ($string, $count = undef) {
    $count //= 5;
```

testapi.pm
```diff
@@ -1353,14 +1301,13 @@ Special characters naming:
 
 =cut
 
-sub send_key {
-    my $key = shift;
+sub send_key ($key, @) {
     my %args = (@_ == 1) ? (do_wait => +shift()) : @_;
```
perlpunk, line 1305: `@_` and `shift`

testapi.pm
```diff
@@ -1407,11 +1352,7 @@ Throws C<FailedNeedle> exception if needle is not matched until C<$counter> is 0
 
 =cut
 
-sub send_key_until_needlematch {
-    my ($tag, $key, $counter, $timeout) = @_;
-
-    $counter //= 20;
-    $timeout //= 1;
+sub send_key_until_needlematch ($tag, $key, $counter = 20, $timeout = 1) {
```
perlpunk, line 1355:
```suggestion
sub send_key_until_needlematch ($tag, $key, $counter = undef, $timeout = undef) {
    $counter //= 20;
$timeout //= 1;
```

testapi.pm
```diff
@@ -1451,16 +1392,9 @@ enter a command line.
...
     # special argument handling for backward compat
-    my $string = shift;
-    my %args;
-    if (@_ == 1) {    # backward compat
-        %args = (max_interval => $_[0]);
-    }
-    else {
-        %args = @_;
-    }
+    my %args = @_ == 1 ? (max_interval => $_[0]) : @_;
```
perlpunk, line 1397: `@_`

testapi.pm
```diff
@@ -1529,8 +1461,8 @@ You can pass the same optional parameters as for C<type_string> function.
 
 =cut
 
-sub enter_cmd {
-    type_string shift, lf => 1, @_;
+sub enter_cmd ($string, @) {
+    type_string $string, lf => 1, @_;
```
perlpunk, line 1465: `@_`

testapi.pm
```diff
@@ -1559,9 +1489,7 @@ Default hold time is 0.15s
 
 =cut
 
-sub mouse_click {
-    my $button = shift || 'left';
-    my $time = shift || 0.15;
+sub mouse_click ($button, $time = 0.15) {
```
perlpunk, line 1492:
```suggestion
sub mouse_click ($button = undef, $time = undef) {
    $button ||= 'left';
    $time ||= 0.15;
```

testapi.pm
```diff
@@ -1576,9 +1504,7 @@ Same as mouse_click only for double click.
 
 =cut
 
-sub mouse_dclick(;$$) {
-    my $button = shift || 'left';
-    my $time = shift || 0.10;
+sub mouse_dclick ($button, $time = 0.10) {
```
perlpunk, line 1507:
```suggestion
sub mouse_dclick ($button = undef, $time = undef) {
    $button ||= 'left';
    $time ||= 0.10;
```

testapi.pm
```diff
@@ -1597,9 +1523,7 @@ Same as mouse_click only for triple click.
 
 =cut
 
-sub mouse_tclick(;$$) {
-    my $button = shift || 'left';
-    my $time = shift || 0.10;
+sub mouse_tclick ($button, $time = 0.10) {
```
perlpunk, line 1526:
```suggestion
sub mouse_tclick ($button = undef, $time = undef) {
    $button ||= 'left';
    $time ||= 0.10;
```

testapi.pm
```diff
@@ -1622,8 +1546,7 @@ Hide mouse cursor by moving it out of screen area.
 
 =cut
 
-sub mouse_hide(;$) {
-    my $border_offset = shift || 0;
+sub mouse_hide ($border_offset = 0) {
```
perlpunk, line 1549:
```suggestion
sub mouse_hide ($border_offset = undef) {
    $border_offset ||= 0;
```

testapi.pm
```diff
@@ -1843,16 +1761,13 @@ I<Only supported by qemu backend.>
 
 =cut
 
-sub start_audiocapture {
-    my $fn = $autotest::current_test->capture_filename;
+sub start_audiocapture ($fn) {
```
perlpunk, line 1764: I don't understand that change. It seems wrong.

testapi.pm
```diff
@@ -1922,9 +1832,7 @@ Returns true on success and false if C<$timeout> timeout is hit. Default timeout
 
 =cut
 
-sub check_shutdown {
-    my ($timeout) = @_;
-    $timeout //= ONE_MINUTE;
+sub check_shutdown ($timeout = ONE_MINUTE) {
```
perlpunk, line 1835:
```suggestion
sub check_shutdown ($timeout = undef) {
    $timeout //= ONE_MINUTE;
```

testapi.pm
```diff
@@ -1954,9 +1862,7 @@ if C<$timeout> timeout is hit. Default timeout is 60s.
 
 =cut
 
-sub assert_shutdown {
-    my ($timeout) = @_;
-    $timeout //= ONE_MINUTE;
+sub assert_shutdown ($timeout = ONE_MINUTE) {
```
perlpunk, line 1865:
```suggestion
sub assert_shutdown ($timeout = undef) {
    $timeout //= ONE_MINUTE;
```

testapi.pm
```diff
@@ -2326,8 +2216,7 @@ A typical call would look like:
 
 =cut
 
-sub compat_args {
-    my ($def_args, $fix_keys) = splice(@_, 0, 2);
+sub compat_args ($def_args, $fix_keys, @) {
     my %ret;
     for my $key (@{$fix_keys}) {
         $ret{$key} = shift if (scalar(@_) >= 1 && (!defined($_[0]) || !grep { $_ eq $_[0] } keys(%{$def_args})));
```
perlpunk, line 2222: `@_`

testapi.pm
```diff
@@ -616,7 +596,7 @@ similarity_level is 50.
 
 =cut
 
-sub wait_screen_change(&@) {
+sub wait_screen_change : prototype($@) {  # no:style:signatures
```
perlpunk, line 599:
```suggestion
sub wait_screen_change : prototype(&@) {  # no:style:signatures
```
That one is the culprit for the syntax errors

testapi.pm
okurz: That's correct. How should I change that here specifically?

testapi.pm
okurz: yes, I see. What do you suggest?

testapi.pm
okurz: can you explain, why again?

okurz: ok, I get it. To keep the behaviour completely unchanged I need the default parameter *and* the default value if `$timeout` is undef because otherwise `assert_shutdown(undef)` would behave different and actually end up with no valid value in the end.

testapi.pm
okurz: ok. In many more places I now avoided any change but just marked the line explicitly to be not raised by the signature style check

testapi.pm
okurz: I kept the original code in places without going to full signatures

okurz:
In many more places I now avoided any change but just marked the line explicitly to be not raised by the signature style check. With this an automatic style check ensures that we use proper signatures in all places except where explicitly marked as ignorable.

testapi.pm
kalikiana: The [documented signature](https://open.qa/api/testapi/#_send_key) is `send_key($key [, wait_screen_change => $wait_screen_change]);` but I guess we need this for backward compatibility... maybe just leave it, since it'll look awkward even if you give the array argument a name

testapi.pm
kalikiana: I guess `# no:style:signatures` addresses that (you only see that in the full diff now)

testapi.pm
kalikiana: `# no:style:signatures` is fine by me I guess

testapi.pm
perlpunk: for completeness, I don't see the problem with replacing `@_` with `@args`:
```
sub send_key ($key, @args) {
    my %args = (@args == 1) ? (do_wait => +shift args) : @args;
```
I think this was mentioned elsewhere or in a nother signature PR already.
But you decided to leave these for now. Just mentioning my solution.

okurz: ok, left it in the old way for now.

okurz:
Validating in production on w7.o.o:

```
systemctl mask --now openqa-continuous-update.service
curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/1696.patch | patch -p1 --directory=/usr/lib/os-autoinst
```

monitoring jobs

testapi.pm
```diff
@@ -161,7 +158,7 @@ You can use distribution object to implement distribution specific helpers.
 
 =cut
 
-sub set_distribution {
+sub set_distribution {    # no:style:signatures
```
kraih, line 161: Why can't this be converted? Looks straight forward. 🤔

testapi.pm
```diff
@@ -1511,7 +1474,7 @@ You can pass the same optional parameters as for C<type_string> function.
 
 =cut
 
-sub type_password {
+sub type_password {    # no:style:signatures
```
kraih, line 1477: Also looks easy to convert.

testapi.pm
```diff
@@ -1529,7 +1492,7 @@ You can pass the same optional parameters as for C<type_string> function.
 
 =cut
 
-sub enter_cmd {
+sub enter_cmd {    # no:style:signatures
```
kraih, line 1495: Another easy one.

testapi.pm
okurz: I got weird errors that I did not understand converting this. But it's easy to try, just change to signatures and run `prove -I. t/03-testapi.t` and you should see the error – or not when you are doing it better than I could do :)

testapi.pm
okurz: Yes, but I failed :)

testapi.pm
```diff
@@ -812,8 +788,7 @@ Return the given variable as array reference (split variable value by , | or ; )
 
 =cut
 
-sub get_var_array {
-    my ($var, $default) = @_;
+sub get_var_array ($var, $default) {
```
okurz, line 791: ok, this should have been obvious. Failed in https://openqa.opensuse.org/tests/2359937

```
Too few arguments for subroutine 'testapi::get_var_array' at /usr/lib/os-autoinst/testapi.pm line 807.
	testapi::get_var_array("SCC_ADDONS") called at /usr/lib/os-autoinst/testapi.pm line 807
	testapi::check_var_array("SCC_ADDONS", "tcm") called at opensuse/lib/main_common.pm line 1288
	main_common::load_consoletests() called at opensuse/products/opensuse/main.pm line 387
	require opensuse/products/opensuse/main.pm called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 196
	OpenQA::Isotovideo::Utils::try {...} () called at /usr/lib/perl5/vendor_perl/5.26.1/Try/Tiny.pm line 102
	eval {...} called at /usr/lib/perl5/vendor_perl/5.26.1/Try/Tiny.pm line 93
	Try::Tiny::try(CODE(0x563d02a976a0), Try::Tiny::Catch=REF(0x563d01e52150)) called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 213
	OpenQA::Isotovideo::Utils::load_test_schedule() called at /usr/bin/isotovideo line 245
Compilation failed in require at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 196.
	OpenQA::Isotovideo::Utils::try {...} () called at /usr/lib/perl5/vendor_perl/5.26.1/Try/Tiny.pm line 102
	eval {...} called at /usr/lib/perl5/vendor_perl/5.26.1/Try/Tiny.pm line 93
	Try::Tiny::try(CODE(0x563d02a976a0), Try::Tiny::Catch=REF(0x563d01e52150)) called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 213
	OpenQA::Isotovideo::Utils::load_test_schedule() called at /usr/bin/isotovideo line 245

 at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 212.
	OpenQA::Isotovideo::Utils::catch {...} ("Too few arguments for subroutine 'testapi::get_var_array' at "...) called at /usr/lib/perl5/vendor_perl/5.26.1/Try/Tiny.pm line 123
	Try::Tiny::try(CODE(0x563d02a976a0), Try::Tiny::Catch=REF(0x563d01e52150)) called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Utils.pm line 213
	OpenQA::Isotovideo::Utils::load_test_schedule() called at /usr/bin/isotovideo line 245
5356: EXIT 1
```

testapi.pm
```diff
@@ -477,11 +462,8 @@ been matched at the time of the call.
 
 =cut
 
-sub match_has_tag {
-    my ($tag) = @_;
-    if ($last_matched_needle) {
-        return $last_matched_needle->{needle}->has_tag($tag);
-    }
+sub match_has_tag ($tag) {
```
okurz, line 465: https://openqa.opensuse.org/tests/2359829#step/start_install/2 fails with

```
# Test died: Too many arguments for subroutine 'testapi::match_has_tag' at opensuse/tests/installation/start_install.pm line 89.
	testapi::match_has_tag("inst-overview-error-found", 0) called at opensuse/tests/installation/start_install.pm line 89
```

introduced in

```
commit cf919f2f6
Author: Dumitru Gutu <dgutu@suse.de>
Date:   Wed Feb 8 18:49:43 2017 +0100

    UEFI boot partition on raid1

diff --git a/tests/installation/start_install.pm b/tests/installation/start_install.pm
index 6ac47ff09..830bbb0e8 100644
--- a/tests/installation/start_install.pm
+++ b/tests/installation/start_install.pm
@@ -81,6 +81,7 @@ sub run() {
     else {
         sleep 2;    # textmode is sometimes pressing alt-i too early
         send_key $cmd{install};
+        wait_screen_change { send_key 'alt-o' } if match_has_tag('inst-overview-error-found', 0);
         while (check_screen([qw(confirmlicense startinstall)], 5)) {
             last if match_has_tag("startinstall");
             send_key $cmd{acceptlicense}, 1;
```

`git grep 'match_has_tag.*, [0-9]'` in os-autoinst-distri-opensuse reveals multiple other occurences *sigh*

-> https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14944

okurz:
As this PR still turns out to be too big to be easily review I am splitting out yet another PR:
* https://github.com/os-autoinst/os-autoinst/pull/2069

testapi.pm
```diff
@@ -798,8 +775,7 @@ Returns true if test variable C<$variable> is equal to C<$value> or returns C<un
 
 =cut
 
-sub check_var {
-    my ($var, $val) = @_;
+sub check_var ($var, $val) {
```
perlpunk, line 778: There are places where only one argument is used (and the uninitialized warning in the log probably ignored ;-)
```
lib/grub_utils.pm
62:    if (check_var("BOOTFROM", "d") && check_var("AUTOUPGRADE") && check_var("PATCH")) {

lib/kdump_utils.pm
216:    assert_script_run('yast2 kdump fadump enable', 180) if check_var('FADUMP');
```

testapi.pm
```diff
@@ -1904,10 +1856,7 @@ Trigger backend specific power action, can be C<'on'>, C<'off'>, C<'acpi'> or C<
 
 =cut
 
-sub power {
-
-    # params: (on), off, acpi, reset
-    my ($action) = @_;
+sub power ($action) {
```
perlpunk, line 1859:
```
tests/xfstests/run.pm
474:            power('reset', keepconsole => is_pvm);
```

testapi.pm
okurz: I guess I can live with such invalid code causing failures, especially if people ignore warnings.

testapi.pm
okurz: yeah I think someone introduced dead code that is never called. Because this should likely be

```
power_action('reset', keepconsole => is_pvm);
```

perlpunk (commented):
I think I checked pretty much every function now

testapi.pm
perlpunk: Well, these are the only two places and we can make a PR...

testapi.pm
okurz: Merged, fixed

testapi.pm
okurz: Adressed in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14996 but as explained not blocking as it's obviously dead-code

testapi.pm
okurz: Adressed in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14997

testapi.pm
```diff
@@ -197,23 +192,21 @@ influenced.
 
 =cut
 
-sub record_soft_failure {
-    my ($reason) = @_;
+sub record_soft_failure ($reason) {
```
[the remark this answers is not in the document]
okurz: Fixed in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14998

Martchus:
Is it ready for review? Looks generally good (after a brief look).

testapi.pm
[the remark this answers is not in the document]
okurz: merged, deployed, we can go on

testapi.pm
okurz: merged, deployed, we can go on.

okurz:
> Is it ready for review? Looks generally good (after a brief look).

Right, thank you. There were some necessary changes in os-autoinst-distri-opensuse of incorrect use which is fixed and deployed for long. We can now go on. I rebased. But as we don't have complete coverage I am considering splitting the change again to have the fully covered changes separated from the uncovered changes and merge them separately.

okurz:
I created
* https://github.com/os-autoinst/os-autoinst/pull/2078
including only the changes that are covered by automatic tests.

okurz:
deployed again as in https://github.com/os-autoinst/os-autoinst/pull/1696#issuecomment-1133031086

okurz:
Found one problem in https://openqa.opensuse.org/tests/2417074#step/amarok/23


```
# Test died: Too few arguments for subroutine 'testapi::_check_or_assert_sound' at /usr/lib/os-autoinst/testapi.pm line 1825.
	testapi::_check_or_assert_sound("DTMF-159D") called at /usr/lib/os-autoinst/testapi.pm line 1825
```

okurz:
Haven't found anymore problems on openqaworker7. I suggest we continue with some unit tests in a mob programming session.

Martchus (approved):

okurz:
Rebased on top of https://github.com/os-autoinst/os-autoinst/pull/2256 with all CI checks passed, i.e. the changed lines are fully covered in tests (not the complete functions but only the signature relevant lines). So finally the complete PR should be ready including enforcement of using signatures in all perl files in this project.

t/03-testapi.t
```diff
@@ -1047,16 +1049,34 @@ like(exception { ensure_installed }, qr/implement.*for your distri/, 'ensure_ins
...
+
+subtest 'assert/check recorded sound' => sub {
+    $cmds = [];
+    lives_ok { start_audiocapture } 'start_audiocapture can be called';
+    like $cmds->[0]->{filename}, qr/captured\.wav/, 'audiocapture started with expected args' or diag explain $cmds;
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->noop('_snd2png');
+    $mock_basetest->noop('verify_sound_image');
+    ok assert_recorded_sound('foo'), 'assert_recorded_sound can be called';
+    ok check_recorded_sound('foo'), 'check_recorded_sound can be called';
```
Martchus, line 1061: It would be nice to test whether these functions do what they're expected to do.

Martchus (approved):

t/03-testapi.t
okurz: yes, it would be nice but as I explained in https://github.com/os-autoinst/os-autoinst/pull/1696#issuecomment-1410316512 so far I only made sure the functions can be called as-is regardless of what they actually do. There is still work to be done outside this PR to have full test coverage of testapi.pm but this PR is only about using signatures everywhere.

kalikiana (approved):

testapi.pm
```diff
@@ -1959,8 +1941,8 @@ I<Currently only qemu backend is supported.>
 
 =cut
 
-sub save_storage_drives {
-    my $filename ||= $autotest::current_test->{name};
+sub save_storage_drives ($filename = undef) {
```
kraih, line 1944: This seems like a new feature that's not really reflected in the commit message or tests.

okurz: Hm, I thought it to be effectively equivalent to the previous behaviour. What cases do you see where use cases behave differently?

kraih: The previous version of the function did not accept an argument and only ever used `$autotest::current_test->{name}` (the `my $filename ||=` was probably a typo).

kraih: > save_storage_drives([$filename]);

Looks like it was meant to have an argument though, judging by the documentation. More reason to actually test it. :)

okurz: Yeah sure. The previous version did not accept any argument so what I did is change the function to merely do what the documentation promises but still only as an option.

PR #1697, opened 2021-06-20
Simplify bmwqemu::fileContent and mark as deprecated before removal

okurz:
* bmwqemu: Mark fileContent function as deprecated before removing
* Replace bmwqemu::fileContent with cleaner Mojo::File usage
* Remove unused bmwqemu fileContent and save_vars references
* Remove unused use-statements in backend::amt
* Simplify bmwqemu::fileContent with Mojo::File

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -30,6 +30,7 @@ use Time::HiRes qw(sleep gettimeofday);
 use IO::Socket::UNIX 'SOCK_STREAM';
 use IO::Handle;
 use POSIX qw(strftime :sys_wait_h mkfifo);
+use Mojo::File 'path';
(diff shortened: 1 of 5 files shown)


backend/qemu.pm
```diff
@@ -23,7 +23,7 @@ use autodie ':all';
 use base 'backend::virt';
 
 use File::Basename 'dirname';
-use File::Path 'mkpath';
+use File::Path qw(path mkpath);
```
Martchus, line 26: Isn't it supposed to use that function from the `Mojo::File` module?

okurz: absolutely :)

Martchus (approved):

PR #1699, opened 2021-06-21
Fix disabling of compression in qemu

foursixnine:
Commit c670720d452b536eed3544987f116fe01ee0a52f didn't take into account the parameter to disable compression.

1b9b23b28728d0f808a9707f7a936137fc6f8548 in #1695 addresses this, however this is a higher level approach.


kalikiana:
Do we still want this fix? Or is #1695 which got merged enough?

foursixnine:
> Do we still want this fix? Or is #1695 which got merged enough?

Givin' it a rebase, let's see how it turns out

okurz (approved):

PR #1700, opened 2021-06-21
Add test coverage for mouse_tclick testapi function

KevinKipkorir:
Fixes: https://progress.opensuse.org/issues/94315

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -467,6 +467,19 @@ subtest 'assert_and_click' => sub {
     is_deeply($cmds->[-1], {cmd    => 'backend_mouse_set', x      => 100,     y   => 100}, 'assert_and_click succeeds and move to old mouse set');
 };
 
+subtest 'assert_and_tclick' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->redefine(assert_screen => {area => [{x => 1, y => 2, w => 3, h => 4}]});
+    ok(assert_and_tclick('foo', mousehide => 1));
+    for (-2, -4) {
+        is_deeply($cmds->[$_], {bstate => 0, button => 'left', cmd => 'backend_mouse_button'}, 'assert_and_tclick succeeds with bstate => 0');
+    }
+    for (-3, -5) {
+        is_deeply($cmds->[$_], {bstate => 1, button => 'left', cmd => 'backend_mouse_button'}, 'assert_and_tclick succeeds with bstate => 1');
+    }
+    is_deeply($cmds->[-1], {cmd => 'backend_mouse_hide', offset => 0}, 'assert_and_tclick succeeds and hides mouse with mousehide => 1');
+};
+
 subtest 'assert_and_dclick' => sub {
     my $mock_testapi = Test::MockModule->new('testapi');
     $mock_testapi->redefine(assert_screen => {area => [{x => 1, y => 2, w => 3, h => 4}]});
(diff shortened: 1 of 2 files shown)


t/03-testapi.t
```diff
@@ -467,6 +467,19 @@ subtest 'assert_and_click' => sub {
     is_deeply($cmds->[-1], {cmd    => 'backend_mouse_set', x      => 100,     y   => 100}, 'assert_and_click succeeds and move to old mouse set');
 };
 
+subtest 'assert_and_tclick' => sub {
```
okurz, line 470: This seems to be a near-duplicate of the `assert_and_dclick` functionality. To have better maintainable code I suggest to reduce some duplication. One thing you could easily do is to not duplicate the complete "subtest" block but instead extend the existing "assert_and_dclick" code to not need to redefine the mock again. E.g. like the following

```perl
subtest 'assert_and_dclick and assert_and_tclick' => sub {
    my $mock_testapi = Test::MockModule->new('testapi');
    $mock_testapi->redefine(assert_screen => {area => [{x => 1, y => 2, w => 3, h => 4}]});
    # assert_and_dclick tests …
    # assert_and_tclick tests …

File: t/03-testapi.t [hunk 2]
```diff
@@ -479,7 +492,6 @@ subtest 'assert_and_dclick' => sub {
     }
     is_deeply($cmds->[-1], {cmd => 'backend_mouse_hide', offset => 0}, 'assert_and_dclick succeeds and hides mouse with mousehide => 1');
 };
-
```
okurz, line 482: unrelated whitespace changes, please revert

testapi.pm
```diff
@@ -47,6 +47,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   enter_cmd
   hold_key release_key
 
+  assert_screen check_screen assert_and_tclick save_screenshot
```
okurz, line 50: you are adding a new test function but also duplicating other entries. Every function name should only be mentioned once in this list. Also, I don't think we need the function "assert_and_tclick" at all and test API extensions should not be taken lightly. I suggest to not extend the file testapi.pm at all also especially as your commit is called "Add test coverage for …", not "Add new test API function"

okurz (changes requested):
See inline comments.

Also, the commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a progress ticket, the choices you made, etc.

testapi.pm
KevinKipkorir: Just seen your comments.Let me make those changes. Thanks

BillAnastasiadis:
Hi Kevin! Just general pull request etiquette advice, use present tense in the title. In your examples:

- **Make** changes to the file, instead of **made** changes to the file (this also is too vague, doesn't say anything about the changes)
- **Add** a new subtest ... instead of **Added** a new subtest ...
- **Revert** to old state, instead of **Reverting** to old state

Think of it as if you're **ordering** git to do something, instead of explaining what you did.

Martchus:
And most importantly: No merge-commits on your branch. Please rebase your changes instead of merging (e.g. use `pull` with `--rebase`).

KevinKipkorir:
> Hi Kevin! Just general pull request etiquette advice, use present tense in the title. In your examples:
> 
>     * **Make** changes to the file, instead of **made** changes to the file (this also is too vague, doesn't say anything about the changes)
> 
>     * **Add** a new subtest ... instead of **Added** a new subtest ...
> 
>     * **Revert** to old state, instead of **Reverting** to old state
> 
> 
> Think of it as if you're **ordering** git to do something, instead of explaining what you did.

Okay got it.

PR #1702, opened 2021-06-28
Add test coverage for mouse_tclick testapi function.

KevinKipkorir:


okurz (changes requested):
So far I see only test description string content changes. Please also adress the suggestions that were raised in previous attempts. I suggest to prefer updating existing pull requests rather than opening new ones, especially if the suggestions are not addressed as that makes the job of reviewers rather hard to find out which points had been addressed yet and which ones not yet.

okurz:
@KevinKipkorir hi, I think you added https://github.com/os-autoinst/os-autoinst/pull/1702/commits/53a65fa1a27ff77e9e707b1bc78c050fe6251813 as a second commit. Likely this was a mistake because this commit belongs to https://github.com/os-autoinst/os-autoinst/pull/1711 instead. I suggest you use interactive rebase on your branch "new" to remove the second commit again.

t/17-basetest.t
```diff
@@ -355,4 +355,45 @@ subtest 'execute_time' => sub {
...
+subtest result => sub {
+    my $basetest_class = 'basetest';
+    my $mock_basetest  = Test::MockModule->new($basetest_class);
+    $mock_basetest->redefine(_result_add_screenshot => sub { });
+
+    my $basetest = bless({
+            result     => undef,
+            details    => [],
+           
+    }, $basetest_class);
```
Martchus, line 367: This could fit in one line.

t/17-basetest.t
```diff
@@ -355,4 +355,45 @@ subtest 'execute_time' => sub {
     is($test->{execution_time}, 42, 'the execution time is correct');
 };
 
+subtest result => sub {
+    my $basetest_class = 'basetest';
+    my $mock_basetest  = Test::MockModule->new($basetest_class);
+    $mock_basetest->redefine(_result_add_screenshot => sub { });
```
Martchus, line 361:
```suggestion
    $mock_basetest->noop('_result_add_screenshot');
```

t/17-basetest.t
```diff
@@ -355,4 +355,45 @@ subtest 'execute_time' => sub {
...
+    is_deeply($basetest->record_testresult('softfail'), {result => 'softfail'}, 'adding one more "softfail" result');
+    is($basetest->{result}, 'fail', 'test result is still "fail"');
+
+    is_deeply($basetest->record_testresult(), {result => 'unk'}, 'adding one more "unk" result');
+    is($basetest->{result}, 'fail', 'test result is still "fail"');
+
+    is_deeply($basetest->record_testresult('softfail', force_status => 1), {result => 'softfail'}, 'adding one more "softfail" result but forcing the status');
+    is($basetest->{result}, 'softfail', 'test result was forced to "softfail"');
+
+    is($basetest->{result},   ('ok', 'fail', 'softfail'), "True");
```
Martchus, line 396: I guess this code doesn't do what you'd expect it to do.

PR #1704, opened 2021-06-28
Change root to bernhard for s390x

skriesch:
As described in https://progress.opensuse.org/issues/93949, you can not login in the test reconnect_mgmt_console  any more.
That affects all s390x tests. Therefore, I want to switch to the user Bernhard.

That has been recommended in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12757

### DIFF
--- a/consoles/sshIucvconn.pm
+++ b/consoles/sshIucvconn.pm
@@ -32,7 +32,7 @@ sub connect_remote {
     my $zvmguest = get_var('ZVM_GUEST');
 
     # ssh connection to SUT for agetty
-    my $ttyconn = $self->backend->new_ssh_connection(hostname => $hostname, password => $args->{password}, username => 'root');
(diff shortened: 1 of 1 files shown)


okurz:
I don't think the username "bernhard" has any use so far within the scope of os-autoinst. For this also see https://github.com/os-autoinst/os-autoinst/search?p=3&q=bernhard . Also, some lines later there is again "root" being used. How about making the username configurable by a test variable if that helps?

skriesch:
bernhard has been defined as a user in the file testapi.pm via the full name. The test failure has got his roots with this error message "Error connecting to <root@s390linux144>:  at /usr/lib/os-autoinst/testapi.pm line 1750. " in this file.

Vogtinator (changes requested):
I don't think `bernhard` has the necessary permissions to run the commands, so you'll probably have to use `su -c 'cmd'`, read the password prompt and type in the password. The `iucvconn` command below probably needs the same treatment.

There's an additional complication: During installation, it can't use `bernhard` because only root is available at that point.

PR #1705, opened 2021-06-29
Skip empty ISO_X variables when configuring block devices

dcermak:
We are currently checking whether ISO is actually defined but do not perform the
same check for ISO_X, which could be an empty string. In that case os-autoinst
calls qemu on an empty file and fails in a non-obvious way.
Instead we now skip adding ISO_X as a blockdevice if the value is not truthy.

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -401,4 +401,19 @@ subtest 'qemu was killed due to the system being out of memory' => sub {
     unlink("./Core-7.2.iso");
 };
 
+subtest 'qemu is not called on an empty file when ISO_1 is an empty string' => sub {
+    my $mock_proc  = Test::MockModule->new('OpenQA::Qemu::Proc');
+    my $call_count = 0;
+    $mock_proc->redefine(get_img_size => sub {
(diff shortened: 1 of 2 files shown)


Martchus (approved):
Tidy doesn't like it.

t/18-qemu.t
```diff
@@ -401,4 +401,19 @@ subtest 'qemu was killed due to the system being out of memory' => sub {
...
+    $mock_proc->redefine(get_img_size => sub {
+            my ($iso) = @_;
+            $call_count++;
+            die 'get_img_size called on an empty string' unless $iso;
+    });
+
+    my %empty_iso_vars = (ISO_1 => '', NUMDISKS => 0);
+
+    OpenQA::Qemu::Proc->new()->configure_blockdevs('disk', 'raid', \%empty_iso_vars);
+    is($call_count, 0, 'get_img_size call count check');
```
okurz, line 416:
```suggestion
    is $call_count, 0, 'No ISOs with empty names are configured';
```

PR #1711, opened 2021-07-05
Increase code coverage of os-autoinst basetest.pm

KevinKipkorir:
Add test case for subtest function.
Check https://progress.opensuse.org/issues/94952


okurz:
In your git commit message please delete the dot at the end of line

Martchus (changes requested):
My comments on https://github.com/os-autoinst/os-autoinst/pull/1702 apply here as well. Besides, this doesn't add test coverage for a subtest. It adds/extends tests for the `record_testresult` functions. So adjust the commit message accordingly.

PR #1713, opened 2021-07-06
Simplify some code in baseclass

okurz:


backend/baseclass.pm
```diff
@@ -296,17 +291,9 @@ sub run_capture_loop {
 sub check_select_rate {
     my ($buckets, $wait_time_limit, $hits_limit, $id) = @_;
 
-    my $time        = gettimeofday;
+    my $time = gettimeofday;
+    $buckets->{TIME} //= $time;
```
perlpunk, line 295: In the old code `$lower_limit` will be set to `$buckets->{TIME}` if it is set.
I don't see that in the new code.

okurz: right. How about

```suggestion
    my $lower_limit = $buckets->{TIME} //= $time;
```

backend/baseclass.pm
```diff
@@ -168,145 +166,133 @@ sub _write_buffered_data_to_file_handle {
...
-
-=out
-
-=item timeout
-
-run the loop this long in seconds, indefinitely if undef, or until the
-$self->{cmdpipe} is closed, whichever occurs first.
+sub do_capture {
+    my ($timeout, $starttime, $buckets, $wait_time_limit, $hits_limit) = @_;
+    last unless $self->{cmdpipe};
```
Martchus, line 171: `last` outside a loop?

backend/baseclass.pm
```diff
@@ -168,145 +166,133 @@ sub _write_buffered_data_to_file_handle {
...
 
-            my $time_to_next = min($time_to_screenshot, $time_to_update_request, $time_to_timeout);
-            my ($read_set, $write_set) = IO::Select->select($self->{select_read}, $self->{select_write}, undef, $time_to_next);
+sub capture_loop {
+    my ($self, $timeout, $starttime) = @_;
+    # Time slot buckets
+    my $buckets         = {};
+    my $wait_time_limit = $bmwqemu::vars{_CHKSEL_RATE_WAIT_TIME} // 30;
+    my $hits_limit      = $bmwqemu::vars{_CHKSEL_RATE_HITS}      // 15_000;
+    $self->do_capture($timeout, $starttime, $buckets, $wait_time_limit, $hits_limit) while 1;
```
Martchus, line 262: Is this loop supposed to be interrupted by a `die`?

backend/baseclass.pm
```diff
@@ -1017,13 +999,6 @@ sub _time_to_assert_screen_deadline {
     return $self->assert_screen_deadline - time;
 }
 
-sub reduce_deadline {
```
Martchus, line 1020: Let's assume no test distributions use this (the openSUSE one doesn't).

Martchus (changes requested):

backend/baseclass.pm
okurz: Well, backend/baseclass isn't exactly an "outwards facing API" so I don't have doubts removing this :)

backend/baseclass.pm
okurz: ok, that's going too far :) I removed that commit from the PR for now.

backend/baseclass.pm
okurz: reverted

Martchus (approved):
I guess this is good but I've been adding the "Not-ready" label to give more people the chance to review.

okurz:
I guess this was enough time, removing "not-ready" label again

PR #1718, opened 2021-07-08
Improve error message when cloning git repository

perlpunk:
Issue: https://progress.opensuse.org/issues/94678

If a user sees the error message

    Could not find 'abcdef' in complete history

then it's not clear where the git repo was cloned from.

The git directory or URL is now added to the error message.

Also this commit fixes the removal of the `.git` suffix.

### DIFF
--- a/t/34-git.t
+++ b/t/34-git.t
@@ -1,0 +1,103 @@
+#!/usr/bin/perl
+#
+# Copyright (c) 2021 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use File::Path qw(rmtree);
+use FindBin '$Bin';
+use Test::Output qw(combined_from);
+use OpenQA::Isotovideo::Utils qw(checkout_git_repo_and_branch);
+use OpenQA::Test::TimeLimit '5';
+use Test::Warnings ':report_warnings';
+
+use Mojo::File 'tempdir';
+my $tmpdir    = tempdir("/tmp/$FindBin::Script-XXXX");
+my $git_dir   = "$tmpdir/tmpgitrepo";
+my $clone_dir = "$Bin/tmpgitrepo";
+
+chdir $Bin;
+
+my $head        = initialize_git_repo();
+my $case_dir_ok = "file://$git_dir#$head";
+my $case_dir    = "file://$git_dir#abcdef";
+
+subtest 'failing clone' => sub {
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir,
+    );
+    my $path;
+    my $out = combined_from {
+        eval { $path = checkout_git_repo_and_branch('CASEDIR') };
+    };
+    my $error = $@;
+    like $error, qr{Could not find 'abcdef' in complete history in cloned Git repository '\Q$case_dir\E'}, "Error message when trying to clone wrong git hash";
+    like $out,   qr{Cloning git URL.*Fetching more remote objects.*git fetch:}s,                           'git fetch was called to get more commits';
+};
+
+cleanup();
+
+subtest 'successful clone' => sub {
+    my $path;
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir_ok,
+    );
+    my $out = combined_from {
+        $path = checkout_git_repo_and_branch('CASEDIR');
+    };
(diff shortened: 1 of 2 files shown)


t/34-git.t
```diff
@@ -0,0 +1,93 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use File::Path qw(rmtree);
+use FindBin '$Bin';
+use Test::Output qw(combined_from);
```
okurz, line 21: can you please add

```
use OpenQA::Test::TimeLimit '5';
use Test::Warnings ':report_warnings';
```

t/34-git.t
```diff
@@ -0,0 +1,93 @@
...
+subtest 'failing clone' => sub {
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir,
+    );
+    my $path;
+    my $out = combined_from {
+        eval { $path = checkout_git_repo_and_branch('CASEDIR') };
+    };
+    my $error = $@;
+    like $error, qr{Could not find 'abcdef' in complete history in cloned Git repository '\Q$case_dir\E'}, "Error message when trying to clone wrong git hash";
```
okurz, line 44: I would prefer to use single-tick strings whenever we do not want to evaluate variables within strings, e.g. the test descriptions

t/34-git.t
```diff
@@ -0,0 +1,93 @@
...
+
+rmtree $clone_dir;
+
+subtest 'successful clone' => sub {
+    my $path;
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir_ok,
+    );
+    my $out = combined_from {
+        eval { $path = checkout_git_repo_and_branch('CASEDIR') };
```
okurz, line 56: why do you need the eval here?

t/34-git.t
```diff
@@ -0,0 +1,93 @@
...
+        my $git_add = qq{cd $git_dir; echo $_ >>README; git add README; git commit -m"Commit $_" >/dev/null};
+        system $git_add;
+    }
+    chomp(my $head = qx{git -C $git_dir rev-parse HEAD});
+    return $head;
+}
+
+sub cleanup {
+    rmtree $clone_dir;
+}
```
okurz, line 89: I would prefer

```perl
sub cleanup { rmtree $clone_dir }
```

okurz (changes requested):

t/34-git.t
perlpunk: done

t/34-git.t
perlpunk: done

t/34-git.t
perlpunk: an oversight

t/34-git.t
perlpunk: done

perlpunk:
I also added another test without cleaning up the clone directory, because that code path wasn't covered yet.

t/34-git.t
okurz: hm, I see like every *second* test description using single ticks but for example this line double-ticks

t/34-git.t
```diff
@@ -0,0 +1,103 @@
...
+
+subtest 'successful clone' => sub {
+    my $path;
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir_ok,
+    );
+    my $out = combined_from {
+        $path = checkout_git_repo_and_branch('CASEDIR');
+    };
+    is $path, $clone_dir, "checkout_git_repo_and_branch returned correct path";
```
okurz, line 60: same here

t/34-git.t
```diff
@@ -0,0 +1,103 @@
...
+    like $out, qr{Cloning git URL.*Fetching more remote objects}s, 'git clone was called again to fetch a git hash';
+
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir_ok,
+    );
+    $out = combined_from {
+        $path = checkout_git_repo_and_branch('CASEDIR');
+    };
+    is $path, $clone_dir, "checkout_git_repo_and_branch returned correct path";
```
okurz, line 69: same here, also please make the message unique

t/34-git.t
```diff
@@ -0,0 +1,103 @@
...
+    );
+    my $out = combined_from {
+        $path = checkout_git_repo_and_branch('CASEDIR');
+    };
+    is $path, $clone_dir, "checkout_git_repo_and_branch returned correct path";
+    like $out, qr{Cloning git URL.*Fetching more remote objects}s, 'git clone was called again to fetch a git hash';
+
+    %bmwqemu::vars = (
+        CASEDIR => $case_dir_ok,
+    );
```
okurz, line 65: do you actually need to overwrite vars again?

t/34-git.t
perlpunk: hm, I'm simply too used to it. Should be done now.

t/34-git.t
perlpunk: done

t/34-git.t
perlpunk: Yes, because `checkout_git_repo_and_branch` overwrites the var itself

okurz (approved):
line 46 of t/34-git.t still has double-ticks but nevermind :)

PR #1720, opened 2021-07-14
Remove contradicting timezone stamps

KevinKipkorir:
Delete obsolete/contradicting timezone markers observed  in the autoinst-log.txt file.
See: https://progress.opensuse.org/issues/95149

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -517,8 +517,7 @@ sub format_vtt_timestamp {
     $caption .= " --> ";
     $caption .= sprintf(POSIX::strftime("%T.%%03d\n", gmtime($frametime_ms / 1000)), $frametime_ms % 1000);
     # clock value as caption text
-    $caption .= sprintf(POSIX::strftime("[%FT%T.%%03d]\n", localtime($walltime)), 1000 * ($walltime - int($walltime)));
-
+    
     return $caption;
 }
 


okurz (changes requested):
Please see the CI failures in https://github.com/os-autoinst/os-autoinst/pull/1720/checks?check_run_id=3071389694#step:3:587 due to your change

Delete the dot at the end of line of the git commit message subject line. Also, the commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, am example output before and after, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

backend/baseclass.pm
```diff
@@ -517,8 +517,7 @@ sub format_vtt_timestamp {
     $caption .= " --> ";
     $caption .= sprintf(POSIX::strftime("%T.%%03d\n", gmtime($frametime_ms / 1000)), $frametime_ms % 1000);
     # clock value as caption text
-    $caption .= sprintf(POSIX::strftime("[%FT%T.%%03d]\n", localtime($walltime)), 1000 * ($walltime - int($walltime)));
```
Martchus, line 520: This function is for formatting the VTT subtitles. Removing this line will break them. I suppose @StefanBruens had a reason to do the computation this way. The last line is supposed to represent the "wallclock" time (as opposed to the raw frametime).

Martchus (changes requested):
This line is for formatting VTT subtitles and has nothing to do with the referenced issue. To fix the referenced issue (which doesn't mention the video at all), you would need to amend the last `diag` call line in `sub runtest` in `basetest.pm`.

backend/baseclass.pm
perlpunk: Yeah, this is the wrong line. Have a look at `basetest.pm` @KevinKipkorir

PR #1721, opened 2021-07-15
Remove contradicting timezone stamps

KevinKipkorir:
Delete contradicting timezone stamps observed in the autoinst-log.txt
Amending basetest.pm deletes the second timestamp that is UTC contradicting with the initial CEST observed at the beginning of the line
The changes should be should produce the following example ;
  [2021-07-05T17:14:58.348 CEST] [debug] ||| finished grub_test installation (8 s)       instead of
  [2021-07-05T17:14:58.348 CEST] [debug] ||| finished grub_test installation at 2021-07-05 15:14:58 (8 s)
Reference:   #1720
See: https://progress.opensuse.org/issues/95149

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -392,7 +392,7 @@ sub runtest {
 
     $self->done();
     $self->{execution_time} = execution_time($starttime);
-    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), $self->{execution_time}));
+    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, $self->{execution_time}));
     return;
 }
 


basetest.pm
```diff
@@ -392,7 +392,7 @@ sub runtest {
 
     $self->done();
     $self->{execution_time} = execution_time($starttime);
-    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, POSIX::strftime('%F %T', gmtime), $self->{execution_time}));
+    bmwqemu::diag(sprintf("||| finished %s %s at %s (%d s)", $name, $self->{category}, $self->{execution_time}));
```
Martchus, line 395: This change looks wrong, The placeholder string is unchanged (and still contains 4 placeholders) while one of the parameters has been removed. The placeholder for the timestamp should have been removed as well, e.g. `"||| finished %s %s (runtime: %d s)"`. I also added (`runtime: `) because now it isn't that clear anymore what the parenthesis are about.

Martchus: I shortly created a fix: https://github.com/os-autoinst/os-autoinst/pull/1722

perlpunk: Good catch.
I wonder why this wasn't catched as a warning:
```
% perl -wE'say sprintf "%s %s", 23'
Missing argument in sprintf at -e line 1.
23
```

Martchus: Good question. Maybe this is only covered by a forked process (which we don't take into account when checking for warnings)?

okurz: hm, sorry. I thought our tests would have covered that

Martchus: I'm sure the line is invoked somewhere within our tests. Whether we have a check is unfortunately a different story.

PR #1727, opened 2021-07-20
Fix warning about expected termination

okurz:


backend/qemu.pm
```diff
@@ -1137,7 +1137,7 @@ sub process_qemu_output {
     my ($buffer) = @_;
     for my $line (split(/\n/, $buffer)) {
         die "QEMU: Shutting down the job" if $line =~ m/key event queue full/;
-        if ($line =~ /^qemu-system-.*: (?!terminating on signal)/) {
+        if ($line =~ /^ qemu-system-.*: (?!terminating on signal)/) {
```
Martchus, line 1140: Maybe just allow any amount of white-space characters here?

okurz: I tried and failed :,( Will try again

okurz: Sorry, I don't know what's wrong. Trying manually `print "COLOR WARN: $line\n" if $line =~ /\s*qemu-system-.*/` works fine but if I use `$line =~ /^\s*qemu-system-.*: (?!terminating on signal)/` in backend/qemu.pm it also matches the line with "terminating on signal"

Martchus: `\s?` should do it

okurz: that's what I used and stated that it does not work, right?

perlpunk: it seems to work in a simple script though:
```perl
@s = (
  "terminating on signal",
  "qemu-system-x: terminating on signal",
  " qemu-system-x: terminating on signal",
  "  qemu-system-x: terminating on signal",
  "qemu-system-x: foo",
  " qemu-system-x: foo",
  "  qemu-system-x: foo",
);
for (@s) {
  say ">>$_<<";
  say 1 if m/^\s*qemu-system-.*: (?!terminating on signal)/
}
__END__
>>terminating on signal<<
>>qemu-system-x: terminating on signal<<
>> qemu-system-x: terminating on signal<<
>>  qemu-system-x: terminating on signal<<
>>qemu-system-x: foo<<
1
>> qemu-system-x: foo<<
1
>>  qemu-system-x: foo<<
1
```

Martchus: @okurz I only saw your "Will try again" comment when I wrote mine. Very strange that it doesn't work.

okurz: could you take over the changes locally and try where I am doing something wrong?

Martchus: I'll try

perlpunk: The problem is the second `:` in the line, e.g.
```
qemu-system-i386: rminating on signal 15 from pid 1500 (/path/to/os-autoinst/t/../isotovideo: backen)
```
So `qemu-system-.*:` matches until the `isotovideo:`

Try:
```
/^\s*qemu-system-[^:]+: (?!terminating on signal)/
```

okurz: Thanks a lot to both of you

Martchus:
Looks like the test fails now at a further check:

```
 8: #     doesn't match '(?^:save_tmp_file returned expected file)'
8: Bailout called.  Further testing stopped:  testresults/result-modify_and_upload_file.json failed
8: 
8: #   Failed test 'Result in testresults/result-modify_and_upload_file.json is ok'
8: #   at ./99-full-stack.t line 83.
8: #          got: 'fail'
8: #     expected: 'ok'
8: FAILED--Further testing stopped: testresults/result-modify_and_upload_file.json failed
```

okurz:
Yes, that's the problem I mentioned we now have in all os-autoinst tests

perlpunk:
I debugged the failure. It is failing at the wget call in `t/data/tests/tests/modify_and_upload_file.pm`
```
Connecting to 10.0.2.2:15223 (10.0.2.2:15223)
# wget: can't connect to remote host (10.0.2.2): Network is unreachable
```

okurz:
interesting, so something busted network?

Martchus:
Ah, btw, I've also noticed the problem locally but thought it would be a local issue.

PR #1723, opened 2021-07-20
git subrepo pull (merge) --force external/os-autoinst-common

okurz:
subrepo:
  subdir:   "external/os-autoinst-common"
  merged:   "9acf1fb0"
upstream:
  origin:   "git@github.com:os-autoinst/os-autoinst-common.git"
  branch:   "master"
  commit:   "9acf1fb0"
git-subrepo:
  version:  "0.4.3"
  origin:   "???"
  commit:   "???"

PR #1729, opened 2021-07-23
Display curl progress meter in upload_logs() & upload_assets()

pdostal:
In `upload_logs()` the `curl` does not display the progress meter.
This is because `curl` by default displays the response output.
We display the meter by redirecting output to `/dev/$serialdev`.

 This is enabled only when `UPLOAD_METER` variable is present.

 * Test runs: [mau-extratests1](http://pdostal-server.suse.cz/tests/12380) [mau-extratests1 UPLOAD_METER](http://pdostal-server.suse.cz/tests/12381) [publiccloud_containers UPLOAD_METER](http://pdostal-server.suse.cz/tests/12379) [publiccloud_containers](http://pdostal-server.suse.cz/tests/12378)

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -2233,6 +2233,7 @@ sub upload_logs {
     my $basename = basename($file);
     my $upname   = $args{log_name} || ($autotest::current_test->{name} . '-' . $basename);
     my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
+    $cmd .= show_curl_progress_meter();
     $cmd .= autoinst_url("/uploadlog/$basename");
     if ($failok) {
         # just use script_run so we don't care if the upload fails
@@ -2280,6 +2281,7 @@ sub upload_asset {
     bmwqemu::log_call(file => $file, public => $public, nocheck => $nocheck);
     my $cmd = "curl --form upload=\@$file ";
     $cmd .= "--form target=assets_public " if $public;
+    $cmd .= show_curl_progress_meter();
     my $basename = basename($file);
     $cmd .= autoinst_url("/upload_asset/$basename");
     if ($nocheck) {
@@ -2312,5 +2314,19 @@ sub compat_args {
     map { $ret{$_} //= $def_args->{$_} } keys(%{$def_args});
     return %ret;
 }
+
+=head2 show_curl_progress_meter
+
+Helper function to alter the curl command to show progress meter.
+Progress meter is shown only when the server output is redirected.
+This works only when uploading where the output is not lately used.
+
+    show_curl_progress_meter( $cmd )
+
+A typical call would look like:
+
+    $cmd .= show_curl_progress_meter($cmd);
+=cut
+sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
 1;
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -35,6 +35,7 @@ PAUSE_ON_SCREEN_MISMATCH;boolean;0;Pause test execution on the next screen misma
 PAUSE_ON_NEXT_COMMAND;boolean;0;Pause test execution on the next test API command. Same notes as for `PAUSE_AT` apply.
 _QUIET_SCRIPT_CALLS;boolean;0;Add quiet flag to all the calls to script_run, script_output and validate_script_output. It will omit all the squares "wait_serial expected" on the Details view of the test case. This option might be useful for serial terminal tests.
 AUTOINST_URL_HOSTNAME;string;;hostname or IP address of host running the autoinst webserver endpoint, defaults to the local IP address within the qemu network for the qemu backend or the `WORKER_HOSTNAME` otherwise.
+UPLOAD_METER;boolean;0;Display curl progress meter in `upload_logs()` and `upload_assets()` subroutines.
 
 |====================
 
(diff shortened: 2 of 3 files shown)


asmorodskyi (approved):
LGTM

testapi.pm
```diff
@@ -2232,7 +2232,8 @@ sub upload_logs {
     bmwqemu::log_call(file => $file, failok => $failok, timeout => $timeout, %args);
     my $basename = basename($file);
     my $upname   = $args{log_name} || ($autotest::current_test->{name} . '-' . $basename);
-    my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
+    # Redirect the output so the progress meter is displayed 
+    my $cmd = "curl -o /dev/$serialdev --form upload=\@$file --form upname=$upname ";
```
okurz, line 2236: this command is used *a lot*. And adding more to type can have a big accumulated effect. I would rather not add that by default but only when it's necessary. Maybe this is something you want to optionally enable with a variable?

okurz (changes requested):

testapi.pm
pdostal: I chose a variable called `UPLOAD_PROGRESS` - is there a place for documenting it?

okurz: the place that every test writer should actually know :) https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc

testapi.pm
```diff
@@ -2232,7 +2232,9 @@ sub upload_logs {
     bmwqemu::log_call(file => $file, failok => $failok, timeout => $timeout, %args);
     my $basename = basename($file);
     my $upname   = $args{log_name} || ($autotest::current_test->{name} . '-' . $basename);
-    my $cmd      = "curl --form upload=\@$file --form upname=$upname ";
+    my $cmd      = "curl -o /dev/$serialdev --form upload=\@$file --form upname=$upname ";
```
okurz, line 2235: still need to remove the `-o /dev/$serialdev` from here

testapi.pm
```diff
@@ -2280,6 +2282,8 @@ sub upload_asset {
     bmwqemu::log_call(file => $file, public => $public, nocheck => $nocheck);
     my $cmd = "curl --form upload=\@$file ";
     $cmd .= "--form target=assets_public " if $public;
+    # Redirect the output so the progress meter is displayed 
+    $cmd .= "-o /dev/$serialdev " if get_var('UPLOAD_METER');
```
okurz, line 2286: this is duplication. Even if it's just two lines better extract into a function that alters the variable `$cmd`

okurz (changes requested):
Please also add a test for this, e.g. in t/03-testapi.t

testapi.pm
```diff
@@ -2313,4 +2315,24 @@ sub compat_args {
...
+    $cmd .= show_curl_progress_meter($cmd);
+=cut
+sub show_curl_progress_meter {
+    my $cmd = shift;
+
+    # Redirect the output so the progress meter is displayed 
+    $cmd .= "-o /dev/$serialdev " if get_var('UPLOAD_METER');
+
+    return $cmd;
+}
```
okurz, line 2336: with this you are adding `$cmd` twice. Ok, I think it's clearer if we let the function just return necessary additional parameters. How about just

```suggestion
# Redirect the output so the progress meter is displayed
sub show_curl_progress_meter { "-o /dev/$serialdev " if get_var('UPLOAD_METER') }
```

and then use like this:

```perl
$cmd .= show_curl_progress_meter;
```

okurz (changes requested):

okurz (approved):

okurz (changes requested):
Oops, sorry. Please also add a test for this, e.g. in t/03-testapi.t

pdostal:
@okurz sorry that it took me a while… I now finally have passing test for `show_curl_progress_meter()` but:
 1) Neither `upload_assets()` neither `upload_logs()` are covered in `t/03-testapi.t` and *codecov* is still failing.
 2) When `UPLOAD_METER` is not set it returns `'0'` - isn't that problem?

testapi.pm
```diff
@@ -2313,4 +2315,18 @@ sub compat_args {
...
+Progress meter is shown only when the server output is redirected.
+This works only when uploading where the output is not lately used.
+
+    show_curl_progress_meter( $cmd )
+
+A typical call would look like:
+
+    $cmd .= show_curl_progress_meter($cmd);
+=cut
+sub show_curl_progress_meter { "-o /dev/$serialdev " if get_var('UPLOAD_METER') }
```
okurz, line 2330: > @okurz sorry that it took me a while… I now finally have passing test for `show_curl_progress_meter()` but:
> 
>     1. Neither `upload_assets()` neither `upload_logs()` are covered in `t/03-testapi.t` and _codecov_ is still failing.

Well, would be good to have that covered as well but I am not demanding that from you :)

>     2. When `UPLOAD_METER` is not set it returns `'0'` - isn't that problem?

well, I guess this is something you could cover with tests for `upload_assets/upload_logs` ;)

Better make the function always return something well defined, e.g.

```suggestion
sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
```

pdostal:
Now, I'm running verification runs and the tests should be fine.
However the CI is failing and I don't think that it is my fault this time.

perlpunk:
> Now, I'm running verification runs and the tests should be fine.
> However the CI is failing and I don't think that it is my fault this time.

Seems this is because https://github.com/os-autoinst/os-autoinst/pull/1723 was merged with a failing CI already

okurz (approved):

pdostal:
Verification runs updated. Thank for your help everyone!

perlpunk:
@pdostal rebasing should fix the CI

PR #1730, opened 2021-07-26
mmapi: Prevent API failures due to short server unavailabilities

okurz:
When retrying multi-machine API calls to a server we should wait between
retries to increase the likelyhood to encounter a server that is able to
accept the request again.

Related progress issue: https://progress.opensuse.org/issues/96010


Amrysliu:
LGTM.

ilausuch (approved):

PR #1732, opened 2021-07-27
Update tidy to 20210717

okurz:


okurz:
checks fail still finding an older tidy version in the tumbleweed container repos. https://build.opensuse.org/package/show/openSUSE:Factory/perl-Perl-Tidy is up-to-date, I guess I need to wait for some snapshot publishing before the container is rebuilt with the updated version

okurz:
https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst_dev/containers/x86_64 shows that tidy is in with version 20210717 so now we should be good

perlpunk (approved):

PR #1734, opened 2021-07-27
Simplify code in qemu modules

okurz:


backend/qemu.pm
```diff
@@ -534,72 +508,42 @@ sub load_snapshot {
...
-    my @names;
-    if ($bmwqemu::vars{VIRTIO_CONSOLE}) {
-        for (my $i = 0; $i < ($bmwqemu::vars{VIRTIO_CONSOLE_NUM} // 1); $i++) {
-            my $name = 'virtio_console' . ($i ? $i : '');
-            push(@names, $name);
-        }
-    }
-    return @names;
+sub virtio_console_names {
+    return undef unless $bmwqemu::vars{VIRTIO_CONSOLE};
```
perlpunk, line 527: As `virtio_console_names` is called in list context, an `undef` item will end up in the array, e.g. `my @virtio_consoles = virtio_console_names`

Better `return ()`

doc/backend_vars.asciidoc
```diff
@@ -85,7 +85,7 @@ CDMODEL;see qemu -device ?;undef;Storage device for virtualized CD
 DELAYED_START;boolean;;delay vm cpu start until resume_vm() is called
 HDDFORMAT;;;HDD format
 HDDMODEL;see qemu -device ?;virtio-blk;Storage device for virtualized HDD.
-HDDMODEL_$i;see qemu -device ?;virtio-blk;Storage device for virtualized HDD. Overrides global HDDMODEL for HDD_$i
+HDDMODEL_$_;see qemu -device ?;virtio-blk;Storage device for virtualized HDD. Overrides global HDDMODEL for HDD_$i
```
perlpunk, line 88: What is this change about?

okurz: This is a necessary change as our CI checks include a crosscheck of variable references in documentation. As we do not reference `HDDMODEL_$i` anymore but now `HDDMODEL_$_` this is what we need to mention here in the doc table as well

backend/qemu.pm
okurz: done, thanks

backend/qemu.pm
```diff
@@ -534,72 +508,42 @@ sub load_snapshot {
...
 }
 
 sub delete_virtio_console_fifo {
     my $self = shift;
-    for my $name ($self->get_virtio_console_fifo_names) {
-        if (-e $name) {
-            unlink($name) or bmwqemu::fctwarn("Could not unlink $name: $!");
-        }
-    }
+    unlink or bmwqemu::fctwarn("Could not unlink $_ $!") for grep { -e } virtio_console_fifo_names;
```
Martchus, line 546: Even if `$_` is passed implicitly here to `unlink` I would rather pass it explicitly.

backend/qemu.pm
```diff
@@ -534,72 +508,42 @@ sub load_snapshot {
...
+sub virtio_console_fifo_names { map { $_ . '.in', $_ . '.out' } virtio_console_names }
+
+sub console_fifo {
+    my ($name) = @_;
+    return bmwqemu::fctwarn("Fifo pipe '$name' already exists!") if -e $name;
+    mkfifo($name, 0666) or bmwqemu::fctwarn("Failed to create pipe $name: $!");
 }
 
 sub create_virtio_console_fifo {
     my $self = shift;
```
Martchus, line 540: Looks like you can remove this line here as well.

backend/qemu.pm
```diff
@@ -534,72 +508,42 @@ sub load_snapshot {
...
-            bmwqemu::fctwarn("Fifo pipe '$name' already exists!");
-        } else {
-            mkfifo($name, 0666) or bmwqemu::fctwarn("Failed to create pipe $name: $!");
-        }
-    }
+    console_fifo($_) for virtio_console_fifo_names;
 }
 
 sub delete_virtio_console_fifo {
     my $self = shift;
```
Martchus, line 545: Looks like you can remove this line here as well.

okurz: done

backend/qemu.pm

backend/qemu.pm

Martchus (approved):

PR #1717, opened 2021-07-08
Set default value for HDDSIZEGB as in QEMU or PVM backends

mloviska:
As a follow-up for [SLE Micro: update schedule to properly boot s390x image](os-autoinst/os-autoinst-distri-opensuse#12759)
and to provide fix for [image size failure check](https://openqa.suse.de/tests/6397443#step/image_checks/7) we need
to create backing image file of specified size as it is in [SLEM-qemu](https://openqa.suse.de/tests/6299772#step/image_checks/4) or [MicroOS](https://openqa.opensuse.org/tests/1829744#step/image_checks/4).

```
[2021-07-08T10:50:30.408 CEST] [debug] <<< backend::console_proxy::__ANON__(wrapped_call={
    "console" => "svirt",
    "function" => "add_disk",
    "args" => [
                {
                  "file" => "/var/lib/openqa/share/factory/hdd/SUSE-MicroOS.s390x-5.1.0-Default-kvm-Build6.3.qcow2",
                  "dev_id" => "a",
                  "backingfile" => 1,
                  "size" => "30G"
                }
              ],
    "wantarray" => undef
  })
```

PR #1736, opened 2021-07-29
Create backing file of arbitrary size for disk image

mloviska:
For bootable disk image testing in SLEM,MicroOS or JeOS we need to
create a disk image of different size than the image itself is shipped.
For QEMU backend this works already perfectly as this is handled well in
`OpenQA/Qemu` code.
The previous [attempt](https://github.com/os-autoinst/os-autoinst/pull/1717) lead to openQA test run failures in loading [pre-installed systems](https://openqa.suse.de/tests/6408034#step/boot_to_desktop/9)

- [Original issue in slem@s390x z/KVM](https://openqa.suse.de/tests/6590998#step/image_checks/7)
- VRs:
    * [sle-15-SP4-Online-s390x-Build18.1-create_hdd_textmode@s390x-kvm-sle12](http://kepler.suse.cz/tests/6274)
    * [sle-15-SP4-Online-s390x-Build18.1-qemu@s390x-kvm-sle12](http://kepler.suse.cz/tests/6276)
    * [slem image-checks](http://kepler.suse.cz/tests/6268#step/image_checks/4)
    * [JeOS with HDDSIZEGB_1=100](http://kepler.suse.cz/tests/6259)
    * [JeOS with HDDSIZEGB_1=undef](http://kepler.suse.cz/tests/6258)

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -27,6 +27,7 @@ use XML::LibXML;
 use File::Temp 'tempfile';
 use File::Basename;
 use Class::Accessor 'antlers';
+use Mojo::JSON qw(decode_json);
 
 use backend::svirt;
 use testapi qw(get_var get_required_var check_var set_var);
@@ -417,7 +418,13 @@ sub add_disk {
             }
             else {
                 $file = $basedir . $file;
-                $self->run_cmd(sprintf("qemu-img create '${file}' -f qcow2 -b '$basedir/%s'", $file_basename))
+                # requested expected value in GBs, if there is any passed as an argument
+                my $size = $args->{size} // 0;
+                # expected value in Bytes
+                my (undef, $json) = $self->run_cmd("qemu-img info --output=json $args->{file}", wantarray => 1);
+                my $image_vsize = decode_json($json)->{'virtual-size'};
+                $size = (($size * 1024 * 1024 * 1024) <= $image_vsize) ? $image_vsize : $size . 'G';
+                $self->run_cmd(sprintf("qemu-img create '${file}' -f qcow2 -b '$basedir/%s' ${size}", $file_basename))
                   && die 'qemu-img create with backing file failed';
             }
         }


mloviska:
Hello @Soulofdestiny @jlausuch @Vogtinator, you have a look, pls?

okurz (changes requested):
Please add tests.

Also I consider the function add_disk very complicated already. I suggest to extract some parts before adding more.

consoles/sshVirtsh.pm
```diff
@@ -417,7 +418,17 @@ sub add_disk {
             }
             else {
                 $file = $basedir . $file;
-                $self->run_cmd(sprintf("qemu-img create '${file}' -f qcow2 -b '$basedir/%s'", $file_basename))
+                # requested expected value in GBs, if there is any passed as an argument
+                my $size = $args->{size} // 0;
+                # expected value in Bytes
+                my (undef, $json) = $self->run_cmd("qemu-img info --output=json $args->{file}", wantarray => 1);
+                my $image_vsize = decode_json($json)->{'virtual-size'};
+                if (($size * 1024 * 1024 * 1024) <= $image_vsize) {
```
jlausuch, line 426:
```suggestion
                $size = (($size * 1024 * 1024 * 1024) <= $image_vsize) ? $image_vsize : $size . 'G'
```

if you prefer 1 liner, but ok with current implementation :)

jlausuch:
LGTM, but I guess you need to add the unit tests.

consoles/sshVirtsh.pm
```diff
@@ -417,7 +418,17 @@ sub add_disk {
...
             else {
                 $file = $basedir . $file;
-                $self->run_cmd(sprintf("qemu-img create '${file}' -f qcow2 -b '$basedir/%s'", $file_basename))
+                # requested expected value in GBs, if there is any passed as an argument
+                my $size = $args->{size} // 0;
+                # expected value in Bytes
+                my (undef, $json) = $self->run_cmd("qemu-img info --output=json $args->{file}", wantarray => 1);
+                my $image_vsize = decode_json($json)->{'virtual-size'};
+                $size = (($size * 1024 * 1024 * 1024) <= $image_vsize) ? $image_vsize : $size . 'G'
+                    $size = $image_vsize;
```
jlausuch, line 427: I guess you don´t need these lines any more.

consoles/sshVirtsh.pm
```diff
@@ -417,7 +418,17 @@ sub add_disk {
             }
             else {
                 $file = $basedir . $file;
-                $self->run_cmd(sprintf("qemu-img create '${file}' -f qcow2 -b '$basedir/%s'", $file_basename))
+                # requested expected value in GBs, if there is any passed as an argument
+                my $size = $args->{size} // 0;
+                # expected value in Bytes
+                my (undef, $json) = $self->run_cmd("qemu-img info --output=json $args->{file}", wantarray => 1);
+                my $image_vsize = decode_json($json)->{'virtual-size'};
+                $size = (($size * 1024 * 1024 * 1024) <= $image_vsize) ? $image_vsize : $size . 'G'
```
jlausuch, line 426:
```suggestion
                $size = (($size * 1024 * 1024 * 1024) <= $image_vsize) ? $image_vsize : $size . 'G';
```

mloviska: I have rebased it, thanks!

jlausuch (commented):
LGTM but would like os-autoinst experts to review.

okurz:
As you can see in the failed coverage check the patch coverage is too low. Meaning that there is no or too little test coverage of the changed lines of code. So https://github.com/os-autoinst/os-autoinst/pull/1736#pullrequestreview-718236676 still applies. If you state explicitly that you don't want to or can't do that then we can try to help or accept the change without as exception but for long-term sustainability I suggest to look into proper tests

jlausuch:
> As you can see in the failed coverage check the patch coverage is too low. Meaning that there is no or too little test coverage of the changed lines of code. So [#1736 (review)](https://github.com/os-autoinst/os-autoinst/pull/1736#pullrequestreview-718236676) still applies. If you state explicitly that you don't want to or can't do that then we can try to help or accept the change without as exception but for long-term sustainability I suggest to look into proper tests

Since this is a minor change on something that was not even covered before (tests are not covering function `add_disk`), would it be ok to postpone the UT for later (a poo could be created for it) and merge this as it is? Anyway, with this PR, the coverage is just reduced 0.04% which is not very significant...

jlausuch:
Thanks!

PR #1737, opened 2021-07-30
qemu: remove medium from cd when ejecting

marmarek:
Just ejecting the cd does not prevent the SUT to access it - the tray
can be injected again (even by the OS itself). This makes the 'eject_cd'
much less useful in tests that verify if the system indeed (or the
installer itself) works independently of the boot CD/DVD.

Improve this by executing also 'blockdev-remove-medium' QMP command, that
indeed remove the medium after ejecting the CD. Note this command
accepts only 'id' parameter (doesn't support 'device' param), so use
only that. In practice, on calls just 'eject_cd()' anyway.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -104,10 +104,13 @@ sub power {
 
 sub eject_cd {
     my ($self, $args) = @_;
+    die "'device' parameter is not supported anymore, use 'id'" if defined $args->{device};
+    my $id = $args->{id} // 'cd0-device';
     $self->handle_qmp_command({execute => 'eject', arguments => {
-                (defined $args->{id} || !defined $args->{device} ? (id => $args->{id} // 'cd0-device') : (device => $args->{device})),
+                id    => $id,
                 force => (!defined $args->{force} || $args->{force} ? Mojo::JSON->true : Mojo::JSON->false)
     }});
+    $self->handle_qmp_command({execute => 'blockdev-remove-medium', arguments => {id => $id}});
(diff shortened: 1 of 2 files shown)


backend/qemu.pm
```diff
@@ -108,6 +108,9 @@ sub eject_cd {
                 (defined $args->{id} || !defined $args->{device} ? (id => $args->{id} // 'cd0-device') : (device => $args->{device})),
                 force => (!defined $args->{force} || $args->{force} ? Mojo::JSON->true : Mojo::JSON->false)
     }});
+    $self->handle_qmp_command({execute => 'blockdev-remove-medium', arguments => {
+                id => $args->{id} // 'cd0-device'
```
Martchus, line 112: I'd avoid repeating the `$args->{id} // 'cd0-device'` part (and rather save it in a variable). Then this invocation likely fits into one line.

Martchus (changes requested):
* I would prefer to remove handling for the `device` parameter completely then. I remember that last time I touched the code I kept the `device` parameter for the sake of compatibility. However, it is not used anyways (at least in the openSUSE distribution) so I guess we don't need to be that strict. Otherwise I'd only execute the 2nd qmp command if the device parameter is not used to at least retain the current behavior in that case.
* Please extend/adjust the "eject cd" subtest in `18-backend-qemu.t`.

okurz (approved):

Martchus (approved):

PR #1738, opened 2021-07-30
qemu: Ensure chattr ends within reasonable time to prevent potential blockage

okurz:
Somes openQA tests calling isotovideo hand indefinitely until the job is
aborted after multi-hour timeout. The last output is seen from the
"chattr" command which might hang or just be the last successfully
executed command. To ensure that chattr is not stuck too long this
commits adds an according timeout command.

Related progress issue: https://progress.opensuse.org/issues/81828

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -335,6 +335,7 @@ sub init_blockdev_images {
         die "init_blockdev_images: '@$qicmd' failed after $tries tries: $@" if $@;
     }
 
+    bmwqemu::diag('init_blockdev_images: Finished creating block devices');
     $self->blockdev_conf->mark_all_created();
 }
 
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -754,6 +754,7 @@ sub start_qemu {
     # do not use autodie here, it can fail on tmpfs, xfs, ...
     run_diag('/usr/bin/chattr', '-f', '+C', $basedir);
 
+    bmwqemu::diag('Configuring storage controllers and block devices');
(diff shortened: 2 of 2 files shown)


okurz:
seems like t/34-git.t destroyed my git config of os-autoinst

OpenQA/Qemu/Proc.pm
```diff
@@ -335,6 +335,7 @@ sub init_blockdev_images {
         die "init_blockdev_images: '@$qicmd' failed after $tries tries: $@" if $@;
     }
 
+    die "reached init_blockdev_images after runcmd";
```
Martchus, line 338: This means the next line is dead-code. Is this a leftover from debugging?

backend/qemu.pm
```diff
@@ -762,6 +765,7 @@ sub start_qemu {
         $self->{proc}->configure_blockdevs($bootfrom, $basedir, $vars);
         $self->{proc}->configure_pflash($vars);
     }
+    bmwqemu::diag('initializing block device images');
```
Martchus, line 768: You could start the message with a capital letter, just for being consistent with the previous message.

Martchus (changes requested):
I've already wanted to mention that the commit meta-data is broken.

OpenQA/Qemu/Proc.pm
okurz: yes, replaced by "bmwqemu::diag" with a better message string

backend/qemu.pm

okurz:
> seems like t/34-git.t destroyed my git config of os-autoinst

Fixed in https://github.com/os-autoinst/os-autoinst/pull/1739

Martchus (approved):

PR #1740, opened 2021-08-09
Allow one to easily configure a build with commonly useful warnings

Martchus:
* This change does *not* change anything unless one adds
  `-DENABLE_WARNINGS` or `-DTREAT_WARNINGS_AS_ERRORS` to the CMake
  arguments.
* At this point this is mainly intended to configure a development build
  with useful warnings enabled more easily. In other context (e.g. during
  rpm package builds) one can still configure warnings as they make sense
  in this context (e.g. by using certain rpm macros).
* If all warnings are fixed, it would make sense to enable both settings
  for CI builds.


Martchus:
Btw, I leave it for later to actually fix the problems (or drop some of the warnings at least for some source files again) so we can enable them within CI runs.

asdil12 (approved):
LGTM

PR #1741, opened 2021-08-10
Another attempt/method for running tests on physical laptops

marmarek:
This is attempt to run tests on laptops that do not support Intel AMT. This uses a HDMI->IP converter (https://www.berrybase.de/en/audio-video/changeover-switch-distributor-repeater-converter/hdmi-sender-f-252-r-video-252-ber-ip to be very specific), plus a RaspberryPI playing as an USB keyboard/mouse gadget.
But it's meant to be flexible enough to support various other setups too. Some more details in the commit messages.

For the DYI RPi setup, it needs a script to actually handle the events and configure USB gadget. This part is in very much in development still (it works already when connected to Linux, but fails to control BIOS/bootloader). Here it is: https://gist.github.com/marmarek/5c44ffeb2f36e106b4d34e7e8780c208. It is much more specific to my setup than this PR, but should be still useful for others. Do you want me to add it to this PR too (which dir?) or keep it separate?
I experiment with emulating USB mass storage there too (use to connect `ISO` to the device this way), but it doesn't work for me yet - at least not for booting from it.

Marking it as WIP, as I'm still finishing integrating it with real tests and may still find a need for significant changes. But simple tests already pass with this. The code will need some cleanup too.

Any thoughts about the design and/or the general idea?

Some things I've stumbled on here:
 - forced scaling to 1024x768 is weird here - the device captures at 1920x1080 and the aspect ratio mismatch makes needles sharing with other tests pretty much useless; can this be disabled? or at least preserve the aspect ratio?
 - I've decided to not support separate press+release keyboard events, as it simplifies interactions with the external script. This seems to work for me, but I wonder if there are some important use cases that would be broken by this decision? Basically `hold_key` and `release_key` are not supported here.
 - my handling of external processes, pipes etc in Perl may be awkward, I use both `autoflush(1)` and sometimes `flush` explicitly. I've added also `disable_consoles` call in `do_stop_vm`, to cleanup processes. This works for me, but I'd greatly appreciate hints what is "the right way" to do it.


marmarek:
Also:
- I fully expect unit tests to fail
- I'd appreciate hints how to write unit tests for this new console driver and/or generalhw backend additions.

Martchus:
* I guess forcing a certain resolution could be lifted somehow, at least explicitly by this backend. Maybe this also requires adjustments within the web UI (needle diff, live view, needle editor).
* I'd say if certain functions don't work it is ok if the limitations are documented.
* I also don't know whether stopping ffmpeg within `do_stop_vm` is the right way but it doesn't seem too odd. The backend's base class also stops the video encoder there. The QEMU backend is using `Mojo::IOLoop::ReadWriteProcess` to start QEMU but I'm not sure whether using this module would be a big win here. The backend's base class also just uses normal Perl process handling to start sub processes (like ffmpeg which can also be used as video encoder).

marmarek:
>     * I guess forcing a certain resolution could be lifted somehow, at least explicitly by this backend. Maybe this also requires adjustments within the web UI (needle diff, live view, needle editor).

Maybe have a variable that defines the resolution? Having it fully dynamic (changing when SUT decides to change the resolution) may be too fragile, but being able to specify it in the machine settings should solve the aspect ratio matching problem, I think.

>     * I also don't know whether stopping ffmpeg within `do_stop_vm` is the right way but it doesn't seem too odd. The backend's base class also stops the video encoder there. The QEMU backend is using `Mojo::IOLoop::ReadWriteProcess` to start QEMU but I'm not sure whether using this module would be a big win here. The backend's base class also just uses normal Perl process handling to start sub processes (like ffmpeg which can also be used as video encoder).

Is there expectation for (any) console to work after `do_stop_vm`? Or, can it be harmful to call `disable_consoles` there? Otherwise, I think calling `disable_consoles` in `do_stop_vm` should be the right thing to do (and at least harmless for other backends, but that I have _not_ tested).

Martchus:
> Maybe have a variable that defines the resolution?

Sounds reasonable.

> Is there expectation for (any) console to work after do_stop_vm?

The baseclass also closes all SSH connections when stopping the VM and these connections are used for certain SSH-based consoles. So it should be ok (but maybe someone else knows more; I'm not dealing with that code very often).

marmarek:
Just an update: in the meantime I came across pikvm.org and I took some inspiration from there (especially: using HDMI adapter connected directly, instead of using IP in the middle). The parts in this PR require minimal changes, but setup-specific scripts need way more changes, and I want to test it properly first.

okurz:
awesome! Looking forward for your updates. If you need any help with something that we can do feel free to ask

marmarek:
Here is updated os-autoinst integration. Setting up RPi4 with all the required parts is a bit of a challenge, example issues:
 - custom kernel build is needed to have `tc358743` driver
 - `/boot/efi/config.txt` needs `gpu_mem=128` and `start_x=1` - they _must_ be in `config.txt` specifically, not `extraconfig.txt`, as the very early boot code does not support `include`. `config.txt` is forcefully overridden on `raspberrypi-firmware-config` package update
 - `start4x.elf` and `fixup4x.dat` (part of RPi firmware) is not included in OpenSUSE package yet
 - finding appropriate EDID (with refresh rage below 60Hz) - and finding the issue is about refresh rate in the first place...

The resolution/aspect ratio issue I "solved" by setting 1024x768 on the SUT. I don't like it, as its yet another deviation of the test setup (that can hide some issue) from the actual product, but lets start with something.

Besides this, I have a bunch of supporting infrastructure, including:
 - everything needed for network boot (tftp server, DHCP, a bit tricky grub2-efi config etc)
 - remotely controlled power strip

I can share most of the scripts used there, but putting it all together is also a bit of a challenge.

consoles/video_stream.pm
```diff
@@ -0,0 +1,395 @@
+# Copyright © 2021 Marek Marczykowski-Górecki
+# Copyright © 2021 SUSE LLC
```
okurz, line 2:
```suggestion
# Copyright 2021 Marek Marczykowski-Górecki
# Copyright 2021 SUSE LLC
```

as in https://github.com/os-autoinst/os-autoinst/blob/master/consoles/console.pm#L2

consoles/video_stream.pm
```diff
@@ -0,0 +1,395 @@
...
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
```
okurz, line 15:
```suggestion
# SPDX-License-Identifier: GPL-2.0-or-later
```

as in https://github.com/os-autoinst/os-autoinst/blob/master/consoles/console.pm#L3

consoles/video_stream.pm
```diff
@@ -0,0 +1,395 @@
...
+
+package consoles::video_stream;
+
+use strict;
+use warnings;
```
okurz, line 20:
```suggestion
use Mojo::Base -strict, -signatures;
```

okurz (changes requested):
Looks nice in general.

I provided some minor remarks regarding copyright and license. I could provide many more suggestions regarding the code style that I prefer to have addressed before we can merge this PR but I leave it up to you if you would like to do such changes yourself or if for example I should update your branch myself with commits on top. Up to you.

@ggardet @coolo as recent authors in backend::generalhw your feedback would be appreciated regarding these changes

marmarek:
Thanks. I have some further changes, including support for storing assets. And few fixes for few bugs found along the way. Should I file separate PR for those, or is it fine to include here (in separate commits ofc)?

ggardet:
> Here is updated os-autoinst integration. Setting up RPi4 with all the required parts is a bit of a challenge, example issues:
> 
> * custom kernel build is needed to have `tc358743` driver

openSUSE Tumbleweed support it since kernel 5.8: https://github.com/openSUSE/kernel-source/blob/master/config/arm64/default#L6495


> * `/boot/efi/config.txt` needs `gpu_mem=128` and `start_x=1` - they _must_ be in `config.txt` specifically, not `extraconfig.txt`, as the very early boot code does not support `include`. `config.txt` is forcefully overridden on `raspberrypi-firmware-config` package update

This is part of https://bugzilla.opensuse.org/show_bug.cgi?id=1192047

> * `start4x.elf` and `fixup4x.dat` (part of RPi firmware) is not included in OpenSUSE package yet

This is part of `raspberrypi-firmware-extra-pi4` package.

> * finding appropriate EDID (with refresh rage below 60Hz) - and finding the issue is about refresh rate in the first place...
> 
> The resolution/aspect ratio issue I "solved" by setting 1024x768 on the SUT. I don't like it, as its yet another deviation of the test setup (that can hide some issue) from the actual product, but lets start with something.

At some point we may need to switch to today's resolution, such as 1920x1080.

> 
> Besides this, I have a bunch of supporting infrastructure, including:
> 
> * everything needed for network boot (tftp server, DHCP, a bit tricky grub2-efi config etc)

You could also use a boot from SD or USB.

> * remotely controlled power strip

I guess you saw the scripts in https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data/generalhw_scripts

> 
> I can share most of the scripts used there, but putting it all together is also a bit of a challenge.

Adding your scripts to https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data/generalhw_scripts is likely the right place.

ggardet:
Could you give us an example how to use `GENERAL_HW_INPUT_CMD` and `GENERAL_HW_INPUT_ARGS` please?

ggardet:
> Thanks. I have some further changes, including support for storing assets. And few fixes for few bugs found along the way. Should I file separate PR for those, or is it fine to include here (in separate commits ofc)?

New features could be a separate PR, but please include bug fixes here (except if they are completely unrelated).

marmarek:
> openSUSE Tumbleweed support it since kernel 5.8: https://github.com/openSUSE/kernel-source/blob/master/config/arm64/default#L6495

Yes, indeed the module is there, but for some reason it doesn't work. I ended up building a kernel from Raspberry Foundation's sources. Documentation mentions two video APIs and migrating/upstreaming them, so maybe that's part that works only in one of them. I could be missing some step, but I gave up looking how to get it working with default kernel, since I've got some configuration working already.

> You could also use a boot from SD or USB.

In fact, on some machines I do. But I load iPXE there, to have remotely-controlled boot order control. I'm preparing a writeup about this all, there will be more details :)

> I guess you saw the scripts in https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data/generalhw_scripts

In fact, I did not! Thanks!

> Could you give us an example how to use `GENERAL_HW_INPUT_CMD` and `GENERAL_HW_INPUT_ARGS` please?

See gist linked in the PR description. I'll add it to the generalhw_scripts dir.

marmarek:
> I'll add it to the generalhw_scripts dir.

Yhm, no, that's another repo. Maybe https://github.com/QubesOS/openqa-tests-qubesos then :)

marmarek:
This is getting into more or less the final shape. Tomorrow I'll update tests.

I'd appreciate comments especially about https://github.com/os-autoinst/os-autoinst/pull/1741/commits/8dc89fdf38e6179aaa9477bde8637d3cc7fa2fdf - while it does solve the issue, I'm not entirely sure if that's the right approach, or whether I got the I/O handled correctly there.

ggardet:
> > openSUSE Tumbleweed support it since kernel 5.8: https://github.com/openSUSE/kernel-source/blob/master/config/arm64/default#L6495
> 
> Yes, indeed the module is there, but for some reason it doesn't work. I ended up building a kernel from Raspberry Foundation's sources. Documentation mentions two video APIs and migrating/upstreaming them, so maybe that's part that works only in one of them. I could be missing some step, but I gave up looking how to get it working with default kernel, since I've got some configuration working already.

Could you create a bug report on https://bugzilla.opensuse.org, please?

> > Could you give us an example how to use `GENERAL_HW_INPUT_CMD` and `GENERAL_HW_INPUT_ARGS` please?
> 
> See gist linked in the PR description. I'll add it to the generalhw_scripts dir.

I saw the link, but I am not sure how I am supposed to use it within openQA, or even just test it manually.
BTW, what Raspberry Pi version and setup are you using for the USB emulation?

backend/generalhw.pm
```diff
@@ -27,39 +28,43 @@ sub get_cmd ($self, $cmd) {
...
+
+    my ($stdout, $stderr, $ret);
+    eval {
+        my ($path, @args) = @full_cmd;
+        my $p = process(execute => $path, args => \@args)->start;
+        $p->to_ioloop->on(read => sub {
+                $stdout .= $p->read_all_stdout;
+                $stderr .= $p->read_all_stderr;
+        });
+        Mojo::IOLoop->singleton->start() unless Mojo::IOLoop->singleton->is_running;
```
Martchus, line 59: `start` will block until `stop` is called which doesn't seem to be the case here. So this code will likely block here infinitely (unless the loop is already running anyways). I would have to check the Mojo::IOLoop::ReadWriteProcess documentation (and possibly implementation details) myself to tell you what the right approach would be, tough.

Martchus (changes requested):

backend/generalhw.pm
marmarek: > I would have to check the Mojo::IOLoop::ReadWriteProcess documentation (and possibly implementation details) myself to tell you what the right approach would be, tough.

I did. The documentation is not much helping (just one sentence for most methods, and no example how to "run a program and collect all its output" - just some examples for interactive things, like read one line, write one line etc). Inspecting implementation got me to this version, since `read_all_stdout` isn't really reading all stdout - it reads all stdout that the process produced so far (and exit early if nothing new is available).

I'm mostly worried about the possible deadlock case, where the process produces a lot of output on stdout _and_ stderr, filling either of the pipes, and the other end is waiting reading from the other pipe...

> So this code will likely block here infinitely (unless the loop is already running anyways).

Can I assume (and assert for, maybe) that the loop is already running? It isn't clear to me from code inspection... But given this version works, I guess it is the case.

Martchus: The documentation (and implementation) both are not very helpful, indeed.

> Can I assume (and assert for, maybe) that the loop is already running? It isn't clear to me from code inspection... But given this version works, I guess it is the case.

I don't think the Mojo IO loop is running in the backend process (which in fact runs its own "capture loop" using `IO::Select`). You could try to read from the `IO::Handle` (`$process->read_stream`) directly. Or you stop the `Mojo::IOLoop` in some event handler (e.g. the process' `stop` event).

Martchus: > Otherwise sometimes Mojo's loop collects the exit code of a process started by IPC::Run::run, which results in a false failure.

Note that it is not Mojo's loop which does that. It is the handler we register for the `Mojo::IOLoop::ReadWriteProcess::Session` which uses Mojo's event system but not the event loop (at least that's how it looks like after a brief look; the event is emitted from the `SIG_CHLD` handler).

I'm wondering whether switching to RWP makes any difference in that regard.

I'm also wondering whether the failure is really caused by the `SIG_CHLD` handler. All we do in that handler is logging.

marmarek: What I really want to achieve here, is to avoid `Mojo` collecting exit code _elsewhere_ of the for child process started here. What happens is that `session->on( collected_orphan => ...)` in `backend/driver.pm` is collecting it. If `Mojo::IOLoop` isn't normally running there, maybe I could prevent that thing from happening either? I guess it's registering SIGCHLD handler - maybe I could temporarily disable it (`$SIG[CHLD] = "DEFAULT"`?) instead of switching to `Mojo::IOLoop::ReadWriteProcess`? That would still risk some weird interactions (`IPC::Run` collecting an exit code that was supposed to be collected by `Mojo::IOLoop`), but maybe that's less of an issue?

What would you prefer:
1. temporarily starting (and stopping) `Mojo::IOLoop`, or
2. trying to make it work with `IPC:Run`

?

marmarek: > All we do in that handler is logging.

Since the message does include the PID and exit status, the exit status was retrieved, so `IPC::Run` can't do it anymore.

marmarek: > I'm wondering whether switching to RWP makes any difference in that regard.

I guess it does make a difference, because RWP can coordinate with the signal handler registered by another part of `Mojo::IOLoop`.

marmarek:
>  I'm preparing a writeup about this all, there will be more details :)

A draft is here already: https://github.com/QubesOS/qubes-posts/pull/91/files

backend/generalhw.pm
Martchus: Then it'll likely be the easiest to start/stop the event loop as it would indeed avoid unwanted interactions between the process management libraries.

backend/generalhw.pm
```diff
@@ -27,39 +28,46 @@ sub get_cmd ($self, $cmd) {
...
+        }
+        $ret = $p->wait_stop->exit_status;
+    };
+    die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
     chomp $stdout;
     chomp $stderr;
 
-    die "$args[0]: $stderr" unless $ret;
-    bmwqemu::diag("IPMI: $stdout");
+    die "$cmd(ret $ret): stdout: $stderr, stderr: $stderr" if $ret;
```
ggardet, line 69: I think it should be:
```
die "$cmd(ret $ret): stdout: $stdout, stderr: $stderr" if $ret;
```

marmarek:
I get _non-deterministically_ this:
```
GENERAL_HW_POWEROFF_CMD(ret 72057594037927935): stdout: , stderr:  at /home/abuild/rpmbuild/BUILD/os-autoinst-4.6.1637172504.0501ede8/backend/generalhw.pm line 69.
```
Both in CI (you can see once it passed and once failed) and locally. Any idea?

marmarek:
> Could you create a bug report on https://bugzilla.opensuse.org, please?

https://bugzilla.opensuse.org/show_bug.cgi?id=1192047

ggardet:
> > Could you create a bug report on https://bugzilla.opensuse.org, please?
> 
> https://bugzilla.opensuse.org/show_bug.cgi?id=1192047

Seems to be https://bugzilla.suse.com/show_bug.cgi?id=1192809 instead

Martchus:
> Both in CI (you can see once it passed and once failed) and locally. Any idea?

Maybe the IO loop is aborted before all events could be processed.  Maybe @kraih has an idea.

I'm still wondering about the interference of RWP's session and `SIGCHLD` handling with `IPC::Run` which doesn't register a `SIGCHLD` handler but only relies on `waitpid`. So `Mojo-IOLoop-ReadWriteProcess/lib/Mojo/IOLoop/ReadWriteProcess/Session.pm` screws even `waitpid`?

marmarek:
I also wonder why I hit it in this case only. One potential difference is the process I started takes rather long time (minutes), maybe that matters?
In theory, if `SIGCHLD` handler (from RWP's session) does `waitpid` on its own, it will definitely interfere with `waitpid` called elsewhere. But I have no idea why this issue affects some cases but not the others.

Maybe a better solution is to prevent RWP's session to register `SIGCHLD` handler in the first place (likely by not using the session at all)? But I'm not sure about the implications (for example whether it wouldn't break RWP itself, or result in zombies of  "unknown" children).

marmarek:
Hmm, maybe set [`RWP::session::collect_status`](https://metacpan.org/pod/Mojo::IOLoop::ReadWriteProcess::Session#collect_status) to 0?

Martchus:
The session's `SIGCHILD` handler calls `waitpid`, indeed. *Maybe* that's what is interfering (depending on how the Perl runtime is scheduling signal handlers). Considering RPW's own `waitpid` calls always use `session->_protect` that's likely the case.

I'm also not sure about the implications of disabling RWP's session. I wasn't involved in the introduction of RWP. I suppose one could use `session->collect_status(0);` to make its `SIGCHLD` handler temporarily a no-op and re-enable it again later (``session->collect_status(1);`).

Note that regardless of the way the process is spawned it is always problematic to block too long in the backend because the main capture loop should generally stay responsive.

marmarek:
> I'm also not sure about the implications of disabling RWP's session. I wasn't involved in the introduction of RWP. I suppose one could use `session->collect_status(0);` to make its `SIGCHLD` handler temporarily a no-op and re-enable it again later (``session->collect_status(1);`).

Looking at the session doc, another option is querying it to see if the exit status got collected there (`->contains($pid)`).

> Note that regardless of the way the process is spawned it is always problematic to block too long in the backend because the main capture loop should generally stay responsive.

Hmm, should I then make handling of those callback processes async? In this particular case, I don't care that much - they are running either before starting "the VM" or after stopping it, so there isn't really anything to capture.

Martchus:
> Hmm, should I then make handling of those callback processes async? In this particular case, I don't care that much - they are running either before starting "the VM" or after stopping it, so there isn't really anything to capture.

Then blocking should be fine.

backend/generalhw.pm
```diff
@@ -139,4 +182,14 @@ sub stop_serial_grab ($self, @) {
 
 # serial grab end
 
+sub do_extract_assets {
+    my ($self, $args) = @_;
```
kraih, line 186: This method should probably use a signature.

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
+# Copyright 2021 Marek Marczykowski-Górecki
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package consoles::video_stream;
+
+use Mojo::Base -strict, -signatures;
+use feature 'say';
```
kraih, line 8: This line does nothing, `Mojo::Base` already imports `say`.

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
...
+
+# speed limit: 30 keys per second
+use constant STREAM_TYPING_LIMIT_DEFAULT => 30;
+
+use constant DV_TIMINGS_CHECK_INTERVAL => 3;
+
+use constant STALL_THRESHOLD => 4;
+
+sub screen {
+    my ($self) = @_;
```
kraih, line 27: You activate `-signatures` at the top, so you should probably use signatures in the methods. 😁

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
...
+            return;
+        }
+        $self->{dv_timings} = _v4l2_ctl($url, '--get-dv-timings');
+    }
+
+    my @cmd = ('/usr/bin/ffmpeg', '-loglevel', 'fatal', '-i', $url);
+    push(@cmd, ('-vcodec', 'ppm', '-f', 'rawvideo', '-r', '2', '-'));
+
+    my $ffmpeg;
+    $self->{ffmpegpid} = open($ffmpeg, '-|', @cmd);
```
kraih, line 128: You're not checking if `open` fails.

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
...
+
+sub connect_remote_input {
+    my ($self, $cmd) = @_;
+
+    $self->{mouse} = {x => -1, y => -1};
+
+    diag "Connecting input device";
+
+    my $input_pipe;
+    $self->{inputpid} = open($input_pipe, '|' . $cmd);
```
kraih, line 145: You're not checking if `open` fails.

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
...
+    my ($self) = @_;
+
+    my $ffmpeg = $self->{ffmpeg};
+    $ffmpeg or die 'ffmpeg does not exist. Probably your backend instance could not start or died.';
+    $ffmpeg->blocking(0);
+    my $ret = $ffmpeg->read(my $header, 20);
+    $ffmpeg->blocking(1);
+
+    if (!$ret) {
+        return;
```
kraih, line 163: Better use an explicit return value, even if it is `undef`;

consoles/video_stream.pm
```diff
@@ -0,0 +1,380 @@
...
+    $self->_send_key_event($args->{key});
+    $self->backend->run_capture_loop(.2);
+    return {};
+}
+
+sub _send_key_event {
+    my ($self, $key) = @_;
+
+    return unless $self->{input_pipe};
+    $self->{input_pipe}->write($key . "\n");
```
kraih, line 294: You probably need exception handling for every `write` call.

backend/generalhw.pm
```diff
@@ -27,39 +28,46 @@ sub get_cmd ($self, $cmd) {
...
+
+    my ($stdout, $stderr, $ret);
+    eval {
+        my ($path, @args) = @full_cmd;
+        my $p = process(execute => $path, args => \@args)->start;
+        $p->to_ioloop->on(read => sub {
+                $stdout .= $p->read_all_stdout;
+                $stderr .= $p->read_all_stderr;
+        });
+        unless (Mojo::IOLoop->singleton->is_running) {
```
kraih, line 59: Looking at the surrounding code it would appear that this is never false. Which would mean that you don't really need RWP and could get away with the blocking `IPC::Run`, which is much easier to use.

backend/generalhw.pm
```diff
@@ -27,39 +28,46 @@ sub get_cmd ($self, $cmd) {
...
+            $p->on(collected => sub { Mojo::IOLoop->singleton->stop() });
+            Mojo::IOLoop->singleton->start();
```
kraih, line 61: If it turns out you do really need a solution for the running inside and outside of an existing event loop problem i can give you a more idiomatic promise based solution. But i don't think you need it.

backend/generalhw.pm
marmarek: See the discussion above, about `SIGCHLD` handler mostly. Anyway, I think there is indeed a way to keep using `IPC::Run`, I'll try that.

backend/generalhw.pm
kraih: You can always temporarily override the signal handler in the `IPC::Run` block, it's blocking after all. Just `local $SIG{CHLD} = ...` or similar at the start of it.

backend/generalhw.pm
```diff
@@ -27,39 +28,38 @@ sub get_cmd ($self, $cmd) {
...
+    push @full_cmd, @extra_args;
 
     my ($stdin, $stdout, $stderr, $ret);
+
+    # Do not let the SIGCHLD handler of Mojo::IOLoop steal the exit code from IPC::Run
+    session->collect_status(0);
     eval { $ret = IPC::Run::run([@full_cmd], \$stdin, \$stdout, \$stderr) };
-    die "Unable to run command '@full_cmd' (deduced from test variable $args[0]): $@\n" if $@;
+    die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
+    session->collect_status(1);
```
Martchus, line 57: This wouldn't be restored in the error case. Maybe it makes sense to use https://docs.mojolicious.org/Mojo/Util#scope_guard here.

marmarek:
Ok, I think I addressed all the earlier comments, and also it is feature complete. Ready for review.

backend/generalhw.pm
```diff
@@ -27,39 +29,40 @@ sub get_cmd ($self, $cmd) {
...
+sub run_cmd ($self, $cmd, @extra_args) {
+    my @full_cmd = split / /, $self->get_cmd($cmd);
+
+    push @full_cmd, @extra_args;
 
     my ($stdin, $stdout, $stderr, $ret);
-    eval { $ret = IPC::Run::run([@full_cmd], \$stdin, \$stdout, \$stderr) };
-    die "Unable to run command '@full_cmd' (deduced from test variable $args[0]): $@\n" if $@;
+
+    # Do not let the SIGCHLD handler of Mojo::IOLoop steal the exit code from IPC::Run
```
kraih, line 54: `Mojo::IOLoop` does not actually have a `SIGCHLD` handler, it is only an RWP feature.

backend/generalhw.pm
```diff
@@ -27,39 +29,40 @@ sub get_cmd ($self, $cmd) {
...
+    session->collect_status(0);
+    {
+        my $guard = scope_guard sub { session->collect_status(1); };
```
kraih, line 57: Good use of a scope guard. I would probably just replace the two calls of `collect_status` with a single `local $SIG{CHLD};` here though.

marmarek: I'm a bit worried of consequences of RWP-started processes, if any exits at this very time. Wouldn't that be a problem? If not, then sure, I can change the SIGCHLD handler.

marmarek: Would you like me to change it to `local $SIG{CHLD}`?

kraih: Both should be fine here. But you should change the comment, since `Mojo::IOLoop` doesn't have a `SIGCHLD` handler (it is RWP only).

Martchus: It is true that if an RWP-started process exists at this very time it will fail to emit the collected event for that process. However, we only lose a log message so I suppose we can live with that. And yes, this is RWP-specific (and nothing `Mojo::IOLoop` does in general).

marmarek:
OBS build says: "TBB Warning: The number of workers is currently limited to 3. The request for 7 workers is ignored. Further requests for more workers will be silently ignored until the limit changes." and later is killed on 1.5h timeout. Is it something for me to look into, or unrelated CI issue?

okurz:
> OBS build says: "TBB Warning: The number of workers is currently limited to 3. The request for 7 workers is ignored. Further requests for more workers will be silently ignored until the limit changes." and later is killed on 1.5h timeout. Is it something for me to look into, or unrelated CI issue?

That CI problem recently appeared. Gladly @Martchus could already fix it in https://github.com/os-autoinst/os-autoinst/pull/1860 so when you rebase your changes the check should run fine

consoles/video_stream.pm
```diff
@@ -0,0 +1,354 @@
...
+
+    for my $letter (split("", $args->{text})) {
+        next if ($letter eq "\r");
+        my $charmap = {
+            "\t" => 'tab',
+            "\n" => 'ret',
+            "\b" => 'backspace',
+            "\e" => 'esc',
+            " " => 'space',
+        };
```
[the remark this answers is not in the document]
kraih: Yes, it would be more efficient to define this at the outermost scope. I would put it like this at the top of the file:
```perl
my $CHARMAP = {
            "\t" => 'tab',
            "\n" => 'ret',
            "\b" => 'backspace',
            "\e" => 'esc',
            " " => 'space',
        };
```

marmarek:
```
[  216s] 3: #   Failed test 'log output for backend driver creation'
[  216s] 3: #   at ./29-backend-driver.t line 24.
[  216s] 3: # STDOUT & STDERR:
[  216s] 3: # hannel_in 20
[  216s] 3: # [0m
[  216s] 3: # don't match:
[  216s] 3: # (?^u:(Blocking SIGCHLD|channel_out))
[  216s] 3: # as expected
[  217s] 3: # Looks like you failed 1 test of 12.
[  217s] 3: [15:15:44] ./29-backend-driver.t .................... 
[  217s] 3: Dubious, test returned 1 (wstat 256, 0x100)
[  217s] 3: Failed 1/12 subtests 
```
Doesn't look related to my changes, right?

marmarek:
I think I've addressed all requests, anything else is needed here?

okurz:
There are warnings about newly introduced code that is not yet covered by tests, e.g. in https://github.com/os-autoinst/os-autoinst/pull/1741/files#diff-4e823d39debb914f0d3c6566265ac0542327bdb48548f4bed485c20dd831b48aR104

As your PR has grown quite a bit maybe we can merge some commits in a separate PR first to get them "ouf of the way" already :)

ggardet:
I am testing this PR with `generalhw` backend to check there is no regression. Let's see how it goes: https://openqa.opensuse.org/tests/2094040

okurz (approved):
ok, I checked the changed code again as well as test results and I can now approve also in the current state. Good work!

marmarek:
> @marmarek This may be just stderr now shown?

Yes, that's [intentional change](https://github.com/os-autoinst/os-autoinst/pull/1741/commits/3a353791a34e26ee5eb0e41fa606ad729da3f29c) -  having just stdout _or_ stderr was rather painful for diagnosing issues with those scripts.

doc/backend_vars.asciidoc
```diff
@@ -272,6 +273,11 @@ GENERAL_HW_POWEROFF_CMD;string;;Shell Command to power off the SUT (in CMD_DIR)
 GENERAL_HW_POWEROFF_ARGS;string;;Arguments to pass GENERAL_HW_POWEROFF_CMD Shell script
 GENERAL_HW_FLASH_CMD;string;;Shell Command to flash a disk image on SUT (in CMD_DIR), optional
 GENERAL_HW_FLASH_ARGS;string;;Arguments to pass GENERAL_HW_FLASH_CMD Shell script
+GENERAL_HW_STORE_ASSET_CMD;string;;Shell Command to extract disk image from SUT (in CMD_DIR), optional
+GENERAL_HW_STORE_ASSET_ARGS;string;;Arguments to pass GENERAL_HW_STORE_ASSET_CMD Shell script. The script will get also extra arguments: disk number and file path to save it into.
```
kalikiana, line 277: Pun intended? 😉 The descriptions already refer to the disk image, so maybe `IMAGE` is a more descriptive term to use here and a single word fits in better with existing ones?
```suggestion
GENERAL_HW_IMAGE_ARGS;string;;Arguments to pass GENERAL_HW_IMAGE_CMD Shell script. The script will get also extra arguments: disk number and file path to save it into.
```

consoles/video_stream.pm
```diff
@@ -0,0 +1,353 @@
...
+    my $ret = $self->disable_video;
+    if ($self->{input_pipe}) {
+        close($self->{input_pipe});
+        waitpid($self->{inputpid}, 0);
+    }
+    return $ret;
+}
+
+sub _v4l2_ctl ($device, $cmd) {
+    my @cmd = ("v4l2-ctl", "-d", $device, "-k");
```
kalikiana, line 59: Maybe long options are a little clearer? At least I had to lookup "device" and  "koncise". Well, no strong opinion
```suggestion
    my @cmd = ("v4l2-ctl", "--device", $device, "--concise");
```

consoles/video_stream.pm
```diff
@@ -0,0 +1,353 @@
...
+        } else {
+            $self->{dv_timings_supported} = 0;
+            bmwqemu::diag "DV timings not supported";
+        }
+    } else {
+        # applies to v4l only
+        $self->{dv_timings_supported} = 0;
+    }
+
+    bmwqemu::diag "Starting receiving video stream at $args->{url}";
```
kalikiana, line 102:
```suggestion
    bmwqemu::diag "Starting to receive video stream at $args->{url}";
```

consoles/video_stream.pm
```diff
@@ -0,0 +1,353 @@
...
+    $self->{input_pipe} = $input_pipe;
+    $self->{input_pipe}->autoflush(1);
+
+    return $input_pipe;
+}
+
+
+sub _receive_frame ($self) {
+    my $ffmpeg = $self->{ffmpeg};
+    $ffmpeg or die 'ffmpeg does not exist. Probably your backend instance could not start or died.';
```
kalikiana, line 152: `exist` makes me think not installed, I assume as the message says this is about the process aborting

```suggestion
    $ffmpeg or die 'ffmpeg is not running. Probably your backend instance could not start or died.';
```

consoles/video_stream.pm
```diff
@@ -0,0 +1,353 @@
...
+            $self->connect_remote_video($self->{args}->{url});
+        }
+    }
+}
+
+sub type_string ($self, $args) {
+    my $seconds_per_keypress = 1 / (get_var('STREAM_TYPING_LIMIT', STREAM_TYPING_LIMIT_DEFAULT) || 1);
+
+    # further slow down if being asked for.
+    # 250 = magic default from testapi.pm
```
kalikiana, line 233: I'm guessing you copied this from vnc... maybe time to add a constant

consoles/video_stream.pm
```diff
@@ -0,0 +1,353 @@
...
+
+    $self->_mouse_move($x, $y);
+    return {absolute => $self->{mouse}};
+}
+
+sub mouse_set ($self, $args) {
+    die "Need x/y arguments" unless (defined $args->{x} && defined $args->{y});
+
+    # TODO: for framebuffers larger than 1024x768, we need to upscale
+    #$self->_mouse_move(int($self->{width}/1024 * $args->{x}), int($self->{height}/768 * $args->{y}));
```
kalikiana, line 327: I suggest to drop these comments and file a ticket or an issue instead, because in my experience they will stay untouched for a long time.

kalikiana (dismissed):

kalikiana (changes requested):
> Has been adressed

Almost. GitHub correctly identified the one outstanding comment 😉️

kalikiana (approved):
Thank you for taking my comments into account. This is looking really nice!

OpenQA/Isotovideo/Utils.pm
```diff
@@ -128,7 +128,6 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} eq 'qemu';
```
foursixnine, line 131: @marmarek fyi this line broke other backends, I wonder if this was something intended for your use case, or was lefover of some debugging...

perlpunk: We reverted the whole PR for now, btw: #1892

kalikiana: See also #1895 for a poc of implementing it in other backends - currently we don't require this and we don't test it

marmarek: It is intentional change, as this PR adds assets extracting to another backend too. I may be missing something, but it looks to call the actual backend method only when there is some asset to store (explicitly requested via settings). Do you mean some tests have `STORE_HDD_*` or `PUBLISH_HDD_*` set even though selected backend does not support it? Sounds like a problem with test settings...

perlpunk: The test linked in https://progress.opensuse.org/issues/104499 has `PUBLISH_HDD_1` set

marmarek: And a backend not supporting storing assets? (I can't access that test) How is that supposed to work?

kalikiana: @marmarek the test above uses https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm which doesn't implement `handle_generated_assets` - technically the test passes

marmarek: So, the test relied on `PUBLISH_HDD_1` being silently ignored by this backend. What about adding dummy `do_extract_assets` to the base backend class that warns about ignoring asset publish for given backend? That would avoid the regression, while also having some info what have happened.

kalikiana: You got me an idea. Instead of making the method mandatory we could return based on having assets: #1897

perlpunk: More findings why `PUBLISH_HDD_1` was set:
https://progress.opensuse.org/issues/104499#note-7

marmarek: That makes sense now. Is it ok to go with "The code above should be moved to backend::svirt, and after that we can hopefully remove the check for qemu." or should I adjust my changes to keep ignoring `PUBLISH_HDD_1` for non-supporting backends?

kalikiana: @marmarek I'd say rebase w/o the commit 32862bc5eb314ceddc250a50731e1ef04a564bb0 (after these two points have been addressed) and you shouldn't need any special handling

marmarek: But then, saving assets using `generalhw` backend won't work... I can modify it to `return unless $bmwqemu::vars{BACKEND} eq 'qemu' or $bmwqemu::vars{BACKEND} eq 'generalhw'` - which is ugly, but should be safe...

kalikiana: > But then, saving assets using generalhw backend won't work
I assume thanks to #1894 a test using that would fail with `backend method 'do_extract_assets' not implemented for class backend::generalhw` and that can be fixed in a separate branch?

perlpunk: I would be fine with testing for both backends for now.
But I vote against `unless` with more than one condition. It makes my brain turn upside down ;-)
```
return unless $bmwqemu::vars{BACKEND} =~ m/^(qemu|generalhw)$/;
```

PR #1746, opened 2021-08-24
 Rework add drives mechanism in svirt 

mloviska:
SLEM 5.1 builds bootable, ready to use images for z/KVM. To make sure
that images can be expanded we need to add additional functionality to
`add_disk` to create a backingfile of given size.
For `svirt` backend ( except of hyperv ), we should be able to create
drives in following modes.

1. Create empty drive
2. Create a backing image file of given size
3. Re-use given image
4. Add cd-rom drive with given iso image

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -16,11 +16,9 @@
 
 package consoles::sshVirtsh;
 
-use strict;
 use warnings;
+use Mojo::Base 'consoles::sshXtermVt', -strict, -signatures;
 use autodie ':all';
-
-use base 'consoles::sshXtermVt';
 
 require IPC::System::Simple;
 use XML::LibXML;
@@ -31,6 +29,7 @@ use Mojo::JSON qw(decode_json);
 
 use backend::svirt;
 use testapi qw(get_var get_required_var check_var set_var);
+use Number::Format qw(unformat_number);
 
 has instance   => (is => "rw", isa => "Num");
 has name       => (is => "rw", isa => "Str");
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
 
+sub to_byte {
+    my $sp              = shift;
+    my $size_multiplier = {
+        b => 1,
+        k => 1024, K => 1024,
+        M => 1024 * 1024,
+        G => 1024 * 1024 * 1024,
+        T => 1024 * 1024 * 1024 * 1024,
+    };
```
okurz, line 336: https://metacpan.org/pod/Number::Format might be more helpful here. There is also https://metacpan.org/pod/Number::Bytes::Human which is currently not in openSUSE:Factory but we could submit that

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
+    $dest .= basename($src);
+    die "Downloading $src failed!" if $self->run_cmd("rsync -av $src $dest");
+    return $dest;
+}
+
+sub decompress_cmd {
+    return sprintf(q(xz --decompress --keep --verbose --no-warn --quiet '%s'), shift);
+}
+
 sub add_disk {
```
okurz, line 401: The method is quite big and complex. Could you please split the method in a separate commit before changing it? If you are unsure how to do that I could try myself to suggest a change.

okurz (changes requested):

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
 }
 
+sub to_byte {
+    my $sp              = shift;
+    my $size_multiplier = {
+        b => 1,
+        k => 1024, K => 1024,
+        M => 1024 * 1024,
+        G => 1024 * 1024 * 1024,
+        T => 1024 * 1024 * 1024 * 1024,
```
Martchus, line 335: In Perl there's a operator for computing an exponentiation, e.g.:
```suggestion
        T => 1024 ** 3,
```
But `perl-Number-Format` is already in Factory so you might just depend on that.

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
     return;
 }
 
+sub to_byte {
+    my $sp              = shift;
```
Martchus, line 329: It would be nice to use signatures in new functions, e.g.:
```suggestion
sub to_byte ($sp) {
```

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
+    my $size_multiplier = {
+        b => 1,
+        k => 1024, K => 1024,
+        M => 1024 * 1024,
+        G => 1024 * 1024 * 1024,
+        T => 1024 * 1024 * 1024 * 1024,
+    };
+
+    #in case of invalid format return error
```
Martchus, line 338:
```suggestion
    # in case of invalid format return error
```

Martchus (changes requested):

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
+        return undef;
+    }
+
+    return $+{num} * $size_multiplier->{$+{unit} // 'b'};
+}
+
+sub gen_kvm_img {
+    my ($self, %args) = @_;
+    my $byte_size = to_byte($args{size});
+    die('Missing argument size or invalid format') unless (defined($byte_size));
```
Martchus, line 349: I would generally prefer to use less parentheses like this:
```suggestion
    die 'Missing argument size or invalid format' unless defined $byte_size;
```

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
+    for (1 .. $tries) {
+        ($ret, $stdout, $stderr) = $self->run_cmd(join(" ", @cmd), wantarray => 1);
+        if ($stderr) {
+            bmwqemu::diag("gen_kvm_img failed: $stderr, try $_ out of $tries");
+            sleep 5;
+        }
+        last unless $ret;
+    }
+    die "Too many attempts to format HDD" if $stderr;
+    bmwqemu::diag("gen_kvm_img: Finished creating image file...");
```
Martchus, line 378: If no interpolation is required I'd prefer to use single quotes.

consoles/sshVirtsh.pm
```diff
@@ -325,111 +325,169 @@ sub add_interface {
...
+        my $basedir     = "/vmfs/volumes/$vmware_datastore/openQA/";
+        my $nfs_basedir = '/vmfs/volumes/' . get_required_var('VMWARE_NFS_DATASTORE');
+        $file = sprintf("%s%s.vmdk", $self->{name}, $args->{dev_id});
+        my $vmware_disk_path = $basedir . $file;
+
+        # Power VM off, delete it's disk image, and create it again.
+        # Than wait for some time for the VM to *really* turn off.
+        my $cmd = <<"EOT";
+( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk '/$self->{name}/ { print \$1 }');
+if [ \$vmid ]; then
```
Martchus, line 423: Is it actually required to escape the `$` here?

mloviska: Yes, that should be treated as a shell variable.

consoles/sshVirtsh.pm
mloviska: Yeah, sure. But wasn't there a reason behind to avoid signatures in Perl?

Martchus: Compatibility - but at this point we decided to drop support for too old Perl versions.

consoles/sshVirtsh.pm
Martchus: You're right, apparently interpolation is happening here. (`<<'EOT'` would avoid it but it looks like you actually need it here.)

consoles/sshVirtsh.pm
mloviska: > https://metacpan.org/pod/Number::Format might be more helpful here. There is also https://metacpan.org/pod/Number::Bytes::Human which is currently not in openSUSE:Factory but we could submit that

`Number::Format` seems to work nicely, thanks!

consoles/sshVirtsh.pm
okurz: yes, years ago. But to my understanding signatures are fine as long as we avoid some more exotic features.

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            push @last_ssh_args,     [%args];
+
+            my $ret = shift @ssh_cmd_return;
+            return undef unless defined $ret;
+
+            if (ref($ret) eq 'ARRAY') {
+                return @$ret;
+            } else {
+                return $ret;
+            }
```
okurz, line 258:
```suggestion
            return ref($ret) eq 'ARRAY' ? @$ret : $ret;
```

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            is(defined($source_node),              1,                    '<source> node exists');
+            is($source_node->getAttribute('file'), $basedir . $exp_file, 'file attribute of <source> is correct');
+
+            is(defined($driver_node),               1,        '<driver> node exists');
+            is($driver_node->getAttribute('name'),  'qemu',   'name attribute of <driver> is correct');
+            is($driver_node->getAttribute('type'),  'qcow2',  'type attribute of <driver> is correct');
+            is($driver_node->getAttribute('cache'), 'unsafe', 'cache attribute of <driver> is correct');
+        };
+
+
```
okurz, line 466: delete one blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+        subtest 'family svirt-xen-hvm backingfile=1' => sub {
+
```
okurz, line 468: delete blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+                    backingfile => 1,
+                    dev_id      => $dev_id,
+                    file        => '/my/path/to/this/file/' . $file,
+                    size        => 12
+            });
+            like($last_ssh_commands[0], qr%^rsync.*/my/path/to/this/file/$file $basedir$file%, 'Use rsync to copy file');
+            is($last_ssh_commands[-1], "qemu-img create -f qcow2 -b $basedir$file ${basedir}openQA-SUT-1$dev_id.img 12G", 'Used image size > backingfile size');
+        };
+
+
```
okurz, line 484: delete one blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            $svirt->add_disk({
+                    backingfile => 1,
+                    dev_id      => $dev_id,
+                    file        => '/my/path/to/this/file/' . $file,
+                    size        => 5
+            });
+            like($last_ssh_commands[0], qr%^rsync.*/my/path/to/this/file/$file $basedir$file%, 'Use rsync to copy file');
+            is($last_ssh_commands[-1], "qemu-img create -f qcow2 -b $basedir$file ${basedir}openQA-SUT-1$dev_id.img $_10gb", 'Used image size <= backingfile size');
+
+
```
okurz, line 500: delete one blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            is($source_node->getAttribute('file'), $basedir . "openQA-SUT-1$dev_id.img", 'file attribute of <source> is correct');
+
+            is(defined($driver_node),               1,        '<driver> node exists');
+            is($driver_node->getAttribute('name'),  'qemu',   'name attribute of <driver> is correct');
+            is($driver_node->getAttribute('type'),  'qcow2',  'type attribute of <driver> is correct');
+            is($driver_node->getAttribute('cache'), 'unsafe', 'cache attribute of <driver> is correct');
+
+            throws_ok { $svirt->add_disk({backingfile => 1, dev_id => $dev_id}) } qr/file/, "Die on missing file argument";
+        };
+        subtest 'family svirt-xen-hvm cdrom=1' => sub {
```
okurz, line 518: add one blank line for separation

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+
+            is(defined($source_node),              1,                '<source> node exists');
+            is($source_node->getAttribute('file'), $basedir . $file, 'file attribute of <source> is correct');
+
+            is(defined($driver_node),               1,      '<driver> node exists');
+            is($driver_node->getAttribute('name'),  'qemu', 'name attribute of <driver> is correct');
+            is($driver_node->getAttribute('type'),  'raw',  'type attribute of <driver> is correct');
+            is($driver_node->getAttribute('cache'), undef,  'cache attribute of <driver> is not defined');
+
+
```
okurz, line 545: delete one blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            throws_ok { $svirt->add_disk({cdrom => 1, dev_id => $dev_id}) } qr/file/, "Die on missing file argument";
+        };
+    };
+    subtest 'family kvm' => sub {
+
```
okurz, line 550: delete one blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+            }
+
+            my $dev_id   = 'dev_id_014_NO_SIZE';
+            my $exp_file = $svirt->name . $dev_id . ".img";
+            $svirt->add_disk({create => 1, dev_id => $dev_id});
+            is($last_ssh_commands[-1], "qemu-img create -f qcow2 $basedir$exp_file 20G", "Default size is 20G");
+        };
+
+        subtest 'family svirt-xen-hvm backingfile=1' => sub {
+
```
okurz, line 598: delete blank line

t/22-svirt.t
```diff
@@ -234,4 +235,479 @@ subtest 'Method backend::svirt::open_serial_console_via_ssh()' => sub {
...
+                    cdrom     => 1,
+                    dev_id    => $dev_id,
+                    file      => '/my/path/to/this/file/' . $file,
+                    bootorder => 'MyArbitraryBootOrderArgument'
+            });
+            my $target_nodelist = $svirt->{domainxml}->findnodes('domain/devices/disk[@type="file" and @device="cdrom"]/boot[@order="MyArbitraryBootOrderArgument"]');
+            is($target_nodelist->size(), 1, 'Boot order entry was created');
+        }
+    };
+
```
okurz, line 711: delete blank line

t/22-svirt.t
mloviska: I find more readable the multiline part. Sorry, I would prefer to keep it that way. The same applies above.

consoles/sshVirtsh.pm
mloviska: Yes, please, it would be nice if you can try to split the original method. Thanks

consoles/sshVirtsh.pm
mloviska: Maybe I am doing something wrong, but is if enable signatures with `use feature 'signatures'` I get warnings the feature is experimental

Martchus: I would recommend enabling them via `use Mojo::Base -strict, -signatures;` (or just add `, -signatures` to the `use` args where `Mojo::Base` is already used).

mloviska: > I would recommend enabling them via `use Mojo::Base -strict, -signatures;` (or just add `, -signatures` to the `use` args where `Mojo::Base` is already used).

Thanks, this was the magic. Good to know how to enable signatures in Perl. :)

consoles/sshVirtsh.pm
okurz: I did so in https://github.com/os-autoinst/os-autoinst/pull/1752 . I am sure this will cause multiple conflicts for you and I have only tested it with t/22-svirt.t so feel free to incorporate that in anyways helpful to you and please test accordingly.

okurz:
now please rebase

PR #1750, opened 2021-08-24
Allow more modern perl style with consistent "use Mojo::Base -strict"

okurz:
Used

```
sed -i "s/^use strict;$/use Mojo::Base -strict;/" $(git ls-files)
```

to convert all files.

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -338,6 +338,7 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
     loadtest 'start';
 };
 
+pass autotest::loadtest('tests/test.py'), 'can load python test module at all';
 loadtest 'test.py', 'we can also parse python test modules';
 
 done_testing();
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -1,7 +1,6 @@
 package OpenQA::Benchmark::Stopwatch;
 
-use strict;
-use warnings;
+use Mojo::Base -strict;
 
 our $VERSION = '0.05';
 use Time::HiRes;
--- a/OpenQA/Exceptions.pm
+++ b/OpenQA/Exceptions.pm
@@ -15,8 +15,7 @@
 
 package OpenQA::Exceptions;
 
-use strict;
-use warnings;
+use Mojo::Base -strict;
 
 use Exception::Class (
     'OpenQA::Exception::InternalException' => {
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -15,8 +15,7 @@
 
 package OpenQA::Isotovideo::Interface;
 
-use strict;
-use warnings;
+use Mojo::Base -strict;
 
 # version of the test API and the API relevant to the worker
 # -> increment on every change of such APIs
--- a/backend/amt.pm
+++ b/backend/amt.pm
@@ -15,8 +15,7 @@
 
 package backend::amt;
 
-use strict;
-use warnings;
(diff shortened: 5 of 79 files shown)


Martchus (approved):

okurz:
Failed in OBS with

```
[  275s] 3: error on tests/test.py: Error -- py_eval raised an exception at /usr/lib/perl5/vendor_perl/5.32.1/x86_64-linux-thread-multi/Inline/Python.pm line 177.
[  275s] 3: BEGIN failed--compilation aborted at (eval 236) line 5.
[  275s] 3: # Tests were run but no plan was declared and done_testing() was not seen.
[  275s] 3: # Looks like your test exited with 255 just after 89.
[  275s] 3: [18:40:53] ./08-autotest.t .......................... 
[  275s] 3: Dubious, test returned 255 (wstat 65280, 0xff00)
[  275s] 3: All 89 subtests passed 
```

I could identify the second commit `Allow more modern perl style with consistent "use Mojo::Base -strict"` as the culprit but don't know what python complains about in particular. I thought we had improved the error feedback from python but seems like we did not cover this part yet. Anyone has an idea how to debug?

Martchus:
The real error message is maybe printed before the exception is thrown but it is suppressed by prove. Maybe it helps to call it with `-v`.

okurz:
> Maybe it helps to call it with `-v`.

no, same. Also same for `perl -I. t/08-autotest.t`

Martchus:
I just had a quick look at the test and it uses `Test::Output`. Maybe the error message is hidden by that.

okurz:
Replacing the wrapper method "loadtest" which does `stderr_like` with a direct call to autotest::loadtest yields

```
SystemError: Objects/listobject.c:307: bad argument to internal function

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/okurz/local/os-autoinst/os-autoinst/testapi.py", line 21, in <module>
    for i in dir(perl.testapi):
SystemError: <built-in method __dir__ of _perl_pkg object at 0x7f5e53554648> returned a result with an error set
```

that's a bit better although I don't what this means :)

okurz:
Updated with one additional commit to have a better more explicit python test module test which shows the above error. And also changed the original commit to *not* use `Mojo::Base -strict` for now in autotest.pm which triggered the above problem. Let's see if other tests fail.

okurz:
currently https://build.opensuse.org/package/show/devel:openQA:TestGithub:OPR-1750/os-autoinst is "unresolvable" against openSUSE_Factory with "nothing provides libIlmImf-2_5.so.25()(64bit) needed by libopencv_imgcodecs4_5", likely unrelated.

t/08-autotest.t
```diff
@@ -338,6 +338,7 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
     loadtest 'start';
 };
 
+pass autotest::loadtest('tests/test.py'), 'can load python test module at all';
```
perlpunk, line 341: https://progress.opensuse.org/issues/99675
```
Useless use of a constant ("can load python test module at a"...) in void context at ./08-autotest.t line 342.
```

Martchus: Looks like it should have been `ok`.

perlpunk: Yeah, I tried that, but that fails for me locally and in CI

Martchus: Then it should possibly use `Test::Exception::lives_ok`.

okurz: `ok` also does not work because loadtest does not return anything. Fix in https://github.com/os-autoinst/os-autoinst/pull/1797

PR #1751, opened 2021-08-25
Add YAML file for commit message checks

BillAnastasiadis:
Related ticket: https://progress.opensuse.org/issues/95581
info: https://github.com/marketplace/actions/gs-commit-message-checker

The rules that are checked against are:
- commit subject must begin either with a capital letter (A-Z) or with a tag ([A-Za-z0-9_]:).
- commit subject must be no longer than 72 characters
- commit subject and commit body must be separated by at least one blank line
- each line in commit body must be no longer than 80 characters


Martchus (changes requested):
> each line in commit body must be no longer than 80 characters

We've already decided for openQA that this is too less and ended up dropping it completely: https://github.com/os-autoinst/openQA/pull/4125
It is just very counterproductive if you cannot describe details of your change.

okurz:
What do you think, should we consider putting these commit style checks into github.com/os-autoinst/os-autoinst-common/ ?

Martchus:
Good idea if it doesn't cause too much technical complication.

BillAnastasiadis:
> > each line in commit body must be no longer than 80 characters
> 
> We've already decided for openQA that this is too less and ended up dropping it completely: [os-autoinst/openQA#4125](https://github.com/os-autoinst/openQA/pull/4125)
> It is just very counterproductive if you cannot add logs/output/examples related to the change.

Missed that. Ok, removed from this one as well.

Martchus (approved):

PR #1758, opened 2021-08-27
tests: Verify that serialization works after cloning

kalikiana:
See: https://progress.opensuse.org/issues/97574


okurz:
nice, does it fail with https://github.com/os-autoinst/os-autoinst/pull/1757 reverted?

kalikiana:
> nice, does it fail with #1757 reverted?

Yes it does, I checked that locally.

okurz:
OBS checks fail due to "unresolvable: nothing provides libIlmImf-2_5.so.25()(64bit) needed by libopencv_imgcodecs4_5 " but only in the openSUSE_Factory repository. I have observed the same in another PR and I am convinced it's unrelated. Also I expect that this will be worked on.

PR #1759, opened 2021-08-31
Fix render calls in error handling of asset upload

Martchus:
It looks like using `message` does not work here:
```
Could not render a response at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Controller.pm line 151.
        Mojolicious::Controller::render(Mojolicious::Controller=HASH(0x1000e203040), "message", "File is too big", "status", 400) called at /usr/lib/os-autoinst/commands.pm line 163
```

`message` isn't documented¹ so I assume `text` is supposed to be used here.

---

¹ https://metacpan.org/pod/Mojolicious::Controller#render

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -165,7 +165,8 @@ sub upload_file {
 
     # choose 'target' field from curl form, otherwise default 'assets_private', assume the pool directory is the current working dir
     my $target = $self->param('target') || 'assets_private';
-    mkdir($target) or die "Unable to create directory for upload: $!" unless -d $target;
+    eval { mkdir $target unless -d $target };
+    if (my $error = $@) { return $self->render(text => "Unable to create directory for upload: $error", status => 500) }
 
     my $upname   = $self->param('upname');
     my $filename = basename($upname ? $upname : $self->param('filename'));
--- a/commands.pm
+++ b/commands.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2020 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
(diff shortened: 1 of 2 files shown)


okurz (commented):
ok, could be. A good opportunity to add tests :)

commands.pm
```diff
@@ -160,12 +160,13 @@ sub get_asset {
...
-    return $self->render(message => 'File is too big', status => 400) if $self->req->is_limit_exceeded;
-    return $self->render(message => 'Upload file content missing', status => 400) unless my $upload = $self->req->upload('upload');
+    return $self->render(text => 'File is too big', status => 400) if $self->req->is_limit_exceeded;
+    return $self->render(text => 'Upload file content missing', status => 400) unless my $upload = $self->req->upload('upload');
 
     # choose 'target' field from curl form, otherwise default 'assets_private', assume the pool directory is the current working dir
     my $target = $self->param('target') || 'assets_private';
-    mkdir($target) or die "Unable to create directory for upload: $!" unless -d $target;
+    eval { mkdir $target unless -d $target };
+    if (my $error = $@) { return $self->render(text => "Unable to create directory for upload: $error", status => 500) }
```
okurz, line 169: shouldn't autodie handle mkdir problems already?

Martchus: It is better to report the error to the user instead of letting the exception escape and just print some generic HTML: https://openqa.nue.suse.com/tests/6975596#step/svirt_upload_assets/8

okurz (approved):

Martchus:
Only the packaging check fails due to `nothing provides libIlmImf-2_5.so.25()(64bit) needed by libopencv_imgcodecs4_5` (know problem). So I'm merging manually.

PR #1760, opened 2021-08-31
Allow overriding Mojolicious limits for the file upload

Martchus:
There are test failures in production because the currently hard-coded
limit of 20 GiB is exceeded:
https://openqa.nue.suse.com/tests/6975596#step/svirt_upload_assets/8

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2020 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -236,9 +236,9 @@ sub get_temp_file {
 sub run_daemon {
     my ($port, $isotovideo) = @_;
 
-    # allow up to 20GB - hdd images
-    $ENV{MOJO_MAX_MESSAGE_SIZE}   = 1024 * 1024 * 1024 * 20;
-    $ENV{MOJO_INACTIVITY_TIMEOUT} = 300;
+    # allow up to 20 GiB for uploads of big hdd images
+    $ENV{MOJO_MAX_MESSAGE_SIZE}   //= ($bmwqemu::vars{UPLOAD_MAX_MESSAGE_SIZE_GB} // 20) * 1024**3;
+    $ENV{MOJO_INACTIVITY_TIMEOUT} //= ($bmwqemu::vars{UPLOAD_INACTIVITY_TIMEOUT}  // 300);
 
     # avoid leaking token
     app->mode('production');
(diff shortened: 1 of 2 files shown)


commands.pm
```diff
@@ -236,9 +236,9 @@ sub get_temp_file {
 sub run_daemon {
     my ($port, $isotovideo) = @_;
 
-    # allow up to 20GB - hdd images
-    $ENV{MOJO_MAX_MESSAGE_SIZE}   = 1024 * 1024 * 1024 * 20;
-    $ENV{MOJO_INACTIVITY_TIMEOUT} = 300;
+    # allow up to 20 GiB for uploads of big hdd images
+    $ENV{MOJO_MAX_MESSAGE_SIZE}   //= $bmwqemu::vars{FILE_UPLOAD_MAX_MESSAGE_SIZE}   // 1024 * 1024 * 1024 * 20;
+    $ENV{MOJO_INACTIVITY_TIMEOUT} //= $bmwqemu::vars{FILE_UPLOAD_INACTIVITY_TIMEOUT} // 300;
```
okurz, line 241: Please add in docs/backend_vars.asciidoc

commands.pm
```diff
@@ -236,9 +236,9 @@ sub get_temp_file {
 sub run_daemon {
     my ($port, $isotovideo) = @_;
 
-    # allow up to 20GB - hdd images
-    $ENV{MOJO_MAX_MESSAGE_SIZE}   = 1024 * 1024 * 1024 * 20;
-    $ENV{MOJO_INACTIVITY_TIMEOUT} = 300;
+    # allow up to 20 GiB for uploads of big hdd images
+    $ENV{MOJO_MAX_MESSAGE_SIZE}   //= $bmwqemu::vars{FILE_UPLOAD_MAX_MESSAGE_SIZE}   // 1024 * 1024 * 1024 * 20;
```
okurz, line 240: I guess the calculation `1024 * 1024 * 1024 * 20` can be simplified with the perl helper module we recently referenced in https://github.com/os-autoinst/os-autoinst/pull/1746.

Also I suggest to use a variable for the size in GB, e.g.

```suggestion
    $ENV{MOJO_MAX_MESSAGE_SIZE}   //= ($bmwqemu::vars{FILE_UPLOAD_MAX_MESSAGE_SIZE_GB} // 20) * 1024**3;
```

okurz (changes requested):

commands.pm
Martchus: I would rather keep it consistent with the upstream `MOJO_MAX_MESSAGE_SIZE` variable and not specify this in GiB.

Martchus: Fine, but *normally* HTTP limits are not in GiB.

okurz: yeah but *normally* users would not exceed 20 GiB then ;)

okurz (approved):

Martchus:
Only the packaging check fails due to `nothing provides libIlmImf-2_5.so.25()(64bit) needed by libopencv_imgcodecs4_5` (know problem). So I'm merging manually.

PR #1761, opened 2021-09-02
Increase MOJO_MAX_MESSAGE_SIZE to 40 due to SP4 migration

dzedro:
MOJO_MAX_MESSAGE_SIZE in test does not affect value on the worker

Fail: https://openqa.suse.de/tests/6986119
VR: https://openqa.suse.de/tests/6987492

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -236,8 +236,8 @@ sub get_temp_file {
 sub run_daemon {
     my ($port, $isotovideo) = @_;
 
-    # allow up to 20GB - hdd images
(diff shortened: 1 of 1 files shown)


Martchus:
I've been merging #1760. This makes it possible to override the limit using test variables or environment variables. So once it is deployed it is likely sufficient to set the test variable in the relevant test suite.

dzedro:
> I've been merging #1760. This makes it possible to override the limit using test variables or environment variables. So once it is deployed it is likely sufficient to set the test variable in the relevant test suite.

I personally don't understand the need for this variable, just another variable which is not needed.
Whatever size does published qcow2 have it can't be decreased despite anybody wants.
But it will work. :smile:

PR #1762, opened 2021-09-03
consoles::sshVirtsh: Remove unnecessary second argument

okurz:
The problem was introduced in
https://github.com/os-autoinst/os-autoinst/pull/1753. This should fix
some failed openQA tests that fail on this method.

Related progress issue: https://progress.opensuse.org/issues/98087

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -495,7 +495,7 @@ sub resume ($self) {
 
 sub get_remote_vmm () { get_var('VMWARE_REMOTE_VMM', '') }
 
-sub define_and_start ($self, $args) {
+sub define_and_start ($self) {
     my $remote_vmm = "";
     if ($self->vmm_family eq 'vmware') {
         my ($fh, $libvirtauthfilename) = tempfile(DIR => "/tmp/");


perlpunk (changes requested):
From my findings, `$args` can be removed completely, it's not used, and code calling this method never pass a second argument.

okurz:
Right, changed. Thank you

perlpunk (approved):

mloviska:
After some changes done to svirt related files ( I guess the signatures update has revealed it ) we can see [Too few arguments for subroutine 'consoles::sshVirtsh::stop_serial_grab' at /usr/lib/os-autoinst/backend/baseclass.pm line 838.
 at /usr/lib/os-autoinst/backend/console_proxy.pm line 65](https://openqa.suse.de/tests/7029546#step/patch_and_reboot/128) failures.

If I haven't overlooked something `stop_serial_grab` does not make use of `$args` either.
https://github.com/os-autoinst/os-autoinst/blob/master/backend/svirt.pm#L340 
https://github.com/os-autoinst/os-autoinst/blob/master/consoles/sshVirtsh.pm#L567

perlpunk:
@mloviska handled in #1763

mloviska:
> @mloviska handled in #1763

Thanks!

PR #1764, opened 2021-09-08
Fix `qemu-img` failure with QEMU 6.1.0 due to missing file format

Martchus:
* Specify format of backing file on image creation with `qemu-img`
* See https://progress.opensuse.org/issues/98117

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -84,7 +84,7 @@ $bdc->add_existing_drive('hd1', '/abs/path/sle15-minimal.qcow2', 'virtio-blk', 2
 @gcmdl = $bdc->gen_cmdline();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu command line for single existing drive');
 
-@cmdl  = ([qw(create -f qcow2 -b /abs/path/sle15-minimal.qcow2 raid/hd1-overlay0 22548578304)]);
+@cmdl  = ([qw(create -f qcow2 -F qcow2 -b /abs/path/sle15-minimal.qcow2 raid/hd1-overlay0 22548578304)]);
 @gcmdl = $bdc->gen_qemu_img_cmdlines();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu-img command line for single existing drive');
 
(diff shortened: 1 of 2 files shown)


kraih (approved):

kalikiana (approved):
That's even easier than I thought. Nice!

t/18-qemu.t
```diff
@@ -114,7 +114,7 @@ $proc  = qemu_proc('-foo', \%vars);
 @gcmdl = $proc->gen_cmdline();
 is_deeply(\@gcmdl, \@cmdl, 'Generate qemu command line for single existing UEFI disk using vars');
 
-@cmdl  = ([qw(create -f qcow2 -b), "$Bin/data/Core-7.2.iso", qw(raid/hd0-overlay0 11116544)]);
+@cmdl  = ([qw(create -f qcow2 -F qcow2 -b), "$Bin/data/Core-7.2.iso", qw(raid/hd0-overlay0 11116544)]);
```
AdamWill, line 117: This is not right. The backing format for an ISO should be raw, not qcow2. If you look at the logs of a run with qemu 6.0.0 or earlier, you will see that it auto-detects raw, not qcow2, as the backing format. I already pointed this out in the ticket as the reason why I did not send this trivial PR myself.

AdamWill:
:-1: 
As I wrote in the ticket, the fix *should* be this simple, but I don't believe it is. I believe this uses the wrong backing format for ISO (and also firmware, though we don't test that) images.

t/18-qemu.t
Martchus: To me it looks like it is just this test which is a bit odd because we specify an ISO as `HDD_1` here. But maybe it would actually make sense to nevertheless check the extension here. Note that the further checks below actually check whether the cd overlays have the format `raw`.

AdamWill:
Uh. Why did you merge this with an outstanding thumbs down and specific question about its correctness?

t/18-qemu.t
Martchus: https://github.com/os-autoinst/os-autoinst/pull/1765

kalikiana:
> Uh. Why did you merge this with an outstanding thumbs down and specific question about its correctness?

So, it seems what happened is this. We had two approvals (myself and @kraih). After that I enabled automerge. And after that you added more comments but didn't mark it as "needs changes". A 👎🏾 does not affect Mergify.

PR #1765, opened 2021-09-08
Deduce format of backing file from extension

Martchus:
* So even if we assign an ISO as HDD_x we would use the correct parameter
  `-F raw` (and not `-F qcow2`)
* See https://github.com/os-autoinst/os-autoinst/pull/1764/files#r704549754
  and https://progress.opensuse.org/issues/98117

### DIFF
--- a/OpenQA/Qemu/BlockDev.pm
+++ b/OpenQA/Qemu/BlockDev.pm
@@ -33,7 +33,7 @@ weakened.
 =cut
 
 package OpenQA::Qemu::BlockDev;
-use Mojo::Base 'OpenQA::Qemu::MutParams';
+use Mojo::Base 'OpenQA::Qemu::MutParams', -signatures;
 
 use Scalar::Util 'weaken';
 use OpenQA::Qemu::SnapshotConf;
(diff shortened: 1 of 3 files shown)


AdamWill:
This looks like it should work, but it just kinda niggles me that we have the code in `BlockDevConf.pm` which also effectively does this, but it's somehow not being used in this case. It'd be nice to know why the disk 'instances' being used here apparently weren't produced by `add_pflash_drive` / `add_iso_drive` and whether we could tweak things to avoid effective duplication there. But for a quick fix I think this looks good, thanks.

Amrysliu (approved):
LGTM

AdamWill:
so looking into it a bit harder, this isn't quite as good as qemu's previous detection. We (Fedora) actually use some base disk images whose filename is `foobar.img` but which are in qcow2 format (I don't remember why I didn't call them `foobar.qcow2`. qemu 6.0.0 and earlier correctly detect `qcow2` as the correct backing format for these, but this simple filename-based detection will pick `raw`. I don't know how qemu was actually doing the detection before, but I assume it looked at the file's magic number or something.

okurz:
I was thinking of maybe letting the user specify the type by a variable but in that case I guess one could also just rename the files to give the "hint". So, yeah, if you can live with renaming the files then it should be fine.

PR #1766, opened 2021-09-10
mmapi: Add whoami() function

cfconrad:
Add a function to get an easy way to retrieve the current jobID.
This jobID could be used later for external tools like pcw[1] to track
these jobs.

E.g. pcw track created public_cloud instances, these instances should
only live as long as the job is running. If a job get closed
unconditionally and the cleanup wasn't take place accordingly, pcw could
delete these instances right away. Currently we set a TTL and delete
these instances by age.

Ticked: https://progress.opensuse.org/issues/63604
VR: http://cfconrad-vm.qa.suse.de/tests/8292#step/clemix_job_id_tag/15

### DIFF
--- a/mmapi.pm
+++ b/mmapi.pm
@@ -22,7 +22,7 @@ use base 'Exporter';
 our @EXPORT = qw(get_children_by_state get_children get_parents
   get_job_info get_job_autoinst_url get_job_autoinst_vars
   wait_for_children wait_for_children_to_start api_call
-  api_call_2 handle_api_error
+  api_call_2 handle_api_error get_current_job_id
 );
 
 require bmwqemu;
@@ -287,6 +287,18 @@ sub wait_for_children_to_start {
         last unless $n;
         sleep $poll_interval;
     }
+}
+
+=head2 get_current_job_id
+
+    get_current_job_id();
+
+Query openQA's API to retrieve the current job ID 
+=cut
+sub get_current_job_id {
+    my $tx = api_call_2(get => 'whoami', undef, $CODES_EXPECTED_BY_MMAPI);
+    return undef if handle_api_error($tx, 'whoami', $CODES_EXPECTED_BY_MMAPI);
+    return $tx->res->json('/id');
 }
 
 1;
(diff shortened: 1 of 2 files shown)


mmapi.pm
```diff
@@ -289,4 +289,16 @@ sub wait_for_children_to_start {
     }
 }
 
+=head2 whoami
+
+    whoami();
+
+Query openqa api to retrive current jobID
```
Martchus, line 296:
```suggestion
Query openQA's API to retrieve the current job ID 
```

mmapi.pm
```diff
@@ -289,4 +289,16 @@ sub wait_for_children_to_start {
     }
 }
 
+=head2 whoami
+
+    whoami();
+
+Query openqa api to retrive current jobID 
+=cut
+sub whoami {
```
Martchus, line 298: `whoami` isn't really a meaningful name, what about:
```suggestion
sub query_current_job_id {
```

Martchus (changes requested):

mmapi.pm
cfconrad: Thx @Martchus
I choose the name `get_current_job_id()` as other functions in that module are called `get_*` too, even if they query openQA as well.

Martchus (approved):

asmorodskyi:
I would say that this is overkill . Isn't there simply JOBID variable which I can get with `get_var()` ? Looks odd that `whoami` function needs to ask 3d party about itself ...

Martchus:
> Isn't there simply JOBID variable which I can get with get_var() ?

I don't think the worker sets such a variable for all jobs because within the context of os-autoinst the openQA job ID shouldn't matter much.

cfconrad:
I'm just curious, where is the difference between `JOBTOKEN` and `JOBID` from the workers view. I mean, both should be unique to each `OPENQA_HOST`.

Martchus:
I guess at some point (before I was involved in that project) the `JOBTOKEN` was also supposed to be some kind of job-specific credential. Of course this is a bit pointless (at least at this point) because job variables are basically public.

cfconrad:
I see, thx for explanation. I think it is still valid as a `jobs/<id>` api request doesn't offer the TOKEN, or?

Martchus:
Out of my head I'd say the token is only accepted in a very limited number of routes. So you generally need the normal ID.

PR #1767, opened 2021-09-10
Allow whitespaces as delimiter for SCHEDULE variable

cfconrad:
This could make it convenient, if someone specify such variable as a
folded YAML multiline like:

```yaml
settings:
  SCHEDULE: >-
    boot/boot_to_desktop
    launch/my/test
```


Martchus (commented):
I'm personally fine with it. Spaces shouldn't be used in filenames here anyways, I suppose.

perlpunk (commented):
I would also be ok with it

OpenQA/Isotovideo/Utils.pm
```diff
@@ -200,7 +200,7 @@ sub load_test_schedule {
         unshift @INC, '.' unless path($bmwqemu::vars{CASEDIR})->is_abs;
         bmwqemu::fctinfo 'Enforced test schedule by \'SCHEDULE\' variable in action';
         $bmwqemu::vars{INCLUDE_MODULES} = undef;
-        autotest::loadtest($_ =~ qr/\./ ? $_ : $_ . '.pm') foreach split(',', $bmwqemu::vars{SCHEDULE});
+        autotest::loadtest($_ =~ qr/\./ ? $_ : $_ . '.pm') foreach split(/,|\s+/, $bmwqemu::vars{SCHEDULE});
```
perlpunk, line 203: Maybe it would be enough to allow a literal space, but not all kinds of whitespace `\s`

cfconrad: I thought about only space here as well, but I do not see a problem with `\s` and somehow I liked the idea to allow newlines or tabs as well.


> I'm personally fine with it. Spaces shouldn't be used in filenames here anyways, I suppose.


I checked `os-autoinst-dist-opensuse` with `find . | grep -P '\s'` there was nothing..

okurz (commented):
+1

OpenQA/Isotovideo/Utils.pm
cfconrad: @perlpunk  do you think we should go with space only? Like this:
```suggestion
        autotest::loadtest($_ =~ qr/\./ ? $_ : $_ . '.pm') foreach split(/,| +/, $bmwqemu::vars{SCHEDULE});
```
or you ok with `\s` ?

cfconrad:
any news? @okurz @perlpunk @Martchus

okurz:
@cfconrad I would be ok to merge if you apply the bit more conservative "space-only" approach. Otherwise I suggest to give the time to perlpunk to resolve the conversation with you

cfconrad:
I changed to conservative splitting `/[, ]+/`. With this we also strip spaced around a `,`.

Martchus (approved):

PR #1769, opened 2021-09-13
Recommend dependency icewm required by consoles/localXvnc.pm

asdil12:
Ticket: https://progress.opensuse.org/issues/97751


perlpunk (approved):

Martchus (changes requested):
Before you add a PR for every single application: We also need `xterm`, `xterm-console` and `xdotool`.

asdil12:
> Before you add a PR for every single application: We also need `xterm`, `xterm-console` and `xdotool`.

Added those to the PR.

Martchus (approved):

okurz (approved):

PR #1770, opened 2021-09-14
Install fonts-config needed by x3270 and require everything

asdil12:
Recommends don't work in containers

Only leave `dumponlyconsole` as a recommend as nothing
seems to provide it.

Ticket: https://progress.opensuse.org/issues/97751

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -81,10 +81,11 @@ Source0:        %{name}-%{version}.tar.xz
 %define test_requires %build_requires %spellcheck_requires %test_base_requires %yamllint_requires perl(Inline::Python) perl(YAML::PP)
 # The following line is generated from dependencies.yaml
 %define devel_requires %python_style_requires %test_requires ShellCheck perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy)
+%define s390_zvm_requires /usr/bin/xkbcomp /usr/bin/Xvnc x3270 icewm xterm xterm-console xdotool fonts-config mkfontdir mkfontscale
 BuildRequires:  %test_requires %test_version_only_requires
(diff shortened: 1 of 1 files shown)


Martchus (commented):
I would rather provide a split package `os-autoinst-full` similar to `qemu-kvm` which we already provide that has all the dependencies as `Requires`. You just needed to add another `%package` block below. (To avoid redundancy, use a macro instead of listing dependencies twice.)

Martchus:
Just for the record: We've just been debugging x3270 startup problems in a minimal containerized environment and the missing font-config prevented it to start. Not sure why it was never a problem before (even when we used containers previously on another host).

okurz (changes requested):
As stated elsewhere I am against requiring all these s390x specific components into generic os-autoinst.
For the ticket at hand please prioritize coming up with a quick-and-dirty solution, e.g. `podman run … bash` to have an interactive prompt in the container, do all what is necessary in the interactive bash session to have at least one s390x job completing and store that in a dirty shell script.

asdil12:
> As stated elsewhere I am against requiring all these s390x specific components into generic os-autoinst.
> For the ticket at hand please prioritize coming up with a quick-and-dirty solution, e.g. `podman run … bash` to have an interactive prompt in the container, do all what is necessary in the interactive bash session to have at least one s390x job completing and store that in a dirty shell script.

This would lead to broken jobs (the worker is already running when running `podman exec`) or raise conditions (if the worker is in that moment installing `os-autoinst-distri-opensuse-deps`).

If I create such a dirty bash script, that should be https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/install_deps.sh. Or https://github.com/os-autoinst/openQA/blob/master/container/worker/Dockerfile.

And with the new subpkg we should have a single additional package to put in there.

Martchus (approved):

PR #1771, opened 2021-09-14
Fix parameters of attach_to_running

Martchus:
The 2nd parameter is actually optional, this has been broken by
#1752.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -550,8 +550,10 @@ __END"
     return;
 }
 
-sub attach_to_running ($self, $args) {
-    my $name = ref($args) ? $args->{name} : $args;
+sub attach_to_running ($self, $args = undef) {
+    $args = {name => $args} unless ref $args;
+
+    my $name = $args->{name};
     $self->name($name) if $name;
     $self->backend->start_serial_grab($self->name);
 
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -559,9 +559,7 @@ sub attach_to_running ($self, $args = undef) {
 
     # Setting SVIRT_KEEP_VM_RUNNING variable prevents destruction of a perhaps valuable VM
     # outside of openQA. Set 'stop_vm' argument should the VM be destroyed at the end.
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -550,16 +550,16 @@ __END"
     return;
 }
 
-sub attach_to_running ($self, $args) {
-    my $name = ref($args) ? $args->{name} : $args;
+sub attach_to_running ($self, $args = undef) {
```
cfconrad, line 553: Just curious, would this be working as well, without the other changes:
```suggestion
sub attach_to_running ($self, $args = {}) {
```

Martchus: If one calls the function like `$console->attach_to_running(compute_name())` then the `$args = {}` version breaks if `compute_name()` returns `undef` but this worked before converting the function to signatures. So I restored that compatibility. (The only incompatibility left is that one cannot call the function with more than two arguments but I suppose that's ok.)

consoles/sshVirtsh.pm
```diff
@@ -550,16 +550,16 @@ __END"
...
+    $args = {name => $args} unless ref $args;
+
+    my $name = $args->{name};
```
mloviska, line 556: There are 2 callers of above subroutine in `./tests/installation/bootloader_hyperv.pm` and `./lib/opensusebasetest.pm`. As long as I have not misread it, the hash argument is not used to pass a name. The name parameter is usually part of the object. Would you mind to double-check, please?

okurz (commented):
+1. please clarify the question in https://github.com/os-autoinst/os-autoinst/pull/1771/files#r708962866 though I don't see it directly related to fixing the signature compatibility.

consoles/sshVirtsh.pm
Martchus: We *could* change the function to not allow setting the name - at least judging by the uses in "our" test distribution. However, I would like to limit the scope of this PR to fixing the parameter passing and not introduce further changes breaking backwards compatibility.

mloviska: Good point! I haven't thought outside our box. Thanks! :)

PR #1772, opened 2021-09-14
Minor improvements

okurz:
* lockapi: Simplify record_info call
* t: Improve needle test to handle wrong test execution

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -73,7 +73,7 @@ sub _log {
         $autotest::current_test->remove_last_result;
         $subject .= ' ' . int($args{amend} / ONE_MINUTE) . 'm' . $args{amend} % ONE_MINUTE . 's';
     }
-    testapi::record_info $subject, $msg;
+    record_info $subject, $msg;
 }
 
 sub _api_call_with_logging_and_error_handling {


needle.pm
```diff
@@ -80,7 +80,7 @@ sub new {
         catch {
             warn "broken json $jsonfile: $_";
         };
-        return undef unless $json;
+        return undef unless $json && $json eq 'HASHREF';
```
Martchus, line 83: Where does `HASHREF` come from? This looks similar to the usual `ref $json eq 'HASH'` check. Did you meant that?

Martchus (changes requested):

needle.pm
okurz: yes, I did. But even if I change it now tests fail and I remember at the time of writing tests could not pass. I will remove that commit.

lockapi.pm
```diff
@@ -73,7 +73,7 @@ sub _log {
         $autotest::current_test->remove_last_result;
         $subject .= ' ' . int($args{amend} / ONE_MINUTE) . 'm' . $args{amend} % ONE_MINUTE . 's';
     }
-    testapi::record_info $subject, $msg;
+    record_info $subject, $msg;
```
Martchus, line 76: You will need to import that function if you call it without package name. However, I'm not sure how this is simplifying things.

Martchus (changes requested):

lockapi.pm
okurz: Yeah, this is just a ridiculous commit. At the time I am sure it had more meat :)

PR #1774, opened 2021-09-15
Regression in 'consoles::sshVirtsh::get_remote_vmm'

mloviska:
`Too many arguments for subroutine 'consoles::sshVirtsh::get_remote_vmm'
at /usr/lib/os-autoinst/backend/baseclass.pm line 838, <$fh> line 45.`

- https://openqa.suse.de/tests/overview?arch=&flavor=&machine=svirt-vmware65&test=&modules=&module_re=&distri=sle&version=15-SP4&build=1.40&groupid=131#

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -493,7 +493,7 @@ sub resume ($self) {
     bmwqemu::diag "VM " . $self->name . " resumed";
 }
 
-sub get_remote_vmm () { get_var('VMWARE_REMOTE_VMM', '') }
+sub get_remote_vmm ($self) { get_var('VMWARE_REMOTE_VMM', '') }
 
 sub define_and_start ($self) {
     my $remote_vmm = "";


mloviska:
VR: http://kepler.suse.cz/tests/6849#

okurz (changes requested):
Can you instead change the test code to call the function not as an object method?

mloviska:
> Can you instead change the test code to call the function not as an object method?

Do you mean to shift a backend related code to the test code? Sure, could be done, but I assume it was added here for some other reason, and it would be better to keep backend code consistent

Martchus (approved):
I suppose this is really supposed to be a method, even if `$self` is not used (currently).

cfconrad:
From my understanding, the `VMWARE_REMOTE_VMM` is dynamically set in `consoles::sshVirtsh` and for me the best way is to get it from the object via `$self->get_remote_vmm()`.

Actually I don't understand, why we set it as openqa VAR. For me it looks like something which could be avoided.

Conclusion, I wouldn't go to make this function static, even if the current implementation makes it possible!

okurz (approved):

PR #1776, opened 2021-09-17
Add aarch64 qemu binary

nicksinger:
I'm not sure why we only have x86_64 and ppc64 so I think it makes sense to also include aarch64 here :)


okurz (approved):
I did not find reason against, let's see what this breaks ;)

PR #1777, opened 2021-09-20
Add OPENQA_HOSTNAME to xml metadata

foursixnine:
As a follow up to 3b277957a and os-autoinst/openQA/4216 add more information to the xml metadata to avoid guessing where a remote worker
is running


okurz (changes requested):
Hm, not a friend of making os-autoinst depend on anything openQA specific. I understand the use case. Maybe we can find something that does not prescribe openQA per se? So from os-autoinst perspective you would just run a test "from local". How about if just openQA worker writes something into it's logfile?

foursixnine:
> Hm, not a friend of making os-autoinst depend on anything openQA specific. I understand the use case. Maybe we can find something that does not prescribe openQA per se? 

Well if you tell me that there's a solution comming for a worker to report to all of it's webUI's that it's busy, and that it's settings don't collide with the others... Or do you have another idea in mind?

Otherwise, No. I doubt anybody is running os-autoinst just for svirt, and if they do, they'll come with pitchforks when needed. 

> So from os-autoinst perspective you would just run a test "from local". 

Except when you're running os-autoinst using a backend that relies on a jump host, that can also be used by others.

> How about if just openQA worker writes something into it's logfile?

No.

In this case, I'll quote you:

> how you would find jobs if not over the webUI unless this is about workers rich don't delete it's pool dir. Could you explain again for dummies like me?

The change is being added exactly because of openQA: There was the situation already where a worker was being shared among multiple webUI's, settings collided, and it took guessing to figure out where was the problem.

Explaining it for dummies:

```
openQA Instance 64: 00006848-sle-micro-5.1-DVD-s390x-Build60.1_12.19-sle-micro_installation+update@s390x-kvm-sle12

 Interface   Type     Source   Model    MAC
-----------------------------------------------------------
 macvtap3    direct   eth0     virtio   52:54:00:8f:b9:23

openQA Instance 5: 07071977-sle-15-SP2-Server-DVD-Updates-s390x-Build20210910-2-sssd_389ds_functional@s390x-kvm-sle12

 Interface   Type     Source   Model    MAC
-----------------------------------------------------------
 -           direct   eth0     virtio   52:54:00:8f:b9:23
 ```

Unless you know where `openQA instance 64` and `openQA instance 5` are running, it's impossible to know in which webUI should the logfiles be checked; this problem was happening often but clearly not looked at, either by lack of expertise or interest.

okurz (approved):
ok, I see that OPENQA_HOSTNAME is already used elsewhere in here, e.g. in OpenQA/Isotovideo/NeedleDownloader.pm so ok.

PR #1778, opened 2021-09-20
mmapi: Allow for longer and more retries after observed failures

okurz:
The initial retry to mmapi calls as th the sleep periods introduced in
dee5d4cf seem to still run into temporary connection outages which we
should be able to pass with more and longer retries.

Related progress issue: https://progress.opensuse.org/issues/98940

### DIFF
--- a/mmapi.pm
+++ b/mmapi.pm
@@ -30,8 +30,8 @@ require bmwqemu;
 use Mojo::UserAgent;
 use Mojo::URL;
 
-our $retry_count    = $ENV{OS_AUTOINST_MMAPI_RETRY_COUNT}    // 3;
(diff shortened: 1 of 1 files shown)


kalikiana (approved):
So that's up to 10 attempts, 300s total max.

It feels like a lot, so just being verbose here on what we're accepting.

okurz:
> So that's up to 10 attempts, 300s total max.
> 
> It feels like a lot, so just being verbose here on what we're accepting.

I know, it's a huge step from before. I targeted 5m ~ 300s. The question about the specific values can be: What would we do if timeout is hit after let's say 40s. I am pretty sure people would still suggest to wait longer. But 300s should cover even complete webUI restarts in case they really take that long and even the webUI hosts reboots completely or something like that.

PR #1779, opened 2021-09-22
Remove check for qemu >= 4 in qemu backend

mimi1vx:
This change fixes poo#98946 and also drop support for
distributions older than Leap 15.2 / SLE 15-SP2

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -683,7 +683,7 @@ sub start_qemu {
         # newer qemu needs safe cache capability level quirk settings
         # https://progress.opensuse.org/issues/75259
         my $caps = ',cap-cfpc=broken,cap-sbbc=broken,cap-ibs=broken';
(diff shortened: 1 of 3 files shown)


asdil12:
Clickable link for the lazy ones: https://progress.opensuse.org/issues/98946

Martchus (changes requested):
* Please use the full URL to reference to progress tickets. Otherwise accessing them is rather tedious.
* I'm ok with removing the check in our Perl code but I don't think it is necessary or useful to remove the check in `dependencies.yaml` and the `.spec`-file.

asdil12 (approved):
LGTM

asdil12:
> * Please use the full URL to reference to progress tickets. Otherwise accessing them is rather tedious.
> * I'm ok with removing the check in our Perl code but I don't think it is necessary or useful to remove the check in `dependencies.yaml` and the `.spec`-file.

I think the test was actually moved to the spec as it is now a hard requirement as we don't workaround old qemu odds.

Martchus (approved):

PR #1780, opened 2021-09-23
Add support for timeout on SSH based commands if possible

okurz:
Net::SSH2 supports a timeout command which so far was not used by us.
The command is available since libssh2 >= 1.2.9 which we commonly have
by now. For example in openSUSE Leap 15.2 there is libssh2-1-1.9.0.
Applying a timeout on individual SSH commands should prevent openQA jobs
hitting "timeout_exceeded" with the additional benefit that jobs can
fail earlier than the default 2h openQA job timeout.

To be backward compatible here we do not apply any timeout. To configure
the value the new test variable SSH_COMMAND_TIMEOUT_S can be set to a
different value.

Tested with `t/23-baseclass.t t/26-ssh_screen.t t/31-sshSerial.t` as
regression tests and manual experimentation with the parameter in `perl
-de1`.

Related progress issue: https://progress.opensuse.org/issues/99123

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1,5 +1,5 @@
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2020 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -40,6 +40,7 @@ use List::MoreUtils 'uniq';
 use Scalar::Util 'looks_like_number';
 use Mojo::File 'path';
 use OpenQA::Exceptions;
+use Time::Seconds;
 
 # should be a singleton - and only useful in backend process
 our $backend;
@@ -1240,7 +1241,8 @@ sub new_ssh_connection {
         }
     }
 
-    my $ssh = Net::SSH2->new;
+    # timeout requires libssh2 >= 1.2.9 so not all versions might have it
+    my $ssh = Net::SSH2->new(timeout => ($bmwqemu::vars{SSH_COMMAND_TIMEOUT_S} // 0) * 1000);
 
     # Retry multiple times, in case of the guest is not running yet
     my $counter    = $bmwqemu::vars{SSH_CONNECT_RETRY} // 5;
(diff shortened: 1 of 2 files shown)


okurz:
Updated the PR to actually not apply a timeout by staying at the default of having no timeout. This allows us to merge even without validation tests on production test scenarios. After merge on a case by case base a timeout can be applied and even eventually potentially we can introduce a sensible default as timeout.

backend/baseclass.pm
```diff
@@ -1241,6 +1242,8 @@ sub new_ssh_connection {
     }
 
     my $ssh = Net::SSH2->new;
+    # timeout requires libssh2 >= 1.2.9 so not all versions might have it
+    $ssh->timeout($bmwqemu::vars{SSH_COMMAND_TIMEOUT_S} * 1000) if $ssh->can('timeout');
```
Martchus, line 1246: I suppose the `timeout` method will actually always be available as it is provided by the Perl module and not the underlying libssh2 library. I would just set the option via the constructor:

```suggestion
    # timeout requires libssh2 >= 1.2.9
    my $ssh = Net::SSH2->new(timeout => ($bmwqemu::vars{SSH_COMMAND_TIMEOUT_S} // 10 * ONE_MINUTE) * 1000);
```

Martchus, line 1246: Now that you've changed it to `$bmwqemu::vars{SSH_COMMAND_TIMEOUT_S} * 1000`, won't this print a warning if the variable is `undef`?

Martchus (changes requested):

backend/baseclass.pm
okurz: Argh, right, the `* 1000`, grmbl. Updated along with your other suggestion

okurz:
created https://progress.opensuse.org/issues/99126 for the CI failure

backend/baseclass.pm
```diff
@@ -1240,7 +1241,8 @@ sub new_ssh_connection {
         }
     }
 
-    my $ssh = Net::SSH2->new;
+    # timeout requires libssh2 >= 1.2.9 so not all versions might have it
```
kalikiana, line 1244: So what does this mean? The following could can fail? What is the comment telling me?

okurz: I expect that the timeout would be ignored

kalikiana (approved):

kalikiana:
For reference:

```
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
./tools/install-new-deps.sh: line 21:    14 Aborted                 sudo zypper --no-refresh install -y -C "${DEPS[@]}"
```

Martchus (approved):
Looks good. You'll have to rebase if you want to give the CI a chance to pass.

PR #1784, opened 2021-09-24
Use Leap 15.3 container

Martchus:
* Follow-up of efbeca4fd892c311d
* See https://progress.opensuse.org/issues/99126

### DIFF
--- a/tools/docker_run_ci
+++ b/tools/docker_run_ci
@@ -39,7 +39,7 @@ done
 
 CI_ENV=$(bash <(curl -s https://codecov.io/env))
 
-IMAGE=registry.opensuse.org/devel/openqa/containers/os-autoinst_dev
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

okurz (approved):

Martchus:
Now tests fail due to a missing Python dependency. Apparently the conditions for Python versions don't work as intended when simply linking packages from Factory. I tried to define the variable in the condition within the project config but it didn't work because it is apparently overridden in the package build (`python_flavor=python3`). So I amended the package spec file itself which worked (resulting in a linkdiff, not sure whether that's a problem: https://build.opensuse.org/package/rdiff/devel:openQA:Leap:15.3/python-click?opackage=python-click&oproject=openSUSE%3AFactory&rev=2). I've re-triggered the container build.

PR #1786, opened 2021-09-29
Use simpler 'podman' as drop-in replacement for our checks

okurz:
podman can be used without a system daemon.

### DIFF
--- a/tools/static_check_containers
+++ b/tools/static_check_containers
@@ -8,5 +8,5 @@ if [ "$(curl -s -w "%{http_code}" https://ghcr.io/v2/ -o /dev/null)" = "401" ]; 
 else
     img=ghcr.io/hadolint/hadolint
 fi
-cre="${cre:-"docker"}"
+cre="${cre:-"podman"}"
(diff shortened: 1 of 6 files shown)


tools/static_check_containers
```diff
@@ -8,5 +8,5 @@ if [ "$(curl -s -w "%{http_code}" https://ghcr.io/v2/ -o /dev/null)" = "401" ];
 else
     img=ghcr.io/hadolint/hadolint
 fi
-cre="${cre:-"docker"}"
+cre="${cre:-"podman"}"
```
kalikiana, line 11: Should there be a change to remove or disable the docker daemon?

okurz: I don't think we explicitly install docker anywhere. Let's see what happens if I replace more. Updated the PR to include a second commit which generalizes more places.

okurz:
There should be no coverage change. That's likely flaky coverage that we should fix

perlpunk (approved):

PR #1789, opened 2021-10-01
Use proper perl "signatures" in consoles::VNC

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -117,9 +117,7 @@ my @encodings = (
     },
 );
 
-sub login {
-    my ($self, $connect_timeout) = @_;
-    $connect_timeout //= 10;
+sub login ($self, $connect_timeout = 10) {
     # arbitrary
     my $connect_failure_limit = 2;
 
@@ -178,9 +176,7 @@ sub login {
     }
 }
 
-sub _handshake_protocol_version {
-    my ($self) = @_;
-
+sub _handshake_protocol_version ($self) {
     my $socket = $self->socket;
     $socket->read(my $protocol_version, 12) || die 'unexpected end of data';
     my $protocol_pattern = qr/\A RFB [ ] (\d{3}\.\d{3}) \s* \z/xms;
@@ -207,9 +203,7 @@ sub _handshake_protocol_version {
     $socket->print($protocol_version);
 }
 
-sub _handshake_security {
-    my $self = shift;
-
+sub _handshake_security ($self) {
     my $socket = $self->socket;
 
     # Retrieve list of security options
@@ -363,17 +357,12 @@ sub _handshake_security {
     }
 }
 
-sub _client_initialization {
-    my $self = shift;
-
+sub _client_initialization ($self) {
     my $socket = $self->socket;
-
     $socket->print(pack('C', !$self->ikvm));    # share
 }
 
-sub _server_initialization {
-    my $self = shift;
-
+sub _server_initialization ($self) {
     my $socket = $self->socket;
     $socket->read(my $server_init, 24) || die 'unexpected end of data';
 
@@ -501,9 +490,7 @@ sub _server_initialization {
     }
 }
 
-sub _send_key_event {
-    my ($self, $down_flag, $key) = @_;
-
+sub _send_key_event ($self, $down_flag, $key) {
     # A key press or release. Down-flag is non-zero (true) if the key is now pressed, zero
     # (false) if it is now released. The key itself is specified using the "keysym" values
     # defined by the X Window System.
@@ -522,13 +509,11 @@ sub _send_key_event {
         ));
 }
 
-sub send_key_event_down {
-    my ($self, $key) = @_;
+sub send_key_event_down ($self, $key) {
     $self->_send_key_event(1, $key);
 }
 
-sub send_key_event_up {
-    my ($self, $key) = @_;
+sub send_key_event_up ($self, $key) {
     $self->_send_key_event(0, $key);
 }
 
@@ -608,7 +593,6 @@ my $keymap_ikvm = {
 };
 
 sub shift_keys {
-
     # see http://en.wikipedia.org/wiki/IBM_PC_keyboard
     return {
         '~' => '`',
@@ -643,9 +627,7 @@ sub shift_keys {
 
 ## use critic
 
-sub init_x11_keymap {
-    my ($self) = @_;
-
+sub init_x11_keymap ($self) {
     return if $self->keymap;
     # create a deep copy - we want to reuse it in other instances
     my %keymap = %$keymap_x11;
@@ -669,9 +651,7 @@ sub init_x11_keymap {
     $self->keymap(\%keymap);
 }
 
-sub init_ikvm_keymap {
-    my ($self) = @_;
-
+sub init_ikvm_keymap ($self) {
     return if $self->keymap;
     my %keymap = %$keymap_ikvm;
     for my $key ("a" .. "z") {
@@ -694,9 +674,7 @@ sub init_ikvm_keymap {
 }
 
 
-sub map_and_send_key {
-    my ($self, $keys, $down_flag, $press_release_delay) = @_;
-
+sub map_and_send_key ($self, $keys, $down_flag, $press_release_delay) {
     die "need delay" unless $press_release_delay;
 
     if ($self->ikvm) {
@@ -742,8 +720,7 @@ sub map_and_send_key {
     }
 }
 
-sub send_pointer_event {
-    my ($self, $button_mask, $x, $y) = @_;
+sub send_pointer_event ($self, $button_mask, $x, $y) {
     bmwqemu::diag "send_pointer_event $button_mask, $x, $y, " . $self->absolute;
 
     my $template = 'CCnn';
@@ -761,9 +738,7 @@ sub send_pointer_event {
 
 # drain the VNC socket from all pending incoming messages.  return
 # true if there was a screen update.
-sub update_framebuffer {    # upstream VNC.pm:  "capture"
-    my ($self) = @_;
-
+sub update_framebuffer ($self) {
     try {
         local $SIG{__DIE__} = undef;
         my $have_recieved_update = 0;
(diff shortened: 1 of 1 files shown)


consoles/serial_screen.pm
```diff
@@ -168,9 +161,7 @@ An undefined timeout will cause to wait indefinitely. A timeout of 0 means to
 just read once.
 
 =cut
-sub do_read
-{
-    my ($self, undef, %args) = @_;
+sub do_read ($self, undef, %args) {
```
Martchus, line 164: This one looks quite strange (include the previous version). It is called like this:

```
        my $read = $self->do_read($buf, max_size => $buflen / 2, timeout => remaining($sttime, $timeout));
        next READ unless (defined($read));
```

So `$buf` is basically ignored. Maybe we should get rid of the parameter in both functions and the invocation.

consoles/serial_screen.pm
```diff
@@ -231,8 +222,7 @@ C<{ matched => 0, string => 'text from the terminal' }>
 on failure.
 
 =cut
-sub read_until {
-    my ($self, $pattern, $timeout) = @_[0 .. 2];
+sub read_until ($self, $pattern, $timeout) {
     my $fd       = $self->{fd_read};
     my %nargs    = @_[3 .. $#_];
```
Martchus, line 227: This line must go into the signature as well, otherwise the function only expects 3 arguments.

consoles/ssh_screen.pm
```diff
@@ -25,9 +25,8 @@ has ssh_channel    => undef;
 
 use constant TYPE_STRING_TIMEOUT => ONE_MINUTE;
 
-sub new {
-    my $class = shift;
-    my $self  = bless @_ ? @_ > 1 ? {@_} : {%{$_[0]}} : {}, ref $class || $class;
+sub new ($class) {
+    my $self = bless @_ ? @_ > 1 ? {@_} : {%{$_[0]}} : {}, ref $class || $class;
```
Martchus, line 29: I don't think this is going to work. The function will only accept one argument and is `@_` really implicitly shifted?

Martchus (changes requested):
Hard to tell whether it'll work without checking all invocations. I guess we'll have to risk it but it would make nevertheless sense to run at least some openSUSE tests locally.

consoles/ssh_screen.pm
okurz: Right. Changed that to a second argument `@args`

consoles/serial_screen.pm

consoles/serial_screen.pm
okurz: For now I changed the second parameter to be `$` to ignore. I will look into your suggestion

consoles/serial_screen.pm
```diff
@@ -166,20 +136,15 @@ sub do_read
...
     my $read;
     while (!defined($read)) {
         $read = sysread($fd, $buffer, $args{max_size});
-        if (!defined($read) && !($ERRNO{EAGAIN} || $ERRNO{EWOULDBLOCK})) {
-            croak "Failed to read from virtio/svirt serial console char device: $ERRNO";
-        }
+        croak "Failed to read from virtio/svirt serial console char device: $ERRNO" if !defined($read) && !($ERRNO{EAGAIN} || $ERRNO{EWOULDBLOCK});
     }
-    $_[1] = $buffer;
+    $buf = $buffer;
```
cfconrad, line 147: We use this called-by-reference, its equal to the `sysread(..., $buffer,...)` not sure how to do this with signatures.
```suggestion
    $_[1] = $buffer; 
```

Martchus: We could require one to specify a scalar ref here. That would make it more obvious that the variable *outside* the function is re-assigned.

okurz: do you have an idea why this test would only fail within OBS and not GHA?

cfconrad: Looks like some GHA glitch it failes: https://github.com/os-autoinst/os-autoinst/runs/4386959135?check_suite_focus=true#step:3:572

kalikiana:
I see a failure, although it doesn't fail CI:

```
8: [20:35:09] ./13-osutils.t ........................... ok     1221 ms ( 0.01 usr  0.00 sys +  1.05 cusr  0.17 csys =  1.23 CPU)
8: 
8:     #   Failed test 'isotovideo exit code'
8:     #   at ./14-isotovideo.t line 128.
8:     #          got: '1'
8:     #     expected: '0'
8: 
8:     #   Failed test 'requested image is published even though the job failed'
8:     #   at ./14-isotovideo.t line 131.
```

consoles/serial_screen.pm
perlpunk: Assigning to `$_[...]` willl still work with signatures, as `@_` is still working as before.

There is a dicussion on removing `@_`, though:
https://www.nntp.perl.org/group/perl.perl5.porters/2021/11/msg262052.html

okurz:
pulled out one part with https://github.com/os-autoinst/os-autoinst/pull/1899

consoles/serial_screen.pm
```diff
@@ -118,19 +104,13 @@ sub normalise_pattern ($pattern, $no_regex) {
         my $hr = shift @$pattern;
         if (@$pattern > 0) {
             my $re = qr/($hr)/;
-            for my $r (@$pattern) {
-                $re .= qr/|($r)/;
-            }
+            $re .= qr/|($_)/ for @$pattern;
```
perlpunk, line 107: hm, looks like all this can be simplified (and corrected) by:
```
my $re = join "|", map { "($_)" } @$pattern;
return qr{$re};
```
This way also in the case of one element it will get parens `()`

perlpunk: I wonder why the various patterns should all get their own parens.
`(a|b|c)` should be the same as `(a)|(b)|(c)`, and then the code which uses the regex can just use `$1` instead of `@@LAST_MATCH_START`

okurz: The above does not work for me. Fails with

```
t/10-terminal.t .. 1/?   
#   Failed test '[Child 22536] direct: use array of regexs'
#   at t/10-terminal.t line 366.
#                   '
# 
# 
# Welcome to SUSE Linux Enterprise Server 12 SP2 RC3 (x86_64) - Kernel 4.4.21-65-default (hvc0).
# 
# linux-5rw7'
#     doesn't match '(?^u:(?^u:.*\
# \
# Welcome\ to\ SUSE\ Linux\ Enterprise\ Server\ 12\ SP2\ RC3\ \(x86_64\)\ \-\ Kernel\ 4\.4\.21\-65\-default\ \(hvc0\)\.\
# \
# linux\-5rw7\ login\:\ ))'
# Looks like you failed 1 test of 31.
```

perlpunk: Then the test should be adjusted. The regex is needlessly verbose. But I guess that's for another PR.

perlpunk: Yeah. Somehow I wasn't able to mark the whole block for my comment.

okurz:
An openQA fullstack test against this change was also successful.

okurz:
Updated with a 100% statement coverage unit test for consoles::vnc_base

consoles/VNC.pm
```diff
@@ -122,9 +122,7 @@ my @encodings = (
     },
 );
 
-sub login {
-    my ($self, $connect_timeout) = @_;
-    $connect_timeout //= 10;
+sub login ($self, $connect_timeout = 10) {
```
Martchus, line 125: This will break. Setting a default via signatures is *not* equivalent to the current default initialization. We call the function like this:

```
$self->{vnc}->login($args->{connect_timeout});
```

If `$args->{connect_timeout}` is `undef` we don't get the default value with your change.

Martchus (changes requested):

consoles/VNC.pm
okurz: So what do you suggest? Does

```suggestion
sub login ($self, $connect_timeout = undef) {
    $connect_timeout //= 10;
```

work?

Martchus: Yes, this should work.

okurz:
@Martchus

> Unfortunately coverage for consoles/VNC.pm is missing.

I agree. Though I have the feeling some reporting for that is missing. See e.g. https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/1789/diff?src=pr&el=tree#diff-Y29uc29sZXMvVk5DLnBt, It can't be true that only 5% is covered, considering t/99-full-stack.t. Maybe we need to put some additional `Devel::Cover::report() if Devel::Cover->can('report')` in the code somewhere but I don't know yet where.

Martchus:
It should be covered by the fullstack test but that one runs the driver code in a sub process. So yes, `Devel::Cover::report() if Devel::Cover->can('report')` is likely needed somewhere. We already have it in `baseclass.pm` but maybe not in all required places and maybe the condition `Devel::Cover->can('report')` is not true when running the driver process as part of the fullstack test.

consoles/VNC.pm
okurz: ok, updated the change accordingly

okurz:
Updated with further small improvements and more tests, e.g. for coverage in consoles::serial_screen. No relevant changes to consoles::VNC coverage however as I still don't know how to approach that.

okurz:
After https://github.com/os-autoinst/os-autoinst/pull/1913 I should check again if we actually covered all consoles

okurz:
18-qemu-options.t fails, don't understand the specific problem yet. Probably should bisect.

okurz:
Split out https://github.com/os-autoinst/os-autoinst/pull/1926 and updated this PR to now only include the single left commit about consoles::VNC. The patch coverage will be still be our automatic acceptance threshold due to background processes not collecting coverage properly. Please review https://github.com/os-autoinst/os-autoinst/pull/1926 first or approve both here if you can

perlpunk (approved):

baierjan (approved):

PR #1792, opened 2021-10-01
Use proper perl "signatures" in t

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.


kalikiana (approved):

t/01-test_needle.t
```diff
@@ -36,8 +36,7 @@ throws_ok(
     'died when constructing needle without prior call to needle::init()'
 );
 
-sub needle_init {
-    my $ret;
+sub needle_init ($ret) {
```
perlpunk, line 39:
```suggestion
sub needle_init () {
```

t/04-check_vars_docu.t
```diff
@@ -117,9 +116,7 @@ EO_BACKEND_FOOTER
     }
 }
 
-sub read_backend_pm {
-    my ($backend) = $_ =~ /^([^\.]+)\.pm/;
-    return unless $backend;
+sub read_backend_pm ($backend) {
```
perlpunk, line 119: That should be reverted.
`$backend` gets the value of `$_`, not `$_[0]`

t/10-terminal.t
```diff
@@ -232,13 +226,11 @@ sub fake_terminal {
...
 
-sub is_matched {
-    my ($result, $expected, $name) = @_;
+sub is_matched ($result, $expected, $name) {
     report_child_test(is_deeply => $result, {matched => 1, string => $expected}, $name);
 }
 
-sub test_terminal_directly {
-    my $tb = Test::Most->builder;
+sub test_terminal_directly ($tb) {
```
perlpunk, line 233: Should be reverted, it doesn't get an argument

t/30-mmapi.t
```diff
@@ -40,9 +41,9 @@ $bmwqemu::vars{OPENQA_URL} = 'http://not/relevant';
 $bmwqemu::vars{JOBTOKEN}   = 'fake-jobtoken';
 
 # define helper to call a function by its name
-sub call {
-    my $function_name = shift;
-    __PACKAGE__->can($function_name)->(@_);
+#sub call ($function_name, @) {
```
Martchus, line 44: Looks like a leftover.

Martchus (changes requested):

t/30-mmapi.t
okurz: fixed

t/03-testapi.t
```diff
@@ -486,9 +482,11 @@ subtest 'record_info' => sub {
     like(exception { record_info('my title', 'output', result => 'not supported', resultname => 'foo') }, qr/unsupported/, 'invalid result');
 };
 
-sub script_output_test {
-    my $is_serial_terminal = shift;
-    my $mock_testapi       = Test::MockModule->new('testapi');
+# unsure why the warning "The signatures feature is experimental" shows up
+# despite the signature enablement in the file
+no warnings 'experimental::signatures';
```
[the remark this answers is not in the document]
kraih: The problem is [here](https://github.com/os-autoinst/os-autoinst/blob/f34a40a3258d1fee5e8e413cbbebf11c1b92bb77/basetest.pm#L24), the `use basetest;` in the test re-enables all warnings.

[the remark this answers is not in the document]
okurz: Oh, I see. Do you have a suggestion what to do there? Move the `no warnings` there?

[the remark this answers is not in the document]
kraih: Yea, i would move the `no warnings ...` right below the `use basetest;`. Or move the `use basetest;` above the `use Mojo::Base -strict, -signatures;`. All `use` statements are executed at compile time anyway, might as well have them all at the top of the file.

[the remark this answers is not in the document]
kraih: Disabling signature warnings in `basetest` should be fine too though.

PR #1799, opened 2021-10-03
Use consistent "Copyright" header without problematic characters

okurz:
* Use consistent "Copyright" header without problematic characters

If we accept these changes I plan to apply the same to other repositories
within the scope of https://github.com/os-autoinst

Previously this PR included another commit "Simplify Copyright entry for SUSE to not need to update recurringly
" https://github.com/os-autoinst/os-autoinst/commit/d0af9e9b143820311c0348383b54731c17701919 which I removed based on https://github.com/os-autoinst/os-autoinst/pull/1800#pullrequestreview-769896377

Related progress issue: https://progress.opensuse.org/issues/98616

### DIFF
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -262,7 +262,7 @@ Inspiration from my colleagues at L<http://www.nestoria.co.uk>
 
 =head1 COPYRIGHT
 
-Copyright (C) 2006 Edmund von der Burg. All rights reserved.
+Copyright 2006 Edmund von der Burg. All rights reserved.
 
 This module is free software; you can redistribute it and/or modify it under
 the same terms as Perl itself. If it breaks you get to keep both pieces.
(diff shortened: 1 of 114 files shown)


kalikiana (approved):
I like that this change comes with a unit test.

AdamWill:
Just FWIW, RH's current guidance on this is:

Brief Summary
There are two recommended approaches for copyright notices in original source files in Red Hat-maintained open source projects. You may choose either approach depending on the preferences or needs of your project.

Method One: State “Copyright the XYZ Authors” (or similar formulation — see below) where XYZ is the project name. Under this approach, the project does not make use of traditional individual and company-specific copyright notices.

Method Two: State “Copyright Red Hat”.

For either of these two methods, don’t bother including years, copyright symbols (i.e., © or approximations like (C)), or (for the second method) the “Inc.” in the Red Hat corporate name.

i.e. our lawyers don't reckon it's necessary/useful to include years, or things like "Inc.", "Ltd." or "LLC".

okurz:
Ok, cool. Haven't thought about the LLC but I am already recommending to leave out the year in new files

PR #1800, opened 2021-10-03
Use abbreviated SPDX license identifiers

okurz:
Follow-up to https://github.com/os-autoinst/os-autoinst/pull/1799

Following https://spdx.dev/ids/ we can replace verbatim license text in
source codes with convenient SPDX license identifiers. This helps to
save (screen-)space, provides consistency, prevents misuse and is both
better readable by humans and automations.

I used the command

```
sed -i ':a;N;$!ba;s@#\n# This program is free software.*if not, see <http://www.gnu.org/licenses/>.@# SPDX-License-Identifier: GPL-2.0-or-later@g' $(git grep -I -l 'Copyright.*[0-9]\{4\}.*SUSE' ':!COPYING' ':!external/')
```

to replace all files with usual exceptions. A similar command has been
included in a style check test.

The file osutils.pm needed manual correction as the previous license
header text was incomplete.

Some files are without an explicit license statement which I have kept
as is. Effectively the license of the project applies which is
"GPL-2.0-or-later" as well.

There are some files which are marked with the text

```
This module is free software; you can redistribute it or modify it
under the same terms as Perl itself.
```

meaning

```
Artistic-1.0 OR GPL-1.0-or-later
```

see https://dev.perl.org/licenses/

Those I adapted as well manually.

https://progress.opensuse.org/issues/98616

### DIFF
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -262,13 +262,9 @@ Inspiration from my colleagues at L<http://www.nestoria.co.uk>
 
 =head1 COPYRIGHT
 
-Copyright 2006 Edmund von der Burg. All rights reserved.
+Copyright 2006 Edmund von der Burg
 
-This module is free software; you can redistribute it and/or modify it under
-the same terms as Perl itself. If it breaks you get to keep both pieces.
-
-THERE IS NO WARRANTY.
-
+SPDX-License-Identifier: Artistic-1.0 OR GPL-1.0-or-later
 =cut
 
 1;
--- a/autotest.pm
+++ b/autotest.pm
@@ -1,18 +1,6 @@
 # Copyright 2009-2013 Bernhard M. Wiedemann
 # Copyright 2012-2021 SUSE LLC
-#
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License along
-# with this program; if not, see <http://www.gnu.org/licenses/>.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
 package autotest;
 
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -1,18 +1,6 @@
 # Copyright 2009-2013 Bernhard M. Wiedemann
 # Copyright 2012-2021 SUSE LLC
-#
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License along
-# with this program; if not, see <http://www.gnu.org/licenses/>.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
 package consoles::localXvnc;
 
--- a/OpenQA/Commands.pm
+++ b/OpenQA/Commands.pm
@@ -1,17 +1,5 @@
 # Copyright 2018-2019 SUSE LLC
-#
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License along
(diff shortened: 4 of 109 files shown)


OpenQA/Benchmark/Stopwatch.pm
```diff
@@ -262,13 +262,9 @@ Inspiration from my colleagues at L<http://www.nestoria.co.uk>
...
 
-Copyright (C) 2006 Edmund von der Burg. All rights reserved.
-
-This module is free software; you can redistribute it and/or modify it under
-the same terms as Perl itself. If it breaks you get to keep both pieces.
-
-THERE IS NO WARRANTY.
+Copyright 2006 Edmund von der Burg
 
+SPDX-License-Identifier Artistic-1.0 OR GPL-1.0-or-later
```
coolo, line 267: the format is with `:` though

autotest.pm
```diff
@@ -1,18 +1,6 @@
...
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License along
-# with this program; if not, see <http://www.gnu.org/licenses/>.
+# Copyright 2009-2013 Bernhard M. Wiedemann
+# Copyright 2012 SUSE LLC
```
coolo, line 2: somehow you lost all the `-2021` for SUSE

okurz: yes, thanks for noticing. That was the intention of https://github.com/os-autoinst/os-autoinst/pull/1800/commits/d0af9e9b143820311c0348383b54731c17701919 in particular. Please see the individual commit messages where I tried to provide sufficient detailed reasoning

okurz:
@bmwiedemann @lnussel WDYT about this change? Please see the individual commit messages where I tried to provide sufficient detailed reasoning

OpenQA/Benchmark/Stopwatch.pm
okurz: thx, I did that wrong in my manual edits of the two "artistic" files. Corrected as well as the C/C++ files I missed. Updated PR

autotest.pm
coolo: I disagree with that change. Even your googe link says adding the current year is fine on modifications. https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html explicity says to add the current year on non-trivial changes. And I doubt it's up to you to dedicde *now* what were non-trivial changes at the time.

consoles/localXvnc.pm
```diff
@@ -1,18 +1,6 @@
...
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License along
-# with this program; if not, see <http://www.gnu.org/licenses/>.
+# Copyright 2009-2013 Bernhard M. Wiedemann
+# Copyright 2012 SUSE LLC
```
coolo, line 2: Please add back all copyright years I added personally for SUSE.

coolo (changes requested):

autotest.pm
okurz: No problem, I updated https://github.com/os-autoinst/os-autoinst/pull/1799 accordingly and will update this PR later.

consoles/localXvnc.pm
okurz: I updated #1799 accordingly and will update this PR later.

autotest.pm
okurz: Updated. I am not saying that the changes are non-trivial. Copying the commit description from the commit which now removed:
Commonly the copyright extends only if significant changes are applied to files
and even in that case renewal of the copyright is automatic based on the actual
content differences, not needing an explicit statement in the copyright header.
We usually extended the copyright with every new year as soon as changes have
been made. This has been seen as a nuisance by many and causes unnecessary
changes in the git log.

Other references like https://opensource.google/docs/copyright/ and
https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code
and https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ also
suggest to only include the original year of publication in the copyright
notice. In practice it would likely not matter if a 70 year period of copyright
is extended by just a handful of years anyway.

It is correct that gnu.org states that we should include subsequent years on changes:
> To update the list of year numbers, add each year in which you have made nontrivial changes to the package. (Here we assume you’re using a publicly accessible revision control server, so that every revision installed is also immediately and automatically published.)"

So technically I was proposing to not follow the GNU GPL recommendations but what I understood to be the current "legal practice".

Also interesting is the following suggestion from GNU GPL:
> When you add the new year, it is not required to keep track of which files have seen significant changes in the new year and which have not. It is recommended and simpler to add the new year to all files in the package, and be done with it for the rest of the year.

I understand this as: Use at best an automatic tool to update all files to the current year at the beginning of each year.

coolo: That's what a lot of GNU projects do, yes. I don't propose that though

bmwiedemann:
I'm not a friend of auto-updating all copyright years as it does not reflect if and when something copyright-worthy has been done.
The 70 years after death of author effectively means, they never expire, so we would not have to update it.
And even if it expired one day, the worst that happens is that it would enter the public domain, so someone can make a propietary version without attribution.
And we can still update the year as part of commits that do major updates to those files.

About redundant `(C)` symbols, I don't care.

coolo (commented):
I'm good now but I'm afraid to approve with all your auto merge magic :)

okurz:
> I'm good now but I'm afraid to approve with all your auto merge magic :)

yeah because you are a member with power :)

okurz:
so, someone else wants to approve as coolo gives +1?

Martchus (approved):

Martchus:
I guess I've been adding the label to late (after approving and reading @coolo 's comment).

okurz:
all good

PR #1802, opened 2021-10-04
Set offset in serial output buffer

mloviska:
Test case [gdb](https://progress.opensuse.org/issues/93901) fails regularly outside serial console.
The issue happens only when 2 consequent `wait_serial` calls appear and
the first call contains also the string that should have been match by
the following one.

This should be handled by setting offset on the serial file after each
match.

- VRs:
    - [sle-15-SP3-JeOS-for-kvm-and-xen-QR-x86_64-Build3.3.5-jeos-base+sdk+desktop@svirt-xen-hvm](http://kepler.suse.cz/tests/7282#step/gdb/1)
    - [sle-15-SP3-JeOS-for-VMware-QR-x86_64-Build3.3.5-jeos-base+sdk+desktop@svirt-vmware65](http://kepler.suse.cz/tests/7281#step/gdb/1)
    - [sle-15-SP3-JeOS-for-MS-HyperV-QR-x86_64-Build3.3.5-jeos-base+sdk+desktop@svirt-hyperv-uefi](http://kepler.suse.cz/tests/7284#step/gdb/1)

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -29,6 +29,7 @@ use Scalar::Util 'looks_like_number';
 use Mojo::File 'path';
 use OpenQA::Exceptions;
 use Time::Seconds;
+use English -no_match_vars;
 
 # should be a singleton - and only useful in backend process
 our $backend;
@@ -838,7 +839,7 @@ previous test's serial output. Call this before you start doing something new
 =cut
 
 sub set_serial_offset {
-    my ($self, $args) = @_;
+    my ($self, $keep) = @_;
 
     $self->{serial_offset} = -s $self->{serialfile};
     return $self->{serial_offset};
@@ -889,20 +890,27 @@ sub wait_serial {
     }
 
     $regexp = [$regexp] if ref $regexp ne 'ARRAY';
-    my $initial_time = time;
+    my $initial_time   = time;
+    my $current_offset = $self->{serial_offset};
     while (time < $initial_time + $timeout) {
         $str = $self->serial_text();
         for my $r (@$regexp) {
-            $matched = ref $r eq 'Regexp' ? $str =~ $r : $str =~ m/$r/;
-            if ($matched) {
-                $regexp = "$r";
+            if (!$args->{no_regex} && $str =~ m/$r/) {
+                $current_offset += $LAST_MATCH_END[0];
+                $str     = substr($str, 0, $LAST_MATCH_END[0]);
+                $matched = 1;
+                last;
+            } elsif ($args->{no_regex} && (my $i = index($str, $r)) >= 0) {
+                $current_offset += length($r) + $i;
+                $str     = substr($str, 0, $i + length($r));
+                $matched = 1;
                 last;
             }
         }
         last if ($matched);
         $self->run_capture_loop(1);
     }
-    $self->set_serial_offset();
+    $self->{serial_offset} = $current_offset;
     return {matched => $matched, string => $str};
 }
 
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -831,15 +831,15 @@ sub proxy_console_call {
     return $wrapped_result;
 }
 
-=head2 set_serial_offset
+=head2 clear_serial_buffer
 
 Determines the starting offset within the serial file - so that we do not check the
 previous test's serial output. Call this before you start doing something new
 
 =cut
 
-sub set_serial_offset {
-    my ($self, $keep) = @_;
+sub clear_serial_buffer {
+    my ($self) = @_;
 
     $self->{serial_offset} = -s $self->{serialfile};
     return $self->{serial_offset};
@@ -848,7 +848,7 @@ sub set_serial_offset {
 
 =head2 serial_text
 
-Returns the output on the serial device since the last call to set_serial_offset
+Returns the output on the serial device since the last call to clear_serial_buffer
 
 =cut
 
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -259,7 +259,7 @@ sub _handle_command_set_current_test {
     my ($self, $response) = @_;
 
     # Note: It is unclear why we call set_serial_offset here
-    $bmwqemu::backend->_send_json({cmd => 'set_serial_offset'});
+    $bmwqemu::backend->_send_json({cmd => 'clear_serial_buffer'});
 
     my ($test_name, $full_test_name) = ($response->{name}, $response->{full_name});
     my $pause_test_name = $self->pause_test_name;
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -13,7 +13,7 @@ use Mojo::File 'path';
 use Mojo::JSON 'decode_json';
 use backend::baseclass;
 use POSIX 'tzset';
-use Mojo::File 'tempdir';
+use Mojo::File qw(tempdir path);
 use Mojo::Util qw(scope_guard);
 use bmwqemu ();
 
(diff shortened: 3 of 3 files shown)


mloviska:
@Vogtinator @jlausuch @cfconrad

Martchus (changes requested):
I guess this goto-like control flow deserves at least some unit tests.

backend/baseclass.pm
```diff
@@ -838,9 +839,11 @@ previous test's serial output. Call this before you start doing something new
 =cut
 
 sub set_serial_offset {
-    my ($self, $args) = @_;
+    my ($self, $keep) = @_;
```
Vogtinator, line 842: With the method's name, I'd expect that the (only) argument is the offset, not some weird "keep" value which is substracted for an (undocumented) reason

backend/baseclass.pm
```diff
@@ -889,20 +892,23 @@ sub wait_serial {
     }
 
     $regexp = [$regexp] if ref $regexp ne 'ARRAY';
-    my $initial_time = time;
-    while (time < $initial_time + $timeout) {
+    my $initial_time  = time;
+    my $bytes_to_keep = 0;
```
Vogtinator, line 896: It's possible that the file grows during this function, which would then result in a wrong offset.

I think it's simpler and more reliable to save the current offset here and add the match end to that instead, i.e.:

```
my $current_offset = $self->{serial_offset};
...
            if (!$args->{no_regex} && $str =~ m/$r/) {
                $current_offset += $LAST_MATCH_END[0];
                last MATCH;
            }
...
$self->{serial_offset} = $current_offset;
```

backend/baseclass.pm
jlausuch: Also, if you keep $self and you are calling this function `$self->set_serial_offset($bytes_to_keep);`... not sure how it will work, cause it stores the value of `$bytes_to_keep` in $self... 
So, either you really use %args and call the function like  `$self->set_serial_offset(keep => $bytes_to_keep);`
or you just  do `my ($keep) = @_;`

backend/baseclass.pm
mloviska: I have initially thought that `$self->{serial_offset}` should be handled within `set_serial_offset` as the code calls the method at the end anyway.

backend/baseclass.pm
```diff
@@ -889,20 +892,23 @@ sub wait_serial {
...
+  MATCH: while (time < $initial_time + $timeout) {
         $str = $self->serial_text();
         for my $r (@$regexp) {
-            $matched = ref $r eq 'Regexp' ? $str =~ $r : $str =~ m/$r/;
-            if ($matched) {
-                $regexp = "$r";
-                last;
+            if (!$args->{no_regex} && $str =~ m/$r/) {
+                $bytes_to_keep = length($str) - $LAST_MATCH_END[0];
```
Vogtinator, line 901: This no longer sets `$matched`

Vogtinator (changes requested):

backend/baseclass.pm
Vogtinator: I'd be in favor of dropping that method, but it's also used by `_handle_command_set_current_test`...

Maybe it could be renamed "clear_serial_buffer` or something like that.

Martchus:
Likely `23-baseclass.t` would be a good place to add tests. And yes, it might be that the function is so far only covered in test API tests.

mloviska:
> Likely `23-baseclass.t` would be a good place to add tests. And yes, it might be that the function is so far only covered in test API tests.

thanks! I will have a look.

Vogtinator:
Assuming that http://kepler.suse.cz/tests/7237#step/gdb/30 was with the current revision of this PR, there's a bug or just inconsistency with the output in the UI. The collected serial match is still the whole chunk (more than the match), but only the match is skipped in the buffer. Thus the not matched part is printed multiple times.

backend/baseclass.pm
```diff
@@ -889,20 +890,25 @@ sub wait_serial {
     }
 
     $regexp = [$regexp] if ref $regexp ne 'ARRAY';
-    my $initial_time = time;
-    while (time < $initial_time + $timeout) {
+    my $initial_time   = time;
+    my $current_offset = $self->{serial_offset};
+  MATCH: while (time < $initial_time + $timeout) {
```
okurz, line 895: I would prefer to not use goto-like statements. Can we find a better approach?

mloviska:
> Assuming that http://kepler.suse.cz/tests/7237#step/gdb/30 was with the current revision of this PR, there's a bug or just inconsistency with the output in the UI. The collected serial match is still the whole chunk (more than the match), but only the match is skipped in the buffer. Thus the not matched part is printed multiple times.

I think that is expected as the two following matches are checking the same chunk of text. The first one matches the beginning "GNU gdb". After successful match, it sets new value of offset and the second in looking for the gdb prompt. The output comes from one cli command invocation http://kepler.suse.cz/tests/7237#step/gdb/9.

Maybe I am overlooking something, but it feels it the similar as in serial [terminal case](https://openqa.suse.de/tests/7293116#step/gdb/10).

backend/baseclass.pm
mloviska: I can check `$matched && last` could work, too. Lemme try

backend/baseclass.pm
```diff
@@ -889,20 +890,25 @@ sub wait_serial {
...
-            $matched = ref $r eq 'Regexp' ? $str =~ $r : $str =~ m/$r/;
-            if ($matched) {
-                $regexp = "$r";
-                last;
+            if (!$args->{no_regex} && $str =~ m/$r/) {
+                $current_offset += $LAST_MATCH_END[0];
+                $matched = 1;
+                last MATCH;
+            }
+            if ($args->{no_regex} && (my $i = index($str, $r)) >= 0) {
```
cfconrad, line 903:
```suggestion
            } 
            elsif ($args->{no_regex} && (my $i = index($str, $r)) >= 0) {
```

Vogtinator:
> I think that is expected as the two following matches are checking the same chunk of text. The first one matches the beginning "GNU gdb". After successful match, it sets new value of offset and the second is looking for the gdb prompt.

This is how it should look: https://openqa.opensuse.org/tests/1952302#step/gdb/10

The issue is that `$str` needs to be chopped at the match end as well.

Vogtinator:
https://openqa.opensuse.org/tests/1952302#step/gdb/11 even shows a test bug... It checks for `(gdb)`, but as regex. So it matches plain `gdb` without parens.

mloviska:
> https://openqa.opensuse.org/tests/1952302#step/gdb/11 even shows a test bug... It checks for `(gdb)`, but as regex. So it matches plain `gdb` without parens.

Yes, I have noticed that :) I will address that in the followup PR

backend/baseclass.pm
```diff
@@ -889,20 +890,24 @@ sub wait_serial {
...
-                $regexp = "$r";
-                last;
+            if (!$args->{no_regex} && $str =~ m/$r/) {
+                $current_offset += $LAST_MATCH_END[0];
+                $matched = 1;
+            }
+            if ($args->{no_regex} && (my $i = index($str, $r)) >= 0) {
+                $current_offset += length($r) + $i;
+                $matched = 1;
             }
```
Vogtinator, line 905: Needs `$matched && last;` here as well

mloviska: Added

cfconrad (commented):
Nice that we start align the matching function with `consoles::serial_screen::read_until()`!
For the return value, maybe we should not return full `$str` but only the a substr with the lenght of `serial_offset` like we do https://github.com/os-autoinst/os-autoinst/blob/master/consoles/serial_screen.pm#L291

backend/baseclass.pm
mloviska: I had it initially like that however, I had no useful usage for `else`. Any ideas ? :)

cfconrad: that we do not run the `if` condition again if the first match. It's just eyecandy.

mloviska: yes, ofc it is better. I just prefer complete `if -> elsif -> else` syntax and not skipping the `else`. As I had no use for `else` I chose 2x `if`. :)

backend/baseclass.pm
```diff
@@ -889,20 +890,29 @@ sub wait_serial {
...
+                $str     = substr($str, 0, $LAST_MATCH_END[0]);
+                $matched = 1;
                 last;
+            } elsif ($args->{no_regex} && (my $i = index($str, $r)) >= 0) {
+                $current_offset += length($r) + $i;
+                $str     = substr($str, 0, $i + length($r));
+                $matched = 1;
+                last;
+            } else {
+                bmwqemu::diag("Given regex ($r) does not match!");
```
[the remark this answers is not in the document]
Vogtinator: I also don't think it's that helpful, because whether a expression matches or not is obvious anyway given the return value.

Vogtinator (approved):
LGTM!

`clear_serial_buffer` could be tested as well, by asserting that the matched string does not contain any previous content.

The `ref $r eq 'Regexp'` test is gone, but I don't know whether that is actually necessary...

Martchus (approved):

PR #1803, opened 2021-10-05
Allow running tidy on single files

foursixnine:
As a follow up to 3b277957a and os-autoinst/openQA/4216 add more
information to the xml metadata to avoid guessing where a remote worker
is running

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -94,7 +94,16 @@ test -e tools/tidy || exit 1
 
 cleanup
 
-[[ -f $filename ]] && perltidy --pro=.../.perltidyrc $filename || find-files | xargs perltidy --pro=.../.perltidyrc
+if [[ -n "$filename" ]]; then
+   if [[ -f "$filename" ]]; then
+       perltidy --pro=.../.perltidyrc "$filename"
+   else
+       echo "\"$filename\" is not a valid file, please provide only one file"
+       exit 22
+   fi
+else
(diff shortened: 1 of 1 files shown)


tools/tidy
```diff
@@ -69,6 +75,11 @@ fi
 
 find-files() {
     local files=()
+
+    if [[ -f $filename ]]; then
+        ls $filename
```
okurz, line 80: do you need to list the file instead of just `echo $filename`?

foursixnine: I’m a bit paranoid, either way should be ok.

Martchus: The parameter should be quoted and considering the `if` it makes more sense to use `echo`.

okurz: Actually I think the whole function `find_files` should only be called if no file parameter is supplied, e.g. change line R102 

```diff
- find-files | xargs perltidy --pro=.../.perltidyrc
+ [[ -f $filename ]] && perltidy --pro=.../.perltidyrc $filename || find-files | xargs perltidy --pro=.../.perltidyrc

okurz (changes requested):

tools/tidy
foursixnine: @okurz I like that! done.

okurz (approved):

okurz:
It's ok

PR #1804, opened 2021-10-06
qemu: fall back to HDDSIZEGB value, even for backing file

ggardet:


Martchus:
It would be great if the commit message had a few more details like why this is required.

ggardet:
> `Failed test 'Generate qemu-img command line for single existing UEFI disk'`. in `t/18-qemu.t`
> 
> I don't think we resize backing disk?

We do with `HDDSIZEGB_$i` only which is inconsistent.

ggardet:
> I might not understand something but it feels like you are changing the expected format (with/without "G" suffix) in a not well described commit that states something different. We must be more careful here. Also your change makes it incompatible with doc/backend_vars.asciidoc. Changing the behaviour for backing file might be ok but I don't think it's wise to change the expected variable value format

I did not change any behavior and `doc/backend_vars.asciidoc` is still valid. The `G` is just required for `qemu-img` to resize the image.

ggardet:
> It's more consistent but IMO not useful behaviour. There are only a handful of cases when it's actually desired to resize an already existing disk, and in those cases it's IMO better to do so explicitly.

This is the purpose of `HDDSIZEGB` and `HDDSIZEGB_$i`, right? If we do not need a resize, then we should not define this var.

ggardet:
> The main issue here is that `HDDSIZEGB` is defined by the machine in openQA, so pretty much all tests have it set by default.

IMO, this means we use a broken workflow. But I agree this should be fixed prior to the merge.
Should I pursue here, or should we just ignore this behavior? WDYT?

PR #1807, opened 2021-10-07
 testapi: Add die_on_timeout option to testapi::script::run()

cfconrad:
The testapi function `script_run()` is widely used to execute one command 
after the other, without checking for exit code. But it could also
happen that the command timed out and in such situation the
test-developer need to handle it separately.
To solve this, introduce a new option `die_on_timeout`. The default is now like
we have it currently **don't die**. But we will see a log message like:
```
[2021-11-22T14:47:47.739305+01:00] [warn] !!! testapi::script_run: DEPRECATED call of script_run() in tests/clemix/script_run.pm:16 add `die_on_timeout => ?` to avoid this warning
```
if the user doesn't specified `die_on_timeout=>(0|1)` explicit.

==== Typical miss usage of script_run() ====

Example of possible miss usage of `script_run()`:
```perl
script_run('command_a');
script_run('command_b');
```
if the first call just times out, it is very likely that the shell is
_not_ at the prompt for command_b. Thus something unexpected happen,
even if we don't care of the exit value from `command_a`.

Also widely used is:
```perl
if (script_run(...) != 0) {
  some_sane_error_handling();
}
```
But in case of timeout, this error handling isn't triggered and the test
just continue.

With this change, we also need to increase the API version specified in
OpenQA::Isotovideo::Interface::version .

== testapi.t: Add test for script_run() timeout behavior ==

Changed other `script_run()` calls to avoid unexpected warnings.

We add `quiet => 1` to the script_run() calls, because otherwise the test
`result (to create a new needl from) has been added` failed, as the
detail picture has the next number.

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 23;
+our $version = 24;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -8,9 +8,10 @@ use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '5';
 use Test::Mock::Time;
 use File::Temp;
-use Test::Output qw(combined_like stderr_like);
+use Test::Output qw(combined_like stderr_like stderr_unlike);
 use Test::Fatal;
 use Test::Warnings qw(:all :report_warnings);
+use Test::Exception;
 use Scalar::Util 'looks_like_number';
 
 use OpenQA::Isotovideo::Interface;
@@ -24,6 +25,7 @@ my $cmds;
 use Test::MockModule;
 my $mod = Test::MockModule->new('myjsonrpc');
 my $fake_exit = 0;
+my $fake_matched = 1;
 
 # define variables for 'fake_read_json'
 my $report_timeout_called = 0;
@@ -42,7 +44,7 @@ sub fake_read_json ($fd) {
     if ($cmd eq 'backend_wait_serial') {
         my $str = $lcmd->{regexp};
         $str =~ s,\\d\+(\\s\+\\S\+)?,$fake_exit,;
-        return {ret => {matched => 1, string => $str}};
+        return {ret => {matched => $fake_matched, string => $str}};
     }
     elsif ($cmd eq 'backend_select_console') {
         return {ret => {activated => 0}};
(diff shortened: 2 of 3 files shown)


jlausuch:
> To get killing working, I would recommend to use the `timeout` command like we do in `script_retry()`, see: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1495

That makes sense, thanks!

Vogtinator:
> > If this PR is actually killing the process that timed-out and returning back the prompt (without killing the job), I am 100% with this.

Is there a use case of intentionally letting a command time out? If not, failing the module and restoring a snapshot is likely the only way to get back to a sane state.

cfconrad:
@okurz I have changed the default and create now the deprecated message
```
[2021-11-22T14:47:47.739305+01:00] [warn] !!! testapi::script_run: DEPRECATED call of script_run() in tests/clemix/script_run.pm:16 add `die_on_timeout => ?` to avoid this warning
```
if `die_on_timeout` isn't explicit set.

Vogtinator:
I can't come up with a valid reason to use `script_run` while allowing (or even expecting) that the script times out.

So IMO it should be safe to fail `script_run` on timeout by default. It's not unlikely that it will uncover some test bugs though.

I don't think the `os-autoinst` API needs to strive for that level of "bug compatibility".

okurz (approved):
We can make it fail soon but let's give it a grave time of at least some days or weeks as the deprecation message is already in

Vogtinator:
> We can make it fail soon but let's give it a grave time of at least some days or weeks as the deprecation message is already in

I'll make a prediction:

1. All tests will spam the deprecation notice, but nobody cares because it's everywhere
2. Someone makes a huge PR to add `die_on_timeout` everywhere, with conflicts and all the other kinds of "fun" such a huge PR causes
3. Something might fail after merging that PR and gets fixed
4. The default of `die_on_timeout` gets flipped
5. `die_on_timeout => 1` gets removed everywhere, because it's the default
6. Nothing uses `die_on_timeout` anymore, and the parameter gets removed

So FWICT this new optional parameter will just cause extra work without any benefit.

cfconrad:
But exactly that is what I try to do, `6.` is the final goal, but we have `5.` which hopefully makes the tests a bit more stable.

Vogtinator:
> But exactly that is what I try to do, `6.` is the final goal, but we have `5.` which hopefully makes the tests a bit more stable.

Yes, but by making `die_on_timeout => 1` the new implicit default, we can skip all steps except for 3.

cfconrad:
looks like I was to scary

okurz:
Now I suggest, @cfconrad could you please directly create the PR for the follow-up steps, we will mark it as "not-ready" and wait some time before we merge?

jlausuch:
I agree with @Vogtinator. 
If we are strict and die by default, maybe some tests start to fail and people will complain, but the reason for those new failures is not this PR, the reason is a wrong usage of `script_run` which should be corrected on the test side.
I just hope that his prediction doesn't come true :)

okurz:
In hindsight seeing that the PR was open for more than a month I agree that we should have made the wrong behaviour fatal :) Anyway, I hope that cfconrad does that now in a follow-up PR which we can then even merge with less waiting time

PR #1809, opened 2021-10-08
distribution: Fix sporadic here-script input failure

cfconrad:
We encounter a stability problem with simple `script_output('w')` call.
1 out of ~50 calls failed, because the content of the script is
```
w\r
```
instead of
```
w\n
```

The actual root cause is unknown. But it seems that we encounter similar
problems as fixed with 87f895f3. The problem is, that a `wait_screen_change()`
call around the `type_string("\n")` didn't solve the problem, as the
changed screen where sometimes detected to early. During tests, we
encountered a minimal sleep of 800ms.
To not use a native `sleep` the decision was made to use
`wait_still_screen` functionality of `type_string`. This is a minimum
sleep, but guarded that the screen did not change within this
period.

The behavior was only seen on hyperv and vmware hosts. Thus we added a
`testapi::backend_get_wait_still_screen_on_here_doc_input()` function which
return the amount of seconds to wait for still screen, before start
typing the script.

The openqa variable `_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT` can be used to
overwrite this value.

Ticked: https://progress.opensuse.org/issues/60566

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -309,4 +309,17 @@ sub stop_serial_grab ($self, @) {
     return;
 }
 
+sub get_wait_still_screen_on_here_doc_input ($self, $args) {
+    # We encountered a sporadic error when type into the here-document in the
+    # distribution::script_output() function (poo#60566). This issue was only
+    # seen by svirt backends from VMM_FAMILY hyperv or vmware.
+    #
+    # With wait_still_screen we actually do a sleep, but the given duration is
+    # the minimum and will be extended till there is no change on the screen.
+    # by comparing the screen and checking that nothing else will write on it.
+    # So if the here-document input is really slow, we hope the wait_still_screen
+    # takes even longer.
+    ($bmwqemu::vars{VIRSH_VMM_FAMILY} // '') =~ qr/^hyperv|vmware$/ ? 1 : 0;
+}
+
 1;
--- a/distribution.pm
+++ b/distribution.pm
@@ -279,8 +279,9 @@ sub script_output {
         testapi::wait_serial("$marker-0-", quiet => $args{quiet});
     }
     elsif ($args{type_command}) {
-        my $cat = "cat - > $script_path;\n";
+        my $cat = "cat - > $script_path;";
         testapi::type_string($cat);
+        testapi::type_string("\n", wait_still_screen => testapi::backend_get_wait_still_screen_on_here_doc_input());
         testapi::type_string($script . "\n", timeout => $args{timeout});
         testapi::send_key('ctrl-d');
     }
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -35,6 +35,7 @@ PAUSE_AT;string;;Test module (name or fullname) to pause test execution at. To b
 PAUSE_ON_SCREEN_MISMATCH;boolean;0;Pause test execution on the next screen mismatch. Same notes as for `PAUSE_AT` apply.
 PAUSE_ON_NEXT_COMMAND;boolean;0;Pause test execution on the next test API command. Same notes as for `PAUSE_AT` apply.
 _QUIET_SCRIPT_CALLS;boolean;0;Add quiet flag to all the calls to script_run, script_output and validate_script_output. It will omit all the squares "wait_serial expected" on the Details view of the test case. This option might be useful for serial terminal tests.
+_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT;float;0;If this value is greater then 0, it is used by `wait_still_screen` before starting to write the script into the here document used in `testapi::script_output()` function (see: poo#60566). By default this depends on the backend.
 AUTOINST_URL_HOSTNAME;string;;hostname or IP address of host running the autoinst webserver endpoint, defaults to the local IP address within the qemu network for the qemu backend or the `WORKER_HOSTNAME` otherwise.
 UPLOAD_METER;boolean;0;Display curl progress meter in `upload_logs()` and `upload_assets()` test API functions.
 UPLOAD_MAX_MESSAGE_SIZE_GB;integer;20;Specifies the max. upload size in GiB for the test API functions `upload_logs()` and `upload_assets()` and the underlying command server API.
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -689,4 +689,14 @@ subtest 'Method consoles::sshVirtsh::add_disk()' => sub {
         }
     };
 };
+
+subtest 'get_wait_still_screen_on_here_doc_input' => sub {
+    set_var(VIRSH_VMM_FAMILY => 'hyperv');
+    is($svirt->get_wait_still_screen_on_here_doc_input({}) > 0, 1, 'wait_still_screen on here doc is set for hyperv');
+    set_var(VIRSH_VMM_FAMILY => 'vmware');
+    is($svirt->get_wait_still_screen_on_here_doc_input({}) > 0, 1, 'wait_still_screen on here doc is set for vmware');
+    set_var(VIRSH_VMM_FAMILY => 'kvm');
+    is($svirt->get_wait_still_screen_on_here_doc_input({}), 0, 'wait_still_screen on here doc is not set for kvm');
+};
+
 done_testing;
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -350,4 +350,6 @@ subtest check_select_rate => sub {
     };
 };
 
+is($baseclass->get_wait_still_screen_on_here_doc_input({}), 0, 'wait_still_screen on here doc is off by default!');
+
 done_testing;
--- a/testapi.pm
+++ b/testapi.pm
@@ -2335,4 +2335,17 @@ A typical call would look like:
 
 sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
+=head2 backend_get_wait_still_screen_on_here_doc_input 
+
+Function to query the backend if it has the known bug from
+https://progress.opensuse.org/issues/60566 which is that typing too fast into
+the here-document input can yield invalid script content.
+This function returns the value to be used by C<wait_still_screen> before 
+starting to write the script into the here document.
+=cut
+sub backend_get_wait_still_screen_on_here_doc_input {
+    state $ret;
+    $ret = query_isotovideo('backend_get_wait_still_screen_on_here_doc_input', {}) unless defined($ret);
+    return get_var(_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT => $ret);
+}
 1;
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -723,6 +723,8 @@ sub is_serial_terminal ($self, $args) {
     return {yesorno => $self->{current_console}->is_serial_terminal};
 }
 
+sub get_wait_still_screen_on_here_doc_input ($self, $args) { 0 }
+
 sub capture_screenshot ($self) {
     return unless $self->{current_screen};
 
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -83,7 +83,9 @@ sub fake_read_json ($fd) {
     elsif ($cmd eq 'backend_mouse_set') {
         return {ret => {x => 100, y => 100}};
     }
-    else {
+    elsif ($cmd eq 'backend_get_wait_still_screen_on_here_doc_input') {
+        return {ret => 0};
+    } else {
         note "mock method not implemented \$cmd: $cmd\n";
     }
     return {};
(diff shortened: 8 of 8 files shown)


okurz:
not exactly. You could use the "object oriented approach" and apply the screen wait within the affected backends or consoles. But maybe this is not really applicable for the backend in all cases but specific to the worker hosts? In that case I would additionally make that behaviour optional with a test variable that can be set within the worker config

distribution.pm
```diff
@@ -222,6 +222,17 @@ sub script_sudo {
...
+sub need_wait_screen_change_on_script_output
+{
+    if (testapi::check_var('BACKEND', 'svirt') &&
+        (testapi::check_var('VIRSH_VMM_FAMILY', 'hyperv') ||
+            testapi::check_var('VIRSH_VMM_FAMILY', 'vmware'))) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
```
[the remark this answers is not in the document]
cfconrad: @okurz plz check the latest version

testapi.pm
```diff
@@ -2335,4 +2335,15 @@ A typical call would look like:
 
 sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
+=head2 backend_need_extra_wait_on_script_output
+
+Function to query the backend, if it has the known bug from poo#60566, when
+typing to fast into the here-script intput, the script content becomes wrong.
+=cut
+sub backend_need_extra_wait_on_script_output
```
Martchus, line 2343:
```suggestion
sub backend_needs_extra_wait_on_script_output
```

testapi.pm
```diff
@@ -2335,4 +2335,15 @@ A typical call would look like:
 
 sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
+=head2 backend_need_extra_wait_on_script_output
+
+Function to query the backend, if it has the known bug from poo#60566, when
+typing to fast into the here-script intput, the script content becomes wrong.
```
Martchus, line 2341:
```suggestion
typing too fast into the here-script intput, the script content becomes wrong.
```

distribution.pm
```diff
@@ -279,8 +281,8 @@ sub script_output {
         testapi::wait_serial("$marker-0-", quiet => $args{quiet});
     }
     elsif ($args{type_command}) {
-        my $cat = "cat - > $script_path;\n";
-        testapi::type_string($cat);
+        testapi::type_string("cat - > $script_path;\n");
+        sleep 0.8 if testapi::backend_need_extra_wait_on_script_output();
```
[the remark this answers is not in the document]
Martchus: It is also weird to hard-code the delay here considering you've already put effort into making it dependent on the backend. Wouldn't it make more sense if the backend would return the required delay?

But as @okurz says, this workaround isn't ideal at all anyways.

Martchus (changes requested):

distribution.pm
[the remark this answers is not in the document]
cfconrad: I agree, we could get the sleep value from the backend as well :+1: 

The reason why `sleep` and not `wait_screen_change=>1` is this test: http://kepler.suse.cz/t8367
Where we already detected a screen change on the first loop and it failed. Not on the first attempt but after some time.
So I don't see a safe solution beside sleeping for some time.

backend/svirt.pm
```diff
@@ -309,4 +309,8 @@ sub stop_serial_grab ($self, @) {
     return;
 }
 
+sub need_extra_wait_on_script_output ($self, $args) {
+    check_var('VIRSH_VMM_FAMILY', 'hyperv') || check_var('VIRSH_VMM_FAMILY', 'vmware');
+}
```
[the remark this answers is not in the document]
cfconrad: ok

distribution.pm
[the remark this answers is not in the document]
Martchus: By the way, it would still make sense to make it configurable via a test variable (also when the value comes from the backend).

testapi.pm
cfconrad: I changed the name to `backend_get_sleep_on_here_script_intput()`

testapi.pm
cfconrad: thx

distribution.pm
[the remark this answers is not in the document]
cfconrad: ok...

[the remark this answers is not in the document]
cfconrad: > The reason why sleep and not wait_screen_change=>1 is this test: http://kepler.suse.cz/t8367
Where we already detected a screen change on the first loop and it failed. Not on the first attempt but after some time.
So I don't see a safe solution beside sleeping for some time.

@okurz still complaining about `sleep()`? plz see:

backend/svirt.pm
```diff
@@ -309,4 +309,8 @@ sub stop_serial_grab ($self, @) {
     return;
 }
 
+sub get_sleep_on_here_script_input ($self, $args) {
+    ($bmwqemu::vars{VIRSH_VMM_FAMILY} // '') =~ qr/^hyperv|vmware$/ ? 0.8 : 0;
```
okurz, line 313: I *still* consider this a bad idea. My statement about "sleep" statement is not handled in a better way through this approach. If you *still* want to go ahead please add a comment here explaining why this is necessary

t/22-svirt.t
```diff
@@ -689,4 +689,14 @@ subtest 'Method consoles::sshVirtsh::add_disk()' => sub {
         }
     };
 };
+
+subtest 'get_sleep_on_here_script_input' => sub {
+    set_var(VIRSH_VMM_FAMILY => 'hyperv');
+    is($svirt->get_sleep_on_here_script_input({}) > 0, 1, 'Sleep on here script is set for hyperv');
+    set_var(VIRSH_VMM_FAMILY => 'vmware');
+    is($svirt->get_sleep_on_here_script_input({}) > 0, 1, 'Sleep on here script is set for hyperv');
```
okurz, line 697: copy-paste error in description string. Needs to be s/hyperv/vmware/

t/23-baseclass.t
```diff
@@ -315,4 +315,8 @@ BdsDxe: starting Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x8,0x0)'}, '
     is_deeply($baseclass->wait_serial({%dargs, regexp => 'something wrong', timeout => 1, no_regex => 1}), {matched => 0, string => "\nWelcome to GRUB!\n"}, "Test string literal mismatch");
 };
 
+subtest 'get_sleep_on_here_script_input' => sub {
+    is($baseclass->get_sleep_on_here_script_input({}), 0, 'Sleep on here script is off by default!');
+};
+
```
okurz, line 321:
```suggestion
is $baseclass->get_sleep_on_here_script_input({}), 0, 'Sleep on here-script is off by default';
```

okurz (changes requested):

backend/svirt.pm
cfconrad: ok, I mean we could also use this `wait_still_screen` thing. Which is in the end a hidden sleep for 1s. I did a test last night containing the following path. And I didn't notice a error on 2800 iterations.

diff --git a/distribution.pm b/distribution.pm
index a15132a0..adbe4538 100644
--- a/distribution.pm
+++ b/distribution.pm
@@ -279,9 +279,11 @@ sub script_output {
         testapi::wait_serial("$marker-0-", quiet => $args{quiet});
     }
     elsif ($args{type_command}) {
-        my $cat = "cat - > $script_path;\n";
+        my $cat = "cat - > $script_path;";
         testapi::type_string($cat);
-        sleep testapi::backend_get_sleep_on_here_script_input();
+        #        sleep testapi::backend_get_sleep_on_here_script_input();
+        my %args = testapi::backend_get_sleep_on_here_script_input() ? (similarity_level => 30, wait_still_screen => 1) : ();
+        testapi::type_string("\n", %args);
         testapi::type_string($script . "\n", timeout => $args{timeout});
         testapi::send_key('ctrl-d');
     }
```

The stats of one `script_output('w')` call of `wait_still_screen=>1` code (http://kepler.suse.cz/t8981)
```
Loop: 100
Allover: 906.873669862747
Avg: 9.06101491212844
Current:9.09190988540649
```

Compared to the sleep variation of the PR: (http://kepler.suse.cz/tests/8963#step/clemix/302)
```
Loop: 100
Allover: 838.612976074219
Avg: 8.37837042331696
Current:8.38157916069031
```

Compared to the current master (http://kepler.suse.cz/tests/8932#step/clemix/302)
```
Loop: 100
Allover: 757.928626060486
Avg: 7.57161826133728
Current:7.56396198272705
```

As you can see, the sleep is actually the better performing, but for me it is more important to get this fix for hyperv and vmware then doing one or the other way of slow down the hit level in that area.

Comment (okurz) [reply backend/svirt.pm:313]: Yes, `wait_still_screen` is a hidden sleep. As I stated, if you *still* want to go ahead please add a comment here explaining why this is necessary if https://github.com/os-autoinst/os-autoinst/pull/1809#discussion_r754640715 does not help and you can't find a better solution that is not a sleep.

File: distribution.pm (hunk 4 above)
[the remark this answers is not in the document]
Comment (cfconrad) [reply distribution.pm]: continue here: https://github.com/os-autoinst/os-autoinst/pull/1809#discussion_r756802924

File: testapi.pm [hunk 9]
```diff
@@ -2335,4 +2335,15 @@ A typical call would look like:
 
 sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
+=head2 backend_get_sleep_on_here_script_input 
+
+Function to query the backend, if it has the known bug from poo#60566, when
+typing too fast into the here-script intput, the script content becomes wrong.
+This function return the needed sleep duration as float.
```
Martchus, line 2342:
```suggestion
This function returns the needed sleep duration as float.
```

cfconrad:
Changed to use `wait_still_screen`.
VR with last version: http://kepler.suse.cz/tests/9286/file/autoinst-log.txt

okurz:
please rebase after I merged your other PR :)

backend/svirt.pm
```diff
@@ -309,4 +309,16 @@ sub stop_serial_grab ($self, @) {
     return;
 }
 
+sub get_wait_still_screen_on_here_doc_input ($self, $args) {
+    # We encounterd a sporadic error when type into the here-document in the
+    # distribution::script_output() function (poo#60566). This issue was only
+    # seen by svirt backends from VMM_FAMILY hyperv or vmware.
+    #
+    # With wait_still_screen we actually do a sleep, but this sleep is guared
```
okurz, line 317: "guared"?

okurz (changes requested):
Tested manually with os-autoinst checked out from this branch and executing the openQA full stack test with that.

backend/svirt.pm
cfconrad: done

cfconrad:
ready from my side

testapi.pm
```diff
@@ -2335,4 +2335,16 @@ A typical call would look like:
 
 sub show_curl_progress_meter { get_var('UPLOAD_METER') ? "-o /dev/$serialdev " : '' }
 
+=head2 backend_get_wait_still_screen_on_here_doc_input 
+
+Function to query the backend, if it has the known bug from poo#60566, when
+typing too fast into the here-document input, the script content becomes wrong.
+This function returns value used with C<wait_still_screen>, before start writing
+the script into the here document.
```
okurz, line 2343:
```suggestion
Function to query the backend if it has the known bug from
https://progress.opensuse.org/issues/60566 which is that typing too fast into the
here-document input can yield invalid script content.
This function returns the value to be used by C<wait_still_screen> before starting to
write the script into the here document.
```

okurz (changes requested):

doc/backend_vars.asciidoc
```diff
@@ -35,6 +35,7 @@ PAUSE_AT;string;;Test module (name or fullname) to pause test execution at. To b
 PAUSE_ON_SCREEN_MISMATCH;boolean;0;Pause test execution on the next screen mismatch. Same notes as for `PAUSE_AT` apply.
 PAUSE_ON_NEXT_COMMAND;boolean;0;Pause test execution on the next test API command. Same notes as for `PAUSE_AT` apply.
 _QUIET_SCRIPT_CALLS;boolean;0;Add quiet flag to all the calls to script_run, script_output and validate_script_output. It will omit all the squares "wait_serial expected" on the Details view of the test case. This option might be useful for serial terminal tests.
+_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT;float;0;If this value is greater then 0, we will use the amount of seconds for wait_still_screen before start typing into the here document in "script_output()" function (see: poo#60566). By default this depends on the backend.
```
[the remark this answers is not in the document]
cfconrad: changed to 
```suggestion
_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT;float;0;If this value is greater then 0, it is used by `wait_still_screen` before starting to write the script into the here document used in `testapi::script_output()` function (see: poo#60566). By default this depends on the backend.
```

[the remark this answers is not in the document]
mloviska: > 

If the initial proposal is not good enough, I would prefer to go for this one as it is crystal clear what is the effect of setting the variable :)

okurz (approved):
CI failed due to https://progress.opensuse.org/issues/103422, rest looks fine

PR #1810, opened 2021-10-08
.mergify.yml: Make rules stricter requiring patch coverage

okurz:
https://progress.opensuse.org/issues/98448


Martchus:
Could be. I needed to check what image is used here.

okurz:
switching over to a branch from my fork: https://github.com/os-autoinst/os-autoinst/pull/1815

PR #1815, opened 2021-10-10
.mergify.yml: Make rules stricter requiring patch coverage

okurz:
https://progress.opensuse.org/issues/98448


okurz:
Now checks passed (compared to #1810)

okurz:
why did you mark this as not-ready? Should we not merge this?

PR #1816, opened 2021-10-11
Use proper perl "signatures" in OpenQA::Isotovideo

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.


okurz:
> I suppose this is _likely_ ok but hard to tell. You could run the openQA fullstack and developer mode test against this to be more sure.

Good idea. I successfully ran "t/full-stack.t t/33-developer_mode.t"

PR #1822, opened 2021-10-14
Exclude all currently generated files in git

okurz:


okurz:
Yes I would prefer to have tests clean up after themselves, same as you do. As soon as we are there we can remove the files again from .gitignore . Checking the git status at the end of each test result is a good idea.

PR #1823, opened 2021-10-14
t: Cleanup temporary files

perlpunk:
With this, there are no more untracked files after running the tests.
I checked with a module that does a `git status` at the end of every test.
Will provide a PR for that soon.


okurz:
Sounds good. @perlpunk can you rebase on https://github.com/os-autoinst/os-autoinst/pull/1826#pullrequestreview-781868232 ?

PR #1828, opened 2021-10-18
myjsonrpc: Prevent error 'Can't use an undefined value as a symbol reference'

okurz:
Can't use an undefined value as a symbol reference

See https://progress.opensuse.org/issues/98952

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -37,12 +37,10 @@ sub send_json ($to_fd, $cmd) {
     }
     $json .= "\n";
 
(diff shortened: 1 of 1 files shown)


okurz:
> Won't this simply delay the problem by making it even harder to distinguish successful send_json calls from unsuccessful ones?

"unsuccessful" meaning without a valid file descriptor, yes. What we could do is the other way and turn the warning into a fatal error by catching the invalid file description and bailing with a stack trace so that we know the caller and prevent the incorrect calls in the first place?

PR #1833, opened 2021-10-22
Add coverage check for t/ folder

okurz:
* Integrate coverage check as part of CI pipeline
* tools: Add coverage check ensuring 100% coverage in t/

### DIFF
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -135,6 +135,12 @@ if (COVER_PATH AND PROVE_PATH)
     )
     add_dependencies(coverage symlinks)
     add_custom_target(
+        check-coverage-t
+        COMMENT "Check coverage in t/ folder of all functional tests"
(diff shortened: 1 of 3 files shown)


cmake/test-targets.cmake
```diff
@@ -156,7 +162,7 @@ if (COVER_PATH AND PROVE_PATH)
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
     add_dependencies(coverage-codecovbash symlinks)
-
+    add_custom_target(coverage-ci)
```
Martchus, line 165: This is still a draft but adding a custom target without dependencies won't do much.

okurz:
I'd say https://github.com/os-autoinst/os-autoinst/commit/0da7fd6ce8a5b2729bf32cd1ef56a421a67995ca is enough

PR #1834, opened 2021-10-22
qemu: Automatically create swtpm device

kalikiana:
- Create device if QEMUTPM is set
- Recommend swtpm packages

See: https://progress.opensuse.org/issues/101015

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -143,6 +143,15 @@ Requires:       qemu-tools
 Convenience package providing os-autoinst and qemu-x86 dependencies.
 %endif
 
+%package swtpm
+Summary:        Convenience package providing os-autoinst+swtpm
+Group:          Development/Tools/Other
+Requires:       os-autoinst
+Requires:       swtpm
+
+%description swtpm
+Convenience package providing os-autoinst and swtpm dependencies.
+
 %package s390-deps
 Summary:        Convenience package providing os-autoinst + s390 worker jumphost deps
 Group:          Development/Tools/Other
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -512,6 +512,33 @@ sub qemu_params_ofw ($self) {
     return 1;
 }
 
+sub setup_tpm ($self, $arch) {
+    my $vars = \%bmwqemu::vars;
+    return unless ($vars->{QEMUTPM});
+    my $tpmn = $vars->{QEMUTPM} eq 'instance' ? $vars->{WORKER_INSTANCE} : $vars->{QEMUTPM};
+    my $vmpath = "/tmp/mytpm$tpmn";
+    mkdir $vmpath unless -d $vmpath;
+    my $vmsock = "$vmpath/swtpm-sock";
+    unless (-e $vmsock) {
+        my @args = ('swtpm', 'socket', '--tpmstate', "dir=$vmpath", '--ctrl', "type=unixio,path=$vmsock", '--log', 'level=20', '-d');
+        push @args, '--tpm2' if (($vars->{QEMUTPM_VER} // '2.0') == '2.0');
+        runcmd(@args);
+    }
+    sp('chardev', "socket,id=chrtpm,path=$vmsock");
+    sp('tpmdev', 'emulator,id=tpm0,chardev=chrtpm');
+    if ($arch eq 'aarch64') {
+        sp('device', 'tpm-tis-device,tpmdev=tpm0');
+    }
+    elsif ($arch eq 'ppc64le') {
+        sp('device', 'tpm-spapr,tpmdev=tpm0');
+        sp('device', 'spapr-vscsi,id=scsi9,reg=0x00002000');
+    }
+    else {
+        # x86_64
+        sp('device', 'tpm-tis,tpmdev=tpm0');
+    }
+}
+
 sub start_qemu ($self) {
     my $vars = \%bmwqemu::vars;
 
@@ -801,22 +828,7 @@ sub start_qemu ($self) {
             sp('append', "dhcp && sanhook iscsi:$vars->{WORKER_HOSTNAME}::3260:1:$vars->{NBF}", no_quotes => 1);
         }
 
-        if ($vars->{QEMUTPM}) {
-            my $tpmn = $vars->{QEMUTPM} eq 'instance' ? $vars->{WORKER_INSTANCE} : $vars->{QEMUTPM};
-            sp('chardev', "socket,id=chrtpm,path=/tmp/mytpm$tpmn/swtpm-sock");
-            sp('tpmdev', 'emulator,id=tpm0,chardev=chrtpm');
-            if ($arch eq 'aarch64') {
-                sp('device', 'tpm-tis-device,tpmdev=tpm0');
-            }
-            elsif ($arch eq 'ppc64le') {
-                sp('device', 'tpm-spapr,tpmdev=tpm0');
-                sp('device', 'spapr-vscsi,id=scsi9,reg=0x00002000');
-            }
-            else {
-                # x86_64
-                sp('device', 'tpm-tis,tpmdev=tpm0');
-            }
-        }
+        $self->setup_tpm($arch);
 
         my @boot_args;
         # Enable boot menu for aarch64 workaround, see bsc#1022064 for details
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -122,7 +122,8 @@ QEMUMACHINE;see qemu -machine ?;undef;Machine and chipset to emulate
 QEMUPORT;integer;20002 + worker instance * 10;Port on which QEMU monitor should listen
 QEMURAM;integer;1024;Size of RAM of VM in MiB
 QEMUTHREADS;integer;0;Number of cpu threads used by VM
-QEMUTPM;'instance' or appropriate value for local swtpm config;undef;Configure VM to use a TPM emulator device, with appropriate args for the arch. sysadmin is responsible for running swtpm with a socket at /tmp/mytpmX, where X is the value of QEMUTPM or the worker instance number if QEMUTPM is set to 'instance'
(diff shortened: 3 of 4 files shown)


kalikiana:
FYI expecting one test failure for the "default" case. I'm guessing I'm missing something obvious but I can't make much sense of the output and I don't see the issue in my code.

Martchus (commented):
I suppose it would work but have you conducted a complete test run?

okurz (approved):

kalikiana:
> FYI expecting one test failure for the "default" case. I'm guessing I'm missing something obvious but I can't make much sense of the output and I don't see the issue in my code.

Seems like that was artifacts from debug code I'd forgotten to revert. However I did forget to commit the `update-deps` changes.
Not sure what's causing the OBS error, though 🤔 

    error: /home/abuild/rpmbuild/SOURCES/os-autoinst.spec:154: Got a %endif with no %if

> I suppose it would work but have you conducted a complete test run?

FYI the test here runs qemu and swtpm, although it doesn't examine the device further. I was going to run the test from [poo#101015](https://progress.opensuse.org/issues/101015) but that deleted, and I couldn't find existing tests relying on swtpm. Trying to get that from Richard.

dist/rpm/os-autoinst.spec
```diff
@@ -131,6 +131,17 @@ Requires:       qemu-tools
...
+%package swtpm
+Summary:        Convenience package providing os-autoinst+swtpm
+Group:          Development/Tools/Other
+Requires:       os-autoinst
+Requires:       libtpms0
+Requires:       swtpm
+
+%description swtpm
+Convenience package providing os-autoinst and swtpm dependencies.
+%endif
```
perlpunk, line 143: Where's the `%if` belonging to that new `%endif`?

okurz:
https://build.opensuse.org/package/show/devel:openQA:TestGithub:OPR-1834/os-autoinst shows that swtpm can not be resolved so this won't work. Do we actually need swtpm for testing? I would prefer if we don't need it at all. If you are just checking the log for what *would* be called I guess we should be ok :)

kalikiana:
> https://build.opensuse.org/package/show/devel:openQA:TestGithub:OPR-1834/os-autoinst shows that swtpm can not be resolved so this won't work. Do we actually need swtpm for testing? I would prefer if we don't need it at all. If you are just checking the log for what _would_ be called I guess we should be ok :)

I think it's rather good to know that it works. And we run a real qemu already

kalikiana:
I was confused why `setup_tpm` needed the re-defined `sp` and `qemu_params_ofw` doesn't. After fixing the deprecation message for `nowait` the test passes w/o it.

okurz (changes requested):
Please rephrase the git commit message subject line to answer the "Why", not the "What". Consider reading the changelog on openqa.opensuse.org/changelog which only shows the git commit message subject line. Think about what you would want to read as a user.

Also, what changes did you apply regarding https://github.com/os-autoinst/os-autoinst/pull/1834#issuecomment-964504795 ?

kalikiana:
> Also, what changes did you apply regarding [#1834 (comment)](https://github.com/os-autoinst/os-autoinst/pull/1834#issuecomment-964504795) ?

I made the dependency Tumbleweed-specific similar to what we do with e.g. opencv. We don't run the test on OBS either way.

dist/rpm/os-autoinst.spec
```diff
@@ -28,6 +28,7 @@ Source0:        %{name}-%{version}.tar.xz
 %if 0%{?suse_version} > 1500
 # openSUSE Tumbleweed
 %define opencv_require pkgconfig(opencv4)
+BuildRequires:  swtpm
```
okurz, line 31: This makes it look like only on Tumbleweed we would actually require the package and we don't need the package anywhere else while still delivering the same functionality which is simply not true. I think we can do better than this. I am thinking, seeing on https://build.opensuse.org/search?search_text=swtpm that swtpm is available in both Tumbleweed as well as SLE15-SP3 aka. Leap 15.3 I see the following options:
1. simply hard-require the package and drop support for Leap 15.2 and others that don't provide swtpm
2. same as 1 but we try to build swtpm ourselves, e.g. in devel:openQA:Leap:15.2 . This might be tricky however as also in https://build.opensuse.org/package/show/security/swtpm the package is unresolvable on Leap 15.2
3. As suggested in before don't require it and rewrite the test to not rely on the package itself but mock the dependency and keep the convenience package only
4. Turn the "convenience" package into a proper package with the swtpm only provided in that package, e.g. by extracting the functionality into a dedicated child-backend like "qemu-swtpm" and deliver the according file(s) only in the according package which requires swtpm

I still prefer 3., next to 1., but let's others have their opinion heard as well

okurz (changes requested):

dist/rpm/os-autoinst.spec
kalikiana: This is only relevant for tests.

kalikiana: I assume `BuildRequires` relates to the build host. We also uses that for `%test_requires`

okurz: I know that. But I don't like to re-introduce special test-dependency handling into the specfile. We have dependencies.yaml so we should use it

kalikiana: I don't see how `dependencies.yaml` can be used here, and I don't see how any of your suggestions helps with what I intended to do. That's why I thought you didn't understand it.

kalikiana: Tests seem to pass unmodified without `swtpm` installed so I'll leave it at that.

kalikiana:
> > I suppose it would work but have you conducted a complete test run?
> 
> FYI the test here runs qemu and swtpm, although it doesn't examine the device further. I was going to run the test from [poo#101015](https://progress.opensuse.org/issues/101015) but that deleted, and I couldn't find existing tests relying on swtpm. Trying to get that from Richard.

    openqa-clone-job --from http://openqa.suse.de --skip-deps --host http://openqa.suse.de 7619588 --skip-download _GROUP=0 CASEDIR="https://github.com/rfan1/os-autoinst-distri-opensuse.git#swtpm_test_only"
    Created job #7691337: sle-15-SP4-Online-x86_64-Build61.1-create_swtpm_hdd@64bit -> http://openqa.suse.de/t7691337

Unfortunately [it fails like so in the`build_hdd` module](https://openqa.suse.de/tests/7691337#step/build_hdd/14):

    fpmS7
    /tmp/scriptfpmS7.sh: line 1: ll: command not found

rfan1:
> > > I suppose it would work but have you conducted a complete test run?
> > 
> > 
> > FYI the test here runs qemu and swtpm, although it doesn't examine the device further. I was going to run the test from [poo#101015](https://progress.opensuse.org/issues/101015) but that deleted, and I couldn't find existing tests relying on swtpm. Trying to get that from Richard.
> 
> ```
> openqa-clone-job --from http://openqa.suse.de --skip-deps --host http://openqa.suse.de 7619588 --skip-download _GROUP=0 CASEDIR="https://github.com/rfan1/os-autoinst-distri-opensuse.git#swtpm_test_only"
> Created job #7691337: sle-15-SP4-Online-x86_64-Build61.1-create_swtpm_hdd@64bit -> http://openqa.suse.de/t7691337
> ```
> 
> Unfortunately [it fails like so in the`build_hdd` module](https://openqa.suse.de/tests/7691337#step/build_hdd/14):
> 
> ```
> fpmS7
> /tmp/scriptfpmS7.sh: line 1: ll: command not found
> ```
> 
> @rfan1 Can you confirm if this is the right reproducer?

No problem, let me revise my code and see.

rfan1:
> > > > I suppose it would work but have you conducted a complete test run?
> > > 
> > > 
> > > FYI the test here runs qemu and swtpm, although it doesn't examine the device further. I was going to run the test from [poo#101015](https://progress.opensuse.org/issues/101015) but that deleted, and I couldn't find existing tests relying on swtpm. Trying to get that from Richard.
> > 
> > 
> > ```
> > openqa-clone-job --from http://openqa.suse.de --skip-deps --host http://openqa.suse.de 7619588 --skip-download _GROUP=0 CASEDIR="https://github.com/rfan1/os-autoinst-distri-opensuse.git#swtpm_test_only"
> > Created job #7691337: sle-15-SP4-Online-x86_64-Build61.1-create_swtpm_hdd@64bit -> http://openqa.suse.de/t7691337
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Unfortunately [it fails like so in the`build_hdd` module](https://openqa.suse.de/tests/7691337#step/build_hdd/14):
> > ```
> > fpmS7
> > /tmp/scriptfpmS7.sh: line 1: ll: command not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @rfan1 Can you confirm if this is the right reproducer?
> 
> No problem, let me revise my code and see.

@kalikiana you can start a new job with the below one and add needed "swtpm" configuration:  :)

`openqa-clone-job --from http://openqa.suse.de --skip-deps --host http://openqa.suse.de 7656835 --skip-download _GROUP=0 CASEDIR="https://github.com/rfan1/os-autoinst-distri-opensuse.git#swtpm_test_only"`

kalikiana:
> openqa-clone-job --from http://openqa.suse.de --skip-deps --host http://openqa.suse.de 7656835 --skip-download _GROUP=0 CASEDIR="https://github.com/rfan1/os-autoinst-distri-opensuse.git#swtpm_test_only"

That seems to work, see https://progress.opensuse.org/issues/101015#note-15

kalikiana:
The test is not fully covered because of the `if ($arch eq 'aarch64') {` handling... this is not new code, though, I moved it into a new function, so I wonder why it wasn't covered before 🤔

okurz (approved):
> so I wonder why [this code] wasn't covered before thinking

Because automated tests did not exist before 2021 maybe ;P

Martchus (approved):

Martchus:
It is just that it looks like this branch is actually covered. Maybe the coverage (of the forked process) isn't collected.

okurz:
@kalikiana why did you merge now? I am ok if you merge and state why you want to overrule checks. IMHO would have been a good opportunity to increase test coverage beyond the patch coverage check limit

PR #1842, opened 2021-11-05
Don't include the previous tests fail hook output in the search for issues

Vogtinator:
After the main test code runs, search_for_expected_serial_failures looks
through the generated serial output for certain patterns. In the case that the
test failed, the post_fail_hook runs after that, which might also produce output
on the serial console. Before this commit, this was not skipped and so the call
to search_for_expected_serial_failures on the next module sees it.

This can even have a cascading effect if there's a OOM message in the log files,
if it was missed in an earlier module. Then the post_fail_hook would print that
to the serial console and thus cause the next module to fail, which in turn
prints the message again and the cycle repeats.

See also https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13610
Related ticket: https://progress.opensuse.org/issues/81382

Before: http://10.160.67.86/tests/1122
After: http://10.160.67.86/tests/1127

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -320,6 +320,10 @@ sub run_post_fail {
         eval { $self->post_fail_hook; };
         bmwqemu::diag("post_fail_hook failed: $@") if $@;
         $self->{post_fail_hook_running} = 0;
+
+        # There might be more messages on serial now.
+        # Read them now to not stumble upon them in the next module.
+        $self->get_new_serial_output();
     }
     $self->fail_if_running();
     die $msg . "\n";
@@ -669,11 +673,13 @@ sub search_for_expected_serial_failures {
     }
 }
 
-sub get_serial_output_json {
+sub get_new_serial_output {
     my ($self) = @_;
 
     myjsonrpc::send_json($autotest::isotovideo, {cmd => 'read_serial', position => $serial_file_pos});
-    return myjsonrpc::read_json($autotest::isotovideo);
+    my $json = myjsonrpc::read_json($autotest::isotovideo);
+    $serial_file_pos = $json->{position};
+    return $json->{serial};
 }
 
 sub parse_serial_output_qemu {
@@ -681,11 +687,11 @@ sub parse_serial_output_qemu {
     # serial failures defined in distri (test can override them)
     my $failures = $self->{serial_failures};
 
-    my $json = $self->get_serial_output_json($serial_file_pos);
+    my $serial = $self->get_new_serial_output();
     my $die = 0;
     my %regexp_matched;
     # loop line by line
-    for my $line (split(/^/, $json->{serial})) {
+    for my $line (split(/^/, $serial)) {
         chomp $line;
         for my $regexp_table (@{$failures}) {
             my $regexp = $regexp_table->{pattern};
@@ -713,7 +719,6 @@ sub parse_serial_output_qemu {
             }
         }
     }
-    $serial_file_pos = $json->{position};
     die "Got serial hard failure" if $die;
     return;
 }
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -24,6 +24,32 @@ use needle;
 # define 'write_with_thumbnail' to fake image
 sub write_with_thumbnail { }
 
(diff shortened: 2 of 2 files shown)


basetest.pm
```diff
@@ -320,6 +320,11 @@ sub run_post_fail {
         eval { $self->post_fail_hook; };
         bmwqemu::diag("post_fail_hook failed: $@") if $@;
         $self->{post_fail_hook_running} = 0;
+
+        # There might be more messages on serial now.
+        # Update the offset to not stumble upon them in the next module.
+        my $json = $self->get_serial_output_json($serial_file_pos);
+        $serial_file_pos = $json->{position};
```
Martchus, line 327: It doesn't look like we're ever calling `get_serial_output_json` without this assignment. So maybe move `$serial_file_pos = $json->{position};` into that function?

Martchus (commented):
I suppose that makes sense. I have only one remark and also don't approve immediately to give other reviewers the chance to have a look (before mergify would merge it automatically).

basetest.pm
Vogtinator: I changed `get_serial_output_json` a bit, how's the new approach? VR is updated.

Martchus: Looks good.

okurz: https://github.com/os-autoinst/os-autoinst/runs/4117429312?check_suite_focus=true#step:3:581 says "basetest::get_serial_output_json does not exist! at ./17-basetest.t line 69."

Vogtinator: Fixed

Martchus (approved):

Martchus:
It would be nice to have test coverage for the function, though.

Vogtinator:
> It would be nice to have test coverage for the function, though.

Yep. Unfortunately that requires mocking the backend function, which would have more logic than the function being tested...

Vogtinator:
Ping

okurz:
In the end we still want to have better test coverage. If you are motivated to do that, this would be great, if not, we will merge without and look into adding test coverage afterwards.

Vogtinator:
> In the end we still want to have better test coverage. If you are motivated to do that, this would be great, if not, we will merge without and look into adding test coverage afterwards.

I'm motivated, but it needs much more time (and/or perl expertise) than I have.

If you have an example which mocks a backend method with a non-trivial function, that would simplify it quite a bit.

okurz:
I think t/17-basetest.t itself is a good example. Maybe also https://github.com/os-autoinst/os-autoinst/blob/master/t/29-backend-ipmi.t and https://github.com/os-autoinst/os-autoinst/blob/master/t/29-backend-generalhw.t . In general I prefer trivial functions over non-trivial so maybe slightly refactor to make mocking easier :)

Vogtinator:
> I think t/17-basetest.t itself is a good example. Maybe also https://github.com/os-autoinst/os-autoinst/blob/master/t/29-backend-ipmi.t and https://github.com/os-autoinst/os-autoinst/blob/master/t/29-backend-generalhw.t . In general I prefer trivial functions over non-trivial so maybe slightly refactor to make mocking easier :)

The hard part is that to test this PR, it's necessary to mock at the `myjsonrpc` or `backend` layers. It would be easy to mock `get_new_serial_output`, but then the code of this PR wouldn't be used during the test.

okurz:
Mocking of read_json is already done in t/03-testapi.t so maybe that helps as reference

mloviska:
[The cascading effect](http://kepler.suse.cz/tests/8877#step/rails/31). If there are no objections, could we merge the PR as it affects all subsequent test cases ?

Vogtinator:
> Mocking of read_json is already done in t/03-testapi.t so maybe that helps as reference

That does indeed help.

I created a new unit test specifically for `get_new_serial_output`, see the second commit. The complexity of the test code is still more than the function which it tests (and also needed more attempts to get it working!), but it does work.

Vogtinator:
@okurz @Martchus Is it intentional that the approval is "sticky" even across major changes like this, i.e. mergify merges those even though they weren't actually reviewed?

Martchus:
I'd say it is deemed questionable but also accepted for now.

PR #1844, opened 2021-11-08
backend::qemu: Fix bug that qemu alive status was never tested

okurz:


Martchus (approved):

backend/qemu.pm
```diff
@@ -953,7 +953,7 @@ sub start_qemu {
         $self->select_console({testapi_console => 'sut'});
     }
     catch {
-        if (!raw_alive) {
+        if (!$self->raw_alive) {
```
kalikiana, line 956: The commit says `qemu alive status was never tested` but I don't see any new tests here

kalikiana (changes requested):

backend/qemu.pm
okurz: See? This is why I had this commit as part of my signatures PR as the problem was only showing up from there. Calling `raw_alive` instead of `$self->raw_alive` means that there is an actual change of parameters and as now `$self` is required in https://github.com/os-autoinst/os-autoinst/pull/1790 as first parameter we have that covered by tests. You still have the choice to approve this PR or #1790 with the change included in the commit list

kalikiana: Discussed briefly. I questioned that the code is used at all, since we don't see this fail and we have no ticket about this being broken so this might be dead code. @Martchus suggests the failure with signatures tells us that the code path works. @okurz didn't see a failure even inserting a `die` in the code path and running fullstack tests.

Martchus (commented):
`select_console` already catches exceptions so this code is indeed not very useful. You can also remove `use Try::Tiny;` in `backend/qemu.pm`.

okurz:
> You can also remove use Try::Tiny; in backend/qemu.pm.

done

Martchus (approved):

kalikiana (approved):
Very nice! Thank you for going the extra steps to clean it up!

PR #1847, opened 2021-11-09
dist: Do not build for i586

DimStar77:
The resulting binaries are not installable, as some dependencies
(e.g. ShellCheck) are not being built for i586.


Martchus (approved):
Likely we really don't need it at this point anymore. (Note that shellcheck should generally be considered an optional dependency, though.)

okurz (approved):

PR #1848, opened 2021-11-09
Declare likely obsolete backend::pvm as deprecated

okurz:
This is an alternative to https://github.com/os-autoinst/os-autoinst/pull/1790 which includes all changes as well. The only difference is in missing signatures. This PR will cause a conflict with https://github.com/os-autoinst/os-autoinst/pull/1790 on the signatures so the decision should be if this needs to be merged first or if instead https://github.com/os-autoinst/os-autoinst/pull/1790 should be used instead.

During our work on https://progress.opensuse.org/issues/99663 we
encountered that we only covered backend::spvm with testing, not
backend::pvm. Trying to develop a test it feels more and more like
backend::pvm is unused. A select count(*) from job_settings where
key='BACKEND' and value = 'pvm'; reveals no hits for both o3 and osd and
the git history also looks like everyone for PowerVM is using either
pvm_hmc or spvm. I think the best approach would be to declare pvm
itself deprecated and fade it out of use.

Related progress issue: https://progress.opensuse.org/issues/102146


backend/pvm.pm
```diff
@@ -21,17 +21,19 @@ use osutils qw(dd_gen_params gen_params runcmd);
...
     $self->{pid} = undef;
     $self->{children} = [];
     $self->{pidfilename} = 'pvm.pid';
-    $self->{pvmctl} = '/usr/bin/pvmctl';
-    $self->{masterlpar} = substr($masterlpar, 0, -1);
+    $self->{pvmctl} = $ENV{PVMCTL} // '/usr/bin/pvmctl';
+    $self->{masterlpar} = substr(_masterlpar, 0, -1);
     die "pvmctl not found" unless -x $self->{pvmctl};
-
+    warn 'DEPRECATED: backend::pvm is unsupported and planned to be removed from os-autoinst eventually';
```
Martchus, line 36: This would include a line number and makes it almost look like an internal warning (coming from some Perl module the user doesn't care about). Maybe just use `bmwqemu::fctwarn` to make it in line with other user facing warnings?

Martchus (changes requested):

backend/pvm.pm
okurz: good idea. Done

Martchus (approved):

okurz:
@AdamWill @os-autoinst/tests-maintainer @phil-hands please be aware about this change. I assume you don't need backend::pvm anywhere. If you use it please let us know.

AdamWill:
thanks for the heads up! No, we're not using this backend. We still only use the qemu backend, in fact.

PR #1853, opened 2021-11-15
qemu: Use wait=off instead of the deprecated nowait

kalikiana:
Otherwise the tests fail because of a deprecation message:

    warning: short-form boolean option 'server' deprecated
    Please use wait=off instead

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -885,7 +885,7 @@ sub start_qemu ($self) {
         }
 
         my $qmpid = 'qmp_socket';
-        sp('chardev', [qv "socket path=$qmpid server nowait id=$qmpid logfile=$qmpid.log logappend=on"]);
+        sp('chardev', [qv "socket path=$qmpid server wait=off id=$qmpid logfile=$qmpid.log logappend=on"]);
         sp('qmp', "chardev:$qmpid");
         sp('S');
     }


kalikiana:
I splitt this off from #1834. And it seems like the new tests won't behave correctly without this.

okurz (changes requested):
As stated in https://github.com/os-autoinst/os-autoinst/pull/1834#pullrequestreview-806023732

changing the commands on the command line might introduce backward incompatible changes. You might fix a warning on Tumbleweed but introduce problems for older Leap or SLE for example.

I suggest you run tests with this change on openSUSE Leap 15.2 and Leap 15.3 as well, e.g. simply from a container. I don't know out of my head which tests we already cover automatically on OBS but for sure I know that we currently don't check on Leap 15.3 automatically for each PR so that would need to be covered manually so far.

kalikiana:
> changing the commands on the command line might introduce backward incompatible changes. You might fix a warning on Tumbleweed but introduce problems for older Leap or SLE for example.

On both 15.2 and 15.3 I can't get tests to pass locally. I used `zypper in openQA-devel os-autoinst-devel os-autoinst-distri-opensuse-deps 'perl(Time::Moment)' 'perl(Test::MockRandom)' 'perl(File::chdir)' 'perl(Inline::Python)'` but still get errors. Is there any way to install all deps required by the git repo? It's hard to tell if I'm missing something:

    # Could not find platform independent libraries <prefix>
    # Could not find platform dependent libraries <exec_prefix>
    # Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
    # ImportError: No module named site

I would guess "python" from that, but that's already installed and working 🤔

okurz:
> > changing the commands on the command line might introduce backward incompatible changes. You might fix a warning on Tumbleweed but introduce problems for older Leap or SLE for example.
> 
> On both 15.2 and 15.3 I can't get tests to pass locally. I used `zypper in openQA-devel os-autoinst-devel os-autoinst-distri-opensuse-deps 'perl(Time::Moment)' 'perl(Test::MockRandom)' 'perl(File::chdir)' 'perl(Inline::Python)'` but still get errors

How did you test? And do you mean you can't get tests to pass locally on origin/master or with the changes from this PR?

> Is there any way to install all deps required by the git repo?

os-autoinst-devel should be enough on openSUSE. Unfortunately so far for years I have not managed to change the CI so that it uses the same which would ensure that this consistently works. CI uses the container image which should have all dependencies already pre-installed. The image is built in https://build.opensuse.org/package/show/devel:openQA/os-autoinst_dev and it's available from registry.opensuse.org/devel/openqa/containers/os-autoinst_dev:latest

kalikiana:
> os-autoinst-devel should be enough on openSUSE

As discussed, [os-autoinst is not very clear on that](https://github.com/os-autoinst/os-autoinst#installing-dependencies) so I'm proposing

kalikiana:
The test is not fully covered because of the `if ($arch eq 'aarch64') {` handling... this is not new code, though, I moved it into a new function, so I wonder why it wasn't covered before 🤔

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1858 has been merged but certainly does not block this PR, so what's next?

kalikiana:
> #1858 has been merged but certainly does not block this PR, so what's next?

Based on your previous comments I assume you know that there's no well-defined manual or automated way to test this on 15.2 and 15.3, and that's exactly what is next.

kalikiana:
> > #1858 has been merged but certainly does not block this PR, so what's next?

There's no 15.2 version of os-autoinst_dev, so this is with the default container:

    toolbox -u -i registry.opensuse.org/opensuse/leap:15.2 -t leap15.2
    releasever=15.2; sudo zypper ar -p 95 -f 'http://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_$releasever' devel_openQA; sudo zypper ar -p 90 -f 'http://download.opensuse.org/repositories/devel:openQA:Leap:$releasever/openSUSE_Leap_$releasever' devel_openQA_Leap; sudo zypper dup --allow-vendor-change
    cd os-autoinst; cd build; cmake ..; make test

```
Running tests...
Test project /home/kalikiana/openQA-dev/repos/os-autoinst/build
    Start 1: test-installed-files
1/8 Test #1: test-installed-files .............   Passed    9.85 sec
    Start 2: test-local-yaml-syntax
2/8 Test #2: test-local-yaml-syntax ...........   Passed    0.40 sec
    Start 3: test-local-python-style
3/8 Test #3: test-local-python-style ..........***Failed    0.01 sec
    Start 4: test-local-tidy
4/8 Test #4: test-local-tidy ..................   Passed   10.53 sec
    Start 5: test-local-shellcheck
5/8 Test #5: test-local-shellcheck ............   Passed    0.18 sec
    Start 6: test-local-perl-style
6/8 Test #6: test-local-perl-style ............   Passed    7.01 sec
    Start 7: test-doc-testapi-spellchecking
7/8 Test #7: test-doc-testapi-spellchecking ...   Passed    0.19 sec
    Start 8: test-perl-testsuite

```

Something's getting stuck here 🤔

Martchus:
And you've installed `os-autoinst-devel` inside that container? Just because your commands look like you've only added the repos and updated it.

kalikiana:
> And you've installed `os-autoinst-devel` inside that container? Just because your commands look like you've only added the repos and updated it.

    zypper in os-autoinst-devel
    No update candidate for 'os-autoinst-devel-4.6.1637922757.627473ea-lp152.946.1.x86_64'. The highest available version is already installed.

okurz:
That won't be enough as this only looks for new versions within the repositories from which packages were originally installed from. Likely you would need to use `zypper dup --allow-vendor-change` assuming that this is by default is not enabled. Anyway, in the meantime there's https://github.com/os-autoinst/os-autoinst/pull/1865 which also changes another parameter. As now also AdamWill knows what changes I feel confident to merge that PR

PR #1854, opened 2021-11-15
ipmi: Make persistent SOL console enabled by default

czerw:
Follow of  https://progress.opensuse.org/issues/60437 and https://github.com/os-autoinst/os-autoinst/pull/1654 where we agreed that this behavior should be enabled by default.

Verification runs:
* old default: http://10.100.12.105/tests/1377#step/kdump/42
* new default:  http://10.100.12.105/tests/1387


okurz (commented):
+1

czerw:
@alice-suse  FYI

alice-suse:
@czerw Thanks for letting us know.

@waynechen55 @Julie-CAO  FYI. If any VT openqa job fails more on sol console, maybe this is related, and we can take an action to disable it on our job group level.

okurz (approved):

PR #1855, opened 2021-11-19
 osutils: Do not hide errors of exec call in run() and run_diag()

cfconrad:
This change does actually 3 changes.
  * It fix the issue poo#81828 [1], as it bypass the race error of  `ReadWriteProcess(code => sub {})` (see: [2])
  * It show the output of the executed process.
  * run_diag() show a stacktrace if someone try to run a command which does not exists.
    
[1] https://progress.opensuse.org/issues/81828
[2] https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/pull/18

Example output of `run_diag()` looks like:
```
[2021-11-21T22:39:40.973951+01:00] [debug] Run command `/usr/bin/chattr +C /var/lib/openqa/pool/3/raid`
[2021-11-21T22:39:41.002731+01:00] [debug] Command `/usr/bin/chattr +C /var/lib/openqa/pool/3/raid` exit with 1
  /usr/bin/chattr: Operation not supported while setting flags on /var/lib/openqa/pool/3/raid
```

Example output of a command which doesn't exists:
```
[2021-11-21T22:39:41.002857+01:00] [debug] Run command `/I_do_not_exists +C /var/lib/openqa/pool/3/raid`
[2021-11-21T22:39:41.016686+01:00] [debug] Command `/I_do_not_exists +C /var/lib/openqa/pool/3/raid` failed fatal with: open3: exec of /I_do_not_exists +C /var/lib/openqa/pool/3/raid failed: No such file or directory at /usr/lib/perl5/5.26.1/IPC/Open3.pm line 283.
  	eval {...} called at /usr/lib/perl5/5.26.1/IPC/Open3.pm line 285
  	IPC::Open3::_open3("open3", GLOB(0x564903ecd848), GLOB(0x564903ecd860), undef, "/I_do_not_exists", "+C", Mojo::File=SCALAR(0x564903edbd48)) called at /usr/lib/perl5/5.26.1/IPC/Open3.pm line 359
  	IPC::Open3::open3(GLOB(0x564903ecd848), GLOB(0x564903ecd860), undef, "/I_do_not_exists", "+C", Mojo::File=SCALAR(0x564903edbd48)) called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 128
  	Mojo::IOLoop::ReadWriteProcess::_open(Mojo::IOLoop::ReadWriteProcess=HASH(0x564903e403c0), "/I_do_not_exists", "+C", Mojo::File=SCALAR(0x564903edbd48)) called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 477
  	Mojo::IOLoop::ReadWriteProcess::start(Mojo::IOLoop::ReadWriteProcess=HASH(0x564903e403c0)) called at /usr/lib/os-autoinst/osutils.pm line 85
  	osutils::run("/I_do_not_exists", "+C", Mojo::File=SCALAR(0x564903edbd48)) called at /usr/lib/os-autoinst/osutils.pm line 100
  	eval {...} called at /usr/lib/os-autoinst/osutils.pm line 98
  	osutils::run_diag("/I_do_not_exists", "+C", Mojo::File=SCALAR(0x564903edbd48)) called at /usr/lib/os-autoinst/backend/qemu.pm line 754
  	backend::qemu::start_qemu(backend::qemu=HASH(0x564905883930)) called at /usr/lib/os-autoinst/backend/qemu.pm line 116
  	backend::qemu::do_start_vm(backend::qemu=HASH(0x564905883930)) called at /usr/lib/os-autoinst/backend/baseclass.pm line 382
  	backend::baseclass::start_vm(backend::qemu=HASH(0x564905883930), undef) called at /usr/lib/os-autoinst/backend/baseclass.pm line 73
  	backend::baseclass::handle_command(backend::qemu=HASH(0x564905883930), HASH(0x564903edbdf0)) called at /usr/lib/os-autoinst/backend/baseclass.pm line 539
  	backend::baseclass::check_socket(backend::qemu=HASH(0x564905883930), IO::Handle=GLOB(0x56490574c7a0)) called at /usr/lib/os-autoinst/backend/qemu.pm line 1100
  	backend::qemu::check_socket(backend::qemu=HASH(0x564905883930), IO::Handle=GLOB(0x56490574c7a0), 0) called at /usr/lib/os-autoinst/backend/baseclass.pm line 229
  	backend::baseclass::do_capture(backend::qemu=HASH(0x564905883930), undef, 1637530780.91973) called at /usr/lib/os-autoinst/backend/baseclass.pm line 258
  	eval {...} called at /usr/lib/os-autoinst/backend/baseclass.pm line 258
  	backend::baseclass::run_capture_loop(backend::qemu=HASH(0x564905883930)) called at /usr/lib/os-autoinst/backend/baseclass.pm line 127
  	backend::baseclass::run(backend::qemu=HASH(0x564905883930), 13, 16) called at /usr/lib/os-autoinst/backend/driver.pm line 73
  	backend::driver::__ANON__(Mojo::IOLoop::ReadWriteProcess=HASH(0x5649057b6788)) called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 326
  	eval {...} called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 326
  	Mojo::IOLoop::ReadWriteProcess::_fork(Mojo::IOLoop::ReadWriteProcess=HASH(0x5649057b6788), CODE(0x564905f88588)) called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 477
  	Mojo::IOLoop::ReadWriteProcess::start(Mojo::IOLoop::ReadWriteProcess=HASH(0x5649057b6788)) called at /usr/lib/os-autoinst/backend/driver.pm line 74
  	backend::driver::start(backend::driver=HASH(0x5648ff091508)) called at /usr/lib/os-autoinst/backend/driver.pm line 39
  	backend::driver::new("backend::driver", "qemu") called at /usr/bin/isotovideo line 211
  	main::init_backend() called at /usr/bin/isotovideo line 262
```

## qemu: Don't call chattr with timeout and remove -f arg

The problem with chattr isn't about chattr it self, it seems to be
located in Mojo::IOLoop::ReadWriteProcess.
Also remove the `-f` argument, so we get error messages visible!

### DIFF
--- a/osutils.pm
+++ b/osutils.pm
@@ -78,28 +78,31 @@ sub quote {
 }
 
 sub run {
-    bmwqemu::diag "running " . join(' ', @_);
-    my @args = @_;
-    my $out;
-    my $buffer;
-    open my $handle, '>', \$buffer;
-    my $p = process(sub { local *STDERR = $handle; exec(@args) });
-    $p->channels(0)->quirkiness(1)->internal_pipes(0)->separate_err(0)->start;
-    $p->on(stop => sub {
-            while (defined(my $line = $p->getline)) {
-                $out .= $line;
-            }
-            bmwqemu::diag $buffer if defined $buffer && length($buffer) > 0;
-    });
-    $p->wait_stop;
+    my @cmd = @_;
+
+    bmwqemu::diag "running `@cmd`";
+    my $p = process(execute => shift @cmd, args => [@cmd]);
+    $p->quirkiness(1)->separate_err(0)->start()->wait_stop();
+
+    my $stdout = join('', $p->read_stream->getlines());
+    chomp $stdout;
+
     close($p->$_ ? $p->$_ : ()) for qw(read_stream write_stream error_stream);
 
-    chomp($out) if $out;
-    return $p->exit_status, $out;
+    return $p->exit_status, $stdout;
 }
 
 # Do not check for anything - just execute and print
-sub run_diag { my $o = (run(@_))[1]; bmwqemu::diag($o) if $o; $o }
+sub run_diag {
+    my ($exit_status, $output);
+    eval {
+        local $SIG{__DIE__} = undef;
+        ($exit_status, $output) = run(@_);
+        bmwqemu::diag("Command `@_` terminated with $exit_status" . (length($output) ? "\n$output" : ''));
+    };
+    bmwqemu::diag("Fatal error in command `@_`: $@") if ($@);
+    return $output;
+}
 
 # Open a process to run external program and check its return status
 sub runcmd {
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -716,8 +716,6 @@ sub start_qemu ($self) {
     mkpath($basedir);
 
     # do not use autodie here, it can fail on tmpfs, xfs, ...
-    # potential endless loop on chattr call,
-    # see https://progress.opensuse.org/issues/81828
     run_diag('/usr/bin/chattr', '+C', $basedir);
 
     bmwqemu::diag('Configuring storage controllers and block devices');
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -716,9 +716,9 @@ sub start_qemu ($self) {
     mkpath($basedir);
 
     # do not use autodie here, it can fail on tmpfs, xfs, ...
-    # timeout with arbitrary value to catch a potential indefinite blocking
-    # chattr call, also see https://progress.opensuse.org/issues/81828
-    run_diag('timeout 30 /usr/bin/chattr', '-f', '+C', $basedir);
+    # potential endless loop on chattr call,
+    # see https://progress.opensuse.org/issues/81828
+    run_diag('/usr/bin/chattr', '+C', $basedir);
 
     bmwqemu::diag('Configuring storage controllers and block devices');
     my $keephdds = $vars->{KEEPHDDS} || $vars->{SKIPTO};
--- a/t/13-osutils.t
+++ b/t/13-osutils.t
@@ -10,7 +10,7 @@ use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '5';
 use Test::MockModule;
 use Test::Warnings qw(:all :report_warnings);
-use Test::Output qw(stderr_like);
+use Test::Output qw(stderr_like stderr_unlike);
 
 subtest qv => sub {
     use osutils 'qv';
@@ -150,13 +150,43 @@ subtest runcmd => sub {
 
     my @cmd = ('qemu-img', 'create', '-f', 'qcow2', 'image.qcow2', '1G');
     my $ret;
-    stderr_like { $ret = runcmd(@cmd) } qr/running qemu-img/, 'debug runcmd progress output';
+    stderr_like { $ret = runcmd(@cmd) } qr/running `qemu-img/, 'debug runcmd progress output';
     is $ret, 0, "qemu-image creation and get its return code";
-    stderr_like { $ret = runcmd('rm', 'image.qcow2') } qr/running rm/, 'debug runcmd output with rm';
+    stderr_like { $ret = runcmd('rm', 'image.qcow2') } qr/running `rm/, 'debug runcmd output with rm';
     is $ret, 0, "delete image and get its return code";
     local $@;
     stderr_like { eval { runcmd('ls', 'image.qcow2') } } qr/No such file or directory/, 'no image found as expected';
     like $@, qr/runcmd 'ls image.qcow2' failed with exit code \d+/, "command failed and calls die";
+};
+
+subtest run_diag => sub {
+    use osutils 'run_diag';
+
+    stderr_like {
+        is(run_diag(qw(echo foo)), 'foo', 'Return stdout')
+    } qr/terminated with 0/, 'Exit code appear in log';
+
+    stderr_like {
+        is(run_diag('echo foo 1>&2'), 'foo', 'Return stderr')
+    } qr/running `echo/, 'Command appear in log';
+
+    stderr_unlike {
+        is(run_diag('false'), '', 'Empty string, if command does not produce output')
+    } qr/^\s*$/m, 'No empty line, if command does not produce output';
+
+
+
+    stderr_like {
+        run_diag('echo "foo$$bar"');
+    } qr/foo\d+bar/, 'Output appear in the log';
+
+    stderr_like {
+        run_diag('echo "foo$$bar" 1>&2');
+    } qr/foo\d+bar/, 'STDERR output appear in the log';
+
+    stderr_like {
+        is(run_diag('/I_do_not_exists'), undef, 'Return undef on execution error and do not die')
+    } qr/No such file or directory/, 'Error message appear in log';
 };
 
 subtest attempt => sub {


osutils.pm
```diff
@@ -78,28 +78,26 @@ sub quote {
...
-            bmwqemu::diag $buffer if defined $buffer && length($buffer) > 0;
-    });
-    $p->wait_stop;
+    my @cmd = @_;
+
+    bmwqemu::diag "Run command `@cmd`";
+    my $p = process(sub { exec(@cmd) });
+    $p->quirkiness(1)->separate_err(0)->start()->wait_stop();
+
+    my $stdout = join('', $p->read_stream->getlines());
```
Martchus, line 87: I assume this is to workaround https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/pull/16.

Martchus (approved):

okurz (approved):
This looks great! Awesome :)

okurz (changes requested):
The code change looks great but the decrease in code coverage is worrysome. Might be a fluke. Please crosscheck

cfconrad:
I changed the code once more. I think the error with `chattr` is this: https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/pull/17
And I also think, we use `ReadWriteProcess` process the wrong anyway. The much more better way is using it with 
```perl
process(execute =>'foo')
```
And this doesn't have the race problem at all. So we fix two things with one shot :)

I would question if it is a good idea to not die, if someone call a command which doesn't exists or isn't executable!
The old implementation didn't, so I don't do it with `run_diag()` as well. But `run()` will `die()`!

osutils.pm
cfconrad: Right

cfconrad:
> The code change looks great but the decrease in code coverage is worrysome. Might be a fluke. Please crosscheck

Before I start to write tests, plz lets first make a decision if we keep the `process(execute =>'foo')` approach.

okurz:
I don't mind `process(execute =>'foo')` but why should it be better?

cfconrad:
If we use `code=>` then we get executed here https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/blob/master/lib/Mojo/IOLoop/ReadWriteProcess.pm#L326, there is some tricky code to get error codes from perl code but it wouldn't be called if we do `exec`.  And also it use use `exit($@)` to write exit_status with the output of the exception. And I actually prefer to get the exception directly. 

But most important is that we do not have the race in `execute=>` path.

Martchus (changes requested):
Looks like `18-backend-qemu.t` needs to be adapted.

okurz:
> If we use `code=>` ...

that perfectly makes sense to me. Thanks a lot for explaining. I would not mind if you put this explanation into the git commit message as well :)

osutils.pm
```diff
@@ -78,28 +78,31 @@ sub quote {
...
-    $p->on(stop => sub {
-            while (defined(my $line = $p->getline)) {
-                $out .= $line;
-            }
-            bmwqemu::diag $buffer if defined $buffer && length($buffer) > 0;
-    });
-    $p->wait_stop;
+    my @cmd = @_;
+
+    bmwqemu::diag "Run command `@cmd`";
```
okurz, line 83: this sounds like a command to the reader but that's wrong because this line is just saying what is done. So this is why previously it said "running". Please keep that.

osutils.pm
```diff
@@ -78,28 +78,31 @@ sub quote {
...
 # Do not check for anything - just execute and print
-sub run_diag { my $o = (run(@_))[1]; bmwqemu::diag($o) if $o; $o }
+sub run_diag {
+    my ($exit_status, $output);
+    eval {
+        local $SIG{__DIE__} = undef;
+        ($exit_status, $output) = run(@_);
+        bmwqemu::diag("Command `@_` exit with $exit_status" . (length($output) ? "\n$output" : ''));
+    };
+    bmwqemu::diag("Command `@_` failed fatal with: $@") if ($@);
```
okurz, line 103: "failed fatal" is incorrect language. Likely you mean "failed fatally"? But then maybe better write the complete string as something like

```suggestion
    bmwqemu::diag("Fatal error in command `@_`: $@") if ($@);
```

Though I wonder if "fatal" is not a bit too much considering that it's not isotovideo that fails but *only* a command that is spawned within.

okurz (changes requested):

osutils.pm
cfconrad: I was trying to differ between a command which just exit with `!=0` and a call which failed because of e.g. permission denied or command not found.

osutils.pm
cfconrad: ok. If you don't mind, I will keep the quotes.

osutils.pm
cfconrad: I took your suggestion, ok?

osutils.pm
```diff
@@ -78,28 +78,31 @@ sub quote {
...
 }
 
 # Do not check for anything - just execute and print
-sub run_diag { my $o = (run(@_))[1]; bmwqemu::diag($o) if $o; $o }
+sub run_diag {
+    my ($exit_status, $output);
+    eval {
+        local $SIG{__DIE__} = undef;
+        ($exit_status, $output) = run(@_);
+        bmwqemu::diag("Command `@_` exit with $exit_status" . (length($output) ? "\n$output" : ''));
```
okurz, line 101: "exit" is a command, like when you tell me "Please exit my car" :) It should stay "exited" but maybe better is "terminated"?

okurz (changes requested):
one minor nit-pick left, sorry about that. Rest looks fine. I still see a problem with coverage. Maybe something else in our coverage analysis broke. Maybe it looks better if you rebase? Still, we should know exactly why our coverage looks like it does right now before we can merge. Please try to rebase.

osutils.pm
okurz: "exit" is a command, like when you tell me "Please exit my car" :) It should stay "exited" but maybe better is "terminated"?

okurz (approved):

Martchus (approved):

perlpunk:
I'm currently investigating sporadic 13-osutils.t failures: https://progress.opensuse.org/issues/103422
```
8: 
8:     #   Failed test 'Exit code appear in log'
8:     #   at ./13-osutils.t line 167.
8:     # STDERR:
8:     # [2021-12-02T11:23:54.401579Z] [debug] running `echo foo`
8:     # [2021-12-02T11:23:54.405060Z] [debug] Command `echo foo` terminated with 1
8:     #   foo
8:     # 
8:     # doesn't match:
8:     # (?^u:terminated with 0)
8:     # as expected
8:     # Looks like you failed 1 test of 10.
8: 
8: #   Failed test 'run_diag'
8: #   at ./13-osutils.t line 190.
8: # Looks like you failed 1 test of 9.
8: [11:23:54] ./13-osutils.t ........................... 
8: Dubious, test returned 1 (wstat 256, 0x100)
8: Failed 1/9 subtests 
```
(Note that since Oct 28 (because of a testing bug I introduced) failures didn't make CI fail)

Since that test is calling `run_diag`, could it be possible that this PR introduced it?

cfconrad:
Same as here https://github.com/mudler/Mojo-IOLoop-ReadWriteProcess/issues/20 . I hope we will get rid of it, once https://build.opensuse.org/request/show/935940 is accepted.

PR #1856, opened 2021-11-19
baseclass: Improve error message on "half-open socket"

cfconrad:
## baseclass: Improve error message on "half-open socket"
Ticket: https://progress.opensuse.org/issues/60161

The current error message is miss leading. As we die here because of
some read attempts threshold. And it must not be correlated to the
current active console.

This try to improve the error message and point to the actual fd which
hit the limit. Also we can name the fd's so it is more easy to identify
them on debugging.

## baseclass: Add test for check_select_rate() function

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -30,6 +30,7 @@ use Mojo::File 'path';
 use OpenQA::Exceptions;
 use Time::Seconds;
 use English -no_match_vars;
+use OpenQA::NamedIOSelect;
 
 # should be a singleton - and only useful in backend process
 our $backend;
@@ -107,10 +108,10 @@ sub run ($self, $cmdpipe, $rsppipe) {
 
     bmwqemu::diag "started mgmt loop with pid $$";
 
-    my $select_read = $self->{select_read} = IO::Select->new;
-    my $select_write = $self->{select_write} = IO::Select->new;
-    $select_read->add($self->{cmdpipe});
-    $select_write->add($self->{cmdpipe});
+    my $select_read = $self->{select_read} = OpenQA::NamedIOSelect->new;
+    my $select_write = $self->{select_write} = OpenQA::NamedIOSelect->new;
+    $select_read->add($self->{cmdpipe}, "baseclass::cmdpipe");
+    $select_write->add($self->{cmdpipe}, "baseclass::cmdpipe");
 
     $self->last_update_request("-Inf" + 0);
     $self->last_screenshot(undef);
@@ -183,7 +184,7 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
         }
 
         my $time_to_next = min($time_to_screenshot, $time_to_update_request, $time_to_timeout);
-        my ($read_set, $write_set) = IO::Select->select($self->{select_read}, $self->{select_write}, undef, $time_to_next);
+        my ($read_set, $write_set) = IO::Select->select($self->{select_read}->select(), $self->{select_write}->select(), undef, $time_to_next);
 
         # We need to check the video encoder and the serial socket
         my ($video_encoder, $external_video_encoder, $other) = (0, 0, 0);
@@ -221,7 +222,13 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
                 # The maximum hits per 10 seconds saw on a half open socket was >100k
                 if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
                     my $console = $self->{current_console}->{testapi_console};
-                    OpenQA::Exception::ConsoleReadError->throw(error => "The console '$console' is not responding (half-open socket?). Make sure the console is reachable or disable stall detection on expected disconnects with '\$console->disable_vnc_stalls', for example in case of intended machine shutdown");
+                    my $fd_nr = fileno $fh;
+                    my $cnt = $buckets->{BUCKET}{$fd_nr};
+                    my $name = $self->{select_read}->get_name($fh);
+                    my $msg = "The file descriptor $fd_nr ($name) hit the read attempts threshold of $hits_limit/${wait_time_limit}s by $cnt. ";
+                    $msg .= "Active console '$console' is not responding, it could be a half-open socket or you need to increase _CHKSEL_RATE_HITS value. ";
+                    $msg .= "Make sure the console is reachable or disable stall detection on expected disconnects with '\$console->disable_vnc_stalls', for example in case of intended machine shutdown.";
+                    OpenQA::Exception::ConsoleReadError->throw(error => $msg);
                 }
             }
 
@@ -497,11 +504,11 @@ sub enqueue_screenshot ($self, $image) {
           if defined $external_video_encoder_cmd_pipe;
     }
     my $encoder_pipe = $self->{encoder_pipe};
-    $self->{select_read}->add($encoder_pipe);
-    $self->{select_write}->add($encoder_pipe);
+    $self->{select_read}->add($encoder_pipe, 'baseclass::encoder_pipe');
+    $self->{select_write}->add($encoder_pipe, 'baseclass::encoder_pipe');
     if (defined $external_video_encoder_cmd_pipe) {
-        $self->{select_read}->add($external_video_encoder_cmd_pipe);
-        $self->{select_write}->add($external_video_encoder_cmd_pipe);
+        $self->{select_read}->add($external_video_encoder_cmd_pipe, 'baseclass::external_video_encoder_cmd_pipe');
+        $self->{select_write}->add($external_video_encoder_cmd_pipe, 'baseclass::external_video_encoder_cmd_pipe');
     }
     $self->{video_frame_number} += 1;
 
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -220,7 +220,7 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
             if (fileno $fh && fileno $fh != -1) {
                 # Very high limits! On a working socket, the maximum hits per 10 seconds will be around 60.
                 # The maximum hits per 10 seconds saw on a half open socket was >100k
-                if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
+                if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh, time())) {
                     my $console = $self->{current_console}->{testapi_console};
                     my $fd_nr = fileno $fh;
                     my $cnt = $buckets->{BUCKET}{$fd_nr};
@@ -270,8 +270,7 @@ sub run_capture_loop ($self, $timeout = undef) {
 
 # wait_time_limit = seconds
 # This is not sliding buckets. All the IDs inside the bucket must be over the limit!
-sub check_select_rate ($buckets, $wait_time_limit, $hits_limit, $id) {
-    my $time = gettimeofday;
+sub check_select_rate ($buckets, $wait_time_limit, $hits_limit, $id, $time) {
     my $lower_limit = $buckets->{TIME} //= $time;
     my $upper_limit = $lower_limit + $wait_time_limit;
     if ($time > $upper_limit) {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ set(PERL_FILES
     OpenQA/Isotovideo/Interface.pm
     OpenQA/Isotovideo/NeedleDownloader.pm
     OpenQA/Isotovideo/Utils.pm
+    OpenQA/NamedIOSelect.pm
     OpenQA/Qemu/BlockDevConf.pm
     OpenQA/Qemu/BlockDev.pm
     OpenQA/Qemu/ControllerConf.pm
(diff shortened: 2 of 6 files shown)


backend/baseclass.pm
```diff
@@ -2,6 +2,46 @@
 # Copyright 2012-2021 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+# Helper class to identify file descriptors in error case, see do_capture()
+package NamedIOSelect;
```
Martchus, line 6: This is better added to a new file.

backend/baseclass.pm
```diff
@@ -2,6 +2,46 @@
...
+use Mojo::Base -base;
+use IO::Select;
+use Carp;
+
+sub select {
+    return shift->{select_obj} //= IO::Select->new();
```
Martchus, line 12: For new code we should better use signatures right away.

backend/baseclass.pm
```diff
@@ -2,6 +2,46 @@
...
+}
+
+sub names {
+    return shift->{names_hash} //= {};
+}
+
+sub add
+{
```
Martchus, line 20: You'll likely need to run tidy.

backend/baseclass.pm
```diff
@@ -2,6 +2,46 @@
...
+    my ($self, $fd, $name) = @_;
+    my $fd_nr = fileno($fd);
+    if (!defined($name)) {
+        my ($package, $filename, $line) = caller;
+        $name = sprintf('NamedIOSelect::add(%d) called at %s:%d', $fd_nr, $filename, $line);
+    }
```
Martchus, line 26: Is this a leftover from debugging?

backend/baseclass.pm
```diff
@@ -221,7 +261,17 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
                 # The maximum hits per 10 seconds saw on a half open socket was >100k
                 if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
                     my $console = $self->{current_console}->{testapi_console};
-                    OpenQA::Exception::ConsoleReadError->throw(error => "The console '$console' is not responding (half-open socket?). Make sure the console is reachable or disable stall detection on expected disconnects with '\$console->disable_vnc_stalls', for example in case of intended machine shutdown");
+                    my $fd_nr = fileno $fh;
+                    my $cnt = $buckets->{$fd_nr}->{cnt};
+                    my $name = $self->{select_read}->get_name($fh);
+                    my $msg = "The file descriptor $fd_nr ($name) hit the read attempts threashold of $hits_limit/${wait_time_limit}s by $cnt. ";
+                    $msg .= "Active console '$console' is not responding (half-open socket?). ";
```
Martchus, line 268: Didn't you say in the PR description

> And it must not be correlated to the current active console.

So this is possibly still misleading. Maybe add a "might" somewhere in the sentence or make it otherwise clear that it can also have other causes?

Martchus (changes requested):
This looks very helpful for debugging. However, I have some remarks. See the inline comments and also inline comments from codecov.

backend/baseclass.pm
cfconrad: Actually not, If someone is using it the `IO::Select->add()` way (without specify a name), we get descriptive name by using caller.

backend/baseclass.pm
cfconrad: ok like `Active console is '$console', it could be a be a half-open socket.` ?

cfconrad: Actually I'm not a fan of this `half-open` socket thing anyway. Normally if a socket is clothed by the other end, you will realize it within the next read/write cycle.

backend/baseclass.pm
cfconrad: done

backend/baseclass.pm
cfconrad: thx

backend/baseclass.pm
cfconrad: I moved it to a **new** folder `Utils` as I didn't found something else which suites. WDYT?

backend/baseclass.pm
cfconrad: The full error message is now:
```
[2021-11-22T10:52:20.050854+01:00] [info] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  The file descriptor 13 (baseclass::cmdpipe) hit the read attempts threashold of 15000/30s by 32004. Active console 'root-virtio-terminal' is not responding, it could be a half-open socket or you need to increase _CHKSEL_RATE_HITS value. Make sure the console is reachable or disable stall detection on expected disconnects with '$console->disable_vnc_stalls', for example in case of intended machine shutdown.
   fd:13 (baseclass::cmdpipe) => 32004 age:30.00s
   fd:17 (qemu::start_qemu::qemu_pipe) => 1 age:58.13s
```

It looks like this
Screenshot from 2021-11-22 10-57-54

I didn't used newlines for each message part, as then this hint in the Web-UI does only show the first line!

cfconrad:
@Martchus can you help me with
```
 The following tests FAILED:
	  1 - test-installed-files (Failed)
Errors while running CTest
FAILED: CMakeFiles/check 
```

backend/baseclass.pm
```diff
@@ -265,22 +276,16 @@ sub run_capture_loop ($self, $timeout = undef) {
...
-        for (keys %{$buckets->{BUCKET}}) {
-            if ($buckets->{BUCKET}{$_} < $hits_limit) {
-                $buckets->{BUCKET} = {$id => 1};
-                return 0;
-            }
-        }
-
-        return 1;
+    my $bucket = $buckets->{$id} //= {cnt => 0, time => $time};
+
```
Martchus, line 280:
```suggestion
    my $bucket = $buckets->{$id} //= {cnt => 0, time => $time};
```

Martchus (commented):
Is the use of `$/` (instead of just `\n`) really the right approach for formatting such error messages? I'm just wondering because we normally just use `\n`.

Utils/NamedIOSelect.pm
```diff
@@ -0,0 +1,40 @@
+# Copyright 2012-2021 SUSE LLC
```
Martchus, line 1: This should fix the style check:
```suggestion
# Copyright 2021 SUSE LLC
```

Utils/NamedIOSelect.pm
```diff
@@ -0,0 +1,40 @@
...
+# Helper class to to use descriptive names for file descriptors within IO::Select
+
+package Utils::NamedIOSelect;
+use Mojo::Base -base, -signatures;
+use IO::Select;
+use Carp;
+
+sub select ($self) {
+    return $self->{select_obj} //= IO::Select->new();
+}
```
okurz, line 13:
```suggestion
sub select ($self) { $self->{select_obj} //= IO::Select->new() }
```

Utils/NamedIOSelect.pm
```diff
@@ -0,0 +1,40 @@
...
+
+sub names ($self) {
+    return $self->{names_hash} //= {};
+}
```
okurz, line 17:
```suggestion
sub names ($self) { $self->{names_hash} //= {} }
```

backend/baseclass.pm
```diff
@@ -221,7 +222,17 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
                 # The maximum hits per 10 seconds saw on a half open socket was >100k
                 if (check_select_rate($buckets, $wait_time_limit, $hits_limit, fileno $fh)) {
                     my $console = $self->{current_console}->{testapi_console};
-                    OpenQA::Exception::ConsoleReadError->throw(error => "The console '$console' is not responding (half-open socket?). Make sure the console is reachable or disable stall detection on expected disconnects with '\$console->disable_vnc_stalls', for example in case of intended machine shutdown");
+                    my $fd_nr = fileno $fh;
+                    my $cnt = $buckets->{$fd_nr}->{cnt};
+                    my $name = $self->{select_read}->get_name($fh);
+                    my $msg = "The file descriptor $fd_nr ($name) hit the read attempts threashold of $hits_limit/${wait_time_limit}s by $cnt. ";
```
okurz, line 228: s/threashold/threshold/

backend/baseclass.pm
okurz: I can't spot the difference. What did you suggest to change? :D

okurz (changes requested):

backend/baseclass.pm
cfconrad: thx

backend/baseclass.pm
cfconrad: :question:

cfconrad: Remove the empty line?

cfconrad:
> Is the use of `$/` (instead of just `\n`) really the right approach for formatting such error messages? I'm just wondering because we normally just use `\n`.

I always thought using `$/` is the os independent way. But if we already have a common behavior here. I will just change it to `\n`, no problem. Should I?

Martchus:
Not sure. I'm wondering what @kraih and @perlpunk think about this.

kraih:
> Not sure. I'm wondering what @kraih and @perlpunk think about this.

Don't use `$/` for this, its long name is `$INPUT_RECORD_SEPARATOR` for a reason. 😉

backend/baseclass.pm
okurz: "Utils" is likely just as good as putting it into the top folder but I don't have a better suggestion.

okurz:
> @Martchus can you help me with
> 
> ```
>  The following tests FAILED:
> 	  1 - test-installed-files (Failed)
> Errors while running CTest
> FAILED: CMakeFiles/check 
> ```

Youd need to reference the new file in https://github.com/os-autoinst/os-autoinst/blob/master/CMakeLists.txt#L12

backend/baseclass.pm
okurz: The OBS checks fail because you added a new folder which so far is not referenced in the `%files` section of dist/rpm/os-autoinst.spec but I suggest anyway to not introduce a new folder. "Utils" too quickly deteriorates, similar to "misc" ;) Unless other people suggest something different I suggest `backend/`

okurz: @Martchus why did you suggest to move this to a new file when it's only used within backend/baseclass anyway?

backend/baseclass.pm
Martchus: Yes, I wanted to suggest removing the empty line.

cfconrad:
I remove the commit https://github.com/os-autoinst/os-autoinst/pull/1856/commits/67275faf8c5ab2f6dab9b3c70497a9fb28fbaffc from this PR, it changed the bucket logic and made it actually more strict. Which I'm now a bit scary of, also because the comments around that code explicit state it and I don't understand what we try to catch with this code. So maybe someone who has better knowledge of this part can improve/remove it.

But I have added a test for `check_select_rate()`, hope I will now reach the coverage :)

backend/baseclass.pm
Martchus: Because baseclass is already > 1000 lines and mixing multiple packages in a single file doesn't make it better. I'd just put it under `OpenQA` or under the top-level. I'd avoid `backend` because we don't want to load it as backend via `require "backend/$name.pm";` in `driver.pm`.

okurz: ok. I also think `OpenQA` is better than `Utils`

t/36-utils.t
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/perl
+
+use Test::Most;
+use Mojo::Base -strict, -signatures;
+
+subtest NamedIOSelect => sub {
+    use Utils::NamedIOSelect;
```
Martchus, line 7: This will be used at compile time in any case so just move it at the top as usual.

Martchus (changes requested):

backend/baseclass.pm
cfconrad: :+1:

t/36-utils.t
cfconrad: :+1:

cfconrad:
ready from my side. 
I'm not sure about the codecov, I think all my "new" code is covered within this change.

okurz (changes requested):

Martchus (approved):

cfconrad:
@okurz something to add?

okurz:
Runtime in CI is perfect with

```
8: [17:59:49] ./36-openqa-namedioselect.t .............. ok      307 ms ( 0.00 usr  0.00 sys +  0.27 cusr  0.04 csys =  0.31 CPU)
```

and no unhandled output. That's good.

The check for "codecov/patch" still fails because only 56% of the code that you change or newly introduce is covered with tests. Especially for new packages we should aim for 100% statement coverage (not complete test coverage, only statement for a now). So I would appreciate if you can improve that. If you really can't or don't want to we can merge without but you have to state that explicitly :)

cfconrad:
Ok I rebased on master, to see a fresh codecov. But from what I understand, all my added coded is covered.

All these `$xxx->->add($fd, "name");`  calles and `backend::baseclass:do_capture()` doesn't have coverage. But these functions had no coverage before.

So I would like to take this PR, even if I hit the coverage threshold!

okurz (approved):
I tested manually with checking out your branch and running the t/full-stack.t from openQA using that os-autoinst version and all good

PR #1857, opened 2021-11-19
baseclass: Fix errors in some tests by increasing _CHKSEL_RATE_HITS

cfconrad:
Ticket: https://progress.opensuse.org/issues/60161

This just double the allowed read attempts in 30s to 30k. This is needed, as
faster workers can process more then 15k read attempts on one socket in
30s. In general idea of accounting reads per socket isn't a good. 
But we do, and the default was to low these days.
Of course all tests, which do heavy console interaction, can set this value
for it's own, but I think it's not obvious that a console error might be this
variable. So having a better default value makes sense.

We can easily trigger up to 55k by a simple valid testcode like this
```perl
while(1) { script_output('true', quiet => 1) };
```

http://openqa-3.wicked.suse.de/tests/31388 with _CHKSEL_RATE_HITS=15000 => FAILED
http://openqa-3.wicked.suse.de/tests/31387 with _CHKSEL_RATE_HITS=100000 => SUCCESSFUL

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -149,7 +149,7 @@ sub do_capture ($self, $timeout = undef, $starttime = undef) {
     # Time slot buckets
     my $buckets = {};
     my $wait_time_limit = $bmwqemu::vars{_CHKSEL_RATE_WAIT_TIME} // 30;
-    my $hits_limit = $bmwqemu::vars{_CHKSEL_RATE_HITS} // 15_000;
+    my $hits_limit = $bmwqemu::vars{_CHKSEL_RATE_HITS} // 30_000;
 
     while (1) {
         last unless $self->{cmdpipe};


Martchus (approved):
I suppose this is ok. I've been adding the "not-ready" label to give more people the chance to review (before mergify would merge it automatically).

Martchus:
By the way, since this is read from a test variable you could also just increase the value on tests which need it.

cfconrad:
Yeah sure, this is what I do so far. In my environment I set this value to 60k in the workers config.

okurz (changes requested):
Can you please rephrase the git commit message subject line (and maybe also the details) to describe the *Why* more than the *What*, also see https://chris.beams.io/posts/git-commit/ about this. So far I don't understand why you do this, what you want to achieve, what problem to fix, what's the effect and potential side-effects. Think how a reader of openqa.opensuse.org/changelog would see your subject (only the subject is shown for each commit) "baseclass: Increase default value of _CHKSEL_RATE_HITS" and it will likely read like "Change foo … 'cause we can" :)

cfconrad:
@okurz done, plz check

okurz:
I am sorry, it's still not clear. Why as a user should I care that you increase the default value of FOO? Try to write it in a more awesome way :) Maybe something like "Fix sporadic job failures on …" or something

Martchus:
The commit message also contains typos.

cfconrad:
@okurz There is no sporadic error or something. It is just bad code. And a good fix would be to remove it and find the cause of the actual problems that where tried to catch with it. But there is this wired parameter and it just doesn't apply to our workers power anymore. So with increasing it, we at least do not run in that **false positive** so quick anymore.

I actually think that the commit message (not looking at the typos) is ok and if later someone try to understand what was the reason for that change, he/she will get an answer from it (if not, there is the ticked and this PR). Also if someone find this variable and do a `git log --grep` he will find it.

Plz give me a complete proposal, or feel free to change it by your self.

okurz:
> Plz give me a complete proposal, or feel free to change it by your self.

If I would know something better I would tell you. The thing is I don't know why you change it at all

cfconrad:
> > Plz give me a complete proposal, or feel free to change it by your self.
> 
> If I would know something better I would tell you. The thing is I don't know why you change it at all

Because the current default value of 15k cause errors in some tests and I think the default is just not a good choice.

okurz:
> > > Plz give me a complete proposal, or feel free to change it by your self.
> > 
> > 
> > If I would know something better I would tell you. The thing is I don't know why you change it at all
> 
> Because the current default value of 15k cause errors in some tests and I think the default is just not a good choice.
> 
> Look at this comment https://progress.opensuse.org/issues/60161#note-20 it's one of the workarounds you can do to avoid it. > But I think, such a variable doesn't fit into a testsuite... so adopting the default is a better solution.

I agree. That's better. But can you please phrase the git commit message accordingly? How about "Fix errors in some tests by increasing _CHKSE_RATE_HITS" for the subject line?

PR #1858, opened 2021-11-23
docs: Clearly mention devel:openQA in README

kalikiana:
Installing from default repos is not enough.

image

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -189,9 +189,14 @@ requests for consideration or create an issue with a code change proposal.
 
 === Installing dependencies
 
-On openSUSE one can install the package
-link:https://build.opensuse.org/project/show/devel:openQA[`os-autoinst-devel`] which
-pulls all dependencies.
+On openSUSE one can install the package `os-autoinst-devel` which provides all
+the dependencies to build and run os-autoinst for the corresponding version of
+the sources. To build a current version of `os-autoinst` it is recommended to
+install `os-autoinst-devel` from
+link:https://build.opensuse.org/project/show/devel:openQA[devel:openQA] as the
+distribution-provided packages might be too old or miss dependencies. This is
+particularly true for openSUSE Leap. Also see
+link:https://open.qa/docs/#_development_version_repository[the openQA docs].
 
 The required dependencies are also declared in `dependencies.yaml`. (The names listed
 within that file are specific to openSUSE but can be easily transferred to other


perlpunk (approved):

README.asciidoc
```diff
@@ -190,8 +190,8 @@ requests for consideration or create an issue with a code change proposal.
 === Installing dependencies
 
 On openSUSE one can install the package
-link:https://build.opensuse.org/project/show/devel:openQA[`os-autoinst-devel`] which
-pulls all dependencies.
+link:https://build.opensuse.org/project/show/devel:openQA[`os-autoinst-devel from devel:openQA`] which
```
Martchus, line 193: "os-autoinst-devel from " should not be part of the link.

README.asciidoc
```diff
@@ -190,8 +190,8 @@ requests for consideration or create an issue with a code change proposal.
...
+pulls in all dependencies. See also link:https://open.qa/docs/#_development_version_repository[`the openQA docs`].
```
Martchus, line 194: The link text shouldn't be formatted as code, remove the \` quotes.

Martchus (changes requested):

README.asciidoc
perlpunk: Why not?

Martchus: Because the link only points to the project page and not to the package page.

README.asciidoc
perlpunk: Why not? :)

README.asciidoc
perlpunk: ok. I am fine with either way

README.asciidoc
Martchus: Because it isn't code or a command.

kalikiana: That was in fact by accident :-D

Martchus (approved):

README.asciidoc
perlpunk: Oh I misread this and thought it was also about the previous link...

README.asciidoc
```diff
@@ -189,9 +189,9 @@ requests for consideration or create an issue with a code change proposal.
 
 === Installing dependencies
 
-On openSUSE one can install the package
-link:https://build.opensuse.org/project/show/devel:openQA[`os-autoinst-devel`] which
-pulls all dependencies.
+On openSUSE one can install the package `os-autoinst-devel` from
+link:https://build.opensuse.org/project/show/devel:openQA[devel:openQA] which
+pulls in all dependencies. See also link:https://open.qa/docs/#_development_version_repository[the openQA docs].
```
okurz, line 194: Not entirely correct. The package is within both openSUSE Leap as well as Tumbleweed and works to develop for the *corresponding version*. For Tumbleweed most likely it will nearly always work because it's recent enough.

okurz (changes requested):

README.asciidoc
kalikiana: Neither Leap nor Tumbleweed is mentioned here. It works across versions so I would prefer to leave this generic.

kalikiana: Btw installing os-autoinst-devel from default repos on Leap 15.2 is what I originally tried. So I'm against anything that makes it sound like a good idea to use default repos in general.

okurz: I still think it's a good idea to extend the README but with including that notion. How about just explaining that aspect and letting the reader decide.

kalikiana: > one definitely needs additional repos which is not what I want to encourage. The preferred base should be Tumbleweed and only Tumbleweed.

So Tumbleweed should work fine with default repos? In that case I would make it a separate section.

Martchus: What about:
> On openSUSE one can install the package `os-autoinst-devel` which pulls in all dependencies. Under Leap it is recommend to install it from devel:openQA as the distribution-provided packages are usually too old.

okurz: Not entirely correct (as I stated in https://github.com/os-autoinst/os-autoinst/pull/1858#discussion_r755851692 ) but acceptable.

How about

```suggestion
On openSUSE one can install the package `os-autoinst-devel` which provides all
the dependencies to build and run os-autoinst for the corresponding version of
sources. To build a current version of `os-autoinst` it is recommended to
install `os-autoinst-devel` from
link:https://build.opensuse.org/project/show/devel:openQA[devel:openQA] as the
distribution-provided packages might be too old or miss dependencies. This is
particularly true for openSUSE Leap. Also see
link:https://open.qa/docs/#_development_version_repository[the openQA docs].
```

okurz (approved):

PR #1859, opened 2021-11-23
Log reason for stopping the test execution early

Martchus:
This should make the behavior more obvious (which can otherwise not be
obvious at all, see https://progress.opensuse.org/issues/102539).

Note that the behavior is generally intended, see 9d81cdb8e. It is also
documented, see openQA commit 712f2db38f.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -229,6 +229,12 @@ sub load_snapshot {
     query_isotovideo('backend_start_serial_grab');
 }
 
+sub _terminate () {
+    close $isotovideo;    # uncoverable statement
+    Devel::Cover::report() if Devel::Cover->can('report');    # uncoverable statement
+    _exit(0);    # uncoverable statement
+}
+
 sub run_all {
     my $died = 0;
     my $completed = 0;
@@ -242,9 +248,7 @@ sub run_all {
         bmwqemu::save_vars(no_secret => 1);
         myjsonrpc::send_json($isotovideo, {cmd => 'tests_done', died => $died, completed => $completed});
     };
-    close $isotovideo;
-    Devel::Cover::report() if Devel::Cover->can('report');
-    _exit(0);
+    _terminate;
 }
 
 sub handle_sigterm {
--- a/autotest.pm
+++ b/autotest.pm
@@ -379,6 +379,12 @@ sub runalltests {
                 query_isotovideo('backend_save_memory_dump', {filename => $fullname});
             }
             if ($t->{fatal_failure} || $flags->{fatal} || (!exists $flags->{fatal} && !$snapshots_supported) || $bmwqemu::vars{TESTDEBUG}) {
+                my $reason = ($t->{fatal_failure} || $flags->{fatal})
+                  ? 'after a fatal test failure'
+                  : ($bmwqemu::vars{TESTDEBUG}
+                    ? 'because TESTDEBUG has been set'
+                    : 'because snapshotting is disabled/unavailable and "fatal => 0" has NOT been set explicitly');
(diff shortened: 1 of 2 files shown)


autotest.pm
```diff
@@ -379,6 +379,12 @@ sub runalltests {
                 query_isotovideo('backend_save_memory_dump', {filename => $fullname});
             }
             if ($t->{fatal_failure} || $flags->{fatal} || (!exists $flags->{fatal} && !$snapshots_supported) || $bmwqemu::vars{TESTDEBUG}) {
+                my $reason = ($t->{fatal_failure} || $flags->{fatal})
+                  ? ' after a fatal test failure'
+                  : ($bmwqemu::vars{TESTDEBUG}
+                    ? 'because TESTDEBUG has been set'
+                    : 'because snapshotting is disabled/unavailable and "fatal => 0" has NOT been set explicitely');
```
perlpunk, line 386: `s/explicitely/explicitly/`

pdostal (approved):
This will help us so much! Thank you Marius

jlausuch (approved):

okurz (changes requested):
Please add according test coverage

Martchus:
Here you go :-)

t/08-autotest.t
```diff
@@ -86,88 +84,99 @@ stderr_like { autotest::run_all } qr/finished/, 'run_all outputs status on stder
...
+        $mock_basetest->redefine(runtest => sub { die "test died\n" });
+        $vm_stopped = 0;
+        stderr_like { autotest::run_all } qr/.*stopping overall test execution after a fatal test failure.*/, 'reason logged';
+        ($died, $completed) = get_tests_done;
+        is $died, 0, 'tests still not considered died if only a test module failed';
+        is $completed, 0, 'tests not considered completed';
+        is $reverts_done, 0, 'no rollbacks done';
+        is $snapshots_made, 0, 'no snapshots made';
+        $reverts_done = $snapshots_made = 0;
+        @sent = ();
```
okurz, line 165: this part is always repeated. Do you think it would work to define something like

```suggestion
sub snapshot_subtest (...) {
        // execute the "sub" reference provided
        // ...
        $reverts_done = $snapshots_made = 0;
        @sent = ();
};

snapshot_subtest 'stopping overall test execution early due to fatal test failure' => sub {
        // subtest instrutions without repeating the same cleanup code repeatedly
};
```

okurz:
looks like the coverage is not yet updated? the patch check also says "0% of diff hit"

Martchus:
I pushed a change to avoid the redundant resetting. Let's see whether the coverage is updated now. The lines are now definitely covered including all cases of the ternary operators.

t/08-autotest.t
okurz: awesome!

Martchus:
My last commit to fix the coverage collection worked introduced a few lines itself which cannot be covered. I'm marking them as uncoverable.

t/08-autotest.t
```diff
@@ -59,7 +59,7 @@ my $mock_basetest = Test::MockModule->new('basetest');
 $mock_basetest->noop('_result_add_screenshot');
 # stop run_all from quitting at the end
 my $mock_autotest = Test::MockModule->new('autotest', no_auto => 1);
-$mock_autotest->noop('_exit');
+$mock_autotest->noop('_terminate');
```
kalikiana, line 62: As asked in the daily, would be nice if you could extend the comment to also say that we can't use coverage because it's not a fork

okurz (approved):

Martchus:
It just failed due to the TBB problem so I've been rebasing.

PR #1860, opened 2021-11-23
Avoid running into TBB's soft limit for the max. number of threads

Martchus:
When executing the signalblocker test in Factory builds (which use
`libtbb12-2021.4.0`) we currently see the warning
    
```
TBB Warning: The number of workers is currently limited to 5. The request for 7 workers is ignored. Further requests for more workers will be silently ignored until the limit changes.
```
    
and the test deadlocks.
    
This change should avoid it by staying below this limit.
    
See https://progress.opensuse.org/issues/103029

### DIFF
--- a/ppmclibs/tinycv.h
+++ b/ppmclibs/tinycv.h
@@ -6,6 +6,7 @@
 
 // opaque type to separate perl from opencv
 struct Image;
+int opencv_default_thread_count();
 void create_opencv_threads(int thread_count);
 void image_destroy(Image *s);
 Image *image_new(long width, long height);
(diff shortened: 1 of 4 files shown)


okurz:
https://github.com/os-autoinst/os-autoinst/pull/1858 and potentially other PRs are blocked by the TBB restriction. Can we undraft and merge this?

Martchus:
Annoyingly in this case the number of workers was even limited to 2:

```
[  162s] 3: TBB Warning: The number of workers is currently limited to 2. The request for 7 workers is ignored. Further requests for more workers will be silently ignored until the limit changes.
```

So this PR wouldn't have helped. Likely I'll have to find a better way.

okurz (approved):

PR #1863, opened 2021-11-26
imgsearch: disable OpenCV logging

AdamWill:
Otherwise the output to stdout - which is intended to be (only)
valid JSON - may include OpenCV logging messages.


AdamWill:
not sure if this is entirely the best way of doing this, but it seems to work, anyway.

okurz (approved):

PR #1868, opened 2021-11-30
Use proper perl "signatures" in some consoles - Part 1

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

Part 1 of https://github.com/os-autoinst/os-autoinst/pull/1789


okurz:
The insufficient patch coverage was expected. Still I recommend to go ahead as this change is meant to prevent further regressions with static signature checks.

consoles/localXvnc.pm
```diff
@@ -22,19 +22,15 @@ use testapi qw(get_var);
 # ssh disconnect and console switching. Ssh console may not display returned result of
 # long-time run test without these options. TCPKeepAlive ensures that if network goes down
 # or the remote host dies, machines will be properly noticed
-sub sshCommand {
-    my ($self, $username, $host, $gui) = @_;
-
+sub sshCommand ($self, $username, $host, $gui) {
```
Martchus: `$gui` must be an optional parameter defaulting to a falsy value.

Martchus (approved):
Looks like you've just fixed things I wanted to comment about.

PR #1869, opened 2021-11-30
loadtest: Retain original test path in error messages

kalikiana:
The error handling refers to different paths than what was specified:

    SCHEDULE=console/rust
    tests died: unable to load console/rust.pm, check the log for the cause (e.g. syntax error)

    SCHEDULE=rust
    isotovideo died: loadtest: script path '../rust.pm' does not match required pattern \w.+/[^/]+.p[my]

XXX: No fix yet, I'm documenting the cases and adding unit tests first.

### DIFF
--- a/t/20-openqa-isotovideo-utils.t
+++ b/t/20-openqa-isotovideo-utils.t
@@ -43,7 +43,7 @@ subtest 'error handling when loading test schedule' => sub {
         my $state = decode_json($base_state->slurp);
         if (is(ref $state, 'HASH', 'state file contains object')) {
             is($state->{component}, 'tests', 'state file contains component');
-            like($state->{msg}, qr/unable to load foo\/bar\.pm/, 'state file contains error message');
+            like($state->{msg}, qr@unable to load foo/bar,@, 'state file contains error message');
         }
     };
     subtest 'invalid productdir' => sub {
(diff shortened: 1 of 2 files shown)


t/20-openqa-isotovideo-utils.t
```diff
@@ -43,7 +43,7 @@ subtest 'error handling when loading test schedule' => sub {
         my $state = decode_json($base_state->slurp);
         if (is(ref $state, 'HASH', 'state file contains object')) {
             is($state->{component}, 'tests', 'state file contains component');
-            like($state->{msg}, qr/unable to load foo\/bar\.pm/, 'state file contains error message');
+            like($state->{msg}, qr@unable to load foo/bar,@, 'state file contains error message');
```
okurz, line 46: I would keep this as is. But feel free to extend t/08-autotest.t for the additional, specific information you want

kalikiana: Not sure what you mean. `foo/bar` is the original value here, which I want to preserve. Currently the test expects a `.pm` extension that the user did not specify.

okurz: That's not what the message says. It says what it tried to load, not what the user specified but the user needs to understand how the file is actually called

kalikiana: I think `console/rust.pm does not exist` would have been helpful, so maybe that needs to be a separate error condition.

okurz: Yes, can be. Let me try to rephrase what I intended initially: t/20-openqa-isotovideo-utils.t is there to test stuff in OpenQA/Isotovideo/Utils.pm and mainly that. So tests specific to the error message should be in t/08-autotest.t which is the correct test to test autotest.pm which includes the serialization of the error message you are interested in

Issue #1686, opened 2021-06-14
os-autoinst OBS checks fail in t/10-virtio_terminal.t


okurz:
@cfconrad your change causes failures within OBS checks. please check https://build.opensuse.org/package/live_build_log/devel:openQA/os-autoinst/openSUSE_Tumbleweed/x86_64

```
[  513s] 3: [33m[0m[33m[0m[33m[0m
[  513s] 3:     #   Failed test '[snap1] virtio_terminal stored all available data'
[  513s] 3:     #   at ./10-virtio_terminal.t line 136.
[  513s] 3:     #          got: ''
[  513s] 3:     #     expected: 'Test data foo'
[  513s] 3: 
[  513s] 3:     #   Failed test '[snap1] Snapshots data verified'
[  513s] 3:     #   at ./10-virtio_terminal.t line 139.
[  513s] 3:     #     Structures begin differing at:
[  513s] 3:     #          $got->{buffer} = ''
[  513s] 3:     #     $expected->{buffer} = 'Test data foo'
[  513s] 3: [33m[0m[33m[0m[33m[0m[33m[0m[33m[0m
[  513s] 3:     #   Failed test '[snap1] carry over buffer sucessful loaded'
[  513s] 3:     #   at ./10-virtio_terminal.t line 153.
[  513s] 3:     #          got: ''
[  513s] 3:     #     expected: 'Test data foo'
[  514s] 3: [33m[0m    # Looks like you failed 3 tests of 22.
[  514s] 3: 
[  514s] 3: #   Failed test 'Test snapshot handling'
[  514s] 3: #   at ./10-virtio_terminal.t line 169.
[  514s] 3: # Looks like you failed 1 test of 3.
[  514s] 3: [04:17:15] ./10-virtio_terminal.t ................... 
[  514s] 3: Dubious, test returned 1 (wstat 256, 0x100)
[  514s] 3: Failed 1/3 subtests 
```

_Originally posted by @okurz in https://github.com/os-autoinst/os-autoinst/issues/1675#issuecomment-860554060_

okurz:
The test 10-virtio_terminal.t was removed from OBS tests with 7688f7b7 but the problem (likely) still persists if the test module would be re-added to OBS tests.

okurz:
@cfconrad did you see this issue?

cfconrad:
As https://github.com/os-autoinst/os-autoinst/pull/1870 state, this is solved. @okurz I cannot close the issue, maybe you can?

PR #1870, opened 2021-11-30
Fix sporadic error in t/10-virtio_terminal.t

cfconrad:
I wasn't able to reproce the problem, but this is what I guess.
We call `$console->peak()` before we wrote the data into the other side of
the pipe.
Now we use USR2 signal to sync between "sender" and "receiver".
Also calling `flush()` explicit, so we ensure data is not buffered by
perl.

Ticked: https://progress.opensuse.org/issues/94991
Issue: https://github.com/os-autoinst/os-autoinst/issues/1686

### DIFF
--- a/t/10-virtio_terminal.t
+++ b/t/10-virtio_terminal.t
@@ -15,6 +15,8 @@ use consoles::virtio_terminal;
 use testapi;
 use bmwqemu;
 
+my $pipe_data_written;
+sub wait_till_pipe_data_written() { sleep 1 while (!$pipe_data_written) }
 
 sub prepare_pipes ($socket_path, $write_buffer = undef) {
     my $pipe_in = $socket_path . ".in";
@@ -25,6 +27,9 @@ sub prepare_pipes ($socket_path, $write_buffer = undef) {
         mkfifo($_, 0666) or die("Cannot create fifo pipe $_");
     }
 
+    $pipe_data_written = 0;
+    $SIG{USR2} = sub { $pipe_data_written = 1 };
+
     my $pid = fork || do {
         my $running = 1;
         $SIG{USR1} = sub { $running = 0; };
@@ -32,14 +37,17 @@ sub prepare_pipes ($socket_path, $write_buffer = undef) {
             die('Timeout for pipe other side helper');
         };
         alarm ONE_MINUTE;
-
-        open(my $fd_r, "<", $pipe_in)
+        my $fd_r = IO::Handle->new();
+        my $fd_w = IO::Handle->new();
+        open($fd_r, "<", $pipe_in)
           or die "Can't open in pipe for writing $!";
-        open(my $fd_w, ">", $pipe_out)
+        open($fd_w, ">", $pipe_out)
           or die "Can't open out pipe for reading $!";
 
         syswrite($fd_w, $write_buffer) if ($write_buffer);
+        $fd_w->flush();
 
+        kill 'USR2', getppid;
         sysread($fd_r, my $buf, 1024) while ($running);
         exit 0;
     };
@@ -126,6 +134,7 @@ subtest "Test snapshot handling" => sub {
     is_deeply($term->{snapshots}, {}, "Snapshots are empty");
 
     $term->select();
+    wait_till_pipe_data_written();
     is_deeply($term->{snapshots}, {}, "Snapshots are empty after select/activate");
 
     $term->save_snapshot('snap1');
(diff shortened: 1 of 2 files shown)


t/10-virtio_terminal.t
```diff
@@ -15,6 +15,8 @@ use consoles::virtio_terminal;
 use testapi;
 use bmwqemu;
 
+my $pipe_data_written;
+sub wait_till_pipe_data_written() { sleep 1 while (!$pipe_data_written) }
```
Martchus, line 19: I think Perl let's you write this more nicely:
```suggestion
sub wait_till_pipe_data_written() { sleep 1 until $pipe_data_written }
```

---

Maybe one second is also a bit long here? Or will the signal handler interrupt the sleep anyways? In this case a comment like `  # sleep interrupted by SIGUSR2` would make sense.

t/10-virtio_terminal.t
```diff
@@ -25,21 +27,27 @@ sub prepare_pipes ($socket_path, $write_buffer = undef) {
...
+        open($fd_r, "<", $pipe_in)
           or die "Can't open in pipe for writing $!";
-        open(my $fd_w, ">", $pipe_out)
+        open($fd_w, ">", $pipe_out)
           or die "Can't open out pipe for reading $!";
 
         syswrite($fd_w, $write_buffer) if ($write_buffer);
+        $fd_w->flush();
 
+        kill 'USR2', getppid;
```
Martchus, line 50: Wouldn't we need the parent processes' PID here?

Martchus (changes requested):
I suppose this could generally work but see the inline comments.

okurz:
As the problem showed up on OBS and as you already changed the spec file the test is also executed on OBS. That's good. But as that was a sporadic problem I suggest to run much more often. E.g. you could try to run a for-loop within OBS checks temporarily as part of this PR. Or you create a custom OBS branch project checking out from git.

t/10-virtio_terminal.t
cfconrad: It actually doesn't matter that much. 
Sleep will interrupt if the SIGUSR2 will arise and if it was already emitted, then we do not sleep at all. Only a very slow chance left to hit the `sleep 1` unnecessary.
But sure, why not to wait 0.5.

t/10-virtio_terminal.t
cfconrad: note the **p**pid https://perldoc.perl.org/functions/getppid

cfconrad:
Thats what I currently do in here: https://build.opensuse.org/package/show/home:cfconrad:branches:devel:openQA/os-autoinst
And I use this script:
```bash
LOOP=0

osc_r() {
    osc r | grep openSUSE_Factory
}

is_building() {
    test $(osc_r | awk '{print $4}' | grep -P 'building|blocked|scheduled|finished' | wc -l) -gt 0
}

is_failed() {
    test $(osc_r | awk '{print $4}' | grep -v succeeded | wc -l) -ne 0
}

osc_rebuild() {
    for i in openSUSE_Factory_PowerPC openSUSE_Factory_ARM openSUSE_Factory; do
        osc rebuild -r $i
    done
    sleep 10;
}

msg() {
    echo "[$LOOP]" $@
}

while true; do
    LOOP=$(( $LOOP + 1 ));
    while is_building; do 
        msg "Wait for builds are ready"
        osc_r
        sleep 1;
    done

    if is_failed; then
        msg "Something failed!!!"
        osc_r
        exit 1 
    fi

    osc_rebuild
done
```

I'm on loop 16. Any hints to improve this?

okurz:
> I'm on loop 16. Any hints to improve this?

Great! Your approach is better than what I had in mind because you do rebuild the complete OBS job so you might hit different OBS workers with differing performance. Well, to improve you could first try to gather the fail-ratio *before* your fix so that you know what to test for :) But I guess it's ok

okurz (approved):
I have reviewed the CI runs and found no unhandled output and comparing to the last successful run of CI tests in master in https://github.com/os-autoinst/os-autoinst/runs/4340445436?check_suite_focus=true I see a runtime of 8s and we still have 8s in your PR

t/10-virtio_terminal.t
Martchus: Ah, I've missed the 2nd "p". Then it makes sense.

t/10-virtio_terminal.t
Martchus: If the sleep is interrupted anyways then the interval shouldn't matter much (except the interrupt happens between the evaluation of the loop condition and the execution of the loop body).

Martchus (approved):

PR #1872, opened 2021-12-02
Add execution time for failed test modules

foursixnine:
Currently when a test fails, we don't know how long did it take, or how long did the post fail hooks took, unless we look at the timestamps on the logs, displaying total execution fixes the first problem and printing a message in the logs for the post fail hooks, fixes the second.

VR: http://phobos.qa.suse.de/tests/3857623

```
[2021-12-02T18:15:47.191256+01:00] [debug] ||| finished gpg console (runtime: 527 s)
[2021-12-02T18:15:47.191297+01:00] [debug] ||| post fail hooks runtime: 474 s
```

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -4,7 +4,7 @@
 
 package basetest;
 
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 use autodie ':all';
 
 use bmwqemu ();
@@ -338,8 +338,7 @@ sub run_post_fail {
 
 sub execution_time { time - shift }
 
-sub compute_test_execution_time {
-    my ($self) = @_;
+sub compute_test_execution_time ($self) {
     # Set the execution time for a general time spent
     $self->{execution_time} = execution_time($self->{test_start_time});
     bmwqemu::diag(sprintf("||| finished %s %s (runtime: %d s)", $self->{name}, $self->{category}, $self->{execution_time}));
--- a/basetest.pm
+++ b/basetest.pm
@@ -46,6 +46,7 @@ sub new {
     $self->{autoinst_failures} = [];
     $self->{fatal_failure} = 0;
     $self->{execution_time} = 0;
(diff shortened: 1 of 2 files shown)


Martchus:
The VR doesn't look successful.

foursixnine:
> The VR doesn't look successful.

The clone is: http://phobos.qa.suse.de/tests/3857623, updated the referfence

basetest.pm
```diff
@@ -325,15 +328,26 @@ sub run_post_fail {
         # Read them now to not stumble upon them in the next module.
         $self->get_new_serial_output();
     }
+
     $self->fail_if_running();
+    $self->compute_test_execution_time();
+    my $post_fail_hook_execution_time = execution_time($post_fail_hook_start_time);
+    bmwqemu::diag(sprintf("||| post fail hooks runtime: %d s", $post_fail_hook_execution_time));
```
kalikiana, line 335: Would you mind adding a unit test for that?

basetest.pm
```diff
@@ -325,15 +328,26 @@ sub run_post_fail {
...
     die $msg . "\n";
 }
 
 sub execution_time { time - shift }
 
+sub compute_test_execution_time {
+    my ($self) = @_;
```
Martchus, line 342: It would be nice to use signatures in new code.

Martchus (commented):
The change is already fully covered by tests but maybe it makes sense to add an additional check somewhere within the tests.

basetest.pm
foursixnine: @Martchus I will try to add signatures on a separate commit, as no methods have them within this package, let's see how it goes.

basetest.pm
foursixnine: Done, now we look for the specific pfh message.

basetest.pm
foursixnine: We should be good to go then 🏎️

PR #1874, opened 2021-12-03
ci: Fix OpenQA::Test::CheckGitStatus exit status

perlpunk:
In the previous commit #1873  the `exit 1` was happening when $? was localized,
so the actual purpose of the module wasn't working anymore.

Now we just set the value of $?, I think that's the right way to do it.

Issue: https://progress.opensuse.org/issues/103422

### DIFF
--- a/t/lib/OpenQA/Test/CheckGitStatus.pm
+++ b/t/lib/OpenQA/Test/CheckGitStatus.pm
@@ -16,19 +16,22 @@ if ($CHECK_GIT_STATUS) {
 }
 
 sub check_status {
-    local $?;
-    chdir $cwd;
-    my $git = File::Which::which('git');
-    return unless $git;
-    my $cmd = 'git rev-parse --git-dir';
-    my $out = qx{$cmd};
-    return if $? != 0;
-    $cmd = 'git status --porcelain=v1 2>&1';
-    my @lines = qx{$cmd};
-    die "Problem running git:\n" . join '', @lines if $? != 0;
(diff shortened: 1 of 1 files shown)


Martchus (approved):

t/lib/OpenQA/Test/CheckGitStatus.pm
```diff
@@ -16,19 +16,22 @@ if ($CHECK_GIT_STATUS) {
...
+        my $out = qx{$cmd};
+        return if $? != 0;
+        $cmd = 'git status --porcelain=v1 2>&1';
+        @lines = qx{$cmd};
+        die "Problem running git:\n" . join '', @lines if $? != 0;
+    }
     if (@lines > 0) {
         Test::More::diag("Error: modified or untracked files\n" . join '', @lines);
-        exit 1;
+        $? = 1;
```
kalikiana, line 34: I need to ask the obvious, #1873 was passing CI - how do I know this will work? Can we check this anywhere?

Martchus: @perlpunk was able to reproduce the problem locally so I assume she's tested it locally.

perlpunk: Yes, I tested both cases locally now by
* adding a `fail()` to 13-os-utils.t to see if `prove` really fails if a test fails
* adding an untracked file to see if the status check still works

PR #1875, opened 2021-12-03
Streamline the use of "|||" log marker with new "modstate" function

okurz:
Previously we had a function "modstart" which was used only once to log
the start of a test module execution. A similar output but using custom
diag+sprintf was done at the end of a test module so the output looks
like

```
[2021-12-03T14:36:20.474509+01:00] [debug] ||| finished bootloader installation (runtime: 88 s)
[2021-12-03T14:36:20.475852+01:00] [debug] ||| starting rcshell_start tests/microos/rcshell_start.pm
```

of two consecutive test module lines but the second printed in bold and
blue font (not visible here).

This commit renames the "modstart" function with a more generic name
"modstate" and uses that in both above mentioned cases.

d5eb3309 motivated me to do this.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -361,7 +361,7 @@ sub runalltests {
         }
 
         my $name = $t->{name};
-        bmwqemu::modstart "starting $name $t->{script}";
+        bmwqemu::modstate "starting $name $t->{script}";
         $t->start();
 
         # avoid erasing the good vm snapshot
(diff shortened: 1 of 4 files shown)


perlpunk (approved):

kalikiana (approved):
Streamline meaning, rename the function and use it instead of diag

Nice tweak, though

Martchus (approved):

PR #1877, opened 2021-12-06
t: Prevent unnecessary diag output in 28-signalblocker.t

okurz:
8363ce36 introduced a change with a "diag" for something that is just
informative and not an error and hence should not be output in normal
terse test output.

### DIFF
--- a/t/28-signalblocker.t
+++ b/t/28-signalblocker.t
@@ -48,7 +48,7 @@ $SIG{CHLD} = sub { $received_sigchld += 1; note "received SIGCHLD $received_sigc
     # make the number of threads to spawn configurable
     my $thread_count = tinycv::default_thread_count();
     my $thread_count_for_testing = $ENV{OS_AUTOINST_TEST_THREAD_COUNT} || $thread_count;
-    diag "threads used: $thread_count_for_testing of $thread_count";
+    note "threads used: $thread_count_for_testing of $thread_count";
 
     tinycv::create_threads($thread_count_for_testing);
     $last_thread_count = thread_count;
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -203,6 +203,9 @@ export CI=1
 # account for sporadic slowness in build environments
 # https://progress.opensuse.org/issues/89059
 export OPENQA_TEST_TIMEOUT_SCALE_CI=20
+# Enable verbose test output as we can not store test artifacts within package
+# build environments in case of needing to investigate failures
+export PROVE_ARGS="--timer -v"
 cd %{__builddir}
 %cmake_build check-pkg-build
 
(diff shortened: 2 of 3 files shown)


perlpunk (approved):

Martchus (changes requested):
I'd like to keep it as it is important to know when figuring out why tests hang/fail (within a CI or OBS environment). The information isn't unnecessary at all. Note that otherwise there wouldn't necessarily be any information about the number of threads because when it breaks the code might just deadlock and cannot print any useful output anymore.

t/28-signalblocker.t
```diff
@@ -48,7 +48,7 @@ $SIG{CHLD} = sub { $received_sigchld += 1; note "received SIGCHLD $received_sigc
     # make the number of threads to spawn configurable
     my $thread_count = tinycv::default_thread_count();
     my $thread_count_for_testing = $ENV{OS_AUTOINST_TEST_THREAD_COUNT} || $thread_count;
-    diag "threads used: $thread_count_for_testing of $thread_count";
+    note "threads used: $thread_count_for_testing of $thread_count";
```
okurz, line 51: ok, Martchus explained me that he sees this line as really necessary because tests could run into timeouts and then it's unclear what the thread count was. We don't use this approach anywhere else but we discussed different options how to do better. We talked about saving the output into a debug log file and outputting that in a `END { }` hook but this won't ever be called if we have a deadlock. I could simulate a problem with

```diff
diff --git a/ppmclibs/tinycv_impl.cc b/ppmclibs/tinycv_impl.cc
index ac1f354e..c8386975 100644
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -360,10 +360,10 @@ void create_opencv_threads(int thread_count)
     std::mutex m;
     std::condition_variable cv;
     int threads_spawned = 0;
-    parallel_for_(Range(0, thread_count), RunFunctionInParallel([&] (const Range &) {
+    parallel_for_(Range(0, 100), RunFunctionInParallel([&] (const Range &) {
         // keep the thread idling until the expected number of threads has been spawned
         std::unique_lock<std::mutex> lock(m);
-        if (++threads_spawned >= thread_count) {
+        if (++threads_spawned >= 100) {
             cv.notify_all(); // note: The last thread releases all previously spawned threads.
         } else {
             cv.wait(lock);
```

and `make && prove -I. -v t/28-signalblocker.t` which makes t/28-signalblocker.t block indefinitely, and not be aborted by OpenQA::Test::TimeLimit . So our next suggestion is to implement a simple testing specific alarm signal handler in ppmclibs/tinycv_impl.cc that would output such debug information.

perlpunk: We could write to a debug logfile and upload it as an artififact in a later step, right?

Martchus: This is also about OBS runs. (In fact, if such a problem would ever occur again we'd likely see it again on OBS first.)

okurz: I added a commit to enable verbose test output in OBS. I think we had this originally but lost it over time. This should help to address the original requirement

Martchus (approved):

PR #1878, opened 2021-12-10
mergify: Add more useful checks

okurz:
While we still need to check explicitly for the success of checks (see
https://docs.mergify.com/conditions/#validating-all-status-checks) we
can specify to additionally not merge on any failures or pending checks.
This should cover any potential new checks which we introduce but not
mention in the mergify config explicitly.


okurz:
Updated to resolve merge conflict. Ready for second approval

Martchus (approved):

PR #1879, opened 2021-12-10
Makefile: Fix build dir when calling make with "-C" itself

okurz:


Makefile
```diff
@@ -1,13 +1,14 @@
 # This is a convenience Makefile wrapping cmake calls
 # All targets should be defined in CMake
 
+build := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/build
 .PHONY: all
 all: build/CMakeCache.txt
-	ninja -C build/ symlinks
+	ninja -C ${build} symlinks
```
Martchus, line 7: Maybe this needs quoting now?

Makefile
```diff
@@ -1,13 +1,14 @@
...
 
 .DEFAULT: build/CMakeCache.txt
-	ninja -C build/ $@
+	ninja -C ${build} $@
 
 build/CMakeCache.txt:
-	@mkdir -p build
-	@cmake -B build -S . -G Ninja
+	@mkdir -p ${build}
+	@cmake -B ${build} -S . -G Ninja
```
Martchus, line 14: Maybe this needs quoting now?

Martchus (commented):
I suppose you've tested it locally.

Makefile
okurz: You mean if spaces would be included? I have tested but only with space-less paths so far

okurz: I tried and found http://savannah.gnu.org/bugs/?712 as well as many references explaining that the best is to avoid spaces in filenames when using make. So no quoting "needed" because it does not do what you expect nor fix problems in make (unlike bash)

okurz:
> I suppose you've tested it locally.

Tested with

```
cd /tmp
make -C /path/to/os-autoinst
```

Martchus (approved):

PR #1881, opened 2021-12-13
Add hardware console log to IPMI and PowerVM backends

czerw:
Enhancement poo#103903: IPMI and PowerVM backends miss log from system
boot/shutdown or kernel crash. This feature enables direct capture of
ipmi/mkvterm sessions on IPMI and PowerVM backends. It is disabled by
default, managed by variable HARDWARE_CONSOLE_LOG. New log file is
hardware-console-log.txt.

Progress: https://progress.opensuse.org/issues/103903
Needles: none

Verification:
ipmi: http://10.100.12.105/tests/1536/logfile?filename=hardware-console-log.txt
spvm: http://10.100.12.105/tests/1530/logfile?filename=hardware-console-log.txt
hmc: http://10.100.12.105/tests/1537/logfile?filename=hardware-console-log.txt

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -93,7 +93,11 @@ sub do_start_vm ($self, @) {
     $self->get_mc_status;
     $self->restart_host unless $bmwqemu::vars{IPMI_DO_NOT_RESTART_HOST};
     $self->truncate_serial_file;
-    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', {persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 1});
+    my $sol = $testapi::distri->add_console(
+        'sol', 'ipmi-xterm',
+        {
+            persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 1,
+            log => $bmwqemu::vars{HARDWARE_CONSOLE_LOG} // 0});
     $sol->backend($self);
     return {};
 }
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -12,6 +12,7 @@ use base 'consoles::vnc_base';
 use IPC::Run ();
 require IPC::System::Simple;
 use Socket;
+use File::Path 'mkpath';
 use File::Which;
 use Time::Seconds;
 use testapi qw(get_var);
@@ -38,6 +39,10 @@ sub callxterm ($self, $command, $window_name) {
     die('Missing "Xvnc"') unless which('Xvnc');
     die('Missing "icewm"') unless which('icewm');
     die('Missing "xterm"') unless which('xterm');
+    if ($self->{args}->{log}) {
+        mkpath 'ulogs';
+        $command = "script -f ulogs/hardware-console-log.txt -c \"$command\"";
+    }
     eval { system("DISPLAY=$display $xterm_vt_cmd -title $window_name -e bash -c '$command' & echo \"xterm PID is \$!\""); };
     die "cant' start xterm on $display (err: $! retval: $?)" if $@;
 }
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -89,6 +89,10 @@ test_base_requires:
   qemu-x86:
   procps:
   python3-setuptools:
+  xterm-console:
+  xorg-x11-Xvnc:
+  icewm:
+  xterm:
 
 test_version_only_requires:
   perl(Mojo::IOLoop::ReadWriteProcess): '>= 0.28'
--- a/t/29-backend-ipmi.t
+++ b/t/29-backend-ipmi.t
@@ -13,10 +13,12 @@ use Test::Output qw(combined_like stderr_like);
 use Test::Warnings qw(:all :report_warnings);
 
 BEGIN { *backend::ipmi::system = sub { 1 } }
+BEGIN { *consoles::localXvnc::system = sub { "@_" =~ /hardware-console-log/ ? 1 : 0 } }
 
 use backend::ipmi;    # SUT
 
 $bmwqemu::vars{WORKER_HOSTNAME} = 'localhost';
+$bmwqemu::vars{"HARDWARE_CONSOLE_LOG"} = 1;
 ok my $backend = backend::ipmi->new(), 'backend can be created';
 $bmwqemu::vars{"IPMI_$_"} = "fake_$_" foreach qw(HOSTNAME USER PASSWORD);
 my @ipmi_cmdline = $backend->ipmi_cmdline;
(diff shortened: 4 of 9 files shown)


consoles/localXvnc.pm
```diff
@@ -38,6 +38,7 @@ sub callxterm ($self, $command, $window_name) {
     die('Missing "Xvnc"') unless which('Xvnc');
     die('Missing "icewm"') unless which('icewm');
     die('Missing "xterm"') unless which('xterm');
+    $command = $self->{args}->{log} ? "script -f hardware-console-log.txt -c \"$command\"" : $command;
```
Martchus, line 41: This would be a bit easier to read in my opinion:
```suggestion
    $command = "script -f hardware-console-log.txt -c \"$command\"" if $self->{args}->{log};
```

Martchus (commented):
I suppose this should work. The duplication for the different backends isn't nice but I suppose it is ok. `script` is part of `util-linux` so we likely don't have to add it as dependency.

czerw:
>The duplication for the different backends isn't nice but I suppose it is ok. 

Do you mean duplication in variables doc? Yeas, I was not happy as well, but not sure how to do it  better.

consoles/localXvnc.pm
czerw: Implemented.

Martchus (approved):

Martchus:
The pipeline error is unrelated:

```
(11/11) Installing: perl-App-cpanminus-1.7043-bp153.1.12.noarch [...........done]
Abort, retry, ignore? [a/r/i/...? shows all options] (a): a
File './repodata/2ec0d1f23dc67a2387797d873e045a03a9b89f0db8c421f59a4b5678fdbc9582-deltainfo.xml.gz' not found on medium 'http://download.opensuse.org/update/leap/15.3/sle/'
Error building the cache:
Warning: Skipping repository 'Update repository with updates from SUSE Linux Enterprise 15' because of the above error.
[repo-sle-update|http://download.opensuse.org/update/leap/15.3/sle/] Valid metadata not found at specified URL
History:
 - ABORT request: Aborting requested by user
 - Can't provide ./repodata/2ec0d1f23dc67a2387797d873e045a03a9b89f0db8c421f59a4b5678fdbc9582-deltainfo.xml.gz
```

In fact, I've just seen this error in a completely different context. Let's see whether a retry helps.

By the way, maybe you can also extend unit tests to cover this case.

backend/ipmi.pm
```diff
@@ -93,7 +93,11 @@ sub do_start_vm ($self, @) {
     $self->get_mc_status;
     $self->restart_host unless $bmwqemu::vars{IPMI_DO_NOT_RESTART_HOST};
     $self->truncate_serial_file;
-    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', {persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 1});
+    my $sol = $testapi::distri->add_console(
+        'sol', 'ipmi-xterm',
+        {
+            persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 1,
+            log => $bmwqemu::vars{HARDWARE_CONSOLE_LOG} // 0});
```
Martchus, line 100: Maybe tidy let's you do something like this (which would look less ugly):
```suggestion
    my %sol_args = (
        persistent => $bmwqemu::vars{IPMI_SOL_PERSISTENT_CONSOLE} // 1,
        log => $bmwqemu::vars{HARDWARE_CONSOLE_LOG} // 0,
    );
    my $sol = $testapi::distri->add_console('sol', 'ipmi-xterm', \%sol_args);
```

czerw:
> By the way, maybe you can also extend unit tests to cover this case.

But how? It depends on network and real hardware, to be honest i'm not good at this mysterious perl mock area :)

okurz (changes requested):
IMHO the duplication is acceptable. I would have approved if you would not have called the PR WIP+EXPERIMENTAL :) So, as I saw something working in the referenced openQA job I consider this great already and assume it's fine? What else do you plan to do before I would be able to approve?

Martchus:
There are already existing tests. For instance `t/29-backend-ipmi.t` checks the `ipmitool` invocation. You could just call `$backend->ipmi_cmdline` again with your new test variable being set and check whether `@ipmi_cmdline` now returns `script …`.

czerw:
>  What else do you plan to do before I would be able to approve?

I need to add unit tests.

okurz:
ok, cool. I would have accepted without but the suggestion from Martchus is good. I would have done the same after we would have merged this PR ;) Give it a try

czerw:
> There are already existing tests. For instance `t/29-backend-ipmi.t` checks the `ipmitool` invocation. You could just call `$backend->ipmi_cmdline` again with your new test variable being set and check whether `@ipmi_cmdline` now returns `script …`.

`$backend->ipmi_cmdline` contains only ipmitool command not  `script` at all. Newly added `script` command is part of `callxterm` function in `localXvnc.pm`, it is used in `eval` section, so it is not part of any return value. Should I somehow mock `eval` function ?

backend/ipmi.pm
czerw: I was just following "style" from spvm and hmc backends for this spart, would like to have it same.

t/29-backend-ipmi.t
```diff
@@ -36,18 +37,17 @@ $ipmi->redefine(ipmitool => sub { $_[1] =~ /power status/ ? $ipmitool_mock->ipmi
...
 $testapi::distri = distribution->new;
 
 ok $backend->do_start_vm, 'can call do_start_vm';
 ok $backend->do_stop_vm, 'can call do_stop_vm';
 ok !$backend->check_socket(undef), 'check_socket not returning true by default';
 ok $backend->get_mc_status, 'can call get_mc_status';
 
+is $testapi::distri->{consoles}->{sol}->{args}->{log}, '1';
+$testapi::distri->{consoles}->{sol}->{DISPLAY} = "display";
+ok !$testapi::distri->{consoles}->{sol}->callxterm('ipmi', "console"), "can create console with log enabled";
```
okurz, line 49: This fails in CI now with

```
8: 
8: #   Failed test at ./29-backend-ipmi.t line 47.
8: #          got: '0'
8: #     expected: '1'
8: Missing 'xterm-console' at /opt/consoles/localXvnc.pm line 37.
8: # Tests were run but no plan was declared and done_testing() was not seen.
8: # Looks like your test exited with 2 just after 14.
8: [13:31:37] ./29-backend-ipmi.t ...................... 
8: Dubious, test returned 2 (wstat 512, 0x200)
8: Failed 1/14 subtests 
```

because "xterm-console" is not installed in the development environment. I suggest you add the package "xterm-console" in dependencies.yaml after line 91 in the test requirements and call `make update-deps` and submit the changes as well

okurz (changes requested):

dependencies.yaml
```diff
@@ -89,6 +89,10 @@ test_base_requires:
   qemu-x86:
   procps:
   python3-setuptools:
+  xterm-console:
+  Xvnc:
```
okurz, line 93: s/Xvnc/xorg-x11-Xvnc/

Martchus (approved):

pevik (approved):

consoles/localXvnc.pm
pevik: > `script` is part of `util-linux` so we likely don't have to add it as dependency.

@Martchus FYI `script` has been used long time ago in `backend/svirt.pm:272

https://github.com/os-autoinst/os-autoinst/blob/ae94c4bd4c48eb9b8cc0e09b1728c886d0f27c95/backend/svirt.pm#L267-L279

PR #1883, opened 2021-12-13
tools: Properly call tests from root dir

okurz:
I think for a long time tests have been called from the t/ (or xt/)
directory because some tests relied on that but by now we should have
fixed all tests to be able to be called from the root folder of the
project (or any folder). With this change we call prove from the root
dir which should help us to get rid of some special workarounds as well
as allow to collect coverage information consistently from test code as
well.

### DIFF
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -42,7 +42,17 @@ db="$build_directory/cover_db"
 # set Perl module include path and coverage options
 export PERL5LIB="$source_directory:$source_directory/ppmclibs/blib/lib:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
 if [[ $WITH_COVER_OPTIONS ]]; then
-    select="-ignore,^/|^external/|/tmp|CheckGitStatus.pm"
+    # As described on
+    # https://stackoverflow.com/a/22733744
+    # whenever you put coverage into a coverage DB you should ensure that you
+    # are doing it from the same directory, so that Devel::Cover can match and
+    # locate the files in the coverage DB.
+    # Meaning that chdir in tests is harmful. We either try to run everything
+    # from the root source folder or everything from another folder,
+    # previously from within t/ but of course we could also run everything
+    # from the build folder or a temporary directory
+
+    select="-select,$source_directory,-ignore,^/|external|/tmp|/CheckGitStatus.pm"
     # add ' -MOpenQA::Test::PatchDeparse' for older OS versions to avoid warnings
     export PERL5OPT="-I$source_directory/t/lib -I$source_directory/external/os-autoinst-common/lib -MOpenQA::Test::CheckGitStatus -MDevel::Cover=-db,$db,-select,$select,-coverage,statement"
 fi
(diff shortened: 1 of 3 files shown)


okurz:
well, ok, there is a codecov report but it does not look complete. Compare e.g. https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/bmwqemu.pm showing 100% statement coverage with all (non-excluded) files labeled green vs. https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/1883/changes showing 96% coverage but no lines light up and showing some less covered lines. anyone?

autotest.pm
```diff
@@ -231,6 +231,8 @@ sub load_snapshot {
 
 sub _terminate () {
     close $isotovideo;    # uncoverable statement
+    use FindBin '$Bin';
+    chdir $Bin;
```
okurz, line 235: Trying to change to the root source folder just before creating a coverage report from subprocesses does not seem to make a difference. The coverage still reports the same as in before in this PR, with decreased coverage from files that we assume are only significantly covered from subprocesses

tools/invoke-tests
```diff
@@ -42,21 +42,16 @@ db="$build_directory/cover_db"
 # set Perl module include path and coverage options
 export PERL5LIB="$source_directory:$source_directory/external/os-autoinst-common/lib:$source_directory/ppmclibs:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
 if [[ $WITH_COVER_OPTIONS ]]; then
-    path_regex="^|$source_directory/|\\.\\./"
-    select="($path_regex)(OpenQA|backend|consoles|ppmclibs)/|($path_regex)isotovideo|($path_regex)[^/]+\.pm,-ignore,\.t|data/tests/|fake/tests/|/CheckGitStatus.pm|$prove_path"
-    export PERL5OPT="-I$source_directory/t/lib -MOpenQA::Test::CheckGitStatus -MDevel::Cover=-db,$db,-select,$select,-coverage,statement"
+    #path_regex="^|$source_directory/|\\./"
+    #select="($path_regex)(OpenQA|backend|consoles|ppmclibs)/|($path_regex)isotovideo|($path_regex)[^/]+\.pm,-ignore,external/|data/tests/|fake/tests/|/CheckGitStatus.pm|$prove_path"
+    #export PERL5OPT="-I$source_directory/t/lib -MOpenQA::Test::CheckGitStatus -MDevel::Cover=-db,$db,-select,$select,-coverage,statement"
+    export PERL5OPT="-I$source_directory/t/lib -MOpenQA::Test::CheckGitStatus -MDevel::Cover=-db,$db,-coverage,statement"
```
okurz, line 48: With most of these Devel::Cover options disabled I don't see error reports in the CI job anymore and the results look very sane. We should likely still exclude stuff like "external/"

okurz:
With https://github.com/os-autoinst/os-autoinst/pull/2074 @perlpunk did the awesome job to fix what I could not achieve over such long time :) I still updated this PR to see if there is any beneficial diff left?

Martchus:
I'd say there's no beneficial diff left.

PR #1884, opened 2021-12-13
Makefile: Fix initial call with test target

okurz:
If the build directory does not yet exist but make is called with a
target initially then the previous default rule did not correctly ensure
that the CMake build directory is correctly initialized.

Tested with

```
rm -rf build/ && make coverage TESTS=t/03-testapi.t && make && rm -rf build && make && rm -rf os-autoinst/build/ && make -C os-autoinst coverage TESTS=t/03-testapi.t && make -C os-autoinst && rm -rf os-autoinst/build/ && make -C os-autoinst
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,14 @@
 # This is a convenience Makefile wrapping cmake calls
 # All targets should be defined in CMake
 
-build := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/build
+build := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))build
 .PHONY: all
 all: build/CMakeCache.txt
 	ninja -C ${build} symlinks
 
-.DEFAULT: build/CMakeCache.txt
+# empty default to ensure build dir is created when make called with arguments
+Makefile: ;
+%: build/CMakeCache.txt
 	ninja -C ${build} $@
 
 build/CMakeCache.txt:
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -175,6 +175,8 @@ sed  -i 's/ my $thisversion = qx{git.*rev-parse HEAD}.*;/ my $thisversion = "%{v
 for i in 07-commands 13-osutils 14-isotovideo 18-qemu-options 18-backend-qemu 99-full-stack; do
     rm t/$i.t
 done
+# Remove test relying on a git working copy
+rm xt/30-make.t
 
 %build
 %define __builder ninja
--- a/xt/30-make.t
+++ b/xt/30-make.t
@@ -1,0 +1,27 @@
+#!/usr/bin/env perl
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+use Test::Output 'stderr_like';
+use Mojo::File qw(path tempdir);
+use Mojo::Util 'scope_guard';
+use FindBin '$Bin';
+use lib "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '20';
+
+my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
+my $srcdir = path("$Bin/..")->realpath;
+stderr_like { is qx{git -C $dir clone $srcdir os-autoinst}, '', 'prepare working copy with git' } qr/Cloning/, 'git clone';
+chdir "$dir/os-autoinst" or die "Failed to change directory to $dir/os-autoinst";
+my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
+is $?, 0, 'prepare working copy with git (exit code)';
+like qx{make -n}, qr/ninja.*symlinks/, 'build default';
+is $?, 0, 'build default (exit code)';
+like qx{rm -rf build && make help}, qr/help: HELP/, 'call specific make target initially';
+chdir '..' or die 'Failed to change directory to ..';
+like qx{rm -rf os-autoinst/build && make -C os-autoinst help}, qr/help: HELP/, 'call make with target initially from outside';
+like qx{make -C os-autoinst help}, qr/help: HELP/, 'call make again with target from outside';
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -9,6 +9,10 @@ all: build/CMakeCache.txt
 .DEFAULT: build/CMakeCache.txt
 	ninja -C ${build} $@
 
+# empty default to ensure build dir is created when make called with arguments
+%:
+	#
```
kalikiana, line 14: Please add something to CI that doesn't work without this fix. If I understand correctly it could be a simple thing like `make test TESTS=`

kalikiana (changes requested):

Makefile
okurz: I don't want to build a dependency circle with our tests testing the call of themselves. That does not make sense

kalikiana: I don't know what that means. I'm suggesting to make sure we call make targets in a way that reproduces the issue in CircleCI or GHA

okurz: Test added with https://github.com/os-autoinst/os-autoinst/pull/1884/commits/c2a17030e5f05198f062810ae52ee68cda3fb438 which is using a local git clone to a temporary directory to not mess up the original source directory

okurz:
Updated:
* Fixed the problem by going back to `%:` but with proper build rules and an exception for the Makefile file as target itself
* Added test for calling make on that level

okurz:
Anyone has a clue why it looks like we are in the wrong directory in
https://github.com/os-autoinst/os-autoinst/runs/4532211740?check_suite_focus=true#step:3:626

```
8: make: *** No targets specified and no makefile found.  Stop.
8: 
8: #   Failed test 'build default'
8: #   at ./30-make.t line 20.
8: #                   ''
8: #     doesn't match '(?^:Symlinking created executables into source directory)'
8: 
8: #   Failed test 'build default (exit code)'
```

Martchus:
Maybe an `ok chdir …, "changed to $dir/os-autoinst"; diag `tree "$dir/os-autoinst`;` can answer your question.

It seems a bit wasteful to conduct the build again just for checking the `Makefile`. Maybe you can mock the actual build or simply use the `Makefile` for conducting the CI build we do anyways (which would of course mean not being able to test all usages)?

okurz:
> Maybe an `` ok chdir …, "changed to $dir/os-autoinst"; diag `tree $dir/os-autoinst`; `` can answer your question.

ah, good point, `chdir` might fail. I added a check for that and also now explicitly clone to a dir "os-autoinst", might resolve to a different target dir otherwise.

> It seems a bit wasteful to conduct the build again just for checking the `Makefile`. Maybe you can mock the actual build

Hm, so far I consider it ok-ish. See for example

```
$ ts-cmd prove -v xt/30-make.t
00:00:00.707574 00:00:00.708323 xt/30-make.t .. 
00:00:00.707679 00:00:00.000091 ok 1 - prepare working copy with git
00:00:00.707977 00:00:00.000389 ok 2 - git clone
00:00:00.708331 00:00:00.000341 ok 3 - prepare working copy with git (exit code)
00:00:04.368668 00:00:03.660291 ok 4 - build default
00:00:04.368715 00:00:00.000055 ok 5 - build default (exit code)
00:00:05.286264 00:00:00.917573 ok 6 - call specific make target initially
00:00:06.247441 00:00:00.961140 ok 7 - call make with target initially from outside
00:00:06.259796 00:00:00.012341 ok 8 - call make again with target from outside
00:00:09.291614 00:00:03.031859 ok 9 - call make from outside without arguments
00:00:09.291784 00:00:00.000177 ok 10 - no (unexpected) warnings (via done_testing)
00:00:09.291976 00:00:00.000224 1..10
00:00:09.487447 00:00:00.195398 ok
00:00:09.487924 00:00:00.000499 All tests successful.
00:00:09.487961 00:00:00.000038 Files=1, Tests=10, 10 wallclock secs ( 0.02 usr  0.01 sys + 25.48 cusr  2.67 csys = 28.18 CPU)
00:00:09.487982 00:00:00.000020 Result: PASS
```

so overall in my environment 10s with the "default build without arguments" being the biggest contributors. We could of course just leave these steps out and just keep the rather cheap "make help" calls in. I guess it would be ok if we just call `make -n` to have dry runs and check if the would-be called commands look sane:

```
ts-cmd prove -v xt/30-make.t
00:00:00.723258 00:00:00.739274 xt/30-make.t .. 
00:00:00.723345 00:00:00.000089 ok 1 - prepare working copy with git
00:00:00.723587 00:00:00.000344 ok 2 - git clone
00:00:00.723914 00:00:00.000320 ok 3 - prepare working copy with git (exit code)
00:00:00.727798 00:00:00.003877 ok 4 - build default
00:00:00.728126 00:00:00.000307 ok 5 - build default (exit code)
00:00:01.648767 00:00:00.920650 ok 6 - call specific make target initially
00:00:02.578487 00:00:00.929701 ok 7 - call make with target initially from outside
00:00:02.590753 00:00:00.012280 ok 8 - call make again with target from outside
00:00:02.595031 00:00:00.004270 ok 9 - call make from outside without arguments
00:00:02.595675 00:00:00.000645 ok 10 - no (unexpected) warnings (via done_testing)
00:00:02.595830 00:00:00.000172 1..10
00:00:02.800766 00:00:00.204930 ok
00:00:02.801211 00:00:00.000461 All tests successful.
00:00:02.801254 00:00:00.000032 Files=1, Tests=10,  3 wallclock secs ( 0.03 usr  0.00 sys +  1.96 cusr  0.77 csys =  2.76 CPU)
00:00:02.801274 00:00:00.000012 Result: PASS
```

Ok, updated accordingly.

> or simply use the `Makefile` for conducting the CI build we do anyways (which would of course mean not being able to test all usages)?

You mean in tools/container_run_ci which we call from GHA pipelines? Hm, we could. But we use slightly different parameters in there, e.g. `CMAKE_BUILD_TYPE=Release`

Anyway, updated to not really conduct any build anymore as explained above

okurz:
@kalikiana can you approve?

Martchus (approved):

kalikiana (approved):

PR #1887, opened 2021-12-16
WIP Fix for Devel::Cover

perlpunk:
eval $code;

is called several times, and $code is different in each case, and as a result, has different number of statements.

This leads to a coverage report which looks wrong, reporting many many
statements for that line.

Let's see how codecov numbers change in contrast to Devel::Cover numbers

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -96,24 +96,31 @@ sub loadtest {
     my $fullname = "$category-$name";
     # perl code generating perl code is overcool
     my $code = "package $name;";
-    $code .= "use lib '.';" unless path($casedir)->is_abs;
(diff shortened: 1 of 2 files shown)


perlpunk:
So compared to the last run in master:
```
/opt/autotest.pm                             90.3   90.3
```
the Devel::Cover report for autotest dropped:
```
/opt/autotest.pm                             80.5   80.5
```
but still the line coverage reported by codecov is the same (except for my changed/added lines of course).

Before the report would look something like this, with many lines for the `eval $code` statement (cut down to a few here in my local tests):
devel-cover-extra-statements

PR #1889, opened 2021-12-21
Automatically detect KVM in qemu if available

okurz:
Previously KVM was expected to be present when running QEMU based tests
unless the test variable "QEMU_NO_KVM" was set. Any jobs running in an
environment where KVM is not available would fail. This commit now
automatically looks up the KVM device file while keeping the
configuration option to disable the use of KVM. Also changing the two
test modules t/18-qemu-options.t and t/99-full-stack.t to not hardcore
QEMU_NO_KVM as they can benefit from faster runtimes in KVM-enabled
environments as well. This additionally helps to actually test
KVM-accelerated virtualization environments directly.

Related progress issue: https://progress.opensuse.org/issues/101262

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -876,7 +876,7 @@ sub start_qemu ($self) {
             }
         }
 
-        sp('enable-kvm') unless $vars->{QEMU_NO_KVM};
+        sp('enable-kvm') if -r '/dev/kvm' && !$vars->{QEMU_NO_KVM};
         sp('no-shutdown');
 
         if ($vars->{VNC}) {
--- a/t/18-qemu-options.t
+++ b/t/18-qemu-options.t
@@ -35,7 +35,6 @@ my @common_options = (
     ARCH => 'i386',
     BACKEND => 'qemu',
     QEMU => 'i386',
-    QEMU_NO_KVM => 1,
     QEMU_NO_TABLET => 1,
     QEMU_NO_FDC_SET => 1,
     CASEDIR => "$data_dir/tests",
--- a/t/99-full-stack.t
+++ b/t/99-full-stack.t
@@ -35,7 +35,6 @@ print $var <<EOV;
    "ARCH" : "i386",
    "BACKEND" : "qemu",
    "QEMU" : "i386",
-   "QEMU_NO_KVM" : "1",
    "QEMU_NO_TABLET" : "1",
    "QEMU_NO_FDC_SET" : "1",
    "CASEDIR" : "$data_dir/tests",


mimi1vx (approved):

backend/qemu.pm
```diff
@@ -876,7 +876,7 @@ sub start_qemu ($self) {
             }
         }
 
-        sp('enable-kvm') unless $vars->{QEMU_NO_KVM};
+        sp('enable-kvm') if -r /dev/kvm && !$vars->{QEMU_NO_KVM};
```
perlpunk, line 879:
```suggestion
        sp('enable-kvm') if -r '/dev/kvm' && !$vars->{QEMU_NO_KVM};
```

backend/qemu.pm
```diff
@@ -876,7 +876,7 @@ sub start_qemu ($self) {
             }
         }
 
-        sp('enable-kvm') unless $vars->{QEMU_NO_KVM};
+        sp('enable-kvm') if -r '/dev/kvm' && !$vars->{QEMU_NO_KVM};
```
okurz, line 879: There was https://github.com/os-autoinst/os-autoinst/pull/1890#discussion_r773273819 "So what happens if there's no /dev/kvm device? It gets switched off silently?"

To give the answer here: "Yes" :)

I am wondering, maybe we should keep it more explicit

kalikiana: My issue is its being coupled with `QEMU_NO_KVM`. I would not be opposed to automatically enabling or disabling kvm (and making sure it's clear from the logs).

kalikiana: In other words, we have a very explicit variable and now you'e adding heuristics to it. This feels like we're asking for confusion.

kalikiana: It would be a lot easier if we had something like `QEMU_KVM=auto`

okurz: yes. I guess you are right

okurz:
Tested locally:

```
$ echo "No KVM" && prove -I. t/99-full-stack.t && git stash && echo "with KVM" && prove -I. t/99-full-stack.t 
No KVM
t/99-full-stack.t .. ok    
All tests successful.
Files=1, Tests=25, 180 wallclock secs ( 0.09 usr  0.00 sys + 69.06 cusr  4.47 csys = 73.62 CPU)
Result: PASS
Saved working directory and index state WIP on fullstack_hardware_test2: f87733af README: Add a section how to verify a runtime environment
with KVM
t/99-full-stack.t .. ok    
All tests successful.
Files=1, Tests=25, 162 wallclock secs ( 0.09 usr  0.00 sys + 42.59 cusr  5.85 csys = 48.53 CPU)
Result: PASS
```

so not a big difference but still

backend/qemu.pm
okurz: but wait, I am only changing the default case when the variable `QEMU_NO_KVM` isn't set.

pevik (approved):
IMHO good improvement.

kalikiana (approved):

okurz:
Ok so let's try it this way

PR #1890, opened 2021-12-21
README: Add a section how to verify a runtime environment

okurz:
Related progress issue: https://progress.opensuse.org/issues/101262

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -157,6 +157,20 @@ programming language is provided.
 Find more details about how to write tests on
 http://open.qa/docs/#_how_to_write_tests
 
+== Verifying a runtime environment
+
+To check if your hardware is able to successfully execute os-autoinst based
+tests one can execute openQA tests, all the development tests or simply call
+something like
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

backend/qemu.pm
```diff
@@ -876,7 +876,7 @@ sub start_qemu ($self) {
             }
         }
 
-        sp('enable-kvm') unless $vars->{QEMU_NO_KVM};
+        sp('enable-kvm') if -r '/dev/kvm' && !$vars->{QEMU_NO_KVM};
```
kalikiana, line 879: So what happens if there's no `/dev/kvm` device? It gets switched off silently?

okurz: Yes

okurz: I actually moved this change out to https://github.com/os-autoinst/os-autoinst/pull/1889

kalikiana (approved):

PR #1891, opened 2021-12-22
dist: Move swtpm "files" out of unnecessary ifarch

kalikiana:
The swtpm package is built for all architectures. Yet the "files" is
not.

Fixes: https://progress.opensuse.org/issues/103683

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -257,8 +257,8 @@ cd %{__builddir}
 %ifarch x86_64
 %files qemu-kvm
 %files qemu-x86
-%files swtpm
 %endif
+%files swtpm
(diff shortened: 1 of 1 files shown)


okurz (approved):
I wrongly assumed you would have put the package in https://github.com/os-autoinst/os-autoinst/pull/1891/files#diff-2bf7294cf30e1f4bda6077c3c165bfbe7b6b67fed2d3afc7e4fe02664ae1f060R146 in a "if x86_64" but that's not the case so your change looks good. I also checked on OBS and confirmed that the package can be built against aarch64 and ppc64le, see https://build.opensuse.org/package/binaries/security/swtpm/openSUSE_Leap_15.3

perlpunk (approved):

PR #1893, opened 2021-12-29
backend: do_extract_assets takes two arguments

kalikiana:

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -427,7 +427,7 @@ sub can_handle ($self, @) {
     return;    # sorry, no
 }
 
-sub do_extract_assets ($self) { notimplemented }
+sub do_extract_assets ($self, $args) { $self->notimplemented }
 
 sub is_shutdown ($self, @) { -1 }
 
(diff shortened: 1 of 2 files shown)


backend/baseclass.pm
```diff
@@ -427,7 +427,7 @@ sub can_handle ($self, @) {
     return;    # sorry, no
 }
 
-sub do_extract_assets ($self) { notimplemented }
+sub do_extract_assets ($self, $@) { notimplemented }
```
perlpunk, line 430:
```suggestion
sub do_extract_assets ($self, $args) { notimplemented }
```

kalikiana: We don't give names to arguments in stubs currently - but if you don't mind inconsistency I'm happy to change it

kalikiana: c.f.

    backend/null.pm
    sub do_extract_assets ($self, @) { }

perlpunk: Well, `$@` was a syntax error. And using `@` says it can have more than two arguments, which is wrong. So if you don't want to give it a name, use `$`.

kalikiana:
#1894 addresses this more completely

PR #1895, opened 2021-12-29
svirt: Implement do_extract_assets

kalikiana:
See: https://progress.opensuse.org/issues/104520

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -128,7 +128,7 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} =~ m/^(qemu|generalhw)$/;
+    return undef unless $bmwqemu::backend->{backend}->can('do_extract_assets');
     my @toextract;
     my $nd = $bmwqemu::vars{NUMDISKS};
     if ($command_handler->test_completed) {
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -5,9 +5,13 @@
 package backend::svirt;
 use Mojo::Base 'backend::virt', -signatures;
 use File::Basename;
+use File::Path 'mkpath';
 use IO::Scalar;
 use Time::HiRes 'usleep';
 use bmwqemu;
+use osutils qw(runcmd);
+
+use constant IMAGE_STORAGE => '/var/lib/libvirt/images/';
 
 use constant SERIAL_CONSOLE_DEFAULT_PORT => 0;
 use constant SERIAL_CONSOLE_DEFAULT_DEVICE => 'console';
@@ -165,6 +169,25 @@ sub load_snapshot ($self, $args) {
     bmwqemu::diag "LOAD snapshot $snapname to $vmname, return code=$rsp";
     $self->die if $rsp;
     return $post_load_snapshot_command;
+}
+
+sub do_extract_assets ($self, $args) {
+    my $format = $args->{format};
+    return undef if (($format ne 'raw') and ($format ne 'qcow2'));
+
+    my $first_hdd = $bmwqemu::vars{S390_ZKVM} ? 'a' : 'b';
+    my $name = $args->{name};
+    my $hdd_num = $args->{hdd_num} - 1;
+    my $vmname = $self->console('svirt')->name;
+    my $svirt_img_name = IMAGE_STORAGE . $vmname . chr(ord($first_hdd) + $hdd_num) . '.img';
+    my $img_dir = $args->{dir};
+    mkpath($img_dir);
+
+    my @args = ();
+    my $qemu_compress_qcow = $bmwqemu::vars{QEMU_COMPRESS_QCOW2} // 1;
+    push @args, '-c' if $qemu_compress_qcow;
+    runcmd('nice', 'ionice', 'qemu-img', 'convert', '-p', '-O', $format, $svirt_img_name, "$img_dir/$name", @args);
+    return undef;
 }
 
 sub get_ssh_credentials ($self, $domain = 'default') {
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -17,8 +17,10 @@ use distribution;
 use Net::SSH2;
 use testapi qw(get_var get_required_var check_var set_var);
 use backend::svirt qw(SERIAL_CONSOLE_DEFAULT_PORT SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT);
-use Mojo::File 'tempdir';
+use Mojo::File qw(tempdir path);
 use Mojo::Util qw(scope_guard);
+use OpenQA::Isotovideo::Utils qw(handle_generated_assets);
+use OpenQA::Isotovideo::CommandHandler;
 
 my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
 chdir $dir;
@@ -127,7 +129,7 @@ subtest 'starting VMware console' => sub {
     is_deeply \@cmds, [
         $s . ' destroy openQA-SUT-1 |& grep -v "\\(failed to get domain\\|Domain not found\\)"',
         $s . ' undefine --snapshots-metadata openQA-SUT-1 |& grep -v "\\(failed to get domain\\|Domain not found\\)"',
-        $s . ' define /var/lib/libvirt/images/openQA-SUT-1.xml',
+        $s . ' define ' . backend::svirt::IMAGE_STORAGE . 'openQA-SUT-1.xml',
         'echo bios.bootDelay = \\"10000\\" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
         $s . ' start openQA-SUT-1 2> >(tee /tmp/os-autoinst-openQA-SUT-1-stderr.log >&2)',
         $s . ' dumpxml openQA-SUT-1'
@@ -451,7 +453,7 @@ subtest 'Method consoles::sshVirtsh::add_disk()' => sub {
     subtest 'family svirt-xen-hvm' => sub {
         set_var(VIRSH_VMM_FAMILY => 'xen');
         set_var(VIRSH_VMM_TYPE => 'hvm');
-        my $basedir = '/var/lib/libvirt/images/';
+        my $basedir = backend::svirt::IMAGE_STORAGE;
 
         my $svirt = consoles::sshVirtsh->new('svirt');
         $svirt->backend(backend::baseclass->new);
@@ -590,7 +592,7 @@ subtest 'Method consoles::sshVirtsh::add_disk()' => sub {
     subtest 'family kvm' => sub {
         set_var(VIRSH_VMM_FAMILY => 'kvm');
         set_var(VIRSH_VMM_TYPE => undef);
-        my $basedir = '/var/lib/libvirt/images/';
+        my $basedir = backend::svirt::IMAGE_STORAGE;
 
         my $svirt = consoles::sshVirtsh->new('svirt');
         $svirt->backend(backend::baseclass->new);
@@ -740,6 +742,35 @@ subtest 'get_wait_still_screen_on_here_doc_input' => sub {
     is($svirt->get_wait_still_screen_on_here_doc_input({}) > 0, 1, 'wait_still_screen on here doc is set for vmware');
     set_var(VIRSH_VMM_FAMILY => 'kvm');
     is($svirt->get_wait_still_screen_on_here_doc_input({}), 0, 'wait_still_screen on here doc is not set for kvm');
+};
+
+# mock backend/driver
+{
+    package FakeBackendDriver;
+    sub new ($class, $name) {
+        my $self = bless({class => $class}, $class);
+        require "backend/$name.pm";
+        $self->{backend} = "backend::$name"->new();
+        return $self;
+    }
+    sub extract_assets ($self, $args) {
+        $self->{backend}->do_extract_assets($args);
+    }
+}
+
+subtest 'upload assets' => sub {
+    my $osutils_mock = Test::MockModule->new('osutils');
+    $osutils_mock->redefine(run => sub (@cmd) {
+            is join(' ', @cmd), 'nice ionice qemu-img convert -p -O qcow2 ' . backend::svirt::IMAGE_STORAGE . 'barfoob.img assets_public/test.qcow2 -c', 'got expected command';
+            return 0, '';
+    });
+
+    my $command_handler = OpenQA::Isotovideo::CommandHandler->new();
+    $bmwqemu::vars{NUMDISKS} = 1;
+    $bmwqemu::vars{FORCE_PUBLISH_HDD_1} = 'test.qcow2';
+    $bmwqemu::backend = FakeBackendDriver->new('svirt');
+    my $return_code = handle_generated_assets($command_handler, 1);
+    is $return_code, 0, 'The asset was uploaded successfully';
 };
 
 done_testing;
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -432,8 +432,6 @@ sub can_handle ($self, @) {
     return;    # sorry, no
 }
 
-sub do_extract_assets ($self, $args) { $self->notimplemented }
-
 sub is_shutdown ($self, @) { -1 }
 
 sub save_memory_dump ($self, $args) { $self->notimplemented }
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -430,7 +430,7 @@ sub add_disk ($self, $args) {
     my $cdrom = $args->{cdrom};
     my $name = $self->name;
     my $file = $name . $args->{dev_id} . ($self->vmm_family eq 'vmware' ? '.vmdk' : '.img');
-    my $basedir = '/var/lib/libvirt/images/';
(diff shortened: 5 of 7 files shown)


perlpunk:
And what about other backend classes?

kalikiana:
Since it turned out the clean solution would be to port [tests/shutdown/svirt_upload_assets.pm#](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/c27b0d46363113b1ec1506128d7b8687fc7aed1a/tests/shutdown/svirt_upload_assets.pm#L15) I put the proof of concept here. Not tested or expected to work for now.

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
     return $post_load_snapshot_command;
 }
 
+sub do_extract_assets ($self, $args) {
+    my $first_hdd = $bmwqemu::vars{S390_ZKVM} ? 'a' : 'b';
+    my $name = $args->{name};
+    die "No name given" unless $args->{name};
+    $name =~ /\.([[:alnum:]]+)$/;
```
perlpunk, line 182: This line can be removed, the result is not used, and the content is alraedy in `$args->{format}`

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
...
+    die "No HDD specified" unless $args->{hdd_num};
+    my $hdd_num = $args->{hdd_num} - 1;
+    die "No format specified" unless $args->{format};
+    my $format = $args->{format};
+    return undef if (($format ne 'raw') and ($format ne 'qcow2'));
+    my $svirt = $self->select_console('svirt');
+    my $svirt_name = $svirt->name . chr(ord($first_hdd) + $hdd_num);
+
+    my $image_storage = '/var/lib/libvirt/images';
```
perlpunk, line 191: I think the asset has to be stored in `$args->{dir}` to be picked up

perlpunk: Sorry, I meant that the final file needs to be in `$args->{dir}`

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
...
+    my $svirt_img_name = $image_storage . '/' . $args->{svirt_name} . '.img';
```
perlpunk, line 192: I think `$args->{svirt_name}` is not set. `$args` comes from `OpenQA::Isotovideo::Utils::handle_generated_assets`

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
...
+    my $first_hdd = $bmwqemu::vars{S390_ZKVM} ? 'a' : 'b';
+    my $name = $args->{name};
+    die "No name given" unless $args->{name};
+    $name =~ /\.([[:alnum:]]+)$/;
+    die "No HDD specified" unless $args->{hdd_num};
+    my $hdd_num = $args->{hdd_num} - 1;
+    die "No format specified" unless $args->{format};
+    my $format = $args->{format};
+    return undef if (($format ne 'raw') and ($format ne 'qcow2'));
+    my $svirt = $self->select_console('svirt');
```
perlpunk, line 188: I think the usage is:
`$self->select_console({testapi_console => 'svirt'});`

perlpunk: Maybe `select_console` isn't even needed and just
`my $svirt_name = $self->console('svirt')->name . ...;`

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
...
+    my $svirt_name = $svirt->name . chr(ord($first_hdd) + $hdd_num);
+
+    my $image_storage = '/var/lib/libvirt/images';
+    my $svirt_img_name = $image_storage . '/' . $args->{svirt_name} . '.img';
+
+    my @args = ();
+    my $cmd = "";
+    my $qemu_compress_qcow = $bmwqemu::vars{QEMU_COMPRESS_QCOW2} // 1;
```
perlpunk, line 196: I know it's from the original code, but using a `QEMU...` named variable here seems odd?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -128,7 +128,6 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} eq 'qemu';
```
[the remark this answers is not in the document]
perlpunk: `if backend::foo->can('do_extract_assets')` should be enough to check, no extra method needed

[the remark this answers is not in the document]
kalikiana: Shouldn't it be enough to not set `PUBLISH_HDD_1` and friends in the settings?

backend/svirt.pm
kalikiana: Actually I'll use `my $vmname = $self->console('svirt')->name` here like elsewhere, so this is consistent

OpenQA/Isotovideo/Utils.pm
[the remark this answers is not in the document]
Martchus: @perlpunk That's also a good idea but requires to get rid of `sub do_extract_assets ($self, $args) { $self->notimplemented }` in `baseclass.pm` of course.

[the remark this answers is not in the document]
perlpunk: oh right. stupid me.
It would need to be `if defined &backend::svirt::do_extract_assets`
This would also need to be fixed in https://github.com/os-autoinst/os-autoinst/pull/1895
Or we really get rid if the baseclass implementation, right.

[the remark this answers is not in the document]
Martchus: I actually like the idea because

> Or we really get rid if the baseclass implementation, right.

doesn't seem too intrusive to me. I doubt that so far some code has tried to invoke `do_extract_assets` looking for the "not implemented" exception.

backend/svirt.pm
kalikiana: it is `qemu-img`, though

perlpunk: ah ok

backend/svirt.pm
```diff
@@ -174,6 +176,32 @@ sub load_snapshot ($self, $args) {
...
+    return undef if (($format ne 'raw') and ($format ne 'qcow2'));
+
+    my $first_hdd = $bmwqemu::vars{S390_ZKVM} ? 'a' : 'b';
+    my $name = $args->{name};
+    my $hdd_num = $args->{hdd_num} - 1;
+    my $vmname = $self->console('svirt')->name;
+    my $svirt_name = $vmname . chr(ord($first_hdd) + $hdd_num);
+    my $img_dir = $args->{dir};
+    mkpath($img_dir);
+    my $svirt_img_name = "$img_dir/$svirt_name.img";
```
perlpunk, line 195: This is the place where the img file is stored, but in `$args->{dir}` we are getting the path where the final file should be stored.
Compare the original code https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/shutdown/svirt_upload_assets.pm#L22
```
    my $image_storage = '/var/lib/libvirt/images';
    my $svirt_img_name = $image_storage . '/' . $args->{svirt_name} . '.img';

    my $cmd = "nice ionice qemu-img convert -p -O $format $svirt_img_name $image_storage/$name";

    upload_asset("$image_storage/$name", 1, 1);
```

kalikiana: Right. I was/am trying to match `qemu.pm` which only calls `export_blockdev_images` which uses `qemu-img` with `$args->{dir}` as the location... there's no upload code there as far as I can tell. I do have a feeling I might be missing something.

kalikiana: In my test I'm checking the expected qemu-img command. `t/18-qemu` doesn't test `do_extract_assets` as such afair. Ideally I'd know what needs to be unit tested, to then make it do the right thing.

kalikiana: I considered using `OpenQA::Qemu::Proc` directly, but that seems to require the drives and block devices... only to generate the qemu-img command in the end - although I'm leaning towards extracting `Qemu/DriveDevice::gen_qemu_img_convert` which then could be shared?

perlpunk: I'll try to write down what I mean:
```perl
my $image_storage = '/var/lib/libvirt/images'; # maybe there is already a var or constant we can use, though
my $target_dir = $args->{dir}; # will contain the value 'assets_public'
my $svirt_img_name = "$image_storage/$svirt_name.img"; # this is where the source file is located
my $cmd = "nice ionice qemu-img convert -p -O $format $svirt_img_name $target_dir/$name"; # result file will end up in 'assets_public'
```

t/22-svirt.t
```diff
@@ -699,4 +699,16 @@ subtest 'get_wait_still_screen_on_here_doc_input' => sub {
...
+subtest 'Upload assets' => sub {
+    my $osutils_mock = Test::MockModule->new('osutils');
+    $osutils_mock->redefine(run => sub {
+            my @cmd = @_;
+            is 'nice ionice qemu-img convert -p -O qcow2 assets/barfoob.img assets/foo -c', join(' ', @cmd), 'got expected command';
+            return 0, '';
+    });
+
+    my $args = {name => 'foo', hdd_num => 1, format => 'qcow2', dir => 'assets'};
+    is($svirt->do_extract_assets($args), undef, 'Assets extracted');
```
perlpunk, line 711: This is the code for qemu:
https://github.com/os-autoinst/os-autoinst/blob/master/t/14-isotovideo.t#L178
```
    combined_like {
        $return_code = handle_generated_assets($command_handler, 1)
    } qr/Requested to force the publication/, 'forced publication of asset';
    is $return_code, 0, 'The asset was uploaded successfully' or die path(bmwqemu::STATE_FILE)->slurp;
    my $force_publish_asset = $pool_dir . '/assets_public/force_publish_test.qcow2';
    ok(-e $force_publish_asset, 'test.qcow2 image exists');
    ok(!-e $pool_dir . '/assets_public/publish_test.qcow2', 'the asset defined by PUBLISH_HDD_X would not be generated in an incomplete job');
```

kalikiana: Thanks! I changed my test now to also use `handle_generated_assets` which I think is better than directly calling into the backend

backend/svirt.pm
kalikiana: I introduced a constant `IMAGE_STORAGE` so we don't need to hard-code this folder. And using `dir` for the target should be fine, since the comment in the distri code saying `private asset` is not what the code does.

OpenQA/Isotovideo/Utils.pm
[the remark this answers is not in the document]
kalikiana: I have `defined &backend::svirt::do_extract_assets`, so no need to remove the "notimplemented" function - I assume we'd rather keep it if we can do that it can raise an error in other backends.

backend/svirt.pm
```diff
@@ -174,6 +175,30 @@ sub load_snapshot ($self, $args) {
...
+
+    my $image_storage = '/var/lib/libvirt/images';
+    my $svirt_img_name = $image_storage . '/' . $args->{svirt_name} . '.img';
+
+    my @args = ();
+    my $cmd = "";
+    my $qemu_compress_qcow = $bmwqemu::vars{QEMU_COMPRESS_QCOW2} // 1;
+    push @args, '-c' if $qemu_compress_qcow;
+    mkpath($image_storage);
+    runcmd('nice', 'ionice', 'qemu-img', 'convert', '-p' . '-O', $format . $svirt_img_name, "$image_storage/$name", @args);
```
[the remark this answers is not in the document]
kalikiana: As mentioned in another comment, I think using the existing Qemu module doesn't make sense. So I'd rather leave refactoring for a separate PR.

backend/svirt.pm
kalikiana: We have `IMAGE_STORAGE` defined in the backend now.

OpenQA/Isotovideo/Utils.pm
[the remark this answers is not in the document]
perlpunk: Using `defined` should do it for now.
But `can()` would be more correct because it uses inheritance. Two backend classes could share the same code.
So removing the "notimplemented" methods, doing a `can()` check instead and having an informative message would save the same purpose as the "notimplemented" methods.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -128,7 +128,7 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} =~ m/^(qemu|generalhw)$/;
+    return unless defined &backend::svirt::do_extract_assets;
```
perlpunk, line 131: That's now completely ignoring what kind of backend we have :)

kalikiana: You're right 🤦🏾 I was thinking to change it but totally forgot.

kalikiana: Sorry about the delay. Turns out this way didn't work with the backend object, so I changed it to "can" like you recommended before since I had to change it anyway.

perlpunk: > Turns out this way didn't work with the backend object

Can you elaborate on why/how this didn't work?

t/22-svirt.t
```diff
@@ -742,4 +744,33 @@ subtest 'get_wait_still_screen_on_here_doc_input' => sub {
...
+    }
+    sub extract_assets ($self, $args) {
+        $self->{backend}->do_extract_assets($args);
+    }
+}
+
+subtest 'upload assets' => sub {
+    my $osutils_mock = Test::MockModule->new('osutils');
+    $osutils_mock->redefine(run => sub (@cmd) {
+            is 'nice ionice qemu-img convert -p -O qcow2 ' . backend::svirt::IMAGE_STORAGE . 'barfoob.img assets_public/test.qcow2 -c', join(' ', @cmd), 'got expected command';
```
perlpunk, line 764: This looks wrong. The usage is
`is $got, $expected, $test_name;`

OpenQA/Isotovideo/Utils.pm
perlpunk: e.g. my suggestion would have been:
```
return if ($bmwqemu::vars{BACKEND} ne `qemu` and
  ($bmwqemu::vars{BACKEND} ne `svirt` or not defined &backend::svirt::do_extract_assets));
```

kalikiana: > Can you elaborate on why/how this didn't work?

`&$bmwqemu::backend::backend::do_extract_assets` gives me `Use of uninitialized value $bmwqemu::backend::backend::do_extract_assets in subroutine dereference` and without the deref it doesn't work.

perlpunk: That's not what I suggested in the first place, but checking `&backend::svirt::do_extract_assets` **in case** the backend is `svirt`.
Checking something like that with a variable name needs a bit more complicated syntax.

perlpunk: Anyway, since you deleted the "not implemented" method in this PR, the `can()` is ok.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -128,7 +128,7 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} =~ m/^(qemu|generalhw)$/;
+    return undef unless $bmwqemu::backend->{backend}->can('do_extract_assets');
```
perlpunk, line 131: In case the method is not implemented, this will return silently.
Before it would die.
Maybe at least a warning or info should be written in the log?

kalikiana: Well, it would never die in practice because the backends live in the same codebase. Also, a log message here would be visible in every single test run unless this feature is implemented - do we want to implement it in all backends? Otherwise I'd consider this harmful.

perlpunk: Ok, I should have noted that the warning should only appear if `PUBLISH_HDD_*` is set.
Not sure how we handle other cases where special vars trigger certain code.

okurz:
Merged https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13974

OpenQA/Isotovideo/Utils.pm
kalikiana: Ack. I'd consider that a new feature, though? And rather do it in another branch with a test to check the warning is logged.

t/22-svirt.t
```diff
@@ -742,4 +744,33 @@ subtest 'get_wait_still_screen_on_here_doc_input' => sub {
...
+    }
+    sub extract_assets ($self, $args) {
+        $self->{backend}->do_extract_assets($args);
+    }
+}
+
+subtest 'upload assets' => sub {
+    my $osutils_mock = Test::MockModule->new('osutils');
+    $osutils_mock->redefine(run => sub (@cmd) {
+            is join(' ', @cmd), 'nice inice qemu-img convert -p -O qcow2 ' . backend::svirt::IMAGE_STORAGE . 'barfoob.img assets_public/test.qcow2 -c', 'got expected command';
```
perlpunk, line 764: typo: `s/inice/ionice/`

perlpunk (approved):

okurz (approved):

PR #1896, opened 2021-12-29
utils: Extract only if assets are available

kalikiana:
Instead of checking the backend, we can see if we have assets that need
to be extracted via e.g. PUBLISH_HDD_1.

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -128,7 +128,6 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} eq 'qemu';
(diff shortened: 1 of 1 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -158,6 +157,8 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
         bmwqemu::diag "Requested to force the publication of '$name'";
         push @toextract, _store_asset($i, $name, 'assets_public');
     }
+    # Nothing to do here if we don't have any assets to extract
+    return unless @toextract;
```
perlpunk, line 161: I think we aren't using `return` without a value.
You should return `$return_code`, however, that is already done after the loop (which isn't executed if the array is empty anyway), so I don't see how that would change anything.

perlpunk:
This is the same as #1897 and 1897 was already mentioned elsewhere, so I'll close this one.

PR #1897, opened 2021-12-29
utils: Extract asset regardless of backend

kalikiana:
We don't need to check the backend if it's implemented correctly. And we can test that extraction is only called when requested.

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -128,7 +128,6 @@ configuration variables.
 sub handle_generated_assets ($command_handler, $clean_shutdown) {
     my $return_code = 0;
     # mark hard disks for upload if test finished
-    return unless $bmwqemu::vars{BACKEND} eq 'qemu';
     my @toextract;
     my $nd = $bmwqemu::vars{NUMDISKS};
     if ($command_handler->test_completed) {
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -184,6 +184,20 @@ subtest 'upload the asset even in an incomplete job' => sub {
     ok(!-e $pool_dir . '/assets_public/publish_test.qcow2', 'the asset defined by PUBLISH_HDD_X would not be generated in an incomplete job');
 };
 
+subtest 'asset handling in unsupported backends' => sub {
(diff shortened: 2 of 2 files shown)


kalikiana:
This is an alternative to #1895 which assumes the method should always be implemented. The function returns after looking for viable assets rather than bailing based on the backend.

perlpunk:
Is this the same as #1896?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -158,6 +157,8 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
         bmwqemu::diag "Requested to force the publication of '$name'";
         push @toextract, _store_asset($i, $name, 'assets_public');
     }
+    # Nothing to do here if we don't have any assets to extract
+    return unless @toextract;
```
perlpunk, line 161: I think we aren't using `return` without a value.
You should return `$return_code`, however, that is already done after the loop (which isn't executed if the array is empty anyway), so I don't see how that would change anything.

kalikiana: We are in line 131 ;-) But that wasn't the point.
My intention was to skip `$bmwqemu::backend->extract_assets($asset)` which calls `do_extract_assets` even though for example `PUBLISH_HDD_1` was defined. That's what https://openqa.suse.de/tests/7917723 / mru-install-minimal-with-addons@s390x-kvm-sle12 does.

kalikiana: Maybe I'm not tracing the code path correctly. Maybe line 131 is not what breaks the abve test. I don't have a unit test for this yet.

perlpunk: Well, currently the code would only enter the loop and call `extract_assets` if there are elements in `@toextract`.
Returning before the loop doesn't make a difference in this regard.
Simplified code:
```
return unless @array;
for my $element (@array) {
    do_something($element);
}
return;
```
The first return statement is useless here if `@array` is empty.

The fact that `extract_assets` was called in our case, shows that `@toextract` was not empty.

kalikiana:
Discussed briefly w/ Tina, this branch is redundant now and we don't really want tests for the lack of optional code paths.

PR #1898, opened 2021-12-30
 Another attempt/method for running tests on physical laptops (try2)

marmarek:
Retry #1741 , this time do not require `do_extract_assets` in all backends. Commit https://github.com/os-autoinst/os-autoinst/commit/32862bc5eb314ceddc250a50731e1ef04a564bb0 is replaced with https://github.com/os-autoinst/os-autoinst/commit/95c5a7dd1ef33889fabca7296edff75130a15187 ("Enable assets collection for generalhw backend too").

See discussion at https://progress.opensuse.org/issues/104499 and https://github.com/os-autoinst/os-autoinst/pull/1741#pullrequestreview-841490852


okurz (approved):
This looks ok although of course there can still be corner cases which could be problematic. 
There are smaller changes to adress the original problem in a more sustainable and also contained way with https://github.com/os-autoinst/os-autoinst/pull/1895 and https://github.com/os-autoinst/os-autoinst/pull/1897 . 
In case we won't be able to manage to have them merged the following weeks I would still be ok to merge this PR in the current form.

consoles/video_stream.pm
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    bmwqemu::diag "Starting to receive video stream at $args->{url}";
+    $self->connect_remote_video($args->{url});
+
+    $self->connect_remote_input($args->{input_cmd}) if $args->{input_cmd};
+}
+
+sub _get_ffmpeg_cmd ($self, $url) {
+    my @cmd = ('/usr/bin/ffmpeg', '-loglevel', 'fatal', '-i', $url);
```
Martchus, line 109: Is hard-coding the absolute path required here? It would be nice to make this configurable (via `PATH` or a specific env variable).

marmarek: Apparently using just `ffmpeg` simply works (via `PATH`), so I changed this.

okurz:
For lower risk when merging I would like to split the PR and pull out some commits that we can merge first before going on with later commits. I will try to split out some commits into one or multiple PR but feel free to do so yourself as well.

marmarek:
One option is to pull everything related to video-stream console into separate PR, without actually connecting it into generalhw backend - is that something that you'd be fine with merging (without followup PR, it would be a dead code)?

okurz:
> One option is to pull everything related to video-stream console into separate PR, without actually connecting it into generalhw backend - is that something that you'd be fine with merging (without followup PR, it would be a dead code)?

I guess that would work but I think we can also do kinda the opposite and split out changes to general_hw which don't necessarily rely on video-stream. I tried that with https://github.com/os-autoinst/os-autoinst/pull/1918 . If you can approve that PR and tests are good then we can have that merged first, rebase this PR here then and see what is left.

marmarek:
Hmm, in fact, [connecting it into generalhw](https://github.com/os-autoinst/os-autoinst/pull/1898/commits/29892fa8452a4035ea5c7a4d855149ba673a64e7) should be safe too, as it doesn't change anything unless someone uses `GENERAL_HW_VIDEO_STREAM_URL` setting.

marmarek:
BTW, since yesterday it is running on the production instance at https://openqa.qubes-os.org already ([example job](https://openqa.qubes-os.org/tests/28211)).

consoles/video_stream.pm
```diff
@@ -0,0 +1,350 @@
...
+        # typical max_interval values are
+        #   4ish:  veeery slow
+        #   15ish: slow
+        $seconds_per_keypress = $seconds_per_keypress + 1 / sqrt($args->{max_interval});
+    }
+
+    for my $letter (split("", $args->{text})) {
+        next if ($letter eq "\r");
+        $letter = $CHARMAP->{$letter} || $letter;
+        $self->_send_key_event($letter);
```
Martchus, line 253:
```suggestion
        $self->_send_key_event($CHARMAP->{$letter} || $letter);
```

consoles/video_stream.pm
```diff
@@ -0,0 +1,350 @@
...
+    my $mask = 0;
+    if ($button eq 'left') {
+        $mask = $bstate;
+    }
+    elsif ($button eq 'right') {
+        $mask = $bstate << 2;
+    }
+    elsif ($button eq 'middle') {
+        $mask = $bstate << 1;
+    }
```
Martchus, line 343: It would be nice *not* to C&P this code from the VNC console (especially since the code will be improved by another pending PR).

Martchus (commented):
I have mainly found some style issues.

consoles/video_stream.pm
marmarek: Do you suggest a different API for sending button presses, or just different way to write this part?

Martchus: I suggest to avoid code duplication because this code looks very familiar.

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1918 merged. We will see any impact as soon as tomorrow morning. Please rebase on top and update this PR.

consoles/video_stream.pm
marmarek: Technically, it is a different protocol, even if indeed both have mouse buttons as a bitmask. For example VNC uses a single command for both mouse move and click, while this one have separate commands. But I can make the improvement similar to VNC's here too.

consoles/video_stream.pm
```diff
@@ -0,0 +1,340 @@
...
+    $self->backend->run_capture_loop(.1);
+
+    return;
+}
+
+sub mouse_hide ($self, $args) {
+    $args->{border_offset} //= 0;
+
+    #my $x = $self->{width} - 1;
+    #my $y = $self->{height} - 1;
```
okurz, line 307: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

consoles/video_stream.pm
```diff
@@ -0,0 +1,340 @@
+# Copyright 2021 Marek Marczykowski-Górecki
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package consoles::video_stream;
+
+use Mojo::Base -strict, -signatures;
+
+use base 'consoles::network_console';
```
okurz, line 9: I suggest to introduce a new base class `consoles::video_base` which you can base `consoles::video_stream` and `consoles::vnc_base` on. Then you can prevent any additional duplication like the methods "get_last_mouse_set", "screen", "type_string", the button call Martchus mentioned, etc.

okurz (changes requested):

consoles/video_stream.pm
okurz: Sure, this is not about the difference in protocol but just simply the technical implementation regarding OOP. Please see https://github.com/os-autoinst/os-autoinst/pull/1898#discussion_r786487750

okurz:
I did some scratch work to show what I mean. Maybe https://github.com/os-autoinst/os-autoinst/pull/1922/commits/21d59f417fcaa66f486110c1faa6e892f3e54bd8 gives you the idea?

ggardet:
@marmarek Could you rebase, please?

marmarek:
Rebased and extracted video_base to reduce some code duplication.

okurz:
Thank you. The change is still big and with the changes in common base classes that are not covered by any automatic tests we should be particularly careful.

@os-autoinst/tools-team one of you would like to give it a try to have at least unit tests for the newly introduced code locations like the video_base class?

@os-autoinst/tests-maintainer @os-autoinst/tools-team can some of you run additional tests with this change applied, e.g. on bare-metal IPMI tests or s390x svirt tests?

Martchus (commented):
It is nice to generalize video handling to reuse so far VNC-specific code elsewhere. I hope it doesn't break anything but at least our tests seem to still pass (without adjustments). As mentioned by @okurz it would be nice to cover all of the new code.

Martchus:
For further testing: We could hot-patch the change on one of the production workers (which also run special backends) and monitor the jobs closely.

Martchus:
@ggardet Thanks for adapting the test distribution - although it let's me wondering: Shouldn't we avoid making backwards-incompatible changes of that kind in the first place (also considering the openSUSE's test distribution isn't the only one)?

ggardet:
> @ggardet Thanks for adapting the test distribution - although it let's me wondering: Shouldn't we avoid making backwards-incompatible changes of that kind in the first place (also considering the openSUSE's test distribution isn't the only one)?

My update is to be able to use the new features, not to avoid a breakage.
Also, I think openSUSE and QubesOS are the only users testing with generalhw. And @marmarek is the maintainer for QubesOS. 
So, it should be fine.

marmarek:
I've just pushed one fix. With this applied, I can confirm it works for Qubes OS too.

marmarek:
As discussed at https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14208#pullrequestreview-877689584, I've reverted de-duplication of `mouse_button`, and adjusted the one in video_stream to use buttons mask as USB HID will need. The gadget-control script is updated too.

okurz:
still what I see what needs to be done

> @os-autoinst/tools-team one of you would like to give it a try to have at least unit tests for the newly introduced code locations like the video_base class?
> 
> @os-autoinst/tests-maintainer @os-autoinst/tools-team can some of you run additional tests with this change applied, e.g. on bare-metal IPMI tests or s390x svirt tests?

marmarek:
> Current head [ae80034](https://github.com/os-autoinst/os-autoinst/commit/ae8003481c230cefb4ba338e80ce4f7bdc4ca230) differs from pull request most recent head [6463b4e](https://github.com/os-autoinst/os-autoinst/commit/6463b4e54369ae7a23e248d4f6890393351d616e). Consider uploading reports for the commit [6463b4e](https://github.com/os-autoinst/os-autoinst/commit/6463b4e54369ae7a23e248d4f6890393351d616e) to get more accurate results

Codecov is confused here. But I _think_ the only parts of video_base not covered with tests are the stubs that are overloaded in subclasses anyway.

marmarek:
Any update on the above (outdated codecov report, tests on other generalhw instances)? For codecov, I can try rebasing, maybe it will help...

okurz (approved):
I trust that we are in a much better state. There are good unit tests for the newly introduced module as well as the changed backend. Also ggardet tested the changes on real hardware so I can approve.

mimi1vx (approved):

okurz:
@mergifyio rebase

okurz:
The CI status currently does not show this but we are waiting for https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-1898

PR #1901, opened 2022-01-03
Ensure to import os-autoinst-common in all tests consistently

okurz:
Instead of adding "os-autoinst-common" to the include path we can simply
include the corresponding paths within all tests which we already did in
most.


perlpunk (approved):

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1902 to address the change in coverage

Martchus (approved):

PR #1902, opened 2022-01-04
t: Ensure 100% statement coverage of consoles/virtio_terminal.pm

okurz:


consoles/virtio_terminal.pm
```diff
@@ -140,7 +139,7 @@ sub open_pipe ($self) {
     sysopen(my $fd_r, $self->{pipe_prefix} . '.out', O_NONBLOCK | O_RDONLY)
       or die "Can't open out pipe for reading $!";
 
-    my $newsize = get_var('VIRTIO_CONSOLE_PIPE_SZ', path('/proc/sys/fs/pipe-max-size')->slurp());
+    my $newsize = $bmwqemu::vars{VIRTIO_CONSOLE_PIPE_SZ} // path('/proc/sys/fs/pipe-max-size')->slurp();
```
perlpunk, line 142: Why did you replace the function call with direct access to the hash?

okurz: That's explained in https://github.com/os-autoinst/os-autoinst/pull/1902/commits/7ffc466fec6d303a36ccf9008a89d06986274bc3 . We should avoid using an external facing API in internal calls,. And they provide no real benefit here and can even be harmful for the dependency handling. Think circular dependencies.

kalikiana (approved):

mimi1vx (approved):

consoles/virtio_terminal.pm
perlpunk: I wouldn't call this an "internal call" though.
Ideally the vars hash will be a `my` variable at some point in the future, and outside of the module where it is defined it will only be accessible via a function call.
But I guess as we have a tied hash now and we have tasks with higher priority, it's fine.

PR #1900, opened 2022-01-03
Control access to %bmwqemu::vars and warn about invalid keys

perlpunk:
Issue: https://progress.opensuse.org/issues/98577

%bmwqemu::vars is writeable from everywhere, and we can't change that
for backwards compatibility.

But by making it a tied hash, we can control access to it.

PR #1903, opened 2022-01-04
Fix 02-test_ocr.t after bmwqemu::vars change

perlpunk:
Brokem since #1900 39b18520

The test writes to bmwqemu::vars before the `tie` happens.
Ideally any code should load the module before writing to %vars,
but we can solve it also by passing the current value of %vars to tie.

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -55,7 +55,7 @@ our @ovmf_locations = (
 );
 
 our %vars;
-tie %vars, 'bmwqemu::tiedvars';
+tie %vars, 'bmwqemu::tiedvars', %vars;
 
 sub result_dir { 'testresults' }
 


t/02-test_ocr.t
```diff
@@ -11,6 +11,7 @@ use Test::Warnings ':report_warnings';
 use File::Which 'which';
 use File::Basename;
 
+#use bmwqemu ();
```
okurz, line 14: ?

okurz (commented):
why the disabled line

t/02-test_ocr.t
perlpunk: forgotten, already removed :)

okurz (approved):

PR #1905, opened 2022-01-05
Fix circular dependency

perlpunk:
`bmwqemu -> backend::driver -> myjsonrpc -> bmwqemu`

However, the use of backend::driver was already thrown out of
bmwqemu in 2016 in 59ba44fa8e97db0be34c33b75be4265a59ea9feb, just the
use statement survived.

Now bmwqemu doesn't use any internal os-autoinst related module anymore.

The effects of the circular dependency can be watched when doing `perl -I. -c bwmqemu.pm`:
```
Subroutine result_dir redefined at bmwqemu.pm line 60.
Subroutine logger redefined at bmwqemu.pm line 62.
Subroutine init_logger redefined at bmwqemu.pm line 64.
Subroutine serialize_state redefined at bmwqemu.pm line 69.
...
```

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -25,7 +25,6 @@ use Exporter 'import';
 our $VERSION;
 our @EXPORT_OK = qw(diag fctres fctinfo fctwarn modstate save_vars);
 
-use backend::driver;
 require IPC::System::Simple;
 
 sub mydie;
@@ -227,6 +226,7 @@ sub modstate {
 
 sub current_test {
     require autotest;
+    no warnings 'once';
     return $autotest::current_test;
 }
 
--- a/isotovideo
+++ b/isotovideo
@@ -53,6 +53,7 @@ BEGIN {
 }
 
 use bmwqemu;
+use backend::driver;
 use needle;
 use autotest ();
 use commands;
--- a/t/10-terminal.t
+++ b/t/10-terminal.t
@@ -22,6 +22,7 @@ my $main_pid = $$;
 
 use consoles::virtio_terminal;
 use testapi ();
+use myjsonrpc;
 use bmwqemu ();
 
 our $VERSION;


bmwqemu.pm
```diff
@@ -227,6 +226,7 @@ sub modstate {
 
 sub current_test {
     require autotest;
+    no warnings 'once';
```
perlpunk, line 229: I can't remember right now if there is a better fix. a warning doesn't appear when using `use` instead of require.

kraih: Wonder if the `require` above should be moved to the top of the file, or if it really should be loaded very late at runtime.

perlpunk: Actually `autotest` does `use bmwqemu`, so this is another example of a circular dependency, and doing the `require autotest` at runtime at least avoids compile time warnings.

I don't see a quick fix for this, as we are exposing the `$autotest::current_test` variable as a global var *sigh*.
(Looking into os-autoinst-distri-opensuse, I can see it used only in 3 files, but still, we don't know what other code is out there using it. I guess there should have been a testapi function providing it.)
The correct fix for this circular dependency would be to move the current_test name into its own module, that `autotest` and `bmwqemu` can use.
To not break code, this could be solved with making `$autotest::current_test` a tied scalar, setting the real value in the `STORE` function call.

Unless we want to invest time with this right now, I would say we keep the `no warnings` here for now.

kraih (approved):

kalikiana:
@perlpunk Is this [poo#104670](https://progress.opensuse.org/issues/104670), which then should be mentioned in the description and be **in progress**?

Or if somehow it's not, I assume it's still related and should be mentioned

bmwqemu.pm
perlpunk: I created https://progress.opensuse.org/issues/104670 for this one

perlpunk:
I created https://progress.opensuse.org/issues/104670 for the other circular dep, see https://github.com/os-autoinst/os-autoinst/pull/1905#discussion_r779485615

This one doesn't have a ticket

kalikiana (approved):

perlpunk:
OBS tests failed:
```
[  121s] 3: 
[  121s] 3: #   Failed test 'log output for backend driver creation'
[  121s] 3: #   at ./29-backend-driver.t line 24.
[  121s] 3: # STDOUT & STDERR:
[  121s] 3: # hannel_in 22
[  121s] 3: # [0m
[  121s] 3: # don't match:
[  121s] 3: # (?^u:(Blocking SIGCHLD|channel_out))
[  121s] 3: # as expected
[  121s] 3: # Looks like you failed 1 test of 12.
[  121s] 3: [20:34:59] ./29-backend-driver.t .................... 
[  121s] 3: Dubious, test returned 1 (wstat 256, 0x100)
[  121s] 3: Failed 1/12 subtests 
```
I'll retrigger

PR #1909, opened 2022-01-11
Also hide common password test variables from vars.json

okurz:
We already don't write any variable with "_SECRET_" in the name to vars.json
for security reasons. Within os-autoinst we have some security relevant data,
e.g. passwords within backends that we should likely treat the same. Cloning
openQA tests would mean that the passwords are not taken over however most
likely they are filled by worker settings when jobs are started anywhere.

Related progress issue: https://progress.opensuse.org/issues/104751

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -99,7 +99,7 @@ sub save_vars {
     my $write_vars = \%vars;
     if ($args{no_secret}) {
         $write_vars = {};
-        $write_vars->{$_} = $vars{$_} for (grep !/^_SECRET_/, keys(%vars));
+        $write_vars->{$_} = $vars{$_} for (grep !/(^_SECRET_|_PASSWORD)/, keys(%vars));
     }
 
     # make sure the JSON is sorted
--- a/t/12-bmwqemu.t
+++ b/t/12-bmwqemu.t
@@ -106,7 +106,7 @@ subtest 'save_vars' => sub {
 
 subtest 'save_vars no_secret' => sub {
     my $dir = "$data_dir/tests";
-    create_vars({CASEDIR => $dir, _SECRET_TEST => 'my_credentials'});
+    create_vars({CASEDIR => $dir, _SECRET_TEST => 'my_credentials', MY_PASSWORD => 'secret'});
     $bmwqemu::openqa_default_share = $data_dir;
 
     eval {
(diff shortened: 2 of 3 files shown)


bmwqemu.pm
```diff
@@ -99,7 +99,7 @@ sub save_vars {
     my $write_vars = \%vars;
     if ($args{no_secret}) {
         $write_vars = {};
-        $write_vars->{$_} = $vars{$_} for (grep !/^_SECRET_/, keys(%vars));
+        $write_vars->{$_} = $vars{$_} for (grep !/(^_SECRET_|_PASSWORD)/, keys(%vars));
```
kalikiana, line 102: I'm thinking even better would be to make this the configurable default. Or make the backends add passwords to a different vars hash, if that's your goal, because that's the only way to be certain this works for all passwords used.

kalikiana (changes requested):
I'm

bmwqemu.pm
kalikiana: > Consider what happens when cloning such jobs. Do they fail because the password is missing?
Also this. If backends use a different hash, a clone can fail with a clear error message like `variable FOO_PASSWORD not copied when cloning`

okurz: I don't need this to be used for "all passwords" but just what we use within our backends.



>> Consider what happens when cloning such jobs. Do they fail because the password is missing?

> Also this. If backends use a different hash, a clone can fail with a clear error message like variable FOO_PASSWORD not copied when cloning

Not sure what you mean here. I don't want any clone to fail because we can't know what variables are defined within worker settings, which is commonly the case for such password variables to access remote machines.

okurz: @kalikiana for security purposes I suggest to not make this a configurable variable. Otherwise it would be really hard to control who can change this variable which can also be abused for malicious intent.

kalikiana: > Not sure what you mean here. I don't want any clone to fail because we can't know what variables are defined within worker settings, which is commonly the case for such password variables

We know what variables the backend requires e.g. `$bmwqemu::vars{IPMI_PASSWORD}` or `$bmwqemu::vars{NOVALINK_PASSWORD}`. The same ones that are listes in `doc/backend_vars.asciidoc`... actually, we seem to be missing HMC and ZVM... so what I'm saying is, if we know the variables we can make sure they are all covered or lead to a well-defined error.

Martchus (approved):

kraih (approved):

bmwqemu.pm
kalikiana: Discussed in the unblock call. I'll propose a PR for undocumented vars and we don't plan to add more variable names here

kalikiana (approved):

PR #1910, opened 2022-01-12
doc: Document ZVM backend variables

kalikiana:


okurz (changes requested):
We should only include variables that our check flags. We need https://github.com/os-autoinst/os-autoinst/pull/1912 for which we likely need this change.

kalikiana:
> We should only include variables that our check flags. We need #1912 for which we likely need this change.

If this branch needs to be merged first, what other changes did you have in mind?

okurz:
Following TDD I want to make sure that we actually need the changes to fix the check after correcting the variable use in backends and consoles

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1912 includes your change

doc/backend_vars.asciidoc
```diff
@@ -43,6 +43,16 @@ UPLOAD_INACTIVITY_TIMEOUT;integer;300;Specifies the inactivity timeout in second
 
 |====================
 
+.ZVM backend
+[grid="rows",format="csv"]
+[options="header",cols="^m,^m,^m,v",separator=";"]
+|====================
+Variable;Values allowed;Default value;Explanation
+ZVM_HOST;string;;Sets the remote host name.
```
Martchus, line 51:
```suggestion
ZVM_HOST;string;;Sets the remote hostname.
```

doc/backend_vars.asciidoc
```diff
@@ -43,6 +43,16 @@ UPLOAD_INACTIVITY_TIMEOUT;integer;300;Specifies the inactivity timeout in second
...
+ZVM_GUEST;string;;Sets the remote user name.
```
Martchus, line 52:
```suggestion
ZVM_GUEST;string;;Sets the remote username.
```
(Just for consistency with your other PR.)

Martchus (changes requested):

doc/backend_vars.asciidoc
okurz: fixed in https://github.com/os-autoinst/os-autoinst/pull/1912

okurz:
covered by https://github.com/os-autoinst/os-autoinst/pull/1912 where your commit is included

PR #1911, opened 2022-01-12
doc: Document HMC backend variables

kalikiana:


okurz (approved):
The actual problem seems to be that we (mis-)use the testapi `get_var` or `get_required_var` functions in the backends when we should use the bmwqemu variable directly. I already prepared some other changes where I fixed that. I will continue with that.

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1912 includes your change

Martchus (approved):

PR #1913, opened 2022-01-12
Use signatures in consoles::vnc_base and consoles::serial_screen

okurz:
* Simplify consoles::vnc_base
* Simplify consoles::serial_screen
* Use proper perl "signatures" in consoles::serial_screen
* Use proper perl "signatures" in consoles::vnc_base

Pulled out from https://github.com/os-autoinst/os-autoinst/pull/1789 to have a separate, easy to merge PR for the files which have complete statement coverage and keep consoles::VNC separately as merging it might be more risky.

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -5,32 +5,25 @@
 package consoles::vnc_base;
 
 use Mojo::Base -strict, -signatures;
-use feature 'say';
 
 use base 'consoles::network_console';
 
 use consoles::VNC;
-use List::Util 'max';
 use Time::HiRes qw(usleep);
 
-use Try::Tiny;
 use bmwqemu ();
 
 # speed limit: 30 keys per second
 use constant VNC_TYPING_LIMIT_DEFAULT => 30;
 
-sub screen ($self) {
-    return $self;
-}
+sub screen ($self) { $self }
 
 sub disable ($self) {
     close($self->{vnc}->socket) if ($self->{vnc} && $self->{vnc}->socket);
     $self->{vnc} = undef;
 }
 
-sub get_last_mouse_set ($self) {
-    return $self->{mouse};
-}
+sub get_last_mouse_set ($self) { $self->{mouse} }
 
 sub disable_vnc_stalls ($self) {
     return unless $self->{vnc};
@@ -178,8 +171,6 @@ sub mouse_hide ($self, $args) {
 
 sub mouse_set ($self, $args) {
     die "Need x/y arguments" unless (defined $args->{x} && defined $args->{y});
-
-    # TODO: for framebuffers larger than 1024x768, we need to upscale
     $self->_mouse_move(int($args->{x}), int($args->{y}));
     return {};
 }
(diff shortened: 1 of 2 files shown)


consoles/vnc_base.pm
```diff
@@ -191,31 +168,17 @@ sub mouse_hide {
...
-    if ($button eq 'left') {
-        $mask = $bstate;
-    }
-    elsif ($button eq 'right') {
-        $mask = $bstate << 2;
-    }
-    elsif ($button eq 'middle') {
-        $mask = $bstate << 1;
-    }
+    my $mask = $button eq 'left' ? $bstate : $button eq 'right' ? $bstate << 2 : $bstate << 1;
```
Martchus, line 181: Pressing the middle button when the specified button is actually invalid seems a bit random. Previously the `$mask` was zero in that case and I'd rather preserve that behavior and possibly print a warning or even raise an error in that case. Besides, in the previous code `1` wasn't a magic number because it was clear that it mean "middle". In the changed version that's no longer clear.

Martchus (changes requested):

consoles/vnc_base.pm
okurz: yes, you are right. Updated with a nicer lookup-table approach.

Martchus (approved):

consoles/vnc_base.pm
Martchus: Looks good :-)

consoles/vnc_base.pm
```diff
@@ -192,31 +169,17 @@ sub mouse_hide {
...
-        $mask = $bstate;
-    }
-    elsif ($button eq 'right') {
-        $mask = $bstate << 2;
-    }
-    elsif ($button eq 'middle') {
-        $mask = $bstate << 1;
-    }
+    my %mask = (left => $bstate, right => $bstate << 2, middle => $bstate << 1);
+    my $mask = $mask{button} // 0;
```
Martchus, line 182: As pointed out by @perlpunk in the other PR, this misses a `$`.

Martchus (changes requested):

consoles/vnc_base.pm
perlpunk: @okurz  it's really confusing to have the same commits in several open pull requests.
Good that I didn't start to review this one yet.

okurz: That's way I cross-reference PRs. The number of commits can also be an indication. Start with the ones with least commits

okurz: > > Start with the ones with least commits
> 
> No, I refuse to. I requested changes on #1914 and I will not go over the commits in several other PRs to request the same changes.

Sure, fine. I am happy about review feedback either way.

> Please mark your PRs in a way that people don't have to figure that out themselves.

Sorry, I try to. The description in https://github.com/os-autoinst/os-autoinst/pull/1914 says "Based on #1913 and #1789". I thought that makes it clear. Maybe together we can find a better general approach.

Anyway, PR updated to include your proposals in this PR as well as the other.

okurz:
Updated according to comments

consoles/vnc_base.pm
Martchus: It is still missing the `$` sign.

Martchus (changes requested):

consoles/vnc_base.pm
okurz: oh, geez. That's embarrassing. Sorry about that. Fixed

Martchus (approved):

perlpunk (approved):

PR #1914, opened 2022-01-12
Simplify base inheritance statement with Mojo::Base everywhere

okurz:
Based on
* #1913
* #1926
* #1789

The patch coverage will be still be our automatic acceptance threshold due to background processes not collecting coverage properly. If you can approve in the current state please approve which I will see as implicit approval of the other three PRs, otherwise please review the PRs above in the stated order.

EDIT: I managed to exclude the consoles::VNC changes which cause the codecov patch coverage check to fail so this PR now only includes the one commit about "base inheritance using Mojo::Base"

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -1,11 +1,7 @@
 package consoles::VNC;
 
-use Mojo::Base -strict;
+use Mojo::Base 'Class::Accessor::Fast', -signatures;
 use bytes;
-use feature 'say';
-
-use base 'Class::Accessor::Fast';
-
 use IO::Socket::INET;
 use bmwqemu qw(diag fctwarn);
 use Time::HiRes qw( sleep gettimeofday time );
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -4,10 +4,7 @@
 
 package consoles::vnc_base;
 
-use Mojo::Base -strict, -signatures;
-
-use base 'consoles::network_console';
-
+use Mojo::Base 'consoles::network_console', -signatures;
 use consoles::VNC;
 use Time::HiRes qw(usleep);
 
--- a/backend/amt.pm
+++ b/backend/amt.pm
@@ -3,11 +3,8 @@
 
 package backend::amt;
 
-use Mojo::Base -strict, -signatures;
+use Mojo::Base 'backend::baseclass', -signatures;
 use autodie ':all';
-
-use base 'backend::baseclass';
-
 use Time::HiRes qw(sleep gettimeofday);
 use Data::Dumper;
 require Carp;
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -6,11 +6,8 @@
 
 package backend::generalhw;
 
-use Mojo::Base -strict, -signatures;
+use Mojo::Base 'backend::baseclass', -signatures;
 use autodie ':all';
-
-use base 'backend::baseclass';
-
 use bmwqemu;
 use IPC::Run ();
 require IPC::System::Simple;
--- a/backend/ikvm.pm
+++ b/backend/ikvm.pm
@@ -4,10 +4,8 @@
 
 package backend::ikvm;
 
-use Mojo::Base -strict, -signatures;
+use Mojo::Base 'backend::ipmi', -signatures;
 use autodie ':all';
-
-use base 'backend::ipmi';
 
 sub new ($class) { $class->SUPER::new }
 
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -3,12 +3,8 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package backend::ipmi;
-
-use Mojo::Base -strict, -signatures;
+use Mojo::Base 'backend::baseclass', -signatures;
 use autodie ':all';
-
-use base 'backend::baseclass';
-
 use Time::HiRes qw(sleep);
 use Time::Seconds;
 use IPC::Run ();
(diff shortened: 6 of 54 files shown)


Martchus:
It would be good if @kraih could have a look (although at the moment this is only build on top of several other changes which should likely be merged before).

kraih (approved):
It's a large patch, so easy to overlook something. But it seems fine to me.

consoles/VNC.pm
```diff
@@ -1093,13 +1030,11 @@ sub _receive_colour_map ($self) {
     $self->socket->read(my $map_infos, 5);
     my ($padding, $first_colour, $number_of_colours) = unpack('Cnn', $map_infos);
 
-    for (my $i = 0; $i < $number_of_colours; $i++) {
+    for (0 .. $number_of_colours) {
         $self->socket->read(my $colour, 6);
         my ($red, $green, $blue) = unpack('nnn', $colour);
-        tinycv::set_colour($self->vncinfo, $first_colour + $i, $red / 256, $green / 256, $blue / 256);
+        tinycv::set_colour($self->vncinfo, $first_colour + $_, $red / 256, $green / 256, $blue / 256);
```
perlpunk, line 1033: this includes `$number_of_colours`. before it was `$i < $number_of_colours`.

perlpunk (changes requested):

consoles/vnc_base.pm
```diff
@@ -178,26 +171,15 @@ sub mouse_hide ($self, $args) {
...
-        $mask = $bstate;
-    }
-    elsif ($button eq 'right') {
-        $mask = $bstate << 2;
-    }
-    elsif ($button eq 'middle') {
-        $mask = $bstate << 1;
-    }
+    my %mask = (left => $bstate, right => $bstate << 2, middle => $bstate << 1);
+    my $mask = $mask{button} // 0;
```
perlpunk, line 179: `$mask{button}` misses a `$`.
```suggestion
    my $mask = {left => $bstate, right => $bstate << 2, middle => $bstate << 1}->{$button} // 0;
```

consoles/serial_screen.pm
```diff
@@ -155,9 +147,7 @@ An undefined timeout will cause to wait indefinitely. A timeout of 0 means to
 just read once.
 
 =cut
-sub do_read
-{
-    my ($self, undef, %args) = @_;
+sub do_read ($self, $buf, %args) {
```
perlpunk, line 125:
```suggestion
sub do_read ($self, $, %args) {
```
like in `consoles/ssh_screen.pm`

consoles/vnc_base.pm
```diff
@@ -52,17 +46,15 @@ sub connect_remote {
     return $self->{vnc};
 }
 
-sub request_screen_update {
-    my ($self) = @_;
+sub request_screen_update ($self, @) {
```
perlpunk, line 39: is there a case where the method is called with an argument?

perlpunk:
> Based on #1913 and #1789

Is there a reason that you created this PR based on two other PRs?
Are you expecting people to review this PR although it contains the same commits as other PRs?
Which one should people review?
Mark other PRs as draft or try to help in any other way that people don't have double work, please.

consoles/vnc_base.pm
okurz: I am not exactly sure but I think so when used as a callback target over the RPC implementation which somehow ends up with more unused, unfilled arguments passed. I might have overdone it in one or two cases but normally I would only add the argument if some tests trigger an error or warning

okurz:
> > Based on #1913 and #1789
> 
> Is there a reason that you created this PR based on two other PRs?

Yes, because otherwise changes would trigger conflicts. I could invest the additional effort to split out commits ito separate PRs for easier review. As soon as one PR is merged the other will show up as conflicting and can not be merged so I would need to rebase and fix the conflict and upload again which you or others might see as need to review again so it's all additional effort that I would like to avoid if I can.

> Are you expecting people to review this PR although it contains the same commits as other PRs?

Well, I hoped that you would have seen the description so that it's clear which commits are unique to this PR. Sorry, I would like if we could make this more clear in github somehow.

> Which one should people review? Mark other PRs as draft or try to help in any other way that people don't have double work, please.

Yes, I will research if there are better solutions. I did mark all PRs as ready for review because in the best case there would have been no problems and in that case there would not be any more work necessary.

consoles/serial_screen.pm

consoles/vnc_base.pm
okurz: done, interesting syntax but makes sense and looks nice :)

consoles/VNC.pm
okurz: right, fixed. thx

consoles/VNC.pm
```diff
@@ -1139,38 +1023,30 @@ sub _receive_ikvm_encoding {
...
 
-sub _receive_colour_map {
-    my $self = shift;
-
+sub _receive_colour_map ($self) {
     $self->socket->read(my $map_infos, 5);
     my ($padding, $first_colour, $number_of_colours) = unpack('Cnn', $map_infos);
 
-    for (my $i = 0; $i < $number_of_colours; $i++) {
+    for (0 .. $number_of_colours - 1) {
```
kraih, line 1030: `for my $i (0 ...` is cleaner than relying on `$_`.

okurz: done

kraih (approved):

Martchus:
I must say I haven't dared to approve the VNC changes yet (since the test coverage is unclear).

kraih:
Better test coverage would be nice indeed. But the changes themselves are not very risky and easy enough to revert if necessary.

okurz:
I managed to exclude the consoles::VNC changes which cause the codecov patch coverage check to fail so this PR now only includes the one commit about "base inheritance using Mojo::Base". Let's see what the codecov check says now.

baierjan (approved):

PR #1915, opened 2022-01-12
Add test with 100% statement coverage for consoles::localXvnc

okurz:


kalikiana (approved):

okurz:
> I hope disable isn't called somewhere in the test distribution

I have found neither localXvnc nor any call to a "disable" method in os-autoinst-distri-opensuse so I think we are good

> Tidy is complaining about the formatting of the comments.

Fixed

Updated to also mock the call to "xdotool".

PR #1917, opened 2022-01-13
testapi: Allow distri class provide default for die_on_timeout

marmarek:
script_run warns if die_on_timeout is not provided (regardless if is 0
or 1). Adjusting all the usages to always use the argument is
unnecessary boilerplate, so let the distri class provide the default.
This way, test developers can set it everywhere without actually
changing every single call. It is still possible to override it
per-call.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -15,6 +15,7 @@ sub new {
     $self->{consoles} = {};
     $self->{serial_failures} = [];
     $self->{autoinst_failures} = [];
+    $self->{script_run_die_on_timeout} = -1;
 
 =head2 serial_term_prompt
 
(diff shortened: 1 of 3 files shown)


Martchus (approved):

marmarek:
> **codecov/patch ** — 59.49% of diff hit (target 69.82%)

Should I worry about it? I believe that's about the changed warning text, I don't see any test for that case.

kalikiana (approved):

marmarek:
I think codecov got confused what is part of this PR - there are a bunch of comments about "Added line ... was not covered by tests", about in files not changed in this PR at all.

okurz (approved):

PR #1919, opened 2022-01-17
Use more log functions from bmwqemu

okurz:
Instead of custom formats in the generic diag function let's make use of
specific functions fit for the purpose.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -52,13 +52,13 @@ BEGIN {
     unshift @INC, "$installprefix";
 }
 
-use bmwqemu;
 use backend::driver;
+use bmwqemu qw(diag fctwarn);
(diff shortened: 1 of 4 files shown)


Martchus (approved):

isotovideo
```diff
@@ -58,7 +58,7 @@ use needle;
 use autotest ();
 use commands;
 use distribution;
-use testapi 'diag';
+use bmwqemu qw(diag fctwarn);
```
perlpunk, line 61: that should be moved up to line 55 where the module is already used

okurz: yep, agreed. Done

perlpunk (approved):

PR #1925, opened 2022-01-19
Extract all log functions into new module "log"

okurz:
The module "bmwqemu" was always a not so nicely designed catch-all
ticket. We can split out the low-level logging functions which can be
also a first step towards better structuring which can also make it
easier for us to manage dependencies within the project.

All log functions in bmwqemu are currently kept in place, forwarding to
the functions in the log module accordingly.


bmwqemu.pm
```diff
@@ -57,9 +53,32 @@ tie %vars, 'bmwqemu::tiedvars', %vars;
...
-sub logger { $logger //= Mojo::Log->new(level => 'debug', format => \&log_format_callback) }
+# deprecated functions, moved to log module
+sub log_format_callback { log::log_format_callback(@_) }
+sub diag { log::diag(@_) }
+
+sub fctres {
+    my ($text, $fname) = @_;
+    $fname //= (caller(1))[3];
+    log::fctres($text, $fname);
+}
```
Martchus, line 64: Something like that might work as well:
```suggestion
*fctres = \&log::fctres;
}
```
Maybe e.g. @kraih or @perlpunk can help with the exact syntax in case my untested suggestion doesn't work.

---

Of course the same applies to all the other "forwarding" functions.

log.pm
```diff
@@ -0,0 +1,76 @@
...
+
+sub log_format_callback {
+    my ($time, $level, @items) = @_;
+
+    my $lines = join("\n", @items, '');
+
+    # ensure indentation for multi-line output
+    $lines =~ s/(?<!\A)^/  /gm;
+
+    return '[' . Time::Moment->now . "] [$level] " . $lines;
```
Martchus, line 32:
```suggestion
    return '[' . Time::Moment->now . "] [$level] $lines";
```

bmwqemu.pm
```diff
@@ -40,10 +40,6 @@ our $screenshotpath = "qemuscreenshot";
 
 # global vars
 
-our $logger;
```
Martchus, line 43: I suppose you can remove the comment above as well.

Martchus (changes requested):

bmwqemu.pm
kraih: I don't actually know if aliasing works there. 🤔

kraih: The usual way to deal with this is a global variable containing the caller level, and that can be changed with a `local`. https://github.com/mojolicious/mojo/blob/4093223cae00eb516e38f2226749d2963597cca3/lib/Test/Mojo.pm#L304

perlpunk: Using an alias like `*fctres = \&log::fctres;` should do the right thing, without having anything to do about the callstack.
I just tested a small proof of concept, and with the alias both calls (`bmwqemu::fctres` and `log::fctres`) result in the same `$fname`.
This should actually be testable, right?

bmwqemu.pm
okurz: No, we still have global vars, e.g. %vars.

log.pm
okurz: done

bmwqemu.pm

Martchus (approved):

perlpunk (changes requested):
I guess we should also make sure with a test that the deprecated functions still work. I haven't checked if that is the case.
Otherwise all good.

okurz:
> I guess we should also make sure with a test that the deprecated functions still work. I haven't checked if that is the case. Otherwise all good.

Done. Extended the test to still test bmwqemu logging wrapper functions as well as new ones in log.pm

PR #1929, opened 2022-01-21
Add OBS workflow

perlpunk:
Issue: https://progress.opensuse.org/issues/102464

Example PR: https://github.com/perlpunk/os-autoinst/pull/12

OBS Limitation: "Due to a limitation, the initial "pending" build status of
  packages with multibuild flavors is not reported. The build status for those
  flavors will however still be reported when the build finishes."
https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.setup.status_reporting

Because of this limitation, I had to add all the jobs we expect to mergify.

Only when this is merged, new PRs will show the new tests.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,6 +11,13 @@ pull_request_rules:
         # wait for OBS package build as well which is triggered independantly
         # See https://progress.opensuse.org/issues/55346 for details
         - "check-success=OBS Package Build"
+        # Multibuild results will not report the "pending" status, so we
+        # need to require them explicitly, see
+        # https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.setup.status_reporting
+        - "check-success~=OBS: os-autoinst - openSUSE_Leap_.*/x86_64"
+        - "check-success=OBS: os-autoinst - openSUSE_Tumbleweed/x86_64"
+        - "check-success~=OBS: os-autoinst:os-autoinst-test - openSUSE_Leap_.*/x86_64"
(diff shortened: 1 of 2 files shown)


.mergify.yml
```diff
@@ -11,6 +11,15 @@ pull_request_rules:
         # wait for OBS package build as well which is triggered independantly
         # See https://progress.opensuse.org/issues/55346 for details
         - "check-success=OBS Package Build"
+        # Multibuild results will not report the "pending" status, so we
+        # need to require them explicitly, see
+        # https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.setup.status_reporting
```
okurz, line 16: Is there a feature request for this? Because hard coding here shows how this is problematic also for other use cases

perlpunk: I don't know of a feature request.

okurz: Can you create one? Like the bug report you created for the open build service, just a different GitHub issue template :)

By the way, as alternative to spelling out each check for mergify we can also look for a count of checks that must be successful. This way we wouldn't need to hardcore everything. If you don't like that please use at least a generic pattern for the Leap version so that we don't need to change all these numbers with every point-release of Leap

perlpunk: I changed the check for Leap to a regex now

perlpunk: Created an issue: https://github.com/openSUSE/open-build-service/issues/12121

okurz (approved):

Martchus (approved):

PR #1930, opened 2022-01-22
Mark backend::amt as deprecated as it seems to be unused

okurz:
As the commit "backend::amt: Remove unsupported call to remove crash
file" showed backend::amt had an error since 4 years so it is likely
completely unused. Also on openqa.suse.de and openqa.opensuse.org I
called SQL `select * from machines where backend ~ 'amt';` and found no
matches.

Related progress issue: https://progress.opensuse.org/issues/105217

### DIFF
--- a/backend/amt.pm
+++ b/backend/amt.pm
@@ -31,6 +31,7 @@ sub new ($class) {
     $ENV{'WSMAN_USER'} = 'admin';
     $ENV{'WSMAN_PASS'} = $bmwqemu::vars{AMT_PASSWORD};
 
+    bmwqemu::fctwarn 'DEPRECATED: backend::amt is unsupported and planned to be removed from os-autoinst eventually';
     return $class->SUPER::new;
 }
 
--- a/backend/amt.pm
+++ b/backend/amt.pm
@@ -38,7 +38,7 @@ sub wsman_cmdline ($self) {
     return ('wsman', '-h', $bmwqemu::vars{AMT_HOSTNAME}, '-P', '16992');
 }
 
-sub wsman ($self, $cmd, $stdin) {
+sub wsman ($self, $cmd, $stdin = undef) {
     my @cmd = $self->wsman_cmdline();
     push(@cmd, split(/ /, $cmd));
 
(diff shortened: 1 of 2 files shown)


okurz:
@marmarek you are the author of backend::amt, right? It looks like the backend is not usable since 2018 so I assume it is also not further required hence marking it as DEPRECATED in preparation for removal. Do you agree?

kalikiana (approved):
Nice. Thanks!

backend/amt.pm
```diff
@@ -95,19 +96,11 @@ sub set_power_state ($self, $power_state) {
...
-        $amt_bootdev = 'Intel(r) AMT: Force PXE Boot';
-    }
-    else {
-        die "Unsupported boot device $bootdev";
-    }
+    my $amt_bootdev = 'Intel(r) AMT: Force ' . {
+        cddvd => 'CD/DVD Boot',
+        hdd => 'Hard-drive Boot',
+        pxe => 'PXE Boot',
+    }->{$bootdev};
```
Martchus, line 103:
```suggestion
    }->{$bootdev} or die "Unsupported boot device $bootdev";
```

perlpunk: Good idea, but that must be in parens because of precedence:
```
my $amt_bootdev = 'Intel(r) AMT: Force ' . ({
...
}->{$bootdev} or die "Unsupported boot device $bootdev");
```

okurz: I guess ` .. || or die "...";` is a common style. Did that now.

perlpunk: That won't help :)
The `"string" . { ... }->{$key}` concatenation is the problem.
```
my $key = "a";
my $str = "foo " . { 
  x => 1, y => 2, z => 3
}->{$key} || die "nope"; 
say "str: $str";
__END__
Use of uninitialized value in concatenation (.) or string at -e line 5.
str: foo 
```

perlpunk: And apart from that, `or die` is the more common style, and `|| die` can be wrong sometimes, e.g.
```
open my $fh, "/tmp/nonsense" || die "Could not open";
```
this won't die.

perlpunk (changes requested):
see comment

backend/amt.pm
okurz: Ok, I thought || is more common. I will change

okurz: done

Martchus (approved):

perlpunk (approved):

marmarek:
> @marmarek you are the author of backend::amt, right? It looks like the backend is not usable since 2018 so I assume it is also not further required hence marking it as DEPRECATED in preparation for removal. Do you agree?

Yes, I agree. Due to lack of (more) AMT-enabled machines, I've shifted my effort to generalhw backend.

okurz:
Ok, that's cool. Thanks for the feedback

PR #1931, opened 2022-01-22
WIP -- Who has an idea how to fix warning 'not a glob reference' -- t: Add test for consoles::amtSol

okurz:


perlpunk:
> Who has an idea how to fix warning 'not a glob reference'

```
my $line = <$amt_console>;
```
For that `$amt_console` needs to be a glob reference, but it is only a Test::MockObject which is probably a hashref and doesn't really support mocking file handles.

Instead of mocking open, kill, IO::Select, I would try to create a pipe with a command like `cat some-test-file`.

PR #1932, opened 2022-01-24
Continue further checks in fullstack test after one fails

Martchus:
This eases debugging as in the failure case all check results will be
available.

Note that this is basically a revert of 90f5eadd1 which does not state why
it was needed/useful.

Related ticket: https://progress.opensuse.org/issues/104971


okurz (approved):

perlpunk (approved):

okurz:
This might need a rebase as the mergify config in master already applies meaning needing the new OBS checks

perlpunk:
The workflow is not yet enabled

PR #1934, opened 2022-01-25
Add simple test for backend::ikvm

okurz:


t/29-backend-ikvm.t
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/perl
+
+use Test::Most;
+use Mojo::Base -strict, -signatures;
+
+use Test::Mock::Time;
+use Test::MockModule;
+use Test::Output qw(stderr_like);
+use Test::Warnings qw(:all :report_warnings);
```
perlpunk, line 9: I'm wondering about importing all functions.
I think none of them is used here, so the `:all` can just be removed.

(Looks like this is true for several other existing tests)

okurz: Right. I was trying to add more test statements where I used these imports but couldn't fix these other tests so I have now removed the unused imports, only in this file though.

perlpunk (approved):

Martchus (approved):

PR #1936, opened 2022-01-26
svirt: Store vmname early for use after test run

kalikiana:
See: https://progress.opensuse.org/issues/104520

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -50,6 +50,7 @@ sub do_start_vm ($self, @) {
         });
 
     $ssh->backend($self);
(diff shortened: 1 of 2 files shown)


Martchus:
Just in case you want to do some fullstack testing: https://github.com/Martchus/openQA-helper#testrun-svirt-backend-locally

(The section "Using QEMU/KVM to test completely locally" says "might not work" because I haven't done it for a long time and it didn't work with all kinds of tests.)

perlpunk (approved):

okurz (approved):

PR #1937, opened 2022-01-26
Run openQA full-stack test as part of os-autoinst CI tests

kraih:
Progress: https://progress.opensuse.org/issues/103788

### DIFF
--- a/.github/workflows/openqa_fullstack.yml
+++ b/.github/workflows/openqa_fullstack.yml
@@ -1,0 +1,21 @@
+---
+name: openQA Fullstack
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  fullstack:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Make project folder user-writable
+        run: sudo chown -R $USER ../../os-autoinst
+      - name: Clone openQA
+        run: git clone --depth 1 https://github.com/os-autoinst/openQA.git ../openQA
+      - name: Install dependencies
+        run: (cd ../openQA; bash -x tools/ci/build_cache.sh)
+      - name: Build os-autoinst
+        run: make symlinks
+      - name: Run unit tests
+        run: make -C ../openQA test-fullstack


.github/workflows/openqa_fullstack.yml
```diff
@@ -0,0 +1,21 @@
...
+on: [push, pull_request]
+jobs:
+  fullstack:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Make project folder user-writable
+        run: sudo chown -R $USER ../../os-autoinst
```
okurz, line 13: I wonder if this is necessary. Can you try without?

.github/workflows/openqa_fullstack.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Clone openQA
+        run: git clone https://github.com/os-autoinst/openQA.git ../openQA
+      - name: Install dependencies
+        run: (cd ../openQA; bash -x tools/ci/build_cache.sh)
+      - name: Build os-autoinst
+        run: bash -x tools/ci/build_autoinst.sh
+      - name: Run unit tests
+        run: (cd ../openQA; tools/ci/run_unit_tests.sh fullstack)
```
okurz, line 21: I suggest

```suggestion
        run: make -C ../openQA test-fullstack
```

because all the additional parameters set by tools/ci/run_unit_tests.sh are likely not useful here, like saving in xunit compatible format and with coverage analysis. That's all additional overhead which we don't use here unless you upload the coverage and XUnit files as GHA artifacts

.github/workflows/openqa_fullstack.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Make project folder user-writable
+        run: sudo chown -R $USER ../../os-autoinst
+      - name: Clone openQA
+        run: git clone https://github.com/os-autoinst/openQA.git ../openQA
+      - name: Install dependencies
+        run: (cd ../openQA; bash -x tools/ci/build_cache.sh)
+      - name: Build os-autoinst
+        run: bash -x tools/ci/build_autoinst.sh
```
okurz, line 19: Instead of a new script file it can be just as simple as

```suggestion
        run: make symlinks
```

and not needing the build script at all

.github/workflows/openqa_fullstack.yml
```diff
@@ -0,0 +1,21 @@
...
+  fullstack:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Make project folder user-writable
+        run: sudo chown -R $USER ../../os-autoinst
+      - name: Clone openQA
+        run: git clone https://github.com/os-autoinst/openQA.git ../openQA
```
okurz, line 15:
```suggestion
        run: git clone --depth 1 https://github.com/os-autoinst/openQA.git ../openQA
```

okurz (changes requested):

.github/workflows/openqa_fullstack.yml
kraih: It is, the directories are not writable by default.

okurz (approved):
I wanted to wait for the CI checks to finish. At least the full stack tests are already successfully completed so that's nice. Bonus would be to check that tests would actually fail if one changes os-autoinst in an incompatible way but today is trust-day so approved :)

Martchus (approved):

PR #1939, opened 2022-01-27
Force full screen update shortly before check screen would fail

Martchus:
Sometimes we are left with a distorted VNC screen despite no stall is
detected. This problem is sometimes even reproducible when running the
fullstack test (fail rate is 5 % on my system), see
https://progress.opensuse.org/issues/104971.

This change forces a full (non-incremental) update of the VNC screen
shortly before a check screen would fail to workaround the issue.

---

I'm currently still running the fullstack test in a loop locally to see whether
it makes a difference. (So far it looks good.)

### DIFF
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -228,7 +228,7 @@ sub as_data ($self) {
 
         $running_total += $lapcopy{time};
         $lapcopy{cumulative} = $running_total;
-        $lapcopy{fraction} = $lapcopy{time} / $data{total_time};
+        $lapcopy{fraction} = $lapcopy{time} / $data{total_time} if $data{total_time};
 
         push @{$data{laps}}, \%lapcopy;
     }
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -32,6 +32,8 @@ use Time::Seconds;
 use English -no_match_vars;
 use OpenQA::NamedIOSelect;
 
+use constant FULL_SCREEN_SEARCH_FREQUENCY => $ENV{OS_AUTOINST_FULL_SCREEN_SEARCH_FREQUENCY} // 5;
+
 # should be a singleton - and only useful in backend process
 our $backend;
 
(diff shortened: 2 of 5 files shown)


okurz (commented):
+1 I thought about this myself and was stuck at the thought that if we fix a situation by enforcing a full screen update then eventually we will run into tests that reproducibly run till the end of timeout and "just work" in the last screen check. So passing test but wasting time. Makes one think why not rely on full screen updates every cycle. What would be the drawback?

coolo:
Next to CPU cycles and network traffic spent not many I'd think.

coolo:
I would think that triggering the !incremental while waiting for assert_screen makes sense. Not only at the end of it. While the test is e.g. typing you better save the CPU.

Martchus:
I've also been thinking about triggering !incremental earlier and more often while waiting for a screen. However, I first wanted to see with how little extra overhead we can already fix the problem of test failures. I suppose with a little more overhead we can indeed also avoid waiting needlessly long in those cases. Considering that the cases are supposedly rare it is hard to say whether the extra overhead pays off.

okurz:
> […] Considering that the cases are supposedly rare it is hard to say whether the extra overhead pays off.

Yes but as I stated I see the risk that we would have a very hard time seeing the other cases at all. So How about triggering !incremental screen captures during check_or_assert_screen on every default iteration which is not "no_wait"?

coolo:
The !incremental update is basically bandwith limited. We only support RAW and ZRLE encoding (both aren't creating these jpg artefacts that normal VNC is so eager to break needle matching with). So if you have a 1024x768 screen in RAW, that's 2.4MB. Requesting that every 2s shouldn't kill any network.

coolo:
See de19bdcc68392abd37adbfe151c0506d02f3525a and 743fe76c3e56447a0bcb2262f4ea5479d42af343 btw. You might face VNC servers that say "screw you, I just gave you a full screen!", but you'll notice quickly I'd say :)

Martchus:
I've split this into multiple commits, maybe it is useful to check the commits separately when reviewing.

okurz (changes requested):
All good with the exception of the lowered code coverage in two locations. The VNC location might be related. Would be good if you could look into both files though.

Martchus:
I only see one affected location and it seems unrelated. The patch coverage is 100 %.

Martchus:
By the way, I was also checking our VNC code for errors (as dropping updates could also be a problem on our side). It looks good except:

```
        # work around buggy addrlink VNC
        next if $encoding_type > 0 && $w * $h == 0;
```

Wouldn't this early return cause our VNC parser to become out-of-sync with the stream as it does not at least seek forward to the next rectangle? But I suppose if that was really our problem we would likely run also in some of the hard errors very soon (and that never happened so far).

coolo:
Well, there is nothing in the stream after that. It's `number of reactangles` and then for each of them x, y, w,h, encoding - and if w*h=0 there is no payload for that rectangle (and any server who sends empty rectangles is pretty stupid). It's possible to imagine an empty rectangle being sent in zrle where you have zlib data after that - and it's possible to imagine zlib data to be more than 0 bytes for an empty rectangle, but I don't believe so.

Martchus:
Ok, if there's no payload (except in contrived cases) we should be good.

okurz (approved):
Ok, reported https://progress.opensuse.org/issues/105759 for the unrelated coverage change. The drop of coverage in consoles/VNC is because of the use of signatures in your patch which is parted of test-covered code so the other code with no coverage is increasing in ratio. That's acceptable for here :)

baierjan (approved):

PR #1940, opened 2022-01-27
Fix include of extension-less perl files in tidy config

okurz:


okurz:
Wait, I don't think this works. Seems to select now *only* files with shebang

baierjan (approved):

Martchus (approved):

PR #1941, opened 2022-01-28
Revert "svirt: Store vmname early for use after test run" due to poo#105690

okurz:
Reverts os-autoinst/os-autoinst#1936, see
https://progress.opensuse.org/issues/105690

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -50,7 +50,6 @@ sub do_start_vm ($self, @) {
         });
 
     $ssh->backend($self);
-    $self->{vmname} = $self->console('svirt')->name;
 
     bmwqemu::save_vars();    # update variables
     return {};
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

backend/svirt.pm
```diff
@@ -179,7 +178,8 @@ sub do_extract_assets ($self, $args) {
     my $first_hdd = $bmwqemu::vars{S390_ZKVM} ? 'a' : 'b';
     my $name = $args->{name};
     my $hdd_num = $args->{hdd_num} - 1;
-    my $svirt_img_name = IMAGE_STORAGE . $self->{vmname} . chr(ord($first_hdd) + $hdd_num) . '.img';
+    my $vmname = $self->console('svirt')->name;
```
kalikiana, line 181: I would expect that this brings us back to `unable to extract assets: Can't call method "name" on an undefined value`?

kalikiana:
See #1942

PR #1943, opened 2022-01-28
WIP -- just for Martchus to get an idea -- scratch suggestions for #1939

okurz:


Martchus:
I've added some of the ideas in https://github.com/os-autoinst/os-autoinst/pull/1939.

PR #1947, opened 2022-02-04
Adjust mergify config for new OBS checks

perlpunk:
Issue: https://progress.opensuse.org/issues/102464

PR #1948, opened 2022-02-07
Avoid non-incremental screen updates in `localXvnc` console

Martchus:
It apparently breaks in some cases, see
https://progress.opensuse.org/issues/106017 where the VNC connection is
attempted to be re-established as it is considered stale (but it fails).
The additional non-incremental update requests have maybe an impact on the
stale VNC detection making it more aggressive. Note that when watching the
test referenced in the ticket with the non-incremental screen update
requests commented out it passes but the screen is not updated for a while.

This is basically a revert of 944b3626c89b47baf7162013a85f2d54b7858131 for
localXvnc.

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -43,6 +43,13 @@ sub callxterm ($self, $command, $window_name) {
     die "cant' start xterm on $display (err: $! retval: $?)" if $@;
 }
 
+sub request_screen_update ($self, $args = undef) {
+    # avoid non-incremental screen updates, see poo#106017
+    # note: It apparently breaks some cases causing the stale VNC detection to force a re-connect which then
+    #       fails with "Connection refused".
+    $self->SUPER::request_screen_update($args) if !(exists $args->{incremental}) || $args->{incremental};
+}
+
 sub fullscreen ($self, $args) {
     my $display = $self->{DISPLAY};
     my $window_name = $args->{window_name};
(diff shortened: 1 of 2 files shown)


okurz (commented):
As you only added a function to localXvnc I assume that unit tests for non-localXvnc cases suffice. But I suggest you hotpatch openqaworker2 and test on there just as we did test the commented lines. After you did that see my comment as approval.

Martchus:
That's what I was doing (uncommented the lines in baseclass, applied the diff from this PR). I started https://openqa.suse.de/tests/8114426#live.

Martchus:
Looks like it works, it is at least passed the critical job module.

okurz (approved):
I agree. The test passed the critical module.

okurz:
with https://github.com/os-autoinst/os-autoinst/pull/1947 not yet merged we need to handle this manually

Martchus:
By the way, I've looked into the code to understand a little bit better what's going on:

* The VNC stall detection likely became more aggressive due to the additional non-incremental screen updates because the stall detection basically happens before sending an update request. The additional non-incremental screen updates are done in their own interval ignoring the backends normal `update_request_interval`. It now feels a bit odd having it implemented that way but I suppose that's ok considering the non-incremental screen updates are only requested when waiting for a specific needle to match.
* I *suspect* that our Xvnc instance will never survive re-establishing the connection because the `-inetd` flags is used and I presume it causes Xvnc to terminate as soon as the socket is closed. So at least the re-connect done in the VNC stall detection doesn't seem to make much sense for `localXvnc` in general. If that's true we might want to disable the VNC stall detection for `localXvnc`.

Martchus:
Here's a draft showing how it would look like: https://github.com/os-autoinst/os-autoinst/pull/1949

PR #1949, opened 2022-02-07
Disable checking for VNC stalls in localXvnc

Martchus:
It would not work anyways because `Xvnc` stops when closing the socket
(which is done when a VNC stall is detected before establishing a new
VNC connection).

This should make it possible to revert https://github.com/os-autoinst/os-autoinst/commit/bf8e7f7ae71673046ad18a948cc2e9f7a7be9ced, see https://github.com/os-autoinst/os-autoinst/pull/1948#issuecomment-1031564322.

I needed to test it on some worker first, of course.


Martchus:
I've just tested our Xvnc setup via vncviewer (using https://github.com/os-autoinst/os-autoinst/commit/8f4205a4087bd888ff18db2b29137fa087206779) and it is indeed impossible to re-connect (when closing the connection Xvnc terminates). So I suppose this change makes generally sense. Alternatively we could change the setup to avoid Xvnc being terminated automatically (by avoiding `-inetd`). However, that would likely complicate the setup.

kalikiana:
> I've just tested our Xvnc setup via vncviewer (using [8f4205a](https://github.com/os-autoinst/os-autoinst/commit/8f4205a4087bd888ff18db2b29137fa087206779)) and it is indeed impossible to re-connect (when closing the connection Xvnc terminates). So I suppose this change makes generally sense. Alternatively we could change the setup to avoid Xvnc being terminated automatically (by avoiding `-inetd`). However, that would likely complicate the setup.

How about going with this since and a time-boxed ticket to change the setup? It could be useful, but maybe it has new side-effects

Martchus:
It took quite a while until my test job was scheduled so I've created another one (to be sure it really ran with my changes): https://openqa.suse.de/tests/8163465

> How about going with this since and a time-boxed ticket to change the setup? It could be useful, but maybe it has new side-effects

Likely we have more important things in the backlog. So I'd refrain from it unless we see a necessity for the VNC stale detection when using Xvnc. (I suppose we only saw problems with QEMU so far, right?)

Martchus:
It works, the test job already passed the critical point (where it otherwise always reliably failed without the change being reverted here).

consoles/localXvnc.pm
```diff
@@ -96,12 +89,14 @@ sub activate ($self) {
...
     close($s);
 
-    $self->connect_remote(
+    my $vnc = $self->connect_remote(
         {
             hostname => "localhost",
             port => $port,
             ikvm => 0
         });
+    # disable checking VNC stalls as this setup would not survive possible re-connects anyways
```
okurz, line 98: As you put in a good explanation in the method "request_screen_update" that you remove with this change I suggest you move at least part of that explanation here to explain why.

okurz (changes requested):
CI tests fail both in GHA and OBS

Martchus:
Updated the comment and adapted tests (as the mocking was no longer sufficient).

okurz (approved):

kalikiana (approved):

PR #1952, opened 2022-02-11
Allow network cable to be unplugged

foursixnine:
There are cases where during a test, we need to isolate the SUT from the network one way or the other, in the physical world we would unplug the network cable, so we can do the same for qemu via qmp commands.

For the time being, the network name is hardcoded to the default one, if system has two nics, code will have to be adjusted (if ever needed)

VR: http://phobos.qa.suse.de/tests/3857658#step/systemd_sapstart_check/45

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -279,6 +279,13 @@ sub _migrate_to_file ($self, %args) {
     return $self->_wait_for_migrate();
 }
 
+sub switch_network ($self, $args) {
+    $self->handle_qmp_command({execute => 'set_link', arguments => {
+                name => $args->{network_link_name} // "qanet0",
+                up => (!defined $args->{network_enabled} || $args->{network_enabled} ? Mojo::JSON->true : Mojo::JSON->false)
+    }}, fatal => 1);
+}
+
 sub save_memory_dump ($self, $args) {
     my $fdname = 'dumpfd';
     my $vars = \%bmwqemu::vars;
--- a/t/18-backend-qemu.t
+++ b/t/18-backend-qemu.t
@@ -87,6 +87,23 @@ subtest 'eject cd' => sub {
     is_deeply $called{handle_qmp_command}[1], \%custom_remove_params, 'blockdev-remove-medium called with custom parameters';
 };
 
+subtest 'switch_network' => sub {
+    my %switch_network_params = (arguments => {name => 'qanet0', up => Mojo::JSON->false}, execute => 'set_link');
+    $called{handle_qmp_command} = undef;
+
+    $backend->switch_network({network_enabled => 0});
+    ok(exists $called{handle_qmp_command}, 'network must be disabled');
+    is_deeply($called{handle_qmp_command}[0], \%switch_network_params, 'qmp command for setlink is passed');
+
+    $called{handle_qmp_command} = undef;
+    $backend->switch_network({network_enabled => 1, network_link_name => 'bingo'});
+    %switch_network_params = (arguments => {name => 'bingo', up => Mojo::JSON->true}, execute => 'set_link');
+    ok(exists $called{handle_qmp_command}, 'a qmp command has been called');
+    is_deeply($called{handle_qmp_command}[0], \%switch_network_params, 'Network name can be specified, network can be enabled');
+
+    $called{handle_qmp_command} = undef;
+};
+
 subtest 'execute arbitrary QMP command' => sub {
     my %query = (execute => 'foo', arguments => {bar => 1});
     $called{handle_qmp_command} = undef;
--- a/testapi.pm
+++ b/testapi.pm
@@ -51,6 +51,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   record_soft_failure record_info force_soft_failure
   become_root x11_start_program ensure_installed eject_cd power
 
+  switch_network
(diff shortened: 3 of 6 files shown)


backend/qemu.pm
```diff
@@ -279,6 +279,15 @@ sub _migrate_to_file ($self, %args) {
     return $self->_wait_for_migrate();
 }
 
+
+sub toggle_network ($self, $args) {
+    bmwqemu::diag $self->handle_qmp_command({execute => 'set_link', arguments => {
+                name => "qanet0",
```
okurz, line 285: you can just use `$bmwqemu::vars{NETWORK_LINK_NAME} // "qanet0"`

testapi.pm
```diff
@@ -1982,6 +1983,20 @@ sub eject_cd {
...
 
+=head2 toggle_network
+
+  toggle_network network_enabled => $boolean;
+
+Changes the default qanet0 device's state, fatal if fails.
+
+=cut
+
+sub toggle_network {
```
okurz, line 1994: as you add a new function to the test API please bump the API version in OpenQA/Isotovideo/Interface.pm

testapi.pm
```diff
@@ -1982,6 +1983,20 @@ sub eject_cd {
     query_isotovideo(backend_eject_cd => \%nargs);
 }
 
+=head2 toggle_network
```
okurz, line 1986: I wonder if we can find a better name. "toggle" sounds like it would switch back and forth, not necessarily knowing if it was on or off in before. Maybe "switch_network"? Or separate "enable_network", "disable_network"?

okurz (changes requested):

testapi.pm
foursixnine: Had the same thought before submitting the PR: `switch_network` fits better here

alvarocarvajald (approved):
LGTM

Martchus (approved):

okurz (approved):

t/18-backend-qemu.t
```diff
@@ -87,6 +87,17 @@ subtest 'eject cd' => sub {
...
+    my %switch_network_params = ('arguments' => {'name' => 'qanet0', 'up' => Mojo::JSON->false}, 'execute' => 'set_link');
+    $called{handle_qmp_command} = undef;
+
+    $backend_mock->redefine(handle_qmp_command => sub { push @{$called{handle_qmp_command}}, $_[1] });
+    $backend->switch_network({network_state => 0});
+    ok(exists $called{handle_qmp_command}, 'a qmp command has been called');
+    is_deeply($called{handle_qmp_command}[0], \%switch_network_params, 'qmp command for setlink is passed') or diag explain $called{handle_qmp_command};
+    $called{handle_qmp_command} = undef;
+};
+
```
okurz, line 100: something from here is causing unhandled test output with a bare "1" that should be prevented or caught with Test::Output, see https://github.com/os-autoinst/os-autoinst/runs/5156844864?check_suite_focus=true#step:3:589

okurz (changes requested):

t/18-backend-qemu.t
foursixnine: Yeah, was actually a `bmwqemu::diag` call on the `backend::qemu::switch_network` implementation

okurz:
5: /opt/t/18-backend-qemu.t: Hash key 'arguments' with quotes at line 91, column 34.  Avoid useless quotes.  (Severity: 5)
5: /opt/t/18-backend-qemu.t: Hash key 'name' with quotes at line 91, column 50.  Avoid useless quotes.  (Severity: 5)
5: /opt/t/18-backend-qemu.t: Hash key 'up' with quotes at line 91, column 70.  Avoid useless quotes.  (Severity: 5)
5: /opt/t/18-backend-qemu.t: Hash key 'execute' with quotes at line 91, column 98.  Avoid useless quotes.  (Severity: 5)

foursixnine:
Last commit should do it :)

t/18-backend-qemu.t
```diff
@@ -87,6 +87,23 @@ subtest 'eject cd' => sub {
...
+    $called{handle_qmp_command} = undef;
+
+    $backend->switch_network({network_state => 0});
+    ok(exists $called{handle_qmp_command}, 'a qmp command has been called');
+    is_deeply($called{handle_qmp_command}[0], \%switch_network_params, 'qmp command for setlink is passed') or diag explain $called{handle_qmp_command};
+
+    $called{handle_qmp_command} = undef;
+    $backend->switch_network({network_state => 1, network_link_name => 'bingo'});
+    %switch_network_params = (arguments => {name => 'bingo', up => Mojo::JSON->false}, execute => 'set_link');
+    ok(exists $called{handle_qmp_command}, 'a qmp command has been called');
```
foursixnine, line 101: This should have failed

okurz: so what should I do now?

foursixnine: see the last commit :)

okurz (approved):

backend/qemu.pm
```diff
@@ -279,6 +279,13 @@ sub _migrate_to_file ($self, %args) {
     return $self->_wait_for_migrate();
 }
 
+sub switch_network ($self, $args) {
```
foursixnine, line 282: The libvirt equivalent to this would be: `virsh domif-setlink`

PR #1953, opened 2022-02-11
Fix default VNC connect timeout

Martchus:
Fix

```
Use of uninitialized value $connect_timeout in addition (+) at /usr/lib/os-autoinst/consoles/VNC.pm line 137.
```

See https://progress.opensuse.org/issues/106654

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -117,7 +117,8 @@ my @encodings = (
     },
 );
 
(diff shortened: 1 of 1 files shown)


Martchus:
This is basically the fix suggested here (which somehow didn't make it into the PR): https://github.com/os-autoinst/os-autoinst/pull/1789/files#r782174191

okurz (approved):

PR #1956, opened 2022-02-15
Avoid warning because of missing second argument

perlpunk:
Issue: https://progress.opensuse.org/issues/106684

The second argument is not required, but when it is undef, it will
create a warning further down the road.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -233,6 +233,7 @@ sub record_info {
     my ($title, $output, %nargs) = @_;
     $nargs{result} //= 'ok';
     die 'unsupported $result \'' . $nargs{result} . '\'' unless _is_valid_result($nargs{result});
+    $output //= '';
     bmwqemu::log_call(title => $title, output => $output, %nargs);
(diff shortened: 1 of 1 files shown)


okurz (approved):

testapi.pm
```diff
@@ -233,6 +233,7 @@ sub record_info {
     my ($title, $output, %nargs) = @_;
     $nargs{result} //= 'ok';
     die 'unsupported $result \'' . $nargs{result} . '\'' unless _is_valid_result($nargs{result});
+    $output //= '';
```
okurz, line 236: The synopsis in line 216 makes it look like `$output` is required but we never introduced that as a really mandatory parameter … until we will have full signature checks in https://github.com/os-autoinst/os-autoinst/pull/1696/files#diff-3420c7e01f73caee7d97053af39990e3297daf42d06e0fc8848f78af1a5e009eR228

I guess I will change that to provide `''` as default

okurz (approved):

kalikiana (approved):

Martchus (approved):

PR #1957, opened 2022-02-15
t: Unset GIT env variables during test

perlpunk:
Issue: https://progress.opensuse.org/issues/106865

Unfortunately not easy to test because during rebase it will first have the older version of the script without the fix.

I found out by running a copy of the test with verbose output and saw from the git output that it was doing the operations in the os-autoinst git directory. Then I had a suspicion and dumped `%ENV`.


okurz (approved):
Awesome :)

perlpunk:
I also added a `die` in case the git init stuff fails. A verbose output of the commands would be best, though.

okurz (approved):

PR #1960, opened 2022-02-18
snapshots: Record info files to visualize snapshot loading

kalikiana:
Since snapshot loading can have side-effects, a recorded info file allows the web UI to clearly indicate when a snapshot was loaded.

See: https://progress.opensuse.org/issues/103791

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -345,9 +345,11 @@ sub runalltests {
         if (!$vmloaded && $fullname eq $firsttest) {
             if ($bmwqemu::vars{SKIPTO}) {
                 if ($bmwqemu::vars{TESTDEBUG}) {
+                    $current_test->record_resultfile('loading snapshot', 'TESTDEBUG set');
                     load_snapshot('lastgood');
                 }
                 else {
+                    $current_test->record_resultfile('loading snapshot', 'skipping to first test');
                     load_snapshot($firsttest);
                 }
             }
@@ -393,12 +395,15 @@ sub runalltests {
                 return 0;
             }
             elsif (!$flags->{no_rollback} && $last_milestone) {
+                $current_test->record_resultfile('loading snapshot', 'milestone requested');
                 load_snapshot('lastgood');
                 $last_milestone->rollback_activated_consoles();
             }
         }
         else {
             if (!$flags->{no_rollback} && $last_milestone && $flags->{always_rollback}) {
+                die 'Unable to rollback as no tests have been loaded' unless $current_test;
+                $current_test->record_resultfile('loading snapshot', 'milestone requested, always rollback flag set');
                 load_snapshot('lastgood');
                 $last_milestone->rollback_activated_consoles();
             }
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -89,8 +89,11 @@ my ($reverts_done, $snapshots_made) = (0, 0);
 $mock_autotest->redefine(load_snapshot => sub { $reverts_done++ });
 $mock_autotest->redefine(make_snapshot => sub { $snapshots_made++ });
 $mock_autotest->redefine(query_isotovideo => 0);
+my @records;    # array of recorded info files (sans nargs)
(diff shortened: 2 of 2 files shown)


autotest.pm
```diff
@@ -345,9 +345,11 @@ sub runalltests {
         if (!$vmloaded && $fullname eq $firsttest) {
             if ($bmwqemu::vars{SKIPTO}) {
                 if ($bmwqemu::vars{TESTDEBUG}) {
+                    testapi::record_info('loading snapshot', 'TESTDEBUG set');
```
okurz, line 348:
```suggestion
                    $current_test;->record_resultfile('loading snapshot', 'TESTDEBUG set', result => 'ok');
```

Please don't call the external facing test API from internal functions. This is following "design by contract".

okurz (changes requested):

autotest.pm
```diff
@@ -345,9 +345,11 @@ sub runalltests {
         if (!$vmloaded && $fullname eq $firsttest) {
             if ($bmwqemu::vars{SKIPTO}) {
                 if ($bmwqemu::vars{TESTDEBUG}) {
+                    $current_test->record_resultfile('loading snapshot', 'TESTDEBUG set');
```
okurz, line 348: are you sure this works without adding `result => 'ok'`?

autotest.pm
```diff
@@ -393,12 +395,15 @@ sub runalltests {
...
             }
             elsif (!$flags->{no_rollback} && $last_milestone) {
+                $current_test->record_resultfile('loading snapshot', 'milestone requested');
                 load_snapshot('lastgood');
                 $last_milestone->rollback_activated_consoles();
             }
         }
         else {
             if (!$flags->{no_rollback} && $last_milestone && $flags->{always_rollback}) {
+                die "ERROR: no tests loaded" unless $current_test;
```
okurz, line 405: did you ever encounter this case? I don't see you adding a test case for this so I wonder under what circumstances this would be needed. Please also use single tick strings by default

Martchus: It looks like the case where one would use the `assert`-macro in C. Likely we should trust our understanding of the code that this case can never happen or we at least use a better message with a bit more context (e.g. "Unable to roll back as no tests have been loaded.") and add a test.

autotest.pm
Martchus: Good question. And "works" means there's no `undef` being recorded (but a meaningful result) - even if it has no visual impact.

autotest.pm
okurz: Right. I wouldn't have worried if the complete module has been introduced with this. I mostly wonder why the author introduced this line :)

autotest.pm
kalikiana: It's documented to be allowed, and results in `unk` afair

autotest.pm
kalikiana: And I tried to inestigate it but honestly I don't get how we can be loading a snapshot when there's no test here

autotest.pm
okurz: Well, but as here there is nothing failing I recommend to mark the step as "ok", or why do you want to differ from the original proposal?

autotest.pm
okurz: Sure, trying to load a snapshot when there is no test won't work. The question was how you actually encountered the case of no test being  defined here. You stated you found out by "running the tests". I don't understand what you mean. You haven't added tests for this and previously CI tests have passed, so which tests do you mean? Manual tests?

kalikiana: The unit tests don't pass because they die due to the undefined value, even without the explicit `die` message. You can see this in CI.

okurz: I tried out myself and progressed a bit but haven't succeeded completely with the `record_resultfile` approach. This is my diff:

okurz:
cdywan and me came up with an alternative: https://github.com/os-autoinst/os-autoinst/pull/1987

kalikiana:
#1987 is a much better approach

PR #1963, opened 2022-02-23
Fix flaky poweroff in openQA full stack tests with direct "sudo poweroff"

okurz:
Instead of ending up with typing "sudo su" and then typing (too soon)
"poweroff" in a prompt that is not ensured to be present and responsive
we should just type "sudo poweroff" to avoid any need for checking the
subsequent root prompt readyness. This is straightforward, simpler than
the previous approach and should fix those instabilities which we saw
especially in the openQA full stack test using the os-autoinst test
module definitions from here.

Related progress issue: https://progress.opensuse.org/issues/106912

### DIFF
--- a/t/data/tests/tests/shutdown.pm
+++ b/t/data/tests/tests/shutdown.pm
@@ -5,8 +5,7 @@ use Mojo::Base 'basetest', -signatures;
 use testapi;
 
 sub run {
-    type_string "sudo su\n";
-    type_string "poweroff\n";
+    enter_cmd 'sudo poweroff';
     assert_shutdown(get_var('INTEGRATION_TESTS') ? 90 : undef);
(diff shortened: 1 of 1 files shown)


Martchus (approved):
It is definitely better but I'm wondering whether it'll really do the trick. It seems that in the failure I've investigated only the last return was missing. It would likely make sense to create a stability test PR.

kalikiana (approved):
Nice catch

kalikiana:
> It is definitely better but I'm wondering whether it'll really do the trick. It seems that in the failure I've investigated only the last return was missing. It would likely make sense to create a stability test PR.

Maybe we shouldn't need a PR for that anymore than we do for openQA? #1968

PR #1964, opened 2022-02-23
PART 2: t: Use more 'enter_cmd' where it simplifies things - after #1963

okurz:
Depends on
* #1963


kalikiana (approved):
Not sure I see a significant benefit, but I like this for exercising this API more

baierjan (approved):

PR #1974, opened 2022-02-28
testapi: Fix validate_script_output() argument parsing

cfconrad:
Not all possible arguments where specified in `validate_script_output()`
`testapi::compat_args()` usage. This caused problems when calling it like
```
validate_script_output( $script, $match, proceed_on_failure => 1);
```
Here `testapi::compat_args()` would use the string 'proceed_on_failure'
as positioned timeout value, as it isn't given as a known key.

Also the default values for `script_output()` call inside of
`validate_script_output()` should not be set. E.g. the timeout was
explicit set to 30 in `validate_script_ouput()`.

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -98,7 +98,7 @@ use testapi qw(is_serial_terminal :DEFAULT);
 use basetest;
 
 my $mock_basetest = Test::MockModule->new('basetest');
-$mock_basetest->redefine(_result_add_screenshot => sub { my ($self, $result) = @_; });
+$mock_basetest->noop('_result_add_screenshot');
 $autotest::current_test = basetest->new();
 
 # we have to mock out wait_screen_change for the type_string tests
@@ -108,13 +108,10 @@ my $mod2 = Test::MockModule->new('testapi');
 my $mock_bmwqemu = Test::MockModule->new('bmwqemu');
 
 subtest 'type_string' => sub {
-    ## no critic (ProhibitSubroutinePrototypes)
-    sub fake_wait_screen_change : prototype(&@) {
-        my ($callback, $timeout) = @_;
-        $callback->() if $callback;
-    }
-
-    $mod2->redefine(wait_screen_change => \&fake_wait_screen_change);
+    $mod2->redefine(wait_screen_change => sub : prototype(&@) {
+            my ($callback, $timeout) = @_;
+            $callback->() if $callback;
+    });
 
     stderr_like { type_string 'hallo' } qr/<<< testapi::type_string/, 'type_string log output';
     is_deeply $cmds, [{cmd => 'backend_type_string', max_interval => 250, text => 'hallo'}], 'type_string called';
@@ -259,7 +256,7 @@ is(current_console, 'a-console', 'Current console is the a-console');
 
 subtest 'script_run' => sub {
     # just save ourselves some time during testing
-    $mock_bmwqemu->redefine(wait_for_one_more_screenshot => sub { sleep 0; });
+    $mock_bmwqemu->noop('wait_for_one_more_screenshot');
 
     $testapi::serialdev = 'null';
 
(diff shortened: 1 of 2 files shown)


t/03-testapi.t
```diff
@@ -682,6 +690,7 @@ subtest 'compat_args' => sub {
 subtest 'check quiet option on script runs' => sub {
     $bmwqemu::vars{_QUIET_SCRIPT_CALLS} = 1;
     my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->redefine(script_output => sub { return 'output'; });
```
okurz, line 693:
```suggestion
    $mock_testapi->redefine(script_output => 'output');
```

t/03-testapi.t
```diff
@@ -580,6 +580,14 @@ subtest 'validate_script_output' => sub {
...
         qr/coderef or regexp/
     );
+
+    my $exp_args = [123, proceed_on_failure => 1, type_command => 1];
+    $mock_testapi->redefine(script_output => sub {
+            my ($script, @args) = @_;
+            is_deeply(\@args, $exp_args, 'Arguments successful passed');
+            return 'foo';
+    });
+    validate_script_output('script', qr/foo/, @$exp_args);
```
okurz, line 590: We should use proper function signatures where we can. We have converted all tests to use signatures but not yet the testapi.pm module itself. Please try if this works

```suggestion
    $mock_testapi->redefine(script_output => sub ($script, @args) {
            is_deeply(\@args, $exp_args, 'Arguments successful passed');
            return 'foo';
    });
    validate_script_output('script', qr/foo/, @$exp_args);
```

but I recommend against putting Test::More functions into mocked functions because on failure the output would point to the source of the mocked function, not to where this is called. So better to avoid actual test instructions within any mocked or actually any helper functions and just use them only on the highest level of the test.

okurz (changes requested):
I unfortunately fail to parse your git commit message. Could you please rephrase "Equal def arguments for …"?

t/03-testapi.t
cfconrad: done

t/03-testapi.t
cfconrad: done

cfconrad:
> I unfortunately fail to parse your git commit message. Could you please rephrase "Equal def arguments for …"?

Neither do I, thx for pointing ;-)

Martchus (approved):

t/03-testapi.t
perlpunk: I still see the old code...

t/03-testapi.t
```diff
@@ -564,22 +564,31 @@ subtest 'script_output' => sub {
...
+    throws_ok {
+        validate_script_output('script', sub { m/error/ })
+    } qr/output not validating/, 'Die on output not match';
+    throws_ok {
+        validate_script_output('script', ['Invalid parameter'])
+    } qr/coderef or regexp/, 'Die on invalid parameter';
+
+    $mock_testapi->redefine(script_output => sub ($script, @args) {
+            return join(',', @args);
+    });
```
okurz, line 580:
```suggestion
    $mock_testapi->redefine(script_output => sub ($script, @args) { join(',', @args) });
```

sorry I missed this from my earlier suggestion

t/03-testapi.t
okurz: Me too. I think you only changed https://github.com/os-autoinst/os-autoinst/pull/1974/files#diff-97e29b53bd6b0cc520266e30da42f72439278f368a9d72db83678f535f1c8c63R567

okurz (changes requested):
Please also observe the suggestion regarding the API version

t/03-testapi.t
cfconrad: ups, yeah. I didn't touched this part, thats why I did not changed it. 
Np I will clean this part up too.

okurz (changes requested):
Code change is fine. Just rephrase the second commit from "03-testapi.t: Code refactoring" to "t: Refactor 03-testapi.t"

okurz (approved):

PR #1978, opened 2022-03-07
Consistently provide all scripts in script/

okurz:


Martchus (changes requested):
You still need to adapt paths used to reference those scripts.

okurz:
https://github.com/os-autoinst/os-autoinst/actions/runs/9971827632/job/27553725203?pr=1978 fails with

```
Can't locate log.pm in @INC
```

don't know why

script/isotovideo
```diff
@@ -0,0 +1,201 @@
...
+# Details: https://progress.opensuse.org/issues/90371
+use JSON::PP;
+
+my $installprefix;    # $bmwqemu::topdir
+
+BEGIN {
+    # the following line is modified during make install
+    $installprefix = undef;
+
+    my ($wd) = $0 =~ m-(.*)/-;
```
perlpunk, line 100: This needs adjustment now.
```
perl script/worker --isotovideo=../os-autoinst/isotovideo --verbose
# $0 == /home/tina/openqa-toolbox/repos/os-autoinst/script/isotovideo
Can't locate log.pm in @INC (you may need to install the log module) (@INC contains: /home/tina/openqa-toolbox/repos/os-autoinst/script ...
```

perlpunk, line 100:
```suggestion
    use FindBin '$RealBin';
    use Cwd qw(realpath);
    my ($wd) = realpath "$RealBin/..";
```

okurz: I strongly suggest we keep any real content changes to the scripts outside this PR. I really only moved isotovideo to script/isotovideo here. I don't understand what `my ($wd) = $0 =~ m-(.*)/-;` is doing. Could you please create your own PR for that change, we apply that first and I will rebase my PR on top of your changes?

perlpunk: yep

perlpunk: #2518 2518

perlpunk: You still need to add the `/..` in your PR though

okurz: done

CMakeLists.txt
```diff
@@ -242,10 +242,10 @@ install(
     COMPONENT "openvswitch"
     EXCLUDE_FROM_ALL
 )
-foreach (EXTRA_SCRIPT check_qemu_oom dewebsockify vnctest)
+foreach (EXTRA_SCRIPT script/check_qemu_oom script/dewebsockify script/vnctest)
     install(
         FILES "${EXTRA_SCRIPT}"
-        DESTINATION "${OS_AUTOINST_DATA_DIR}"
+        DESTINATION "${OS_AUTOINST_DATA_DIR}/script"
```
Martchus, line 248: You created a symlink for `check_qemu_oom`. This symlink should probably be installed here as well.

script/isotovideo
```diff
@@ -0,0 +1,202 @@
...
+
+use FindBin '$RealBin';
+
+my $installprefix;    # $bmwqemu::topdir
+
+BEGIN {
+    # the following line is modified during make install
+    $installprefix = undef;
+
+    my $topdir = "$RealBin/.." || '..';
```
perlpunk, line 102: the `|| ...` is actually useless, as the lefthand side will always be true

```suggestion
    my $topdir = "$RealBin/..";
```

CMakeLists.txt
okurz: I wanted the symlinks only for convenience for developers. Everything in OS_AUTOINST_DATA_DIR isn't supposed to be called by normal users and code that relies on `check_qemu_oom` calls it from the script sub directory. Hence I would not install the symlink, fine?

b10n1k:
I dont understand the motivation of this commit. why move them in script and not in bin? which is empty btw

okurz:
> I dont understand the motivation of this commit. why move them in script and not in bin? which is empty btw

Good question. I don't even clearly remember anymore. Let's say because https://github.com/os-autoinst/openQA/tree/master/script and https://github.com/mojolicious/mojo/tree/main/script do it like this.

Martchus:
`bin` is a relict from the time the project was using autotools. We should remove this directory as we never want to put binaries under version control.

Martchus (approved):
Let's see what this breaks. (I added the "not-ready" label so you can merge this in a time window where you can follow-up on it if necessary.)

okurz:
fixup: #2527

PR #1979, opened 2022-03-07
Use proper perl "signatures" in common files - distribution

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -3,14 +3,12 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package distribution;
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 
 use testapi ();
 use Carp 'croak';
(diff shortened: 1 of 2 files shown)


baierjan (approved):

kalikiana (approved):

okurz:
Patched on openqaworker7 with

```
ssh -J o3 root@openqaworker7 "curl -s https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/1979.patch | patch --backup -p1 -d /usr/lib/os-autoinst/"
```

monitoring some jobs

Martchus (approved):

okurz:
Updated

```diff
- sub new ($class) {
+ sub new ($class, @) {
```

patched w7 and monitoring tests.

PR #1981, opened 2022-03-07
Fix off-by-one error in color calculation

okurz:
Regression introduced in 8ec50014


okurz:
> Great PR! Please pay attention to the following items before merging:
> 
> 
>     * [ ]  Consider running manual verification tests on real hardware, especially for non-qemu backends

I think the risk is negligible for this regression fix of the logic. Just an analysis of the code should suffice.

baierjan (approved):

PR #1982, opened 2022-03-08
Add error handling for uninitialized isotovideo socket

okurz:
When trying to call test API functions from schedule code, e.g. the
main.pm one would get an error message like:

```
myjsonprc: called on undefined file descriptor at
/usr/lib/os-autoinst/myjsonrpc.pm line 40.
```

which is not very helpful to beginners. The problem is that test API
functions which only work as expected from within test modules are
called by users in unexpected cases, e.g. from scheduling code like a
main.pm file. In this case the isotovideo socket was not yet initialized
and we would try to execute a command on the worker, not the system
under test, for example a virtual machine.

This commit introduces an explicit message with a hint to the user what
to check in case they try to call tests in the wrong way where we would
end up calling the send_json function within an uninitialized object.

The problem was originally discovered when supporting a user receiving
such unexpected errors.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -317,7 +317,7 @@ sub query_isotovideo {
     }
     $json{cmd} = $cmd;
 
-    # send the command to isotovideo
+    die "isotovideo is not initialized. Ensure that you only call test API functions from test modules, not schedule code\n" unless defined $isotovideo;
     myjsonrpc::send_json($isotovideo, \%json);
 
     # wait for response (if test is paused, this will block until resume)
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -100,6 +100,7 @@ use basetest;
 my $mock_basetest = Test::MockModule->new('basetest');
 $mock_basetest->noop('_result_add_screenshot');
 $autotest::current_test = basetest->new();
+$autotest::isotovideo = 1;
 
 # we have to mock out wait_screen_change for the type_string tests
 # that use it, as it doesn't work with the fake send_json and read_json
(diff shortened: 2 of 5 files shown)


Martchus (approved):

autotest.pm
```diff
@@ -317,7 +317,7 @@ sub query_isotovideo {
     }
     $json{cmd} = $cmd;
 
-    # send the command to isotovideo
+    die "isotovideo is not initialized. Ensure to call test API calls only from test modules, not schedule code\n" unless defined $isotovideo;
```
kalikiana, line 320:
```suggestion
    die "isotovideo is not initialized. Be sure to only call test API from test modules, not schedule code\n" unless defined $isotovideo;
```

okurz: That's wrong. Now we are missing an article and I think you misjudged "calls" as being a verb but I meant it as plural of "test API calls" :)

okurz:
sporadic issue reported in https://progress.opensuse.org/issues/107998 about t/29-backend-driver.t

autotest.pm
kalikiana: I don't see any missing articles, I just changed it because "ensure to call" feels wrong to me.

How about this:

    ```suggestion
    die "isotovideo is not initialized. Please ensure that you only call test API from test modules, not schedule code\n" unless defined $isotovideo;
```

Comment (okurz) [reply autotest.pm:320]: That's still missing an article and not proper english. "ensure that you only call test API" should rather say "ensure that you only call the test API". But again the word "calls" is missing. We could replace "calls" with "functions" though. So

```suggestion
    die "isotovideo is not initialized. Ensure that you only call test API functions from test modules, not schedule code\n" unless defined $isotovideo;
```

kalikiana: Thank you for saying what you meant. Sounds good to me

kalikiana (approved):

PR #1985, opened 2022-03-08
backend::svirt: Extract methods _is_hyperv/_is_vmware

okurz:


backend/svirt.pm
```diff
@@ -312,7 +314,7 @@ sub get_wait_still_screen_on_here_doc_input ($self, $args) {
     # by comparing the screen and checking that nothing else will write on it.
     # So if the here-document input is really slow, we hope the wait_still_screen
     # takes even longer.
-    ($bmwqemu::vars{VIRSH_VMM_FAMILY} // '') =~ qr/^hyperv|vmware$/ ? 1 : 0;
+    _vmm_family() =~ qr/^hyperv|vmware$/ ? 1 : 0;
```
perlpunk, line 317: Not related to the change, but I guess this regex should rather be: `qr/^(hyperv|vmware)$/` (otherwise it would allow `hypervsomething` or `somethingvmware`)

perlpunk (approved):

backend/svirt.pm
okurz: good catch. I added an according commit to this PR

Martchus (approved):

okurz:
hm, codecov change missing. Trying rebase

@mergifyio rebase

okurz:
that's annoying. We managed to pull in the openQA dependency handling problem also into os-autoinst in https://github.com/os-autoinst/os-autoinst/runs/5484977053?check_suite_focus=true now :,(

PR #1986, opened 2022-03-08
Use signatures in most of backend/ and consoles/ as possible

okurz:


okurz:
Note to self, fails with

```
#   Failed test 'Supported machines listed'
#   at ./18-qemu-options.t line 82.
```

baierjan (approved):

PR #1987, opened 2022-03-09
Record info in next test module after loading snapshot

okurz:
After a test module is rolled back, i.e. a snapshot is loaded, the
system under test can behave different, e.g. not being fully responsive
yet or simply presenting a different starting situation. To make the
potential impact more obvious to test reviewers we now record a specific
test result step informing in the next module that a snapshot was loaded
in before.

Related progress issue: https://progress.opensuse.org/issues/103791

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -372,6 +372,7 @@ sub runalltests {
         eval { $t->runtest; };
         my $error = $@;    # save $@, it might be overwritten
         $t->save_test_result();
+        my $next_test = $testindex eq $#testorder ? undef : $testorder[$testindex + 1];
 
         if ($error) {
             my $msg = $error;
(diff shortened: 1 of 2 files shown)


Martchus (approved):

kalikiana (approved):
I think this is looking rather nice now

autotest.pm
```diff
@@ -372,6 +372,7 @@ sub runalltests {
         eval { $t->runtest; };
         my $error = $@;    # save $@, it might be overwritten
         $t->save_test_result();
+        my $next_test = $testindex eq $#testorder ? undef : $testorder[$testindex + 1];
```
perlpunk, line 375: Why `eq` and not `==`?

perlpunk: Just `my $next_test = $testorder[$testindex + 1]` would have worked, too. Why the extra check? It's perl, not python ;-)

okurz: Ok, sure. @kalikiana and me were just too cautious ;) Created https://github.com/os-autoinst/os-autoinst/pull/1989

PR #1988, opened 2022-03-09
Only load snapshots if there is actually a next module

okurz:
After:
* https://github.com/os-autoinst/os-autoinst/pull/1987

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -381,14 +381,14 @@ sub runalltests () {
                 bmwqemu::stop_vm();
                 return 0;
             }
-            elsif (!$flags->{no_rollback} && $last_milestone) {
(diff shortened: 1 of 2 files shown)


okurz:
This should be tested in a manual test with openQA where we actually load snapshots

okurz:
@Mergifyio rebase

Martchus (commented):
Looks generally good but yes, should be tested manually.

okurz:
Deployed on openqaworker7.o.o with `curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/1988.patch | patch -p1 --directory=/usr/lib/os-autoinst`

Martchus (approved):

mimi1vx (approved):

PR #1990, opened 2022-03-14
myjsonrpc: Don't limit how often select may return EINTR

Vogtinator:
select returns EINTR if the wait was interrupted by any signal. The previous
code errored out if this happened more than five times, which is not wrong
and not unlikely to happen. The limit doesn't really have a purpose anyway,
so just remove it.

Related ticket: https://progress.opensuse.org/issues/107710

Verification run: http://10.160.67.86/tests/1201

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -96,17 +96,13 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
 
         # wait for next read
         my @res = $s->can_read;
-        my $remaining_attempts = 5;
         while (!@res) {
             # throw an error except can_read has been interrupted
             my $error = $!;
             confess "ERROR: unable to wait for JSON reply: $error\n" unless $!{EINTR};
-            confess "ERROR: can_read's underlying system call has been interrupted too many times\n" unless $remaining_attempts;
-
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
Can you please explain in more details "The limit doesn't really have a purpose anyway,
so just remove it."? It's likely the original author had some idea in mind why they added it, right?

Vogtinator:
> Can you please explain in more details "The limit doesn't really have a purpose anyway, so just remove it."? It's likely the original author had some idea in mind why they added it, right?

I don't have more details than that.

> [13:21] <Martchus_> I don't remember why I only allowed 5 attempts. Maybe I didn't want to change it only as much as necassary.

okurz:
> > Can you please explain in more details "The limit doesn't really have a purpose anyway, so just remove it."? It's likely the original author had some idea in mind why they added it, right?
> 
> I don't have more details than that.
> 
> > [13:21] <Martchus_> I don't remember why I only allowed 5 attempts. Maybe I didn't want to change it only as much as necassary.

That reads as: We can increase the tries, e.g. to 100, but not remove it.

Vogtinator:
The limit can definitely be removed, it serves no purpose other than induce random failures. FWICT, @Martchus agrees with that.

okurz (approved):

PR #1992, opened 2022-03-15
Increase default VNC timeout for remote connections

Martchus:
* Make timeout configurable as well
* See https://progress.opensuse.org/issues/76813

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -118,7 +118,6 @@ my @encodings = (
 );
 
 sub login ($self, $connect_timeout = undef) {
-    $connect_timeout //= 10;
     # arbitrary
     my $connect_failure_limit = 2;
 
@@ -135,7 +134,9 @@ sub login ($self, $connect_timeout = undef) {
     my $hostname = $self->hostname || 'localhost';
     my $port = $self->port || 5900;
     my $description = $self->description || 'VNC server';
-
+    my $local_timeout = $bmwqemu::vars{VNC_CONNECT_TIMEOUT_LOCAL} // 10;
+    my $remote_timeout = $bmwqemu::vars{VNC_CONNECT_TIMEOUT_REMOTE} // 60;
+    $connect_timeout //= ($hostname =~ qr/(localhost|127\.0\.0\.\d+|::1)/) ? $local_timeout : $remote_timeout;
     my $endtime = time + $connect_timeout;
 
     my $socket;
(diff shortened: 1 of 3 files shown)


t/27-consoles-vnc.t
```diff
@@ -50,4 +52,23 @@ subtest 'handling VNC stall, malformed RFB protocol on re-connect' => sub {
...
+    _setup_rfb_magic;
+    combined_like {
+        throws_ok { $c->login } 'OpenQA::Exception::VNCSetupError', 'dies on connect timeout' }
+    qr/.*Error connecting to.*/, 'error logged';
+    is $attempts, 7, 'login attempts for local hostname';
+    $attempts = 0;
+    $c->hostname('10.161.145.95');
+    combined_like {
+        throws_ok { $c->login } 'OpenQA::Exception::VNCSetupError', 'dies on connect timeout (2)'
+    } qr/.*Error connecting to.*/, 'error logged (2)';
```
okurz, line 70: Is this trying to actually connect to this adress? I would prefer to mock the real connection attempt or at least use a non-reachable localhost address. What about 127.0.0.2?

consoles/VNC.pm
```diff
@@ -135,7 +134,9 @@ sub login ($self, $connect_timeout = undef) {
     my $hostname = $self->hostname || 'localhost';
     my $port = $self->port || 5900;
     my $description = $self->description || 'VNC server';
-
+    my $local_timeout = $bmwqemu::vars{VNC_CONNECT_TIMEOUT_LOCAL} // 10;
+    my $remote_timeout = $bmwqemu::vars{VNC_CONNECT_TIMEOUT_REMOTE} // 60;
+    $connect_timeout //= ($hostname =~ qr/(localhost|127\.0\.0\.\d+|::1)/) ? $local_timeout : $remote_timeout;
```
Martchus, line 139: @kraih Does Perl or some module like Mojolicious provide a helper for this distinction? (e.g. something in the lines of https://github.com/Martchus/syncthingtray/blob/8e6bd62030db1867fd68d3798a7c1e497933bae5/connector/utils.cpp#L118)

t/27-consoles-vnc.t
okurz: ok, cool

kalikiana (approved):
Looks sensible enough to me!

consoles/VNC.pm
kraih: This should probably be a utility function with tests.

okurz (approved):

PR #1993, opened 2022-03-15
Avoid warnings about redefinition of `has`

Martchus:
* Use `Mojo::Base` everywhere, remove use of `Class::Accessor`
* See https://progress.opensuse.org/issues/108323


perlpunk (approved):

backend/baseclass.pm
```diff
@@ -5,7 +5,7 @@
 # this is an abstract class
 package backend::baseclass;
 
-use Mojo::Base -strict, -signatures;
+use Mojo::Base -base, -strict, -signatures;
```
kraih, line 8: `-base` and `-strict` are redundant. You can just remove the `-strict`;

kraih (approved):

PR #1994, opened 2022-03-16
Fix serialization error introduced by 7be44a78

Martchus:
* Ensure `disable_vnc_stalls` returns something serializable
* See https://progress.opensuse.org/issues/108452

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -25,6 +25,7 @@ sub get_last_mouse_set ($self, $args) { $self->{mouse} }
 sub disable_vnc_stalls ($self) {
     return unless $self->{vnc};
     $self->{vnc}->check_vnc_stalls(0);
+    return 0;
 }
 
 sub connect_remote ($self, $args) {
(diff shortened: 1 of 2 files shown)


okurz (approved):
looks ok. Hotpatched on OSD openqaworker2. I don't know how 7be44a78 should have introduced this though

Martchus:
> I don't know how https://github.com/os-autoinst/os-autoinst/commit/7be44a786c05c215552b15e956292e682ed9b47e should have introduced this though

It is not obvious. The difference between `Class::Accessor` (1) and `Mojo::Base` (2) is that accessors of (1) return the new value back when setting a value where (2) returns the object itself in that case. I've actually checked all uses but I haven't expected that the return value is actually used here. I suppose in Perl we have to consider all return values as used if they occur on the last statement of a function.

okurz:
>     * [x]  Consider running manual verification tests on real hardware, especially for non-qemu backends

Yep, we did that on openqaworker2

PR #1996, opened 2022-03-19
Fix return value of set_assert_screen_timeout

perlpunk:
Mojo accessors return the object on set.

This should fix https://github.com/os-autoinst/openQA/pull/4569


Martchus (approved):

Martchus:
Thanks!

PR #1997, opened 2022-03-22
Fix qemu test to actually wait for the process to be stopped

kraih:
Just calling ->stop does not guarantee that the forked process stops in
time for the cleanup code to create the state file for the test code. A
switch to ->wait_stop should make the test more reliable. If that does
not resolve the issue completely, we might have to hook into an event to
manually wait for the process to stop.

Progress: https://progress.opensuse.org/issues/108443

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -412,8 +412,9 @@ subtest 'qemu was killed due to the system being out of memory' => sub {
     $mock_proc->redefine(check_qemu_oom => sub { return 0; });
     $proc = qemu_proc('-foo', \%vars);
     $proc->exec_qemu;
-    $proc->_process->stop;
+    $proc->_process->sleeptime_during_kill(0.1)->wait_stop;
     my $base_state = path(bmwqemu::STATE_FILE);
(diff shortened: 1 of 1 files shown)


t/18-qemu.t
```diff
@@ -412,8 +412,9 @@ subtest 'qemu was killed due to the system being out of memory' => sub {
     $mock_proc->redefine(check_qemu_oom => sub { return 0; });
     $proc = qemu_proc('-foo', \%vars);
     $proc->exec_qemu;
-    $proc->_process->stop;
+    $proc->_process->wait_stop;
```
Martchus, line 415: This is basically a while loop with `sleep 1` where `1` comes from `$proc`s `sleeptime_during_kill` attribute. Maybe it makes sense to reduce `sleeptime_during_kill` (e.g. to `0.1`).

Martchus (commented):
Looks good, only one suggestion.

t/18-qemu.t
kraih: Seems reasonable.

okurz (approved):

Martchus (approved):

PR #2001, opened 2022-03-24
Fix a bug where the collected event was registered too late

kraih:
Not actually a fix for the issue i'm working on, but a small race condition i found along the way.

Progress: https://progress.opensuse.org/issues/107998

### DIFF
--- a/backend/driver.pm
+++ b/backend/driver.pm
@@ -8,8 +8,8 @@
 # in that 2nd process runs the actual backend, derived from backend::baseclass
 
 package backend::driver;
+use Mojo::Base -base, -signatures;
 
-use Mojo::Base -strict, -signatures;
 use autodie ':all';
 
 use Carp 'croak';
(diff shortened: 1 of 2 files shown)


backend/driver.pm
```diff
@@ -52,8 +53,7 @@ sub start ($self) {
         blocking_stop => 1,
         separate_err => 0,
         subreaper => 1,
-        code => sub {
-            my $process = shift;
+        code => sub ($process) {
```
Martchus, line 56: Judging by RWP code additional arguments might be passed to the callback:
```suggestion
        code => sub ($process, @args) {
```

Martchus (commented):
Looks good, except for the one remark mentioned in the inline comment.

Martchus (approved):

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst/runs/5681532735?check_suite_focus=true#step:3:594 fails in 18-qemu-options.t with an error message I don't really understand.

kraih:
I've restarted the tests a few times, but `18-qemu-options.t` fails consistently in CI. Locally i'm not seeing the problem. Probably a race condition that becomes more common with the patch. I'll investigate and look for a fix.

Martchus:
Strange. Maybe the number of arguments is diverting in certain cases?

kraih:
Maybe i'll try to ssh into the GitHub Actions VM on Monday to track it down further.

okurz (approved):

PR #2002, opened 2022-03-28
Hide secrets in all log_call invocations 

b10n1k:
My first idea was to encrypt the text and then decrypt it when it is needed for the call.

However the concern was about the logs and i found that the only thing that it needs to be done
is to update the log_call, because this is what is called from all the backends. in this way nothing
else seems to need to change in testapi or anywhere else.

http://aquarius.suse.cz/tests/9012/logfile?filename=autoinst-log.txt

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -215,11 +215,13 @@ sub log_call {
     else {
         # key/value pairs
         my @result;
+        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
         while (my ($key, $value) = splice(@_, 0, 2)) {
             if ($key =~ tr/0-9a-zA-Z_//c) {
                 # only quote if needed
                 $key = pp($key);
             }
+            $value = "[masked]" if ($key eq 'text' && $should_hide);
             push @result, join("=", $key, pp($value));
         }
         $params = join(", ", @result);
--- a/t/12-bmwqemu.t
+++ b/t/12-bmwqemu.t
@@ -60,6 +60,11 @@ subtest 'log_call' => sub {
         bmwqemu::log_call(test => $lines);
     }
     stderr_like(\&log_call_indent, qr{\Q<<< main::log_call_indent(test=[\E\n\Q    "a",\E\n\Q    [\E\n\Q      "b"\E\n\Q    ]\E\n\Q  ])}, 'log_call auto indentation');
+
+    sub log_call_test_secret {
+        bmwqemu::log_call(text => "passwd\n", secret => 1);
+    }
+    stderr_like(\&log_call_test_secret, qr{\Q<<< main::log_call_test_secret(text="[masked]", secret=1)}, 'log_call hides sensitive info');
 };
 
 subtest 'update_line_number' => sub {
--- a/testapi.pm
+++ b/testapi.pm
@@ -1465,7 +1465,6 @@ sub type_string {
     $string .= "\n" if $args{lf};
 
     if (is_serial_terminal) {
-        bmwqemu::log_call(text => $log, %args);
         query_isotovideo('backend_type_string', {text => $string, %args});
         return;
     }
--- a/testapi.pm
+++ b/testapi.pm
@@ -1461,7 +1461,6 @@ sub type_string {
     else {
         %args = @_;
     }
-    my $log = $args{secret} ? 'SECRET STRING' : $string;
     $string .= "\n" if $args{lf};
 
     if (is_serial_terminal) {
@@ -1474,7 +1473,7 @@ sub type_string {
     my $wait_still = $args{wait_still_screen} // 0;
     my $wait_timeout = $args{timeout} // 30;
     my $wait_sim_level = $args{similarity_level} // 47;
-    bmwqemu::log_call(string => $log, max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still,
+    bmwqemu::log_call(string => $string, max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still,
         timeout => $wait_timeout, similarity_level => $wait_sim_level);
     my @pieces;
     if ($wait) {


bmwqemu.pm
```diff
@@ -215,11 +215,17 @@ sub log_call {
     else {
         # key/value pairs
         my @result;
+        my $should_stash = 0;
+        my @tmp_array = @_;
+        while (my ($key, $value) = splice(@tmp_array, 0, 2)) {
```
Martchus, line 220: I wouldn't bet on this syntax doing what you want so please provide some unit tests showing it does the right thing. Likely there's also a way to do this without constructing and destructing a temporary array.

bmwqemu.pm
```diff
@@ -215,11 +215,17 @@ sub log_call {
...
+            $should_stash = 1 if ($key eq 'secret' && $value == 1);
```
Martchus, line 221: Isn't it sufficient if `$value` is a truthy value? The check `$value == 1` has the disadvantage of emitting a warning if `$value` isn't convertible to a number.

Martchus:
Maybe "hide" would be a better word than "stash" here?

bmwqemu.pm
```diff
@@ -215,11 +215,17 @@ sub log_call {
...
+        }
         while (my ($key, $value) = splice(@_, 0, 2)) {
             if ($key =~ tr/0-9a-zA-Z_//c) {
                 # only quote if needed
                 $key = pp($key);
             }
+            $value = "*****" if ($key eq 'text' && $should_stash);
```
okurz, line 228: For now I would stay consistent with the previous string "SECRET STRING"

okurz (changes requested):

bmwqemu.pm
b10n1k: I think it is more nice like that. and it implies directly a _secret_ just looking at it. Would you mind to keep it like this?

Martchus: I think using stars or some other placeholder is nicer, indeed. It would make sense to split that change into an additional commit, though.

b10n1k: no problem with that.

bmwqemu.pm
okurz: at this point, yes. We are not doing parameter input validation, IMHO not the right place for that.

bmwqemu.pm
okurz: if we want to change it, how about "(masked)" instead?

bmwqemu.pm
asmorodskyi, line 220: I am not sure why you need extra loop for this , why simply not do same things in loop which we already have which started in line 223 ?

bmwqemu.pm
Martchus: or [masked]

asmorodskyi:
LGTM , also I am closing my version which was trying to achieve the same . Nice work Yannis !

b10n1k:
> LGTM , also I am closing my version which was trying to achieve the same . Nice work Yannis !

I removed the `log_call` from testapi::type_string and i have the feeling that there are more of those around which could be removed and cleanup the os-autoinst.txt. just keep this in mind.

bmwqemu.pm
```diff
@@ -215,11 +215,14 @@ sub log_call {
     else {
         # key/value pairs
         my @result;
+        my $should_hide = 0;
+        $should_hide = 1 if (grep { defined $_ && $_ eq 'secret' } @_);
```
okurz, line 219:
```suggestion
        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
```

okurz (changes requested):
Better rephrase the "Hide" commit message as "Hide secrets in all log_call invocations"
and please wrap your git commit messages consistently at 80 characters per line.

b10n1k:
@okurz @Martchus @asmorodskyi changes are made. But i cant comment on the conversations anymore, for some reason. I guess also that the recommendation from the bot is not needed, right?

bmwqemu.pm
```diff
@@ -215,11 +215,13 @@ sub log_call {
     else {
         # key/value pairs
         my @result;
+        my $should_hide = 1 if (grep { ($_ // '') eq 'secret' } @_);
```
okurz, line 218: No

```suggestion
        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
```

okurz (changes requested):
I am not sure what your intention with the tests is but it should be obvious that we won't merge your PR as long as you have disabled a big chunk of tests with comments :)

okurz:
> I guess also that the recommendation from the bot is not needed, right?

Well, if the only real change is that the string output in log messages changes then the API version does not need to be bumped.

okurz:
@b10n1k please only push updates if you actually address the review comments. I was getting notified multiple times but have seen the changes not implemented.

t/12-bmwqemu.t
```diff
@@ -60,6 +60,11 @@ subtest 'log_call' => sub {
         bmwqemu::log_call(test => $lines);
     }
     stderr_like(\&log_call_indent, qr{\Q<<< main::log_call_indent(test=[\E\n\Q    "a",\E\n\Q    [\E\n\Q      "b"\E\n\Q    ]\E\n\Q  ])}, 'log_call auto indentation');
+
+    sub log_call_test_secret {
+        bmwqemu::log_call(text => "passwd\n", secret => 1);
+    }
+    stderr_like(\&log_call_test_secret, qr{\Q<<< main::log_call_test_secret(text="[masked]", secret=1)}, 'log_call hides sensitive info');
```
Martchus, line 67:
```suggestion
    sub log_call_test_secret {
        bmwqemu::log_call(text => "passwd\n", secret => 1);
    }
    stderr_like { bmwqemu::log_call(text => "passwd\n", secret => 1) } qr{\Q<<< main::log_call_test_secret(text="[masked]", secret=1)}, 'log_call hides sensitive info';
```

bmwqemu.pm
```diff
@@ -215,11 +215,13 @@ sub log_call {
     else {
         # key/value pairs
         my @result;
+        my $should_hide = 1 if grep { ($_ // '') eq 'secret' } @_;
```
Martchus, line 218: Declaring a variable with a nested `if` is a bit weird in my opinion. I'd write it simply as:
```suggestion
        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
```

Martchus (approved):

okurz (approved):

jlausuch:
This breaks the use of `type_password`. 

With this change,  `type_password()` in testapi.pm passes the argument `secret` to `type_string` (see below code), but now `type_string()` is ALWAYS printing `$string` to the log, whatever value it is (if it's secret or not). So, all the strings commands made with `type_password` are logged in autoinst-log.txt

Maybe we should recover `my $log = $args{secret} ? 'SECRET STRING' : $string;`

I created this ticket https://progress.opensuse.org/issues/111010


```
sub type_password {
    my ($string, %args) = @_;
    $string //= $password;
    type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
}
```

PR #2004, opened 2022-03-28
isotovideo: Use wheels specified in wheels.yaml if present

kalikiana:


okurz (commented):
Looks easy enough. Nice.

isotovideo
```diff
@@ -242,6 +243,13 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
 # set a default distribution if the tests don't have one
 $testapi::distri = distribution->new;
 
+if (my $specfile = file('plugins.yaml')->slurp) {
+    my $spec = YAML::PP->new->load_file($specfile);
```
perlpunk, line 247: either use the filename and load_file, or the file content and load

container/os-autoinst_dev/Dockerfile
```diff
@@ -127,6 +127,7 @@ RUN zypper in -y -C \
        'perl(XML::LibXML)' \
        'perl(XML::SemanticDiff)' \
        'perl(YAML::PP)' \
+       'perl(YAML::PP)' \
```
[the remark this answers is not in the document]
perlpunk: Might be a bug in the update-deps script. It should filter duplicate modules from different sections.

isotovideo
```diff
@@ -242,6 +243,13 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
...
+    foreach my $plugin ($spec->{plugins}) {
+        checkout_git_repo_and_branch($plugin);
+    }
+}
+
```
okurz, line 252: so maybe

```suggestion
checkout_git_repo_and_branch($_) for YAML::PP->new->load_file('plugins.yaml')->{plugins} if -e 'plugins.yaml';
```

okurz (changes requested):

isotovideo
Martchus: I suppose in a non-draft version we maybe want error handling here anyways, e.g. some meaningful error message if plugins doesn't exist or is no array would be useful for the user (instead of just some Perl warning).

okurz: You mean if the file exists but is ill-formated?

Martchus: Yes. Maybe it can be validated against a schema (so we don't have to care about every error case ourselves).

Martchus:
Looks generally good (besides issues mentioned by others). I'm not sure about the naming. If we'd just call it "plugins" everywhere that would make more sense to me (compared to using "wheels" in some places which is not consistent and also not as intuitive).

kraih (dismissed):

container/os-autoinst_dev/Dockerfile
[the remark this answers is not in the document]
kalikiana: I filed [poo#109313](https://progress.opensuse.org/issues/109313)

t/14-isotovideo.t
```diff
@@ -105,6 +105,19 @@ subtest 'isotovideo with git refspec specified' => sub {
...
+subtest 'isotovideo with wheels' => sub {
+    chdir($pool_dir);
+    unlink('vars.json') if -e 'vars.json';
+    my $specfile = path("$data_dir/tests/wheels.yaml");
+    $specfile->spurt('wheels: [foo/bar]');
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Cloning.*foo/bar@, 'single plugin picked up';
+    $specfile->spurt('wheels: [foo/bar, spam/eggs]');
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Cloning.*spam/eggs@, 'single plugin picked up';
```
okurz, line 117: please use a distinct, different test description here, e.g. 'a different plugin picked up'

schema/Wheels-01.yaml
```diff
@@ -0,0 +1,14 @@
+$id: http://open.qa/api/schema/Wheels-01.yaml
+$schema: http://json-schema.org/draft-04/schema#
+description: Spec for loadable modules
```
okurz, line 3: modules? Shouldn't that be "wheels" or "plugins" to not introduce yet another word for the same?

container/os-autoinst_dev/Dockerfile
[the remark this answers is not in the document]
perlpunk: if you rebase now the duplicate should be gone

dependencies.yaml
```diff
@@ -169,6 +169,9 @@ main_requires:
   perl(Try::Tiny):
   perl(XML::LibXML):
   perl(XML::SemanticDiff):
+  perl(YAML::PP):
```
okurz, line 172: this duplicates that previously we already required it from https://github.com/os-autoinst/os-autoinst/pull/2004/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6R109 . I suggest to remove https://github.com/os-autoinst/os-autoinst/pull/2004/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6R109 and keep this more low-level dependency

perlpunk: The links don't work for me, but I guess the suggestion is to remove YAML::PP from test_requires, now that it is added to main_requires.

okurz: maybe now you need to manually remove the double `perl(YAML::PP)` from Dockerfile and cpanfile and spec file

t/14-isotovideo.t
```diff
@@ -105,6 +105,33 @@ subtest 'isotovideo with git refspec specified' => sub {
...
+    $specfile->spurt("version: v0.1\nwheels: [https://github.com/foo/bar.git#branch]");
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Cloning.*XXXfoo/bar@, 'single repo with branch';
+    $specfile->spurt("version: v0.1\nwheels: [foo/bar]");
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Cloning.*foo/bar@, 'single repo';
+    $specfile->spurt("version: v0.1\nwheels: [foo/bar, spam/eggs]");
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Cloning.*spam/eggs@, 'two repos';
+    $specfile->remove;
```
okurz, line 132: How much does this increase the runtime? Please compare numbers from before and afterwards. The goal should be to have no significant increase and prefer to reduce test runtimes where possible to reach tests that run well below 1s.

okurz (changes requested):

t/14-isotovideo.t
kalikiana: I don't know. Applying TDD means I'm adding tests first to be sure things will work and this is how existing tests were written.

okurz: I don't really understand what you mean by that. I suggest you call `prove --timer -I. t/14-isotovideo.t` with and without your changes (e.g. after `git stash` to stash your changes) and compare the runtime.

t/14-isotovideo.t
```diff
@@ -105,6 +106,35 @@ subtest 'isotovideo with git refspec specified' => sub {
             opts => "casedir=$data_dir/tests test_git_refspec=deadbeef _exit_after_schedule=1") } qr/Checking.*local.*deadbeef/, 'refspec in local git repository would be checked out';
 };
 
+subtest 'isotovideo with wheels' => sub {
+    chdir($pool_dir);
+    unlink('vars.json') if -e 'vars.json';
+    my $specfile = path("$data_dir/tests/wheels.yaml");
+
+    my $system_mock = Test::MockModule->new('IPC::System::Simple');
+    $system_mock->redefine(capture => sub { return; });
```
kalikiana, line 115: Apparently this doesn't work. `git` still seems to be called. I don't understand it...

kalikiana: Marius suggested it doesn't work because it's not a qualified function on a blessed object (c.f. [Mojo::File / slurp](https://github.com/os-autoinst/os-autoinst/blob/master/t/10-virtio_terminal.t#L74))

t/14-isotovideo.t
kalikiana: I mean that I added *new* tests and they're not even passing. How would you reduce test runtime in that context? Are you seeing any obvious issues with how I'm testing this?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -92,7 +94,8 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
     else {
         bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists";
     }
-    return $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string;
+    return $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string unless $args{repo};
+    return undef;
```
okurz, line 98: I suggest to make it:

```suggestion
    return $args{repo} ? undef : $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string;
```

This way we use the positive if-condition and make it obvious that we always want to return something valid but distinguish based on an option.

t/14-isotovideo.t
okurz: I don't see any obvious issues yet. Of course you can focus first on getting tests to work. I am just pointing out what we also should look out for. Currently if I call `prove -I. t/14-isotovideo.t` locally it takes 13s and we should be careful to not extend that (significantly).

t/14-isotovideo.t
kalikiana: I've decided to drop the tests that would require mocking and tackle that in *pr coming up, please refresh later* instead

t/14-isotovideo.t
```diff
@@ -105,6 +105,20 @@ subtest 'isotovideo with git refspec specified' => sub {
...
+    unlink('vars.json') if -e 'vars.json';
+
+    my $specfile = path("$data_dir/tests/wheels.yaml");
+    $specfile->spurt("wheels: [foo/bar]");
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Invalid.*Missing property@, 'invalid YAML causes error';
+    $specfile->spurt("version: v99\nwheels: [foo/bar]");
+    combined_like { isotovideo(
+            opts => "casedir=$data_dir/tests _exit_after_schedule=1") } qr@Unsupported version@, 'unsupported version';
+    $specfile->remove;
```
perlpunk, line 119: Please use the END block at the end of the test as we do in all our tests to make sure files are also removed in case the test exits unexpectedly.

kalikiana: I removed it because it is executed too late in the successful case

kalikiana: >  removed it because it is executed too late in the successful case
By too late I mean that other tests will pickup the file. Although we could have both I guess 🤔

okurz (approved):
I would prefer to move the new functionality into a different file, especially as you are loading new dependencies. However I can also look into doing that myself as a follow-up outside this PR.

isotovideo
```diff
@@ -246,6 +250,25 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
 # set a default distribution if the tests don't have one
 $testapi::distri = distribution->new;
 
+my $specfile = path($bmwqemu::vars{PRODUCTDIR} . '/wheels.yaml');
+if (-e $specfile) {
+    my $schema_file = dirname(__FILE__) . '/schema/Wheels-01.yaml';
```
kraih, line 255: Since you're already using `Mojo::File`, it can also export a `curfile` function that could be used here (`curfile->dirname`).

kalikiana: I actually changed that to `$installprefix` since the file is somewhere else when it's installed.

kalikiana:
I pulled in all of the tests from #2030 again. Compared to before all but one test are unit tests which rely on `checkout_git_repo_and_branch` being mocked. Instead of the output the expected repositories are verified. The wheel code lives in Utils now. One additional unit test checks that no wheels don't result in an error.

isotovideo
```diff
@@ -245,6 +246,7 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
 # set a default distribution if the tests don't have one
 $testapi::distri = distribution->new;
 
+checkout_wheels($bmwqemu::vars{PRODUCTDIR});
```
okurz, line 249: why productdir? In case of os-autoinst-distri-opensuse that would be something like "products/opensuse/" or "products/sle", shouldn't that rather be "CASEDIR"? Or do you explicitly want to allow loading product-specific plugins? Also how about defining a variable that defaults to PRODUCTDIR if not set?:

```suggestion
checkout_wheels($bmwqemu::vars{WHEELS_DIR} // $bmwqemu::vars{PRODUCTDIR});
```

okurz (changes requested):
Please also shortly describe the idea and usage in the README file

isotovideo
kalikiana: I thought it might be intuitive since PRODUCTDIR defaults to CASEDIR. And having a product-specific wheel configuration will eventually be useful.
WHEELS_DIR is a good idea actually since defaults won't work in all cases anyway.

isotovideo
```diff
@@ -245,6 +246,8 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
 # set a default distribution if the tests don't have one
 $testapi::distri = distribution->new;
 
+$bmwqemu::vars{WHEELS_DIR} ||= $bmwqemu::vars{PRODUCTDIR};
```
okurz, line 249: I would still prefer to only fall back to CASEDIR. I assume wheels will mostly be used by lib/ or test modules in test distributions do not specific per product. So I see no sense in using product dir.

kalikiana: That's why I left the thread unresolved 😄

Maybe @punkioudi you have some thoughts here as a future user? Where would you intuitively put the wheels.yaml?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -98,7 +103,39 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
     else {
         bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists";
     }
-    return $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string;
+    return $args{repo} ? undef : $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string;
```
okurz, line 106:
```perl
return undef if $args{repo};
return $bmwqemu::vars{$dir_variable} = path($local_path)->to_abs->to_string;
```

okurz (changes requested):

okurz (approved):

OpenQA/Isotovideo/Utils.pm
```diff
@@ -98,9 +103,42 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
...
+    checkout_wheels($dir);
+
+Takes a directory which may require wheels and checks out the
+referenced git repository into a local working copy with an additional,
+optional git refspec to checkout.
+If no wheels are configured the function returns early.
+
+=cut
+sub checkout_wheels ($dir) {
+    my $specfile = path($dir . '/wheels.yaml');
```
Martchus, line 121:
```suggestion
    my $specfile = path($dir, 'wheels.yaml');
```

autotest.pm
```diff
@@ -32,6 +32,11 @@ OS Autoinst decides which test modules to run based on a distribution specific
 script called main.pm. This is either located in $vars{PRODUCTDIR} or
 $vars{CASEDIR} (e.g. <distribution>/products/<product>/main.pm).
 
+Wheels can be used to add functionality from other repositories. If a file
+wheels.yaml is present in $vars{PRODUCTDIR} (or $vars{CASEDIR} if unset) the
```
Martchus, line 36: It doesn't look like `PRODUCTDIR` is used anymore in the current state of the PR.

autotest.pm
```diff
@@ -32,6 +32,11 @@ OS Autoinst decides which test modules to run based on a distribution specific
...
+specified git repositories are made available before tests are run. It also
```
Martchus, line 37:
```suggestion
specified git repositories are made available before tests are run. It is also
```

autotest.pm
```diff
@@ -32,6 +32,11 @@ OS Autoinst decides which test modules to run based on a distribution specific
...
+possible to override $vars{WHEELS_DIR} explicitly.
```
Martchus, line 38:
```suggestion
possible to set $vars{WHEELS_DIR} explicitly.
```

OpenQA/Isotovideo/Utils.pm
```diff
@@ -98,9 +103,42 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
...
+    die "Unable to load schema '$schema_file'" unless -f $schema_file;
+    my $validator = JSON::Validator->new;
+    $validator->schema($schema_file);
+    my $spec = YAML::PP->new->load_file($specfile);
+    my @errors = $validator->validate($spec);
+    die 'Invalid YAML: ' . join(',', @errors) if @errors;
+    die 'Unsupported version: Version must be "v0.1"' if $spec->{version} ne 'v0.1';
+
+    foreach my $repo (@{$spec->{wheels}}) {
+        $repo = "https://github.com/$repo.git" unless $repo =~ '^http';
```
Martchus, line 136:
```suggestion
        $repo = "https://github.com/$repo.git" unless $repo =~ qr/^http/;
```

Martchus (changes requested):

PR #2005, opened 2022-03-29
Avoid JSON::PP::Boolean warnings

perlpunk:
When Cpanel::JSON::XS is loaded before JSON::PP, it will
result in warnings

    Subroutine JSON::PP::Boolean::(0+ redefined at .../overload.pm
    ...

This is because Cpanel::JSON::XS creates the same overloads
as JSON::PP. This is handled in newer versions of JSON::PP, so
it will go away at some point, but until then it is always confusing to see
the warnings.
Adding `-mJSON::PP` makes sure it will be loaded first.

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -39,6 +39,10 @@ use Mojo::Base -strict, -signatures;
 use autodie ':all';
 no autodie 'kill';
 
+# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
+# Details: https://progress.opensuse.org/issues/90371
+use JSON::PP;
+
 my $installprefix;    # $bmwqemu::scriptdir
 my $fatal_error;    # the last error message caught by the die handler
 


okurz (approved):

okurz:
Approved although I like your proposal to load in isotovideo (with a comment) better

Martchus (approved):
Considering this is supposedly only affecting tests I assume this fix is good enough.

isotovideo
```diff
@@ -39,6 +39,21 @@ use Mojo::Base -strict, -signatures;
...
+# result in warnings
+#
+#     Subroutine JSON::PP::Boolean::(0+ redefined at .../overload.pm
+#     ...
+#
+# This is because Cpanel::JSON::XS creates the same overloads
+# as JSON::PP. This is handled in newer versions of JSON::PP, so
+# it will go away at some point, but until then it is always confusing to see
+# the warnings.
+# Loading it here first will avoid the warning.
```
okurz, line 54: I suggest to make the comment at most two-line, e.g. the subject "Avoid…" and a reference to any bug or ticket reference for details.

okurz (changes requested):

okurz (approved):

PR #2006, opened 2022-03-29
Use proper perl "signatures" in common files - needle

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.


needle.pm
```diff
@@ -141,17 +136,15 @@ sub save {
     close $fh;
 }
 
-sub unregister {
-    my ($self, $reason) = @_;
+sub unregister ($self, $reason = 'unknown reason') {
```
Martchus, line 139: As mentioned many times before, setting a default here is not equivalent to the previous code. Note that this function is invoked by the test distribution.

okurz: I will never remember that, sorry

okurz: like this with `$reason = undef` and keeping the old `… || 'unknown reason'`?

backend/baseclass.pm
```diff
@@ -748,7 +748,7 @@ sub reload_needles (@) {
     # called from testapi::set_var, so read the vars
     bmwqemu::load_vars();
 
-    $_->unregister() for needle->all();
+    $_->unregister() for needle::all();
```
kalikiana, line 751: This looks quite strange to me, considering all you did was add `()`. But it seems to work I guess

kalikiana (approved):

backend/baseclass.pm
okurz: The syntax `needle->all()` kinda worked but was never intended as `needle::all` was a free function, not an object method

needle.pm
```diff
@@ -281,23 +260,20 @@ sub get_property_value {
...
     my $needle = needle->new($File::Find::name);
     $needles{$needle->{name}} = $needle if $needle;
 }
 
-sub default_needles_dir { "$bmwqemu::vars{PRODUCTDIR}/needles" }
+sub default_needles_dir () { "$bmwqemu::vars{PRODUCTDIR}/needles" }
 
-sub init {
-    my $init_needles_dir = ($bmwqemu::vars{NEEDLES_DIR} // default_needles_dir);
+sub init ($init_needles_dir = $bmwqemu::vars{NEEDLES_DIR} // default_needles_dir) {
```
perlpunk, line 276: So it's now possible to pass `init` an argument? Was that intended?

okurz: hm, honestly by now I don't remember anymore. But I think it's ok if the variable can be initialized from outside

PR #2007, opened 2022-03-29
Do not require additional command to pause on a test module

Martchus:
Make developer mode pause directly when reaching a test module to pause
on (instead of only pausing on the next command within that test module).


OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -250,7 +250,7 @@ sub _handle_command_set_current_test ($self, $response, @) {
         $self->reason_for_pause('reached module ' . $pause_test_name);
     }
     $self->update_status_file;
-    $self->_respond_ok();
+    $self->reason_for_pause ? $self->postponed_answer_fd($self->answer_fd)->postponed_command(undef) : $self->_respond_ok;
```
okurz, line 253: I think this is easier understood if you move it into a method with a good name

okurz (approved):

PR #2008, opened 2022-03-29
Add explicit test for `set_pipe_sz()` to ensure stable test coverage

Martchus:
See https://progress.opensuse.org/issues/105759


kraih (approved):

kraih:
Theoretically this shouldn't work, because the caller also had flaky coverage before. But the latest coverage results would suggest that it does work anyway... so 🤷.

Martchus:
@okurz I'll try that command tomorrow. But if it is still not stable I'm out of ideas anyways.

PR #2010, opened 2022-03-30
t: Speedup 18-qemu-options.t

okurz:
* t: Reduce runtime of 18-qemu-options.t by doing less QEMU_APPEND checks
* t: Further reduce runtime of 18-qemu-options.t by moving tpm tests
* t: Reduce runtime of 18-qemu-options.t by moving hugepages options test
* t: Remove more unused test variables in 18-qemu-options.t
* t: Speedup 18-qemu-options.t a bit

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -519,7 +519,7 @@ sub setup_tpm ($self, $arch) {
     my $vars = \%bmwqemu::vars;
     return unless ($vars->{QEMUTPM});
     my $tpmn = $vars->{QEMUTPM} eq 'instance' ? $vars->{WORKER_INSTANCE} : $vars->{QEMUTPM};
-    my $vmpath = "/tmp/mytpm$tpmn";
+    my $vmpath = ($vars->{QEMUTPM_PATH_PREFIX} // '/tmp/mytpm') . $tpmn;
     mkdir $vmpath unless -d $vmpath;
     my $vmsock = "$vmpath/swtpm-sock";
     unless (-e $vmsock) {
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -139,8 +139,9 @@ QEMUMACHINE;see qemu -machine ?;undef;Machine and chipset to emulate
 QEMUPORT;integer;20002 + worker instance * 10;Port on which QEMU monitor should listen
 QEMURAM;integer;1024;Size of RAM of VM in MiB
 QEMUTHREADS;integer;0;Number of cpu threads used by VM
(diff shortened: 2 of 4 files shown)


Martchus (commented):
Looks generally good but tests don't pass.

okurz:
openQA Fullstack fails due to perl-Mojolicious which is likely fixed by https://github.com/os-autoinst/openQA/pull/4588 meanwhile. A new trigger should solve this. 18-backend-qemu.t now fails trying to execute `swtpm` which previously was done in 18-qemu-options.t . I have not seen this problem locally and the command execution should have been mocked already. Looking into that.

okurz:
I am confused about https://github.com/os-autoinst/os-autoinst/runs/5768272143?check_suite_focus=true#step%3A3%3A595=
so it says

```
: open3: exec of swtpm socket --tpmstate dir=/tmp/18-backend-qemu.t-AO0g/mytpm3 --ctrl type=unixio,path=/tmp/18-backend-qemu.t-AO0g/mytpm3/swtpm-sock --log level=20 -d --tpm2 failed: No such file or directory at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/IOLoop/ReadWriteProcess.pm line 127.
```

help?

okurz:
> I am confused about https://github.com/os-autoinst/os-autoinst/runs/5768272143?check_suite_focus=true#step%3A3%3A595= …

I think I have understood it. I only mocked the "runcmd" call for the cases when I assumed the qemu backend would call swtpm itself which fixed tests locally but I better move the runcmd mock to the top of the subtest so that all invocations are mocked accounting for swtpm not even existing in an environment. I tested that with removing my /usr/bin/swtpm and calling tests.

Martchus (approved):

okurz:
hm, backend/driver.pm was [100% covered in before](https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/backend/driver.pm), now it's not anymore, in particular in the lines https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/2010/changes#D2L35 . Likely the previous code coverage was due to us running qemu longer into the process so that it would run into that child collection message although it was not proper explicit coverage. I guess I will separate the PR and start with parts of the commits first.

okurz:
Going back even one step further as the coverage is still decreased.

okurz:
Ok, I now include all commits again, so back to the state that Martchus approved. I tested coverage collection locally with `rm -rf /tmp/mytpm* && while_inotifywait /bin/sh -c "rm -rf cover_db/ && cover -coverage stmt -test -make 'prove -I. --timer t/18-backend-qemu.t t/18-qemu-options.t; echo $?'"` and the interesting thing is that backend/driver.pm is actually never included when I tested origin/master and my changes so likely the backend/driver.pm coverage comes from other test modules anyway. And with the above command I receive a total coverage of 29.5% vs 28.9% in origin/master so already an improvement. With that I think the PR is still a net benefit.

PR #2017, opened 2022-04-05
[WIP] Add githook-perltidy as pre-commit git hook

b10n1k:
This PR adds `App::githook::perltidy` in the os-autoinst dependencies.
With this in place contributors will ensure that the coding style is
sustained and remains clean and tidy.

To install the script the only requirement is the existence of `.perltidyrc`
which this repo already has. For other cases take a look on
https://metacpan.org/dist/App-githook-perltidy/view/bin/githook-perltidy#githook-perltidy-pre-commit

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -176,6 +176,8 @@ accelerated virtualization support.
 If you want to contribute to this project, please clone and send
 pull requests via https://github.com/os-autoinst/os-autoinst.
 
+Before submission, is recommended to install the <<githooks, git-hooks>>.
+
 More information on the contribution can be found on
 http://os-autoinst.github.io/openQA/contact/, too.
 
@@ -313,6 +315,28 @@ Install files for packaging:
 ----
 make install DESTDIR=…
 ----
+
+=== Install git hooks
+[[githooks]]
+
+Git hook scripts are useful for identifying simple issues before submission to code review.
+Make sure that you have installed the recommended git-hooks on your system before contribution
+to the project.
+
+If `githook-perltidy` has been already installed on your system, check if the `.git/hooks/pre-commit`
+contains the following lines
+
+----
+$ which githook-perltidy
+/usr/bin/githook-perltidy
+$ cat .git/hooks/pre-commit
+#!/bin/sh
+/usr/bin/githook-perltidy pre-commit
+----
+
+Otherwise run `githook-perltidy install` and check again.
+In case you have enabled `pre-commit` hooks by your own use the `-f` flag
+or edit the file
 
 Further notes:
 


Martchus (commented):
Will the newly added dependency survive invoking the `update-deps` target? I'm also wondering why the dependency is needed. Can one who like this workflow not simply invoke the wanted tidy command directly? I'm also wondering whether it makes sense to recommend this workflow in README.asciidoc.

perlpunk:
Yeah, maybe it would make more sense to invoke `tools/tidy` in the hook, as that uses tidyall with caching.

b10n1k:
> Yeah, maybe it would make more sense to invoke `tools/tidy` in the hook, as that uses tidyall with caching.

Let me try this

Martchus:
I usually try to come up with the most simple setup and invoking tidy manually worked so far just fine for me. Not sure whether we should recommend newcomer to create a setup that is more complicated than it needs to be. This seems more like a power-user setup.

It would definitely annoying if committing is unnecessarily delayed or even blocked just because tidy fails, e.g. if I'm editing files which are not covered by tidy anyways or when I'm just creating a WIP/experimental commit I'm so far skipping it. If tidy doesn't work at all, e.g. because I currently have the wrong version installed, I'd still like to be able to commit code at all.

b10n1k:
> I usually try to come up with the most simple setup and invoking tidy manually worked so far just fine for me. Not sure whether we should recommend newcomer to create a setup that is more complicated than it needs to be. This seems more like a power-user setup.
> 
> It would definitely annoying if committing is unnecessarily delayed or even blocked just because tidy fails, e.g. if I'm editing files which are not covered by tidy anyways or when I'm just creating a WIP/experimental commit I'm so far skipping it. If tidy doesn't work at all, e.g. because I currently have the wrong version installed, I'd still like to be able to commit code at all.

i dont see the problem at all. The opposite. Basically what you are saying is "i want to be able to skip the checks". This is one extra argument in the `git` command.

b10n1k:
> Yeah, maybe it would make more sense to invoke `tools/tidy` in the hook, as that uses tidyall with caching.

So i edited the pre-commit and wrote `./tools/tidy -o`. With only this inside everything looks good expect one thing. I think the module was taking care of finishing the commit after the perltidy was run.

README.asciidoc
```diff
@@ -314,6 +316,28 @@ Install files for packaging:
...
 ----
 
+=== Install git hooks
+[[githooks]]
+
+Git hook scripts are useful for identifying simple issues before submission to code review.
+Make sure that you have installed the recommended git-hooks on your system before contribution
+to the project.
+
+If `githook-perltidy` has been already installed on your system, check if the `.git/hooks/pre-commit`
```
Martchus, line 326:
```suggestion
If `githook-perltidy` has been already installed on your system, check if `.git/hooks/pre-commit`
```
or just
```suggestion
`githook-perltidy` has been already installed on your system if `.git/hooks/pre-commit`
```

README.asciidoc
```diff
@@ -314,6 +316,28 @@ Install files for packaging:
...
+contains the following lines
```
Martchus, line 327:
```suggestion
contains the following lines:
```

README.asciidoc
```diff
@@ -314,6 +316,28 @@ Install files for packaging:
...
+----
+$ which githook-perltidy
+/usr/bin/githook-perltidy
+$ cat .git/hooks/pre-commit
+#!/bin/sh
+/usr/bin/githook-perltidy pre-commit
+----
+
+Otherwise run `githook-perltidy install` and check again.
+In case you have enabled `pre-commit` hooks by your own use the `-f` flag
```
Martchus, line 338:
```suggestion
In case you have enabled `pre-commit` hooks by your own, use the `-f` flag
```

README.asciidoc
```diff
@@ -314,6 +316,28 @@ Install files for packaging:
...
+or edit the file
```
Martchus, line 339:
```suggestion
or edit the file.
```
And which file again? Maybe it should be stated explicitly.

PR #2019, opened 2022-04-06
Use more signatures - cv/basetest/autotest

okurz:
* Use proper perl "signatures" in common files - autotest
* Use proper perl "signatures" in common files - basetest
* Use proper perl "signatures" in common files - cv

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -4,7 +4,7 @@
 
 package autotest;
 
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 
 use bmwqemu;
 use Exporter 'import';
@@ -39,8 +39,7 @@ loadtest is called.
 
 =cut
 
-sub find_script {
-    my ($script) = @_;
+sub find_script ($script) {
     my $casedir = $bmwqemu::vars{CASEDIR};
     my $script_override_path = join('/', $bmwqemu::vars{ASSETDIR} // '', 'other', $script);
     if (-f $script_override_path) {
@@ -86,8 +85,7 @@ e.g. by making use of the openQA asset download feature.
 
 =cut
 
-sub loadtest {
-    my ($script, %args) = @_;
+sub loadtest ($script, %args) {
     no utf8;    # Inline Python fails on utf8, so let's exclude it here
     my $casedir = $bmwqemu::vars{CASEDIR};
     my $script_path = find_script($script);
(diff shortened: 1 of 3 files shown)


autotest.pm
```diff
@@ -262,9 +254,7 @@ sub handle_sigterm {
     _exit(1);
 }
 
-sub start_process {
-    my $child;
-
+sub start_process ($child = undef) {
```
perlpunk, line 257: Is it intended that `$child` can be passed as an argument now?

basetest.pm
```diff
@@ -453,11 +418,7 @@ sub record_resultfile {
     $self->write_resultfile($filename, $output);
 }
 
-sub record_serialresult {
-    my ($self, $ref, $res, $string) = @_;
-
-    $string //= '';
-
+sub record_serialresult ($self, $ref, $res, $string = '') {
```
perlpunk, line 421: `$string` might end up as `undef`

basetest.pm
```diff
@@ -499,10 +456,7 @@ test details and returns it.
 
 =cut
 
-sub record_testresult {
-    my ($self, $result, %args) = @_;
-    $result //= 'unk';
-
+sub record_testresult ($self, $result = 'unk', %args) {
```
perlpunk, line 459: `$result` might end up as undef

basetest.pm
```diff
@@ -559,10 +511,7 @@ add screenshot with 'unk' result if an image is available
 
 =cut
 
-sub take_screenshot {
-    my ($self, $res) = @_;
-    $res ||= 'unk';
-
+sub take_screenshot ($self, $res = 'unk') {
```
perlpunk, line 514: same here

autotest.pm
okurz: No, that does not make sense. Thanks.

okurz (commented):
updated

Martchus (changes requested):

basetest.pm
```diff
@@ -499,10 +457,8 @@ test details and returns it.
 
 =cut
 
-sub record_testresult {
-    my ($self, $result, %args) = @_;
+sub record_testresult ($self, $result = undef, %args) {
```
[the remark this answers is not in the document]
okurz: yes because `%args` means "[0;inf["

basetest.pm
```diff
@@ -24,9 +24,7 @@ sub import {
     warnings->unimport('experimental::signatures');
 }
 
-sub new {
-    my ($class, $category) = @_;
-    $category ||= 'unknown';
+sub new ($class, $category = 'unknown') {
```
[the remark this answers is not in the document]
okurz: sorry, you just need to mention it … 27 more times and I will get it :)

Martchus (approved):

basetest.pm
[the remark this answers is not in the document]
perlpunk: Right. An array or hash is like a catchall.

PR #2020, opened 2022-04-07
Use proper perl "signatures" in common files - bmwqemu

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -4,7 +4,7 @@
 
 package bmwqemu;
 
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 use autodie ':all';
 use Fcntl ':flock';
 use Time::HiRes qw(sleep);
@@ -51,7 +51,7 @@ our @ovmf_locations = (
 our %vars;
 tie %vars, 'bmwqemu::tiedvars', %vars;
 
-sub result_dir { 'testresults' }
+sub result_dir () { 'testresults' }
 
 # deprecated functions, moved to log module
 {
@@ -69,16 +69,15 @@ sub result_dir { 'testresults' }
 use constant STATE_FILE => 'base_state.json';
 
 # Write a JSON representation of the process termination to disk
-sub serialize_state {
-    my $state = {@_};
-    bmwqemu::fctwarn($state->{msg}) if delete $state->{error};
-    bmwqemu::diag($state->{msg}) if delete $state->{log};
+sub serialize_state (%state) {
+    bmwqemu::fctwarn($state{msg}) if delete $state{error};
+    bmwqemu::diag($state{msg}) if delete $state{log};
     return undef if -e STATE_FILE;
-    eval { path(STATE_FILE)->spurt(encode_json($state)) };
+    eval { path(STATE_FILE)->spurt(encode_json(\%state)) };
     bmwqemu::diag("Unable to serialize fatal error: $@") if $@;
 }
 
-sub load_vars {
+sub load_vars () {
     my $fn = "vars.json";
     my $ret = {};
     local $/;
@@ -92,8 +91,7 @@ sub load_vars {
     return;
 }
 
-sub save_vars {
-    my (%args) = @_;
+sub save_vars (%args) {
     my $fn = "vars.json";
     unlink "vars.json" if -e "vars.json";
     open(my $fd, ">", $fn);
@@ -118,7 +116,7 @@ our $gocrbin = "/usr/bin/gocr";
 # set from isotovideo during initialization
 our $scriptdir;
 
-sub init {
+sub init () {
     load_vars();
 
     $vars{BACKEND} ||= "qemu";
@@ -134,7 +132,7 @@ sub init {
     log::init_logger;
 }
 
-sub _check_publish_vars {
+sub _check_publish_vars () {
     return 0 unless my $nd = $vars{NUMDISKS};
     my @hdds = map { $vars{"HDD_$_"} } 1 .. $nd;
     for my $i (1 .. $nd) {
@@ -147,7 +145,7 @@ sub _check_publish_vars {
     return 1;
 }
 
-sub ensure_valid_vars {
+sub ensure_valid_vars () {
     # defaults
     $vars{QEMUPORT} ||= 15222;
     $vars{VNC} ||= 90;
@@ -176,13 +174,13 @@ our $backend;
 
 # util and helper functions
 
-sub current_test {
+sub current_test () {
     require autotest;
     no warnings 'once';
     return $autotest::current_test;
 }
 
-sub update_line_number {
+sub update_line_number () {
     return unless current_test;
     return unless current_test->{script};
     my @out;
@@ -198,25 +196,25 @@ sub update_line_number {
 }
 
 # pretty print like Data::Dumper but without the "VAR1 = " prefix
-sub pp {
+sub pp (@args) {
     # FTR, I actually hate Data::Dumper.
-    my $value_with_trailing_newline = Data::Dumper->new(\@_)->Terse(1)->Useqq(1)->Dump();
+    my $value_with_trailing_newline = Data::Dumper->new(\@args)->Terse(1)->Useqq(1)->Dump();
     chomp($value_with_trailing_newline);
     return $value_with_trailing_newline;
 }
 
-sub log_call {
+sub log_call (@args) {
     my $fname = (caller(1))[3];
     update_line_number();
     my $params;
-    if (@_ == 1) {
-        $params = pp($_[0]);
+    if (@args == 1) {
+        $params = pp($args[0]);
     }
     else {
         # key/value pairs
         my @result;
-        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
-        while (my ($key, $value) = splice(@_, 0, 2)) {
+        my $should_hide = grep { ($_ // '') eq 'secret' } @args;
+        while (my ($key, $value) = splice(@args, 0, 2)) {
             if ($key =~ tr/0-9a-zA-Z_//c) {
                 # only quote if needed
                 $key = pp($key);
@@ -234,14 +232,13 @@ sub log_call {
 
 # backend management
 
-sub stop_vm {
+sub stop_vm () {
     return unless $backend;
     my $ret = $backend->stop();
     return $ret;
 }
 
-sub mydie {
-    my ($cause_of_death) = @_;
+sub mydie ($cause_of_death) {
     log_call(cause_of_death => $cause_of_death);
     croak "mydie";
 }
@@ -250,8 +247,7 @@ sub mydie {
 
 
 # store the obj as json into the given filename
-sub save_json_file {
-    my ($result, $fn) = @_;
+sub save_json_file ($result, $fn) {
     open(my $fd, ">", "$fn.new");
     my $json = Cpanel::JSON::XS->new->pretty->canonical->encode($result);
     print $fd $json;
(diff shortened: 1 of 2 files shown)


bmwqemu.pm
```diff
@@ -69,16 +69,16 @@ sub result_dir { 'testresults' }
 use constant STATE_FILE => 'base_state.json';
 
 # Write a JSON representation of the process termination to disk
-sub serialize_state {
-    my $state = {@_};
+sub serialize_state (@args) {
+    my $state = {@args};
```
perlpunk, line 73: Here I would use `%args`

bmwqemu.pm
```diff
@@ -198,24 +197,24 @@ sub update_line_number {
 }
 
 # pretty print like Data::Dumper but without the "VAR1 = " prefix
-sub pp {
+sub pp(@) {
```
perlpunk, line 200: This should be named and `@_` 2 lines later replaced with that.
At some point `@_` will be deprecated in signature subs.

bmwqemu.pm

bmwqemu.pm
```diff
@@ -198,24 +197,24 @@ sub update_line_number {
 }
 
 # pretty print like Data::Dumper but without the "VAR1 = " prefix
-sub pp {
+sub pp(@args) {
     # FTR, I actually hate Data::Dumper.
-    my $value_with_trailing_newline = Data::Dumper->new(\@_)->Terse(1)->Useqq(1)->Dump();
+    my $value_with_trailing_newline = Data::Dumper->new(\@args)->Terse(1)->Useqq(1)->Dump();
```
Martchus, line 202: In cases like this (so this counts for `log_call` as well) signatures have zero benefit. I suppose they incur an additional copy of the argument list into an array. Considering those functions are called a lot (so the additional slow-down accumulates) it is likely not worth it.

bmwqemu.pm
```diff
@@ -51,7 +51,7 @@ our @ovmf_locations = (
 our %vars;
 tie %vars, 'bmwqemu::tiedvars', %vars;
 
-sub result_dir { 'testresults' }
+sub result_dir() { 'testresults' }
```
Martchus, line 54:
```suggestion
sub result_dir () { 'testresults' }
```
I suppose we normally use a space here, don't we? So I'd use it consistently in places where signatures are used.

Martchus (changes requested):

bmwqemu.pm
okurz: you can teach perltidy :)

bmwqemu.pm
perlpunk: I wouldn't say _zero_ benefit.
You can see at one glance that the function takes an array, and this array is also easier to spot than the short `@_` in the body of the sub.
Currently signatures like that might indeed be a _bit_ slower because `@_` is still there, but `@_` will be dropped for performance reason in future versions of perl (only in signature subs of course).

perlpunk (approved):

bmwqemu.pm
Martchus: > Maybe, maybe not.

That's a bad justification for ignoring a possible performance regression. Considering the change is quite needless here I'd just avoid it.

(And yes, maybe I'm wrong and Perl optimizes this case and thus no copy is required. Annoyingly neither https://quick-bench.com nor https://godbolt.org are available for Perl.)

perlpunk: https://metacpan.org/pod/Benchmark comes with perl.

I'm surprised that you are so worried about it. We (well, Oli mostly) added signatures to hundreds of subroutines by now.

bmwqemu.pm
```diff
@@ -198,24 +197,24 @@ sub update_line_number {
...
-        $params = pp($_[0]);
+    if (@args == 1) {
+        $params = pp($args[0]);
     }
     else {
         # key/value pairs
         my @result;
-        my $should_hide = grep { ($_ // '') eq 'secret' } @_;
+        my $should_hide = grep { ($_ // '') eq 'secret' } @args;
         while (my ($key, $value) = splice(@_, 0, 2)) {
```
perlpunk, line 218: missed the `@_` here: `splice(@_, 0, 2)`

perlpunk (changes requested):
See my comment

bmwqemu.pm
Martchus: I'm not that worried. I just see that this change has a negative outcome considering the very little benefit signatures have here. Even if the negative outcome is very little it is still negative and thus I'm reluctant to think it is a good change. However, if you still want it, be welcome to dismiss my review.

okurz: I would prefer if we can adopt the simple rule: "Signatures everywhere" which I would also like to protect with an according style check

bmwqemu.pm
okurz: I did that and updated all other references of `@_` which includes the TiedHash methods which I previously did not change.

bmwqemu.pm
Martchus: You must hate passing parameters in C++ then :-)
I suppose extending the rule by "except args are just passed through" doesn't make it much more complicated so not really an argument. And I suppose in case we use prototypes we already need to make an exception.

bmwqemu.pm
okurz: With

```
git grep 'sub [a-z_A-Z0-9]\+()' | wc -l ; git grep 'sub [a-z_A-Z0-9]\+ ()' | wc -l
```

I found that there are 15 references of subs without separating space before `()` and 27 with a space. I suggest to allow both styles for now and try to make that consistent in a separate step. I prepared a PR for that: https://github.com/os-autoinst/os-autoinst/pull/2021

bmwqemu.pm
okurz: I would be ok if you can
1. Convince everybody else
2. Provide an according style check in xt/01-style.t

bmwqemu.pm
perlpunk: I found 15 and 27 a bit too less :)
```
git grep 'sub [a-z_A-Z0-9]\+(' | wc -l ; git grep 'sub [a-z_A-Z0-9]\+ (' | wc -l
19
665
```

okurz: Let's handle that in #2021 . perltidy already adds the space for any non-zero parameter signature list.

okurz: https://github.com/os-autoinst/os-autoinst/pull/2021 merged, rebase PR. Now we have consisent `sub <name> () {…`

bmwqemu.pm
```diff
@@ -69,16 +69,16 @@ sub result_dir { 'testresults' }
 use constant STATE_FILE => 'base_state.json';
 
 # Write a JSON representation of the process termination to disk
-sub serialize_state {
-    my $state = {@_};
+sub serialize_state (%args) {
+    my $state = {%args};
```
Martchus, line 73:
```suggestion
sub serialize_state (%state) {
```

And then use `%state` directly, e.g. `delete $state{error}` and `encode_json(\%state)`.

Martchus (changes requested):

bmwqemu.pm

okurz:
I think you guys could also help me where I need to add `uncoverable` again now as I think an uncoverable line needs to be extended to cover the signatures line as well

perlpunk (approved):

bmwqemu.pm
Martchus: 1. You could also try to convince me that there's no performance regression. "Maybe, maybe not." doesn't cut it.
2. So you're offloading the task to me?

But I don't care enough so I've just dismissed my review.

perlpunk:
```
#!/usr/bin/env perl
use v5.22;
use warnings;
use experimental qw/ signatures /;

use Data::Dumper;
use Benchmark qw/ timethese /;

sub old {
    my $value_with_trailing_newline = Data::Dumper->new(\@_)->Terse(1)->Useqq(1)->Dump();
}
sub signature (@x) {
    my $value_with_trailing_newline = Data::Dumper->new(\@x)->Terse(1)->Useqq(1)->Dump();
}

my @data = 'a' .. 'e'; # run 1
@data = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"; # run 2
@data = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" x 10; # run 3

timethese($ARGV[0] || -1, {
    old => sub { old(@data) for 1 .. ($ARGV[1] || 10) },
    signature => sub { signature(@data) for 1 .. ($ARGV[1] || 10) },
});
```
```
tina@linux-8vf8:/tmp/bench-signature% perl test.pl 300000 1 # run 1
Benchmark: timing 300000 iterations of old, signature...
       old:  2 wallclock secs ( 2.28 usr +  0.00 sys =  2.28 CPU) @ 131578.95/s (n=300000)
 signature:  3 wallclock secs ( 2.43 usr +  0.00 sys =  2.43 CPU) @ 123456.79/s (n=300000)
tina@linux-8vf8:/tmp/bench-signature% perl test.pl 300000 1 # run 2
Benchmark: timing 300000 iterations of old, signature...
       old:  2 wallclock secs ( 1.80 usr +  0.00 sys =  1.80 CPU) @ 166666.67/s (n=300000)
 signature:  2 wallclock secs ( 1.84 usr +  0.00 sys =  1.84 CPU) @ 163043.48/s (n=300000)
tina@linux-8vf8:/tmp/bench-signature% perl test.pl 300000 1 # run 3
Benchmark: timing 300000 iterations of old, signature...
       old:  3 wallclock secs ( 2.37 usr +  0.00 sys =  2.37 CPU) @ 126582.28/s (n=300000)
 signature:  2 wallclock secs ( 2.41 usr +  0.00 sys =  2.41 CPU) @ 124481.33/s (n=300000)
```

okurz: I ran the benchmark as well but with different modules to resemble better what we use:

```
use strictures;
```

this gives for 500k iterations 99s (old) vs. 95s (signatures) so signatures are slightly faster even. Can we agree on using signatures everywhere? If not I can also leave out the corresponding changes of the two functions where I used `@args` and we can discuss that separately.

perlpunk: it sounds unlikely to me that signatures are faster.
can you post your benchmark? i posted mine and signatures were slightly slower (but not enough that i would argue against them).

the usage of `strictures` or the way we enable signatures should not have any effect on the result.

okurz: Yes, `strictures` shouldn't matter. Full results for the three test arrays 1, 2, 3 in this order:

```
Benchmark: timing 500000 iterations of old, signature...
       old: 82 wallclock secs (82.37 usr +  0.00 sys = 82.37 CPU) @ 6070.17/s (n=500000)
 signature: 89 wallclock secs (89.12 usr +  0.01 sys = 89.13 CPU) @ 5609.78/s (n=500000)

Press ENTER or type command to continue
Benchmark: timing 500000 iterations of old, signature...
       old: 97 wallclock secs (96.75 usr +  0.12 sys = 96.87 CPU) @ 5161.56/s (n=500000)
 signature: 101 wallclock secs (100.41 usr +  0.00 sys = 100.41 CPU) @ 4979.58/s (n=500000)

Press ENTER or type command to continue
Benchmark: timing 500000 iterations of old, signature...
       old: 68 wallclock secs (67.82 usr +  0.00 sys = 67.82 CPU) @ 7372.46/s (n=500000)
 signature: 67 wallclock secs (67.36 usr +  0.01 sys = 67.37 CPU) @ 7421.70/s (n=500000)
```

so signatures seem to be slower for the first and second set but faster for the third set although I would say that all cases are not significant.

Martchus (approved):
Considering the benchmarks

PR #2025, opened 2022-04-11
Use proper perl "signatures" in common files - lockapi

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -4,7 +4,7 @@
 ## synchronization API
 package lockapi;
 
-use Mojo::Base 'Exporter';
+use Mojo::Base 'Exporter', -signatures;
 use Scalar::Util 'looks_like_number';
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

Martchus (approved):

dzedro:
Is this breaking `mutex_wait` ? https://openqa.suse.de/tests/8520127#step/boot_to_desktop/3

okurz:
> Is this breaking `mutex_wait` ? https://openqa.suse.de/tests/8520127#step/boot_to_desktop/3

As it's the only method that is not currently covered properly by unit tests, yes :)

Fix in https://github.com/os-autoinst/os-autoinst/pull/2026

Thank you for the report.

PR #2026, opened 2022-04-12
Fix potentially missing parameters on mutex_wait

okurz:
Based on report in https://github.com/os-autoinst/os-autoinst/pull/2025#issuecomment-1096436103

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -95,7 +95,7 @@ sub mutex_create ($name, @) {
 }
 
 # Wrapper for mutex_lock & mutex_unlock
-sub mutex_wait ($name, $where, $info) {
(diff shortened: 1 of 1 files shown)


okurz:
Hotpatched on OSD with

```
sudo salt --no-color --state-output=changes -C 'G@roles:worker' cmd.run 'curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/2026.patch | patch -p1 --directory=/usr/lib/os-autoinst'
```

perlpunk (approved):

okurz:
Need to merge because obviously we don't have test coverage for the only line that introduced that regression :)

PR #2027, opened 2022-04-12
Enlarge visible windows only in fullscreen method

rakoenig:
See ticket: [poo#109795](https://progress.opensuse.org/issues/109795) for details.

### DIFF
--- a/consoles/localXvnc.pm
+++ b/consoles/localXvnc.pm
@@ -51,7 +51,7 @@ sub fullscreen ($self, $args) {
     die "Missing 'xdotool'" unless $xdotool;
 
     # search for YaST Window and grab the id
-    my $window_id = qx"DISPLAY=$display $xdotool search --sync --limit 1 --name $window_name";
+    my $window_id = qx"DISPLAY=$display $xdotool search --sync --onlyvisible --name $window_name";
     $window_id =~ s/\D//g;
 
     # resize and move window to fit in icewm


okurz:
@rakoenig please see https://github.com/os-autoinst/os-autoinst/pull/2027#issuecomment-1096613270 and provide a manual verification test where we can see that working in action as we currently don't have automatic CI tests covering that behaviour.

Martchus (commented):
Looks good, there's only one typo ("poblems") in the commit message.

Martchus (approved):

rakoenig:
> @rakoenig Did you see this comment?

Yes I saw it and I'm desperately trying to make a verification run. But so far what people suggested on the Slack-channel didn't work for me. Running the worker in the container was a complete fail. I'm a bit lost on how to tweak the workers from remote, but looks like there is not much of a documentation or even a tutorial about that stuff.

okurz:
@Martchus could you help rakoenig to get this tested?

Martchus:
When I remember the chat history correctly, that change was actually tested (by manually logging in on the host and invoking the command manually). Hence I've also already approved the change.

If it was my PR I'd probably go for the approach documented here: https://progress.opensuse.org/projects/openqav3/wiki/#Use-a-production-host-for-testing-backend-changes-locally-eg-svirt-powerVM-IPMI-bare-metal-s390x-etc - This is the approach used so far when developing with the svirt backend.

Alternatively, you can also hot-patch an instance. [Take it out of production](https://progress.opensuse.org/projects/openqav3/wiki/Wiki#Take-machines-out-of-salt-controlled-production) and apply a patch via `patch -d /usr/lib/os-autoinst -p1 -i …`. To run a test on that specific worker, clone it with the worker-specific `WORKER_CLASS` setting. You don't need to restart any services on the host for os-autoinst changes to apply. When you're done, revert your changes and [bring the worker back](https://progress.opensuse.org/projects/openqav3/wiki/Wiki#Bring-back-machines-into-salt-controlled-production).

(All other suggestions were only luring you into uncharted territory. They might be better in the long run but you'll have to figure out details on your own at this point.)

okurz (approved):
I am ok if the command was tried out on the actual production machines

PR #2028, opened 2022-04-12
Correct croak on type_string with wrong function name

b10n1k:
When type_string is failing due to a timeout the error message reffers to the failed subroutine as
`type_screen` instead of the correct name.

### DIFF
--- a/consoles/ssh_screen.pm
+++ b/consoles/ssh_screen.pm
@@ -61,7 +61,7 @@ sub type_string ($self, $nargs) {
     while ($written < length($text)) {
         my $elapsed = consoles::serial_screen::elapsed($stime);
 
-        croak("type_screen(): Timed out after $elapsed seconds.")
+        croak((caller(0))[3] . ": Timed out after $elapsed seconds.")
           if ($elapsed > TYPE_STRING_TIMEOUT);
 
         my $chunk = $self->ssh_channel->write(substr($text, $written));
--- a/t/26-ssh_screen.t
+++ b/t/26-ssh_screen.t
@@ -8,9 +8,25 @@ use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '5';
 use consoles::ssh_screen;
+use consoles::serial_screen;
+use bmwqemu;
+use Test::Exception;
+use Test::MockObject;
+use Test::MockModule;
 
 my $screen = consoles::ssh_screen->new(ssh_connection => 'My_Con', ssh_channel => 'My_Chan');
 is($screen->{fd_read}, 'My_Chan', 'SSH channel is used for reading');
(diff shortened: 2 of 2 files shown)


b10n1k:
@os-autoinst/tools-team

okurz (changes requested):
Some typos in the git commit message, e.g. "reffers" and "i". Please consistently wrap at 80 characters and leave out the "Signed-off-by". We don't need it and IMHO it's not adding any value if you do that yourself.

okurz:
> @os-autoinst/tools-team

No need to ping the team. We commonly review all pull requests in this repo.

consoles/ssh_screen.pm
```diff
@@ -61,7 +61,7 @@ sub type_string ($self, $nargs) {
     while ($written < length($text)) {
         my $elapsed = consoles::serial_screen::elapsed($stime);
 
-        croak("type_screen(): Timed out after $elapsed seconds.")
+        croak("(caller(0))[3]: Timed out after $elapsed seconds.")
```
Martchus, line 64: I doubt you really want to output `(caller(0))[3]` literally here - so the expression should not be within a string literal.

Martchus (changes requested):

consoles/ssh_screen.pm
b10n1k: *this change is ready for review*

okurz:
@b10n1k code change looks good. Have you seen https://github.com/os-autoinst/os-autoinst/pull/2028#issuecomment-1096985819 ? Can you provide a verification run where we see this in action?

Martchus:
A unit test would be sufficient as well.

b10n1k:
@Martchus test added and @okurz test shows also `where you can see this in action` :).

t/26-ssh_screen.t
```diff
@@ -8,9 +8,26 @@ use FindBin '$Bin';
...
+use Test::MockObject;
+use Test::MockModule;
 
 my $screen = consoles::ssh_screen->new(ssh_connection => 'My_Con', ssh_channel => 'My_Chan');
 is($screen->{fd_read}, 'My_Chan', 'SSH channel is used for reading');
 is($screen->{fd_write}, 'My_Chan', 'SSH channel is used for writing');
 
+subtest 'Correct message when type_string timeouts' => sub {
+    use consoles::serial_screen;
+    use bmwqemu;
```
Martchus, line 21: I'd prefer to move them to the top. They're already loaded at compile time anyways.

Martchus (commented):
~~Likely tidy won't like it.~~ Apparently it does.

t/26-ssh_screen.t
b10n1k: Done

Martchus (approved):

okurz (approved):
Very nice

PR #2030, opened 2022-04-19
Git wheel repo tests

kalikiana:


t/14-isotovideo.t
```diff
@@ -105,6 +105,35 @@ subtest 'isotovideo with git refspec specified' => sub {
             opts => "casedir=$data_dir/tests test_git_refspec=deadbeef _exit_after_schedule=1") } qr/Checking.*local.*deadbeef/, 'refspec in local git repository would be checked out';
 };
 
+subtest 'isotovideo with wheels' => sub {
+    chdir($pool_dir);
+    unlink('vars.json') if -e 'vars.json';
+
+    my $utils_mock = Test::MockModule->new('OpenQA::Isotovideo::Utils');
+    $utils_mock->redefine(capture => sub { return; });
```
Martchus, line 113:
```suggestion
    $utils_mock->noop('capture');
```

or if you want to keep the check whether `capture` exists:

```suggestion
    $utils_mock->redefine(capture => undef);
```

isotovideo
```diff
@@ -246,6 +250,25 @@ $bmwqemu::vars{TEST_GIT_HASH} = checkout_git_refspec($bmwqemu::vars{CASEDIR} =>
...
+    my @errors = $validator->validate($spec);
+    die 'Invalid YAML: ' . join(',', @errors) if @errors;
+    die 'Unsupported version: Version must be "v0.1"' if $spec->{version} ne 'v0.1';
+
+    foreach my $repo (@{$spec->{wheels}}) {
+        $repo = "https://github.com/$repo.git" unless $repo =~ '^http';
+        checkout_git_repo_and_branch($specfile, repo => $repo);
+    }
+}
+
```
okurz, line 271: We discussed this in the unblock in 2022-07-06 and agreed that we do not necessarily need to wait for https://progress.opensuse.org/issues/81899. The suggestion is to move this into a module instead of adding to this procedural section of code and provide a test. If no unit test with full mocking seems reasonable or feasible make it work with tiny empty fake git repos created on the fly. I think we also create and checkout git repos in other tests already.

kalikiana:
See #2004

PR #2033, opened 2022-04-20
Use OpenQA::Test::PatchDeparse to avoid warnings

perlpunk:
I tested it on a (curremtly) broken PR (#2032) to check if the warning is really gone.


Martchus (approved):

okurz:
I suggest to exclude external/ from coverage analysis

perlpunk:
> I suggest to exclude external/ from coverage analysis

done

perlpunk:
I also excluded the external directory for perltidy now

okurz (approved):

PR #2034, opened 2022-04-20
Sync tools/tidy with openQA

perlpunk:
* Add --list
* Formatting
* Allow tools/perlfiles for special selection

Issue: https://progress.opensuse.org/issues/110142

For now I'm not adding this to os-autoinst-common because I have to solve the problem that os-autoinst-distri-opensuse uses `readlink -f $0` to get the `os-autoinst/cpanfile`.

See also https://github.com/os-autoinst/openQA/pull/4614

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -3,11 +3,10 @@
 #
 # perltidy rules can be found in ../.perltidyrc
 #
-
 usage() {
     cat << EOF
 Usage:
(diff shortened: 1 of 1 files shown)


okurz (commented):
Looks good. But please make sure that this also still works within os-autoinst-distri-opensuse where tidy is referenced from os-autoinst

Martchus (approved):

okurz (approved):

PR #2036, opened 2022-04-22
PoC: Add initial support for SIE instruction and AP Facility

foursixnine:
This is to support https://progress.opensuse.org/issues/110170 note that likely we will need to do a bit of testing on an s390 host to properly check that the support works, and is the correct one.﻿

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -112,6 +112,22 @@ sub _init_xml ($self, $args = {}) {
     $elem = $doc->createElement('pae');
     $features->appendChild($elem);
 
+    ## Following features are required for s390's SIE instruction
+    # https://share.confex.com/share/117/webprogram/Handout/Session9564/9564%20-%20zVM%20Security%20and%20Integrity.pdf p6
+    # https://qemu.readthedocs.io/en/latest/system/s390x/vfio-ap.html#start-interpretive-execution-sie-instruction
+    
+    if ( defined $bmwqemu::vars{S390_SIE_AP_MATRIX_PATH} ){
+        $elem = $doc->createElement('ap');
+        $features->appendChild($elem);
+
+        $elem = $doc->createElement('apqci');
+        $features->appendChild($elem);
+
+        $elem = $doc->createElement('aptf');
+        $features->appendChild($elem);
+	add_vfio_mediated_device(path => $bmwqemu::vars{S390_SIE_AP_MATRIX_PATH});
+    }
+
     if ($self->vmm_family eq 'xen' and $self->vmm_type eq 'linux') {
         $elem = $doc->createElement('kernel');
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -112,6 +112,22 @@ sub _init_xml ($self, $args = {}) {
     $elem = $doc->createElement('pae');
     $features->appendChild($elem);
 
+    ## Following features are required for s390's SIE instruction
+    # https://share.confex.com/share/117/webprogram/Handout/Session9564/9564%20-%20zVM%20Security%20and%20Integrity.pdf p6
+    # https://qemu.readthedocs.io/en/latest/system/s390x/vfio-ap.html#start-interpretive-execution-sie-instruction
+    
+    if ( defined $bmwqemu::vars{S390_SIE_AP_MATRIX_PATH} ){
+        $elem = $doc->createElement('ap');
+        $features->appendChild($elem);
```
Martchus, line 121: I know you're just copying from above, but I'd prefer:
```suggestion
        $features->appendChild($doc->createElement('ap'));
```

consoles/sshVirtsh.pm
```diff
@@ -112,6 +112,22 @@ sub _init_xml ($self, $args = {}) {
     $elem = $doc->createElement('pae');
     $features->appendChild($elem);
 
+    ## Following features are required for s390's SIE instruction
+    # https://share.confex.com/share/117/webprogram/Handout/Session9564/9564%20-%20zVM%20Security%20and%20Integrity.pdf p6
+    # https://qemu.readthedocs.io/en/latest/system/s390x/vfio-ap.html#start-interpretive-execution-sie-instruction
+    
+    if ( defined $bmwqemu::vars{S390_SIE_AP_MATRIX_PATH} ){
```
Martchus, line 119:
```suggestion
    # enable features required for s390's SIE instruction, see
    # - https://share.confex.com/share/117/webprogram/Handout/Session9564/9564%20-%20zVM%20Security%20and%20Integrity.pdf#page=6
    # - https://qemu.readthedocs.io/en/latest/system/s390x/vfio-ap.html#start-interpretive-execution-sie-instruction
    if ( defined $bmwqemu::vars{S390_SIE_AP_MATRIX_PATH} ){
```
(Note the referencing of the pdf page via `#page=N`.)

Or just make it a function called `enable_features_requires_for_s390_sie_instruction` (of course keeping the comments with references).

Martchus:
Regarding testing: It falls into the same category as #2027 so see [my comment](https://github.com/os-autoinst/os-autoinst/pull/2027#issuecomment-1104919833=) there.

consoles/sshVirtsh.pm
kraih: Agreed. Reusing the same variable over and over is not nice.

PR #2037, opened 2022-04-22
Runner: Move schedule loading and dir setup into module

kalikiana:

### DIFF
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -1,0 +1,87 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Isotovideo::Runner;
+use Mojo::Base -base, -signatures;
+
+use OpenQA::Isotovideo::Utils qw(checkout_git_repo_and_branch checkout_git_refspec);
+
+use bmwqemu;
+use autotest;
+
+use Mojo::File 'path';
+use Try::Tiny;
+
+# a run is comprised of these individual resources
+has [qw(casedir needles_dir productdir)] => undef;
+
+sub new ($class, %args) {
+    my $self = $class->SUPER::new(
+        casedir => undef,
+        needles_dir => undef,
+        productdir => path($args{productdir}),
+    );
+    $bmwqemu::vars{PRODUCTDIR} = $self->productdir->to_string;
+
+    return $self;
+}
+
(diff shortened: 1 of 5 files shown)


OpenQA/Isotovideo/Runner.pm
```diff
@@ -0,0 +1,80 @@
...
+
+sub new ($class, %args) {
+    my $self = $class->SUPER::new(
+        casedir => path($args{casedir}),
+        needles_dir => $args{needles_dir},
+        productdir => path($args{productdir}),
+        test_git_hash => undef,
+    );
+
+    checkout_git_repo_and_branch('CASEDIR');
```
perlpunk, line 29: I would avoid doing too much in a constructor.

okurz (commented):
I don't yet see the benefit of moving from one file to another as long as there is not less in isotovideo. As announced I am currently preparing a change to move more out of isotovideo into modules.

kalikiana:
> I don't yet see the benefit of moving from one file to another as long as there is not less in isotovideo. As announced I am currently preparing a change to move more out of isotovideo into modules.

Which isn't my goal. I'm doing this to enable unit tests.

OpenQA/Isotovideo/Runner.pm
```diff
@@ -0,0 +1,80 @@
...
+# a run is comprised of these individual resources
+has [qw(casedir needles_dir productdir test_git_hash)] => undef;
+
+sub new ($class, %args) {
+    my $self = $class->SUPER::new(
+        casedir => path($args{casedir}),
+        needles_dir => $args{needles_dir},
+        productdir => path($args{productdir}),
+        test_git_hash => undef,
+    );
```
Martchus, line 27:
```suggestion
sub new ($class, @args) {
    my $self = $class->SUPER::new(@args);
```

OpenQA/Isotovideo/Runner.pm
Martchus: Avoiding this invocation directly in the constructor would mean we could avoid having the constructor altogether (which would be a plus, indeed).

Martchus (commented):
I'm wondering whether it makes sense opening two construction sites very close to each other (considering @okurz changes to move things out of isotovideo). But I suppose the refactoring is generally ok if it helps to improve tests.

OpenQA/Isotovideo/Runner.pm
kalikiana: That would mean the caller has to provide `path` objects, though.

PR #2039, opened 2022-04-22
WIP Sync .perltidyrc with openQA (only line length)

perlpunk:
+-l=120   # 120 characters per line


perlpunk:
As we can see here, we more or less only have comments, that are now mostly moved to the let, so they should be reformatted.
Very few actual code lines, so maybe we should go for this.
WDYT?

perlpunk:
os-autoinst-distri-opensuse using the same .perltidyrc - I give up...

PR #2043, opened 2022-04-29
Extend svirt documentation

Martchus:
* I've tested locally whether it works and updated and extended the
  documentation according to what works for me.
* See https://progress.opensuse.org/issues/106867

### DIFF
--- a/doc/backends.md
+++ b/doc/backends.md
@@ -1,52 +1,94 @@
 # Backends in openQA
-OpenQA (or actually os-autoinst) supports multiple backends to run the SUT.
-This document is meant to describe their particularities. So far only svirt
-is covered, though.
+OpenQA (or actually os-autoinst) supports multiple backends to run the SUT. This
+document is meant to describe their particularities. So far only svirt and
+Vagrant are covered, though.
 
-For backend specific variables, there is a [separate documentation](backend_vars.asciidoc).
+For backend specific variables, there is a
+[separate documentation](backend_vars.asciidoc).
 
 ## svirt
 This backend establishes an SSH connection to another machine to start there
 a virtual machine using libvirt/virsh which uses QEMU, Xen, Hyper-V or VMWare
 under the hood.
 
-### Local setup
-Simply configure it to connect to the local machine and install libvirt/virsh.
+The svirt backend is enabled via `BACKEND=svirt` and the VM used under the hood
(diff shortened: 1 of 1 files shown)


doc/backends.md
```diff
@@ -1,52 +1,94 @@
...
-#VIRSH_PASSWORD=# see notes
-VIRSH_GUEST=127.0.0.1
 VIRSH_VMM_FAMILY=kvm
 VIRSH_VMM_TYPE=hvm
 ```
 
-##### Notes
-To allow an SSH connection to your local machine, either put your (root) password
-in the worker configuration or add your key to `$HOME/.ssh/authorized_keys`.
+If you add multiple instance, be sure to assign a different `VIRSH_INSTANCE` and
```
perlpunk, line 40: multiple instances

doc/backends.md
```diff
@@ -1,52 +1,94 @@
...
+
+#### SSH configuration
+To allow an SSH connection to your local machine, put your root password in the
+worker configuration as shown in the previous section. If your root password is
+entered from the configuration, that `type_string` command is logged. So be
+aware that your root password will end up in `autoinst-log.txt`.
+
+In any case, also be sure to allow root login in `/etc/ssh/sshd_config` and that
+`sshd` is actually started (e.g. using `zypper in openssh-server` and
+`systemctl start  sshd.service` under openSUSE).
```
okurz, line 58: Remove the double space after "start" please. Also I recommend to leave out the `.service`

doc/backends.md
```diff
@@ -1,52 +1,94 @@
...
 zypper in virt-manager # a GUI for libvirt, not really required for openQA but sometimes still useful
-systemctl start libvirtd sshd
+systemctl start libvirtd
+```
+
+Otherwise there is no configuration required. The test will configure libvirt on
+its own. By default, the openSUSE test distribution uses the domain name
+`openQA-SUT-1`. So you can for instance use `virsh dumpxml openQA-SUT-1` to
+investigate the configuration of the running virtual machine. I also recommended
+using `virt-manager` for viewing the created VM.
```
okurz, line 72: Better write "One can also use `virt-manager` ..."

okurz (changes requested):

okurz (approved):

kraih (approved):

PR #2045, opened 2022-05-03
Add perl-Template-Toolkit to dependencies for devel

mimi1vx:
Is soft dependency od perl-Devel-Cover and without it commmand
`make coverage` fails


okurz:
s/dependecies/dependencies/

okurz (approved):

Martchus (approved):

PR #2046, opened 2022-05-04
Use proper perl "signatures" in common files - commands

okurz:
Using a command like

```
tools/tidy --only-changed
```

and some manual fixes.

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -4,7 +4,7 @@
 
 package commands;
 
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 use autodie ':all';
 
 require IPC::System::Simple;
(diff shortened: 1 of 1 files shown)


commands.pm
```diff
@@ -20,8 +20,7 @@ BEGIN {
     $ENV{MOJO_REACTOR} = 'Mojo::Reactor::Poll';
 }
 
-# Automatically enables "strict", "warnings", "utf8" and Perl 5.10 features
-use Mojolicious::Lite;
+use Mojolicious::Lite -signatures;
```
perlpunk, line 23: why do we need to enable signatures twice?

okurz: I don't know. Do you think we should just remove this line?

okurz: If I remove the import then t/07-commands.t fails on unknown definition of "app". If I remove the `-signatures` flag then we get "The signatures feature is experimental"

Martchus: Yes, we're using `Mojolicious::Lite` here.

> If I remove the -signatures flag then we get "The signatures feature is experimental"

Maybe that module is special then. So I guess we can just keep the `-signatures` then.

Martchus (approved):

commands.pm
perlpunk: hm, ok. I guess it doesn't hurt to have it twice

perlpunk (approved):

PR #2048, opened 2022-05-05
Add basic backend-svirt.t test

mimi1vx:
https://progress.opensuse.org/issues/109620

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -100,12 +100,7 @@ sub scp_get ($self, $src, $dest) {
 }
 
 sub can_handle ($self, $args) {
-    my $vars = \%bmwqemu::vars;
-    if ($args->{function} eq 'snapshots' && !$bmwqemu::vars{HDDFORMAT} eq 'raw') {
-        # Snapshots via libvirt are supported on KVM and, perhaps, ESXi. Hyper-V uses native tools.
-        return {ret => 1} if _vmm_family() =~ qr/kvm|hyperv|vmware/;
-    }
-    return;
+    $args->{function} eq 'snapshots' && _vmm_family =~ qr/kvm|hyperv|vmware/ ? {ret => 1} : undef;
 }
 
 sub is_shutdown ($self, @) {
@@ -136,7 +131,7 @@ sub save_snapshot ($self, $args) {
         $rsp = $self->run_ssh_cmd("virsh $libvirt_connector snapshot-create-as $vmname $snapname");
     }
     bmwqemu::diag "SAVE VM $vmname as $snapname snapshot, return code=$rsp";
-    $self->die unless ($rsp == 0);
+    $self->die if $rsp;
     return;
 }
 
@@ -297,7 +292,7 @@ sub serial_terminal_log_file ($self) {
     return '/tmp/' . SERIAL_TERMINAL_LOG_PATH . '.' . $bmwqemu::vars{JOBTOKEN};
 }
 
-sub check_socket ($self, $fh, $write = undef) { $self->check_ssh_serial($fh, $write) || $self->SUPER::check_socket($fh, $write) }
+sub check_socket ($self, $fh, $write = undef) { $self->check_ssh_serial($fh, $write) || $self->SUPER::check_socket($fh, $write) }    # uncoverable statement
 
 sub stop_serial_grab ($self, @) {
     $self->stop_ssh_serial;
--- a/t/29-backend-svirt.t
+++ b/t/29-backend-svirt.t
@@ -1,0 +1,124 @@
+#!/usr/bin/perl
+
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use Mojo::Base -strict, -signatures;
+use Test::Warnings qw(:all :report_warnings);
+use Test::Fatal;
+use Test::MockModule;
+use Test::MockObject;
+use Test::Mock::Time;
+use Test::Output qw(stderr_like);
+use Mojo::File qw(tempdir);
+use Mojo::Util qw(scope_guard);
+use FindBin '$Bin';
+use lib "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '5';
+use backend::svirt;
+use distribution;
+
+my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
+chdir $dir;
+my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
+
+my $ssh_object = Test::MockObject->new();
+$ssh_object->set_true(qw/disconnect scp_get blocking/);
+$ssh_object->set_always('error', 'Mock SSH Error');
+
+my $chan_object = Test::MockObject->new();
+$chan_object->set_true(qw/exec/);
+
+my $run_ssh_cmd_mock = Test::MockModule->new('backend::baseclass');
+$run_ssh_cmd_mock->redefine(new_ssh_connection => sub ($self, %args) { return $ssh_object });
+$run_ssh_cmd_mock->redefine(start_ssh_serial => sub { return ($ssh_object, $chan_object) });
+
+sub redefine_ssh ($excepted = 0) {
+    $run_ssh_cmd_mock->redefine(run_ssh_cmd => sub ($self, $cmd, %args) {
+            return ref($excepted) eq 'ARRAY' ? @$excepted : $excepted;
+    });
+}
+
+$bmwqemu::vars{WORKER_HOSTNAME} = 'localhost';
+$bmwqemu::vars{VIRSH_HOSTNAME} = 'foobar';
+$bmwqemu::vars{VIRSH_USERNAME} = 'root';
+$bmwqemu::vars{VIRSH_PASSWORD} = 'password';
+$bmwqemu::vars{JOBTOKEN} = "JobToken";
+
+my $distri = $testapi::distri = distribution->new();
+
+subtest 'Generic svirt backend' => sub {
+    my @diag_log;
+    my $backend = backend::svirt->new;
+    my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+    # silence some log output for cleaner tests
+    $bmwqemu_mock->noop('diag');
+    $bmwqemu_mock->noop('log_call');
+    redefine_ssh;
+    $backend->{need_delete_log} = 1;
+    ok $backend->do_start_vm, 'can start vm';
+    # not kvm/hyperv or vmware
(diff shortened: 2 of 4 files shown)


Martchus (commented):
It would be good to know how it relates to `t/22-svirt.t` and why you haven't extended the existing test file. Additionally, I hope you have checked that these parts aren't already covered (because testing the same code twice is counter productive as it only extends the CI runtime with no additional value).

okurz (changes requested):
@mimi1vx have you seen https://github.com/os-autoinst/os-autoinst/pull/2048#pullrequestreview-963165735 ? Maybe you are doing too much? Also please s/backent/backend/ :)

t/29-backend-svirt.t
```diff
@@ -0,0 +1,106 @@
...
+
+my $run_ssh_cmd_mock = Test::MockModule->new('backend::baseclass');
+
+sub redefine_ssh ($excepted = 0) {
+    $run_ssh_cmd_mock->redefine(run_ssh_cmd => sub ($self, $cmd, %args) {
+            if (ref($excepted) eq 'ARRAY') {
+                return @$excepted;
+            } else {
+                return $excepted;
+            }
```
okurz, line 38:
```suggestion
            return ref($excepted) eq 'ARRAY' ? @$excepted : $excepted;
```

okurz (changes requested):

Martchus:
Have you seen my previous comment?

mimi1vx:
> It would be good to know how it relates to `t/22-svirt.t` and why you haven't extended the existing test file.
this is unit tests like `29-backend-amt.t` as proposed in poo. I think  target of `t/22-svirt.`  is different
> Additionally, I hope you have checked that these parts aren't already covered (because testing the same code twice is co
cunter productive as it only extends the CI runtime with no additional value).

most of tests is new coverage

mimi1vx:
@Martchus any idea why https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/2048/changes#D1L105 isn't reached ... ??

okurz:
> @Martchus any idea why https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/2048/changes#D1L105 isn't reached ... ??

I don't know why it's not reached but I guess you could simply turn around the `return ... if` to `if ... return` and combine the two if-statements

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+subtest 'Generic svirt backend' => sub {
+    my $backend = backend::svirt->new;
+    my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+    # silence some log output for cleaner tests
+    $bmwqemu_mock->noop('diag');
+    $bmwqemu_mock->noop('log_call');
+    redefine_ssh;
+    $backend->{need_delete_log} = 1;
+    ok $backend->do_start_vm, 'can start vm';
+    is $backend->can_handle({function => 'snapshots'}), undef, 'can handle snapshots';
```
Martchus, line 61: The expected value is `undef` and but the test is called 'can handle snapshots'. Either `undef` should be `{ret => 1}` or 'can handle snapshots' should be 'can not handle snapshots' for this to make sense.

The same counts for the subsequent tests. They assume a falsy value is correct but the test name sounds positive.

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+
+    my $backend = backend::svirt->new;
+    my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+    # silence some log output for cleaner tests
+    $bmwqemu_mock->noop('diag');
+    $bmwqemu_mock->noop('log_call');
+    redefine_ssh;
+    $backend->{need_delete_log} = 1;
+    ok $backend->do_start_vm, 'can start vm';
+    is $backend->can_handle({function => 'snapshots'}), undef, 'can handle snapshots';
```
Martchus, line 83: This should really be a truthy value as we should support snapshots for vmware (at least judging by the svrit code). Maybe that explains why https://app.codecov.io/gh/os-autoinst/os-autoinst/compare/2048/changes#D1L105 is not covered.

The same counts for HyperV.

Martchus (changes requested):

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+chdir $dir;
+my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
+
+my $ssh_object = Test::MockObject->new();
+$ssh_object->set_true(qw/disconnect scp_get blocking/);
+$ssh_object->set_always('error', 'Mock SSH Error');
+my $chan_object = Test::MockObject->new();
+$chan_object->set_true(qw/exec/);
+
+
```
Martchus, line 32:
```suggestion
```

t/29-backend-svirt.t
mimi1vx: yes , this is exactly what worries me ...

culprit:
`if ($args->{function} eq 's...' && !bmwqmenu::vars{HDDFORMAT} eq 'raw') {}`

for any value of `HDDFORMAT` it returns **FALSE**

when I modified condition to use `bmwqmenu::vars{HDDFORMAT} ne 'raw'` it works as intended but when is `HDDFORMAT` undefined throws exception ...

so I tried find where and how is used this variable -> only in `backend::qemu` to kill backend as unsupported var

Martchus: `!bmwqmenu::vars{HDDFORMAT} eq 'raw'` looks wrong, indeed. It negates `bmwqmenu::vars{HDDFORMAT}` and *then* does the equality comparison. I'd also switch it to `ne`. To handle the `undef` case, I'd use `(bmwqmenu::vars{HDDFORMAT} // '') ne 'raw'`.

mimi1vx: I'm more in favor to remove `HDDFORMAT`. And also from documentation from backend cars

backend/svirt.pm
```diff
@@ -100,11 +100,7 @@ sub scp_get ($self, $src, $dest) {
 }
 
 sub can_handle ($self, $args) {
-    my $vars = \%bmwqemu::vars;
-    if ($args->{function} eq 'snapshots' && !$bmwqemu::vars{HDDFORMAT} eq 'raw') {
-        # Snapshots via libvirt are supported on KVM and, perhaps, ESXi. Hyper-V uses native tools.
-        return {ret => 1} if _vmm_family() =~ qr/kvm|hyperv|vmware/;
-    }
+    if ($args->{function} eq 'snapshots' && _vmm_family =~ qr/kvm|hyperv|vmware/) { return {ret => 1}; }
     return;
```
Martchus, line 104:
```suggestion
    ($args->{function} eq 'snapshots' && _vmm_family =~ qr/kvm|hyperv|vmware/) ? {ret => 1} : undef;
```

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+    my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+    # silence some log output for cleaner tests
+    $bmwqemu_mock->noop('diag');
+    $bmwqemu_mock->noop('log_call');
+    redefine_ssh;
+    $backend->{need_delete_log} = 1;
+    ok $backend->do_start_vm, 'can start vm';
+    # not kvm/hyperv or vmware
+    is $backend->can_handle({function => 'snapshots'}), undef, 'can not handle snapshots';
+    is $backend->save_snapshot({name => "Snap1"}), undef, 'can save snapshot';
```
Martchus, line 62:
```suggestion
    is $backend->save_snapshot({name => "Snap1"}), undef, 'can not save snapshot';
```
Or maybe just remove the check for the return value if it is meaningless here.

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+    is $backend->load_snapshot({name => "Snap1"}), '', 'can load snapshot';
```
Martchus, line 63:
```suggestion
    is $backend->load_snapshot({name => "Snap1"}), '', 'can not load snapshot';
```
Or maybe just remove the check for the return value if it is meaningless here.

t/29-backend-svirt.t
```diff
@@ -0,0 +1,120 @@
...
+    my $backend = backend::svirt->new;
+    my $bmwqemu_mock = Test::MockModule->new('bmwqemu');
+    # silence some log output for cleaner tests
+    $bmwqemu_mock->noop('diag');
+    $bmwqemu_mock->noop('log_call');
+    redefine_ssh;
+    $backend->{need_delete_log} = 1;
+    ok $backend->do_start_vm, 'can start vm';
+    is $backend->can_handle({function => 'snapshots'})->{ret}, 1, 'can handle snapshots';
+    is $backend->save_snapshot({name => "Snap1"}), undef, 'can save snapshot';
```
Martchus, line 84:
```suggestion
    is $backend->save_snapshot({name => "Snap1"}), 1, 'can save snapshot';
```
If it doesn't return a truthy value then either change the function accordingly or remove the check (which is otherwise quite pointless).

**The same counts for further occurrences.**

Martchus (changes requested):
Please rename [Small fixes for svirt backend and removal of HDDFORMAT check](https://github.com/os-autoinst/os-autoinst/pull/2048/commits/e1fd2cb7428c654f07c8e698b3019543bf53ed0a) to match our formatting style (start with a verb in imperative).

okurz (changes requested):
See comments by Martchus

t/29-backend-svirt.t
mimi1vx: `''` is standard return of load snapshot, only on VMware is `vmware fixup`

t/29-backend-svirt.t
mimi1vx: there is small problem ... `save_snapshot` or `load_snapshot` never checks if backend can_handle snapshot ... so in test it works without problems

t/29-backend-svirt.t
Martchus: If it really returns certain different strings that particular function should be ok.

t/29-backend-svirt.t
mimi1vx: > If it doesn't return a truthy value then either change the function accordingly or remove the check (which is otherwise quite pointless).

change backend functions ? so change behavior to be different for rest of backends ?

Martchus: Right, if that would make it inconsistent with the other backends than better don't change it. However, maybe it still makes sense to remove the check and just invoke the function (if the return value isn't really telling us mach here). Maybe you can add a better check to see whether `save_snapshot` had actually the expected effect, e.g. whether `run_ssh_cmd` has been invoked as expected.

okurz (approved):

Martchus (approved):

PR #2051, opened 2022-05-10
VNC console: Test more ikvm encodings

Martchus:
See https://progress.opensuse.org/issues/107026


Martchus:
> I am really curious, how did you so quickly learn the ikvm specifics?

Maybe I should actually state that as a comment in the test code: These tests are written under the assumption that the current implementation is correct. So I'm merely feeding it data to hit all code branches and assume that what's happening is supposed to happen. (Well, I also do *some* sanity check, e.g. see my remark about possible crashes in the chat.)

mimi1vx (approved):

okurz (approved):

PR #2052, opened 2022-05-11
Add all missing function signatures except testapi

okurz:
Using the command

```
git grep -l 'sub [a-zA-Z0-9_]* {'
```

to find all missing places where we can and should use function
signatures as well.


tools/lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm
```diff
@@ -5,9 +5,9 @@ use Perl::Critic::Utils qw( :severities :classification :ppi );
 
 our $VERSION = '0.0.1';
 
-sub default_severity { return $SEVERITY_HIGH }
-sub default_themes { return qw(openqa) }
-sub applies_to { return qw(PPI::Token::Quote::Single PPI::Token::Quote::Double) }
+sub default_severity () { $SEVERITY_HIGH }
+sub default_themes (@) { qw(openqa) }
```
Martchus, line 9: These types of functions look like they're the same kind. So likely it makes sense to add `@` everywhere.

okurz: done

PR #2054, opened 2022-05-12
Hide string from logs in type_string when using secret=1

jlausuch:
Background: https://github.com/os-autoinst/os-autoinst/pull/2002

This PR is printing $string on the logs regardless if it's a secret or not.

https://progress.opensuse.org/issues/111010

This should avoid things like
```
[2022-05-12T15:59:08.330771+02:00] [debug] tests/jeos/firstrun.pm:116 called testapi::type_password
[2022-05-12T15:59:08.331044+02:00] [debug] <<< testapi::type_string(string="nots3cr3t", max_interval=100, wait_screen_changes=0, wait_still_screen=0, timeout=30, similarity_level=47, secret=1)
```

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -213,13 +213,13 @@ sub log_call (@args) {
     else {
         # key/value pairs
         my @result;
-        my $should_hide = grep { ($_ // '') eq 'secret' } @args;
+        my %args = @_;
         while (my ($key, $value) = splice(@args, 0, 2)) {
             if ($key =~ tr/0-9a-zA-Z_//c) {
                 # only quote if needed
                 $key = pp($key);
             }
-            $value = "[masked]" if ($key eq 'text' && $should_hide);
+            $value = "[masked]" if ($key =~ /string|text/ && $args{secret});
             push @result, join("=", $key, pp($value));
         }
         $params = join(", ", @result);
--- a/testapi.pm
+++ b/testapi.pm
@@ -1474,7 +1474,7 @@ sub type_string {
     my $wait_timeout = $args{timeout} // 30;
     my $wait_sim_level = $args{similarity_level} // 47;
     bmwqemu::log_call(string => $string, max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still,
-        timeout => $wait_timeout, similarity_level => $wait_sim_level);
+        timeout => $wait_timeout, similarity_level => $wait_sim_level, secret => $args{secret});
     my @pieces;
     if ($wait) {
         # split string into an array of pieces of specified size
--- a/t/12-bmwqemu.t
+++ b/t/12-bmwqemu.t
@@ -62,10 +62,15 @@ subtest 'log_call' => sub {
     }
     stderr_like(\&log_call_indent, qr{\Q<<< main::log_call_indent(test=[\E\n\Q    "a",\E\n\Q    [\E\n\Q      "b"\E\n\Q    ]\E\n\Q  ])}, 'log_call auto indentation');
 
-    sub log_call_test_secret {
+    sub log_call_test_secret_text {
         bmwqemu::log_call(text => "passwd\n", secret => 1);
     }
-    stderr_like(\&log_call_test_secret, qr{\Q<<< main::log_call_test_secret(text="[masked]", secret=1)}, 'log_call hides sensitive info');
+    stderr_like(\&log_call_test_secret_text, qr{\Q<<< main::log_call_test_secret_text(text="[masked]", secret=1)}, 'log_call hides sensitive info');
+
+    sub log_call_test_secret_string {
+        bmwqemu::log_call(string => "passwd\n", secret => 1);
+    }
+    stderr_like(\&log_call_test_secret_string, qr{\Q<<< main::log_call_test_secret_string(string="[masked]", secret=1)}, 'log_call hides sensitive info');
 };
 
 subtest 'update_line_number' => sub {


Martchus (commented):
It would be nice to extend unit tests but it looks generally good.

jlausuch:
> It would be nice to extend unit tests but it looks generally good.

Actually there are plenty of UT for `type_string` and `type_password`, I wonder why this wasn't caught before...

jlausuch:
@asmorodskyi @okurz ^

bmwqemu.pm
```diff
@@ -213,13 +213,13 @@ sub log_call (@args) {
     else {
         # key/value pairs
         my @result;
-        my $should_hide = grep { ($_ // '') eq 'secret' } @args;
+        my %args = @_;
         while (my ($key, $value) = splice(@args, 0, 2)) {
```
cfconrad, line 217:
```suggestion
        my %args = @_;
        while (my ($key, $value) = each %args) {
```

perlpunk: That would lose the original order, that might not be wanted. Additionally, it would be random. At least it should be sorted, so then I would suggest to use `sort keys %args`.
Maybe just keep it like it is.

bmwqemu.pm
```diff
@@ -213,13 +213,13 @@ sub log_call (@args) {
...
             if ($key =~ tr/0-9a-zA-Z_//c) {
                 # only quote if needed
                 $key = pp($key);
             }
-            $value = "[masked]" if ($key eq 'text' && $should_hide);
```
cfconrad, line 222: I would suggest not doing this here. This is something which should be done inside the caller.

But if we wanna keep it. it looks better then before :+1: .

testapi.pm
```diff
@@ -1474,7 +1474,7 @@ sub type_string {
     my $wait_timeout = $args{timeout} // 30;
     my $wait_sim_level = $args{similarity_level} // 47;
     bmwqemu::log_call(string => $string, max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still,
-        timeout => $wait_timeout, similarity_level => $wait_sim_level);
+        timeout => $wait_timeout, similarity_level => $wait_sim_level, secret => $args{secret});
```
cfconrad, line 1477:
```suggestion
bmwqemu::log_call(string => ($args{secret}?'[masked]' : $string), max_interval => $max_interval, wait_screen_changes => $wait, wait_still_screen => $wait_still,
        timeout => $wait_timeout, similarity_level => $wait_sim_level, secret => $args{secret});
```

bmwqemu.pm
cfconrad: hm, I didn't checked all `log_call()` calles. There might be some who use `text` as the key for secret value!
As I say, this is just bad code and should not be done in this function!

okurz: Why would you want to handle that in the caller and not here?

okurz (changes requested):
The good thing about our tests is that they show an effect of your change. See https://github.com/os-autoinst/os-autoinst/runs/6407456296?check_suite_focus=true#step:3:637

```
7: 
7:     #   Failed test 'log_call hides sensitive info'
7:     #   at ./12-bmwqemu.t line 68.
7:     # STDERR:
7:     # [2022-05-12T14:10:27.956610Z] [debug] <<< main::log_call_test_secret(text="passwd\n", secret=1)
7:     # 
7:     # doesn't match:
7:     # (?^u:\<\<\<\ main\:\:log_call_test_secret\(text\=\"\[masked\]\"\,\ secret\=1\))
7:     # as expected
7:     # Looks like you failed 1 test of 5.
7: 
7: #   Failed test 'log_call'
7: #   at ./12-bmwqemu.t line 69.
```

To reproduce you should be able to simply call `prove -I. t/12-bmwqemu.t`, then extend the test according to your expectations.

Likely you want to extend t/03-testapi.t if bmwqemu.pm itself is behaving fine but calling `log_call` incorrectly.

asmorodskyi:
is someone understands why #2002 stop working ? because I remember that I was certainly hiding secrets just after #2002 . From changes which Jose is doing I am suspecting that parameter which holding what we actually typing was renamed from "string" to "text" is that the case ?

bmwqemu.pm
jlausuch: It was originally changed here: https://github.com/os-autoinst/os-autoinst/pull/2002/files#diff-3420c7e01f73caee7d97053af39990e3297daf42d06e0fc8848f78af1a5e009eL1464
and it was accepted, I'm just correcting it

jlausuch:
> The good thing about our tests is that they show an effect of your change. See https://github.com/os-autoinst/os-autoinst/runs/6407456296?check_suite_focus=true#step:3:637
> 
> ```
> 7: 
> 7:     #   Failed test 'log_call hides sensitive info'
> 7:     #   at ./12-bmwqemu.t line 68.
> 7:     # STDERR:
> 7:     # [2022-05-12T14:10:27.956610Z] [debug] <<< main::log_call_test_secret(text="passwd\n", secret=1)
> 7:     # 
> 7:     # doesn't match:
> 7:     # (?^u:\<\<\<\ main\:\:log_call_test_secret\(text\=\"\[masked\]\"\,\ secret\=1\))
> 7:     # as expected
> 7:     # Looks like you failed 1 test of 5.
> 7: 
> 7: #   Failed test 'log_call'
> 7: #   at ./12-bmwqemu.t line 69.
> ```
> 
> To reproduce you should be able to simply call `prove -I. t/12-bmwqemu.t`, then extend the test according to your expectations.
> 
> Likely you want to extend t/03-testapi.t if bmwqemu.pm itself is behaving fine but calling `log_call` incorrectly.

jlausuch:
> is someone understands why #2002 stop working ? because I remember that it was certainly hiding secrets just after #2002 . From changes which Jose is doing I am suspecting that parameter which holding what we actually typing was renamed from "string" to "text" is that the case ?

The funny thing is that the only place where we are passing `text` as parameter is from the UT... The rest of the places use `string` as the parameter name. For instance, `type_password` use `string` :
```
type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
```
and I would say this is the most sensitive case where we should consider... I have added a regex for `text|string` in `log_call`... so we handle both parameters...

jlausuch:
BTW, if `type_password` is used with serial_terminal, the string will be printed in `serial_terminal.txt` anyway, so this method should be used only with root-console. Cause I guess there is no workaround for that case
```
    if (is_serial_terminal) {
        query_isotovideo('backend_type_string', {text => $string, %args});
        return;
    }
```

asmorodskyi:
> BTW, if `type_password` is used with serial_terminal, the string will be printed in `serial_terminal.txt` anyway, so this method should be used only with root-console. Cause I guess there is no workaround for that case

you can not state this , for example when you really entering password it won't be displayed . e.g 
```
#sudo -s                                                                                           
[sudo] password for root: 
```

asmorodskyi:
> > is someone understands why #2002 stop working ? because I remember that it was certainly hiding secrets just after #2002 . From changes which Jose is doing I am suspecting that parameter which holding what we actually typing was renamed from "string" to "text" is that the case ?
> 
> The funny thing is that the only place where we are passing `text` as parameter is from the UT... The rest of the places use `string` as the parameter name. For instance, `type_password` use `string` :
> 
> ```
> type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
> ```
> 
> and I would say this is the most sensitive case where we should consider... I have added a regex for `text|string` in `log_call`... so we handle both parameters...

if this is the case than I would assume single line https://github.com/os-autoinst/os-autoinst/pull/2054/files#diff-443ffbdc622ec9236041884826df475d69e3fa8ce40ec03c043c3ba66fae921bR222 would do the job and fix the problem . All other changes which comes together is a refactoring of existing approach and IMHO very questionable one . I really like initial approach suggested by Yannis the only missing piece in it is the fact that we have not only "text" but also "string" ...

jlausuch:
> > > is someone understands why #2002 stop working ? because I remember that it was certainly hiding secrets just after #2002 . From changes which Jose is doing I am suspecting that parameter which holding what we actually typing was renamed from "string" to "text" is that the case ?
> > 
> > 
> > The funny thing is that the only place where we are passing `text` as parameter is from the UT... The rest of the places use `string` as the parameter name. For instance, `type_password` use `string` :
> > ```
> > type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and I would say this is the most sensitive case where we should consider... I have added a regex for `text|string` in `log_call`... so we handle both parameters...
> 
> if this is the case than I would assume single line https://github.com/os-autoinst/os-autoinst/pull/2054/files#diff-443ffbdc622ec9236041884826df475d69e3fa8ce40ec03c043c3ba66fae921bR222 would do the job and fix the problem . All other changes which comes together is a refactoring of existing approach and IMHO very questionable one . I really like initial approach suggested by Yannis the only missing piece in it is the fact that we have not only "text" but also "string" ...

Not only that. 
In Yannis approach, `secret` was never an argument to `log_call`, so the condition he had [here](https://github.com/os-autoinst/os-autoinst/pull/2002/files#diff-443ffbdc622ec9236041884826df475d69e3fa8ce40ec03c043c3ba66fae921bR218) was never hit. This is what I'm trying to solve here.

jlausuch:
> > BTW, if `type_password` is used with serial_terminal, the string will be printed in `serial_terminal.txt` anyway, so this method should be used only with root-console. Cause I guess there is no workaround for that case
> 
> you can not state this , for example when you really entering password it won't be displayed . e.g
> 
> ```
> #sudo -s                                                                                           
> [sudo] password for root: 
> ```

yes, you are right, it was some misunderstanding of mine, please ignore this comment :)

asmorodskyi:
> > > > is someone understands why #2002 stop working ? because I remember that it was certainly hiding secrets just after #2002 . From changes which Jose is doing I am suspecting that parameter which holding what we actually typing was renamed from "string" to "text" is that the case ?
> > > 
> > > 
> > > The funny thing is that the only place where we are passing `text` as parameter is from the UT... The rest of the places use `string` as the parameter name. For instance, `type_password` use `string` :
> > > ```
> > > type_string $string, secret => 1, max_interval => ($args{max_interval} // 100), %args;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and I would say this is the most sensitive case where we should consider... I have added a regex for `text|string` in `log_call`... so we handle both parameters...
> > 
> > 
> > if this is the case than I would assume single line https://github.com/os-autoinst/os-autoinst/pull/2054/files#diff-443ffbdc622ec9236041884826df475d69e3fa8ce40ec03c043c3ba66fae921bR222 would do the job and fix the problem . All other changes which comes together is a refactoring of existing approach and IMHO very questionable one . I really like initial approach suggested by Yannis the only missing piece in it is the fact that we have not only "text" but also "string" ...
> 
> Not only that. In Yannis approach, `secret` was never an argument to `log_call`, so the condition he had [here](https://github.com/os-autoinst/os-autoinst/pull/2002/files#diff-443ffbdc622ec9236041884826df475d69e3fa8ce40ec03c043c3ba66fae921bR218) was never hit, because `type_string` was never passing any parameter `secret` to it. This is what I'm trying to solve here.

than how it is working here https://openqa.suse.de/tests/8736413/logfile?filename=autoinst-log.txt ? ( if you look for `[maksed]` you will find all occasions of `define_secret_variable` calls which calling `type_password` )

bmwqemu.pm
cfconrad: yes, I didn't thought about the order, good catch!

bmwqemu.pm
cfconrad: > Why would you want to handle that in the caller and not here?

Imagine code like this:
```perl
sub setup_secret {
    my (%args) = @_;
   bmwqemu::log_call(@_, secret => 1);
    
    my $my_super_secret_password_is = $args{psk};
}
```

I understand the `log_call()` function in that way, that it log parameters which are given. And there is just **no** format of them. 

A workaround for me could be something like:
```perl
sub setup_secret {
    my (%args) = @_;
   bmwqemu::log_call(@_, -mask => ['psk']);
    
    my $my_super_secret_password_is = $args{psk};
}
```
And hope that no one will use `-mask` as parameters for the calling function :).

jlausuch:
> than how it is working here https://openqa.suse.de/tests/8736413/logfile?filename=autoinst-log.txt ? ( if you look for `[maksed]` you will find all occasions of `define_secret_variable` calls which calling `type_password` )

Right. I don't know. But take a look here:

https://openqa.suse.de/tests/8405529/logfile?filename=autoinst-log.txt (2 months ago)
```
[2022-03-26T12:33:40.095557+01:00] [debug] tests/jeos/firstrun.pm:116 called testapi::type_password
[2022-03-26T12:33:40.095850+01:00] [debug] <<< testapi::type_string(string="SECRET STRING", max_interval=100, wait_screen_changes=0, wait_still_screen=0, timeout=30, similarity_level=47)
```

https://openqa.suse.de/tests/8738443/logfile?filename=autoinst-log.txt (recent)
```
[2022-05-12T06:09:58.666172+02:00] [debug] tests/jeos/firstrun.pm:116 called testapi::type_password
[2022-05-12T06:09:58.666561+02:00] [debug] <<< testapi::type_string(string="nots3cr3t", max_interval=100, wait_screen_changes=0, wait_still_screen=0, timeout=30, similarity_level=47)
```

asmorodskyi:
I don't like that we don't have clear understanding what is going on but trying to "fix" something. You showing me place where password is exposed and I showing another where it is hidden without understanding why it happening like this there is no way to say if your patch make sense or not . I was looking closer into "my" "positive case " and it looks really strange , there is NO logging of `type_password` **at all** : 
```
[2022-05-12T01:01:24.032685+02:00] [debug] <<< consoles::serial_screen::type_string(text="read -sp \"enter value: \" ARM_SUBSCRIPTION_ID", cmd="backend_type_string", json_cmd_token="wottTNYz")
[2022-05-12T01:01:24.032970+02:00] [debug] tests/publiccloud/prepare_instance.pm:68 called publiccloud::basetest::provider_factory -> lib/publiccloud/basetest.pm:99 called publiccloud::azure::init -> lib/publiccloud/azure.pm:29 called publiccloud::azure_client::init -> lib/publiccloud/azure_client.pm:31 called publiccloud::utils::define_secret_variable -> lib/publiccloud/utils.pm:208 called testapi::script_run
[2022-05-12T01:01:24.033092+02:00] [debug] <<< testapi::send_key(key="ret", wait_screen_change=0, do_wait=0)
[2022-05-12T01:01:24.033569+02:00] [debug] <<< consoles::serial_screen::type_string(key="ret", text="\n", json_cmd_token="KbwqPoRP", cmd="backend_send_key")
[2022-05-12T01:01:24.034206+02:00] [debug] <<< consoles::serial_screen::type_string(max_interval=100, secret=1, text="[masked]", cmd="backend_type_string", json_cmd_token="BfskvBUj")
[2022-05-12T01:01:24.034495+02:00] [debug] tests/publiccloud/prepare_instance.pm:68 called publiccloud::basetest::provider_factory -> lib/publiccloud/basetest.pm:99 called publiccloud::azure::init -> lib/publiccloud/azure.pm:29 called publiccloud::azure_client::init -> lib/publiccloud/azure_client.pm:31 called publiccloud::utils::define_secret_variable -> lib/publiccloud/utils.pm:210 called testapi::script_run
[2022-05-12T01:01:24.034650+02:00] [debug] <<< testapi::script_run(cmd="set +a", timeout=30, quiet=undef, output="", die_on_timeout=1)
[2022-05-12T01:01:24.034809+02:00] [debug] tests/publiccloud/prepare_instance.pm:68 called publiccloud::basetest::provider_factory -> lib/publiccloud/basetest.pm:99 called publiccloud::azure::init -> lib/publiccloud/azure.pm:29 called publiccloud::azure_client::init -> lib/publiccloud/azure_client.pm:31 called publiccloud::utils::define_secret_variable -> lib/publiccloud/utils.pm:210 called testapi::script_run
[2022-05-12T01:01:24.034966+02:00] [debug] <<< testapi::wait_serial(regexp="# ", record_output=undef, no_regex=1, expect_not_found=0, buffer_size=undef, timeout=90, quiet=undef)
[2022-05-12T01:01:24.035646+02:00] [debug] <<< consoles::serial_screen::read_until(buffer_size=undef, timeout=90, quiet=undef, regexp="# ", record_output=undef, json_cmd_token="twGmJvzQ", expect_not_found=0, no_regex=1, cmd="backend_wait_serial", pattern=[
    "# "
  ])
```

`type_password` is called at line 209 https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/utils.pm#L209 . Only what I notice that I am actually passing password as param and also adding `\n` at the end which actually not necessary and in all cases where password is exposed function called w/o parameters , but I don't know if it is important or not ...

bmwqemu.pm
cfconrad: A other idea is to mask by value and not by `key` like:
```perl
sub setup_secret {
    my (%args) = @_;
   bmwqemu::log_call(@_, -mask-value => [$args{psk}]);
    
    my $my_super_secret_password_is = $args{psk};
}
```
and the code snippet of `log_call()`
```perl
sub log_call {
    my $mask_value = [];
    my @args;
    while (my $v = shift @_) {
       if($v eq '-mask-value' && @_ ){
            $mask_value = shift @_;
            $mask_value = [$mask_value] unless ref($mask_value) eq 'ARRAY';
       } else {
            push @args, $v;
       }
    }

    foreach my $v (@args){
        foreach my $mask (@$mask) {
              $v =~ s/\Q$mask\E/[masked]/g;
        }
    }

    .... # Print out the arguements
}

cfconrad: @okurz @jlausuch wdyt?

okurz:
> > BTW, if `type_password` is used with serial_terminal, the string will be printed in `serial_terminal.txt` anyway, so this method should be used only with root-console. Cause I guess there is no workaround for that case
> 
> you can not state this , for example when you really entering password it won't be displayed . e.g
> 
> ```
> #sudo -s                                                                                           
> [sudo] password for root: 
> ```

Correct. Same as for the case of VNC consoles: If you use `type_password` to type the password literally into a shell prompt then the password will be visible in the screenshot.

bmwqemu.pm
jlausuch: What is the advantage of your proposal vs giving `secret => 1`? In both cases you need to tell `log_call` that the message is a secret in some way, either `secret =>` or `-maskvalue`...

cfconrad: The advantage is, that you tell `log_call` **what** is the secret and not just that there is one.

okurz: I am not exactly sure I understand your implementation proposal but it sounds like a better approach to explicitly pass a secret string around in an explicitly called variable rather than passing a string or text around and relying on an additional boolean variable to be correctly evaluated

jlausuch: > I am not exactly sure I understand your implementation proposal but it sounds like a better approach to explicitly pass a secret string around in an explicitly called variable rather than passing a string or text around and relying on an additional boolean variable to be correctly evaluated

Is this a question for me?

okurz: My comment was a response to @cfconrad. That implicitly includes a suggestion to you that you could try to implement it like that, if you can :)

jlausuch: Agreed with @cfconrad that he will provide a PR with his implementation so we can compare better an decide what works better for us.

cfconrad: here are my 5cent of this: https://github.com/os-autoinst/os-autoinst/pull/2062

okurz:
#2062 should replace this. Please bring in new changes with tests in case something is still missing

PR #2058, opened 2022-05-13
Simplify/improve error handling in `update_framebuffer`

Martchus:
* Avoid using try-catch so `bmwqemu::fctwarn` will log the actual function
  name
* Return whether there was a screen update or not in a direct/consistent
  way
* Use less verbose coding style


okurz:
> Great PR! Please pay attention to the following items before merging:
> 
> 
>     * [ ]  Consider running manual verification tests on real hardware, especially for non-qemu backends

I consider this important here.

Martchus:
> I consider this important here.

The code has unit tests and the fullstack tests run it as well.

okurz (approved):

PR #2061, opened 2022-05-16
Avoid backend from being stuck on broken VNC connections for too long

Martchus:
* Blocking reads/writes on the VNC socket can cause the backend process to
  hang (and therefore also the test driver to hang) leading to tests
  being stopped by the openQA worker after the overall timeout exceeded
* See https://progress.opensuse.org/issues/111004 for the observation of
  the problem
* Set VNC socket timeout so the backend would not block for too long
  similar to how we already limit the number of connection attempts

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -116,7 +116,7 @@ my @encodings = (
     },
 );
 
-sub login ($self, $connect_timeout = undef) {
+sub login ($self, $connect_timeout = undef, $timeout = undef) {
     # arbitrary
     my $connect_failure_limit = 2;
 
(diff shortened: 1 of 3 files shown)


Martchus:
I'll hot-patch this on openqaworker2. Let's see how the behavior is when it actually runs into the timeout. Maybe it makes sense to tweak the behavior in that case, e.g. add a reconnect.

Martchus:
My test job got nevertheless stuck (https://openqa.suse.de/tests/8765381). Well, at least it hasn't gotten worse. Maybe it hangs at a different place after all¹. I suppose this change is generally a good idea anyways.

---

¹ Unfortunately strace just shows

```
martchus@openqaworker2:~> sudo strace -p 12689
strace: Process 12689 attached
read(24, 
```

okurz (approved):

kalikiana (approved):

Martchus:
I've been updating the PR. I've already tested [Apply VNC timeout also for reading/writing](https://github.com/os-autoinst/os-autoinst/pull/2061/commits/d1adda78adc34c5ac02b5040a2bc0e97eaa83827) on openqaworker2 and it is already an improvement, see https://progress.opensuse.org/issues/111004#note-16. I'll test the last commit for portability on openqaworker2 later.

Martchus:
It seems to work fine on ow2: https://progress.opensuse.org/issues/111004#note-17

---

I've also been digging a little bit in old commits. Apparently it is a known problem that the timeout is only used on connect, see this "FIXME" comment: https://github.com/os-autoinst/os-autoinst/commit/304c5f2ffc49d8333692ee61abbca21166e4b2cd#diff-782670a09b7b934d82616dc71844a2ed8257ab95ba49e87c6c9cc1f5d3eb2dc4R94

So commit 2 and 3 from this PR are actually fixing this FIXME from 2015. (I suppose it would have helped me if we hadn't removed the FIXME from the code.)

By the way, the `$socket->sockopt(Socket::TCP_NODELAY, 1);` call that always fails with `EACCES` and therefore likely does nothing has been introduced in https://github.com/os-autoinst/os-autoinst/commit/967b15fa99e92b5971dd0c23a3c4858ab3cd20fe (see https://progress.opensuse.org/issues/111004#note-17 for the strace log). @coolo Do you remember whether this call ever worked? Maybe we can remove it considering that it isn't working anyways?

Martchus:
By the way, I'm still surprised that simply not hanging on reads helps already that much (without attempting to handle `EAGAIN` specifically). Note that in the strace logs one can clearly see that the actual problem (reads taking long) is still there and occurring several times during the test run. However, with this change the VNC client can move on which is apparently already good enough to let the test pass instead of timing out. Example (from test run https://openqa.suse.de/tests/8771207#live):

```
fcntl(25, F_GETFL)                      = 0x802 (flags O_RDWR|O_NONBLOCK)
fcntl(25, F_SETFL, O_RDWR)              = 0
read(25, "\326\277\257\177X\332\271|\276\366\254\365\253.o^X\274\323\274\370\272\262\374\376\321\17\343m\363~c"..., 8192) = 8192
read(25, "\261x\341fC\1}c\312\242\242HY7z\360\35\215\257\0056>\372>\335\216Ge\261\37\330\275"..., 8192) = 8192
read(25, "J1\327\353\357\1\225\362-\227\201\n+Us\330Q7\n\242\243\366J\3212|3\35\317/L@"..., 8192) = 8192
read(25, 0x56259a17e9e0, 8192)          = -1 EAGAIN (Die Ressource ist zur Zeit nicht verfügbar)
read(25, 
```

Martchus:
https://openqa.suse.de/tests/8771207 just passed.

PR #2062, opened 2022-05-18
bmwqemu: Hide secrets on log output with log_call()

cfconrad:
Parse given parameters for `-masked => <string>` and replace
all occurrence from the given string with `[masked]`.

-- 
Background: https://github.com/os-autoinst/os-autoinst/pull/2002 and https://github.com/os-autoinst/os-autoinst/pull/2054

https://progress.opensuse.org/issues/111010

This should avoid things like
```
[2022-05-12T15:59:08.330771+02:00] [debug] tests/jeos/firstrun.pm:116 called testapi::type_password
[2022-05-12T15:59:08.331044+02:00] [debug] <<< testapi::type_string(string="nots3cr3t", max_interval=100, wait_screen_changes=0, wait_still_screen=0, timeout=30, similarity_level=47, secret=1)
```

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -203,27 +203,49 @@ sub pp (@args) {
     return $value_with_trailing_newline;
 }
 
+# Use special argument `-masked` to hide the given value in log output.
+# It can be specified multiple times and or the value can be a ARRAY_REF or
+# scalar.
 sub log_call (@args) {
     my $fname = (caller(1))[3];
     update_line_number();
+
+    # extract -masked parameter out of argument list
+    my @masked;
+    my @effective_args;
+    while (@args) {
+        my $v = shift @args;
+        if (defined($v) && $v eq '-masked' && @args) {
+            my $mval = shift @args;
+            push @masked, ref($mval) eq 'ARRAY' ? @$mval : $mval;
+        } else {
+            push @effective_args, $v;
+        }
+    }
+
     my $params;
-    if (@args == 1) {
-        $params = pp($args[0]);
+    if (@effective_args == 1) {
+        $params = pp($effective_args[0]);
     }
     else {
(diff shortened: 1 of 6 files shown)


bmwqemu.pm
```diff
@@ -203,27 +203,50 @@ sub pp (@args) {
...
+    while (@args) {
+        my $v = shift @args;
+        if (defined($v) && $v eq '-masked' && @args) {
+            my $mval = shift @args;
+            $masked = [@$masked, ref($mval) eq 'ARRAY' ? @$mval : $mval];
+        } else {
+            push @tmpargs, $v;
+        }
+    }
+    @args = @tmpargs;
```
Martchus, line 225: It would be nice to avoid the copy here. Just use `@tmpargs` in subsequent code (and maybe give it a better name, e.g. `@effective_args`).

bmwqemu.pm
```diff
@@ -203,27 +203,50 @@ sub pp (@args) {
...
     update_line_number();
+
+    # extract -masked parameter out of argument list
+    my $masked = [];
+    my @tmpargs;
+    while (@args) {
+        my $v = shift @args;
+        if (defined($v) && $v eq '-masked' && @args) {
+            my $mval = shift @args;
+            $masked = [@$masked, ref($mval) eq 'ARRAY' ? @$mval : $mval];
```
Martchus, line 220:
```suggestion
            push @$masked, ref($mval) eq 'ARRAY' ? @$mval : $mval;
```

You could also make `$masked` an array (instead of an array ref) which would be more consistent with other local variables.

Martchus (changes requested):

Martchus (approved):

okurz (approved):
This looks more complicated than what I hoped but ok

PR #2066, opened 2022-05-20
Improve error feedback for non-existant CASEDIR

okurz:


perlpunk (approved):

bmwqemu.pm
```diff
@@ -157,7 +157,7 @@ sub ensure_valid_vars () {
     }
 
     die "CASEDIR variable not set, unknown test case directory" if !defined $vars{CASEDIR};
-    die "No scripts in $vars{CASEDIR}" if !-e "$vars{CASEDIR}";
+    die "No scripts in CASEDIR '$vars{CASEDIR}'" unless -e $vars{CASEDIR};
```
Martchus, line 160: Maybe suppress the line number at the end to make it look like a user-error:
```suggestion
    die "No scripts in CASEDIR '$vars{CASEDIR}\n'" unless -e $vars{CASEDIR};
```

okurz: Right. With the better error message we should not need the line anymore :D But the line can still provide some context. Well, ok. Changed.

Martchus (approved):

PR #2067, opened 2022-05-20
Move version replacement from spec file to build script

Martchus:
* Replace the version lookup via Git in CMake code instead of relying on
  the replacement in the spec file (in the same way we already replace the
  install directory)
* Provide a default so users won't end up by default with an installation
  that is complaining about not being within a Git directory
* Handle the case when Git is not installed or when the source directory is
  not a Git checkout with a fallback to "unknown" so current packaging
  scripts will continue to work

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,18 @@ add_custom_target(
 add_dependencies(symlinks videoencoder snd2png debugviewer tinycv)
 
 # create installable versions of isotovideo and cv.pm
+set(OS_AUTOINST_VERSION "" CACHE STRING "os-autoinst version to set for the installation")
+if (OS_AUTOINST_VERSION)
+    set(OS_AUTOINST_VERSION_REPLACE "${OS_AUTOINST_VERSION}")
+elseif (GIT_BIN AND IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+    find_program(GIT_BIN git)
+    execute_process(COMMAND "${GIT_BIN}" rev-parse HEAD
+                    WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                    OUTPUT_VARIABLE OS_AUTOINST_VERSION_REPLACE)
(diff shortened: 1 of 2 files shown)


Martchus:
Checked the package from the OBS checks and they look good. (Of course I also tested it locally.)

perlpunk (approved):

CMakeLists.txt
```diff
@@ -147,6 +147,18 @@ add_custom_target(
...
 # create installable versions of isotovideo and cv.pm
+set(OS_AUTOINST_VERSION "" CACHE STRING "os-autoinst version to set for the installation")
+if (OS_AUTOINST_VERSION)
+    set(OS_AUTINST_VERSION_REPLACE "${OS_AUTOINST_VERSION}")
+elseif (GIT_BIN AND IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+    find_program(GIT_BIN git)
+    execute_process(COMMAND "${GIT_BIN}" rev-parse HEAD
+                    WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                    OUTPUT_VARIABLE OS_AUTINST_VERSION_REPLACE)
+    string(REPLACE "\n" "" OS_AUTINST_VERSION_REPLACE "${OS_AUTINST_VERSION_REPLACE}")
```
okurz, line 158: That's actually not "moving the replacement to CMake", it's something additional. Why care about calling git dynamically here as well? That's a bit duplicating what isotovideo is doing internally. I understand that you want to cover the additional use case of "user installs from a git checkout and wants to have a valid version included" but do we need that?

Martchus: > Why care about calling git dynamically here as well?

We came to this after noticing bad user experience: A user ended up with an installation that complained not being within a Git directory. Just doing the replacement if `OS_AUTINST_VERSION` is defined (with no fallback) would not improve that. Note that `git rev-parse HEAD` is not too much logic in my opinion and we also don't *need* to keep this in sync with isotovideo itself. This is merely to have good enough fallback.

Martchus: I changed the commit message to make the real intention more clear.

okurz (approved):

PR #2071, opened 2022-05-30
Pass -x and -y as array arguments

perlpunk:
Before they were passed as a single string "-x 1 -y 2" which results
in videoencoder only receiving the -x correctly.

Followup to #2070

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -317,7 +317,7 @@ sub start_encoder ($self) {
     my $cwd = Cwd::getcwd;
     my @cmd = (qw(nice -n 19), "$bmwqemu::scriptdir/videoencoder", "$cwd/video.ogv");
     push(@cmd, '-n') if $bmwqemu::vars{NOVIDEO} || ($has_external_video_encoder_configured && !$bmwqemu::vars{EXTERNAL_VIDEO_ENCODER_ADDITIONALLY});
-    push(@cmd, "-x $self->{xres} -y $self->{yres}");
+    push @cmd, '-x', $self->{xres}, '-y', $self->{yres};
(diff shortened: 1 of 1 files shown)


AdamWill:
:+1:

Martchus (approved):

backend/baseclass.pm
```diff
@@ -317,7 +317,7 @@ sub start_encoder ($self) {
     my $cwd = Cwd::getcwd;
     my @cmd = (qw(nice -n 19), "$bmwqemu::scriptdir/videoencoder", "$cwd/video.ogv");
     push(@cmd, '-n') if $bmwqemu::vars{NOVIDEO} || ($has_external_video_encoder_configured && !$bmwqemu::vars{EXTERNAL_VIDEO_ENCODER_ADDITIONALLY});
-    push(@cmd, "-x $self->{xres} -y $self->{yres}");
+    push @cmd, '-x', $self->{xres}, '-y', $self->{yres};
```
kalikiana, line 320: I'm wondering why there is no failing test but also no warning about missing coverage 🤔

Martchus: > I'm wondering why there is no failing test

The line is surly covered by tests. I assume no test actually asserts whether the video resolution is correct (at least not if a custom resolution has been specified). And the video encoder uses C/C++ standard functions to do the number conversion so it simply stops after the first non-numeric character with no error. And the width works because the argument parser allows specifying option and values as one argument (e.g. `-x1024`).

AdamWill:
yeah, I think the original line likely gets hit during `23-baseclass.t`, but it doesn't look like any test checks the actual command line or the resolution used. So we'd only catch if the problem caused the encoder to not run at all, or something.

okurz (approved):

PR #2073, opened 2022-06-02
ci: Upload coverage HTML as an artifact

perlpunk:
Sometimes the codecov report is updated late, sometimes it disagrees with what
Devel::Cover reports.
And while codecov reports only line coverage, Devel::Cover reports statements,
which can sometimes help.

Usage: Click on the CI run "Summary" and download the "coverage.zip"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,13 @@ jobs:
           file: ./build/cover_db/codecov.json
           fail_ci_if_error: true
           verbose: true
+      - run: |
+          tar cvf coverage.tar build/cover_db
+      - uses: actions/upload-artifact@v3
+        with:
+          name: coverage
+          path: coverage.tar
+          retention-days: 15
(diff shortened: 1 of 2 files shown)


kraih (approved):
Nice, i would have thought those reports would be too large for artefacts.

kraih:
Might make sense to limit the retention time a bit: https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/

perlpunk:
> Might make sense to limit the retention time a bit: https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/

ah, I didn't know that, good idea. I set it to 30 days now

perlpunk:
Actually I think I will tar the directory first. Apparently every single file will result in a HTTP request.
That can result in rate limits and might also be slower.

PR #2074, opened 2022-06-02
Fix code coverage and run tests from root dir

perlpunk:
* Add `--workdir` option to isotovideo
* ci: Run tests from toplevel directory

`isotovideo --workdir /path/to/pool` will make isotovideo chdir into that directory at the beginning.

If we run isotovideo in a test from the os-autoinst directory with the `--workdir` option, then the module loading will happen in the os-autoinst directory, which results in Devel::Cover being able to count them as the same files as with the other tests.

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -30,9 +30,11 @@ sub isotovideo (%args) {
     $args{default_opts} //= 'backend=null';
     $args{opts} //= '';
     $args{exit_code} //= 1;
-    my @cmd = ($^X, "$toplevel_dir/isotovideo", '-d', $args{default_opts}, split(' ', $args{opts}));
+    chdir "$Bin/..";
+    my @cmd = ($^X, "$toplevel_dir/isotovideo", '--workdir', $pool_dir, '-d', $args{default_opts}, split(' ', $args{opts}));
+    chdir $pool_dir;
     note "Starting isotovideo with: @cmd";
-    my $output = qx(@cmd);
+    my $output = qx(cd $toplevel_dir && @cmd);
     my $res = $?;
     return fail 'failed to execute isotovideo: ' . $! if $res == -1;    # uncoverable statement
     return fail 'isotovideo died with signal ' . ($res & 127) if $res & 127;    # uncoverable statement
@@ -42,9 +44,9 @@ sub isotovideo (%args) {
 }
 
 subtest 'get the version number' => sub {
-    # Make sure we're in a folder we can't write to, no base_state.json should be created here
-    chdir('/');
-    combined_like { system $^X, "$toplevel_dir/isotovideo", '--version' } qr/Current version is.+\[interface v[0-9]+\]/, 'version printed';
+    chdir "$Bin/..";
+    combined_like { system $^X, "$toplevel_dir/isotovideo", '--workdir', $pool_dir, '--version' } qr/Current version is.+\[interface v[0-9]+\]/, 'version printed';
+    chdir $pool_dir;
     ok(!-e bmwqemu::STATE_FILE, 'no state file was written');
 };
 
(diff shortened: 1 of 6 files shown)


perlpunk:
The overall coverage dropped because of new files that were not listed before.
Here's a vimdiff screenshot:
coverage-fixed
I think the drop comes mostly from `os-autoinst-openvswitch` which has only 25.7% coverage.

There are still some warnings left, though:
```
Devel::Cover: merging data for ppmclibs/blib/lib/tinycv.pm into t/../ppmclibs/blib/lib/tinycv.pm
Devel::Cover: merging data for /opt/backend/driver.pm into backend/driver.pm
Devel::Cover: merging data for /opt/isotovideo into isotovideo
Devel::Cover: merging data for /opt/OpenQA/Isotovideo/CommandHandler.pm into OpenQA/Isotovideo/CommandHandler.pm
Devel::Cover: Can't find digest for OpenQA/Isotovideo/NeedleDownloader.pm
Devel::Cover: Can't find digest for distribution.pm
Devel::Cover: Can't find digest for commands.pm
Devel::Cover: Can't find digest for log.pm
Devel::Cover: Can't find digest for bmwqemu.pm
Devel::Cover: Can't find digest for OpenQA/Isotovideo/Utils.pm
Devel::Cover: Can't find digest for myjsonrpc.pm
Devel::Cover: Can't find digest for signalblocker.pm
Devel::Cover: Can't find digest for autotest.pm
Devel::Cover: Can't find digest for backend/console_proxy.pm
Devel::Cover: Can't find digest for cv.pm
Devel::Cover: Can't find digest for testapi.pm
Devel::Cover: Can't find digest for OpenQA/Exceptions.pm
Devel::Cover: Can't find digest for needle.pm
Devel::Cover: Can't find digest for OpenQA/Benchmark/Stopwatch.pm
Devel::Cover: Can't find digest for OpenQA/Isotovideo/Interface.pm
```
so there might be other tests where commands are run from a different directory.
But I think it's already more correct than before.

perlpunk:
I updated 14-isotovideo.t and updated the screenshot above. Most warnings are gone.

t/14-isotovideo.t
```diff
@@ -30,9 +30,11 @@ sub isotovideo (%args) {
     $args{default_opts} //= 'backend=null';
     $args{opts} //= '';
     $args{exit_code} //= 1;
-    my @cmd = ($^X, "$toplevel_dir/isotovideo", '-d', $args{default_opts}, split(' ', $args{opts}));
+    chdir "$Bin/..";
+    my @cmd = ($^X, "$toplevel_dir/isotovideo", '--workdir', $pool_dir, '-d', $args{default_opts}, split(' ', $args{opts}));
+    chdir $pool_dir;
```
okurz, line 35: Hm, the explanation makes sense how we end up with coverage files referring to differing directories but now it looks like we need to switch back and forth between different directories. Do you think it would be possible to somehow refer to all coverage files by absolute paths instead?
So instead of trying to ensure that we run everything from the os-autoinst root directory do you think it would be possible if we run everything from an external temporary directory so that Devel:Cover would relate to files by absolute paths?

perlpunk: I also thought about that but not sure it will help much.

The back and forth is mostly necessary because the tests are working with files created by the isotovideo call (e.g. `autoinst-log.txt`), but they are referring to those files with a relative path.
So the easiest thing for me was to switch back to the pool dir, instead of changing all the file paths in the tests. The latter would be better IMHO, but it is a bit of work.

okurz: Ok, thanks, I see. Well, we can still look into the alternative later on. The important thing is that you found out the original problem and found a fix

perlpunk:
Ok, I used `Cwd::realpath()` in `cv.pm`, and now also the "warning" for `ppmclibs/blib/lib/tinycv.pm` is gone.

kalikiana (approved):

Martchus (commented):
Nice to see that this has finally been fixed!

PR #2077, opened 2022-06-09
Fix VNC console tests under big-endian systems such as s390x

Martchus:
See https://progress.opensuse.org/issues/111608

---

@AdamWill I don't have a real big-endian system to test. Maybe you can test it within your builds. However, when I set the endian conversion incorrectly, I can reproduce the exact same issue you're seeing on s390x on x86_64. So I'm confident this change makes a difference (and works).

### DIFF
--- a/t/27-consoles-vnc.t
+++ b/t/27-consoles-vnc.t
@@ -39,6 +39,10 @@ is $c->_receive_bell, 1, 'can call _receive_bell';
 is_deeply \@printed, ['RFB 003.006', pack('C', 1)], 'protocol version and security type replied' or diag explain \@printed;
 @printed = ();
 
+# ensure endian conversion is setup correctly (despite initially mocking _server_initialization)
+my $machine_is_big_endian = unpack('h*', pack('s', 1)) =~ /01/ ? 1 : 0;
+$c->_do_endian_conversion($machine_is_big_endian);
+
 subtest 'send update request' => sub {
     $c->width(1024);
     $c->height(512);


baierjan (approved):

okurz:
> See https://progress.opensuse.org/issues/111608
> 
> @AdamWill I don't have a real big-endian system to test.

Well, you have OBS :)

okurz (approved):

AdamWill:
Unfortunately this does not seem to help :( I did a build with this patch applied and my downstream patch to disable the broken checks removed, and the checks fail again in just the same way:

    2:     #   Failed test 'pixel data updated in framebuffer (blue)'
    2:     #   at ./27-consoles-vnc.t line 171.
    2:     #          got: '37'
    2:     #     expected: '41'
    2: 
    2:     #   Failed test 'pixel data updated in framebuffer (green)'
    2:     #   at ./27-consoles-vnc.t line 172.
    2:     #          got: '41'
    2:     #     expected: '37'
    2: 
    2:     #   Failed test 'pixel data updated in framebuffer (red)'
    2:     #   at ./27-consoles-vnc.t line 173.
    2:     #          got: '0'
    2:     #     expected: '31'
    2:     # Looks like you failed 3 tests of 24.
    2: 
    2: #   Failed test 'update framebuffer'
    2: #   at ./27-consoles-vnc.t line 229.

Not sure why it's not working, but...it doesn't seem to work :(

AdamWill:
I added a couple of `diag` lines for testing, one right where we determine `$machine_is_big_endian` and one just before the setting of `my $vncinfo` in the 'update framebuffer' subtest, to check what `$c->_do_endian_conversion` is set to at that moment. On s390x they show this:

2: # machine is big endian: 1
2:     # _do_endian_conversion is: 1

on x86_64 they show this:

2: # machine is big endian: 0
2:     # _do_endian_conversion is: 0

so the test 'works', and the flag is properly set in 'update framebuffer' subtest...but somehow still the test fails...

AdamWill:
I also tested the same diags but with the setting of `$c->_do_endian_conversion($machine_is_big_endian);` commented out. With that change, we get:

2: # machine is big endian: 1
2: Use of uninitialized value in concatenation (.) or string at ./27-consoles-vnc.t line 165.

i.e., without the setting of it that this PR added, it's uninitialized during 'update framebuffer'. So it's not like it was already being set to 1 somehow and the PR "didn't do anything"; the PR definitely "does something", but somehow it doesn't fix the test...

AdamWill:
yeah, it's very odd. I've really no idea why it doesn't work :(

PR #2078, opened 2022-06-09
Use signatures in most of testapi (test covered changes)

okurz:
* Simplify testapi::check_var
* Use proper perl "signatures" in most of testapi
* t: Add test for testapi::x11_start_program
* t: Add test for set_var, get_var_array, check_var_array
* t: Add test for testapi::force_soft_failure
* t: Add test for testapi::init

I created this PR out of
* #1696

including only the changes that are covered by automatic tests.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -777,8 +777,7 @@ Returns true if test variable C<$variable> is equal to C<$value> or returns C<un
 =cut
 
 sub check_var ($var, $val) {
-    return 1 if (defined $bmwqemu::vars{$var} && $bmwqemu::vars{$var} eq $val);
-    return 0;
+    return defined $bmwqemu::vars{$var} && $bmwqemu::vars{$var} eq $val;
 }
 
 =head2 get_var_array
--- a/testapi.pm
+++ b/testapi.pm
@@ -107,12 +107,9 @@ to be used in tests.
 
 =cut
 
-sub _calculate_clickpoint {
-    my ($needle_to_use, $needle_area, $click_point) = @_;
+sub _calculate_clickpoint ($needle_to_use, $needle_area = undef, $click_point = undef) {
     # If there is no needle area defined, take it from the needle itself.
-    if (!$needle_area) {
-        $needle_area = $needle_to_use->{area}->[-1];
-    }
+    $needle_area ||= $needle_to_use->{area}->[-1];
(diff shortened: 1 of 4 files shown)


Martchus (approved):

Martchus:
The style check is still failing, though.

okurz:
> The style check is still failing, though.

fixed

testapi.pm
```diff
@@ -197,23 +192,21 @@ influenced.
...
     bmwqemu::log_call(reason => $reason);
 
     $autotest::current_test->record_soft_failure_result($reason);
 }
 
-sub _is_valid_result {
-    my ($result) = @_;
+sub _is_valid_result ($result) {
     return $result =~ /^(ok|fail|softfail)$/;
 }
```
kalikiana, line 203:
```suggestion
sub _is_valid_result ($result) { $result =~ /^(ok|fail|softfail)$/ }
```

testapi.pm
```diff
@@ -477,11 +463,8 @@ been matched at the time of the call.
...
 =cut
 
-sub match_has_tag {
-    my ($tag) = @_;
-    if ($last_matched_needle) {
-        return $last_matched_needle->{needle}->has_tag($tag);
-    }
+sub match_has_tag ($tag) {
+    return $last_matched_needle->{needle}->has_tag($tag) if $last_matched_needle;
     return;
```
kalikiana, line 468:
```suggestion
    return undef;
```

Might as well fix the return value here. We should always return explicitly

testapi.pm
okurz: I didn't want to change something unrelated to signatures

okurz: ok, changed

testapi.pm

kalikiana (approved):

testapi.pm
```diff
@@ -1559,9 +1530,9 @@ Default hold time is 0.15s
 
 =cut
 
-sub mouse_click {
-    my $button = shift || 'left';
-    my $time = shift || 0.15;
+sub mouse_click ($button = undef, $time = undef) {
+    $button //= 'left';
+    $time //= 0.15;
```
perlpunk, line 1529: Strictly, if `$time` was 0, then it would have been set to `0.15` in the original code. Now only if it's undef.

perlpunk: ok, I just wanted to mention it.

PR #2082, opened 2022-06-09
qemu: Treat qxl as an unsupported backend

kalikiana:
- Setting QEMUVGA to qxl is considered unsupported.
- Add unit test coverage for -vga.

See: https://progress.opensuse.org/issues/111992

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -503,7 +503,7 @@ sub delete_virtio_console_fifo () { unlink $_ or bmwqemu::fctwarn("Could not unl
 
 sub qemu_params_ofw ($self) {
     my $vars = \%bmwqemu::vars;
-    $vars->{QEMUVGA} ||= "std";
+    $vars->{QEMUVGA} ||= 'std';
     $vars->{QEMUMACHINE} //= "usb=off";
     sp('g', '1024x768');
     # newer qemu needs safe cache capability level quirk settings
@@ -681,7 +681,9 @@ sub start_qemu ($self) {
     elsif ($vars->{OFW}) {
         $use_usb_kbd = $self->qemu_params_ofw;
     }
-    sp('vga', $vars->{QEMUVGA}) if $vars->{QEMUVGA};
+    # qxl hard-codes the default resolution, see poo#111992
+    die "qxl is unsupported\n" if ($vars->{QEMUVGA} // '') eq 'qxl';
+    sp('vga', $vars->{QEMUVGA});
 
     my @nicmac;
     my @nicvlan;
--- a/t/18-backend-qemu.t
+++ b/t/18-backend-qemu.t
@@ -192,4 +192,12 @@ subtest qemu_tpm_option => sub {
     like $runcmd, qr|swtpm socket --tpmstate dir=.*mytpm6 --ctrl type=unixio,path=.*mytpm6/swtpm-sock --log level=20 -d|, 'swtpm 1.2 device created';
 };
 
+subtest qemu_vga_option => sub {
+    my $runcmd;
+    $backend_mock->redefine(runcmd => sub (@cmd) { $runcmd = join(' ', @cmd) });
+    unlike qemu_cmdline(QEMUVGA => undef), qr|-vga|, 'nothing specified by default';
+    like qemu_cmdline(QEMUVGA => 'std'), qr|-vga std|, 'std selected explicitly';
+    throws_ok { qemu_cmdline(QEMUVGA => 'qxl') } qr/qxl is unsupported/, 'unsupported vga aborts execution';
+};
+
 done_testing();


Martchus:
So there was no way to make qxl use the right resolution?

backend/qemu.pm
```diff
@@ -681,7 +680,9 @@ sub start_qemu ($self) {
     elsif ($vars->{OFW}) {
         $use_usb_kbd = $self->qemu_params_ofw;
     }
-    sp('vga', $vars->{QEMUVGA}) if $vars->{QEMUVGA};
+    $vars->{QEMUVGA} ||= 'std';
```
okurz, line 683: Please don't set that. Some time ago I have explicitly removed the default to leave it to qemu to select a default. That fixed some problems we had

okurz (changes requested):

kalikiana:
> So there was no way to make qxl use the right resolution?

The only way I'm aware of is by interactively going through the menu.

okurz:
> > So there was no way to make qxl use the right resolution?
> 
> The only way I'm aware of is by interactively going through the menu.

That's what we discussed, yes. I suggest to do this then.

kalikiana:
> > The only way I'm aware of is by interactively going through the menu.
> 
> That's what we discussed, yes. I suggest to do this then.

I wasn't planning on looking into that because it requires steps to be executed in the test rather than backend configuration. The PR fulfills the AC because `qxl is not used anymore`.

okurz:
> The PR fulfills the AC because `qxl is not used anymore`.

Yes, it would. But I will only merge this if test maintainers can confirm that they can live without qxl.

@Vogtinator do you think we can switch to alternative QEMUVGA selections in tests or do you consider selecting a resolution within the OVMF UEFI menu – just like the boot device – a better choice?

Vogtinator:
IMO just not supporting `qxl` is not a valid solution. It is currently used in some tests and we'd lose some coverage there and at the very least would have to modify the windows images to be able to switch those tests to virtio instead.

It's also not just `qxl` that's affected by the default resolution change. It's all video devices, except for a few exceptions like `cirrus`.

What we could try to do is perform some OVMF configuration manually and then use the resulting vars flash file in tests. This would probably have to be repeated for every OVMF update though.

Martchus:
Let me try convincing `qxl` to use our default resolution via some qemu params.

Martchus:
If I just clone https://openqa.opensuse.org/tests/2400280 with `QEMUVGA=qxl` everything works as expected. So that's not good enough for reproducing the issue as it is likely additionally also OVMF-specific. If I clone https://openqa.opensuse.org/tests/2400506 (the job from the issue) then it fails locally with `runcmd '/usr/bin/qemu-img create -f qcow2 -F qcow2 -b /hdd/openqa-devel/openqa/pool/1/windows-10-x86_64-21H1@uefi_win.qcow2 /hdd/openqa-devel/openqa/pool/1/raid/hd0-overlay0 110080' failed with exit code 1: 'qemu-img: /hdd/openqa-devel/openqa/pool/1/raid/hd0-overlay0: Image is not in qcow2 format`. I still need to investigate why that's the case.

Vogtinator:
Yes, the OVMF resolution should mostly impact OVMF (shell and menu) as well as bootloaders. The kernel will also inherit this resolution until the "real" driver is loaded and new configuration applied.

Martchus (dismissed):
Looks like `-device qxl,xres=1024,yres=768` works just fine. Let me create a PR. If it is acceptable I'd close this one.

Vogtinator:
> Looks like `-device qxl,xres=1024,yres=768` works just fine. Let me create a PR. If it is acceptable I'd close this one.

Did you also use `-vga none`? If not, you unintentionally got both `-vga std` and the qxl device attached, with vga std being the default device.

backend/qemu.pm
```diff
@@ -681,7 +681,9 @@ sub start_qemu ($self) {
     elsif ($vars->{OFW}) {
         $use_usb_kbd = $self->qemu_params_ofw;
     }
-    sp('vga', $vars->{QEMUVGA}) if $vars->{QEMUVGA};
+    # qxl hard-codes the default resolution, see poo#111992
+    die "qxl is unsupported\n" if ($vars->{QEMUVGA} // '') eq 'qxl';
```
Martchus, line 685: A fatal `die` would overshoot the goal. "Unsupported" doesn't mean "not allowed at all". As shown in https://github.com/os-autoinst/os-autoinst/pull/2082#issuecomment-1152199464 it is also not *that* bad, e.g. if one would get past the bootloader it would actually work.

backend/qemu.pm
```diff
@@ -681,7 +681,9 @@ sub start_qemu ($self) {
...
+    sp('vga', $vars->{QEMUVGA});
```
Martchus, line 686: `QEMUVGA` might be `undef` here.

Martchus (changes requested):

backend/qemu.pm
kalikiana: Well, we don't know what's installed and wether the test has needles for the right resolution.

PR #2083, opened 2022-06-10
Enforce resolution specified via XRES/YRES in all QEMU video configs

Martchus:
* Use `-device` option with `xres` and `yres` parameters to specify the
  video device (instead of using `-vga` which supposedly doesn't allow
  specifying these parameters)
    * Don't rely on QEMU's default for `xres`/`yres` to match our defaults
    * Make `XRES`/`YRES` parameters introduced by
      https://github.com/os-autoinst/os-autoinst/commit/cfcb4962e65c55225d47221ee8121d35b5990575 work for non-ARM
      architectures as well
* Move variable for ARM-specific variable `QEMU_VIDEO_DEVICE` into the
  right scope
* Fix https://progress.opensuse.org/issues/111992

---

I also added an additional commit to update the documentation which is outdated since https://github.com/os-autoinst/os-autoinst/commit/bca214ee1dfc257c694cf1b4650dec0fa7d984a1.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -671,18 +671,19 @@ sub start_qemu ($self) {
     my $use_usb_kbd;
     my $arch = $vars->{ARCH} // '';
     $arch = 'arm' if ($arch =~ /armv6|armv7/);
-    my $custom_video_device = $vars->{QEMU_VIDEO_DEVICE} // 'virtio-gpu-pci';
 
     if ($arch eq 'aarch64' || $arch eq 'arm') {
+        my $custom_video_device = $vars->{QEMU_VIDEO_DEVICE} // 'virtio-gpu-pci';
         my $video_device = ($vars->{QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64}) ? 'VGA' : "${custom_video_device},xres=$self->{xres},yres=$self->{yres}";
         sp('device', $video_device);
         $arch_supports_boot_order = 0;
         $use_usb_kbd = 1;
     }
-    elsif ($vars->{OFW}) {
-        $use_usb_kbd = $self->qemu_params_ofw;
+    else {
+        $use_usb_kbd = $self->qemu_params_ofw if $vars->{OFW};
+        $vars->{QEMUVGA} //= 'std';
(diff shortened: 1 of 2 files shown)


Martchus:
By the way, I've tested that locally via `script/openqa-clone-job … https://openqa.opensuse.org/tests/2400280 QEMUVGA=qxl UEFI=1 UEFI_PFLASH_CODE=/usr/share/qemu/ovmf-x86_64-ms-code.bin UEFI_PFLASH_VARS=/usr/share/qemu/ovmf-x86_64-ms-vars.bin`. I'll have to do some more tests with other `QEMUVGA` values to be sure it doesn't break everything else.

Vogtinator:
See https://github.com/os-autoinst/os-autoinst/pull/2082#issuecomment-1152262349. I don't think this does what you expect, I fell into that trap as well.

backend/qemu.pm
```diff
@@ -671,18 +671,19 @@ sub start_qemu ($self) {
...
         my $video_device = ($vars->{QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64}) ? 'VGA' : "${custom_video_device},xres=$self->{xres},yres=$self->{yres}";
         sp('device', $video_device);
         $arch_supports_boot_order = 0;
         $use_usb_kbd = 1;
     }
-    elsif ($vars->{OFW}) {
-        $use_usb_kbd = $self->qemu_params_ofw;
+    else {
+        $use_usb_kbd = $self->qemu_params_ofw if $vars->{OFW};
+        $vars->{QEMUVGA} //= 'std';
```
kalikiana, line 684: Probably shouldn't do that here, see https://github.com/os-autoinst/os-autoinst/pull/2082#discussion_r893874084

kalikiana: Well, I don't know what the original issue was. We could avoid setting the variable, but *device* might only work with an explicit argument.

Vogtinator:
Yep, that matches my experiments as well...

Martchus:
And btw, if I use `-g` I get `qemu-system-x86_64: -g 1024x768: Option not supported for this target`. So that's definitely ppc-only.

PR #2088, opened 2022-06-15
Fix HTML Devel::Cover report

perlpunk:
There seems to be a problem with html_basic, and html_minimal does actually
not look too minimal for me.

Related issue: https://progress.opensuse.org/issues/111251


okurz (approved):
> There seems to be a problem with html_basic, and html_minimal does actually
not look too minimal for me.

Hm, is there any kind of "upstream issue" that catches this as well?

mimi1vx (approved):

PR #2089, opened 2022-06-16
[WIP] qemu: Add aio option

ggardet:


ggardet:
`aio` option works with `-drive` not with `-device`. This results in `Property 'virtio-blk-device.aio' not found`.

PR #2091, opened 2022-06-17
Add code to request VMWare VNC access over web sockets and use it

Martchus:
* See particular commit messages and https://progress.opensuse.org/issues/113201
* ~~See commit on https://github.com/Martchus/os-autoinst-distri-opensuse/commits/svirt-vmware-vnc-cfg/lib/susedistribution.pm for required changed to the test distribution~~
    * ~~So this change has only an effect if `VMWARE_VNC_OVER_WS` is set to a truthy value (otherwise the VNC console is used as-is, without attempting to use WebSockets).~~
* One only needs to set `VMWARE_VNC_OVER_WS=1` (as documented in `backend_vars.md`) to use it

### DIFF
--- a/consoles/VMWare.pm
+++ b/consoles/VMWare.pm
@@ -110,8 +110,9 @@ sub launch_vnc_server ($self, $listen_port) {
     }
 }
 
-sub deduce_url_from_vars () {
+sub deduce_url_from_vars ($vnc_console) {
     return undef unless $bmwqemu::vars{VMWARE_VNC_OVER_WS};
+    return undef unless $vnc_console->hostname eq ($bmwqemu::vars{VIRSH_GUEST} // '');
     my $host = $bmwqemu::vars{VMWARE_HOST} or die "VMWARE_VNC_OVER_WS set but not VMWARE_HOST\n";
     my $user = $bmwqemu::vars{VMWARE_USERNAME} // 'root';
     my $password = $bmwqemu::vars{VMWARE_PASSWORD} or die "VMWARE_VNC_OVER_WS set but not VMWARE_PASSWORD\n";
@@ -119,7 +120,7 @@ sub deduce_url_from_vars () {
 }
 
 sub setup_for_vnc_console ($vnc_console) {
-    return undef unless my $ws_url = $vnc_console->vmware_vnc_over_ws_url // deduce_url_from_vars;
+    return undef unless my $ws_url = $vnc_console->vmware_vnc_over_ws_url // deduce_url_from_vars($vnc_console);
     my $self = $vnc_console->{_vmware_handler} //= consoles::VMWare->new;
     log::diag 'Establishing VNC connection over WebSockets via ' . Mojo::URL->new($ws_url)->to_string;
     $vnc_console->hostname('127.0.0.1');
--- a/consoles/VMWare.pm
+++ b/consoles/VMWare.pm
@@ -10,11 +10,13 @@ use Mojo::JSON qw(encode_json);
 use Mojo::UserAgent;
 use Mojo::URL;
 use Mojo::Util qw(xml_escape);
+
+use bmwqemu;
 use log;
 
 has protocol => 'https';
 has host => undef;
-has vm_id => 1;
+has vm_id => undef;
 has username => 'root';
 has password => undef;
 has dewebsockify_pid => undef;
@@ -38,7 +40,7 @@ sub configure_from_url ($self, $url) {
     $url = Mojo::URL->new($url);
     $self->protocol($url->protocol)->host($url->host);
     $self->username($url->username)->password($url->password);
-    $self->vm_id(substr $url->path, 1);
+    $self->vm_id(substr($url->path, 1)) if length $url->path > 1;
 }
 
 sub get_vmware_wss_url ($self) {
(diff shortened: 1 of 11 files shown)


okurz (changes requested):

consoles/VNC.pm
```diff
@@ -116,6 +118,16 @@ my @encodings = (
 );
 
 sub login ($self, $connect_timeout = undef, $timeout = undef) {
+    my $port = $self->port || 5900;
```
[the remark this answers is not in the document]
okurz: I mean the whole block

consoles/VNC.pm
```diff
@@ -12,11 +12,13 @@ use Carp qw(confess cluck carp croak);
 use Data::Dumper 'Dumper';
 use Scalar::Util 'blessed';
 use OpenQA::Exceptions;
+use consoles::VMWare;
```
okurz, line 15: Hm. I would prefer if we can avoid the VNC console to depend on the VMWare console. Can we use specific VMWare consoles that could inherit from VNC?

Martchus: I also thought about using inheritance. It would have been more work for nothing but if things generally work I can do that if it is required. Not sure whether it is the better design (if in doubt, I'd say https://en.wikipedia.org/wiki/Composition_over_inheritance).

Note that my initial thinking was that this is not a special console at all from the side of tests and the VNC console would figure out things on its own. And in my opinion it would still be nice if I could move code from https://github.com/Martchus/os-autoinst-distri-opensuse/commits/svirt-vmware-vnc-cfg/lib/susedistribution.pm to os-autoinst so all the VMWare-specific handling is at one place and this doesn't become a half test distribution specific feature (which is weirdly documented as backend variable here, despite the variable being used in the test distro). However, moving that code to os-autoinst means that from tests it shouldn't make a difference which VNC console to instantiate. So my initial idea of letting the normal VNC console handle it under the hood would actually still make sense. Note that `VMWare.pm` is conceptually not really its own console but just a helper for making the normal VNC console work with VMWare's VNC over WebSockets.

okurz: Yes, I understood that. Well, ok, so that means that consoles::VNC is both the generic VNC console as well as the module combining different implementations I guess

Martchus: > so that means that consoles::VNC is both the generic VNC console

Not really because the generic VNC console is actually `consoles::vnc_base`. `consoles::VNC` is also just some module internally used by that console. Apparently our convention is to use lower case names for actual consoles and upper case names for "console helpers". Of course that's just a technicality. But that's my understanding of our conventions and that's also why I called the new module `consoles::VMWare`.

Martchus:
This was tested by a colleague: http://10.67.129.66/tests/1342#

So despite the failure the run verifies that it generally works. The specialty of the test run is actually only visible in the logs:
```
[2022-07-14T23:37:50.136903+08:00] [debug] Establishing VNC connection to 10.67.131.2:5901
[2022-07-14T23:37:50.138361+08:00] [debug] Establishing VNC connection over WebSockets via https://10.67.131.2/
[2022-07-14 23:37:51.25745] [22986] [info] Establishing WebSocket connection to wss://vh002.qa2.suse.asia:443/ticket/e160cba9321bf71a
[2022-07-14 23:37:51.25862] [22986] [info] Client accepted
[2022-07-14 23:37:51.27830] [22986] [info] WebSocket connection established
```

okurz (approved):

Martchus:
The style check fails now at `No backend or console files use external facing testapi` because I'm using `$testapi::password`. That shouldn't be a failure. I suppose we can live without the fallback so I'm removing the use. However, I don't like that the style check is forbidding that.

consoles/VMWare.pm
```diff
@@ -0,0 +1,132 @@
...
+    my $cookie = $request_wss_txn->req->cookies->[0];
+    die "VMWare did not return a session cookie\n" unless $cookie;
+    return (Mojo::URL->new($wss_url->text), $cookie);
+}
+
+sub _cleanup_previous_dewebsockify_process ($self) {
+    return undef unless my $pid = $self->dewebsockify_pid;
+    kill SIGTERM => $pid;
+    waitpid $pid, 0;
+    $self->dewebsockify_pid(undef);
```
okurz, line 89: I suggest to cover these lines with tests *or* mark as uncoverable. There is no better time than now to do it properly :)

okurz (changes requested):

consoles/VMWare.pm
```diff
@@ -0,0 +1,137 @@
...
+    my $vm_id = xml_escape($self->vm_id || $bmwqemu::vars{VIRSH_VM_ID});
+    my $auth_xml = qq{<Envelope xmlns="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><Header><operationID>esxui-8fb8</operationID></Header><Body><Login xmlns="urn:vim25"><_this type="SessionManager">ha-sessionmgr</_this><userName>$username</userName><password>$password</password></Login></Body></Envelope>};
+    my $request_wss_xml = qq{<Envelope xmlns="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><Header><operationID>esxui-c51d</operationID></Header><Body><AcquireTicket xmlns="urn:vim25"><_this type="VirtualMachine">$vm_id</_this><ticketType>webmks</ticketType></AcquireTicket></Body></Envelope>};
+
+    # request VMWare session
+    my $ua = ($self->{_vmware_ua} //= Mojo::UserAgent->new);
+    $ua->cookie_jar->empty;    # avoid auth error because we're already logged in
+    my ($auth_txn, $auth_headers) = _prepare_vmware_request($ua, $api_url, $auth_xml);
+    $auth_headers->cookie('vmware_client=VMware');
+    $ua->insecure(1);    # so far our setup doesn't have a proper certificate
```
kalikiana, line 62: This might seem like nit-picking but I would prefer if this was explicitly requested rather than done automatically. If I was setting up a machine I would expect to see an error if the certificate wasn't valid

t/27-consoles-vnc.t
```diff
@@ -365,4 +367,28 @@ subtest 'server initialization' => sub {
     is_deeply \@printed, \@expected, 'pixel format and encodings replied' or diag explain \@printed;
 };
 
+subtest 'login on real VNC server via vnctest, request and receive frame buffer' => sub {
+    # This test is using `vnctest` so this script is covered as well. Note that the `vnctest` script has mainly been added to be able to run our VNC client
+    # code manually against a real VNC server (which can sometimes be useful).
+
+    my $display = 20 // $ENV{VNC_TEST_DISPLAY};
```
kalikiana, line 374: What's this supposed to do? This is always 20, no? Did you mean the reverse?

kalikiana (changes requested):

t/27-consoles-vnc.t
Martchus: Yes, I meant the reverse. Good catch.

kalikiana (approved):
Looks very nice. Thank you for taking my comments into account!

PR #2093, opened 2022-06-17
Fix endian conversion in VNC code for s390x/ppc64

Martchus:
* Implement the endian conversion correctly; one needs to shift the read
  byte itself, the current code keeps the byte order as-is
* Tested on a Tumbleweed ppc64 VM
* See https://progress.opensuse.org/issues/111608

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -649,11 +649,11 @@ static uint16_t read_u16(const unsigned char* data, size_t& offset,
 {
     uint16_t pixel;
     if (do_endian_conversion) {
-        pixel = data[offset++] * 256;
-        pixel += data[offset++];
+        pixel =  (data[offset++] >> 0);
+        pixel |= (data[offset++] >> 8);
     } else {
-        pixel = data[offset++];
-        pixel += data[offset++] * 256;
+        pixel = *(uint16_t*)(data + offset);
+        offset += 2;
     }
     return pixel;
 }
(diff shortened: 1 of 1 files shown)


okurz (approved):

AdamWill:
Thanks a lot! I'll test it. Then, uh, all those ppc64 / s390x workers we all have will finally work again...:D

AdamWill:
Unfortunately I still see two fails on s390x:

    2: 
    2:     #   Failed test 'pixel data updated in framebuffer (blue)'
    2:     #   at t/27-consoles-vnc.t line 171.
    2:     #          got: '0'
    2:     #     expected: '41'
    2: 
    2:     #   Failed test 'pixel data updated in framebuffer (green)'
    2:     #   at t/27-consoles-vnc.t line 172.
    2:     #          got: '0'
    2:     #     expected: '37'
    2:     # Looks like you failed 2 tests of 24.
    2: 
    2: #   Failed test 'update framebuffer'
    2: #   at t/27-consoles-vnc.t line 229.
    2: # Looks like you failed 1 test of 21.
    2: [20:06:31] t/27-consoles-vnc.t ......................

ppmclibs/tinycv_impl.cc
```diff
@@ -649,11 +649,11 @@ static uint16_t read_u16(const unsigned char* data, size_t& offset,
 {
     uint16_t pixel;
     if (do_endian_conversion) {
-        pixel = data[offset++] * 256;
-        pixel += data[offset++];
+        pixel =  (data[offset++] >> 0);
+        pixel |= (data[offset++] >> 8);
```
Vogtinator, line 653: Shifting an unsigned char by 8 will always result in 0. I guess you mean `<< 8`?

Vogtinator:
I wonder what the meaning of `do_endian_conversion` is. Previously, it switched between little endian and big endian, while now (with fixed shifting) it can only select between little and native endian, which means that on machines with native little endianess it no longer reads big endian at all.

mloviska:
All s390x osd jobs are [failing](https://openqa.suse.de/tests/8981633#step/bootloader_start/43) could you double-check whether the PR could be the root cause, please?

ppmclibs/tinycv_impl.cc
coolo: which would be * 256 (the compiler does the << 8 for you if you compile with any level of O

okurz:
> All s390x osd jobs are [failing](https://openqa.suse.de/tests/8981633#step/bootloader_start/43) could you double-check whether the PR could be the root cause, please?

Likely so. Proposing an alert for now: https://github.com/os-autoinst/os-autoinst/pull/2095

PR #2095, opened 2022-06-20
Revert "Fix endian conversion in VNC code for s390x/ppc64"

okurz:
Reverts os-autoinst/os-autoinst#2093 due to problems described in https://progress.opensuse.org/issues/111608 as well as https://github.com/os-autoinst/os-autoinst/pull/2093

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -649,11 +649,11 @@ static uint16_t read_u16(const unsigned char* data, size_t& offset,
 {
     uint16_t pixel;
     if (do_endian_conversion) {
-        pixel =  (data[offset++] >> 0);
-        pixel |= (data[offset++] >> 8);
+        pixel = data[offset++] * 256;
+        pixel += data[offset++];
     } else {
-        pixel = *(uint16_t*)(data + offset);
-        offset += 2;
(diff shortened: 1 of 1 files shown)


Martchus (approved):

coolo:
b6c30cd47e6507b675126645a7656acad28ed614 renamed server_endian to no_endian_conversaion flag but never really sets it anywhere - or I heavily overlook this

coolo:
Because `$c->_do_endian_conversion($machine_is_big_endian);` is just wrong in the test - the endian conversion should happen on little endian

Martchus:
Ok, you're right. The test expects the endianess conversion to happen at the wrong place. And the existing endianess conversion is actually correct. My suggested change only fixes the test but the test is just wrong. I suppose Perl's packing behavior doesn't work in the way I think.

Martchus:
> And the existing endianess conversion is actually correct.

As long as it runs on a little-endian system. On a big-endian system it preserves the byte order. My suggested change does the endian-conversion but only if it runs on a big-endian system. So both is kind of bad. With the reverted state our code at least behaves correctly again when it runs on little-endian but I'll have to some up with a fix to do the byte order conversion correctly regardless of the system's endianness (via htons).

PR #2103, opened 2022-06-24
ci: Fully cover OpenQA::Qemu::Proc

perlpunk:

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -8,11 +8,13 @@ use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '5';
 use Test::Fatal;
 use Test::Warnings qw(warnings :report_warnings);
+use Test::Output qw(stderr_from);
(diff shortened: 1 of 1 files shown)


t/scripts/unxz
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+opt_k=$1
+opt_f=$2
+file=$3
+newfile=${file%.xz}
+mv "$file" "$newfile"
```
okurz, line 7: Instead of adding a new file IMHO would be nicer to overwrite the command called within the actual code. Or make it a configurable variable that you overwrite from tests

okurz (changes requested):

mimi1vx (approved):

t/scripts/unxz
perlpunk: done

Martchus (approved):

okurz (approved):

PR #2104, opened 2022-06-24
Fix error from stricter git permission checks

okurz:
git version 2.35 introduced strict checks on directory ownership
refusing to operate on directories not owned by the user calling git.
In the case of openQA tests not running from openQA worker cache
synchronized test distributions this caused the git revision of test
distribution directories to be not parsed and ending up as 'UNKNOWN',
our fallback string.

This commit fixes this problem by adding the requested git directory to
the list of "safe directories" in the user's global git config. This of
course means that the user account must be able to write to the
according config file.

Production verification run: https://openqa.opensuse.org/tests/2454002

Related progress issue: https://progress.opensuse.org/issues/113030

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -12,15 +12,16 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse($dirname) {
+    chomp(my $version = (-d "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN');
+    return $version;
+}
 
 sub calculate_git_hash ($git_repo_dir) {
-    my $dir = getcwd();
-    chdir($git_repo_dir);
-    chomp(my $git_hash = qx{git rev-parse HEAD ||:});
-    $git_hash ||= "UNKNOWN";
-    chdir($dir);
+    my $git_hash = git_rev_parse($git_repo_dir);
     bmwqemu::diag "git hash in $git_repo_dir: $git_hash";
     return $git_hash;
 }
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -16,7 +16,7 @@ our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
   checkout_git_refspec handle_generated_assets load_test_schedule);
 
 sub git_rev_parse($dirname) {
-    chomp(my $version = (-d "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN');
+    chomp(my $version = (-e "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN');
     return $version;
 }
 
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -15,8 +15,13 @@ use Try::Tiny;
 our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
   checkout_git_refspec handle_generated_assets load_test_schedule);
 
-sub git_rev_parse($dirname) {
-    chomp(my $version = (-e "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN');
+sub git_rev_parse ($dirname) {
+    return 'UNKNOWN' unless -e "$dirname/.git";
+    my $checksafe = q{git config --global --get safe.directory | grep -q};
+    my $addsafe = q{HOME=$(mktemp -d --tmpdir os-autoinst-git.XXXXX) && git config --global --add safe.directory};
+    my $version = qx{($checksafe "$dirname" && git -C "$dirname" rev-parse HEAD || $addsafe "$dirname" && git -C "$dirname" rev-parse HEAD && rm -r \$HOME)};
+    $version ||= '(unreadable git hash)';
+    chomp($version);
     return $version;
 }
 
--- a/isotovideo
+++ b/isotovideo
@@ -80,7 +80,7 @@ use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 Getopt::Long::Configure("no_ignore_case");
 use OpenQA::Isotovideo::CommandHandler;
 use OpenQA::Isotovideo::Interface;
-use OpenQA::Isotovideo::Utils qw(checkout_git_repo_and_branch
+use OpenQA::Isotovideo::Utils qw(git_rev_parse checkout_git_repo_and_branch
   checkout_git_refspec handle_generated_assets load_test_schedule);
 
 session->enable;
@@ -94,9 +94,7 @@ sub usage ($r) {
 }
 
 sub _get_version_string () {
-    my $dirname = curfile->dirname;
-    my $thisversion = (-d "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'unknown';
-    chomp $thisversion;
+    my $thisversion = git_rev_parse(curfile->dirname);
     return "Current version is $thisversion [interface v$OpenQA::Isotovideo::Interface::version]";
 }
 
--- a/t/20-openqa-isotovideo-utils.t
+++ b/t/20-openqa-isotovideo-utils.t
@@ -53,4 +53,6 @@ subtest 'error handling when loading test schedule' => sub {
     };
 };
 
+is_deeply OpenQA::Isotovideo::Utils::_store_asset(0, 'foo.qcow2', 'bar'), {hdd_num => 0, name => 'foo.qcow2', dir => 'bar', format => 'qcow2'}, '_store_asset returns correct parameters';
+
 done_testing;
--- a/t/99-full-stack.t
+++ b/t/99-full-stack.t
@@ -29,6 +29,7 @@ note("pool dir: $pool_dir");
 chdir($pool_dir);
 my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
 
+my $casedir = path($data_dir, 'tests');
 path('vars.json')->spurt(<<EOV);
 {
    "ARCH" : "i386",
@@ -36,7 +37,7 @@ path('vars.json')->spurt(<<EOV);
    "QEMU" : "i386",
    "QEMU_NO_TABLET" : "1",
    "QEMU_NO_FDC_SET" : "1",
-   "CASEDIR" : "$data_dir/tests",
+   "CASEDIR" : "$casedir",
    "ISO" : "$data_dir/Core-7.2.iso",
    "CDMODEL" : "ide-cd",
    "HDDMODEL" : "ide-hd",
--- a/t/99-full-stack.t
+++ b/t/99-full-stack.t
@@ -49,6 +49,8 @@ EOV
 path('live_log')->touch;
 system("cd $toplevel_dir && perl $toplevel_dir/isotovideo --workdir $pool_dir -d 2>&1 | tee $pool_dir/autoinst-log.txt");
 my $log = path('autoinst-log.txt')->slurp;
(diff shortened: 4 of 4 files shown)


okurz:
Additionally I would like to check this on production workers to be sure we don't run into problems when the openQA worker user account tries to alter its git config itself.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,17 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $version = (-d "$dirname/.git" ? qx{(git config --global --get safe.directory | grep -q $dirname || git config --global --add safe.directory $dirname) && git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN';
```
perlpunk, line 19: `.git` can also be just a file and not a directory.

mimi1vx (approved):

okurz:
@ggardet Did you test this change on some workers?

Martchus (approved):

OpenQA/Isotovideo/Utils.pm
okurz: ok, added a commit

okurz:
> > @ggardet Did you test this change on some workers?
> 
> No, because the patch does not apply on the installed `os-autoinst` files.

Right. During installation the version within isotovideo is hardcoded to the package version but the patch should still apply for OpenQA/Isotovideo/Utils.pm which is relevant when the test distribution is loaded.

ggardet:
No, it also fails for `OpenQA/Isotovideo/Utils.pm`.

okurz:
> No, it also fails for `OpenQA/Isotovideo/Utils.pm`.

Interesting. That is unexpected. I patched now on openqaworker7 and will monitor there.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,17 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $version = (-e "$dirname/.git" ? qx{(git config --global --get safe.directory | grep -q $dirname || git config --global --add safe.directory $dirname) && git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN';
```
okurz, line 19: This needs write-permissions but normally we have `_openqa-worker` running with a home directory /var/lib/empty which does not offer write permissions for _openqa-worker. Not sure what to do instead.

okurz: Rewritten to read out the git ref manually avoiding git commands.

Martchus (changes requested):
I suppose reading the `.git` files manually is ok. If it doesn't work in some special cases its not that bad for an optional feature. Considering that, maybe it would be good to catch exceptions and provide a fallback instead of a hard fail (e.g. the old code had a fallback to `unknown`).

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,22 @@ use bmwqemu;
...
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    # This is equivalent to "git rev-parse HEAD" but without the security
+    # constraints of more recent git-versions, see
+    # https://github.blog/2022-04-18-highlights-from-git-2-36/#stricter-repository-ownership-checks
+    my $rev = path($dirname . '.git//HEAD')->slurp =~ s/ref: //r;
+    chomp $rev;
+    my $version = path($dirname . "/.git/$rev")->slurp;
```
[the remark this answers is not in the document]
okurz: Actually the proper way is to supply multiple strings by comma. That way the elements are properly joined regardless of the number of `/`.

isotovideo
```diff
@@ -94,9 +94,7 @@ sub usage ($r) {
 }
 
 sub _get_version_string () {
-    my $dirname = curfile->dirname;
-    my $thisversion = (-d "$dirname/.git" ? qx{git -C $dirname rev-parse HEAD} : '') || 'unknown';
-    chomp $thisversion;
+    my $thisversion = git_rev_parse(curfile->dirname);
```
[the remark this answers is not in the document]
okurz: I think we are far from done. Another problem is that we call `git_rev_parse` not necessarily in the root folder of the git repo, e.g. in isotovideo line 243 when we checkout the test distribution thes might be just a subfolder, e.g. in t/14-isotovideo.t it's `./t/data/tests`. Well, git also has a command for that, `git rev-parse --show-toplevel`, which of course suffers from the same security feature. So should we try to walk up all directories until we can find `.git`?

okurz:
I patched again on openqaworker7 with

```
curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/os-autoinst/pull/2104.patch | patch -p1 --directory=/usr/lib/os-autoinst
```

Triggered

```
end=010 openqa-clone-set https://openqa.opensuse.org/tests/2452346 okurz_investigation WORKER_CLASS=openqaworker7
```

-> https://openqa.opensuse.org/tests/overview?distri=opensuse&build=okurz_investigation&version=Tumbleweed (only first 10 results relevant)

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,17 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $version = (-e "$dirname/.git" ? qx{(git config --global --get safe.directory | grep -q $dirname || HOME=\$\(mktemp -d\) && mkdir -p \$HOME && git config --global --add safe.directory $dirname) && git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN';
```
perlpunk, line 19: `mktemp` should already create this directory, so why `mkdir`?

Also I would remove the directory afterwards.

Another thing I consider good practice is to use a template for tempdirs, so if for some reason a lot of them are there, we know at least where they come from.

Also for my taste the line is too long. The backslash stuff can be reduced by doing something like:
```
my $addsafe = q{HOME=$(mktemp -d) && git config --global --add safe.directory};
... qx{git config --global --get safe.directory | grep -q $dirname || $addsafe $dirname ) ... } ...
```

okurz:
Needs https://github.com/os-autoinst/openQA/pull/4734

OpenQA/Isotovideo/Utils.pm
okurz: Done together with a template for the temp dir.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,18 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $addsafe = q{HOME=$(mktemp -d os-autoinst) && git config --global --add safe.directory};
```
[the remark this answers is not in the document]
okurz: done

okurz:
Production verification run: https://openqa.opensuse.org/tests/2454002

okurz:
https://github.com/os-autoinst/openQA/pull/4734 should be merged first

perlpunk (approved):

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,18 @@ use bmwqemu;
...
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $addsafe = q{HOME=$(mktemp -d os-autoinst-git.XXXXX) && git config --global --add safe.directory};
+    my $version = (-e "$dirname/.git" ? qx{(git config --global --get safe.directory | grep -q $dirname || $addsafe $dirname) && git -C $dirname rev-parse HEAD} : '') || 'UNKNOWN';
```
Martchus, line 20: Don't we need to quote `$dirname` here like this?

```suggestion
    my $version = (-e "$dirname/.git" ? qx{(git config --global --get safe.directory | grep -q "$dirname" || $addsafe "$dirname") && git -C "$dirname" rev-parse HEAD} : '') || 'UNKNOWN';
```

OpenQA/Isotovideo/Utils.pm
```diff
@@ -12,15 +12,18 @@ use bmwqemu;
 use autotest;
 use Try::Tiny;
 
-our @EXPORT_OK = qw(checkout_git_repo_and_branch checkout_git_refspec
-  handle_generated_assets load_test_schedule);
+our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
+  checkout_git_refspec handle_generated_assets load_test_schedule);
+
+sub git_rev_parse ($dirname) {
+    my $addsafe = q{HOME=$(mktemp -d os-autoinst-git.XXXXX) && git config --global --add safe.directory};
```
Martchus, line 19: Would this temp dir be cleaned up at some point? Especially since we seem to create always a new temp dir it would make sense. I suppose we could alternatively also just create a directory within the current working directory. That's where isotovideo already stores intermediate/temporary files and it would make sense to keep doing that. Since the worker normally cleans the "pool" directory you'd get the cleanup for free.

Martchus (dismissed):

OpenQA/Isotovideo/Utils.pm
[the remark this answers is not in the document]
perlpunk: `mktemp -d os-autoinst-git.XXXXX --tempdir` will create it in `/tmp`.
But still it needs to be removed afterwards.

I would really still prefer the template, even if we remove it.
It is even more useful in case the directories are *not* removed because of some mistake.
If that happens and we have hundreds of those lying around, we know where it comes from and can react.

OpenQA/Isotovideo/Utils.pm
okurz: Theoretically yes but os-autoinst does not know that openQA behaves like that … or at least shouldn't need to know ;) But please see https://github.com/os-autoinst/os-autoinst/pull/2104#discussion_r914934993 about the deletion. I will implement that proposal including the note from @perlpunk about the template plus the `--tempdir` option

okurz: done

OpenQA/Isotovideo/Utils.pm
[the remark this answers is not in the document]
okurz: Ok, `mktemp -d --tmpdir os-autoinst-git.XXXXX` did it. I now included my proposal from https://github.com/os-autoinst/os-autoinst/pull/2104#discussion_r914934993 plus the template plus the proposal from https://github.com/os-autoinst/os-autoinst/pull/2104#discussion_r914954525 with more quotes

Martchus (commented):
It is really hard to tell what this mix of && and || works as intended. Does it really have to beat one-liner shell command?

okurz:
> It is really hard to tell what this mix of && and || works as intended. Does it really have to be a one-liner shell command?

How would you write it so that it's still a single shell session?

Martchus:
Just extract it, like check_qemu_oom.

okurz:
> Just extract it, like check_qemu_oom.

But we don't need it to be a separate script file. Can we avoid the extra file for now? You can blame me if the `&& ||` combinations don't work ;)

PR #2106, opened 2022-07-04
testapi: Fix use of invalid option wait_screen_changes

kalikiana:
The documented option `wait_screen_changes` is being used in the typing test and in distri opensuse. However the option has no effect. Logs will report the value of `wait_screen_change` instead.
We should document and correctly log `wait_screen_change`.


Martchus (commented):
The commit message should be `testapi: Fix use of invalid option …` to follow our standards. Otherwise it looks good.

Martchus (approved):

PR #2110, opened 2022-07-07
Prevent passwords leaked to log by redundant log-message

okurz:
Commit 4e66dcc0 introduced duplicate log messages with leaked passwords
like this:

```
susedistribution::handle_password_prompt -> lib/susedistribution.pm:70 called testapi::type_password
[2022-07-06T12:06:05.534656+02:00] [debug] <<< testapi::type_string(string="[masked]", max_interval=100, wait_screen_change=0, wait_still_screen=0, timeout=30, similarity_level=47)
[2022-07-06T12:06:05.535894+02:00] [debug] <<< consoles::video_base::type_string(max_interval=100, cmd="backend_type_string", json_cmd_token="HWQIViIt", text="nots3cr3t")
```

We should not need the second log message at all and also it's not
properly handling the password so I am suggesting to simply remove it
completely.

Related progress issue: https://progress.opensuse.org/issues/113312

### DIFF
--- a/consoles/video_base.pm
+++ b/consoles/video_base.pm
@@ -30,8 +30,6 @@ sub _typing_limit () { $bmwqemu::vars{TYPING_LIMIT} // TYPING_LIMIT_DEFAULT || 1
 sub send_key_event ($key, $press_release_delay) { }
 
 sub type_string ($self, $args) {
-    bmwqemu::log_call(%$args, $args->{secret} ? (-masked => $args->{text}) : ());
-
     my $seconds_per_keypress = 1 / _typing_limit;
 
     # further slow down if being asked for.


okurz:
Adding "not-ready" flag to await feedback from multiple persons:
@cfconrad @jlausuch @Martchus @b10n1k @asmorodskyi asking you as authors or reviewers from former changes:
* #2002
* #2054
* #2062

Martchus (approved):

asmorodskyi (approved):
LGTM

jlausuch (approved):
+1

cfconrad:
Thx for pointing. 

I think, the actual error is here https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1454 where we call the `backend_type_string` for all **none** serial terminal backends.
And these calls just miss the parameter `secret` (**and** all others beside `text` and `max_interval`).

But you are right, I added this log inside of https://github.com/os-autoinst/os-autoinst/pull/2062/files#diff-457f23a16e93b308d1b29031db3d25ab7f7b5bc10b63541230aa51891523b9cdR33 and maybe it's not needed.

From my point, it is worth to have it, as it actually point out such errors where the info get lost, if we call backend routines via our json rpc.

But I do not wanna stop merging, as removing it is fast and hides the problem.

baierjan (approved):

kalikiana (changes requested):
> I think, the actual error is here https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1454 where we call the `backend_type_string` for all **none** serial terminal backends. And these calls just miss the parameter `secret` (**and** all others beside `text` and `max_interval`).

`ssh_screen` and `serial_screen` also rely on `-masked`. So if we decide that this is testapi-only those have to be fixed here as well by dropping the equivalent log calls.

> From my point, it is worth to have it, as it actually point out such errors where the info get lost, if we call backend routines via our json rpc.

What kind of errors do we lose in practice? Do you have an example?

cfconrad:
I think I do not understand you. This is how I searched for the infos:
```bash
clemix@linux-7k5u:~/sandbox/os-autoinst> grep -rnE  'ssh_screen|serial_screen' --include "*.pm" | grep new
consoles/sshSerial.pm:53:    $self->{screen} = consoles::ssh_screen->new(
consoles/sshVirtshSUT.pm:44:    $self->{screen} = consoles::ssh_screen->new(ssh_connection => $ssh, ssh_channel => $chan);
consoles/virtio_terminal.pm:146:    $self->{screen} = consoles::serial_screen::->new($self->{fd_read}, $self->{fd_write});
clemix@linux-7k5u:~/sandbox/os-autoinst> grep 'is_serial_terminal' consoles/sshSerial.pm consoles/sshVirtshSUT.pm consoles/virtio_terminal.pm
consoles/sshSerial.pm:sub is_serial_terminal ($self) { 1 }
consoles/sshVirtshSUT.pm:sub is_serial_terminal ($self) { 1 }
consoles/virtio_terminal.pm:sub is_serial_terminal ($self, @) { 1 }
```
And with this, all these get called with https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1434 , which pass all parameters including `secret`.

What I try to say is, that a change like this https://github.com/os-autoinst/os-autoinst/commit/61a1699b0d21ea772800395d11fea1ea97b7a86f should also solve the problem and I think make it even better, so the backend also get's the relevant info. Of course there are smarter ways, cause now we send all parameters including `lf` which might be filtered out first...

kalikiana:
> And with this, all these get called with https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1434 , which pass all parameters including `secret`.

Did you see the specific lines of code I was linking above? They look like this:

    bmwqemu::log_call(%$nargs, $nargs->{secret} ? (-masked => $nargs->{text}) : ());

Same as the line being removed in this PR. So I don't see how the use of `secret` helps, since it's never being passed down to this code. And it would lead to the same duplication of logs.

cfconrad:
As I already mentioned, `backend_type_string` is called differently depending on the backend `is_serial_terminal()`.
And the lines you mentioned belong to screen's which are only instantiated by backends which use `is_serial_terminal == 1`
Simply check this:
```bash
wget -q -O - https://openqa.suse.de/tests/8753119/file/autoinst-log.txt | grep 'serial_screen::type_string' | grep secret=1
```
The reason, why it isn't logged twice is, that there is just no `bmwqemu::log_call` call in that branch in `testapi::type_string`.

Again, I think it is better to call the backend functions always with the same arguments, regardless if they belong to `is_serial_terminal()` or not, but this is cosmetic. So just removing the `log_call()` in that specific console is also ok :)

okurz:
>     * https://github.com/os-autoinst/os-autoinst/blob/master/consoles/serial_screen.pm#L72
> 
>     * https://github.com/os-autoinst/os-autoinst/blob/master/consoles/ssh_screen.pm#L56

I would tend to delete those two. I don't think we should repeat the log (with duplicated code) in each backend or console this way. If you are interested in the full stacktrace there are better ways or we add a generic method that we can call without needing to repeat the code.

cfconrad:
maybe I'm wrong, but I don't think these produce duplicated log, can you show me a autoinst-log.txt where you see this?

okurz:
I agree with @kalikiana. The current PR is an improvement and I will *not* delete the serial+ssh backend log messages but maybe we can improve that later on.

PR #2111, opened 2022-07-07
Add cloud init data for vmware tests

mloviska:
As per [VMware image and cloud-init](https://jira.suse.com/browse/TEAM-5569), we need to test vmware with cloud init. This is possible only in case test backend injects cloud init data in to the vmx file directly

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -24,6 +24,11 @@ sub new ($class, $testapi_console = undef, $args = {}) {
     $self->name("openQA-SUT-" . $self->instance);
     $self->vmm_family($bmwqemu::vars{VIRSH_VMM_FAMILY} // 'kvm');
     $self->vmm_type($bmwqemu::vars{VIRSH_VMM_TYPE} // 'hvm');
+    if ($self->vmm_family eq 'vmware') {
+        $self->{_vmm_user} = $bmwqemu::vars{VMWARE_USERNAME};
+        $self->{_vmm_pword} = $bmwqemu::vars{VMWARE_PASSWORD};
+        $self->{_vmm_host} = $bmwqemu::vars{VMWARE_HOST};
+    }
 
     return $self;
 }
@@ -461,26 +466,19 @@ sub add_disk ($self, $args) {
     return;
 }
 
-sub virsh () {
-    my $virsh = 'virsh';
-    $virsh .= ' ' . $bmwqemu::vars{VMWARE_REMOTE_VMM} if $bmwqemu::vars{VMWARE_REMOTE_VMM};
-    return $virsh;
-}
-
 sub suspend ($self) {
-    $self->run_cmd(virsh() . " suspend " . $self->name) && die "Can't suspend VM ";
+    $self->run_cmd($self->{virsh} . " suspend " . $self->name) && die "Can't suspend VM ";
     bmwqemu::diag "VM " . $self->name . " suspended";
 }
 
 sub resume ($self) {
-    $self->run_cmd(virsh() . " resume " . $self->name) && die "Can't resume VM ";
+    $self->run_cmd($self->{virsh} . " resume " . $self->name) && die "Can't resume VM ";
     bmwqemu::diag "VM " . $self->name . " resumed";
 }
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
-sub define_and_start ($self) {
-    my $remote_vmm = "";
+sub prepare_virsh ($self) {
     if ($self->vmm_family eq 'vmware') {
         my ($fh, $libvirtauthfilename) = File::Temp::tempfile(DIR => "/tmp/");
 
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -489,48 +487,93 @@ sub define_and_start ($self) {
         $self->run_cmd(
             "cat > $libvirtauthfilename <<__END
 [credentials-vmware]
-username=" . ($bmwqemu::vars{VMWARE_USERNAME} or die 'Need variable VMWARE_USERNAME') . "
-password=" . ($bmwqemu::vars{VMWARE_PASSWORD} or die 'Need variable VMWARE_PASSWORD') . "
-[auth-esx-" . ($bmwqemu::vars{VMWARE_HOST} or die 'Need variable VMWARE_HOST') . "]
+username=$self->{vmm_user}
```
Martchus, line 490: Here you're using hash keys but before these values are stored using attributes. That seems inconsistent and is maybe also wrong.

consoles/sshVirtsh.pm
```diff
@@ -489,48 +487,93 @@ sub define_and_start ($self) {
...
+    if ($ci_meta && $ci_user) {
+        $self->run_cmd("echo guestinfo.metadata = \"$ci_meta\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd("echo guestinfo.metadata.encoding = \"gzip+base64\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd("echo guestinfo.userdata = \"$ci_user\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd("echo guestinfo.userdata.encoding = \"gzip+base64\" >> $vmx", domain => 'sshVMwareServer');
+    }
+}
+
+sub define_and_start ($self) {
+    bmwqemu::diag("Method consoles::sshVirtsh::define_and_start will become DEPRECATED!");
```
Martchus, line 564: Why deprecating it? This function is used for other VM types (besides VMWare). I'd generally try to minimize the impact of your change for other VM types. (If you really need to affect other VM types, you should also test these, see https://github.com/os-autoinst/os-autoinst/blob/master/doc/backends.md for some documentation.)

consoles/sshVirtsh.pm
mloviska: Not sure why, but then in is interpreted as this.

```
  username=consoles::sshVirtsh=HASH(0x557622de7e78)->vmm_user
  password=consoles::sshVirtsh=HASH(0x557622de7e78)->vmm_pword
  [auth-esx-consoles::sshVirtsh=HASH(0x557622de7e78)->vmm_host]
```

mloviska: Or should it be like this ?
```
...
username=${self->vmm_user}
...
```

consoles/sshVirtsh.pm
Martchus: Keep in mind that we can only merge your changes when `xen` and `zkvm` work still unaffected. So generally deprecating this function seems still wrong. If you need to do specific things for `vmware` then just do that but keep other code in place.

mloviska: I am not going to remove the function within this PR, just marking that it might be removed in the future. However, if this is not recommended, let's keep it then.

consoles/sshVirtsh.pm
Martchus: To my knowledge you cannot call a function inside a string literal (and an attribute is basically a function). You can store the values in variables before making the string. You can also simply *not* using attributes (consistently). Especially if these fields should only be used by the object internally, it would make sense to use hash-keys with leading underscore like `$self->{_vmm_user}` (which btw. also gives better performance than attributes).

mloviska: Updated, thanks!

consoles/sshVirtsh.pm
```diff
@@ -489,48 +487,93 @@ sub define_and_start ($self) {
...
+    my $ci_meta = $bmwqemu::vars{CLOUD_INIT_META};
+    my $ci_user = $bmwqemu::vars{CLOUD_INIT_USER};
+
+    # set default boot delay
+    $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
+
+    # inject cloud init metadata and userdata required for the image
+    if ($ci_meta && $ci_user) {
+        $self->run_cmd("echo guestinfo.metadata = \"$ci_meta\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd("echo guestinfo.metadata.encoding = \"gzip+base64\" >> $vmx", domain => 'sshVMwareServer');
```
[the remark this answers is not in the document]
mloviska: Added, thanks!

mloviska:
VR: [sle-15-SP4-JeOS-for-VMware-x86_64-Build2.24.32 + cloud-init](http://kepler.suse.cz/tests/17659#step/record_machine_id/9)

consoles/sshVirtsh.pm
kalikiana: Discussed it briefly. As I understand it there is no plans to refactor other backends soon, so I suggest to drop the deprecation message (which would be misleading) and just keep the changes minimal. Also, double-check that coverage looks okay e.g. at least hits the code path in existing tests.

PR #2112, opened 2022-07-08
Increase timeout for local VNC connections to prevent incompletes

Martchus:
Especially on slow workers like aarch64 we've seen incompletes since
d1adda78adc34c5ac02b5040a2bc0e97eaa83827 and
93ff454deae61e573a9cbf88f172304002fb83a4 have been deployed. These commits
applied the VNC timeouts for also for reads/writes on the socket (which
were previously only applied to establishing the connection).

Apparently the timeout of 10 seconds for local connections is too low for
slow systems like the o3 worker aarch64 leading to incompletes like
`backend died: unexpected end of data at
/usr/lib/os-autoinst/consoles/VNC.pm line 183.`.

I increased the timeouts on that worker to the values of this change and
none of the 582 jobs ran into the problem anymore. So it is likely a more
sensible default.

Note that the error handling could still be improved, e.g. to re-connect
instead of just stopping the backend.

Related ticket: https://progress.opensuse.org/issues/113282

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -133,9 +133,9 @@ sub login ($self, $connect_timeout = undef, $timeout = undef) {
     my $port = $self->port || 5900;
     my $description = $self->description || 'VNC server';
     my $is_local = $hostname =~ qr/(localhost|127\.0\.0\.\d+|::1)/;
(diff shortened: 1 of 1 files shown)


ggardet (approved):
Thanks!

kalikiana (approved):

okurz (approved):

PR #2113, opened 2022-07-08
t: Increase timeout for scripts

perlpunk:
Issue: https://progress.opensuse.org/issues/113141

Especially on aarch64 builds it can happen that a command like
`isotovideo --help` exceeds the 3 seconds timeout.

Since we have the overall OpenQA::Test::TimeLimit I think it's ok to
increase the timeout.

### DIFF
--- a/t/44-scripts.t
+++ b/t/44-scripts.t
@@ -24,9 +24,9 @@ for my $key (keys %types) {
 }
 
 for my $script (sort keys %types) {
-    my $out = qx{timeout 3 $Bin/../$script --help 2>&1};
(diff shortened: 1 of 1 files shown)


Martchus (changes requested):
I'm fine with that. Unfortunately 3 seconds can beat bit less for a full blown Perl Script. The only problem is a missing c in the commit message.

perlpunk:
oops. `c` added :)

Martchus (approved):

okurz (approved):

PR #2117, opened 2022-07-12
Document testing svirt/VMWare using local virsh client

Martchus:


doc/backends.md
```diff
@@ -111,6 +111,66 @@ test distribution for details. When cloning an openSUSE test, the
...
+
+In this setup, `virsh` will be invoked with the `-c` parameter to specify the remote
+hypervisor host. If you want to use `virsh` manually, you also need to specify that
+parameter accordingly (e.g. grab the hypervisor URL from the autoinst log while the
+test is running).
+
+Like with "Local setup" one *could* also use `virt-manager`. The `esx://`-URL
+can be added as custom URL and then one is prompted for the password. I only ran
+into the problem that the self-signed certificate of our hypervisor instance was
+not accepted.
```
kalikiana, line 130: Can this be ignored or did you have to update the machine? Might be good to mention.

Martchus: I cannot answer that question. I *suppose* it cannot be ignored but I don't want to write it because I'm not sure. I *suppose* updating the machine might also be an option but again I wouldn't know exactly whether and how it is possible. These instructions really only go as far as I know things for sure. I just wanted to mention that it is possible to add it as custom URL in general but said caveats are expected.

kalikiana: Ack. That's fine by me.

PR #2118, opened 2022-07-13
Stop showing ipmi password in error ouput

Julie-CAO:
Stop showing ipmi passwords in autoinst.txt from a ipmi backend job. 

The error output is changed to:
```
[2022-07-13T03:02:30.777057-04:00] [info] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  ipmitool -I lanplus -H 10.67.135.53 -U tester -P <secret> mc guid: Error: Unable to establish IPMI v2 / RMCP+ session at /usr/lib/os-autoinst/backend/ipmi.pm line 48.
```

- ticket: https://progress.opensuse.org/issues/110227
- verification run: http://10.67.129.102/tests/298

@kalikiana @okurz @alice-suse and others, welcome review!

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -41,8 +41,9 @@ sub ipmitool ($self, $cmd, %args) {
     }
     chomp $stdout;
     chomp $stderr;
+    # Output error message with ipmi password masked
+    die join(' ', map { $_ eq $bmwqemu::vars{IPMI_PASSWORD} ? "[masked]" : $_ } @cmd) . ": $stderr" unless ($ret);
 
-    die join(' ', @cmd) . ": $stderr" unless ($ret);
     bmwqemu::diag("IPMI: $stdout");
     return $stdout;
 }
--- a/t/29-backend-ipmi.t
+++ b/t/29-backend-ipmi.t
@@ -25,6 +25,7 @@ my @ipmi_cmdline = $backend->ipmi_cmdline;
 is_deeply \@ipmi_cmdline, [qw(ipmitool -I lanplus -H fake_HOSTNAME -U fake_USER -P fake_PASSWORD)], 'valid ipmi_cmdline';
 
 my $ipmi = Test::MockModule->new('backend::ipmi');
+throws_ok { $backend->ipmitool('foo') } qr/[masked]/, 'ipmi password masked in error output';
 $ipmi->redefine(ipmi_cmdline => sub { (qw(echo simulating ipmi)) });
 my $ret;
 combined_like { $ret = $backend->ipmitool('foo') } qr/IPMI: simulating ipmi foo/, 'log output for IPMI call';


backend/ipmi.pm
```diff
@@ -41,8 +41,13 @@ sub ipmitool ($self, $cmd, %args) {
     }
     chomp $stdout;
     chomp $stderr;
+    unless ($ret) {
+        # Don't show ipmi password in error output
+        my $secreted_cmd = join(' ', @cmd);
+        $secreted_cmd =~ s/-P\s+\w+/-P <secret>/;
+        die "$secreted_cmd: $stderr";
+    }
```
okurz, line 49: As we already have some special handling for masking secrets I wonder, @cfconrad do you have an idea how something like https://github.com/os-autoinst/os-autoinst/blob/af9bd7805f9b3b51b3516c0334f72cb162605079/bmwqemu.pm#L209 can be used here without duplicating the implementation?

okurz (changes requested):

backend/ipmi.pm
kalikiana: This code parses the `-P` argument from `ipmi`. The secret hiding in `log_call` in `bmwqemu.pm` replaces values. I'm not sure about sharing code, but it should probably behave the same way?

okurz: At the very least we should use the same text `[masked]` and remove the die-message :)

Julie-CAO: Thanks. changed to '[masked]' to behave the same way. Now the new error output is :
```
[2022-07-13T21:21:44.454344-04:00] [info] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  ipmitool -I lanplus -H 10.67.135.53 -U tester -P [masked] mc guid: Error: Unable to establish IPMI v2 / RMCP+ session at /usr/lib/os-autoinst/backend/ipmi.pm line 48.
```
The die-message cannot be removed because there is a die-handler function called in its baseclass:

```
$SIG{__DIE__} = \&die_handler;
...
sub die_handler ($msg) {
    chomp($msg);
    bmwqemu::fctinfo "Backend process died, backend errors are reported below in the following lines:\n$msg";
    bmwqemu::serialize_state(component => 'backend', msg => $msg);
    $backend->stop_vm();
    $backend->close_pipes();
}
```

Julie-CAO:
the codecov/patch check failed with:
```
[2022-07-14T01:39:30.221Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Could not find a repository, try using repo upload token', code='not_found')}
```
what can I do from my side?

backend/ipmi.pm
cfconrad: We could add a list of mask-able strings into the logger and then replace them here: https://github.com/os-autoinst/os-autoinst/blob/af9bd7805f9b3b51b3516c0334f72cb162605079/log.pm#L24
But this is nothing I would do in this PR and maybe it will then again hit the timing concerns you had in the past...

A nitty hint, you could write it with a map as well..
```suggestion
    die join(' ', map { $_ eq $bmwqemu::vars{IPMI_PASSWORD} ? "[masked]" : $_ } @cmd) . ": $stderr" unless ($ret);
```

okurz (changes requested):

okurz:
Please first add unit tests covering the changed code path. Take a look into t/29-backend-ipmi.t. Or better first try to execute it locally, e.g. make && prove -I. t/29-backend-ipmi.t

Julie-CAO:
> Please first add unit tests covering the changed code path. Take a look into t/29-backend-ipmi.t. Or better first try to execute it locally, e.g. make && prove -I. t/29-backend-ipmi.t

Unit test added:
```
jcao@localhost:~> prove -I. t/29-backend-ipmi.t 
t/29-backend-ipmi.t .. ok    
All tests successful.
Files=1, Tests=17,  1 wallclock secs ( 0.02 usr  0.00 sys +  0.46 cusr  0.03 csys =  0.51 CPU)
Result: PASS
```

kalikiana (approved):

PR #2119, opened 2022-07-13
Fix git version readout for symlinked directories

okurz:
Related progress issue: https://progress.opensuse.org/issues/113030


perlpunk (approved):

Martchus (approved):

okurz:
There might still be a problem with a trailing slash. I wil try to reproduce the failure first, e.g. in t/99-full-stack.t

PR #2121, opened 2022-07-14
Fix git version readout from sub-directory + tests

okurz:
* Add test for git version reading as non-owner
* Add back support for reading layered test distribution's git versions
* t: Add unit tests for 'git_rev_parse' and 'checkout_git_refspec'

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -15,12 +15,12 @@ our @EXPORT_OK = qw(git_rev_parse checkout_git_repo_and_branch
   checkout_git_refspec handle_generated_assets load_test_schedule);
 
 sub git_rev_parse ($dirname) {
-    return 'UNKNOWN' unless -e "$dirname/.git";
     $dirname = path($dirname)->realpath;
-    my $checksafe = q{git config --global --get safe.directory | grep -q};
-    my $addsafe = q{HOME=$(mktemp -d --tmpdir os-autoinst-git.XXXXX) && git config --global --add safe.directory};
-    my $version = qx{($checksafe "$dirname" && git -C "$dirname" rev-parse HEAD || $addsafe "$dirname" && git -C "$dirname" rev-parse HEAD && rm -r \$HOME)};
-    $version ||= '(unreadable git hash)';
+    chomp(my $version = qx{git -C "$dirname" rev-parse HEAD 2>&1});
+    return $version if $? == 0;
+    return 'UNKNOWN' unless $version =~ /(git config.*safe.directory.*$)/;
+    my $addsafe = 'TMPDIR=$(mktemp -d --tmpdir os-autoinst-git.XXXXX) && HOME=$TMPDIR && ' . $1;
+    $version = qx{$addsafe && git -C "$dirname" rev-parse HEAD && rm -r \$TMPDIR} || '(unreadable git hash)';
     chomp($version);
     return $version;
 }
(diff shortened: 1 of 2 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -14,13 +14,13 @@ use Try::Tiny;
...
-    return 'UNKNOWN' unless -e "$dirname/.git";
+sub git_rev_parse ($dirname, $cmd_prefix='') {
     $dirname = path($dirname)->realpath;
-    my $checksafe = q{git config --global --get safe.directory | grep -q};
-    my $addsafe = q{HOME=$(mktemp -d --tmpdir os-autoinst-git.XXXXX) && git config --global --add safe.directory};
-    my $version = qx{($checksafe "$dirname" && git -C "$dirname" rev-parse HEAD || $addsafe "$dirname" && git -C "$dirname" rev-parse HEAD && rm -r \$HOME)};
-    $version ||= '(unreadable git hash)';
+    chomp(my $version = qx{$cmd_prefix git -C "$dirname" rev-parse HEAD 2>&1});
+    return $version if $? == 0;
+    return 'UNKNOWN' unless $version =~ /(git config.*safe.directory.*$)/;
```
perlpunk, line 21: Hm, so now we'll get "UNKNOWN" for every error (except the unsafe dir error).
I don't know why you removed the first line.

Martchus: I guess we can live with `UNKNOWN` in case of an error.

Martchus (approved):
Looks like this could actually work :-)
Git even does the quoting for us in case the path contains white-spaces.

OpenQA/Isotovideo/Utils.pm
perlpunk: ok

t/20-openqa-isotovideo-utils.t
```diff
@@ -12,14 +12,25 @@ use Mojo::Util qw(scope_guard);
...
 chdir $dir;
 my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
 mkdir $pool_dir;
 
+
+is git_rev_parse($dir), 'UNKNOWN', 'non-git repo detected as such';
+stderr_like { checkout_git_refspec($dir, 'MY_REFSPEC_VARIABLE') }
+qr/git hash in.*UNKNOWN/, 'checkout_git_refspec also detects UNKNOWN';
+my $toplevel_dir = "$Bin/..";
+my $version = -e "$toplevel_dir/.git" ? qr/[a-f0-9]+/ : 'UNKNOWN';
```
perlpunk, line 28:
```suggestion
my $version = -e "$toplevel_dir/.git" ? qr/[a-f0-9]+/ : qr/^UNKNOWN$/;
```

OpenQA/Isotovideo/Utils.pm
```diff
@@ -14,13 +14,13 @@ use Try::Tiny;
...
+    my $addsafe = 'TMPDIR=$(mktemp -d --tmpdir os-autoinst-git.XXXXX) && HOME=$TMPDIR && ' . $1;
+    $version = qx{$addsafe && git -C "$dirname" rev-parse HEAD && rm -r \$TMPDIR} || '(unreadable git hash)';
```
okurz, line 23: if this line is not covered then this likely means that https://github.com/os-autoinst/os-autoinst/pull/2121/files#diff-42b9d43b3da224eb09923bc4972451da136f214a1e03b139f17033c85160a54dR31 does not return successfully, i.e. `command -v sudo >/dev/null && sudo -u nobody true` does not work within our CI environment.

okurz:
* https://github.com/os-autoinst/os-autoinst/pull/2126
might help

OpenQA/Isotovideo/Utils.pm
Martchus: I suppose marking as uncoverable is good enough if adding (and using) an additional user becomes too complicated.

PR #2122, opened 2022-07-14
Switch development base back to Tumbleweed assuming problems are fixed

okurz:
We used openSUSE Tumbleweed as development base some time ago but needed
to switch to Leap due to some problems concerning container runtime
environments and glibc. As openSUSE Leap 15.4 has been released we
should switch to a more recent version. Let's try with Tumbleweed again.

Related progress issue: https://progress.opensuse.org/issues/111881


Martchus (approved):
The next CRE/glibc incompatibility might be further away than Leap 15.4's EOL so why not :-)

okurz:
Let's see how this goes. The Dockerfile isn't used in CI so far so we need to merge it first so that OBS builds the updated image. Then we update tools/container_run_ci

okurz:
https://github.com/os-autoinst/os-autoinst/pull/2124 as follow-up for switching CI build as well

PR #2123, opened 2022-07-14
send_key_until_needlematch: Avoid check_screen timeout in the first iteration

Vogtinator:
When send_key_until_needlematch is called, the key was not pressed yet, so
the expected state is either already there or won't automatically appear,
so don't waste time waiting for it. Instead press the key immediately if
check_screen fails.

VR: http://10.168.4.168/tests/1232

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 26;
+our $version = 27;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -651,6 +651,7 @@ subtest 'check_assert_shutdown' => sub {
     $mod->redefine(read_json => {ret => 0});
     is(check_shutdown, 0, 'check_shutdown should return "false" if timeout is hit');
     throws_ok { assert_shutdown } qr/Machine didn't shut down!/, 'assert_shutdown should throw exception if timeout is hit';
+    $mod->redefine(read_json => \&fake_read_json);
 };
 
 subtest 'compat_args' => sub {
@@ -920,5 +921,48 @@ subtest 'get_var_array and check_var_array' => sub {
 };
 
 like(exception { x11_start_program 'true' }, qr/implement x11_start_program/, 'x11_start_program needs specific implementation');
+
+subtest 'send_key_until_needlematch' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+
+    $mock_testapi->redefine(wait_screen_change => sub : prototype(&@) {
+            my ($callback, $timeout) = @_;
+            $callback->() if $callback;
+    });
+    $mock_testapi->redefine(_handle_found_needle => sub { return $_[0] });
+    $mock_testapi->redefine(assert_screen => sub { die 'assert_screen reached' });
+
+    my $mock_tinycv = Test::MockModule->new('tinycv');
+    $mock_tinycv->redefine(from_ppm => sub : prototype($) { return bless({} => __PACKAGE__); });
+
+    # Check immediate needle match
+    $fake_needle_found = 1;
+    $cmds = [];
+    send_key_until_needlematch('tag', 'esc');
+    is(scalar @$cmds, 1, 'needle matches immediately, no key sent') || diag explain $cmds;
+    is($cmds->[-1]->{cmd}, 'check_screen');
+    is($cmds->[-1]->{timeout}, 0);
+
+    # Needle never matches
+    $mock_testapi->unmock('send_key');
+    $fake_needle_found = $fake_needle_found_after_pause = 0;
+    $cmds = [];
+    throws_ok(sub { send_key_until_needlematch('tag', 'esc', 2); },
+        qr/assert_screen reached/,
+        'no candidate needle matched tags'
+    );
+
+    my $count_send_key = 0;
+    # Skip the first check_screen
+    shift @$cmds;
+    for my $cmd (@$cmds) {
+        is($cmd->{timeout}, 1, "timeout for other check_screen is nonzero") if $cmd->{cmd} eq 'check_screen';
+        $count_send_key++ if $cmd->{cmd} eq 'backend_send_key';
+    }
+    is($count_send_key, 3, 'tried to send_key three times') || diag explain $cmds;
+    $cmds = [];
+
+    $fake_needle_found = 1;
+};
 
 done_testing;
--- a/testapi.pm
+++ b/testapi.pm
@@ -1378,13 +1378,16 @@ sub send_key_until_needlematch {
 
     $counter //= 20;
     $timeout //= 1;
-    while (!check_screen($tag, $timeout)) {
+
+    my $real_timeout = 0;
+    while (!check_screen($tag, $real_timeout)) {
         wait_screen_change {
             send_key $key;
         };
         if (!$counter--) {
             assert_screen $tag, 1;
         }
+        $real_timeout = $timeout;
     }
 }
 


Vogtinator:
> However I would be fine to merge this if you could add test coverage for the function, e.g. in t/03-testapi.t,

I'll try that if it's not too involved, but I don't really understand why this is a requirement. Having a unit test doesn't really offset the potential that it might break some use of the function, does it?

> plus bump the version number in OpenQA/Isotovideo/Interface.pm

Is that really necessary? The behaviour doesn't really change in a meaningful way and the API remains completely unchanged.

okurz:
> > However I would be fine to merge this if you could add test coverage for the function, e.g. in t/03-testapi.t,
> 
> I'll try that if it's not too involved, but I don't really understand why this is a requirement. Having a unit test doesn't really offset the potential that it might break some use of the function, does it?

Are you arguing with Edsger W. Dijkstra “Program testing can be used to show the presence of bugs, but never to show their absence!”?

> > plus bump the version number in OpenQA/Isotovideo/Interface.pm
> 
> Is that really necessary? The behaviour doesn't really change in a meaningful way and the API remains completely unchanged.

But it *does* chance, just bump it, it's just a number :)

Vogtinator:
> > > However I would be fine to merge this if you could add test coverage for the function, e.g. in t/03-testapi.t,
> > 
> > 
> > I'll try that if it's not too involved, but I don't really understand why this is a requirement. Having a unit test doesn't really offset the potential that it might break some use of the function, does it?
> 
> Are you arguing with Edsger W. Dijkstra “Program testing can be used to show the presence of bugs, but never to show their absence!”?

Not necessarily. I'll try the argument one step further: If this change proves to be somehow worse than before (which I doubt), then writing a unittest for this would be a waste of time. From that PoV, it would actually make sense to apply the change first, wait a bit and only then add a test! Otherwise we'd test that the suboptimal behaviour is present...

> > > plus bump the version number in OpenQA/Isotovideo/Interface.pm
> > 
> > 
> > Is that really necessary? The behaviour doesn't really change in a meaningful way and the API remains completely unchanged.
> 
> But it _does_ chance, just bump it, it's just a number :)

Sure, just wondering.

Vogtinator:
Here's an attempt I spent way too much time on: https://github.com/Vogtinator/os-autoinst/tree/broken-useless-unit-test. I won't waste another millisecond on that, but feel free to continue on your own.

The test is orders of magnitude more fragile and prone to break than the method it's supposed to test. Given the simple logic in the method here, I'm now of the opinion that there shouldn't be a unit test for it at all.

okurz:
> Not necessarily. I'll try the argument one step further: If this change proves to be somehow worse than before (which I doubt), then writing a unittest for this would be a waste of time. From that PoV, it would actually make sense to apply the change first, wait a bit and only then add a test! Otherwise we'd test that the suboptimal behaviour is present...

Yes, tests prove that the current behaviour work as expected, not that it's optimal. I would say you don't need to come up with perfect tests covering the exact behaviour, at least just have some reasonable code that executes the touched code path, i.e. complete statement coverage.

Vogtinator:
> > Not necessarily. I'll try the argument one step further: If this change proves to be somehow worse than before (which I doubt), then writing a unittest for this would be a waste of time. From that PoV, it would actually make sense to apply the change first, wait a bit and only then add a test! Otherwise we'd test that the suboptimal behaviour is present...
> 
> Yes, tests prove that the current behaviour work as expected, not that it's optimal. I would say you don't need to come up with perfect tests covering the exact behaviour, at least just have some reasonable code that executes the touched code path, i.e. complete statement coverage.

Did you check the code in the branch I linked?

While originally I thought this would only improve performance, this might also improve https://progress.opensuse.org/issues/113447, e.g. https://openqa.opensuse.org/tests/2473779#step/bootloader_uefi/11.

Vogtinator:
> > > > Not necessarily. I'll try the argument one step further: If this change proves to be somehow worse than before (which I doubt), then writing a unittest for this would be a waste of time. From that PoV, it would actually make sense to apply the change first, wait a bit and only then add a test! Otherwise we'd test that the suboptimal behaviour is present...
> > > 
> > > 
> > > Yes, tests prove that the current behaviour work as expected, not that it's optimal. I would say you don't need to come up with perfect tests covering the exact behaviour, at least just have some reasonable code that executes the touched code path, i.e. complete statement coverage.
> > 
> > 
> > Did you check the code in the branch I linked?
> 
> I did. Not sure what you want me to tell you.

Whether it's on the "right track" or if not, what is wrong.

> > While originally I thought this would only improve performance, this might also improve https://progress.opensuse.org/issues/113447, e.g. https://openqa.opensuse.org/tests/2473779#step/bootloader_uefi/11.
> 
> Good point. Yes, can very well happen in such cases as well.

okurz:
> > > > > Not necessarily. I'll try the argument one step further: If this change proves to be somehow worse than before (which I doubt), then writing a unittest for this would be a waste of time. From that PoV, it would actually make sense to apply the change first, wait a bit and only then add a test! Otherwise we'd test that the suboptimal behaviour is present...
> > > > 
> > > > 
> > > > Yes, tests prove that the current behaviour work as expected, not that it's optimal. I would say you don't need to come up with perfect tests covering the exact behaviour, at least just have some reasonable code that executes the touched code path, i.e. complete statement coverage.
> > > 
> > > 
> > > Did you check the code in the branch I linked?
> > 
> > 
> > I did. Not sure what you want me to tell you.
> 
> Whether it's on the "right track" or if not, what is wrong.

To me
https://github.com/os-autoinst/os-autoinst/commit/17bd7a5094cd96d24b7b85b41a520bcfcc33625b
looks completely fine. I would go with it.

Vogtinator:
> To me [17bd7a5](https://github.com/os-autoinst/os-autoinst/commit/17bd7a5094cd96d24b7b85b41a520bcfcc33625b) looks completely fine. I would go with it.

Ok, done.

kalikiana (approved):

okurz (changes requested):
Ok, please also just bump the API version number, rest is fine

Vogtinator:
Ping. I'd like to see whether this improves the annoying issue with `bootloader_uefi` fails on aarch64.

Martchus (approved):

Vogtinator:
First fallout! https://openqa.opensuse.org/tests/2507164

I'll make a PR to address that.

Vogtinator:
> First fallout! https://openqa.opensuse.org/tests/2507164
> 
> I'll make a PR to address that.

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15347

DimStar77:
This also seems to be a fallout from here:
https://openqa.opensuse.org/tests/2507350#step/boot_into_snapshot/53

Vogtinator:
> This also seems to be a fallout from here: https://openqa.opensuse.org/tests/2507350#step/boot_into_snapshot/53

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15358

dzedro:
Another few https://openqa.suse.de/tests/9308870#step/change_desktop/17
https://openqa.suse.de/tests/9315850#step/grub2_test/80

Vogtinator:
> Another few https://openqa.suse.de/tests/9308870#step/change_desktop/17 https://openqa.suse.de/tests/9315850#step/grub2_test/80

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15361

bs-suse:
I've not checked if similar ones are already mentioned above, but these are the ones that let me stumble about this:
https://openqa.suse.de/tests/9339832#step/releasenotes/13
https://openqa.suse.de/tests/9339836#step/releasenotes/13
https://openqa.suse.de/tests/9339396#step/releasenotes/13
https://openqa.suse.de/tests/9339400#step/releasenotes/13

Vogtinator:
> I've not checked if similar ones are already mentioned above, but these are the ones that let me stumble about this: https://openqa.suse.de/tests/9339832#step/releasenotes/13 https://openqa.suse.de/tests/9339836#step/releasenotes/13 https://openqa.suse.de/tests/9339396#step/releasenotes/13 https://openqa.suse.de/tests/9339400#step/releasenotes/13

The code is already prepared to handle this, but it's disabled in textmode for some reason:

```
         if (!check_var('VIDEOMODE', 'text')) {
             # Make sure release notes window is shown to avoid sending key too early
             # It takes longer time to show multilple release notes for addons
             assert_screen([qw(release-notes-sle-ok-button release-notes-sle-close-button)], 300);
         }
```

I suggest to just drop the condition.

PR #2129, opened 2022-07-18
CI: Also run container build tests in pull requests on changes

okurz:
We already have container tests but so far ran these only after a push
to the master branch, likely due to them being potentially more
heavy-weight and long-running. What we can do is to run them
unconditionally on pushes to the master branch but also always run them
in pull request CI runs but only if the 'container/' path is touched,
e.g. changes to the Dockerfile files.


baierjan (approved):

kalikiana (approved):
Sounds good. I'm definitely happy if we can do this.

PR #2132, opened 2022-07-20
Make lowercase settings fatal

foursixnine:
Avoid red herring, by removing a warning that should be a fatal error.

https://progress.opensuse.org/issues/113528#note-9

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 28;
+our $version = 29;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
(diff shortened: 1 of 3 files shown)


okurz:
I am thinking, should we bump the testing
test API version?

foursixnine:
> I am thinking, should we bump the testing test API version?

makes sense, done

kalikiana (approved):

okurz (approved):
change is fine but for new behaviour enforcing stricter settings I would like to give more time for others to see this PR to be ready for that. Thinking of roughly 2 weeks before we merge this.

mimi1vx (approved):

grisu48:
Please let's not merge this in August when everyone is on vacation. A change like this might stress out the remaining skeleton crew more then it should.

I vote for postponing this until September when the vacation time is over.

kraih (approved):

grisu48 (approved):

kalikiana:
So it seems @grisu48 is in favor of (updating and) merging this change now. Note that the *not-ready* still keeps this from being merged. I would suggest to remove that **Monday September 5** if others are fine with it (since that's the start of the week).

punkioudi (approved):

kalikiana:
@mergifyio rebase

AdamWill:
> Shouldn't be (I'm downplaying it :p), I would not expect users to rely on lower case settings.

Well, this user was...in our tests we do [stuff like this](https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/lib/utils.pm#_718) as an internal state marker:

    set_var("_welcome_done", 1);

as it's entirely internal to the test logic I didn't bother hitting caps lock for those variables. :/ Guess I'll go update them all to upper case now.

PR #2133, opened 2022-07-25
wheels: Pick up Perl modules and tests

kalikiana:
- Add the lib path of each wheel to INC
- Lookup test modules in wheels
- Unit test a wheel with a test module and a Perl module

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -44,6 +44,9 @@ loadtest is called.
 =cut
 
 sub find_script ($script) {
+    if (defined(my $wheel = (glob Cwd::getcwd . "/*/tests/$script"))) {
+        return $wheel;
+    }
     my $casedir = $bmwqemu::vars{CASEDIR};
     my $script_override_path = join('/', $bmwqemu::vars{ASSETDIR} // '', 'other', $script);
     if (-f $script_override_path) {
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -112,15 +112,13 @@ subtest 'isotovideo with wheels' => sub {
     chdir($pool_dir);
     unlink('vars.json') if -e 'vars.json';
 
+    $bmwqemu::scriptdir = "$Bin/..";
     my $wheels_dir = "$data_dir";
     my $specfile = path("$wheels_dir/wheels.yaml");
     $specfile->spurt("wheels: [foo/bar]");
-    # also verify that isotovideo invokes the wheel code correctly
-    combined_like { isotovideo(
-            opts => "wheels_dir=$wheels_dir casedir=$data_dir/tests _exit_after_schedule=1") } qr@Invalid.*Missing property@, 'invalid YAML causes error';
+    throws_ok { checkout_wheels(
+            "$wheels_dir") } qr@Invalid.*Missing property@, 'invalid YAML causes error';
     $specfile->spurt("version: v99\nwheels: [foo/bar]");
-    # from here on we directly test wheels in a way we can mock
-    $bmwqemu::scriptdir = "$Bin/..";
     throws_ok { checkout_wheels(
             "$wheels_dir") } qr@Unsupported version@, 'unsupported version';
     $specfile->spurt("version: v0.1\nwheels: [https://github.com/foo/bar.git]");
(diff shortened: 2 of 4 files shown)


kalikiana:
As mentioned in the ticket, this is the follow-up to #2004 which includes coverage for executing Perl code from wheels. I ended up doing both in the same branch because I couldn't find a good way to execute just the wheel code otherwise.

t/14-isotovideo.t
```diff
@@ -145,6 +143,16 @@ subtest 'isotovideo with wheels' => sub {
     $specfile->remove;
     is(checkout_wheels("$wheels_dir"), 1, 'no wheels');
     is(scalar @repos, 5, 'git never called');
+
+    # also verify that isotovideo invokes the wheel code correctly
+    $bmwqemu::vars{CASEDIR} = "$data_dir/tests";
+    $specfile->spurt("version: v0.1\nwheels: [copy/writer]");
+    path($pool_dir, 'writer', 'lib', 'Copy', 'Writer')->make_path->child('Content.pm')->spurt("package Copy::Writer::Content; sub write {}; 1");
+    path($pool_dir, 'writer', 'tests', 'pen')->make_path->child('ink.pm')->spurt("use Mojo::Base 'basetest', -signatures; use Copy::Writer::Content 'write'; sub run (\$) {}; 1");
```
perlpunk, line 151: `write` is not a good function name because it's a builtin function. And it's also not exported yet.
For making sure the `run` function is actually called I guess I would make a `diag` call and check for that in the log.

autotest.pm
```diff
@@ -44,6 +44,9 @@ loadtest is called.
 =cut
 
 sub find_script ($script) {
+    foreach my $wheel (glob Cwd::getcwd . "/*/tests/$script") {
+        return $wheel;
```
Martchus, line 48: A loop with an unconditional return?

kalikiana: We discussed it in the daily. I'll change it to use if and [0] to access the first element

t/14-isotovideo.t
kalikiana: For the record, with exporter and exit_after_schedule the test works just fine and is a bit faster - it's just unfortunate that without it perl completely ignores exports, so this will need to be documented clearly for users

kalikiana:
> Cannot read /tmp/14-isotovideo.t-Wta0/pool/writer/lib/Copy/Writer/Content.pm: No such file or directory at /usr/lib/perl5/site_perl/5.36.0/Devel/Cover/Report/Codecovbash.pm line 76.

Discussed this with Tina. The file is probably getting picked up by codecov regardless of having been deleted. So 1) either we ignore files/folders based on a naming pattern like `fake-writer` 2) or the test needs to use files in the repo, similar to how `t/fake/...` is being used by some tests

perlpunk:
I also played with it locally, and my suggestion would be:
* Move the `pen/ink.pm` to a regular git file in `t/data/tests/pen/ink.pm`. No need to copy it in the test. Remove the line where the `ink.pm` is created.

* Rename the `Copy::Writer::Content` to something like `Temp::Copy::Writer::Content` and add `Temp/` to the `-ignore,external/|Temp/|...` in `tools/invoke-tests`

perlpunk:
To test locally:
```
tools/invoke-tests --coverage t/14-isotovideo.t
cover -report html_minimal  cover_db
```

okurz:
> > Cannot read /tmp/14-isotovideo.t-Wta0/pool/writer/lib/Copy/Writer/Content.pm: No such file or directory at /usr/lib/perl5/site_perl/5.36.0/Devel/Cover/Report/Codecovbash.pm line 76.
> 
> Discussed this with Tina. The file is probably getting picked up by codecov regardless of having been deleted. So 1) either we ignore files/folders based on a naming pattern like `fake-writer` 2) or the test needs to use files in the repo, similar to how `t/fake/...` is being used by some tests

Or just add in tools/invoke-tests:

```patch
-    select="($path_regex)(OpenQA|backend|consoles|ppmclibs)/|($path_regex)isotovideo|($path_regex)[^/]+\.pm,-ignore,external/|/CheckGitStatus.pm|$prove_path"
+    select="($path_regex)(OpenQA|backend|consoles|ppmclibs)/|($path_regex)isotovideo|($path_regex)[^/]+\.pm,-ignore,external/|/tmp|/CheckGitStatus.pm|$prove_path"

```

kalikiana:
> * Move the `pen/ink.pm` to a regular git file in `t/data/tests/pen/ink.pm`. No need to copy it in the test. Remove the line where the `ink.pm` is created.

For the record, I didn't realize Tina was testing with the wrong folder. The test from the wheel must reside in the wheel subfolder to match the "already cloned wheel" behavior. And so this isn't an option afterall.

t/14-isotovideo.t
```diff
@@ -285,4 +294,5 @@ done_testing();
 END {
     rmtree "$Bin/data/tests/product";
     unlink("$data_dir/wheels.yaml") if -e "$data_dir/wheels.yaml";
+    rmtree "$pool_dir/writer";
```
perlpunk, line 297: This is not necessary. The `$pool_dir` is a subdirectory of the tempdir which gets removed automatically at the end.

okurz (approved):

t/14-isotovideo.t
Martchus: I think it is actually generally a good idea to do cleanup before running tests because you can never rely on the cleanup that happens at the end of the test (e.g. the code you're testing is indeed broken and therefore the test crashes in the middle).

Martchus (approved):

t/14-isotovideo.t
perlpunk: `tempdir()` creates a directory with random characters, and I don't see why we should delete a directory in the `END` block if the top directory is removed automatically anyway. And I don\t understand how your comment about cleanup up at the beginning of the test relates to that.

We now have a useless `rmtree` there.

perlpunk:
See #2138

t/14-isotovideo.t
Martchus: resolved by https://github.com/os-autoinst/os-autoinst/pull/2138

PR #2135, opened 2022-07-29
Make DISCARD inside SUT unmap blocks in qcow2 file

marmarek:
This makes `fstrim` run inside SUT actually reduce the size of the disk
image. It is especially relevant when the disk image is later uploaded.

### DIFF
--- a/OpenQA/Qemu/BlockDev.pm
+++ b/OpenQA/Qemu/BlockDev.pm
@@ -129,7 +129,8 @@ sub gen_cmdline ($self) {
             join(',', ('driver=' . $self->driver,
                     'node-name=' . $self->node_name,
                     'file=' . $self->node_name . FILE_POSTFIX,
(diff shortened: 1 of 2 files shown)


Martchus (approved):
Seems reasonable

kalikiana (commented):
I couldn't find official documentation of the behavior and the unit tests don't actually test this behavior. So I suggest to monitor some jobs to see that they actually shrink

Martchus:
This is the official documentation: https://qemu-project.gitlab.io/qemu/tools/qemu-nbd.html?highlight=unmap#cmdoption-qemu-nbd-discard

marmarek:
> So I suggest to monitor some jobs to see that they actually shrink

I do run this patch on my openqa instance, and this change reduces disk image of some tests from 50G to 25GB (with compression disabled).

okurz (approved):

PR #2136, opened 2022-07-29
Fix re-connecting via VNC over WebSockets

Martchus:
* Before overriding the VNC hostname to `127.0.0.1` (to connect to
  `dewebsockify`), save the original hostname.
* When checking whether VNC over WebSockets should be enabled, compare the
  configured `VIRSH_GUEST` with the saved hostname and not the overridden
  one.
* This should fix the problem of the first test mentioned in
  https://progress.opensuse.org/issues/114820


perlpunk (approved):

okurz (approved):

okurz:
We haven't tested it manually but let's see

PR #2137, opened 2022-08-01
tools: Remove redundant 'select' for codecov

okurz:
'select' never worked as we intended anyway and does not have an effect
over 'ignore' that we still use.


okurz:
Related to https://github.com/os-autoinst/os-autoinst/pull/2133#issuecomment-1200939763

okurz:
Seems like coverage did not really change at all so this supports my hypothesis that the "select" statement is having no effect.

Though https://github.com/os-autoinst/os-autoinst/runs/7608400908?check_suite_focus=true#step:3:662 shows:

```
Devel::Cover: merging data for t/data/tests/product/main.pm into t/data/tests/main.pm
Devel::Cover: merging data for t/data/tests/foo.pm into t/data/tests/bar/baz.pm
Devel::Cover: ignoring extra statement
```

Let's check if other jobs have that as well

Martchus (approved):

okurz:
Fixed the "extra statement". PR is ready.

perlpunk (approved):

PR #2139, opened 2022-08-02
spec: Exclude t/27-consoles-vmware.t on OBS

kalikiana:
This unit test runs into the timeout on OBS very early and it doesn't look like a longer timeout would help.

See: https://progress.opensuse.org/issues/114881

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -173,8 +173,9 @@ Convenience package providing os-autoinst + s390 worker jumphost dependencies.
 # and exclude known flaky tests in OBS check
 # https://progress.opensuse.org/issues/52652
 # 07-commands: https://progress.opensuse.org/issues/60755
+# 27-consoles-vmware: https://progress.opensuse.org/issues/114881
 # 29-backend-driver: https://progress.opensuse.org/issues/105061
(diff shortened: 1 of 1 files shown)


Martchus (approved):
I couldn't find a solution earlier when looking into the problem as well so I suppose that's the best for now.

dist/rpm/os-autoinst.spec
```diff
@@ -181,6 +181,8 @@ done
 rm xt/00-tidy.t
 # Remove test relying on a git working copy
 rm xt/30-make.t
+# https://progress.opensuse.org/issues/114881
+rm t/27-consoles-vmware.t
```
okurz, line 185: please move it up into line 177

okurz (changes requested):

dist/rpm/os-autoinst.spec
kalikiana: I would rather keep it here, otherwise it becomes less obvious what ticket it relates to and future changes will cause conflicts

okurz: We already reference tickets for individual excluded modules so we should be consistent

okurz (approved):

PR #2141, opened 2022-08-03
Add capability to start vmware VMs with cloud init data

mloviska:
As per [VMware image and cloud-init](https://jira.suse.com/browse/TEAM-5569), we need to test
vmware with cloud init. This is possible only in case test backend injects cloud init data in
to the vmx file directly

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -522,7 +522,22 @@ __END"
     # define the new domain
     $self->run_cmd("virsh $remote_vmm define $xmlfilename") && die "virsh define failed";
     if ($self->vmm_family eq 'vmware') {
-        $self->run_cmd('echo bios.bootDelay = \"10000\" >> /vmfs/volumes/datastore1/openQA/' . $self->name . '.vmx', domain => 'sshVMwareServer');
+        my $vmx = sprintf('/vmfs/volumes/datastore1/openQA/%s.vmx', $self->name);
+
+        # set default boot delay
+        $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
+
+        # inject cloud init metadata and userdata required for the image if there are any
+        my $ci_meta = $bmwqemu::vars{CLOUD_INIT_META};
+        my $ci_user = $bmwqemu::vars{CLOUD_INIT_USER};
+        my $ci_encoding = $bmwqemu::vars{CLOUD_INIT_ENCODING};
+
+        if ($ci_meta && $ci_user && $ci_encoding) {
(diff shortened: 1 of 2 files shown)


mloviska:
Original PR: https://github.com/os-autoinst/os-autoinst/pull/2111
VR: http://kepler.suse.cz/tests/18390# (We will require test case modification, however it proves that cloud-init process was triggered)

mloviska:
Hello @kalikiana, @Martchus! 

Sorry, my previous branch got messed up and after a chit-chat with Cris I have pushed the updated code.

consoles/sshVirtsh.pm
```diff
@@ -522,7 +522,22 @@ __END"
...
-        $self->run_cmd('echo bios.bootDelay = \"10000\" >> /vmfs/volumes/datastore1/openQA/' . $self->name . '.vmx', domain => 'sshVMwareServer');
+        my $vmx = sprintf('/vmfs/volumes/datastore1/openQA/%s.vmx', $self->name);
+
+        # set default boot delay
+        $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
+
+        # inject cloud init metadata and userdata required for the image if there are any
+        my $ci_meta = $bmwqemu::vars{CLOUD_INIT_META};
+        my $ci_user = $bmwqemu::vars{CLOUD_INIT_USER};
+        my $ci_encoding = $bmwqemu::vars{CLOUD_INIT_ENCODING};
```
okurz, line 533: please either die if the according variables are not set or provide a good default.

okurz (changes requested):

consoles/sshVirtsh.pm
mloviska: Sorry, why would you like to add `die`? VMs should be also tested w/o cloud-init as well

okurz: But if you don't define those variables then undefined variables will be used in string concatenations causing at least Perl warnings if not further weird problems

mloviska: If I am not mistaken, `undef` is a falsy value, therefore the code should not get into the below condition and attempt to do string concat.

So it should behave like this, or no?

```
No warning in condition
```

```
#!/usr/bin/perl

use strict;
use warnings;

my $a = undef;

unless ($a) {
    print "No warning in condition\n";
}

if ($a) {
    print "Nothing\n";
}
```

Martchus: Yes, the if-block should do the trick.

Martchus (approved):

kalikiana (approved):

mloviska:
If there are no objections, could we possibly merge it?

consoles/sshVirtsh.pm
okurz: Yes, true. Sorry about that. Haven't looked two lines below :D

PR #2143, opened 2022-08-05
Implement --color=auto with pipe detection

kalikiana:
Related: https://progress.opensuse.org/issues/114878


isotovideo
```diff
@@ -22,6 +22,11 @@ Enable direct output to STDERR instead of autoinst-log.txt
 
 isotovideo will chdir to that directory on startup
 
+=item B<--color=[yes|no|auto]>
+
+Enable or disable color output explicitly. Defaults to "auto". Alternatively
```
perlpunk, line 27: As you implemented it, the default is `yes`.
`auto` usually means that colors should be printed only if the output goes to a terminal (not piped or redirected to a file).

So if we document that and call it `auto`, then it IMHO should behave like that.
Compare `man git-config` and look for `color.ui`

perlpunk: btw, during testing it seems both STDOUT and STDERR are seen as not printing to a terminal, so we can't really test that.

But regarding implementation, something like:
```
# if one of the two filehandles is not opened to a tty, don't use colors
if ($options{color} eq 'auto' and (not -t STDOUT or not -t STDERR)) {
    $ENV{ANSI_COLORS_DISABLED} = 1;
}
```

kalikiana: I extended the condition now with what you suggested. The "default" test is a TODO... since I don't know how we can test it, but this way at least the code paths gets exercised...

perlpunk: Ok. With the changed default (currently we always get colored output), it would be good to run an openQA test to ensure we still get a colored autoinst-log.txt

mimi1vx:
Where is handled 'NO_COLOR' ?

perlpunk:
> Where is handled 'NO_COLOR' ?

It's handled by Term::ANSIColor, but only starting with [version 5.01](https://metacpan.org/dist/Term-ANSIColor/changes#L5) which was [first released with perl 5.32](http://perlpunks.de/corelist/mversion?module=Term%3A%3AANSIColor)

If we fully want to support it, we would need to unset `NO_COLOR` in case of `--color yes` 

https://metacpan.org/pod/Term::ANSIColor#ENVIRONMENT
> Programs that want to enable color despite NO_COLOR being set will need to unset that environment variable

kalikiana:
Discussed this briefly in the daily:
- I will add a line to unset NO_COLOR
- I will open a PR for openQA to set --color=yes since the default would otherwise switch off colors - I'm going for that simply because I consider it the most common default and others had no strong opinion either

t/14-isotovideo.t
```diff
@@ -51,6 +52,20 @@ subtest 'get the version number' => sub {
...
 
+subtest 'color output can be configured via the command-line' => sub {
+    chdir($pool_dir);
+    unlink('vars.json') if -e 'vars.json';
+    my $out = stderr_from { isotovideo(opts => "--color=yes casedir=$data_dir/tests schedule=foo,bar/baz _exit_after_schedule=1") };
+    isnt($out, colorstrip($out), 'logs use colors when requested');
+    $out = stderr_from { isotovideo(opts => "--color=no casedir=$data_dir/tests schedule=foo,bar/baz _exit_after_schedule=1") };
+    is($out, colorstrip($out), 'no colors in logs');
+  TODO: {
+        local $TODO = 'STDERR and STDOUT are always absent here';
```
Martchus, line 63: I doubt STDERR/STDOPUT are completely absent here. They're just not referring to a terminal. Not sure whether we would ever add such a test. So maybe it would be better to remove the whole TODO block.

kalikiana: That's what I mean by absent. I don't know how exactly it is implemented.

The reason to have it is coverage of the code path.

isotovideo
```diff
@@ -104,10 +109,18 @@ sub version () {
...
-GetOptions(\%options, 'debug|d', 'workdir=s', 'help|h|?', 'version|v') or usage(1);
+GetOptions(\%options, 'debug|d', 'workdir=s', 'color=s', 'help|h|?', 'version|v') or usage(1);
 usage(0) if $options{help};
 version() if $options{version};
 
+my $color = $options{color} // 'auto';
+# Default to no color if there is no terminal
+$color = 'no' if $color eq 'auto' and (not -t STDOUT or not -t STDERR);
+# User setting has preference, see https://no-color.org/
+delete $ENV{NO_COLOR} if $color eq 'yes';
```
kalikiana, line 120: Note to self, os-autoinst/openQA/pull/4771 now checks the interface version to pass `--color=no`, so this branch should bump it

isotovideo
```diff
@@ -104,10 +109,18 @@ sub version () {
...
 }
 
-GetOptions(\%options, 'debug|d', 'workdir=s', 'help|h|?', 'version|v') or usage(1);
+GetOptions(\%options, 'debug|d', 'workdir=s', 'color=s', 'help|h|?', 'version|v') or usage(1);
 usage(0) if $options{help};
 version() if $options{version};
 
+my $color = $options{color} // 'auto';
+# Default to no color if there is no terminal
+$color = 'no' if $color eq 'auto' and (not -t STDOUT or not -t STDERR);
```
okurz, line 118: I wonder about this. Do openQA tests still have color recorded in logs? Is that what https://github.com/os-autoinst/openQA/pull/4771 tries to do? Because we should preserve the behaviour.

t/14-isotovideo.t
```diff
@@ -51,6 +52,20 @@ subtest 'get the version number' => sub {
...
+        $out = stderr_from { isotovideo(opts => "--color=auto casedir=$data_dir/tests schedule=foo,bar/baz _exit_after_schedule=1") };
+        isnt($out, colorstrip($out), 'logs use colors by default');
+    }
```
okurz, line 66: I don't feel like the TODO sections in tests work out for us. We should only merge changes without new TODO blocks in tests.

okurz (changes requested):

okurz:
I would suggest to go without the auto-option and console detection and reliance on openQA for now: https://github.com/os-autoinst/os-autoinst/pull/2144

When I say "for now" I mean for days/weeks. Let's get the option in first and experiment with the automatic stuff later

okurz:
Please rebase after we got https://github.com/os-autoinst/os-autoinst/pull/2144 out of the way :) As alternative you can still decide to not go that route and leave out any automatic detection for now.

kalikiana:
I changed the code to use IO::Interactive. But then I remembered the mock won't work unless the code moves somewhere else. I'm leaving it for now since I don't want to introduce a helper function just for testing.

okurz:
As you resolved https://progress.opensuse.org/issues/114878, can you close this draft PR as well as the corresponding one in openQA?

PR #2145, opened 2022-08-10
 Laptop: switch laptop hardware to HP EliteBook 820 G1 model

DimStar77:
Dell laptops try to load a kernel module (dell_smm_hwmon) which is
crashing when loading inside qemu

https://bugzilla.opensuse.org/show_bug.cgi?id=1200616


Martchus (approved):
Makes sense, I suppose. Do you have a verification run?

DimStar77:
> Makes sense, I suppose. Do you have a verification run?

Not yet - Work in progress

Vogtinator:
> > Makes sense, I suppose. Do you have a verification run?
> 
> Not yet - Work in progress

I copied the data files to openqaworker4 and did `openqa-clone-job --within-instance https://openqa.opensuse.org 2506258 _GROUP=38 WORKER_CLASS=openqaworker4 LAPTOP=hp_elitebook_820g1`

https://openqa.opensuse.org/tests/2507949#live

DimStar77:
> https://openqa.opensuse.org/tests/2507949#live

VR passed

Vogtinator:
So this turns out to be a bug in the KVM host kernels, fixed by https://www.spinics.net/lists/kvm/msg280860.html.

okurz:
> So this turns out to be a bug in the KVM host kernels, fixed by https://www.spinics.net/lists/kvm/msg280860.html.

do you still want to have this merged though?

Vogtinator:
IMO yes, as it will take a while for the fix to end up on the workers and it's better not to load modules not meant for the actual "hardware" anyway. And who knows, maybe openQA will find an issue with HP drivers next!

PR #2140, opened 2022-08-02
Fix CPUs assignment in QEMU 6.20

acerv:
QEMU 6.2.0 assigns all CPUs to NUMA node 0 by default when QEMU_NUMA=1
is defined.

Kernel boot output on QEMU 5.2.0 with QEMU_NUMA=1 (correct behavior):

    [    0.003248] numa: Node 0 CPUs: 0
    [    0.003251] numa: Node 1 CPUs: 1
Kernel boot output on QEMU 6.2.0 with QEMU_NUMA=1 (wrong behavior):

    [    0.003181] numa: Node 0 CPUs: 0-1
    [    0.003185] numa: Node 1 CPUs:

PR #2146, opened 2022-08-12
Implement QEMU backend support for arbitrary CPU topology

mdoucha:
Add job settings `QEMUSOCKETS`, `QEMUDIES`, `QEMUCLUSTERS` and `QEMUCORES` for configuring arbitrary CPU topology.

It turns out that PR #2140 was not sufficient to fix the NUMA issues. It helped on PPC64LE but now other archs are complaining about invalid NUMA configuration during boot. The underlying cause is that [QEMU 6.2 changed the default CPU topology](https://www.qemu.org/docs/master/system/invocation.html) from creating multiple CPU sockets to creating multiple cores on the same socket. So let's add the necessary configuration options to create the CPU topology we need in NUMA tests.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -872,7 +872,14 @@ sub start_qemu ($self) {
         }
 
         sp('device', 'usb-kbd') if $use_usb_kbd;
-        sp('smp', $vars->{QEMUTHREADS} ? [qv "$vars->{QEMUCPUS} threads=$vars->{QEMUTHREADS}"] : $vars->{QEMUCPUS});
+
+        my $smp_config = [$vars->{QEMUCPUS}];
+        for my $key (qw(QEMUSOCKETS QEMUDIES QEMUCLUSTERS QEMUCORES QEMUTHREADS)) {
+            my $qkey = lc($key =~ s/^QEMU//r);
+            push @$smp_config, "$qkey=$vars->{$key}" if exists $vars->{$key};
+        }
+        sp('smp', $smp_config);
+
         if ($vars->{QEMU_NUMA}) {
             for my $i (0 .. ($vars->{QEMUCPUS} - 1)) {
                 my $m = int($vars->{QEMURAM} / $vars->{QEMUCPUS});
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -135,12 +135,16 @@ QEMU_QMP_CONNECT_ATTEMPTS;integer;20;The number of attempts to connect to qemu q
 PATHCNT;integer;2;Number of paths in MULTIPATH scenario
 PXEBOOT;boolean or 'once';0;Boot VM from network, on every boot or only once if set to 'once'
 QEMU;QEMU binary filename;undef;Filename of QEMU binary to use
+QEMUCLUSTERS;integer;undef;Number of CPU clusters used by VM
+QEMUCORES;integer;undef;Number of CPU cores used by VM
 QEMUCPU;see qemu -cpu ?;undef;CPU to emulate
 QEMUCPUS;integer;1;Number of CPUs to assign to VM
+QEMUDIES;integer;undef;Number of CPU dies used by VM
 QEMUMACHINE;see qemu -machine ?;undef;Machine and chipset to emulate
 QEMUPORT;integer;20002 + worker instance * 10;Port on which QEMU monitor should listen
 QEMURAM;integer;1024;Size of RAM of VM in MiB
-QEMUTHREADS;integer;0;Number of cpu threads used by VM
+QEMUSOCKETS;integer;undef;Number of CPU sockets used by VM
+QEMUTHREADS;integer;undef;Number of CPU threads used by VM
 QEMUTPM;'instance' or appropriate value for local swtpm config;undef;Configure VM to use a TPM emulator device, with appropriate args for the arch. If a TPM device is available at QEMUTPM_PATH_PREFIX + X, where X is the value of QEMUTPM or the worker instance number if QEMUTPM is set to 'instance', it will be used. Otherwise it will be created at test startup. See QEMUTPM_VER in the latter case.
 QEMUTPM_VER;'1.2' or '2.0' depending on which TPM chip should be emulated;'2.0';If no TPM device has been setup otherwise, swtpm will be used internally to create one with a socket at /tmp/mytpmX
 QEMUTPM_PATH_PREFIX;string;'/tmp/mytpm';Path prefix to use or create TPM emulator device in


mimi1vx (approved):

backend/qemu.pm
```diff
@@ -872,7 +872,15 @@ sub start_qemu ($self) {
         }
 
         sp('device', 'usb-kbd') if $use_usb_kbd;
-        sp('smp', $vars->{QEMUTHREADS} ? [qv "$vars->{QEMUCPUS} threads=$vars->{QEMUTHREADS}"] : $vars->{QEMUCPUS});
+
+        my $smp_config = [$vars->{QEMUCPUS}];
+
+        for my $qkey (qw(sockets dies clusters cores threads)) {
```
perlpunk, line 878: Not sure what others prefer, but I like when I can grep the code for a var name directly.
So writing `qw(QEMUSOCKETS QEMUDIES QEMUCLUSTERS QEMYCORES QEMUTHREADS)` is what I would do here.

backend/qemu.pm
```diff
@@ -872,7 +872,15 @@ sub start_qemu ($self) {
         }
 
         sp('device', 'usb-kbd') if $use_usb_kbd;
-        sp('smp', $vars->{QEMUTHREADS} ? [qv "$vars->{QEMUCPUS} threads=$vars->{QEMUTHREADS}"] : $vars->{QEMUCPUS});
+
+        my $smp_config = [$vars->{QEMUCPUS}];
+
```
Martchus, line 877:
```suggestion
```

backend/qemu.pm
```diff
@@ -872,7 +872,15 @@ sub start_qemu ($self) {
...
+        for my $qkey (qw(sockets dies clusters cores threads)) {
+            my $key = "QEMU" . uc $qkey;
+            push @$smp_config, "$qkey=$vars->{$key}" if exists $vars->{$key};
+        }
+
+        sp('smp', $smp_config);
```
Martchus, line 883:
```suggestion
        sp('smp', $smp_config);

```

Martchus (commented):
The change itself looks good. It would be nice if the PR description would be also in the commit message.

backend/qemu.pm
Martchus: But `$qkey` is also used as-is. Converting it the other way around is likely more annoying. So for me it looks acceptable as it is.

perlpunk: oh, you're right. I missed that one.

mdoucha: Changed it anyway, being able to grep for the variables is good.

backend/qemu.pm
mdoucha: Fixed.

backend/qemu.pm
mdoucha: Fixed.

okurz (changes requested):
What you put into the description of the pull request could also be in the commit message details. I don't know how you created your pull request but your git commit messages has *only* the subject line while the pull request description has more details. Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage.

I can recommend the tool [hub](https://github.com/github/hub) (zypper in rubygem-hub) for easier PR creation. Also, I myself use a script [git-pr-last](http://github.com/okurz/scripts/tree/master/git-pr-last) to create a PR with proper description for these simple one-commit PRs:
```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
git push $target && git show --no-patch --format=%B | hub pull-request -F -
```

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

mdoucha:
> What you put into the description of the pull request could also be in the commit message details. I don't know how you created your pull request but your git commit messages has _only_ the subject line while the pull request description has more details. Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage.

The commit message describes what's in the commit and why. We do not need to bore future generations with a backstory about sudden change of defaults in a completely different project.

backend/qemu.pm
```diff
@@ -872,7 +872,14 @@ sub start_qemu ($self) {
         }
 
         sp('device', 'usb-kbd') if $use_usb_kbd;
-        sp('smp', $vars->{QEMUTHREADS} ? [qv "$vars->{QEMUCPUS} threads=$vars->{QEMUTHREADS}"] : $vars->{QEMUCPUS});
+
+        my $smp_config = [$vars->{QEMUCPUS}];
+        for my $key (qw(QEMUSOCKETS QEMUDIES QEMUCLUSTERS QEMUCORES QEMUTHREADS)) {
+            my $qkey = lc(substr $key, 4);
```
okurz, line 878:
```suggestion
            my $qkey = lc(key =~ s/QEMU//r);
```

to make it clear that the qemu prefix should be stripped regardless of length

doc/backend_vars.asciidoc
```diff
@@ -135,12 +135,16 @@ QEMU_QMP_CONNECT_ATTEMPTS;integer;20;The number of attempts to connect to qemu q
 PATHCNT;integer;2;Number of paths in MULTIPATH scenario
 PXEBOOT;boolean or 'once';0;Boot VM from network, on every boot or only once if set to 'once'
 QEMU;QEMU binary filename;undef;Filename of QEMU binary to use
+QEMUCLUSTERS;integer;0;Number of CPU clusters used by VM
+QEMUCORES;integer;0;Number of CPU cores used by VM
```
okurz, line 139: Is 0 really the default? I guess the default is just an null-string reverting to whatever qemu defaults to which is likely "1" in this case, isn't it?

okurz (changes requested):

backend/qemu.pm
perlpunk: In that case I suggest to add `^`: `s/^QEMU//r` (I know it's not needed here but it says more explicitly what it does and it might be a nanosexond faster)

mdoucha: Fixed.

doc/backend_vars.asciidoc
mdoucha: Changed all 5 default values to `undef`.

okurz (approved):

mdoucha:
Something broke in CI

    Reading installed packages...
    Package 'perl-DBD-Pg-3.15.1' not found.
    Error: Process completed with exit code 104.

perlpunk:
> Something broke in CI
> 
> ```
> Reading installed packages...
> Package 'perl-DBD-Pg-3.15.1' not found.
> Error: Process completed with exit code 104.
> ```

Oh, that needs a dependency update. Will trigger the dependency bot now

Martchus (approved):

perlpunk:
So this needs to wait for https://github.com/os-autoinst/openQA/pull/4775

perlpunk:
I retriggered the fullstack test

PR #2149, opened 2022-08-18
testapi: Allow forcing worker ip in autoinst_url()

asdil12:
Ticket: https://progress.opensuse.org/issues/115478
Verification: http://kazhua.qa/tests/141#step/zypper_log_packages/23
Related: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15379

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
 Return VM's host IP.
 In a kvm instance you reach the VM's host under default 10.0.2.2
 
+Optional named parameter C<inside_sut> in C<$args> will force using actual worker IP/hostname
+even on KVM guests if set to 0.
+
 =cut
 
-sub host_ip () { check_var('BACKEND', 'qemu') ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
+sub host_ip ($args = {}) { (check_var('BACKEND', 'qemu') && ($args->{inside_sut} // 1)) ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
 
 =head2 autoinst_url
 
-  autoinst_url([$path, $query]);
+  autoinst_url([$path, $query, $args]);
 
 returns the base URL to contact the local C<os-autoinst> service
 
@@ -2142,16 +2145,19 @@ Optional C<$path> argument is appended after base url.
 Optional HASHREF C<$query> is converted to URL query and appended
 after path.
 
+Optional named parameter C<inside_sut> in C<$args> will force using actual worker IP/hostname
+even on KVM guests if set to 0.
+
 Returns constructor URL. Can be used inline:
 
   script_run("curl " . autoinst_url . "/data");
 
 =cut
 
-sub autoinst_url ($path = undef, $query = undef) {
+sub autoinst_url ($path = undef, $query = undef, $args = {}) {
     $path //= '';
     $query //= {};
-    my $hostname = get_var('AUTOINST_URL_HOSTNAME', host_ip());
+    my $hostname = get_var('AUTOINST_URL_HOSTNAME', host_ip($args));
     # QEMUPORT is historical for the base port of the worker instance
     my $workerport = get_var("QEMUPORT") + 1;
 


testapi.pm
```diff
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
 Return VM's host IP.
 In a kvm instance you reach the VM's host under default 10.0.2.2
 
+Optional C<$force_worker_hostname> argument will force using actual worker IP
+even on KVM guests.
+
 =cut
 
-sub host_ip () { check_var('BACKEND', 'qemu') ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
+sub host_ip ($force_worker_hostname = 0) { (check_var('BACKEND', 'qemu') && !$force_worker_hostname) ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
```
Martchus, line 2135: It'll likely be called like `autoinst_url($path, $query, 1)` on your side. Unfortunately the meaning of `1` is not very clear then. So maybe make it a named parameter?

```suggestion
sub host_ip ($args = {}) { (check_var('BACKEND', 'qemu') && !$args->{force_worker_hostname}) ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
```

testapi.pm
```diff
@@ -2142,16 +2145,19 @@ Optional C<$path> argument is appended after base url.
...
+even on KVM guests.
+
 Returns constructor URL. Can be used inline:
 
   script_run("curl " . autoinst_url . "/data");
 
 =cut
 
-sub autoinst_url ($path = undef, $query = undef) {
+sub autoinst_url ($path = undef, $query = undef, $force_worker_hostname = 0) {
```
Martchus, line 2157:
```suggestion
sub autoinst_url ($path = undef, $query = undef, $args = {}) {
```

testapi.pm
```diff
@@ -2142,16 +2145,19 @@ Optional C<$path> argument is appended after base url.
...
     $path //= '';
     $query //= {};
-    my $hostname = get_var('AUTOINST_URL_HOSTNAME', host_ip());
+    my $hostname = get_var('AUTOINST_URL_HOSTNAME', host_ip($force_worker_hostname));
```
Martchus, line 2160:
```suggestion
    my $hostname = get_var('AUTOINST_URL_HOSTNAME', host_ip($args));
```

testapi.pm
```diff
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
 Return VM's host IP.
 In a kvm instance you reach the VM's host under default 10.0.2.2
 
+Optional C<$force_worker_hostname> argument will force using actual worker IP
+even on KVM guests.
```
Martchus, line 2131:
```suggestion
Optional C<$force_worker_hostname> argument will force using actual worker IP/hostname
even on KVM guests.
```

Maybe this could also state *why* this is possibly useful.

Vogtinator:
IMO it would be a nicer API to split this into two functions or indicate the use case better. By default, it should return a URL which is usable from the SUT, but on request return a URL usable from test module code. Something like `autoinst_url(inside_sut => 0)`. Not only does that make it more clear what it's actually doing in the function itself, but it's also more obvious on the caller side.

Martchus (changes requested):
I like the idea from @Vogtinator 's comment. It also makes my request for improving the documentation unnecessary.

asdil12:
New VR: http://kazhua.qa/tests/145

testapi.pm
```diff
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
 Return VM's host IP.
 In a kvm instance you reach the VM's host under default 10.0.2.2
 
+Optional C<inside_sut> kwarg will force using actual worker IP/hostname
+even on KVM guests if set to 0.
+
 =cut
 
-sub host_ip () { check_var('BACKEND', 'qemu') ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
+sub host_ip ($args = {}) { (check_var('BACKEND', 'qemu') && $args->{inside_sut} // 1) ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
```
Martchus, line 2135: This should be correct, however, some braces to make the operator precedence more clear wouldn't hurt:
```suggestion
sub host_ip ($args = {}) { (check_var('BACKEND', 'qemu') && ($args->{inside_sut} // 1)) ? get_var('QEMU_HOST_IP', '10.0.2.2') : get_required_var('WORKER_HOSTNAME') }
```

testapi.pm
```diff
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
...
 
 =head2 autoinst_url
 
-  autoinst_url([$path, $query]);
+  autoinst_url([$path, $query, $force_worker_hostname]);
```
Martchus, line 2139: This needs to be updated.

testapi.pm
```diff
@@ -2127,13 +2127,16 @@ sub diag (@args) { bmwqemu::diag(@args) }
 Return VM's host IP.
 In a kvm instance you reach the VM's host under default 10.0.2.2
 
+Optional C<inside_sut> kwarg will force using actual worker IP/hostname
```
Martchus, line 2130: This is not Python and even if it was "kwarg" is an awkward abbreviation.

```suggestion
Optional named parameter C<inside_sut> will force using actual worker IP/hostname
```

Martchus (changes requested):

kalikiana (approved):

Vogtinator:
Apparently `WORKER_HOSTNAME` is not always available: https://openqa.opensuse.org/tests/2579388#step/zypper_log_packages/10

Maybe that should just use `127.0.0.1` or similar, if the worker also listens there?

PR #2150, opened 2022-08-18
Improve test coverage of `baseclass.pm`

Martchus:
See https://progress.opensuse.org/issues/111254 and particular commit messages

### DIFF
--- a/OpenQA/Benchmark/Stopwatch.pm
+++ b/OpenQA/Benchmark/Stopwatch.pm
@@ -158,7 +158,7 @@ sub summary ($self) {
 
         my $duration = $event->{time} - $prev_time;
         my $cumulative = $event->{time} - $self->{start};
-        my $percentage = ($duration / $self->total_time) * 100;
+        my $percentage = $self->total_time > 0 ? (($duration / $self->total_time) * 100) : 0;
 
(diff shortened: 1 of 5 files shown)


OpenQA/Benchmark/Stopwatch.pm
```diff
@@ -158,7 +158,7 @@ sub summary ($self) {
 
         my $duration = $event->{time} - $prev_time;
         my $cumulative = $event->{time} - $self->{start};
-        my $percentage = ($duration / $self->total_time) * 100;
+        my $percentage = $self->total_time > 0 ? (($duration / $self->total_time) * 100) : 0;
```
mimi1vx, line 161: hmm , how perl divide by zero? Nice catch.

mimi1vx (approved):

OpenQA/Benchmark/Stopwatch.pm
Martchus: In my test the execution time is basically zero, so I've noticed that issue. In Perl you actually get an exception when trying to divide by zero (in contrast to JavaScript where you get Infinity which is reasonable as well).

kalikiana (approved):
Nice!

PR #2151, opened 2022-08-19
Fix nr. of keys sent in send_key_until_needlematch

ilmanzo:
Related ticket: https://progress.opensuse.org/issues/107749 

the function `send_key_until_needlematch()` is documented to send the key at maximum _n_ times, but is sending the key _n+1_ times before failing. By looking at the test [`03-testapi.t`](https://github.com/os-autoinst/os-autoinst/blob/515dc27b2e6625c01b3a5b6d90b441cda2d5bba2/t/03-testapi.t#L950) , 'esc' key is sent two times and then [checked for three times](https://github.com/os-autoinst/os-autoinst/blob/515dc27b2e6625c01b3a5b6d90b441cda2d5bba2/t/03-testapi.t#L962). 

This is an issue for some scenarios, for example, if an even number of keypresses is required, as is the case of maximizing and unmaximizing a window, by sending an even number of alt-f10.

This PR aims to correct both the unit test and the API behavior.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -1365,7 +1365,7 @@ sub release_key {
 
   send_key_until_needlematch($tag, $key [, $counter, $timeout]);
 
-Send specific key until needle with C<$tag> is not matched or C<$counter> is 0.
+Send specific key until needle with C<$tag> is matched or C<$counter> is 0.
 C<$tag> can be string or C<ARRAYREF> (C<['tag1', 'tag2']>)
 Default counter is 20 steps, default timeout is 1s
 
@@ -1384,7 +1384,7 @@ sub send_key_until_needlematch {
         wait_screen_change {
             send_key $key;
         };
-        if (!$counter--) {
+        if (--$counter <= 0) {
             assert_screen $tag, 1;
         }
         $real_timeout = $timeout;
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 27;
+our $version = 28;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -947,7 +947,7 @@ subtest 'send_key_until_needlematch' => sub {
     $mock_testapi->unmock('send_key');
     $fake_needle_found = $fake_needle_found_after_pause = 0;
     $cmds = [];
-    throws_ok(sub { send_key_until_needlematch('tag', 'esc', 2); },
+    throws_ok(sub { send_key_until_needlematch('tag', 'esc', 3); },
         qr/assert_screen reached/,
         'no candidate needle matched tags'
     );


testapi.pm
```diff
@@ -1384,7 +1384,7 @@ sub send_key_until_needlematch {
         wait_screen_change {
             send_key $key;
         };
-        if (!$counter--) {
+        if (!--$counter) {
```
Martchus, line 1387:
```suggestion
        if (--$counter <= 0) {
```
So we get into this condition even if somebody passes 0.

Martchus (commented):
I suppose this change is in line with what is already documented so I guess it looks good to merge.

testapi.pm
ilmanzo: I guess you mean `>= 0`  ?

ilmanzo: Good idea! I'd also propose a check if passed parameter '$counter' is >0 ; passing 0 or negative number does not make any sense in my opinion.  What do you think ?

perlpunk (approved):

ilmanzo:
to avoid breaking existing test suites, we are going to [fix all callers](https://progress.opensuse.org/issues/115619) ; I'd suggest to merge / release this change in the same timeframe ,

Martchus:
Ok, then I'm making it as not-ready for now to prevent the auto-merge.

Martchus (approved):

testapi.pm
Martchus: It should be good enough now.

jknphy:
Hi @Martchus how we can sync the deployment in openQA of this change with the other PR for the use the function?
I'm thinking to merge it on Monday for example (not on Friday :) I guess) Should we ping tools squad for manual deployment? otherwise there could be many test failing during the gap, wdyt?

Martchus:
Since packages need to be rebuilt and an out-of-schedule deployment needed to be triggered manually, there might be some time between merging a PR here and its actual deployment. Normally we solve this by avoiding breaking changes in the first place. I haven't thought of this change as one which would affect many tests as I assumed the new behavior is what people expected to happen anyways.

I suppose we could still do it but then the other way around. This change is deployed first and once you get the deployment e-mail on openqa AT suse… you can merge the test PR (which should be deployed rather quickly).

jknphy:
> een merging a PR here and its actual deployme

what deployment email? is there any slack channel?

perlpunk:
Normally we could check `$OpenQA::Isotovideo::Interface::version >= 28`, right? But I guess that's not very nice for a change with over 200 lines

perlpunk:
@jknphy there will be an email sent to the openqa mailing list after deployment.

perlpunk:
I retriggered the Tumbleweed OBS test, there is already a ticket about [the failing 27-consoles-vmware.t test](https://progress.opensuse.org/issues/114881)

b10n1k (approved):
+1

jknphy:
can this be merged without that OBS check? is it just about to remove the label? I don't see any merge button (but maybe I don't have permission in this repo, not sure or that is not visible with the automerge in place).

Martchus:
We could merge it without the OBS check. However, it would be cleaner if you'd rebase against master. (The problem with the test should hopefully already fixed by another PR that has been merged meanwhile.)

jknphy (approved):

Martchus:
Yes, we can remove labels and merge it.

ilmanzo:
So... I'd ask whoever has permission to remove the label and merge; many thanks

ge0r:
> So... I'd ask whoever has permission to remove the label and merge; many thanks

Since it was @Martchus that added the label, could you merge this please?

PR #2152, opened 2022-08-24
Check if SSH password is defined explicitly

sdclarke:
When creating a new SSH connection, if the password is an empty string
the statement `if ($args{password})` will be false

Using `defined()` will instead allow for the correct behaviour, meaning
an empty password can be used correctly

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1143,7 +1143,7 @@ sub new_ssh_connection ($self, %args) {
     while ($counter > 0) {
         if ($ssh->connect($args{hostname}, $args{port})) {
 
-            if ($args{password}) {
+            if (defined($args{password})) {
                 $ssh->auth(username => $args{username}, password => $args{password});
             }
             else {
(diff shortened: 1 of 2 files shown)


perlpunk:
Thanks! I added a test for this and pushed to this branch.

perlpunk (approved):

Martchus (approved):

marmarek:
This breaks using ssh-agent completely, because the password defaults to `testapi::password` in https://github.com/os-autoinst/os-autoinst/blob/master/consoles/sshSerial.pm#L28. Can you please revert, or maybe add an explicit option to use the agent?

Martchus:
`undef testapi::password;` should still work but this might have not been the best idea, indeed.

marmarek:
Indeed, that is not the best idea, I see use the `testapi::password` for other console types (virtio, graphical login etc).

okurz:
I created https://progress.opensuse.org/issues/117784 to work on that quickly

Martchus:
To summarize:
* This change allows one to pass an empty sting or 0 as password. This seems like a valid use-case.
* I assume you used an empty string or 0 before to use ssh-agent. That makes sense as well but of course breaks now.

Maybe it would be the nicest to simply remove the fallback to `testapi::password`. I just don't know whether something currently relies on that fallback. So for the sake of not breaking more it would likely be better to add another option to use the SSH agent explicitly (and it is not the worst because explicitness makes the code usually also easier to read).

PR #2155, opened 2022-08-29
Extend unit tests of `baseclass.pm` and add missinng `use needle`

Martchus:
See https://progress.opensuse.org/issues/111254

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1306,7 +1306,7 @@ sub _stop_children_processes ($self) {
             $ret = waitpid($pid, WNOHANG);
             bmwqemu::diag "waitpid for $pid returned $ret";
             last if ($ret == $pid);
-            sleep 1;
+            sleep 1;    # uncoverable statement
         }
     }
 }
@@ -1315,10 +1315,10 @@ sub _child_process ($self, $code) {
     die "Can't spawn child without code" unless ref($code) eq "CODE";
 
     my $pid = fork();
-    die "fork failed" unless defined($pid);
+    die "fork failed" unless defined($pid);    # uncoverable statement
 
     if ($pid == 0) {
-        $code->();
+        $code->();    # uncoverable statement
     }
     else {
         push @{$self->{children}}, $pid;
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -20,6 +20,7 @@ use POSIX qw(_exit :sys_wait_h);
 use IO::Select;
 require IPC::System::Simple;
 use myjsonrpc;
+use needle;
 use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
 use OpenQA::Benchmark::Stopwatch;
 use MIME::Base64 'encode_base64';
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -16,7 +16,7 @@ use File::Copy 'cp';
 use File::Basename;
 use Time::HiRes qw(gettimeofday time tv_interval);
 use Try::Tiny;
-use POSIX qw(_exit :sys_wait_h);
+use POSIX qw(_exit waitpid WNOHANG);
 use IO::Select;
 require IPC::System::Simple;
 use myjsonrpc;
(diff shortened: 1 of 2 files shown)


baierjan (approved):

kalikiana (approved):

t/23-baseclass.t
```diff
@@ -617,4 +617,106 @@ subtest 'bouncer functions' => sub {
...
+    };
+
+    subtest 'multiple tags specified, multiple needles set for assertion' => sub {
+        my @tags = (qw(inst-welcome not-existing));
+        my $res = $baseclass->set_tags_to_assert({mustmatch => \@tags});
+        is_deeply $res, {tags => \@tags}, 'tags returned' or diag explain $res;
+        my $needles = $baseclass->assert_screen_needles;
+        is scalar @$needles, 2, 'matching needles assigned';
+        is $needles->[0]->{name}, 'welcome.ref', 'needle welcome.ref matched';
+        is $needles->[1]->{name}, 'inst-welcome-20140902', 'needle inst-welcome-20140902 matched';
```
kalikiana, line 703: 7:         #   Failed test 'needle welcome.ref matched'
7:         #   at ./t/23-baseclass.t line 702.
7:         #          got: 'inst-welcome-20140902'
7:         #     expected: 'welcome.ref'
7: 
7:         #   Failed test 'needle inst-welcome-20140902 matched'
7:         #   at ./t/23-baseclass.t line 703.
7:         #          got: 'welcome.ref'
7:         #     expected: 'inst-welcome-20140902'
7:         # Looks like you failed 2 tests of 4.
7: 
7:     #   Failed test 'multiple tags specified, multiple needles set for assertion'
7:     #   at ./t/23-baseclass.t line 704.
7:     # Looks like you failed 1 test of 3.

kalikiana (changes requested):

t/23-baseclass.t
kalikiana: I wonder if you don't simply need to sort the array here

Martchus:
100 % on `[backend/baseclass.pm](https://codecov.io/gh/os-autoinst/os-autoinst/pull/2155/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=os-autoinst#diff-YmFja2VuZC9iYXNlY2xhc3MucG0=)` now :-)

Now I just need to figure out why the package build fails.

kalikiana:
> As if #2156 was already merged (but it is _not_).

Well [this](https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2155/os-autoinst/openSUSE_Tumbleweed/x86_64) is x86_64.

Note that [it also fails in t/23-baseclass.t on OBS](https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2155/os-autoinst/openSUSE_Leap_15.3/x86_64)

Martchus:
Oh, right. It was never completely disabled. So I suppose we should not merge the other PR at this point.

Martchus:
Looks like my latest commit helped to fix the issues of the Leap 15.3/15.4 builds on OBS. So this should be good to merge now. (My Leap 15.4 container where I wanted to reproduce the issue locally is still installing packages. Maybe I can nevertheless experiment a little bit with it because I'm still wondering why the error `Undefined subroutine called at /home/abuild/rpmbuild/BUILD/os-autoinst-4.6.1661874858.f99d1f6/backend/baseclass.pm line 1306.` occurred.)

backend/baseclass.pm
```diff
@@ -16,7 +16,7 @@ use File::Copy 'cp';
 use File::Basename;
 use Time::HiRes qw(gettimeofday time tv_interval);
 use Try::Tiny;
-use POSIX qw(_exit :sys_wait_h);
+use POSIX qw(_exit waitpid WNOHANG);
```
Martchus, line 19: Note that this is not required on TW. However, with the Perl version on Leap 15.3/15.4 the test otherwise fails. I've seen this in OBS but now also in a local container.

The following would work as well:
```suggestion
use POSIX qw(_exit :sys_wait_h waitpid);
```

So it is `waitpid` that is the undefined subroutine in `Undefined subroutine called at /tmp/os-autoinst/backend/baseclass.pm line 1306.`. However, for the sake of consistency it is likely best to just include both symbols explicitly and avoid use of the cryptic `:sys_wait_h` altogether.

kalikiana (approved):

PR #2156, opened 2022-08-30
Execute VMWare test on PowerPC again as instability should be fixed

Martchus:
Hopefully 725152a7d481e0cb7404f683aa22dae093261809 fixed the instability of
that test which lead to sporadic test failures (especially on PowerPC).

See https://progress.opensuse.org/issues/114881


Martchus:
Unfortunately this has not been fixed yet, see https://github.com/os-autoinst/os-autoinst/pull/2155#issuecomment-1232622920. So I'm setting this back to draft.

PR #2160, opened 2022-09-05
Extend unit tests of backend driver

Martchus:
See https://progress.opensuse.org/issues/111254

### DIFF
--- a/backend/driver.pm
+++ b/backend/driver.pm
@@ -23,21 +23,18 @@ use myjsonrpc;
 use signalblocker;
 use log qw(diag fctinfo);
 
+sub _collect_orphan ($session, $p, @) {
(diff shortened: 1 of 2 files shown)


kalikiana (approved):
Makes sense

backend/driver.pm
```diff
@@ -23,21 +23,18 @@ use myjsonrpc;
 use signalblocker;
 use log qw(diag fctinfo);
 
+sub _collect_orphan ($session, $p, @) {
```
perlpunk, line 26: Why the extra `@` args at the end?

Martchus: This function is invoked by `Mojo::RWP` and it is not clear whether this function might be called with more arguments in the future. Besides, I'm merely extending tests here; so changing parameter handling shouldn't be mixed into it (and the `@` preserves the old behavior despite using a different syntax).

perlpunk (approved):

PR #2162, opened 2022-09-06
Disable VMWare test for all archs as it is generally unstable

Martchus:
The test has also timed out a few times on x86_64 builds. The logs are not
very helpful so for now it is likely best to disable the test to keep the
CI going.

See https://progress.opensuse.org/issues/114881


kalikiana (approved):
It can't be helped

perlpunk (approved):

PR #2158, opened 2022-09-02
WIP: Move loop to wait for screen change from autotest to backend

Martchus:
Just a draft for a slight architectural change for `wait_screen_change` to implement See https://progress.opensuse.org/issues/114412, not tested at all.

---

* Make `query_isotovideo` block on `backend_wait_screen_change` until
  either the screen changes or the specified timeout has been exceeded
    * Avoid polling the backend from the autotest process (one less loop
      with a sleep)
    * Be able to adapt the frequency of screen updates as needed on the
      backend side
* See https://progress.opensuse.org/issues/114412

PR #2163, opened 2022-09-06
Extend tests of IPMI backend or mark lines as uncoverable

Martchus:
* Mark lines with sleep as uncoverable as it is likely not worth extending
  the tests for these cases
* Extend tests to cover otherwise uncovered lines
* See https://progress.opensuse.org/issues/111254


perlpunk (approved):

Martchus:
Oh no, this feature branch had the old version of #2158 (before I removed the TODO) in it. I suppose it is not that bad. I'll close #2158 and will rebase #2164.

PR #2166, opened 2022-09-09
Allow following Perl testsuite output during package builds

Martchus:
Turns out that this is only a problem of `prove` when it is not running within a terminal. New commit message:

When running the Perl testsuite via `prove` then output will be buffered
when a pipe is used, e.g. making timestamps on OBS useless.

One can easily see the difference locally by invoking e.g.
```
TESTS="t/27-consoles-vmware.t" PROVE_ARGS=-v  ninja check-pkg-build | ts
```
before and after this change.

---

Old commit message (not really relevant anymore):

So far the Perl testsuite is invoked via CTest when the `check-pkg-build` target is invoked. This unfortunately doesn't allow following the Perl testsuite output and CTest is only emitting the Perl testsuite output as one big chunk at the end, making the timestamps in OBS logs useless.

This change will invoke such tests not within CTest. Since we're using Perl's own test framework here anyways this should be fine. Now, the Perl testsuite is always executed first. If it fails no other tests are executed and the failure summary of the Perl testsuite is clearly visible. If the Perl testsuite succeeds, other tests on CMake level are executed via CTest as usual.

---

This will hopefull help debugging issues like
https://progress.opensuse.org/issues/114881 where it is otherwise not clear whether and at which point a test hangs or is slow.

### DIFF
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -79,8 +79,9 @@ endif ()
 
 # add targets for invoking Perl test suite
 find_program(PROVE_PATH prove)
+find_program(UNBUFFER_PATH unbuffer)
 if (PROVE_PATH)
-    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
+    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
     add_test(
         NAME test-perl-testsuite
         COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -83,6 +83,8 @@ Source0:        %{name}-%{version}.tar.xz
 %define devel_requires %python_style_requires %test_requires ShellCheck perl(Code::TidyAll) perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy) perl(Template::Toolkit)
 %define s390_zvm_requires /usr/bin/xkbcomp /usr/bin/Xvnc x3270 icewm xterm xterm-console xdotool fonts-config mkfontdir mkfontscale
 BuildRequires:  %test_requires %test_version_only_requires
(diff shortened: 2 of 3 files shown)


perlpunk (commented):
But then we can/should enable 27-consoles-vmware.t again, right?

Btw, in the build log from this PR I can't find a behaviour change: https://build.opensuse.org/public/build/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2166/openSUSE_Tumbleweed/x86_64/os-autoinst/_log
The OBS timestamp only seems to change after a single perl test file finished, like before.

Martchus:
> But then we can/should enable 27-consoles-vmware.t again, right?

No, that test still hasn't been fixed on OBS. I have created #2167 for that to be able to reproduce the issue again, though. (Of course now all tests have passed. I suppose I'll have to re-trigger tests there to get a negative result.)

> Btw, in the build log from this PR I can't find a behaviour change

You can. Just compare it to: https://progress.opensuse.org/attachments/13780/obs-vmware-test-failure-2.log.txt

You'll see that the lines of the Perl test suite are prefixed with `3: ` and have all the same timestamp. That's not the case in the log you've linked.

Btw, you can easily test this locally by invoking e.g. `WITH_COVER_OPTIONS=1 TESTS=t/27-consoles-vmware.t PROVE_ARGS=-v  ninja check-pkg-build` locally in your build dir.

Martchus:
> Btw, in the build log from this PR I can't find a behaviour change

Ok, there are indeed different timestamps in the logs also before this change. However, CTest is still only emitting rather big chunks of output at the same time instead of emitting the output as it is produced. So individual Perl tests are generally ending up with just the same timestamp. E.g. when running the following command locally the output of the Perl testsuite is only printed as two big chunks (at least on my system):

```
WITH_COVER_OPTIONS=1 TESTS="t/27-consoles-vnc.t t/27-consoles-vmware.t" PROVE_ARGS=-v ninja check-pkg-build
```

(And when you run the same command on this PR's branch the output is printed as it is produced.)

Martchus:
Updated the PR. Turns out @perlpunk was right and apparently the real problem was on `prove`-level.

Martchus:
Now it looks good. No gaps and different timestamps within one test:

```
[  169s] 3:     ok 6 - data can be read with timeout
[  169s] 3:     ok 7 - read with timeout has correct length
[  169s] 3:     ok 8 - read aborts after timeout on no data
[  169s] 3:     ok 9 - data can be read partially
[  169s] 3:     ok 10 - fragment read as expected
[  170s] 3:     ok 11 - second part of data can be read partially
[  170s] 3:     ok 12 - second part of data correctly read
[  170s] 3:     ok 13 - read until reports failure if search term not found
[  170s] 3:     ok 14 - rest of data has been read
[  170s] 3:     ok 15 - nothing left in re
```

perlpunk (approved):

baierjan (approved):

PR #2168, opened 2022-09-10
improve validate_script_output for clarity

bs-suse:
This is to improve the validate_script_output function.

To improve the information density on the details test result page I did:
- add `title` and `fail_message` parameters to `validate_script_output`
   - the `title` still defaults to `validate_script_output`
   - the `fail_message` still defaults to `output not validating`
- add a `validate_script_output script:` section to the message of the tile
   - the `script / cmd` is shown there right before the `output` of that command
- set the default value of `quiet` to `1` to make less noise

Using the `title` and `fail_message` can improve the readability of test results quite a bit.

```perl
validate_script_output("sapcontrol -nr ${instance_id} -function GetVersionInfo|tee GetVersionInfo_${instance_id}.before_patching", sub { /GetVersionInfo[\r\n]+OK/ }, title => "Versions before");
validate_script_output("sapcontrol -nr ${instance_id} -function StopSystem", sub { /StopSystem[\r\n]+OK/ }, title => "StopSystem");
validate_script_output("sapcontrol -nr ${instance_id} -function StartSystem", sub { /StartSystem/ }, proceed_on_failure => 1, title => "StartSystem");
# compare running versions (patch levels) - they should differ
validate_script_output(
    "diff --side-by-side --width=160 --suppress-common-lines GetVersionInfo_${instance_id}.before_patching GetVersionInfo_${instance_id}.after_patching",
    sub { /patch/ }, # succeeds if at least one patch level is different after patching
    fail_message => 'Patching FAILED. The versions are the same before and after patching :-(',
    proceed_on_failure => 1, # this needs to be here because the diff command will always find at least the difference in the timestamp
    title => "Version compare",
);
```

Successful validate example:
validate_script_output success detail

Failed test:
improve validate_script_output


testapi.pm
```diff
@@ -1202,14 +1202,22 @@ alternatively matches a regular expression. Use it as
...
+    my %args = compat_args(
+        {
+            title              => 'validate_script_output',
+            timeout            => undef,
+            fail_message       => 'output not validating',
+            quiet              => 1, # less noise by default
+        }, ((@args % 2) == 0) ? [] : ['timeout'], @args);
 
-    my $output = script_output($script, @args);
+    my $output = $distri->script_output($script, $args{timeout}, %args);
```
b10n1k, line 1220: Could you explain why this change is needed? i dont think is wrong but i would like to know what is the reasoning behind it?

testapi.pm
```diff
@@ -1227,27 +1235,27 @@ sub validate_script_output {    # no:style:signatures
         my $body = $deparse->coderef2text($check);
 
         $message = sprintf
-          "validate_script_output got:\n%s\n\nCheck function (deparsed code):\n%s",
-          $output, $body;
+          "validate_script_output script:\n%s\n\nvalidate_script_output got:\n%s\n\nCheck function (deparsed code):\n%s",
```
b10n1k, line 1238:
```suggestion
          "validate_script_output script:\n%s\n\nscript_output got:\n%s\n\nCheck function (deparsed code):\n%s",
```

testapi.pm
```diff
@@ -1227,27 +1235,27 @@ sub validate_script_output {    # no:style:signatures
...
     }
     elsif (reftype $check eq 'REGEXP') {
         if ($output !~ $check) {
             $res = 'fail';
             bmwqemu::diag("output does not match the regex:\n$output");
         }
         $message = sprintf
-          "validate_script_output got:\n%s\n\nRegular expression:\n%s",
-          $output, $check;
+          "validate_script_output script:\n%s\n\nvalidate_script_output got:\n%s\n\nRegular expression:\n%s",
```
b10n1k, line 1247:
```suggestion
          "validate_script_output script:\n%s\n\nscript_output got:\n%s\n\nRegular expression:\n%s",
```

b10n1k (changes requested):
Also the documentation now is misleading as it still implies that the `validate_script_output` is a wrapper of the `script_output` of the testapi.

testapi.pm
bs-suse: The parameter passing does not work if `compat_args` is called twice because of `$timeout` as fixed first parameter.
Via calling `$distri->script_output` directly I avoid that issue (and have less function calls).

cfconrad: I think the problem is, that you pass parameters like `title` and `fail_message` in the positional parameter slot, thus `compat_args` can not judge if these are named or positional and take it as positional parameter. This could result in the name `title` used as `timeout`.

If you call it like you did, you could also use the `testapi::script_output()` call. And I would prefer to use it, because the for instance the `quite` parameter is taken from `_QUITE_SCRIPT_CALLS` which isn't in the distribution!

```suggestion
    my $output = script_output($script, $args{timeout}, %args);
```

testapi.pm
```diff
@@ -1191,25 +1191,33 @@ sub get_test_data {
...
 
 sub validate_script_output {    # no:style:signatures
     my ($script, $check, @args) = @_;
+    my %args = compat_args(
+        {
+            title => 'validate_script_output',
+            timeout => undef,
+            fail_message => 'output not validating',
+            quiet => 1,    # less noise by default
+        }, ((@args % 2) == 0) ? [] : ['timeout'], @args);
```
alvarocarvajald, line 1218: I think this `compat_args()` call from L1212-L1218 can be moved below the call to `script_output()` and it should in theory preserve the old functionality. After all, the new named parameters are not needed by neither `testapi::script_output()` nor by `distribution::script_output()`

testapi.pm
alvarocarvajald: What I am not sure about @cfconrad 's workaround is that there are calls to `compat_args()` both in `testapi::script_output()` and in `distribution::script_output()` neither of which have a previous check for the number of arguments in `@_`. If `testapi::script_output()` is already using `compat_args()` to modify positional arguments into named arguments, then doing the same again in `distribution::script_output()` doesn't make much sense to me. I think we've been lucky that this has been working as is.

cfconrad: That is actually the idea of `compat_args()` You can call a function like:
```
func($a, $b, $b);
```perl
and 
```
func('foo' => $a, 'bla' => $b, 'blub' => $c);
```
But also as:
```perl
func($a, $b);
func($a);
func(bla=>$b);
```

And both result in the same behavior as long as you specify compat_args like
```perl
my %args = compat_args(
        {
            foo => 'something',
            bla => 'is',
            blub => 'this',
        }, ['foo', 'bla', 'blub'], @args);
```
see: https://github.com/os-autoinst/os-autoinst/blob/master/t/03-testapi.t#L661

And `compat_args` get problems when we pass not specified arguments in the positional one. For the example above it would:
```perl
func(k => $bob);
```
In that case, `compat_args` do not recognize the key `k` as named parameter, thus it take it as positional for `foo` and the value `$bob` as positional for `bla`... I think we run in this trap here as well, cause we extend the arguments from `script_output()` but do not specify them in the `compat_args` call in `validate_script_output`.

testapi.pm
cfconrad: as I tried to explain here: https://github.com/os-autoinst/os-autoinst/pull/2168/files#r968215741 it would make problems if someone call this function like:
```perl
validate_script_output($script, $check, title => "BLA", timeout => 3);
```

testapi.pm
cfconrad: There would be the case that we can change `compat_args` in that regards, if there are only one positional parameter, we could make some extra judgement to get that solved. But I do not see it with 2 or more positional parameters.

alvarocarvajald: > And `compat_args` get problems when we pass not specified arguments in the positional one. For the example above it would:
> 
> ```perl
> func(k => $bob);
> ```
> 
> In that case, `compat_args` do not recognize the key `k` as named parameter, thus it take it as positional for `foo` and the value `$bob` as positional for `bla`... I think we run in this trap here as well, cause we extend the arguments from `script_output()` but do not specify them in the `compat_args` call in `validate_script_output`.

That was the bit I was missing, and it explains why we have not had issues with the 2 inner calls to `compat_args()` in `testapi::script_output()` and `distribution_script_output()` as both `compat_args()` calls deal with the same 4 named args (timeout, proceed_on_failure, quiet and type_command), and why it's causing @bs-suse some troubles when any of his new named args reach the inner calls.

@bs-suse I maintain that the new named args are not required by either `testapi::script_output()` or by `distribution::script_output()`, but now I suggest instead of having the `compat_args()` call be placed after the call to `script_output()`, you can drop from the hash those named args you introduced that are not used by `script_output()`, so my suggestion would be:

```
sub validate_script_output {    # no:style:signatures
    my ($script, $check, @args) = @_;
    my %args = compat_args(
        {
            title => 'validate_script_output',
            timeout => undef,
            fail_message => 'output not validating',
            quiet => 1,    # less noise by default
        }, ((@args % 2) == 0) ? [] : ['timeout'], @args);
my $title = $args{title};
my $fail_message = $args{fail_message};
delete $args{title};
delete $args{fail_message};
my $output = script_output($script, %args);
```
And in the rest of the method, you refer to `$title` and `$fail_message` instead of `$args{title}` and `$args{fail_message}`.

cfconrad: I think if we will add this "one fixed parameter" case in `compat_args` you do not need to change the call at all. Also this `(@args % 2) == 0` is then not needed anymore. See:
https://github.com/os-autoinst/os-autoinst/pull/2169

testapi.pm
```diff
@@ -1202,14 +1202,27 @@ alternatively matches a regular expression. Use it as
...
+            fail_message => 'output not validating',
+            timeout => undef,
+            proceed_on_failure => undef,
+            quiet => 1,    # less noise by default
+            type_command => undef,
+        }, ((@args % 2) == 0) ? [] : ['timeout'], @args);
+    my $title = $args{title};
+    my $fail_message = $args{fail_message};
+    delete $args{title};
+    delete $args{fail_message};
```
cfconrad, line 1224:
```suggestion
    my $title = delete $args{title};
    my $fail_message = delete $args{fail_message};
```

testapi.pm
cfconrad: As you specified now all parameters. I do not see any need for this workaround...
```suggestion
        }, ['timeout'], @args);
```

t/03-testapi.t
```diff
@@ -591,17 +591,28 @@ subtest 'validate_script_output' => sub {
...
+
     );
-    for my $exp_args (@exp_args_list) {
-        my $joined_args = join(',', @$exp_args);
-        lives_ok { validate_script_output('script', qr/^$joined_args$/, @$exp_args) } "Arguments passed to script_output($joined_args)";
+    while (@exp_args_list) {
+        my $args = shift @exp_args_list;
+        my $exp = shift @exp_args_list;
+        validate_script_output('script', qr//, @$args);
+        is_deeply($arguments, $exp, "Arguments passed to script_output");
```
alvarocarvajald, line 615: `$arguments` is not assigned to anything.

bs-suse: isn't that what is done in line 602 `$mock_testapi->redefine(script_output => sub ($script, @args) { $arguments = {@args}; return '' });` ?

b10n1k (approved):
LGTM

t/03-testapi.t
Martchus, line 615: To follow our coding style and get more output in case of an error:
```suggestion
        is_deeply $arguments, $exp, 'Arguments passed to script_output' or diag explain $arguments;
```

Martchus (commented):
The PR description is quite detailed which is good. It would be nice if the commit message had more details as well.

t/03-testapi.t
alvarocarvajald: > isn't that what is done in line 602 `$mock_testapi->redefine(script_output => sub ($script, @args) { $arguments = {@args}; return '' });` ?

You're correct. I didn't notice that. Resolving this.

bs-suse:
I think I've addressed all the issues with this PR now.
Please have another look and let me know if I missed something.

cfconrad (commented):
LGTM

Martchus (approved):
Although it would still be nice to improve the test coverage of the changes (as the corresponding CI check fails).

t/03-testapi.t
```diff
@@ -587,21 +587,38 @@ subtest 'validate_script_output' => sub {
     throws_ok {
         validate_script_output('script', sub { m/error/ })
     } qr/output not validating/, 'Die on output not match';
+    throws_ok {
+        validate_script_output('script', qr/error/)
+    } qr/output not validating/, 'Die on output not match';
```
okurz, line 592: please use new, unique test descriptions so that the output is not repeated in console output and the specific test locations can be identified in case of failures, e.g.

```suggestion
    } qr/output not validating/, 'Die on output not match for regex';
```

foursixnine (approved):

okurz (changes requested):

t/03-testapi.t
okurz: ok, commented after merge. I would appreciate if you follow-up with these changes in another PR.

okurz:
I have created https://github.com/os-autoinst/os-autoinst/pull/2178 to bump the API version due to the  functional changes in the test API.

t/03-testapi.t
bs-suse: done. 😄 
https://github.com/os-autoinst/os-autoinst/pull/2179

PR #2169, opened 2022-09-13
testapi: Optimize compat_args()'s handling of one fixed parameter

cfconrad:
With only one fixed parameter, the number of arguments specify
if the fixed parameter was given or not.
It makes the check for known named parameter keys obsolete and
allow additional parameters also in the first field (fixed parameter).

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -2280,18 +2280,17 @@ A typical call would look like:
 
 =cut
 
-sub compat_args {    # no:style:signatures
-    my ($def_args, $fix_keys) = splice(@_, 0, 2);
+sub compat_args ($def_args, $fix_keys, @args) {
     my %ret;
     if (@$fix_keys == 1) {
-        $ret{$fix_keys->[0]} = shift if ((@_ % 2) != 0);
+        $ret{$fix_keys->[0]} = shift @args if ((@args % 2) != 0);
     } else {
         for my $key (@{$fix_keys}) {
-            $ret{$key} = shift if (@_ >= 1 && (!defined($_[0]) || !exists $def_args->{$_[0]}));
+            $ret{$key} = shift @args if (@args >= 1 && (!defined($args[0]) || !exists $def_args->{$args[0]}));
         }
     }
-    carp("Odd number of arguments") unless ((@_ % 2) == 0);
-    %ret = (%{$def_args}, %ret, @_);
+    carp("Odd number of arguments") unless ((@args % 2) == 0);
+    %ret = (%{$def_args}, %ret, @args);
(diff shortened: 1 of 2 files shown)


Martchus:
The commit message has grammar/punctuation issues and I'm also not quite sure whether I understand it correctly. "We do not *need*" and "it is *possible* to *just*" makes it sound like this was just and optimization and there's no change in behavior. Is that correct? Then it would make sense to title the commit e.g. "testapi: Optimize `compat_args()`'s handling of one fixed parameter". Otherwise it would make sense to start with e.g. "testapi: Fix …" to state the intention of the change.

testapi.pm
```diff
@@ -2283,8 +2283,14 @@ A typical call would look like:
 sub compat_args {    # no:style:signatures
     my ($def_args, $fix_keys) = splice(@_, 0, 2);
     my %ret;
-    for my $key (@{$fix_keys}) {
-        $ret{$key} = shift if (scalar(@_) >= 1 && (!defined($_[0]) || !grep { $_ eq $_[0] } keys(%{$def_args})));
```
perlpunk, line 2287: Not related to the PR, but I never looked closely at this code. It occurred to me that for looking if a key exists in a hash, you don't need grep:
```
# old:
!grep { $_ eq $_[0] } keys(%{$def_args}
# new
! exists $def_args->{$_[0]}
```
That would be a simple optimization.

testapi.pm
```diff
@@ -2283,8 +2283,14 @@ A typical call would look like:
...
+    if (scalar(@$fix_keys) == 1) {
```
perlpunk, line 2286:
```suggestion
    if (@$fix_keys == 1) {
```

testapi.pm
```diff
@@ -2283,8 +2283,14 @@ A typical call would look like:
...
+        if ((scalar(@_) % 2) != 0) {
```
perlpunk, line 2287:
```suggestion
        if ((@_ % 2) != 0) {
```

testapi.pm
```diff
@@ -2283,8 +2283,14 @@ A typical call would look like:
...
+            $ret{$fix_keys->[0]} = shift(@_);
```
perlpunk, line 2288:
```suggestion
            $ret{$fix_keys->[0]} = shift;
```

testapi.pm
cfconrad: thx, I have added it anyway...

testapi.pm
cfconrad: and I changed it to use `signatures` where we do not have this implicit shift...

kalikiana (approved):

okurz (approved):

PR #2170, opened 2022-09-13
Fix resolution with virtio backend on new QEMU versions

Martchus:
The QEMU version 7.0.0 (currently in Tumbleweed) apparently changed the default EDID information when using the virtio backend so now the screen resolution is wrong once the graphical installer is shown. (The video parameter usually specified within GRUB only affects the console but the installer GUI makes its own EDID lookup.)

This change fixes the problem at least when `QEMUVGA` is set manually to `virtio`.

Note that `cirrus` doesn't seem to be affected.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -680,7 +680,15 @@ sub start_qemu ($self) {
     elsif ($vars->{OFW}) {
         $use_usb_kbd = $self->qemu_params_ofw;
     }
-    sp('vga', $vars->{QEMUVGA}) if $vars->{QEMUVGA};
+    if (my $qemu_vga = $vars->{QEMUVGA}) {
+        if ($qemu_vga eq 'virtio') {
+            # specify EDID information explicitly to get consistent behavior across different QEMU versions
+            sp('device', "virtio-vga,edid=on,xres=$self->{xres},yres=$self->{yres}");
+        }
+        else {
+            sp('vga', $qemu_vga);
+        }
+    }
 
     my @nicmac;
     my @nicvlan;


Martchus:
I've tested it on TW and it works. Of course it is only effective if one specified `QEMUVGA=virtio` explicitly.

This still leaves the default case to be handled. That would be the `std` VGA backend (as of QEMU 2.2, see documentation of `-vga` parameter on https://www.qemu.org/docs/master/system/qemu-manpage.html#hxtool-3) and it also needs fixing (as of QEMU 7.0). However, so far we normally don't specify the `-vga` parameter to keep the default. When no `-vga` parameter is specified the default VGA device is called "VGA" and we *could* pass EDID via `-device VGA,edid=on,xres=1800,yres=200`. I'm not sure whether this will break if QEMU changes the default VGA backend to something else but maybe it is still a worthwhile change so users with QEMU 7.0 don't need to specify `QEMUVGA=virtio` explicitly and the standard just works. (And as a bonus, custom resolutions would also work for the standard backend.)

perlpunk (approved):

Martchus:
I've been creating #2171 to address the default case mentioned in my previous comment. It would likely still make sense to merge this PR first. The other PR will also fix the missing test coverage (but I'm working on that anyways as part of https://progress.opensuse.org/issues/111254 so it will definitely not be forgotten, even if we eventually don't merge #2171).

AdamWill:
So this all would work, but it just feels increasingly messy alongside the ARM-specific code next door (I'm including #2171 in my evaluation here). We now have `QEMUVGA` which exactly sets the `-vga` arg except when it sets `-device` with a bunch of magic extra stuff, but is entirely ignored on ARM. And we have `QEMU_VIDEO_DEVICE` which sets `-device`, but only on ARM even though its name gives you no indication of this. Oh, and `QEMU_OVERRIDE_VIDEO_DEVICE_AARCH64`, which actually also applies to 32-bit ARM.

It's all a bit ugh at this point.

Could we maybe just blow it all up and start over, maybe referencing https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ ? I admit it's a bit tricky to figure out how, but it just feels so ugh right now.

Martchus:
As mentioned in the chat I like your proposal. I think we can nevertheless this PR before (to mitigate the immediate problem; this PR should be ready to be merged, as well as #2171 which also has tests).

AdamWill:
Sure, I can send mine on top of this and #2171 .

Martchus:
Merged as part of #2171

PR #2171, opened 2022-09-13
Fix resolution for std/virtio backend on new QEMU versions

Martchus:
Based on #2170 but also covers the default graphics backend (`QEMUVGA` is `std` or not present at all). Also includes a test.


AdamWill:
As discussed in #2170 , I'd like to rewrite this to be more sane and I have a proposal for that waiting in the wings, but these two do technically do the right thing so :+1: to merging them first.

baierjan (approved):

okurz (approved):

PR #2173, opened 2022-09-13
Deal with "append" options for QEMU backend

jamespthomas:
Currently only the first option to "append" passed via QEMU_APPEND will be passed to the kernel, so check for this and correctly app all command line appends


jamespthomas:
Note, I don't expect this to be the most optimal way of doing this since my perl is non-existent, but hopefully it explains the problem a bit better

Martchus:
It would be great to extend unit tests for this case. To make it easier, feel free to extract this piece of code into its own function.

PR #2175, opened 2022-09-15
Improve performance of `wait_still_screen`

Martchus:
* Move the loop in autotest polling the backend into the backend itself
  removing the sleep in autotest completely (similar to 05aa5cd)
* Lower the update frequency of the backend when `no_wait` is used (instead
  of just polling the backend more often)
* See https://progress.opensuse.org/issues/116554

### DIFF
--- a/t/99-full-stack.t
+++ b/t/99-full-stack.t
@@ -59,7 +59,7 @@ unlike $log, qr/script_run: DEPRECATED call of script_run.+die_on_timeout/, 'no 
 like $log, qr/do not wait_still_screen/, 'test type string and do not wait';
 like $log, qr/wait_still_screen: detected same image for 0.2 seconds/, 'test type string and wait for .2 seconds';
 like $log, qr/wait_still_screen: detected same image for 1 seconds/, 'test type string and wait for 1 seconds';
-like $log, qr/wait_still_screen: detected same image for 0.1 seconds/, 'test type string and wait for .1 seconds';
+like $log, qr/wait_still_screen: detected same image for 0\.1 seconds/, 'test type string and wait for .1 seconds';
 like $log, qr/.*event.*STOP/, 'Machine properly paused';
 like $log, qr/.*event.*RESUME/, 'Machine properly resumed';
 like $log, qr/get_test_data returned expected file/, 'get_test_data test';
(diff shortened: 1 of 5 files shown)


Martchus:
It should be ready for review. I'm still running some tests, though.

t/99-full-stack.t
```diff
@@ -59,7 +59,7 @@ unlike $log, qr/script_run: DEPRECATED call of script_run.+die_on_timeout/, 'no
 like $log, qr/do not wait_still_screen/, 'test type string and do not wait';
 like $log, qr/wait_still_screen: detected same image for 0.2 seconds/, 'test type string and wait for .2 seconds';
 like $log, qr/wait_still_screen: detected same image for 1 seconds/, 'test type string and wait for 1 seconds';
-like $log, qr/wait_still_screen: detected same image for 0.1 seconds/, 'test type string and wait for .1 seconds';
+like $log, qr/wait_still_screen: detected same image for 0\.1 seconds/, 'test type string and wait for .1 seconds';
 like $log, qr/.*event.*STOP/, 'Machine properly paused';
```
kraih, line 63: You could have fixed the `0.2` above too.

kraih (approved):

testapi.pm
```diff
@@ -663,39 +663,23 @@ sub wait_still_screen {    # no:style:signatures
...
+    my $res = query_isotovideo('backend_wait_still_screen', \%args);
+    if (!$res->{timed_out}) {
+        bmwqemu::fctres("detected same image for $stilltime seconds ($res->{elapsed} s elapsed), last detected similarity is $res->{sim}");
+        return 1;
+    }
+    else {
+        $autotest::current_test->timeout_screenshot;
+        bmwqemu::fctres("wait_still_screen timed out after $res->{elapsed} seconds, last detected similarity is $res->{sim}");
+        return 0;
     }
```
okurz, line 682: I'd use an early return on the true-case here but not that important:
```suggestion
    if ($res->{timed_out}) {
        $autotest::current_test->timeout_screenshot;
        bmwqemu::fctres("wait_still_screen timed out after $res->{elapsed} seconds, last detected similarity is $res->{sim}");
        return 0;
    }
    bmwqemu::fctres("detected same image for $stilltime seconds ($res->{elapsed} s elapsed), last detected similarity is $res->{sim}");
    return 1;

```

okurz (approved):

PR #2177, opened 2022-09-21
Consolidate qemu video device setting, deprecate QEMUVGA

AdamWill:
As discussed in #2170, this code has progressively become a bit of a mess as special cases were added for different arches and things changed in qemu. This rewrites it to at least consolidate the mess in one place, and deprecate some things which will allow us to simplify it in future.

Always setting `-device` rather than sometimes `-vga` and sometimes `-device` allows us to be much more consistent. We 'translate' `QEMUVGA` values into `-device` values after warning that it is deprecated. Eventually we can hopefully get rid of the handling of the deprecated vars, and the code will then be much clearer and simpler.

References for this:
https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ https://www.kraxel.org/blog/2019/03/edid-support-for-qemu/

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 31;
+our $version = 32;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -501,7 +501,6 @@ sub delete_virtio_console_fifo () { unlink $_ or bmwqemu::fctwarn("Could not unl
 
 sub qemu_params_ofw ($self) {
     my $vars = \%bmwqemu::vars;
-    $vars->{QEMUVGA} ||= "std";
     $vars->{QEMUMACHINE} //= "usb=off";
     # set the initial resolution on PCC and SPARC
     sp('g', "$self->{xres}x$self->{yres}");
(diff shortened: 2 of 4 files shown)


Martchus (approved):

Martchus:
> Sorry, if I could not yet follow the complete discussion flow in previous pull requests.

Well, I've already insinuated in previous discussions that there might be more people to convince than just me. (Since deprecation doesn't mean breaking existing behavior as it is rather a suggestion to use another variable in accordance with QEMU's upstream changes that make `-device…` the preferred way of specifying a graphics device now.)

AdamWill:
So yeah, let me unpack the problem with `QEMUVGA` a little more, as I see it.

The intent of that variable is: if you set this to foo, we will run `qemu -vga foo`. Only, that has turned out to be problematic in a couple of ways.

The original one was, I think, because of [this](https://listman.redhat.com/archives/libvir-list/2016-September/msg00544.html): using `-vga virtio` on aarch64 with `-machine virt` (which is what we use) doesn't work properly. So if you want virtio graphics on aarch64 it's best to use `-device virtio-gpu-pci`. So we grew a whole slightly janky codepath specific to aarch64 which entirely ignores `QEMUVGA` (though this is not documented) and only lets you pick between virtio and VGA.

The new one is this whole "changing the EDID resolution" stuff. You can't set the EDID resolution options using `-vga`. You have to pass `-device`. So @Martchus 's patches had to kind of translate `QEMUVGA` variables into `-device` instead of `-vga`, but only when using an EDID-capable device.

So that's two situations where setting `-vga` is not actually worthwhile and we have to, awkwardly, hack around the idea of `QEMUVGA`.

When I first thought "let's clean up this mess" I didn't intend to deprecate QEMUVGA, but for me, looking at the mess, it kind of comes naturally. The natural approach to cleaning it up is, as this code does, to always set `-device`. When you've decided that, it feels weird that the main interface to this system - except on aarch64! - is a variable that claims to be setting something else, and which we have to "translate" into a `-device` setting. It feels much more clean to just say "look, it turns out that using `-vga` was a bad idea, so we want to migrate off that, so it makes more sense to set a variable that maps to setting `-device`".

I wasn't necessarily planning to actually go ahead and *remove* `QEMUVGA` for some time, maybe not unless (until?) we run into yet another problem of this type that can only be solved with `-device`. But saying it's deprecated does give us cover to do it if/when it becomes necessary.

kraih (approved):

AdamWill:
A note here: I was planning to also have this patch always set `-vga none`, to ensure qemu doesn't automatically add a(nother) video device. kraxel (qemu dev) told me qemu is supposed to notice when a video `-device` option is set and not do that, but there may be corner cases where it still happens. However, I found in testing that running on ppc64le with `-device virtio-vga -vga none` seems not to work properly, where as just `-device virtio-vga` works OK. I'll look into that some more next week.

kalikiana:
@Mergifyio rebase

kalikiana (approved):

okurz (dismissed):
Please also bump the interface number in OpenQA/Isotovideo/Interface.pm, then I can approve

PR #2180, opened 2022-09-23
Dynamically obtain vmware datastore path

nanzhg:
VMware datastore path should not be hardcode as its name is different on each ESXi server. Its value can be obtain from the variable 'VMWARE_DATASTORE'.

Failed job: https://openqa.suse.de/tests/9574488
```
[debug] [run_ssh_cmd(echo bios.bootDelay = "10000" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-4.vmx)] stderr:
  sh: can't create /vmfs/volumes/datastore1/openQA/openQA-SUT-4.vmx: nonexistent directory
```

https://openqa.suse.de/tests/9574488/file/vars.json
`"VMWARE_DATASTORE": "Datastore2",`

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -522,7 +522,7 @@ __END"
     # define the new domain
     $self->run_cmd("virsh $remote_vmm define $xmlfilename") && die "virsh define failed";
     if ($self->vmm_family eq 'vmware') {
-        my $vmx = sprintf('/vmfs/volumes/datastore1/openQA/%s.vmx', $self->name);
+        my $vmx = sprintf('/vmfs/volumes/%s/openQA/%s.vmx', $bmwqemu::vars{VMWARE_DATASTORE} // 'datastore1', $self->name);
 
         # set default boot delay
         $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -217,7 +217,7 @@ VIRSH_INSTANCE;integer;;VM's instance number on VIRSH_HOSTNAME
 VMWARE_USERNAME;string;;Administrator's username ('@' is '%40')
 VMWARE_PASSWORD;string;;Administrator's password
 VMWARE_HOST;string;;VCS server for authentication
-VMWARE_DATASTORE;string;;VMware datastore
+VMWARE_DATASTORE;string;datastore1;VMware datastore
 VMWARE_NFS_DATASTORE;string;;VMware datastore with openQA NFS directories
 VMWARE_SERIAL_PORT;string;;TCP port where is VM's serial port stream to be expected on the ESX server
 VMWARE_BRIDGE;string;;VMware's bridge name (usual default is 'VM Network')


consoles/sshVirtsh.pm
```diff
@@ -522,7 +522,7 @@ __END"
     # define the new domain
     $self->run_cmd("virsh $remote_vmm define $xmlfilename") && die "virsh define failed";
     if ($self->vmm_family eq 'vmware') {
-        my $vmx = sprintf('/vmfs/volumes/datastore1/openQA/%s.vmx', $self->name);
+        my $vmx = sprintf('/vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/%s.vmx', $self->name);
```
Martchus, line 525: It is nicer to just rely on `sprintf` and still specify `datastore1` as a default:
```suggestion
        my $vmx = sprintf('/vmfs/volumes/%s/openQA/%s.vmx', $bmwqemu::vars{VMWARE_DATASTORE} // 'datastore1', $self->name);
```
(or at least use interpolation)

If you think `datastore1` is too specific to use as a default and there should be no default then you should make it rather throw an exception that the mandatory variable `VMWARE_DATASTORE` is missing (instead of having Perl emit a warning in that case).

t/22-svirt.t
```diff
@@ -162,11 +162,11 @@ subtest 'starting VMware console with Cloud Init' => sub {
...
-        'echo bios.bootDelay = "10000" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
-        'echo guestinfo.metadata = "test@meta" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
-        'echo guestinfo.metadata.encoding = "gzip+base64" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
-        'echo guestinfo.userdata = "test%user" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
-        'echo guestinfo.userdata.encoding = "gzip+base64" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx',
+        'echo bios.bootDelay = "10000" >> /vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/openQA-SUT-1.vmx',
+        'echo guestinfo.metadata = "test@meta" >> /vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/openQA-SUT-1.vmx',
+        'echo guestinfo.metadata.encoding = "gzip+base64" >> /vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/openQA-SUT-1.vmx',
+        'echo guestinfo.userdata = "test%user" >> /vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/openQA-SUT-1.vmx',
+        'echo guestinfo.userdata.encoding = "gzip+base64" >> /vmfs/volumes/' . $bmwqemu::vars{VMWARE_DATASTORE} . '/openQA/openQA-SUT-1.vmx',
```
Martchus, line 169: I'm wondering whether all of these test changes are actually required. I'd rather avoid them considering that in the tests we use a defined set of `$bmwqemu::vars` values and therefore we can actually just keep the expected value here hard-coded (similar to how the name of the vmx file is also just hard-coded). That makes it easier to understand what the test actually expects here.

Martchus (changes requested):

consoles/sshVirtsh.pm
Martchus: Looks like the variable `VMWARE_DATASTORE` is already used somewhere else defaulting to an empty string. Not sure whether that makes sense, though. If you make it defaulting to `datastore1` it would make sense to also update other places accordingly and specify the default in `doc/backend_vars.asciidoc`.

t/22-svirt.t
nanzhg: OK. In fact, it will not impact testing on sle15sp5, will keep 'datastore1' as the default value.

consoles/sshVirtsh.pm
nanzhg: Thanks for your review and comments.
Actually, the default datastore name `datastore1` is nonexistent in our vmware server which caused the job failure. This code change is for the case that `datastore1` is not specified as the default value.
Agree your suggested change, will update it and specify the default value in `doc/backend_vars.asciidoc`.

Martchus (approved):

perlpunk (approved):

PR #2181, opened 2022-09-23
Prevent stalebot from closing stale issues

Martchus:
Those issues might still be valid, see
https://progress.opensuse.org/issues/117136

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,5 +1,5 @@
 daysUntilStale: 90
-daysUntilClose: 7
+daysUntilClose: false
 staleLabel: stale
 markComment: >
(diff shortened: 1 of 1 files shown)


baierjan (approved):

kraih (approved):

.github/stale.yml
```diff
@@ -1,5 +1,5 @@
 daysUntilStale: 90
-daysUntilClose: 7
+daysUntilClose: false
```
okurz, line 2: I assume this affects issues and pull requests in the same way? I don't mind either way

Martchus: Yes, it should affect both.

PR #2184, opened 2022-09-28
Fix missing quotes output in vmx file

nanzhg:
Since vmware 7.0 has more strict inspection on vmx file, we need to ensure the quotation mark output around the value while adding a new property in vmx file.

To resolve https://progress.opensuse.org/issues/117196, adding backslash to output quotation mark.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -525,7 +525,7 @@ __END"
         my $vmx = sprintf('/vmfs/volumes/%s/openQA/%s.vmx', $bmwqemu::vars{VMWARE_DATASTORE} // 'datastore1', $self->name);
 
         # set default boot delay
-        $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd(qq{echo 'bios.bootDelay = "10000"' >> $vmx}, domain => 'sshVMwareServer');
 
         # inject cloud init metadata and userdata required for the image if there are any
         my $ci_meta = $bmwqemu::vars{CLOUD_INIT_META};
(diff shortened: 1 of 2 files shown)


Martchus (commented):
Looks good but you need to adapt the corresponding tests as well (see CI failure).

Martchus:
It would also be great to have the information from the PR description in the commit message body.

consoles/sshVirtsh.pm
```diff
@@ -525,18 +525,18 @@ __END"
         my $vmx = sprintf('/vmfs/volumes/%s/openQA/%s.vmx', $bmwqemu::vars{VMWARE_DATASTORE} // 'datastore1', $self->name);
 
         # set default boot delay
-        $self->run_cmd("echo bios.bootDelay = \"10000\" >> $vmx", domain => 'sshVMwareServer');
+        $self->run_cmd("echo bios.bootDelay = \\\"10000\\\" >> $vmx", domain => 'sshVMwareServer');
```
perlpunk, line 528: I would actually try to make it more readable:
```suggestion
        $self->run_cmd(q{echo 'bios.bootDelay = "10000"' >> $vmx}, domain => 'sshVMwareServer');
```
perl has these special quotes, so that you can both use `'` and `"` in the string, and now it's clear that the double quotes should land in the file.

Martchus:
Have you seen my first comment? (Tests are still failing.)

nanzhg:
> Have you seen my first comment? (Tests are still failing.)

Yes, I was looking into the failure. Will try to fix it. Thanks!

nanzhg:
> Have you seen my first comment? (Tests are still failing.)

Yes, I've been looking into the failure. Will try to fix it. Thanks!

nanzhg:
@Martchus For below CI failure, should I also update `t/22-svirt.t` to match newly added the single quotation in `consoles/sshVirtsh.pm`?
#   Failed test 'expected commands invoked'
#   at ./t/22-svirt.t line 128.
#     Structures begin differing at:
#              $got->[3] = 'echo 'bios.bootDelay = "10000"' >> $vmx'
#     $expected->[3] = 'echo bios.bootDelay = "10000" >> /vmfs/volumes/datastore1/openQA/openQA-SUT-1.vmx'

Martchus (changes requested):
Yes, please update the test so it expects the quoting (and passes again). However, also note that the test reveals a bug in the current implementation (after following @perlpunk 's quoting suggestion). Now the interpolation of `$vmx` isn't done anymore (which is a bug, maybe you need to go back to using `"` unless @perlpunk has a better suggestion).

consoles/sshVirtsh.pm
perlpunk: Oh sorry, it should be:
```
        $self->run_cmd(qq{echo 'bios.bootDelay = "10000"' >> $vmx}, domain => 'sshVMwareServer');
```
thanks @Martchus

perlpunk:
yeah, just use `qq{}` instead of `q{}`. I commented with the corrected line. `q{}` is single quoting, and `qq{}` is double quoting.

Martchus (approved):

perlpunk (approved):

PR #2187, opened 2022-09-28
Add files that are now fully covered to the list in codecov config

Martchus:
See https://progress.opensuse.org/issues/111254#note-17

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,19 +17,7 @@ coverage:
         target: 100.0
         paths:
           - OpenQA
-          - backend/amt.pm
-          - backend/baseclass.pm
-          - backend/console_proxy.pm
-          - backend/driver.pm
-          - backend/generalhw.pm
-          - backend/ipmi.pm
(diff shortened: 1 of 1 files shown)


okurz (approved):

codecov.yml
```diff
@@ -18,10 +18,20 @@ coverage:
...
+          - backend/driver.pm
+          - backend/generalhw.pm
+          - backend/ipmi.pm
           - backend/null.pm
+          - backend/qemu.pm
+          - backend/s390x.pm
           - backend/spvm.pm
+          - backend/svirt.pm
           - backend/vagrant.pm
           - backend/virt.pm
```
okurz, line 32: How about using https://docs.codecov.com/docs/codecovyml-reference#ignore to only ignore pvm and ikvm and add `backend/` here?

okurz (changes requested):

okurz (approved):

kalikiana (approved):

PR #2192, opened 2022-10-10
Allow explicit use of SSH agent after 026c93c broke implicit use

Martchus:
With 026c93c is is possible use a falsy value as password which broke the ability to use a falsy value to imply SSH agent use. This change adds the possibility to use SSH agent back by explicitly specifying `use_ssh_agent => 1`.

See https://github.com/os-autoinst/os-autoinst/pull/2152#issuecomment-1272407524 and https://progress.opensuse.org/issues/117784


baierjan (approved):

perlpunk (approved):

marmarek:
Thanks!

PR #2194, opened 2022-10-26
Move backend process logic into new class

kalikiana:

### DIFF
--- a/OpenQA/Isotovideo/Backend.pm
+++ b/OpenQA/Isotovideo/Backend.pm
@@ -1,0 +1,33 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Isotovideo::Backend;
+use Mojo::Base -base, -signatures;
+
+use Mojo::File qw(path);
+use backend::driver;
+use bmwqemu;
+use log qw(diag);
+
+sub new ($class, @args) {
+    $bmwqemu::vars{BACKEND} ||= "qemu";
+    $bmwqemu::backend = backend::driver->new($bmwqemu::vars{BACKEND});
+
+    path('os-autoinst.pid')->spurt("$$");
+    # might throw an exception
+    $bmwqemu::backend->start_vm;
+    $class->SUPER::new(@args);
+}
+
+sub process ($self) {
+    $bmwqemu::backend->{backend_process};
+}
+
+sub stop ($self) {
+    return undef unless my $process = $self->process;
+    diag('stopping backend process ' . $process->pid);
+    $process->stop if $process->is_running;
+    diag('done with backend process');
+}
+
+1;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ set(PERL_FILES
     OpenQA/Benchmark/Stopwatch.pm
     OpenQA/Commands.pm
     OpenQA/Exceptions.pm
+    OpenQA/Isotovideo/Backend.pm
     OpenQA/Isotovideo/CommandHandler.pm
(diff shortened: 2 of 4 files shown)


OpenQA/Isotovideo/Backend.pm
```diff
@@ -0,0 +1,31 @@
...
+}
+
+has process => sub {
+    return $bmwqemu::backend->{backend_process};
+};
+
+sub stop ($self) {
+    diag('stopping backend process ' . $self->process->pid);
+    $self->process->stop if $self->process->is_running;
+    $self->process = undef;
```
Martchus, line 27:
```suggestion
    $self->process(undef);
```

Martchus (changes requested):

OpenQA/Isotovideo/Backend.pm
```diff
@@ -0,0 +1,31 @@
...
+    path('os-autoinst.pid')->spurt("$$");
+    # might throw an exception
+    $bmwqemu::backend->start_vm;
+}
+
+has process => sub {
+    return $bmwqemu::backend->{backend_process};
+};
+
+sub stop ($self) {
```
Martchus, line 24: Since the process might be `undef`:
```suggestion
sub stop ($self) {
    return undef unless my $process = $self->process;
```
And then you can use `$process` directly (instead of using the attribute repeatedly).

Martchus (dismissed):

OpenQA/Isotovideo/Backend.pm
```diff
@@ -0,0 +1,31 @@
...
+    $bmwqemu::backend = backend::driver->new($bmwqemu::vars{BACKEND});
+
+    path('os-autoinst.pid')->spurt("$$");
+    # might throw an exception
+    $bmwqemu::backend->start_vm;
+}
+
+has process => sub {
+    return $bmwqemu::backend->{backend_process};
+};
```
kraih, line 22: Should this really be cached in `$self->{process}`? Seems like it could leak. A simple getter method might be more appropriate.

kalikiana: I don't know what you mean by leak. It was previously stored in a variable. `stop` needs to do the cleanup like before.

kraih: You're creating a new reference to the value behind `$bmwqemu::backend->{backend_process}` here. If that value ever changes and would have been garbage collected, the new reference would still exist and prevent garbage collection. That's why you generally only want to create these extra references via attribute accessors when you really need them. Which i don't believe to be the case here. Instead you could just do `sub process { $bmwqemu::backend->{backend_process} }`.

OpenQA/Isotovideo/Backend.pm
```diff
@@ -0,0 +1,33 @@
+# Copyright 2018-2022 SUSE LLC
```
okurz, line 1:
```suggestion
# Copyright SUSE LLC
```

OpenQA/Isotovideo/Backend.pm
```diff
@@ -0,0 +1,33 @@
...
+}
+
+sub stop ($self) {
+    return undef unless my $process = $self->process;
+    diag('stopping backend process ' . $process->pid);
+    $process->stop if $process->is_running;
+    diag('done with backend process');
+}
+
+1;
```
Martchus, line 33: Not having a single function that actually uses `$self` makes me think whether we really need to make an object here. I suppose for future extension it cannot hurt, though.

kraih: That would be my train of thought as well. It's for future extensions.

okurz (approved):

PR #2196, opened 2022-10-28
Move check_asserted_screen into CommandHandler

kalikiana:

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -9,12 +9,13 @@ use testapi 'diag';
 use OpenQA::Isotovideo::Interface;
 use Cpanel::JSON::XS;
 use Mojo::File 'path';
+use Time::HiRes qw(gettimeofday tv_interval);
 
 use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
 
 
 # io handles for sending data to command server and backend
-has [qw(cmd_srv_fd backend_fd answer_fd)] => undef;
+has [qw(test_fd cmd_srv_fd backend_fd answer_fd)] => undef;
 
 # the name of the current test (full name includes category prefix, eg. installation-)
 has [qw(current_test_name current_test_full_name)];
@@ -52,6 +53,15 @@ has backend_requester => undef;
 
 # whether the test has already been completed and whether it has died
 has [qw(test_completed test_died)] => 0;
+
+# the time of the last asserted screen
+has [qw(last_check_seconds last_check_microseconds)] => 0;
+
+sub new ($class, @args) {
+    my $self = $class->SUPER::new(@args);
+    $self->_update_last_check;
+    return $self;
+}
 
 sub clear_tags_and_timeout ($self) {
     $self->tags(undef);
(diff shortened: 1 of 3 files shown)


okurz (approved):

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -350,4 +351,37 @@ sub update_status_file ($self) {
     rename $tmp, AUTOINST_STATUSFILE or die $!;
 }
 
+my ($last_check_seconds, $last_check_microseconds) = gettimeofday;
```
Martchus, line 354: These variables could now be fields of the object. Otherwise it'll be a bit unintuitive that different instances influence each other this way.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -350,4 +351,37 @@ sub update_status_file ($self) {
...
+sub _calc_check_delta ($self) {
+    # an estimate of eternity
+    my $delta = $last_check_seconds ? tv_interval([$last_check_seconds, $last_check_microseconds], [gettimeofday]) : 100;
+    # sleep the remains of one second if $delta > 0
+    my $timeout = $delta > 0 ? 1 - $delta : 0;
+    $self->timeout($timeout < 0 ? 0 : $timeout);
+    return $delta;
+}
+
+sub check_asserted_screen ($self, $testfd) {
```
Martchus, line 364: `$testfd` is always the same, right? So maybe it makes sense to make it an attribute of CommandHandler and pass it when instantiating the object like the oder FDs. Then it would be more consistent with how the CommandHandler works in general.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -350,4 +361,38 @@ sub update_status_file ($self) {
...
+        $self->timeout(0.1);
+    }
+    else {
+        $self->_calc_check_delta;
+        # come back later, avoid too often called function
+        return if $self->timeout > 0.05;
+    }
+    my ($last_check_seconds, $last_check_microseconds) = gettimeofday;
+    $self->last_check_seconds($last_check_seconds);
+    $self->last_check_microseconds($last_check_microseconds);
```
Martchus, line 385: What about

```
sub _update_last_check ($self) {
    my ($last_check_seconds, $last_check_microseconds) = gettimeofday;
    $self->last_check_seconds($last_check_seconds);
    $self->last_check_microseconds($last_check_microseconds);
}
```

and use that here and in the constructor?

Martchus (commented):
Looks good, I have only a small suggestion to reduce code duplication.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -53,6 +54,14 @@ has backend_requester => undef;
 # whether the test has already been completed and whether it has died
 has [qw(test_completed test_died)] => 0;
 
+# the time of the last asserted screen
+has [qw(last_check_seconds last_check_microseconds)] => 0;
+
+sub new ($class, @args) {
+    my $self = $class->SUPER::new(@args);
+    $self->_update_last_check;
```
Martchus, line 62: It would be good to return `$self` explicitly.

Martchus (approved):

PR #2197, opened 2022-10-31
Don't lessen overall result badness via serial fail autodetect

asdil12:
Only make things worse

Ticket: https://progress.opensuse.org/issues/119461

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -17,6 +17,15 @@ use Mojo::File 'path';
 my $serial_file_pos = 0;
 my $autoinst_log_pos = 0;
 
+my $fail_severity = {
+    '' => 0,
+    na => 0,
+    ok => 100,
+    softfail => 200,
+    fail => 300,
+    canceled => 800
+};
+
 # enable strictures and warnings in all tests globally but allow signatures
 sub import ($self, @) {
     strict->import;
@@ -658,9 +667,11 @@ sub parse_serial_output_qemu ($self) {
                 elsif ($type =~ 'hard|fatal') {
                     $die = 1;
                     $fail_type = 'fail';
-                    $self->{fatal_failure} = $type eq 'fatal';
+                    $self->{fatal_failure} |= $type eq 'fatal';
                 }
                 $self->record_resultfile($message, $message . " - Serial error: $line", result => $fail_type);
+                # don't make overall result better - only worse
+                next if (defined($self->{result}) && ($fail_severity->{$self->{result}} // 999) > $fail_severity->{$fail_type});
                 $self->{result} = $fail_type;
             }
         }
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -116,6 +116,26 @@ subtest parse_serial_output => sub {
     is($basetest->{result}, 'ok', 'test result set to ok');
     is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
 
+    $basetest->{result} = 'softfail';
+    $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'softfail', 'test result stays at softfail on ok match');
+
+    $basetest->{result} = 'fail';
+    $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'fail', 'test result stays at fail on ok match');
+    $basetest->{serial_failures} = [
+        {type => 'soft', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'fail', 'test result stays at fail on softfail match');
+    $basetest->{result} = undef;
+
     $basetest->{serial_failures} = [
         {type => 'soft', message => 'SimplePattern', pattern => qr/Serial/},
     ];


okurz (approved):

okurz:
@mdoucha

mdoucha (changes requested):

basetest.pm
```diff
@@ -17,6 +17,13 @@ use Mojo::File 'path';
 my $serial_file_pos = 0;
 my $autoinst_log_pos = 0;
 
+my $fail_severity = {
+    '' => 0,
+    'ok' => 100,
+    'softfail' => 200,
+    'fail' => 300
```
mdoucha, line 24: The mapping needs a few more values, otherwise the PR looks good.

    select distinct(result) from job_modules;
       result   
    ------------
     none
     failed
     pass
     running
     passed
     skipped
     softfailed
     canceled

asdil12: I'm not sure if that result strings are used there, as basetest.pm will set `ok` while this here uses `passed`.
So there seems to be some kind of translation somewhere.

asdil12: here we go:
`lib/OpenQA/Schema/Result/JobModules.pm`:
```perl
sub update_result {
    my ($self, $r) = @_; 
    my $result = $r->{result} || 'none';
    $result =~ s,fail,failed,;
    $result =~ s,^na,none,;
    $result =~ s,^ok,passed,;
    $result =~ s,^unk,none,;
    $result =~ s,^skip,skipped,;
    $result = 'softfailed' if ($r->{dents} && $result eq 'passed');
    $self->update({result => $result});
    return $result;
}
```

mdoucha, line 24: The mapping needs a few more values, otherwise the PR looks good.

    select distinct(result) from job_modules;
       result   
    ------------
     none
     failed
     pass
     running
     passed
     skipped
     softfailed
     canceled

mdoucha:
Yes, there is translation in OpenQA WebUI code. But AFAICT, the test module may get into skipped or cancelled state before the serial failures get evaluated.

asdil12:
@mdoucha I added some more entries. Do you thing my ordering makes sense?

mdoucha:
I can see you've force-pushed a new commit but it looks identical to the previous one.

basetest.pm
```diff
@@ -658,9 +665,13 @@ sub parse_serial_output_qemu ($self) {
...
                     $die = 1;
                     $fail_type = 'fail';
-                    $self->{fatal_failure} = $type eq 'fatal';
+                    $self->{fatal_failure} |= $type eq 'fatal';
                 }
                 $self->record_resultfile($message, $message . " - Serial error: $line", result => $fail_type);
+                # don't make overall result better - only worse
+                if (defined($self->{result})) {
+                    next if ($fail_severity->{$self->{result}} > $fail_severity->{$fail_type});
+                }
```
Martchus, line 674: You can avoid multiple ifs:
```suggestion
                next if defined($self->{result}) && $fail_severity->{$self->{result}} > $fail_severity->{$fail_type};
```

But you should also better avoid warnings and unintended updates in case the your mapping is insufficient:

```suggestion
                next if defined($self->{result}) && ($fail_severity->{$self->{result}} // 999) > $fail_severity->{$fail_type};
```

t/17-basetest.t
```diff
@@ -116,6 +116,29 @@ subtest parse_serial_output => sub {
...
     is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
 
+    $basetest->{result} = 'softfail';
+    $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'softfail', 'test result stays at softfail');
+    is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
+    $basetest->{result} = undef;
```
Martchus, line 126: The assigned value is never used (so just remove that line).

Martchus (changes requested):

basetest.pm
```diff
@@ -17,6 +17,13 @@ use Mojo::File 'path';
 my $serial_file_pos = 0;
 my $autoinst_log_pos = 0;
 
+my $fail_severity = {
```
Martchus, line 20: This hash probably fits in one line and you can avoid quoting the keys.

asdil12:
> I can see you've force-pushed a new commit but it looks identical to the previous one.

now it should have the changes.

basetest.pm
asdil12: The list just got longer. Also putting it in one line would make it hard to read.
Regarding the quoting I would rather use quoting and not go too deep into the perl rabbit hole (also grepping for quoted strings is easier that way).

basetest.pm
asdil12: The question is if the default should be 999 or 0 then

t/17-basetest.t
```diff
@@ -116,6 +116,28 @@ subtest parse_serial_output => sub {
...
+    is($basetest->{result}, 'softfail', 'test result stays at softfail');
+    is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
+
+    $basetest->{result} = 'fail';
+    $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'fail', 'test result stays at fail');
+    is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
```
okurz, line 133: please use distinct test description strings

okurz (changes requested):

t/17-basetest.t
asdil12: I already did, but I made them more verbose now.

t/17-basetest.t
```diff
@@ -116,6 +116,28 @@ subtest parse_serial_output => sub {
...
+    is($basetest->{result}, 'softfail', 'test result stays at softfail on ok match');
+    is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
+
+    $basetest->{result} = 'fail';
+    $basetest->{serial_failures} = [
+        {type => 'info', message => 'CPU soft lockup detected', pattern => qr/Serial/},
+    ];
+    $basetest->parse_serial_output_qemu();
+    is($basetest->{result}, 'fail', 'test result stays at fail on ok match');
+    is($message, 'CPU soft lockup detected - Serial error: Serial to match', 'log message matches output');
```
okurz, line 133: there is still three times repeated "log message matches output"

asdil12: I removed that lines as that is already tested before.

basetest.pm
```diff
@@ -17,6 +17,21 @@ use Mojo::File 'path';
...
+    '' => 0,
+    'none' => 0,
+    'skipped' => 50,
+    'canceled' => 50,
+    'skipped' => 50,
```
Martchus, line 25: You have `skipped` twice. However, I think only ``, `na`, `ok`, `softfail`, `fail` and `canceled` (with only one "l" on os-autoinst level) are relevant here because the other results are only set by openQA (and should not occur on os-autoinst level). But combined with my other suggestion you don't even need `canceled` because supposedly special results like that should always be preserved (as they are supposedly not considered "good" results everywhere else, too). (And technically you could then also drop `fail` because it is the most severe result anyways.)

basetest.pm
Martchus: I made the default 999 because it makes generally sense to preserve a "special" result outside the severity scale.

Martchus (changes requested):

okurz (approved):

Martchus (approved):

asdil12:
```
/home/dheidler/devel/os-autoinst/basetest.pm: Hash key 'na' with quotes at line 22, column 5.  Avoid useless quotes.  (Severity: 5)
/home/dheidler/devel/os-autoinst/basetest.pm: Hash key 'ok' with quotes at line 23, column 5.  Avoid useless quotes.  (Severity: 5)
/home/dheidler/devel/os-autoinst/basetest.pm: Hash key 'softfail' with quotes at line 24, column 5.  Avoid useless quotes.  (Severity: 5)
/home/dheidler/devel/os-autoinst/basetest.pm: Hash key 'fail' with quotes at line 25, column 5.  Avoid useless quotes.  (Severity: 5)
/home/dheidler/devel/os-autoinst/basetest.pm: Hash key 'canceled' with quotes at line 26, column 5.  Avoid useless quotes.  (Severity: 5)
```
seriously? it even forces barewords?

okurz:
> seriously? it even forces barewords?

On purpose to avoid discussions like in https://github.com/os-autoinst/os-autoinst/pull/2197#pullrequestreview-1165124164 :)

basetest.pm
```diff
@@ -17,6 +17,15 @@ use Mojo::File 'path';
 my $serial_file_pos = 0;
 my $autoinst_log_pos = 0;
 
+my $fail_severity = {
+    '' => 0,
+    na => 0,
+    ok => 100,
+    softfail => 200,
+    fail => 300,
+    canceled => 800
```
perlpunk, line 26: should be "cancelled" :)

Martchus: For os-autoinst it should actually be "canceled" when I remember correctly.

perlpunk: Oh, I didn't realize that the typo is already in the existing code...

Martchus: "canceled" is AE and "cancelled" BE. So not really a typo either. Just inconsistent spelling.

PR #2198, opened 2022-11-02
Allow pausing on failures via developer mode

Martchus:
* Related ticket: https://progress.opensuse.org/issues/119380

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -255,6 +255,14 @@ sub _handle_command_set_pause_on_next_command ($self, $response, @) {
     $self->_respond_ok();
 }
 
+sub _handle_command_pause_test_execution ($self, $response, @) {
+    return $self->_respond_ok if $self->reason_for_pause;    # do nothing if already paused
+    my $reason_for_pause = $response->{reason} // 'manually paused';
+    $self->reason_for_pause($reason_for_pause);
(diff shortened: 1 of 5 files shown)


okurz:
the test report from circleCI seems incomplete but OBS checks already show relevant test failures

testapi.pm
```diff
@@ -875,12 +875,25 @@ sub x11_start_program {    # no:style:signatures
     return $distri->x11_start_program($program, @args);
 }
 
+sub _croak ($command, $error) {
```
okurz, line 878: I wonder if it would be better to move those helper functions outside the test API so that it stays more clean as the real outside interface. I mean just move it to a separate file, WDYT?

Martchus:
If one wants to test this for real, then it is likely best to use it together with the corresponding UI changes for openQA: https://github.com/os-autoinst/openQA/pull/4906

okurz (approved):

Martchus:
It looks like it doesn't work when the post fail hook dies.

Martchus:
It actually doesn't break when when the post fail hook dies. I've just ran into an issue on the UI-side and pushed a fix on the openQA PR.

However, there's one caveat. When the post fail hook dies then the behavior is unchanged. So the test would *not* pause and be considered failed as usual.

By the way, I did the extra testing today in the light of https://progress.opensuse.org/issues/120786 to make sure that this change doesn't lead to test being wrongly considered incomplete instead of failed. So far I see no problems.

okurz (approved):

baierjan (approved):

PR #2200, opened 2022-11-07
Move loop into CommandHandler

kalikiana:

Note: Requires #2196

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -5,17 +5,19 @@ package OpenQA::Isotovideo::CommandHandler;
 use Mojo::Base 'Mojo::EventEmitter', -signatures;
 
 use bmwqemu;
-use testapi 'diag';
+use log qw(diag fctwarn);
 use OpenQA::Isotovideo::Interface;
+use OpenQA::Isotovideo::NeedleDownloader;
 use Cpanel::JSON::XS;
 use Mojo::File 'path';
+use IO::Select;
 use Time::HiRes qw(gettimeofday tv_interval);
 
 use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
 
 
 # io handles for sending data to command server and backend
-has [qw(test_fd cmd_srv_fd backend_fd answer_fd)] => undef;
+has [qw(test_fd cmd_srv_fd backend_fd backend_out_fd answer_fd)] => undef;
 
 # the name of the current test (full name includes category prefix, eg. installation-)
 has [qw(current_test_name current_test_full_name)];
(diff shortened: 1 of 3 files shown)


isotovideo
```diff
@@ -262,92 +246,33 @@ $backend = OpenQA::Isotovideo::Backend->new;
...
-    $rsp->{tags} = $command_handler->tags;
-    if ($rsp->{found} || $rsp->{timeout}) {
-        myjsonrpc::send_json($testfd, {ret => $rsp});
-        $command_handler->clear_tags_and_timeout();
-    }
-    else {
-        _calc_check_delta unless $no_wait;
-    }
-}
+$command_handler->on(signal => sub ($event, $sig) {
```
Martchus, line 266: The event you're emitting is just called `sig` so this handler won't be executed.

Martchus (commented):
The use of the events looks generally good (especially considering events are already used here anyways).

isotovideo
Martchus: It looks still inconsistent. The signal is now called `signalhandler`.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -53,6 +58,28 @@ has backend_requester => undef;
...
+    my $self = $class->SUPER::new(@args);
+
+    my $signalhandler = sub ($sig) { $self->_signalhandler($sig) };
+    $SIG{TERM} = $signalhandler;
+    $SIG{INT} = $signalhandler;
+    $SIG{HUP} = $signalhandler;
+
+    my ($last_check_seconds, $last_check_microseconds) = gettimeofday;
+    $self->last_check_seconds($last_check_seconds);
+    $self->last_check_microseconds($last_check_microseconds);
```
Martchus, line 74: An explicit return would be nice.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -53,6 +58,28 @@ has backend_requester => undef;
...
+# the time of the last asserted screen
+has [qw(last_check_seconds last_check_microseconds)] => 0;
+
+sub new ($class, @args) {
+    my $self = $class->SUPER::new(@args);
+
+    my $signalhandler = sub ($sig) { $self->_signalhandler($sig) };
+    $SIG{TERM} = $signalhandler;
+    $SIG{INT} = $signalhandler;
+    $SIG{HUP} = $signalhandler;
```
Martchus, line 70: I'd extract that into a function, e.g. called `setup_signal_handler` and have the user calling it after constructing the object. (Assigning global variables unconditionally is not what I'd generally expect when invoking a constructor.)

Martchus (changes requested):

OpenQA/Isotovideo/CommandHandler.pm
kalikiana: That's code from #2196 :-)

Martchus (approved):

okurz (changes requested):
codecov/project/fully_covered failed

okurz:
Now patch coverage :)

kalikiana:
> Now patch coverage :)

Well, if you consider it useful I'll mark it as uncoverable. I don't fully understand why it shows up even though we isotovideo is already exempt...

okurz (approved):

PR #2201, opened 2022-11-14
Make blocking mode configurable for video encoder pipe

sdclarke:
The pipe for the video encoder was hardcoded to be non-blocking. Some people may prefer to make it blocking for better user experince in the WebUI live view in cases where there are large screenshot file sizes

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -339,7 +339,7 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     my $pid = open($self->{$pipe_name}, '|-', @cmd);
     my $pipe = $self->{$pipe_name};
     $self->{video_encoders}->{$pid} = {name => $display_name, pipe => $pipe, cmd => join ' ', @cmd};
-    $pipe->blocking(0);
+    $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
 }
 
 sub _start_external_video_encoder_if_configured ($self) {
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -46,6 +46,7 @@ UPLOAD_INACTIVITY_TIMEOUT;integer;300;Specifies the inactivity timeout in second
 NO_DEPRECATE_BACKEND_$backend;boolean;0;Only warn about deprecated backends instead of aborting
 XRES;integer;1024;Resolution of display on x axis. Sets the resolution of the video encoder, and in qemu, the initial console resolution when OFW is set (Power and SPARC), and the EDID resolution for devices that support EDID
 YRES;integer;768;Resolution of display on y axis. Sets the resolution of the video encoder, and in qemu, the initial console resolution when OFW is set (Power and SPARC), and the EDID resolution for devices that support EDID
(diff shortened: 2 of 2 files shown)


doc/backend_vars.asciidoc
```diff
@@ -46,6 +46,7 @@ UPLOAD_INACTIVITY_TIMEOUT;integer;300;Specifies the inactivity timeout in second
 NO_DEPRECATE_BACKEND_$backend;boolean;0;Only warn about deprecated backends instead of aborting
 XRES;integer;1024;Resolution of display on x axis. Sets the resolution of the video encoder, and in qemu, the initial console resolution when OFW is set (Power and SPARC), and the EDID resolution for devices that support EDID
 YRES;integer;768;Resolution of display on y axis. Sets the resolution of the video encoder, and in qemu, the initial console resolution when OFW is set (Power and SPARC), and the EDID resolution for devices that support EDID
+VIDEO_ENCODER_BLOCKING_PIPE;boolean;0;Whether the pipe for writing data to the video encoder should be blocking or not
```
Martchus, line 49: It would be good to state what effect this has (advantages/disadvantages). And what exactly do you mean with "better user experience"?

Martchus (changes requested):

doc/backend_vars.asciidoc
sdclarke: By better user experience I'm referring to the fact that in the case I had, the live view would not update anywhere close to real time as isotovideo was attempting to send ~4M of data in 64K chunks. Making this blocking meant that one could follow the progress of a test run in the live view as the next test wouldn't proceed until the full image had at least been sent over to the videoencoder.

As for advantages and disadvantages, it's hard to say exactly, one big advantage for me would be the working live view. Without sinking even more time into understanding the code it would be difficult for me personally to work out all of the side effects this may have.

Martchus:
```suggestion
VIDEO_ENCODER_BLOCKING_PIPE;boolean;0;Whether the pipe for writing data to the video encoder should be blocking or not; making it blocking might allow following the live view in realtime despite large screenshot file sizes but it is not a well tested configuration
```

sdclarke: Looks good, I've added it

backend/baseclass.pm
```diff
@@ -339,7 +339,8 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     my $pid = open($self->{$pipe_name}, '|-', @cmd);
     my $pipe = $self->{$pipe_name};
     $self->{video_encoders}->{$pid} = {name => $display_name, pipe => $pipe, cmd => join ' ', @cmd};
-    $pipe->blocking(0);
+    my $blocking = $bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE} ? 1 : 0;
+    $pipe->blocking($blocking);
```
Martchus, line 343:
```suggestion
    $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
```

Martchus (approved):

okurz:
@os-autoinst/tools-team this needs manual testing

okurz (approved):

PR #2203, opened 2022-11-14
Migrate from codecov action v2 to v3

kraih:
Noticed that there is already a v3 of the codecov uploader while investigating a problem with v2.

Progress: https://progress.opensuse.org/issues/120175


Martchus (approved):
The change looks good. Let's see whether the CI still works.

kraih:
Still works. 🎉

okurz (approved):

PR #2206, opened 2022-11-15
Move stop_autotest into CommandHandler

kalikiana:


okurz (approved):

ggardet:
This PR leads to incompletes when post fail hooks fail, see: https://progress.opensuse.org/issues/120786

PR #2209, opened 2022-11-16
Move stop_server logic into CommandHandler

kalikiana:

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -75,7 +75,6 @@ require IPC::System::Simple;
 use POSIX qw(:sys_wait_h _exit);
 use Try::Tiny;
 use Mojo::File qw(curfile);
-use Mojo::UserAgent;
 use Mojo::IOLoop::ReadWriteProcess 'process';
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 Getopt::Long::Configure("no_ignore_case");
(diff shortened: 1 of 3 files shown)


isotovideo
```diff
@@ -220,7 +189,7 @@ load_test_schedule;
 
 # start the command fork before we get into the backend, the command child
 # is not supposed to talk to the backend directly
-($cmd_srv_process, $cmd_srv_fd) = commands::start_server($cmd_srv_port = $bmwqemu::vars{QEMUPORT} + 1);
+my ($cmd_srv_process, $cmd_srv_fd) = commands::start_server(my $cmd_srv_port = $bmwqemu::vars{QEMUPORT} + 1);
```
Martchus, line 192: I didn't know it was possible to have to `my` keywords in the same statement.

t/19-isotovideo-command-processing.t
```diff
@@ -410,6 +411,16 @@ subtest 'No readable JSON' => sub {
...
 };
 
+subtest 'Stopping the command server' => sub {
+    # A fake server and port is enough to exercise the code path
+    $command_handler->cmd_srv_port(1);
+    $command_handler->cmd_srv_process(process(sub { })->start);
+    $bmwqemu::vars{JOBTOKEN} = 'deadbeef';
+    stderr_like {
+        $command_handler->stop_server('test execution ended');
+    } qr/stopping command server.+deadbeef/, 'reason logged';
```
Martchus, line 421:
```suggestion
    } qr/stopping command server.+deadbeef/, 'reason logged';
    is $command_handler->cmd_srv_process, undef, 'process unset';
```

t/19-isotovideo-command-processing.t
```diff
@@ -410,6 +411,16 @@ subtest 'No readable JSON' => sub {
     is($command_handler->loop, 0, 'Loop was stopped');
 };
 
+subtest 'Stopping the command server' => sub {
+    # A fake server and port is enough to exercise the code path
+    $command_handler->cmd_srv_port(1);
+    $command_handler->cmd_srv_process(process(sub { })->start);
```
Martchus, line 417: This test might fail if the sub process already existed when `stop_server` is called (and thus `$self->cmd_srv_process->is_running` returns false). Maybe it makes sense to invoke e.g. `sleep infinity` here so the process keeps idling until it is stopped.

isotovideo
okurz: We use it all the times when not using signatures, e.g.:

```perl
my function {
    my ($param1, $param2, %all_the_rest) = @_;
    …
```

Martchus: But in this statement the `my` keyword is only used once (the `my function` line doesn't count into the statement).

kalikiana:
@mergifyio rebase

isotovideo
kraih: It's not very common, since it tends to make the code less readable.

perlpunk: > We use it all the times when not using signatures, e.g.:
> 
> ```perl
> my function {
>     my ($param1, $param2, %all_the_rest) = @_;
>     …
> ```

where do we use `my function`? somewhere in openqa or os-autoinst?

kalikiana:
This is obsolete by now.

PR #2210, opened 2022-11-18
Remove sub standstill_detected from basetest.pm

r-richardson:
While increasing code coverage for basetest.pm ([Ticket: 94952](https://progress.opensuse.org/issues/94952)) i've found this function `standstill_detected` which seems to not be called anywhere within openqa/os-autoinst:

```
╭─░▒▓    ~/Code/tools/repos ·················································· ✔  2.5.8   at 14:34:11  ▓▒░─╮
╰─ ls                                                                                                              ─╯
openQA  openQA-helper  os-autoinst  os-autoinst-distri-opensuse

╭─░▒▓    ~/Code/tools/repos ·················································· ✔  2.5.8   at 14:34:12  ▓▒░─╮
╰─ grep -r 'standstill_detected' ./                                                                                ─╯
grep: ./os-autoinst/build/cover_db/structure/4c742895140b66adb3e1a588e8bae0d2: binary file matches
./os-autoinst/basetest.pm:sub standstill_detected ($self, $lastscreenshot) {
```
Anyone know if this might be used externally or if it is safe to remove ?

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -591,19 +591,6 @@ Return a listref containing hashrefs like this:
 
 sub ocr_checklist () { [] }
 
-sub standstill_detected ($self, $lastscreenshot) {
-    $self->record_screenfail(
-        img => $lastscreenshot,
-        result => 'fail',
-        overall => 'fail'
-    );
-
-    testapi::send_key('alt-sysrq-w');
-    testapi::send_key('alt-sysrq-l');
-    testapi::send_key('alt-sysrq-d');    # only available with CONFIG_LOCKDEP
-    return;
-}
-
 # this is called if the test failed and the framework loaded a VM
 # snapshot - all consoles activated in the test's run function loose their
 # state


Martchus:
We've already checked the opensuse and openQA test distributions and the function wasn't used.

@AdamWill @phil-hands Would you be fine with removing this function?

okurz:
@r-richardson I suggest you check the git history why/when and by whom the function was introduced

Martchus:
We've already checked. It was refactored by @coolo but it wasn't really clear to me how the refactored function was intended to be used. (I haven't looked very long at it, though.)

phil-hands:
[I replied by mail, but that seems not to have turned up yet, so I'll repeat myself to be sure]

I don't recall using it, and looking though our git history seems not to show it ever being used, so getting rid of it seems fine to me.

AdamWill:
We've never used this either, fire away! Thanks for the heads up.

okurz (approved):

coolo:
It was never meant to be used by tests, 33f52cd18cbd6db7b89a176a99b6d4118abb83f2 split testapi from bmwqemu - before bmwqemu loop itself detected stalls and ran test code, after that it moved to a function in the basetest. Later (681c399e730b4b3ab9c64dc643d3958a84058b99) bmwqemu also lost the main loop and it moved into the backend class - where now stills are detected. And so the function became an orphan

Martchus (approved):

PR #2211, opened 2022-11-21
Check for warnings in isotovideo test

kalikiana:
And remove the misleading and unused output handling.
    
See: https://progress.opensuse.org/issues/120786

Note: This would have shown a clear error message when failing tests in #2206. It doesn't address the problem of CI behaving differently.

### DIFF
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -36,13 +36,12 @@ sub isotovideo (%args) {
     my @cmd = ($^X, "$toplevel_dir/isotovideo", '--workdir', $pool_dir, '-d', $args{default_opts}, split(' ', $args{opts}));
     chdir $pool_dir;
     note "Starting isotovideo with: @cmd";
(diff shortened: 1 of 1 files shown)


Martchus (approved):

perlpunk:
> Note: This would have failed and shown a clear error message on #2206.

Really? I put it on top of #2206 and pushed it to my fork, and CI passed.

I still don't know why I get a failure locally and CI passes, but this PR doesn't address this :)

kalikiana:
> See my comment. I don't see how this helps

It helps by revealing the information we used to understand the problem. As for CI I opened #2213.

okurz (approved):

perlpunk:
The description still says that #2206 would have failed with this.

kalikiana:
> The description still says that #2206 would have failed with this.

I re-phrased it.

perlpunk (approved):
I don't know why removing $output as a return value from `isotovideo()` would be necessary but ok...

PR #2216, opened 2022-11-22
Move backend into CommandHandler as well

perlpunk:
to keep the right order of initialization (autotest::start_process, backend)

This is based on the reverted #2206

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -9,6 +9,7 @@ use autotest ();
 use log qw(diag fctwarn);
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::NeedleDownloader;
+use OpenQA::Isotovideo::Backend;
 use Cpanel::JSON::XS;
 use Mojo::File 'path';
 use IO::Select;
@@ -18,7 +19,7 @@ use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
 
 
 # io handles for sending data to command server and backend
-has [qw(test_fd cmd_srv_fd backend_fd backend_out_fd answer_fd)] => undef;
+has [qw(test_fd cmd_srv_fd backend backend_fd backend_out_fd answer_fd)] => undef;
 
 # the running test process
 has test_process => undef;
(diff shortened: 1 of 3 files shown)


okurz (approved):

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -17,7 +19,10 @@ use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
 
 
 # io handles for sending data to command server and backend
-has [qw(test_fd cmd_srv_fd backend_fd backend_out_fd answer_fd)] => undef;
+has [qw(test_fd cmd_srv_fd backend backend_fd backend_out_fd answer_fd)] => undef;
```
kraih, line 22: `undef` is the default value anyway, even without assigning it.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -17,7 +19,10 @@ use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
...
+
+# the running test process
+has test_process => undef;
```
kraih, line 25: The `undef` assignment is not needed here either.

OpenQA/Isotovideo/CommandHandler.pm
perlpunk: ok, but this was there before

OpenQA/Isotovideo/CommandHandler.pm
Martchus: I guess this is done for consistently avoiding quotes around the accessors name.

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -67,6 +72,28 @@ sub new ($class, @args) {
     return $self;
 }
 
+sub init ($self) {
+    $self->start_process;
+    $self->start_backend;
+    return $self;
+}
+
+sub start_process ($self) {
```
Martchus, line 81: This function is starting autotest specifically, so what about
```suggestion
sub start_autotest ($self) {
```

ggardet:
It seems this PR is causing issues on aarch64 workers: https://progress.opensuse.org/issues/121045
Any idea why?

PR #2218, opened 2022-11-23
qemu backend: Use the value of `HDDSIZEGB` var for backing files as well

ggardet:
This will fix aarch64 JeOS tests using raw images where only `HDDSIZEGB` is defined and avoid to define an additional `HDDSIZEGB_1` variable.

Example of test where the HDD is not resized when only  `HDDSIZEGB` is defined: https://openqa.opensuse.org/tests/2897239
But it was resized once  `HDDSIZEGB_1` was set: https://openqa.opensuse.org/tests/2897830

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -155,7 +155,7 @@ sub configure_blockdevs ($self, $bootfrom, $basedir, $vars) {
         my $backing_file = $vars->{"HDD_$i"};
         my $node_id = 'hd' . ($i - 1);
         my $hdd_serial = $vars->{"HDDSERIAL_$i"} || $node_id;
-        my $size = $vars->{"HDDSIZEGB_$i"};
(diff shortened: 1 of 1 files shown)


ggardet:
> Unit tests need to be adapted. Besides that, I hope no other tests rely on the previous behavior.

Am I changing an expected behavior? If yes, I would like to understand the rational. Otherwise, I can update the test.

Martchus:
I don't know what's the expected behavior is (and what the full impact of this change is). We could give this just a try and find out.

ggardet:
So this will lead to some issues. Example of failures with `HDDSIZEGB_1` set:
* rescue system: https://openqa.opensuse.org/tests/2900882 the original HDD size is 40GB, so resizing it to 20G leads to a failure
* zdup upgrade test : https://openqa.opensuse.org/tests/2900455#step/setup_zdup/40 throws to an emergency shell

PR #2220, opened 2022-11-25
Add pico keyboard support to video_stream console

asdil12:
Ticket: https://progress.opensuse.org/issues/119566
VR: https://openqa.suse.de/tests/10038251#step/prepare_firstboot/2

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -5,6 +5,8 @@
 package consoles::video_stream;
 
 use Mojo::Base 'consoles::video_base', -signatures;
+use Mojo::UserAgent;
+use Mojo::URL;
 
 use List::Util 'max';
 use Time::HiRes qw(usleep);
@@ -19,6 +21,13 @@ use constant STREAM_TYPING_LIMIT_DEFAULT => 30;
 use constant DV_TIMINGS_CHECK_INTERVAL => 3;
 
 use constant STALL_THRESHOLD => 4;
+
+use constant DEFAULT_MAX_RES_X => 1680;
+use constant DEFAULT_MAX_RES_Y => 1050;
+use constant DEFAULT_MAX_RES => DEFAULT_MAX_RES_X * DEFAULT_MAX_RES_Y;
+use constant DEFAULT_BYTES_PER_PIXEL => 3;
+use constant DEFAULT_PPM_HEADER_BYTES => 20;
+use constant DEFAULT_VIDEO_STREAM_PIPE_BUFFER_SIZE => DEFAULT_MAX_RES * DEFAULT_BYTES_PER_PIXEL + DEFAULT_PPM_HEADER_BYTES;
 
 sub screen ($self, @) {
     return $self;
@@ -116,7 +125,8 @@ sub connect_remote_video ($self, $url) {
     $self->{ffmpegpid} = open($ffmpeg, '-|', @$cmd)
       or die "Failed to start ffmpeg for video stream at $url";
     # make the pipe size large enough to hold full frame and a bit
-    fcntl($ffmpeg, Fcntl::F_SETPIPE_SZ, 4194304);
+    my $frame_size = $bmwqemu::vars{VIDEO_STREAM_PIPE_BUFFER_SIZE} // DEFAULT_VIDEO_STREAM_PIPE_BUFFER_SIZE;
+    fcntl($ffmpeg, Fcntl::F_SETPIPE_SZ, $frame_size);
     $self->{ffmpeg} = $ffmpeg;
     $ffmpeg->blocking(0);
 
@@ -144,7 +154,7 @@ sub _receive_frame ($self) {
     my $ffmpeg = $self->{ffmpeg};
     $ffmpeg or die 'ffmpeg is not running. Probably your backend instance could not start or died.';
     $ffmpeg->blocking(0);
-    my $ret = $ffmpeg->read(my $header, 20);
+    my $ret = $ffmpeg->read(my $header, DEFAULT_PPM_HEADER_BYTES);
     $ffmpeg->blocking(1);
 
     return undef unless $ret;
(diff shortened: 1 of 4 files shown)


consoles/video_stream.pm
```diff
@@ -116,7 +118,8 @@ sub connect_remote_video ($self, $url) {
     $self->{ffmpegpid} = open($ffmpeg, '-|', @$cmd)
       or die "Failed to start ffmpeg for video stream at $url";
     # make the pipe size large enough to hold full frame and a bit
-    fcntl($ffmpeg, Fcntl::F_SETPIPE_SZ, 4194304);
+    my $frame_size = 1680 * 1050 * 3 + 20;    # resolution * 3 bytes/pix + header
```
ggardet, line 121: The resolution may not be a fixed value, no?

asdil12: it doesn't seem to have to match exactly - just to be large enough.

okurz: pleaes don't use magic numbers though

okurz (changes requested):
Missing test coverage

asdil12:
> pleaes don't use magic numbers though

I thought I was reducing them with this PR by at least providing a proper calculation with a proper comment.
How would you do it instead?

consoles/video_stream.pm
asdil12: I thought I was reducing them with this PR by at least providing a proper calculation with a proper comment.
How would you do it instead?

Martchus:
```suggestion
    my $frame_size = $bmwqemu::vars{VIDEO_STREAM_PIPE_BUFFER_SIZE} // (1680 * 1050 * 3 + 20);    # resolution * 3 bytes/pix + header
```

Unfortunately it doesn't make sense to use the `XRES`/`YRES` variables here because supposedly the resolution might not match that (which is another problem to be resolved but likely out of scope for now).

consoles/video_stream.pm
```diff
@@ -225,6 +228,28 @@ sub send_key_event ($self, $key, $press_release_delay) {
       or die "failed to send '$key' input event";
 }
 
+sub type_string ($self, $args) {
+    my $keyboard_device_url = $bmwqemu::vars{GENERAL_HW_PICO_KEYBOARD_URL};
+    if ($keyboard_device_url) {
+        # send keyboard events using RPi Pico W based keyboard emulator
+        my $url = Mojo::URL->new($keyboard_device_url)->query(type => $args->{text});
+        bmwqemu::diag(Mojo::UserAgent->new->get($url)->result->body);
```
Martchus, line 236: This diag message is useless without better context information. In this form it'll just make a URL show up in the logs with no hint where it comes from and why it was logged. Either remove the logging completely or provide better context. Note that printing the whole body unconditionally as-is might also be problematic (as an HTTP request might generally return all kinds of data, e.g. possibly binary data).

consoles/video_stream.pm
```diff
@@ -225,6 +228,28 @@ sub send_key_event ($self, $key, $press_release_delay) {
...
+    }
+    return $self->SUPER::type_string($args);
+}
+
+sub send_key ($self, $args) {
+    my $keyboard_device_url = $bmwqemu::vars{GENERAL_HW_PICO_KEYBOARD_URL};
+    if ($keyboard_device_url) {
+        # send keyboard events using RPi Pico W based keyboard emulator
+        my $url = Mojo::URL->new($keyboard_device_url)->query(sendkey => $args->{key});
+        bmwqemu::diag(Mojo::UserAgent->new->get($url)->result->body);
```
Martchus, line 247: The remarks for the other function apply here as well.

consoles/video_stream.pm
```diff
@@ -225,6 +228,28 @@ sub send_key_event ($self, $key, $press_release_delay) {
       or die "failed to send '$key' input event";
 }
 
+sub type_string ($self, $args) {
+    my $keyboard_device_url = $bmwqemu::vars{GENERAL_HW_PICO_KEYBOARD_URL};
```
Martchus, line 232: Likely you could have used different hardware to run that HTTP server on. As far as os-autoinst is converned this is just the URL to *some* HTTP server that'll allows sending keystrokes to the SUT *somehow*. So the variable name doesn't need to be more specific than that:
```suggestion
    my $keyboard_device_url = $bmwqemu::vars{GENERAL_HW_KEYBOARD_URL};
```

Please also add documentation for the newly introduced variable.

consoles/video_stream.pm
Martchus, line 236: Don't instantiate a new `Mojo::UserAgent` for every request (especially since this function expected to be called quite frequently). You could use e.g. `($self->{_ua} //= Mojo::UserAgent->new)` instead or maybe there's even already a ua instantiated somewhere in that context.

consoles/video_stream.pm
```diff
@@ -225,6 +228,28 @@ sub send_key_event ($self, $key, $press_release_delay) {
...
+    if ($keyboard_device_url) {
+        # send keyboard events using RPi Pico W based keyboard emulator
```
Martchus, line 234: This code doesn't seem very RPi Pico W specific at all. You're just making an HTTP request here. I suggest to document the interface of the HTTP server and also the variables you introduce. A good place to add such backend-specific documentation would be https://github.com/os-autoinst/os-autoinst/blob/master/doc/backends.md. Of course you can also give a bit more details about your specific setup there and e.g. provide a link to the code of the specific HTTP server you're using.

Martchus (changes requested):
I had to leave a few remarks. Especially documentation would be important.

consoles/video_stream.pm
asdil12: I don't print the url but the body.
This device will only return "text/plain" on this api method:

https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/5f476954266b55e998acab3ce532aaddf4cc44d4/data/generalhw_scripts/rpi_pico_w_keyboard/rpi_pico_w_keyboard.ino#L189-L209

Martchus: But it still lacks context. So at least add a prefix like `response from typing string via HTTP request: `. Then I'll at least be clear where the trash being logged here comes from. And yes, we must assume that the HTTP server possibly returns trash, even though your C++ code is of course bullet-proof :-)

asdil12: added some context

t/26-video_stream.t
```diff
@@ -205,6 +205,24 @@ subtest 'input events' => sub {
...
+    $user_agent_mock->redefine(start => sub ($ua, $tx) { });
+    $user_agent_mock->redefine(get => sub { Mojo::Transaction::HTTP->new });
+    $http->redefine(result => sub { $i = $i + 1; Mojo::Message::Response->new->code(200)->body("hallo") });
+
+    $console->activate;
+    $console->send_key({key => 'a'});
+    $console->send_key({key => 'ctrl-x'});
+    $console->type_string({text => "some test\n"});
+    $console->disable;
+    is $i, 3, 'right number of http requests called';
```
Martchus, line 225: You could also check whether the right requests have been made, e.g. you could push the attempted requests in an array in your mocking (`$tx->req` should return a https://docs.mojolicious.org/Mojo/Message/Request object).

Martchus (commented):
Thanks for the documentation, it looks good now.

t/26-video_stream.t
```diff
@@ -205,6 +205,27 @@ subtest 'input events' => sub {
...
         "s\n", "o\n", "m\n", "e\n", "spc\n", "t\n", "e\n", "s\n", "t\n", "ret\n",
     ], "correct commands sent";
+
+    $bmwqemu::vars{GENERAL_HW_KEYBOARD_URL} = 'http://127.0.0.42:42000/cmd';
+
+    # mock ua
+    my $urls = [];
+    my $user_agent_mock = Test::MockModule->new('Mojo::UserAgent');
+    my $http = Test::MockModule->new('Mojo::Transaction::HTTP');
+    my $req_mock = Test::MockModule->new('Mojo::Message::Request');
```
Martchus, line 215: Thanks for updating tests. This variable is unused now. Also ensure that the modules `Mojo::Transaction::HTTP` and `Mojo::Message::Response` are properly imported (we should not rely on accidental implicit imports).

Martchus (approved):

consoles/video_stream.pm
```diff
@@ -116,7 +118,8 @@ sub connect_remote_video ($self, $url) {
     $self->{ffmpegpid} = open($ffmpeg, '-|', @$cmd)
       or die "Failed to start ffmpeg for video stream at $url";
     # make the pipe size large enough to hold full frame and a bit
-    fcntl($ffmpeg, Fcntl::F_SETPIPE_SZ, 4194304);
+    my $frame_size = $bmwqemu::vars{VIDEO_STREAM_PIPE_BUFFER_SIZE} // (1680 * 1050 * 3 + 20);    # resolution * 3 bytes/pix + header
```
okurz, line 121: Still a lot of magic numbers

ggardet:
We may add documentation about `VIDEO_STREAM_PIPE_BUFFER_SIZE`?

asdil12:
> We may add documentation about `VIDEO_STREAM_PIPE_BUFFER_SIZE`?

done

okurz (approved):

PR #2191, opened 2022-10-10
Make assert_script_sudo work properly in serial terminal

b10n1k:
When `is_serial_terminal` is used the output doesnt need redirection to the serial device. This causes the assertion to fail when the serial terminal is used.

Signed-off-by: ybonatakis <ybonatakis@suse.com>

PR #2223, opened 2022-11-29
Fix assert_script_sudo invocation

b10n1k:
`assert_script_sudo` tries to assert the output from `script_sudo`. However they use their own marker in each case. Moreover it tries to assert the output from invalid flow. That is because the `assert_script_run` has static $wait time when `script_sudo` involves conditions which gives different output depending on that variable.
Fix consists:
- pass the same arguments between the functions
- simplify `assert_script_sudo` and let `script_sudo` set the marker

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -334,33 +334,26 @@ subtest 'script_run' => sub {
     is(background_script_run('sleep 10', output => 'foo'), '1234', 'background_script_run with output returns valid PID');
 };
 
-sub assert_script_sudo_test () {
+sub assert_script_sudo_test ($waittime, $is_serial_terminal) {
     my $mock_testapi = Test::MockModule->new('testapi');
     $mock_testapi->redefine(_handle_found_needle => sub { return $_[0] });
     $mock_testapi->noop(qw(send_key enter_cmd));
+    my $script_sudo = '';
     $mock_testapi->redefine(hashed_string => 'XXX');
     $mock_testapi->redefine(wait_serial => 'XXX-0-');
-    my $script_sudo = '';
-    $mock_testapi->redefine(script_sudo => sub { $script_sudo = "$_[0]" });
-    $mock_testapi->redefine(_set_assert_marker => 'marker');
-    is assert_script_sudo('echo foo'), undef, 'successful assertion of script_sudo (1)';
-    is $script_sudo, 'echo foo; marker', 'script_sudo called like expected(1)';
-    is assert_script_sudo('bash'), undef, 'successful assertion of script_sudo (2)';
-    is $script_sudo, 'bash; marker', 'script_sudo called like expected(2)';
-}
-
-sub set_assert_marker_test ($is_serial_terminal) {
-    my $mock_testapi = Test::MockModule->new('testapi');
-    $mock_testapi->redefine(_handle_found_needle => sub { return $_[0] });
     $mock_testapi->redefine(is_serial_terminal => $is_serial_terminal);
-    my $expected_output = $is_serial_terminal ? ' > /dev/null' : '';
-    is testapi::_set_assert_marker('XXX'), "echo XXX-\$?-$expected_output", 'marker setter returns correct string to the serial';
+    $mock_testapi->redefine(script_sudo => sub { $script_sudo = "$_[0]"; return "XXX-0-" });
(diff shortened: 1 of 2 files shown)


b10n1k:
tests wont work. try to figure out.

b10n1k:
related to https://progress.opensuse.org/issues/120984 [urgent]

Martchus:
Looks like the tidy check doesn't pass.

t/03-testapi.t
```diff
@@ -334,33 +334,23 @@ subtest 'script_run' => sub {
     is(background_script_run('sleep 10', output => 'foo'), '1234', 'background_script_run with output returns valid PID');
 };
 
-sub assert_script_sudo_test () {
+sub assert_script_sudo_test ($waittime, $is_serial_terminal) {
```
perlpunk, line 337: Why did you remove the setting of `$script_sudo`?
This test is still passing even if the changes in testapi.pm would be reverted.
It's good that the test checks the return value of `assert_script_sudo`, but that doesn't change with your code changes.
My suggestion from chat was:
```
$mock_testapi->redefine(script_sudo => sub { $script_sudo = $_[0]; return "XXX-0-" })
```
So you can set `$script_sudo` this way and check later that it contains what is expected.

testapi.pm
```diff
@@ -1026,6 +1021,9 @@ Run C<$command> via C<script_sudo> and then check by C<wait_serial> if its exit
 status is not zero.
 See C<wait_serial> for default timeout.
 
+For this to work the C<str> needs to agree with the corresponding C<str> marker
```
kalikiana, line 1024: What's `str` in this context? If you're referring to the internal variable I wouldn't mention it here since it's impossible for the caller to set it.

t/03-testapi.t
b10n1k: You were right. I put back the `is $script_sudo` checks and test it against the old code. Seems to work as expected this time.

testapi.pm
b10n1k: `$str` is the hashed_string. i documented because it is easy to miss the fact that this part is associated with the hashed string on the other side ($distri->script_sudo).

b10n1k: i can do so. I just thought to make it more official.

b10n1k: Done

kalikiana (approved):

b10n1k:
Can we merge?

Martchus (approved):

Martchus:
Can you provide a verification run?

perlpunk:
So, to clarify some things, the original code did this:
```
sub assert_script_sudo {
    my ($cmd, $wait) = @_;
    my $str = hashed_string("ASS$cmd");
    script_sudo("$cmd; echo $str-\$?- > /dev/$serialdev", 0);
```
So it always redirected to `/dev/$serialdev`, and this was changed to check `is_serial_terminal()` in #2191 
And now the redirect is removed completely, right?

You mentioned https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16002 in the ticket, but I don't see how this is depending on this fix here. The `assert_script_sudo "cmd", timeout => 180`  is already the wrong usage of the function

perlpunk:
Ok, I can see that the redirect is done in [susedistribution::script_sudo](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm#L356) and now understand that this is never called because `$wait` is "timeout" and not 180 because of the bug in the argument passing: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16002

So this PR here will not fix it alone, only https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16002 will.

I'm only wondering how this was ever working, and I also wonder why [os-autoinst::distribution::script_sudo](https://github.com/os-autoinst/os-autoinst/blob/master/distribution.pm#L194) is so different from [susedistribution::script_sudo](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm#L349), and if the former will ever work, given that it is using a different parameter for the `hashed_string`. But I think this can be merged now.

perlpunk (approved):

Martchus:
There's actually already a verification run mentioned in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16002.

PR #2224, opened 2022-11-29
Show escaped version of keyboard emulator server return

asdil12:
Log lines will look like this:
```
[2022-11-29T14:44:04.481416+01:00] [debug] Keyboard emulator says: "type: root\n"
```

VR: https://openqa.suse.de/tests/10050269

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -235,30 +235,43 @@ sub send_key_event ($self, $key, $press_release_delay) {
       or die "failed to send '$key' input event";
 }
 
-sub type_string ($self, $args) {
+=head2 _send_keyboard_emulator_cmd
+
+	_send_keyboard_emulator_cmd($self, %args)
+
+Send keyboard events using RPi Pico W based keyboard emulator
+
+Args to be used:
+
+	type => "hallo welt\n"
+	sendkey => "ctrl-alt-del"
+
+Intended to be used together with this device:
+https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data/generalhw_scripts/rpi_pico_w_keyboard
+
+=cut
+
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

Martchus (approved):
I'll approve but next time move this duplicated code into a separate function.

perlpunk (changes requested):
Oops, there is a typo in the title and commit message

asdil12:
> Oops, there is a typo in the title and commit message

Fixed

okurz (changes requested):
I suggest to un-duplicate the code. Three lines of identical code is enough to warrant a small refactoring

asdil12:
> I suggest to un-duplicate the code. Three lines of identical code is enough to warrant a small refactoring

you asked for it ;)

consoles/video_stream.pm
```diff
@@ -235,29 +235,43 @@ sub send_key_event ($self, $key, $press_release_delay) {
...
+    chomp($server_response);
+    bmwqemu::diag("Keyboard emulator says: " . bmwqemu::pp($server_response));
+}
+
+
+sub type_string ($self, $args) {
+    if ($bmwqemu::vars{GENERAL_HW_KEYBOARD_URL}) {
+        $self->keyboard_emulator_cmd(type => $args->{text});
         return {};
     }
```
Martchus, line 268: The new function could also return `{}`, then this code and the other occurrence boil down to:
```suggestion
    return $self->keyboard_emulator_cmd(type => $args->{text}) if $bmwqemu::vars{GENERAL_HW_KEYBOARD_URL};
```

perlpunk (approved):

Martchus (approved):

PR #2225, opened 2022-11-30
Resume directly after failure in case `assert_script_run` fails

Martchus:
With this change, resuming the test execution ignoring failures will resume the test execution right after the failing `assert_script_run` command (instead on the next module). So far this is only a tweak for `assert_script_run` but it could be extended for other test API functions.

I've been starting with `assert_script_run` because this function was the starting point for the whole "pause on failure" feature.

Related ticket: https://progress.opensuse.org/issues/119380

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -12,6 +12,7 @@ use File::Basename;
 use Socket;
 use IO::Handle;
 use POSIX '_exit';
+use Carp qw();
 use cv;
 use signalblocker;
 use Scalar::Util 'blessed';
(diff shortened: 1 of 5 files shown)


Martchus:
I've been marking this a draft because it would make sense to merge https://github.com/os-autoinst/openQA/pull/4938 first.

mimi1vx (approved):

kalikiana (approved):
I'm liking how you implemented it. Very elegant.

PR #2226, opened 2022-11-30
Add subtest for search_for_expected_serial_failures

r-richardson:
see: https://progress.opensuse.org/issues/94952

### DIFF
--- a/t/17-basetest.t
+++ b/t/17-basetest.t
@@ -521,4 +521,14 @@ subtest rollback_activated_consoles => sub {
     is_deeply(\@selected_consoles, [{cmd => 'backend_select_console', testapi_console => 'last_milestone_console'}], 'last milestone console selected') or diag explain \@selected_consoles;
 };
 
+subtest search_for_expected_serial_failures => sub {
+    $bmwqemu::vars{BACKEND} = 'qemu';
+    my $basetest = basetest->new();
+    my $mock_basetest = Test::MockModule->new('basetest');
+    $mock_basetest->mock(run => sub { die 'test failure' });
+    $mock_basetest->mock(parse_serial_output_qemu => sub { $basetest->{result} = 'successfully called function' });
+    $basetest->runtest();
+    is($basetest->{result}, 'successfully called function', 'search for expected serial failures is working');
+};
+
 done_testing;


t/17-basetest.t
```diff
@@ -521,4 +521,20 @@ subtest rollback_activated_consoles => sub {
     is_deeply(\@selected_consoles, [{cmd => 'backend_select_console', testapi_console => 'last_milestone_console'}], 'last milestone console selected') or diag explain \@selected_consoles;
 };
 
+subtest search_for_expected_serial_failures => sub {
+    $bmwqemu::vars{BACKEND} = 'qemu';
+    my $basetest_class = 'basetest';
```
Martchus, line 526: I would just put the string literal `'basetest'` where it is needed (because this will always be the same anyways).

t/17-basetest.t
```diff
@@ -521,4 +521,20 @@ subtest rollback_activated_consoles => sub {
...
+    Test::MockModule->new("autotest")->noop('set_current_test');
```
Martchus, line 527: This like is not effective. You need to assign `Test::MockModule->new("autotest")` to a variable because if the object's lifetime ends the mocking will be undone.

t/17-basetest.t
```diff
@@ -521,4 +521,20 @@ subtest rollback_activated_consoles => sub {
...
+    my $mock_basetest = Test::MockModule->new($basetest_class);
+    $mock_basetest->mock(run => sub { die(); });
```
Martchus, line 529: I'd at least give it some message:
```suggestion
    $mock_basetest->mock(run => sub { die 'test failure' });
```

Martchus (changes requested):

Martchus (changes requested):

t/17-basetest.t
```diff
@@ -521,4 +521,19 @@ subtest rollback_activated_consoles => sub {
     is_deeply(\@selected_consoles, [{cmd => 'backend_select_console', testapi_console => 'last_milestone_console'}], 'last milestone console selected') or diag explain \@selected_consoles;
 };
 
+subtest search_for_expected_serial_failures => sub {
+    $bmwqemu::vars{BACKEND} = 'qemu';
+    my $mock_func = Test::MockModule->new("autotest")->noop('set_current_test');
```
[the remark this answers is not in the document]
r-richardson: yes, the line was actually not needed, removed it :)

t/17-basetest.t
```diff
@@ -521,4 +521,18 @@ subtest rollback_activated_consoles => sub {
...
+    my $mock_basetest = Test::MockModule->new('basetest');
+    $mock_basetest->mock(run => sub { die 'test failure' });
+    my $basetest = bless({
+            details => [],
+            name => 'foo',
+            category => 'category1',
+            execute_time => 42,
+    }, 'basetest');
+    $basetest->runtest();
+    is($basetest->{result}, 'fail', 'search for expected serial failures is working');
```
Martchus, line 535: I'm not sure whether that check is good enough to tell that `search_for_expected_serial_failures` has actually worked. That function will call another function to assign that result but other code paths do the same so the result will likely be `fail` anyways. I'm not sure to which level you need functions to be called but if `parse_serial_output_qemu` has already coverage it would make sense to mock it and check whether it is called.

Martchus (changes requested):

Martchus (approved):

kalikiana (approved):

PR #2227, opened 2022-12-06
Move backend into CommandHandler

perlpunk:
Issues: https://progress.opensuse.org/issues/81899 https://progress.opensuse.org/issues/121045

This is a very small step to move variables into the CommandHandler module, as the previous PRs had various issues.

For now we make it a singleton, as the CommandHandler instance has not been initialized at this point.

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -8,6 +8,7 @@ use bmwqemu;
 use log qw(diag fctwarn);
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::NeedleDownloader;
+use OpenQA::Isotovideo::Backend;
 use Cpanel::JSON::XS;
 use Mojo::File 'path';
 use IO::Select;
@@ -68,6 +69,12 @@ sub new ($class, @args) {
     $self->_update_last_check;
     return $self;
 }
+
+my $backend;
+
+sub create_backend ($self) { $backend = OpenQA::Isotovideo::Backend->new }
+sub backend ($self) { $backend }
+
 
 sub setup_signal_handler ($self) {
     my $signal_handler = sub ($sig) { $self->_signal_handler($sig) };
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -72,7 +72,7 @@ sub new ($class, @args) {
 
 my $backend;
 
-sub create_backend ($self) { $backend = OpenQA::Isotovideo::Backend->new }
+sub create_backend ($self) { $backend ||= OpenQA::Isotovideo::Backend->new }
 sub backend ($self) { $backend }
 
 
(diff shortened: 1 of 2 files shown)


okurz (approved):

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -69,6 +70,12 @@ sub new ($class, @args) {
     return $self;
 }
 
+my $backend;
+
+sub create_backend ($self) { $backend = OpenQA::Isotovideo::Backend->new }
+sub backend ($self) { $backend }
```
Martchus, line 76: Wouldn't it make more sense to store the backend as a member variable and return it like this?
```suggestion
sub backend ($self) { $self->{_backend} }
```
Otherwise it is rather weird that this function randomly creates the last backend created by an arbitrary instance.

Martchus (dismissed):

OpenQA/Isotovideo/CommandHandler.pm
perlpunk: Yes, but I wrote "**For now** we make it a singleton".
In #2216 it was done as a member variable, but I wanted to make **small** steps, like I said, because we had this regression which only one person on only one architecture only in cloud appeared.
In a **next** step I would move things further, and then it would be possible to make it a member variable.

isotovideo
```diff
@@ -242,29 +240,29 @@ bmwqemu::save_vars();
 my $testfd;
 ($testprocess, $testfd) = autotest::start_process();
 
-$backend = OpenQA::Isotovideo::Backend->new;
+OpenQA::Isotovideo::CommandHandler->create_backend;
```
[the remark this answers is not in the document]
perlpunk: A singleton is a normal thing and not something that has surprising side effects automatically.
And as I said, it is a first small step bwcause #2216 was too big and finding out what caused the regression proved to be impossible so far, as we can't access the platform ourselves where it happened.

OpenQA/Isotovideo/CommandHandler.pm
perlpunk: and regarding

> creates the last backend created by an arbitrary instance

There will only be one instance anyway.
I made this clear now by using `$backend ||=`.

It's the command handler, there is only one instance of it, and OpenQA::Isotovideo::Backend is a wrapper around `$bmwqemu::backend` (a global variable) anyway, so there can also be only one per process.

perlpunk: With my current change I *can't* make it a member variable, because there is no instance of CommandHandler yet.
That's planned as one of the next steps.

perlpunk: I renamed `$self` to `$class` now to make it clear.

okurz: I think that should suffice.

okurz:
@perlpunk can you please squash the WIP commits? I would like this PR in as soon as possible to be able to proceed.

PR #2233, opened 2022-12-15
Add tesseract-ocr to test OCR in CI

mimi1vx:
Install tessseract so we can test OCR


kalikiana (approved):

baierjan (approved):

coolo:
There is a typo in the commit log. It's dependencies and tesseract is with just 2 s

baierjan (approved):

PR #2236, opened 2022-12-16
Fix uncoverable comments

perlpunk:
The "count:n" is only for cases where the following line has more than one statement, and n actually means the specific number of the statement, not the count of statements to mark uncoverable, and since there is only one statement, count:2 was targeting statement number 2 which doesn't exist.

### DIFF
--- a/t/02-test_ocr.t
+++ b/t/02-test_ocr.t
@@ -25,9 +25,8 @@ cv::init();
 require tinycv;
 
 unless (which('tesseract')) {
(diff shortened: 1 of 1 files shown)


baierjan (approved):

asdil12 (approved):

perlpunk:
@asdil12 why did you merge? It would have been good to let mergify do that and actually see the coverage changes first

PR #2241, opened 2023-01-11
Mark uncoverable statements and remove dead code

mimi1vx:
https://progress.opensuse.org/issues/121357

### DIFF
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -426,9 +426,9 @@ subtest 'qemu is not called on an empty file when ISO_1 is an empty string' => s
     my $mock_proc = Test::MockModule->new('OpenQA::Qemu::Proc');
     my $call_count = 0;
     $mock_proc->redefine(get_img_size => sub {
-            my ($iso) = @_;
-            $call_count++;
-            die 'get_img_size called on an empty string' unless $iso;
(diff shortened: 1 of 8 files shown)


okurz:
s/uncovereable/uncoverable/

t/18-qemu.t
```diff
@@ -426,9 +426,9 @@ subtest 'qemu is not called on an empty file when ISO_1 is an empty string' => s
     my $mock_proc = Test::MockModule->new('OpenQA::Qemu::Proc');
     my $call_count = 0;
     $mock_proc->redefine(get_img_size => sub {
-            my ($iso) = @_;
-            $call_count++;
-            die 'get_img_size called on an empty string' unless $iso;
+            my ($iso) = @_;    # uncoverable statement
```
perlpunk, line 429: Hm, if a mocked function is never called, maybe it doesn't need to be mocked at all.
I just ran the test without this `redefine`, and it passed.

perlpunk: Funnily, this was added by me, so it seems I overlooked something.

mimi1vx: I think this function is called -> if tests fail :D

okurz (approved):

kalikiana (approved):

PR #2249, opened 2023-01-20
Export TESSDATA_PREFIX in check phase of os-autoinst.spec

mimi1vx:
https://progress.opensuse.org/issues/123193

PR #2250, opened 2023-01-20
Set tesseract to quiet mode and remove 'stderr_like' checks

mimi1vx:
https://progress.opensuse.org/issues/123193

New tesseract-ocr in tumbleweed removed debug output and now reports only warnings and errors.


mimi1vx:
merged into #2249

PR #2252, opened 2023-01-21
Improvements in testapi, preparations for #1696

okurz:
* testapi: Improve wrong call handling in get_test_data
* distribution: Use "enter_cmd" where applicable in become_root
* t: Add call tests for multiple missing testapi functions

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -777,6 +777,7 @@ subtest 'check quiet option on script runs' => sub {
     is(assert_script_run('true', quiet => 0), undef, 'assert_script_run with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
     ok(!validate_script_output('script', sub { m/output/ }, quiet => 0), 'validate_script_output with _QUIET_SCRIPT_CALLS=1 and quiet=>0');
     delete $bmwqemu::vars{_QUIET_SCRIPT_CALLS};
+    $mock_testapi->unmock('wait_serial');
 };
 
 subtest 'host_ip, autoinst_url' => sub {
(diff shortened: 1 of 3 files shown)


testapi.pm
```diff
@@ -1144,7 +1144,8 @@ This will return content of the file located in data/autoyast/autoinst.xml
 
 sub get_test_data {
     my ($path) = @_;
-    $path = get_var('CASEDIR') . '/data/' . $path;
+    defined $bmwqemu::vars{CASEDIR} or die 'Need variable CASEDIR';
+    $path = $bmwqemu::vars{CASEDIR} . '/data/' . $path;
```
Martchus, line 1148:
```suggestion
    $path = "$bmwqemu::vars{CASEDIR}/data/$path";
```

kalikiana (approved):

Martchus (approved):

t/03-testapi.t
```diff
@@ -1038,4 +1039,24 @@ subtest 'mouse click' => sub {
...
+lives_ok { become_root } 'become_root can be called';
+like(exception { ensure_installed }, qr/implement.*for your distri/, 'ensure_installed can be called');
+lives_ok { hold_key('ret') } 'hold_key can be called';
+lives_ok { release_key('ret') } 'release_key can be called';
+lives_ok { reset_consoles } 'reset_consoles can be called';
+lives_ok { power('on') } 'power can be called';
+lives_ok { save_memory_dump } 'save_memory_dump can be called';
+like(exception { save_storage_drives }, qr/should be called.*post_fail_hook/, 'save_storage_drives should be called special');
+lives_ok { freeze_vm } 'freeze_vm can be called';
+lives_ok { resume_vm } 'resume_vm can be called';
```
perlpunk, line 1054: Having tests that only check if a function does not die might increase test coverage, but we don't actually test that the functions are doing what we expect. Is that still planned, or do you think we don't need it here for some reason?

okurz: Well, I only added tests to cover the call signatures. I think having better functional tests would be good but that's not related to introducing signatures, right?

PR #2253, opened 2023-01-23
Catch intermediary errors between attempts to clone

kalikiana:
Let the unit test die to mirror the behavior of checkout_git_repo_and_branch.

Also don't retry if history could not be fetched.

See: https://progress.opensuse.org/issues/123556

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -102,7 +102,17 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
     my $local_path = $url->path->parts->[-1] =~ s/\.git$//r;
     my $tries = $args{retry_count};
     while ($tries--) {
-        last unless clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable;
+        last if try {
+            clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable;
+        }
+        catch {
+            my $error_message = $_;
+            # Die if history could not be fetched
+            die $error_message if grep /Could not find.+in complete history/, $error_message;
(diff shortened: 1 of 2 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -102,7 +102,17 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
...
+            clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable;
+        }
+        catch {
+            my $error_message = $_;
+            # Die if history could not be fetched
+            die $error_message if grep /Could not find.+in complete history/, $error_message;
+            # Die on the final attempt
+            die $error_message unless $tries;
+            bmwqemu::diag $error_message;
+        };
```
Martchus, line 115: I think this will work as-is but should be clarified with explicit `return` statements (as the whole `try { … } catch { … };` invocation will simply return whatever the last line in those blocks returned).

The last line in the `catch`-block might in fact return the falsy value `undef` but that's not very clear. So adding a `return 0;` in the catch block would make sense to clarify that we don't want to run into the `last` case then.

The `try` case relies on `clone_git` returning a truthy value in case of success. That's already clear in `clone_git` itself which contains explicit return statements. Maybe it would make sense to add an explicit `return` statement in the `try`-block as well to make it clear that we care about the return value here.

I'm also wondering whether it is the best idea to re-throw exceptions this way. Wouldn't this lead to duplicate `in line …` suffixes (which would be rather confusing)? Maybe it could at least be avoided by using a trailing new line (and then `chomp`).

Martchus (changes requested):

kalikiana:
This has been resolved differently.

PR #2254, opened 2023-01-25
sshVirtsh: Show virsh stderr output on virsh start failure

mdoucha:
Svirt backends sometimes fail to start and the error message shown in webUI is useless most of the time. The real error can be found only in autoinst-log.txt. Print the real error into the test failure box.


okurz (approved):

jlausuch (approved):

perlpunk (approved):

pevik:
@mdoucha @Martchus thanks for further improving my original code.

PR #2255, opened 2023-01-26
Introduce VNC_EXTRA_VARS variable

grisu48:
Introduce the `VNC_EXTRA_VARS` variable that allows users to pass additional parameters to the `-vnc` argument of the QEMU backend.

Related ticket: https://progress.opensuse.org/issues/121261

Verification run: [With VNC_EXTRA_VARS](http://unarmed-1.qa.suse.de/tests/985#settings) | [Without VNC_EXTRA_VARS](http://unarmed-1.qa.suse.de/tests/942#settings)

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -931,7 +931,9 @@ sub start_qemu ($self) {
 
         if ($vars->{VNC}) {
             my $vncport = $vars->{VNC} !~ /:/ ? ":$vars->{VNC}" : $vars->{VNC};
-            sp('vnc', [qv "$vncport share=force-shared"]);
+            my $extravars = $vars->{VNC_EXTRA_VARS};
+            $extravars = defined $extravars ? " $extravars" : '';
+            sp('vnc', [qv "$vncport share=force-shared$extravars"]);
             sp('k', $vars->{VNCKB}) if $vars->{VNCKB};
         }
 
--- a/doc/backend_vars.asciidoc
+++ b/doc/backend_vars.asciidoc
@@ -194,6 +194,7 @@ VDE_PORT;integer;worker instance + 10;number of vde switch port to connect
 VDE_SOCKETDIR;string;.;directory where vde_switch control socket is to be found
 VDE_USE_SLIRP;integer;1;whether to start slirpvde
 VNC;integer;worker instance + 90;Display on which VNC server is running. Actual port is 5900 + VNC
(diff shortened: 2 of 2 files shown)


Martchus (changes requested):

okurz:
+1 good idea

backend/qemu.pm
```diff
@@ -931,7 +931,12 @@ sub start_qemu ($self) {
 
         if ($vars->{VNC}) {
             my $vncport = $vars->{VNC} !~ /:/ ? ":$vars->{VNC}" : $vars->{VNC};
-            sp('vnc', [qv "$vncport share=force-shared"]);
+            my $extravars = $vars->{VNC_EXTRA_VARS};
+            if($extravars) {
+                sp('vnc', [qv "$vncport share=force-shared $extravars"]);
+            } else {
+                sp('vnc', [qv "$vncport share=force-shared"]);
+            }
```
Martchus, line 939: I'd avoid the repetition and use `defined` because someone might for some reason want to append `0`.

```suggestion
            my $extravars = $vars->{VNC_EXTRA_VARS};
            $extravars = defined $extravars ? " $extravars" : '';
            sp('vnc', [qv "$vncport share=force-shared$extravars"]);
```

Martchus (changes requested):

backend/qemu.pm
okurz: Then I guess it can be as simple as:

```suggestion
            my $extravars = $vars->{VNC_EXTRA_VARS} // '';
            sp('vnc', [qv "$vncport share=force-shared$extravars"]);
```

okurz (changes requested):

backend/qemu.pm
grisu48: I took @Martchus suggestion. @okurz suggestion is missing a space AFAICT.

Martchus (approved):

backend/qemu.pm
okurz: You are right

grisu48:
Failed test 'missing documentation for backend QEMU variable VNC_EXTRA_VARS, please update backend_vars'` - However the value has been added in this commit.

Martchus:
Yes, the test seems to be wrong. I don't know what's the problem with it out of my head so I suggest to debug the test locally.

okurz (approved):

PR #2257, opened 2023-02-06
Remove empty directory after failed git clone attempt

baierjan:
Includes:
- #2258

Reference: https://progress.opensuse.org/issues/123888

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -38,7 +38,10 @@ sub calculate_git_hash ($git_repo_dir) {
 }
 
 sub clone_git ($local_path, $clone_url, $clone_depth, $branch, $dir, $dir_variable) {
-    return bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists" if -e $local_path;
+    if (-e $local_path) {
+        bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists";
+        return 1;
+    }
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

OpenQA/Isotovideo/Utils.pm
```diff
@@ -104,6 +104,7 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
     while ($tries--) {
         last unless clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable;
         bmwqemu::diag "Clone failed, retries left: $tries of $args{retry_count}";
+        rmdir $local_path;
```
Martchus, line 107: The directory might not be empty (as Git might have failed somewhere in the middle). I suggest to use https://perldoc.perl.org/File::Path#rmtree%28-%24dir-%29 instead. Also take care of the behavior when the directory has not been created (no exception should be thrown in this case).

Martchus (changes requested):

OpenQA/Isotovideo/Utils.pm
baierjan: I am even no longer sure, if *any* deletion should take place. We might end up debugging weird issues and deleting something completely different for multiple invocations when `$dir_variable` contains same repositories in different namespaces.

As this should be needed to fix the referenced issues, I will take this aside, at least for a while.

baierjan:
This is not needed right now.

PR #2258, opened 2023-02-06
Fix retry logic for git clone

baierjan:
References:
- https://progress.opensuse.org/issues/123873
- https://progress.opensuse.org/issues/123556
- https://progress.opensuse.org/issues/123888

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -38,7 +38,10 @@ sub calculate_git_hash ($git_repo_dir) {
 }
 
 sub clone_git ($local_path, $clone_url, $clone_depth, $branch, $dir, $dir_variable) {
-    return bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists" if -e $local_path;
+    if (-e $local_path) {
+        bmwqemu::diag "Skipping to clone '$clone_url'; $local_path already exists";
+        return 1;
+    }
     bmwqemu::fctinfo "Cloning git URL '$clone_url'";
     my $branch_args = '';
     if ($branch) {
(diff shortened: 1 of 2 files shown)


OpenQA/Isotovideo/Utils.pm
```diff
@@ -101,10 +104,18 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
     my $clone_url = $url->fragment(undef)->to_string;
     my $local_path = $url->path->parts->[-1] =~ s/\.git$//r;
     my $tries = $args{retry_count};
-    while ($tries--) {
-        last unless clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable;
-        bmwqemu::diag "Clone failed, retries left: $tries of $args{retry_count}";
-        sleep($args{retry_interval} // 5);
+  retry: {
```
Martchus, line 107: Where is this label used?

OpenQA/Isotovideo/Utils.pm
```diff
@@ -101,10 +104,18 @@ sub checkout_git_repo_and_branch ($dir_variable, %args) {
...
+        do {
+            my $status;
+            eval { $status = clone_git $local_path, $clone_url, $args{clone_depth}, $branch, $dir, $dir_variable };
+            $error = $@;
+            last if $status;
+            bmwqemu::diag "Clone failed, retries left: $tries of $args{retry_count}";
+            rmdir $local_path;
+            sleep($args{retry_interval} // 5);
+        } while ($tries-- > 0);
+        die $error;
```
Martchus, line 118: This `die` is outside of the loop. Wouldn't we always run into this case?

Martchus (changes requested):
Even if this code makes sense it is perhaps very hard to follow.

OpenQA/Isotovideo/Utils.pm
baierjan: No, we hit it only if the `last` is not used in the inner loop. The `last` will end the named block, not the do-while.

Martchus: Ok, this makes sense. I'm still wondering whether we can simplify this. Having a label that is not really referenced and having two nested loops seems a bit too complicated for this use-case.

baierjan: I really tried, this construction was suggested on https://www.perltutorial.org/perl-do-while/

baierjan: I can probably use `last retry if $status;` if you want the label to be mentioned somewhere.

okurz: Would it work to put this code into a sub-function and return from there instead of a loop with label?

okurz (commented):
Please describe in both commits a bit more explanation. I wonder why you turn of signing. Sounds useful to do but I'd still like to learn that brought you to this. And also for the retry commit

OpenQA/Isotovideo/Utils.pm
baierjan: I managed to rearrange the code slightly so the second loop is no longer needed. It behave slightly differently (setting `$bmwqemu::vars{$dir_variable}` before it dies)  , hopefully it does not matter.

okurz:
+1, @kalikiana you might be interested as well

Martchus (approved):

kalikiana (approved):
Nice. And you improved the unit test coverage as well, I think that's a really good idea here!

PR #2259, opened 2023-02-06
Allow debugging JSON RPC per job

richiejp:
Occasionally it appears that the wait_serial command is sent by the testapi frontend, but never received by the backend. This has only been observed on a single job after a single LTP test.

It is not easily reproducible and so we need to start debugging in production.

We need to confirm or rule out that the JSON RPC is being lost. Meanwhile we do not want to enable debugging for all jobs that a worker runs.

With this change we can enable debugging JSON RPC on a single job.

https://progress.opensuse.org/issues/121774

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -28,12 +28,12 @@ sub send_json ($to_fd, $cmd) {
     $cmdcopy{json_cmd_token} ||= bmwqemu::random_string(8);
 
     my $json = $cjx->encode(\%cmdcopy);
-    if (DEBUG_JSON) {
+    if (DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC}) {
         my $copy = $json;
         # shorten long content
         $copy =~ s/"([^"]{30})[^"]+"/"$1"/g;
         my $fd = fileno($to_fd);
-        bmwqemu::diag("($$) send_json($fd) JSON=$copy");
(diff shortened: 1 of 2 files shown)


Martchus (approved):

t/24-myjsonrpc-debug-var.t
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/perl
+#
+# Copyright 2019-2020 SUSE LLC
```
okurz, line 3:
```suggestion
# Copyright SUSE LLC
```

Martchus (approved):

okurz (approved):

myjsonrpc.pm
```diff
@@ -28,12 +28,12 @@ sub send_json ($to_fd, $cmd) {
...
 
     my $json = $cjx->encode(\%cmdcopy);
-    if (DEBUG_JSON) {
+    if (DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC}) {
         my $copy = $json;
         # shorten long content
         $copy =~ s/"([^"]{30})[^"]+"/"$1"/g;
         my $fd = fileno($to_fd);
-        bmwqemu::diag("($$) send_json($fd) JSON=$copy");
+        bmwqemu::diag("send_json($fd) JSON=$copy");
```
perlpunk, line 36: Did you have a reason for removing the process id `$$`?

PR #2264, opened 2023-02-27
Avoid setting misleading reason for handled errors

Martchus:
The current approach with the die handler might record errors that have actually been handled when they happened (and thus were not fatal). When then an actually fatal error occurs we might end up reporting only handled and actually irrelevant error. This change simply wraps the `isotovideo` code within an `eval` to only catch fatal errors to avoid this.

Note that these code changes only appear huge because of indentation changed.

Related ticket: https://progress.opensuse.org/issues/124934

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -151,21 +151,7 @@ eval {
     }
 
     diag 'isotovideo ' . ($return_code ? 'failed' : 'done');
-
-    my $clean_shutdown;
-    if (!$return_code) {
-        eval {
-            $clean_shutdown = $bmwqemu::backend->_send_json({cmd => 'is_shutdown'});
-            diag('backend shutdown state: ' . ($clean_shutdown // '?'));
-        };
-
-        # don't rely on the backend in a sane state if we failed - just stop it later
-        eval { bmwqemu::stop_vm(); };
-        if ($@) {
-            bmwqemu::serialize_state(component => 'backend', msg => "unable to stop VM: $@", error => 1);
-            $return_code = 1;
-        }
-    }
+    my $clean_shutdown = $runner->handle_shutdown(\$return_code);
 
     # read calculated variables from backend and tests
     bmwqemu::load_vars();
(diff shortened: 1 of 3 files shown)


isotovideo
```diff
@@ -109,73 +108,71 @@ chdir $options{workdir} if $options{workdir};
...
+    diag 'isotovideo ' . ($return_code ? 'failed' : 'done');
+
+    my $clean_shutdown;
+    if (!$return_code) {
+        eval {
+            $clean_shutdown = $bmwqemu::backend->_send_json({cmd => 'is_shutdown'});
+            diag('backend shutdown state: ' . ($clean_shutdown // '?'));
+        };
+
+        # don't rely on the backend in a sane state if we failed - just stop it later
```
perlpunk, line 162: I would add a `local $@` here, otherwise it might survive the outer eval.
(but I'm currently looking at the diff from my phone)

Martchus: Not sure why that would be the case. Nested `eval`s seem to work as expected:

```
  DB<6> eval { print "foo\n"; eval { die "bar\n"; }; if($@) { print "inner: $@" } }; if($@) { print "outer: $@" }
foo
inner: bar

  DB<7> eval { die "foo\n"; eval { die "bar\n"; }; if($@) { print "inner: $@" } }; if($@) { print "outer: $@" }
outer: foo

  DB<8> eval { print "foo\n"; eval { die "bar\n"; }; if($@) { print "inner: $@" } die "baz\n" }; if($@) { print "outer: $@" }
foo
inner: bar
outer: baz
```

kraih: Tina might be thinking of a misfeature in older Perl versions. We now depend on Perl 5.20 though, so this shouldn't be a concern here.

perlpunk: ah ok, thanks

perlpunk (approved):

kraih (approved):

PR #2272, opened 2023-03-01
wip -- bisecting where fully_covered fails - part 2

okurz:


okurz:
Removing

```
          - OpenQA
          - backend
          - consoles/VMWare.pm
          - consoles/VNC.pm
          - consoles/localXvnc.pm
          - consoles/network_console.pm
          - consoles/serial_screen.pm
          - consoles/sshSerial.pm
          - consoles/virtio_terminal.pm
```

showed same result, 98.94% whereas https://github.com/os-autoinst/os-autoinst/pull/2271/files is good with

```
          - bmwqemu.pm
          - log.pm
          - mmapi.pm
          - osutils.pm
          - signalblocker.pm
          - t
          - xt
```

removed so updating to include the original entries again, only excluding `t/`

okurz:
this also shows 100% for fully_covered so seems "t/" is the culprit. Let's try what happens if we explicitly ignore 't' instead of not listing it.

PR #2274, opened 2023-03-02
WIP - Add utilities for DEMO_MODE

perlpunk:
Issue: https://progress.opensuse.org/issues/125378

This is a draft with a simple `mouse_move` function in testapi that just calls `mouse_set` after moving 5 pixels step by step. But maybe that can be done in the backend to be a bit more performant.

* Add demosleep()
* Add mouse_move()
* In assert_and_click use mouse_move in demo mode

Test it with the following repositories:
* https://github.com/perlpunk/openqa-simple-starter
* https://github.com/perlpunk/openqa-simple-starter-needles

use these settings: `DEMO_MODE=1 SCREENSHOTINTERVAL=0.1`

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -36,7 +36,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   hold_key release_key
 
   assert_screen check_screen assert_and_dclick save_screenshot
-  assert_and_click mouse_hide mouse_set mouse_click
+  assert_and_click mouse_hide mouse_set mouse_move mouse_click
   mouse_dclick mouse_tclick match_has_tag click_lastmatch mouse_drag
 
   assert_script_run script_run background_script_run
@@ -55,7 +55,7 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   switch_network
   save_memory_dump save_storage_drives freeze_vm resume_vm
 
-  diag hashed_string
+  diag hashed_string demosleep
 
   save_tmp_file get_test_data
 );
@@ -507,7 +507,13 @@ sub click_lastmatch (%args) {
     # Calculate the absolute click point.
     my ($x, $y) = _calculate_clickpoint($last_matched_needle, $relevant_area, $relative_click_point);
     bmwqemu::diag("clicking at $x/$y");
-    mouse_set($x, $y);
+    if ($bmwqemu::vars{DEMO_MODE}) {
+        mouse_move($x, $y);
+        sleep 1.5;
+    }
+    else {
+        mouse_set($x, $y);
+    }
     if ($args{dclick}) {
         mouse_dclick($args{button}, $args{clicktime});
     }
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -1477,6 +1483,42 @@ sub mouse_set ($mx, $my) {
...
+        }
+        else {
+            $curx += $diffx < 0 ? -$pixels : +$pixels;
+        }
+        if (abs($diffy) < $pixels) {
+            $cury = $y;
+        }
+        else {
+            $cury += $diffy < 0 ? -$pixels : +$pixels;
+        }
```
Martchus, line 1513: I think you want to use the `clamp` function here. It looks like Perl doesn't have such a function but you can find a possible implementation in C++ here: https://en.cppreference.com/w/cpp/algorithm/clamp

Then it would just be `$curx += clamp($x - $curx, -$pixels, +$pixels)` and the same for y.

Martchus: I've just translated the C++ implementation to Perl:

```
sub clamp($v, $lo, $hi) { $v < $lo ? $lo : $hi < $v ? $hi : $v }
```

testapi.pm
```diff
@@ -1477,6 +1483,42 @@ sub mouse_set ($mx, $my) {
...
+Slowly move mouse pointer to given coordinates
+
+=cut
+
+sub mouse_move ($x, $y, $pixels = 5) {
+    my $old = query_isotovideo('backend_get_last_mouse_set');
+    my $curx = $old->{x};
+    my $cury = $old->{y};
+    return if $curx < 0 or $cury < 0;
+    while (1) {
```
Martchus, line 1499: Generally it would be nice to have this whole loop within the backend (to avoid the IPC/serialization overhead for every animation step) but I guess for simplicity we can optimize that later.

testapi.pm
```diff
@@ -1477,6 +1483,42 @@ sub mouse_set ($mx, $my) {
...
+        }
+        if (abs($diffy) < $pixels) {
+            $cury = $y;
+        }
+        else {
+            $cury += $diffy < 0 ? -$pixels : +$pixels;
+        }
+        mouse_set $curx, $cury;
+        last if $curx == $x and $cury == $y;
+        sleep 0.01;
```
Martchus, line 1516: This could be a function parameter:
```suggestion
        sleep $delay_between_steps;
```

Martchus (changes requested):

testapi.pm
perlpunk: Thanks, that works :)

testapi.pm
perlpunk: yep

testapi.pm
perlpunk: Yeah, I mentioned this in the PR description.

testapi.pm
```diff
@@ -507,7 +507,13 @@ sub click_lastmatch (%args) {
     # Calculate the absolute click point.
     my ($x, $y) = _calculate_clickpoint($last_matched_needle, $relevant_area, $relative_click_point);
     bmwqemu::diag("clicking at $x/$y");
-    mouse_set($x, $y);
+    if ($bmwqemu::vars{DEMO_MODE}) {
```
okurz, line 510: the variable "DEMO_MODE" does a similar thing to the variable "NICEVIDEO" in osado, see https://github.com/search?q=repo%3Aos-autoinst%2Fos-autoinst-distri-opensuse%20NICEVIDEO&type=code

PR #2275, opened 2023-03-03
Stream run_cmd output in generalhw backend

asdil12:
This allows the user to observe the output of longer running
scripts in the live log.

### DIFF
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -45,21 +45,13 @@ sub run_cmd ($self, $cmd, @extra_args) {
 
     push @full_cmd, @extra_args;
 
-    my ($stdin, $stdout, $stderr, $ret);
-
-    {
-        # Do not let the SIGCHLD handler of Mojo::IOLoop::ReadWriteProcess::Session steal the exit code from IPC::Run
-        local $SIG{CHLD};
-        eval { $ret = IPC::Run::run(\@full_cmd, \$stdin, \$stdout, \$stderr) };
-        die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
-    }
-    chomp $stdout;
-    chomp $stderr;
-
-    die "$cmd: stdout: $stdout, stderr: $stderr" unless $ret;
-    bmwqemu::diag("$cmd: stdout: $stdout, stderr: $stderr");
-    return $stdout;
+    bmwqemu::diag("Calling $cmd");
+    my $ret = _system(@full_cmd);
+    die "Failed to run command '@full_cmd' (deduced from test variable $cmd): $ret\n" if ($ret != 0);
 }
+
+# wrapper to be mocked in os-autoinst unit tests as it is hard to mock system()
+sub _system (@cmd) { system(@cmd) }    # uncoverable statement
 
 sub poweroff_host ($self) {
     $self->run_cmd('GENERAL_HW_POWEROFF_CMD');
--- a/t/29-backend-generalhw.t
+++ b/t/29-backend-generalhw.t
@@ -45,16 +45,13 @@ $bmwqemu::vars{HDDSIZEGB_1} = 5;
 my $distri = $testapi::distri = distribution->new;
 my $backend = backend::generalhw->new;
 
-# mock IPC::Run and the VNC console
-my $ipc_run_mock = Test::MockModule->new('IPC::Run');
-my $fake_ipc_error;
-$ipc_run_mock->redefine(run => sub {
-        my ($args, $stdin, $stdout, $stderr) = @_;
-        die $fake_ipc_error if $fake_ipc_error;
+# mock backend::generalhw::_system() and the VNC console
+my $cmd_mock = Test::MockModule->new('backend::generalhw');
+my $fake_system_return = 0;
+$cmd_mock->redefine(_system => sub {
+        my $args = \@_;
         push @invoked_cmds, $args;
-        $$stdin = 'stdin';
-        $$stdout = 'stdout';
-        $$stderr = 'stderr';
+        return $fake_system_return;
 });
 my $serial_mock = Test::MockModule->new('backend::generalhw');
 $serial_mock->redefine(start_serial_grab => sub { push @invoked_cmds, 'start_serial_grab' });
@@ -111,11 +108,11 @@ subtest 'stop VM' => sub {
 };
 
 subtest 'error handling' => sub {
-    $fake_ipc_error = 'fake error';
+    $fake_system_return = -1;
     throws_ok(
         sub { $backend->run_cmd('GENERAL_HW_POWEROFF_CMD') },
-        qr/Unable to run command '$cmd_ctl poweroff' \(deduced from test variable GENERAL_HW_POWEROFF_CMD\): fake error/,
-        'IPC error thrown with context'
+        qr/Failed to run command '$cmd_ctl poweroff' \(deduced from test variable GENERAL_HW_POWEROFF_CMD\): -1/,
+        'CMD error thrown with context'
     );
     $bmwqemu::vars{GENERAL_HW_CMD_DIR} = 'does-not-exist';
     throws_ok(


backend/generalhw.pm
```diff
@@ -45,13 +45,22 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
         # Do not let the SIGCHLD handler of Mojo::IOLoop::ReadWriteProcess::Session steal the exit code from IPC::Run
         local $SIG{CHLD};
-        eval { $ret = IPC::Run::run(\@full_cmd, \$stdin, \$stdout, \$stderr) };
-        die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
+        $p = IPC::Run::start(\@full_cmd, '&>', \$out);
+    }
+    while (1) {
+        # stream the cmd output to the live log
+        last unless $p->pumpable();
+        $p->pump() if $p->pumpable();
```
kalikiana, line 58: Why not make it `while ($p->pumpable)`?

backend/generalhw.pm
```diff
@@ -45,13 +45,21 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
-
+    bmwqemu::diag("Calling $cmd");
+    my ($p, $out);
     {
         # Do not let the SIGCHLD handler of Mojo::IOLoop::ReadWriteProcess::Session steal the exit code from IPC::Run
         local $SIG{CHLD};
-        eval { $ret = IPC::Run::run(\@full_cmd, \$stdin, \$stdout, \$stderr) };
-        die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
+        $p = IPC::Run::start(\@full_cmd, '&>', \$out);
+    }
```
perlpunk, line 54:
```suggestion
```

backend/generalhw.pm
```diff
@@ -45,13 +45,21 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
+    while ($p->pumpable) {
+        # stream the cmd output to the live log
+        $p->pump();
+        print($out);
+        $out = '';
+    }
+    die "Failed to run command '@full_cmd' (deduced from test variable $cmd): " . $p->resul() . "\n" unless $p->finish();
+}
```
perlpunk, line 62:
```suggestion
```

backend/generalhw.pm
```diff
@@ -45,20 +45,20 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
-        die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
+        $p = IPC::Run::start(\@full_cmd, '&>', \$out);
     }
-    chomp $stdout;
-    chomp $stderr;
-
-    die "$cmd: stdout: $stdout, stderr: $stderr" unless $ret;
-    bmwqemu::diag("$cmd: stdout: $stdout, stderr: $stderr");
-    return $stdout;
+    while ($p->pumpable) {
```
perlpunk, line 55: The whole while loop should be moved into the block above, that's what I meant

backend/generalhw.pm
```diff
@@ -45,20 +45,20 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
+        # stream the cmd output to the live log
+        $p->pump();
+        print($out);
+        $out = '';
+    }
```
Martchus, line 60: After reading the documentation of IPC::Run I *suppose* this works and is *not* a spinning loop. However, I'm still wondering whether just invoking `system` (with proper error handling) would just do the trick here as well. After all, all this does is immediately forwarding the output as-is. This seems as good as just invoking the process via `system`.

backend/generalhw.pm
```diff
@@ -45,20 +45,20 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
-    my ($stdin, $stdout, $stderr, $ret);
-
+    bmwqemu::diag("Calling $cmd");
+    my ($p, $out);
     {
         # Do not let the SIGCHLD handler of Mojo::IOLoop::ReadWriteProcess::Session steal the exit code from IPC::Run
         local $SIG{CHLD};
-        eval { $ret = IPC::Run::run(\@full_cmd, \$stdin, \$stdout, \$stderr) };
-        die "Unable to run command '@full_cmd' (deduced from test variable $cmd): $@\n" if $@;
+        $p = IPC::Run::start(\@full_cmd, '&>', \$out);
```
perlpunk, line 53: Wouldn't it still make sense to distinguish between stdout and stderr?

asdil12: Then I might interfere with ordering of output.
So I removed decided not to distinguish.

backend/generalhw.pm
asdil12: I would be fine with that too.

backend/generalhw.pm
Martchus: By just using `system` you'd retain that distinction without any extra effort.

asdil12: ah - you mean in regards to what os-autoinst writes to which output FD, yes.
But I hope that everything will hit the autoinst log and also in the right order.

backend/generalhw.pm
```diff
@@ -45,20 +45,10 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
-    }
-    chomp $stdout;
-    chomp $stderr;
-
-    die "$cmd: stdout: $stdout, stderr: $stderr" unless $ret;
-    bmwqemu::diag("$cmd: stdout: $stdout, stderr: $stderr");
-    return $stdout;
+    bmwqemu::diag("Calling $cmd");
+    my $ret = system(@full_cmd);
+    bmwqemu::diag("Command $cmd returned " . ($ret >> 8));
```
Martchus, line 50: This would print nonsense like "Command … returned 72057594037927935" if `system` returns -1.

backend/generalhw.pm
```diff
@@ -45,20 +45,10 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
+    die "Failed to run command '@full_cmd' (deduced from test variable $cmd).\n" if ($ret != 0);
```
Martchus, line 51: It would be nice to print the error message from `$?` (and avoid the problem mentioned in my other comment):
```suggestion
system(@args) == 0
  or die "Failed to run command '@full_cmd' (deduced from test variable $cmd): $?\n";
```

Martchus (changes requested):

backend/generalhw.pm
perlpunk: > to print the error message from `$?`

`$?` doesn't contain an error message, only the exit code.

asdil12: I switched to the return code var directly.

Martchus (changes requested):

backend/generalhw.pm
perlpunk: I meant that before stderr and stdout was collected in two different variables and printed seperately.
By using `system` it's not clear anymore what is stdout and what stderr.
So I don't understand how it would retain that distinction.
Not that I have a strong opinion, I just would like to understand.

backend/generalhw.pm
Martchus: Which leads to the same behavior as my suggestion (but as I said it should be ok).

backend/generalhw.pm
```diff
@@ -45,20 +45,15 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
+    die "Failed to run command '@full_cmd' (deduced from test variable $cmd): $ret\n" if ($ret != 0);
+}
 
-    die "$cmd: stdout: $stdout, stderr: $stderr" unless $ret;
-    bmwqemu::diag("$cmd: stdout: $stdout, stderr: $stderr");
-    return $stdout;
+sub _system {
+    # wrapper to be mocked in os-autoinst unit tests
+    # as it is hard to mock system()
+    return system($@);
```
perlpunk, line 56:
```suggestion
    return system(@_);
```
But better would be a signature:
```
sub _system (@cmd) {
    # wrapper to be mocked in os-autoinst unit tests
    # as it is hard to mock system()
    return system(@cmd);
}
```

backend/generalhw.pm
asdil12: I guess Marius meant that system redirects stdout to parent process stdout and child stderr to parent stderr.
Which should here both end up in the log anyway.

backend/generalhw.pm
```diff
@@ -45,20 +45,15 @@ sub run_cmd ($self, $cmd, @extra_args) {
...
-    chomp $stderr;
+    bmwqemu::diag("Calling $cmd");
+    my $ret = _system(@full_cmd);
+    die "Failed to run command '@full_cmd' (deduced from test variable $cmd): $ret\n" if ($ret != 0);
+}
 
-    die "$cmd: stdout: $stdout, stderr: $stderr" unless $ret;
-    bmwqemu::diag("$cmd: stdout: $stdout, stderr: $stderr");
-    return $stdout;
+sub _system (@cmd) {
```
Martchus, line 53: You'll need to mark the function as non-coverable. Maybe this works best if you make it a one-liner.

Martchus (approved):

perlpunk (approved):

PR #2276, opened 2023-03-05
Implement OCR based on reference string

clanig:
~~**This is a draft!**~~

The previous implementation of the OCR based matching ~~did not compare recognized text of the image area with a reference string~~ (Edit: apparently was removed (partially) a long time ago). This PR replaces the Tesseract implementation with an implementation based on the program GOCR. With this commit, openQA does not require OCR needles to have images provided along with the JSON files. Instead, the reference string can be delivered in the JSON. If the strings do not match, the Levenshtein distance is used to calculate the match percentage.
Implicitly, this commit fixes a bug that prevented running openQA if JSON files without images were provided.

Edit: **Important:** This PR does fix character encoding bugs. It can potentially lead to side effects that make the character encoding look wrong in another place.

What is done?
- [x] Provide purely text based needles as JSON file
- [x] Recognize single line string
- [x] Recognize multi line string
- [x] Report match percentage
- [x] Replace Tesseract unit test
- [x] Check if matching can be improved
- [x] Fix exclude areas for OCR (partially fixed)
- [x] Check if multiple OCR areas work
- [x] Check if mixed needles work
- [x] Check if evil image can harm openQA through OCR
- [x] Ensure dependencies are in Tumbleweed / Take care of dependencies

Dependencies (Packages):
- ~~tesseract-ocr~~
- gocr
- netpbm
- perl-Text-Levenshtein (Repo: https://download.opensuse.org/repositories/devel:/languages:/perl/openSUSE_Tumbleweed/)

Debug log output of POC:
https://gist.github.com/clanig/45f364925993880dae2951015991a3ba

Issues:
- The OCR does often mix up 0 with o. This can be circumvented by reducing the match percentage to e.g. 85%. It seems that GOCR should be improved to fix this.
- When OCR fails to recognize a character, it messes up the output. ~As a result, the Levenshtein distance gets very high. It seems like this can only be fixed in GOCR itself.~ This was due to encoding bugs in front- and backend. (fixed)
- Without patches to the frontend you can't create these needles easily. (fixed in frontend PR)
- Exclude areas: Implemented are: Excluded left/right/top/bottom full edges of areas, as well as dividing exclude areas going through the OCR area horizontally and vertically. Additionally, full exclusion of the match area is supported. All this is done by geometric area refinement/splitting. Unfortunately, for all other exclude areas we would need to resynthesize the image background in the excluded area. For this reason, it is not implemented. We can overcome these issues by giving direct feedback at needle creation what is recognized by OCR in the front end.

How to test:
1. Checkout my branch
2. Point the openQA worker to the isotovideo of the branch: `--isotovideo=<PATH_OF_os-autoinst>/isotovideo`
3. Run tests on that worker

As a POC a test for The battle for Wesnoth containing more than 15 OCR needles was created. Most needles match by 100%.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16539

POO issue: https://progress.opensuse.org/issues/121354


ppmclibs/tinycv.pm
```diff
@@ -22,6 +22,7 @@ bootstrap tinycv $VERSION;
 package tinycv::Image;
 
 use Mojo::Base -strict, -signatures;
+use Text::Levenshtein qw(distance);
```
Martchus, line 25: If you introduce a new dependency, you need to update dependencies.yaml and regenerate the spec/container files.

Martchus (commented):
I haven't reviewed this before because this change doesn't seem to be ready for review. I guess this is still true but I've nevertheless just added two remarks.

I suppose it would be important to add test cases for this feature.

ppmclibs/tinycv.pm
```diff
@@ -75,46 +81,74 @@ sub search_ ($self, $needle, $threshold, $search_ratio, $stopwatch = undef) {
...
-            x => $xmatch,
-            y => $ymatch,
-            w => $area->{width},
-            h => $area->{height},
-            result => 'ok',
-        };
-        if (my $click_point = $area->{click_point}) {
-            $ma->{click_point} = $click_point;
+    if (@ocr) {
+        for my $area (@ocr) {
```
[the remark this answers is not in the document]
clanig: Thanks for the comment. I would like to update this PR today. Though, indeed, this is still in the current code and I will check quickly if I can remove it before pushing the new code.

ppmclibs/tinycv.pm
clanig: Thanks. I will make sure to do this. However, regarding this particular dependency we might need to get it in the Tumbleweed repos before we can merge this PR.

clanig:
I have updated the code to the current state. It is still based on an old master checkout though.
I am going to rebase it.

clanig:
I have rebased everything. I will take care of the remaining two tasks next.

ocr.pm
```diff
@@ -1,31 +1,88 @@
...
 use Mojo::Base -strict, -signatures;
 require IPC::System::Simple;
+use autodie ':all';
+use File::Temp 'tempdir';
+use Scalar::Util 'blessed';
+use bmwqemu 'fctwarn';
+
+=head2 img_to_str
+
+Abstracts the area of an image to a character string and returns the result.
```
Martchus, line 23: I'm not sure whether "Abstracts" is the right word here.

ocr.pm
```diff
@@ -1,31 +1,88 @@
...
+    # -a defines the accuracy threshold per character in percent. GOCR's default
+    # is 95.
+    chomp($ret = `gocr -C '--_.:,;()[]{}<>0-9a-zA-Z~+*^@#\$\%&/=?!' -u § -a 95 $img_fpath 2> $stderr_fpath`);
 
-    if ($area) {
-        $img = $img->copyrect($area->{xpos}, $area->{ypos}, $area->{width}, $area->{height});
+    if ($?) {
+        my $stderr_fh;
+        unless (open($stderr_fh, "<$stderr_fpath")) {
+            fctwarn("Error during OCR occurred but could not read stderr output!");
```
Martchus, line 73:
```suggestion
            fctwarn('Error during OCR occurred and stderr could not be read');
```

t/01-test_needle.t
```diff
@@ -94,27 +94,6 @@ $needle = needle->new('kde.ref.json');
 $res = $img1->search($needle);
 ok(!defined $res, "no match with different art");
 
-subtest 'handle failure to load image' => sub {
```
Martchus, line 97: Removing this whole subtest looks wrong.

Martchus (changes requested):

t/39-tinycv.t
```diff
@@ -0,0 +1,268 @@
...
+    $olvl = tinycv::Image::overlap_lvl($area1, {xpos => 90, ypos => 140, height => 20, width => 20});
+    ok($olvl eq 14, "Check 14 for partial overlap through left edge.");
+
+    undef($olvl);
+    $olvl = tinycv::Image::overlap_lvl($area1, {xpos => 190, ypos => 140, height => 20, width => 20});
+    ok($olvl eq 15, "Check 15 for partial overlap through right edge.");
+};
+
+use needle;
+use OpenQA::Benchmark::Stopwatch;
```
Martchus, line 128: Please put all `use` statements at the top of the file.

clanig:
I have pushed the current (rebased) state. From my perspective the PR is finished with regard to the code.

ocr.pm
clanig: Thank you for the comment. I have rephrased it in the current version.

ocr.pm
```diff
@@ -1,31 +1,88 @@
...
+itself can't be recognized. Keep that in mind for comparison!
+Returns undef on failure.
+
+=cut
+
+sub img_to_str ($img) {
+
+    if (blessed($img)) {
+        unless ($img->isa('tinycv::Image')) {
+            fctwarn("Not a reference to object with type tinycv::Image!");
```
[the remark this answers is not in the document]
clanig: Thank you for the hint. I have tried to fix this everywhere.

ocr.pm
clanig: Thanks. I have changed this accordingly.

t/01-test_needle.t
clanig: Thanks for this hint! I was able to keep most parts.

t/02-test_ocr.t
```diff
@@ -11,44 +11,40 @@ use Test::Warnings ':report_warnings';
...
-
-my $ocr;
-for my $area (@{$res->{needle}->{area}}) {
-    next unless $area->{type} eq 'ocr';
-    $ocr .= ocr::tesseract($img1, $area);
-}
-
-ok defined $ocr, 'OCR area found' and
-  ok($ocr =~ /Memory Test.*Video Mode/s, 'multiple OCR regions');
+fail("OCR: is gocr installed") unless (which('gocr'));
```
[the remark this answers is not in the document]
clanig: Thanks. I have solved this implementing a SKIP block. I hope you are OK with that solution as well.

t/39-tinycv.t
clanig: I have solved that (hopefully).

ocr.pm
```diff
@@ -1,31 +1,121 @@
...
+
+Character string of characters '-_.:,;()[]{}<>0-9a-zA-Z~+*^@#$%&/=?!§',
+where '§' is placed for each unrecognized character. Implicitly, '§'
+itself can't be recognized. Keep that in mind for comparison!
+Returns undef on failure.
+
+=cut
+
+sub img_to_str ($img) {
+    return unless $OCR_INSTALLED;
```
Martchus, line 52: In new code I suggest to use `return undef` (unless you really want this to return an empty list when executed in list-context):

```suggestion
    return undef unless $OCR_INSTALLED;
```

ocr.pm
```diff
@@ -1,31 +1,121 @@
...
+    if (blessed($img)) {
+        unless ($img->isa('tinycv::Image')) {
+            fctwarn('Not a reference to object with type tinycv::Image!');
+            return;
+        }
+    }
+    else {
+        fctwarn('Not a reference to object with type tinycv::Image!');
+        return;
+    }
```
Martchus, line 62: Those two cases can actually be shrunk into one:
```suggestion
    if (!blessed($img) || !$img->isa('tinycv::Image')) {
        fctwarn('Not a reference to object with type tinycv::Image (internal error, probably a bug)!');
        return undef;
    }
```

This error message is likely still useless to a user. I'm not sure whether it can be improved, though.

Martchus (changes requested):
I guess not all my points (see inline comments) are very important.

Do you have some example job? After reading all the code I'm still wondering how well this works.

container/os-autoinst_dev/Dockerfile
```diff
@@ -129,6 +128,7 @@ RUN zypper in -y -C \
        'perl(Test::Pod)' \
        'perl(Test::Strict)' \
        'perl(Test::Warnings)' \
+       'perl(Text::Levenshtein)' \
```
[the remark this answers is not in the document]
Martchus: Unfortunately the SR has been declined. Maybe @perlpunk can help with that.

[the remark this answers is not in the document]
clanig: Oops, I thought that SR would be created differently. Looks like I could not handle it myself then.
It would be great if @perlpunk could help with it.

ocr.pm
```diff
@@ -1,31 +1,121 @@
...
+        my $stderr_fh;
+        unless (open($stderr_fh, "<$stderr_fpath")) {
+            fctwarn('Error during OCR occurred and stderr output could not be read!');
+            return;
+        }
+        my @stderr_lines = <$stderr_fh>;
+        fctwarn('Error occurred during OCR');
+        for my $line (@stderr_lines) {
+            fctwarn("$line");
+        }
```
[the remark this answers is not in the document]
clanig: In this case, only a critical error in `gocr` would cause a warning. And this should never happen, since it just takes an image and creates a string of it.
I don't have the slightest clue what `gocr` could write to STDERR, but to be safe, I have created this functionality in order to log an error and not let it be hidden and disappear in the case it happens (if that would ever be the case is a different question).
As far as I remember I have only extended the debug output for the case that a needle match was found to print the OCR string aside from the match percentage as done in case of images. From my perspective, the current could should create a warning only if the user has an insufficiency in the setup (e.g. no OCR setup but trying to match an OCR needle), or trying to match illegally sized images or caused a programming bug somewhere else invoking with illegal parameter values. I am going to change the lines as you suggested.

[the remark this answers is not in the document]
clanig: I have restructured this section. I hope it is OK now.

ocr.pm
clanig: I have implemented the suggestion.

ocr.pm
clanig: Thanks for the hint. I have tried to find all returns and change them.

t/39-tinycv.t
```diff
@@ -0,0 +1,270 @@
...
+    $needle = needle->new("bootmenu.test.img-nopng.json");
+    stderr_like(
+        sub { is($img->search_($needle, 0.0, 0.0), undef, 'Check if image based needle without png returns undef.'); },
+        qr/skipping .*: missing PNG./,
+        'Check stderr for img based needle without png file.');
+
+    my $ret;
+    my $levenshtein_available = eval { require Text::Levenshtein; 1 } || 0;
+  SKIP: {
+        skip('Either the OCR program or the Levenshtein module are not installed.', 1)
```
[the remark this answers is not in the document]
clanig: I have done this change as suggested.

ocr.pm
```diff
@@ -1,31 +1,121 @@
...
+    }
+    else {
+        fctwarn('Not a reference to object with type tinycv::Image!');
+        return;
+    }
+
+    # Better be safe than sorry. C.f. CVE-2005-1141
+    unless ($img->xres() <= 1024 && $img->yres() <= 768) {
+        fctwarn('The image is too big for OCR. Returning undef...');
```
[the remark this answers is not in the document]
clanig: Thanks. I have changed the message accordingly.

container/os-autoinst_dev/Dockerfile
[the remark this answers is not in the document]
perlpunk: https://build.opensuse.org/request/show/1112337

clanig:
> Do you have some example job? After reading all the code I'm still wondering how well this works.

I can show a screenshot of the Wesnoth test. See below. But this one is made including the frontend changes.
Unfortunately I do not have a dev instance of openQA currently exposed for external access.
I have chosen a screenshot that contains an OCR error, and you can expect to see `§` (which stands for the error) sometimes, but most of the times the OCR works just fine.
However, since it sometimes does recognize two characters as one, and doesn't find out what characters they are, this ends up in a significant Levenshtein distance and as a result, the OCR can't be used for few short words. It could be improved if gocr would become more precise though. The benefit is that it doesn't need to train anything, so no need for any pictures at all. It is just ready to go for almost all cases.

Screenshot from 2023-09-19 18-05-25

container/os-autoinst_dev/Dockerfile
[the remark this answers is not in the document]
clanig: Thanks a lot!

clanig:
netpbm is missing in the test pipeline, which - if I am not wrong - should perhaps be a requirement for gocr. I am going to check if the package is wrong or if another dependency needs to be added to os-autoinst.

clanig:
I am going to fix the coverage quickly.

clanig:
I can't fix the coverage because the coverage tool does not accept the whole branch to be uncoverable using `uncoverable branch true` in `ocr.pm`. This is contrary to the documentation of `Devel::Cover`.

okurz:
> I can't fix the coverage because the coverage tool does not accept the whole branch to be uncoverable using `uncoverable branch true` in `ocr.pm`. This is contrary to the documentation of `Devel::Cover`.

It's actually different. The command `uncoverable branch` only marks the statements as not coverable for branch coverage, not statement coverage. For that you would need `uncoverable statement` on every single line with a compete statement

clanig:
OK. Finally, the whole pipeline is green. There are three lines uncovered. They were tested manually and but covering these would require significant extra effort without any real benefit.

t/39-tinycv.t
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/perl
+# Checks units provided by tinycv.pm module.
+# TODO: This test does not cover all provided functions yet.
```
okurz, line 3: I'd say it's ok to not have all included. Please don't include "TODOs" in the code. Those tend to stay forever :) If you want to remember ideas where to improve have a ticket.

t/39-tinycv.t
```diff
@@ -0,0 +1,270 @@
...
+    $olvl = tinycv::Image::overlap_lvl($area1, {xpos => 140, ypos => 190, height => 20, width => 20});
+    ok($olvl eq 13, 'Check 14 for partial overlap through bottom edge.');
+
+    undef($olvl);
+    $olvl = tinycv::Image::overlap_lvl($area1, {xpos => 90, ypos => 140, height => 20, width => 20});
+    ok($olvl eq 14, 'Check 14 for partial overlap through left edge.');
+
+    undef($olvl);
+    $olvl = tinycv::Image::overlap_lvl($area1, {xpos => 190, ypos => 140, height => 20, width => 20});
+    ok($olvl eq 15, 'Check 15 for partial overlap through right edge.');
```
okurz, line 124: That looks rather verbose. Is it really necessary to test each and every variant in all cases? I suggest to focus on the important "corner" cases

testapi.pm
```diff
@@ -252,8 +252,9 @@ sub _handle_found_needle ($foundneedle, $rsp, $tags) {
     my $frame = $rsp->{frame};
     $autotest::current_test->record_screenmatch($img, $foundneedle, $tags, $rsp->{candidates}, $frame);
     my $lastarea = $foundneedle->{area}->[-1];
-    bmwqemu::fctres(
-        sprintf("found %s, similarity %.2f @ %d/%d", $foundneedle->{needle}->{name}, $lastarea->{similarity}, $lastarea->{x} // 0, $lastarea->{y} // 0));
+    my $debug_info = sprintf("found %s, similarity %.2f @ %d/%d", $foundneedle->{needle}->{name}, $lastarea->{similarity}, $lastarea->{x} // 0, $lastarea->{y} // 0);
+    $debug_info .= " OCR: $lastarea->{ocr_str}" if defined $lastarea->{ocr_str};
+    bmwqemu::fctres($debug_info);
```
okurz, line 257: what codecov says. This needs to be covered :)

t/02-test_ocr.t
```diff
@@ -8,47 +8,124 @@ use FindBin '$Bin';
...
+            'Check stderr for exceeding vertical pixel count');
+
+        $img = tinycv::read(dirname(__FILE__) . '/data' . '/bootmenu.test.png');
+        $xpos = 256;
+        $ypos = 254;
+        $width = 194;
+        $height = 17;
+        $img_area = $img->copyrect($xpos, $ypos, $width, $height);
+        $ocr_str = img_to_str($img_area);
+        diag("Found OCR String:\n$ocr_str");
```
okurz, line 116: Please make sure that in non-verbose mode the tests do not output anything, see https://github.com/os-autoinst/os-autoinst/actions/runs/6347763302/job/17243248731?pr=2276#step:3:635

You can use `note` here. And for all other output I suggest to use Test::Output as you already do in other cases

okurz (changes requested):

t/39-tinycv.t
kalikiana: Maybe just make it a hash with different cases. That would simplifiy it anyway

clanig: Thank you for the suggestion. I have created a Hash mapping all input parameters to the corresponding output value and test case description.

testapi.pm
clanig: Thanks for your opinion. I have implemented the test case.

clanig:
I have updated the code. It should be covering the last three lines as well. I can't reproduce the issue in the test pipeline. From my perspective it might be a sporadic bug of the pipeline, so a rerun might fix it.

okurz (changes requested):
Please look and fix the CI issues. There are style issues related to your changes. In general os-autoinst tests are very stable so while unrelated sporadic test issues are possible they are unlikely. I will look into the code changes after you could at least fix the related style issues

clanig:
Thanks for the info. Previously I have not noticed that aside from `29-backend-generalhw.t` having failed, there were style issues in `39-tinycv.t` as well. I have fixed them so the CI is green now.

t/03-testapi.t
```diff
@@ -118,6 +119,36 @@ my $mod2 = Test::MockModule->new('testapi');
 
 my $mock_bmwqemu = Test::MockModule->new('bmwqemu');
 
+subtest '_handle_found_needle' => sub {
+    my $mocked_needle = Test::MockModule->new('needle');
+    $mocked_needle->mock('new', sub { return bless({name => 'needlename'}) });
+    my $mocked_encode = Test::MockModule->new('Encode');
+    $mocked_encode->mock('decode_base64', sub { return 'A value' });
```
Martchus, line 126:
```suggestion
    $mocked_encode->mock(decode_base64 => 'A value');
```

t/03-testapi.t
```diff
@@ -118,6 +119,36 @@ my $mod2 = Test::MockModule->new('testapi');
...
+    my $mocked_tinycv = Test::MockModule->new('tinycv');
+    $mocked_tinycv->mock('from_ppm', sub { return 'A Value' });
```
Martchus, line 128:
```suggestion
    $mocked_tinycv->mock(from_ppm => 'A Value');
```
The same counts for further occurrences. We usually avoid quotes and you can just pass the scalar value as 2nd argument instead of a function returning it.

t/03-testapi.t
```diff
@@ -118,6 +119,36 @@ my $mod2 = Test::MockModule->new('testapi');
...
+    my $last_matched_needle_sav = $testapi::last_matched_needle;
+    stderr_like(
+        sub { testapi::_handle_found_needle($foundneedle, $fakeresponse, undef); },
+        qr;.*testapi::_handle_found_needle: found needlename, similarity 1.00 @ 0/0 OCR: ocr.*;,
+        'Check correct output for ocr needle');
+
+    $mocked_needle->unmock_all();
+    $mocked_encode->unmock_all();
+    $mocked_tinycv->unmock_all();
+    undef($mocked_current_test);
```
Martchus, line 147: All of this shouldn't be necessary because the variables go out of scope at the end of the block anyways.

t/39-tinycv.t
```diff
@@ -0,0 +1,372 @@
...
+    );
+    my $area1 = {xpos => 100, ypos => 100, width => 100, height => 100};
+
+    foreach my $test_case (@test_case_list) {
+        ok(
+            tinycv::Image::overlap_lvl($area1, $test_case->{area}) eq
+              $test_case->{retval},
+            $test_case->{descr}
+        );
+    }
```
Martchus, line 162:
```suggestion
    is tinycv::Image::overlap_lvl($area1, $_->{area}), $_->{retval}, $_->{descr}
        for @test_case_list;
```

Martchus (changes requested):

bmwqemu.pm
```diff
@@ -272,7 +266,7 @@ sub mydie ($cause_of_death) {
 # store the obj as json into the given filename
 sub save_json_file ($result, $fn) {
     open(my $fd, ">", "$fn.new");
-    my $json = eval { Cpanel::JSON::XS->new->pretty->canonical->encode($result) };
+    my $json = eval { Cpanel::JSON::XS->new->utf8->pretty->canonical->encode($result) };
```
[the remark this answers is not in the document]
okurz: I feel like this deserves its own PR. Also to potentially identify issues more easily and revert. @clanig can you please put this commit into its own PR which we can merge first?

okurz (changes requested):

bmwqemu.pm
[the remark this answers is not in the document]
clanig: Sure, makes sense:
https://github.com/os-autoinst/os-autoinst/pull/2397

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+    }
+
+    $need_ref_img = 0;
+    for my $area (@{$needle->{area}}) {
+        push(@exclude, $area) if $area->{type} eq 'exclude';
+        if ($area->{type} eq 'match') {
+            push(@match, $area);
+            $need_ref_img = 1;
+        }
+        push(@ocr, $area) if $area->{type} eq 'ocr';
```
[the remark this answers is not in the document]
clanig: Thanks for the comment. I have changed it.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+    my $a1top = $area1->{ypos};
+    my $a1bot = $area1->{ypos} + $area1->{height};
+    my $a1left = $area1->{xpos};
+    my $a1right = $area1->{xpos} + $area1->{width};
+    my $a2top = $area2->{ypos};
+    my $a2bot = $area2->{ypos} + $area2->{height};
+    my $a2left = $area2->{xpos};
+    my $a2right = $area2->{xpos} + $area2->{width};
+
+    return 1 if ($a2bot >= $a1bot) and ($a2top <= $a1top)
```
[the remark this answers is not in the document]
clanig: Done.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+
+=item *
+
+i in [0, 15] | i is level of overlap of area2 over area1 where:
+
+=over
+
+=item *
+
+i:= 0 for area2 does not overlap area1
```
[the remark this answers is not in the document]
clanig: I have added `:` to make it clearer.

ocr.pm
```diff
@@ -1,31 +1,126 @@
...
-    open(my $fh, '<:encoding(UTF-8)', $txtfn);
-    local $/;
-    $txt = <$fh>;
-    close $fh;
-    unlink $imgfn;
-    unlink $txtfn;
-    return $txt;
+    # Better be safe than sorry. C.f. CVE-2005-1141
+    unless ($img->xres() <= $MAX_PIX_HORIZONTAL && $img->yres() <= $MAX_PIX_VERTICAL) {
+        fctwarn("Skipping OCR evaluation of image, resolution too high(max. " .
```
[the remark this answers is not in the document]
clanig: I have changed it. Thanks.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
         }
-        $stopwatch->lap("**++ search__: areas exclusion") if $stopwatch;
+        $stopwatch->lap('**++ search__: areas exclusion') if $stopwatch;
     }
+
     my $ret = {ok => 1, needle => $needle, area => []};
+
+    for my $area (@ocr) {
+
+        unless ($LEVENSHTEIN && ocr_installed()) {
```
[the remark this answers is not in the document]
clanig: I have moved it upward. Thanks for the great finding!

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+            fctwarn("skipping $needle->{name}: Perl Levenshtein module or OCR program are not installed.");
```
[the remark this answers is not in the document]
clanig: This is changed accordingly.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+            return undef;
+        }
+        my $refstr = $area->{refstr};
+
+        if ($refstr =~ qr/§/) {
+            fctwarn("skipping $needle->{name}: Illegal placeholder character '\§' in refstr");
```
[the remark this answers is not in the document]
clanig: This is changed accordingly.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+            return undef;
+        }
+        my $sim = 1;
+        my @partareas = ({
+                xpos => $area->{xpos},
+                ypos => $area->{ypos},
+                width => $area->{width},
+                height => $area->{height}
+        });
```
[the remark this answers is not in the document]
clanig: You are right, thanks! It is implemented.

ppmclibs/tinycv.pm
```diff
@@ -34,47 +47,422 @@ sub mean_square_error ($areas) {
...
+
+                # If you change the percentage of 90%, please check if the
+                # test 39-tinycv.t requires changes as well.
+                my $m = ($area->{match} || 90) / 100;
+                if ($sim < $m - $threshold) {
+                    $ma->{result} = 'fail';
+                    $ret->{ok} = 0;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
clanig: Thanks for the hint, it is changed.

t/03-testapi.t
clanig: Thanks. It is done.

t/03-testapi.t
clanig: Thanks. It is done.

t/39-tinycv.t
clanig: Thanks. I have changed it as you suggested. Saves some space.

t/01-test_needle.t
```diff
@@ -168,7 +166,7 @@ needle_init;
 
 my @alltags = sort keys %needle::tags;
 my @needles = @{needle::tags('none') || []};
-is(@needles, 4, "four needles found");
+is(@needles, 21, "21 needles found");
```
okurz, line 169: Where are 21 needles come from now?

t/02-test_ocr.t
```diff
@@ -8,47 +8,124 @@ use FindBin '$Bin';
...
-}
 
-use needle;
-use cv;
-use ocr;
+subtest 'ocr_installed' => sub {
+    my $ocr_installed_sav = $ocr::OCR_INSTALLED;
+    $ocr::OCR_INSTALLED = undef;
+    ok(!ocr_installed(), 'Check retval if OCR program is not installed');
+    $ocr::OCR_INSTALLED = '/usr/bin/ocr_prog';
```
okurz, line 25: I suggest to not rely on this hardcoded path, be flexible please

okurz (changes requested):
I only could point out some minor issues. For a more in depth review I will need to come back to this some time later. Please fix all the open issues first

ocr.pm
```diff
@@ -1,31 +1,126 @@
...
+    my $stderr_fpath = $tmpdir . 'stderr.txt';
+
+    unless ($img->write($img_fpath)) {
+        fctwarn('Writing image to prepare for OCR failed!');
+        return undef;
+    }
+
+    # -a defines the accuracy threshold per character in percent. GOCR's default
+    # is 95.
+    chomp($ret = `gocr -C '--_.:,;()[]{}<>0-9a-zA-Z~+*^@#\$\%&/=?!' -u '§' -a 95 $img_fpath 2> $stderr_fpath`);
```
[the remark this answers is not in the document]
clanig: The explicit declaration of parameters does ensure a consistent behavior of programs independently from changes introduced in futur versions. In addition, it simplifies the maintenance of the existing code by already providing the scheme to tune parameters in the future. Further, making visible a selected amount of internals can visualize how program works to reduce the black box impression to potential maintainers.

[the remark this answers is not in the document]
okurz: hm, not convinced. In my experience, in particular in this project os-autoinst we struggled because too many explicit parameters were specified. I suggest to just trust the upstream making sensible choices about changing parameter default values if any. Would you be ok to just leave out the "-a 95"? Then also nobody questions our choice of the value 95 :)

ocr.pm
```diff
@@ -1,31 +1,126 @@
...
 require IPC::System::Simple;
+use autodie ':all';
+use File::Temp 'tempdir';
+use File::Which;
+use Scalar::Util 'blessed';
+use Exporter 'import';
+use bmwqemu 'fctwarn';
+
+our @EXPORT = qw(img_to_str ocr_installed);
+our $MAX_PIX_HORIZONTAL = 1024;
```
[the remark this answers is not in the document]
clanig: For previous version of GOCR there is a CVSS entry regarding a heap overflow in case of big images. As a result, I have created the limit in case there was an unnoticed a bug that shows similar behavior already in the program or available in the future and to ensure the consistency of the library in case of being reused for different implementations.

t/01-test_needle.t
clanig: The 21 needles are different test cases. The OCR test requires additional ones, for which reason they sum up to 21 instead of the previous four.

needle.pm
```diff
@@ -1,5 +1,5 @@
 # Copyright 2009-2013 Bernhard M. Wiedemann
-# Copyright 2012-2021 SUSE LLC
+# Copyright 2012-2023 SUSE LLC
```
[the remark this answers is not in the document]
clanig: Thanks for the comment. I have fixed it in the latest revision.

ocr.pm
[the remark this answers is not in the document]
clanig: Sure, I can leave it out. Will create another revision.

ocr.pm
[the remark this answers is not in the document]
okurz: I don't see why that should be concern of os-autoinst then. I would leave out that resolution or at least make it overridable

t/02-test_ocr.t
okurz: ok, I see, nevermind

ocr.pm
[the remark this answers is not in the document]
Martchus: I guess at least making it overridable (via test variables) would make sense.

Martchus (approved):

okurz (changes requested):
sorry, I missed to publish some pending review comments

PR #2277, opened 2023-03-06
Rename foo.pm and bar/baz.pm to more specific names

perlpunk:
Also test explicitly that *both* modules have been scheduled, because previously removing bar.pm would not make the test fail, just the error message in the output would have gone unnoticed.


Martchus (approved):

perlpunk:
`codecov/project/fully_covered Successful in 1s — 99.28% (target 100.00%), passed because patch was fully covered by tests with no unexpected coverage changes`
So there's still something wrong

okurz:
yes, I wouldn't have expected this change to magically fix the problem but I see it as an improvement nevertheless. Could you undraft it? I would merge this and continue afterwards

baierjan (approved):

PR #2278, opened 2023-03-07
Fix flaky codecov coverage checks about t/data/tests

okurz:
All relevant code parts in the codecov config section "fully_covered"
are actually 100% statement+branch covered but we now have the check
partially returning a false result.as a test module in t/data/tests is
loaded within t/14-isotovideo.t where the coverage is not reliably
collected from the isotovideo subprocess. But as we actually do not need
to care about the coverage of those trivial test modules we will just
include them from the coverage check in codecov.

Related progress issue: https://progress.opensuse.org/issues/125237


Martchus (approved):

okurz:
This did not work as intended:

```
codecov/project/fully_covered — 99.28% (target 100.00%)
```

I think that we might have reached the 100% *only* when by accident those test modules were actually covered. We could lower the threshold to 99.28% but obviously it wouldn't look nice and just the number does not ensure that we don't loose coverage in other places.

okurz:
https://app.codecov.io/gh/os-autoinst/os-autoinst/pull/2278/indirect-changes looks weird and hardly readable to me. I rebased and pushed again. Looks like the diff to master is too big to make the coverage diff report easy enough to see.

kraih (approved):

PR #2283, opened 2023-03-15
coverage: Ignore via PERL5OPT in addtion to codecov.yaml

kalikiana:
See: https://progress.opensuse.org/issues/125237#note-13


tools/invoke-tests
```diff
@@ -44,7 +44,7 @@ db="$build_directory/cover_db"
 # set Perl module include path and coverage options
 export PERL5LIB="$source_directory:$source_directory/ppmclibs/blib/lib:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
 if [[ $WITH_COVER_OPTIONS ]]; then
-    ignore="external/|t/data/tests/tests/|/tmp|/CheckGitStatus.pm|$prove_path"
+    ignore="external/|t/data/test/|/tmp|/CheckGitStatus.pm|$prove_path"
```
perlpunk, line 47:
```suggestion
    ignore="external/|t/data/tests/|/tmp|/CheckGitStatus.pm|$prove_path"
```

okurz (changes requested):
wait, excluding t/data/tests/tests is IMHO too much. We should actually include t/data/tests/tests/ but only exclude t/data/tests/bar/ and t/data/tests/module1.pm

perlpunk:
t/data/tests/tests was already excluded before

okurz:
> t/data/tests/tests was already excluded before

do you mean in codecov? Yes. But then I wonder why was t/data/tests ever a problem in codecov when it should have been already excluded by Devel::Cover?

okurz:
> > t/data/tests/tests was already excluded before
> 
> do you mean in codecov? Yes. But then I wonder why was t/data/tests ever a problem in codecov when it should have been already excluded by Devel::Cover?

EDIT: Nevermind. I got it know. Then I think we can go even further and revert my latest change in codecov.yml with https://github.com/os-autoinst/os-autoinst/pull/2284 and ignore all data from t/data for the same reason. Please let me know if you think this is a good extension of this PR here or if it goes too far.

PR #2284, opened 2023-03-15
Consolidate ignoring of irrelevant test data for coverage

okurz:
We already excluded t/data/tests/tests due to unreliable code coverage
analysis in test modules and recently added t/data/tests to the
codecov-specific ignore list. Instead we should ignore everything
irrelevant when calling Devel::Cover and not ignore anything in codecov.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -32,8 +32,6 @@ coverage:
           - signalblocker.pm
           - t
           - xt
-        ignore:
-          - t/data/tests
     patch:
       default:
         branches: null
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -44,7 +44,7 @@ db="$build_directory/cover_db"
 # set Perl module include path and coverage options
 export PERL5LIB="$source_directory:$source_directory/ppmclibs/blib/lib:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
 if [[ $WITH_COVER_OPTIONS ]]; then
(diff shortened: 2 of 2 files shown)


Martchus (commented):
This makes generally sense. Not sure why codecov complains, though.

perlpunk (approved):

okurz:
Changed it to ignore t/data/tests instead of complete t/data as I think the coverage of t/data/lib/testdistribution.pm and t/data/main.pm is helpful

okurz:
who can explain again the 99.28% for fully_covered where we expect 100%?

okurz:
https://github.com/os-autoinst/os-autoinst/actions/runs/4430248247/jobs/7771734952#step:3:933 and following reports <100% for t files, how can that happen?

I checked locally with

```
rm -rf cover_db/ && OPENQA_TEST_TIMEOUT_DISABLE=1 tools/invoke-tests --coverage t/14-isotovideo.t && cover -report html_minimal  cover_db
```

and I am getting 99.4% coverage but in cover_db/t-14-isotovideo-t.html all lines are showing up as green. What's the matter?

perlpunk:
Obsoleted by #2292

okurz:
Yeah but I would still prefer if we don't need to ignore anything in codecov when ignoring is already done in Devel::Cover. Do you think your one commit to make the fake files distinct would help here?

perlpunk:
> Yeah but I would still prefer if we don't need to ignore anything in codecov

yes, that's the purpose of #2292

PR #2289, opened 2023-03-30
container: Provide jq for convenience in downstream jobs

okurz:
https://github.com/os-autoinst/os-autoinst-distri-example/actions/runs/4562594649/jobs/8050016153?pr=15
shows how jq is tried to be installed but fails due to problems in the
repositories which we can avoid by providing all dependencies in the
container already.

### DIFF
--- a/container/isotovideo/Dockerfile.qemu-kvm-jq
+++ b/container/isotovideo/Dockerfile.qemu-kvm-jq
@@ -1,0 +1,6 @@
+#!BuildTag: isotovideo:qemu-kvm
+
+FROM opensuse/tumbleweed
+# Provide "jq" as convenience to work with the os-autoinst output
+RUN zypper -n in os-autoinst-qemu-kvm jq && zypper clean
+ENTRYPOINT ["/usr/bin/isotovideo"]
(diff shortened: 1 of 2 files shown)


container/isotovideo/Dockerfile.qemu-kvm
```diff
@@ -1,5 +1,6 @@
 #!BuildTag: isotovideo:qemu-kvm
 
 FROM opensuse/tumbleweed
-RUN zypper -n in os-autoinst-qemu-kvm && zypper clean
+# Provide "jq" as convenience to work with the os-autoinst output
+RUN zypper -n in os-autoinst-qemu-kvm jq && zypper clean
```
kalikiana, line 5: To be honest this seems pretty arbitrary. The clean approach would be to use a dedicated image which includes all tools needed **or** even make it a proper os-autoinst feature if it caters to a common use case.

okurz: ok, updated to use two new explicit container images instead.

Martchus: And I've already wanted to comment why not just putting the tool into the image we already have. I'm not sure what's best but creating a distinct container for each set of tools we ever need seems not like a good idea. I guess since it is just one extra container for now it is ok.

kalikiana (approved):

perlpunk:
@mergifyio rebase

perlpunk:
@Mergifyio rebase

mimi1vx (approved):

PR #2290, opened 2023-03-30
Turn around default of script_run, see #1807

okurz:
The option "die_on_timeout" was introduced as a temporary measure to do
"the right thing" while preserving original behaviour for limited time.
By now we can turn around the default to do the most commonly expected
behaviour by default while still offering the old behaviour as a
workaround although now accompanied with a deprecation message.

Related progress issue: https://progress.opensuse.org/issues/126959

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -923,7 +923,7 @@ sub assert_script_run {    # no:style:signatures
 
 =head2 script_run
 
-  script_run($cmd [, timeout => $timeout] [, output => ''] [, quiet => $quiet] [, die_on_timeout => -1]);
+  script_run($cmd [, timeout => $timeout] [, output => ''] [, quiet => $quiet] [, die_on_timeout => 1]);
 
 Deprecated mode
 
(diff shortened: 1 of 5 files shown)


testapi.pm
```diff
@@ -936,9 +936,10 @@ execution to complete.
 C<$output> can be used as an explanatory text that will be displayed with the execution of
 the command.
 
-With C<die_on_timeout> -1 (default) a warning will be printed to log. To avoid
-this warning, set it explicit to 0. With C<die_on_timeout> equal to 1, this command
-throw an exception, if timeout expires.
+By default C<script_run> will throw an exception if timeout expires. This is
```
Martchus, line 939:
```suggestion
By default C<script_run> will throw an exception if the timeout is exceeded. This is
```
```suggestion
By default C<script_run> will throw an exception if the timeout has expired. This is
```
If you don't like the "the":
```suggestion
By default C<script_run> will throw an exception if C<timeout> has expired. This is
```

testapi.pm
```diff
@@ -962,23 +963,20 @@ sub script_run {    # no:style:signatures
...
-            croak("command '$cmd' timed out") if !defined($ret);
+        if ($die_on_timeout == 0) {
+            # This is to warn users of deprecated behaviour of script_run()
+            my ($package, $filename, $line) = caller;
+            my $casedir = testapi::get_var(CASEDIR => '');
+            $filename =~ s%^\Q$casedir\E/%%;
+            bmwqemu::fctwarn("DEPRECATED call of script_run() in $filename:$line " .
+                  'requested by `die_on_timeout => 0` or set
+                  $distri->{script_run_die_on_timeout}. Adapt the test code to work
+                  with the default. This workaround will be removed in the near future');
```
Martchus, line 977:
```suggestion
                  with the default. This workaround is going to be removed in the near future.');
```
I guess it doesn't really matter much, though - and one generally uses the will future to make threats :-)

okurz: yes. I wanted to make that threat :) I think the message sounds more likely to happen with "will be removed" so that people would really act. I would prefer to keep that.

testapi.pm
okurz: ok, I went with "the timeout has expired"

Martchus (approved):

kalikiana (approved):

perlpunk (approved):

PR #2292, opened 2023-04-01
ci: Fix codecov issues

perlpunk:
Issue: https://progress.opensuse.org/issues/125237

See individual commit messages

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -30,10 +30,10 @@ coverage:
           - mmapi.pm
           - osutils.pm
           - signalblocker.pm
-          - t
+          - t/data
+          - t/fake
+          - t/*.t
           - xt
-        ignore:
-          - t/data/tests
     patch:
       default:
         branches: null
--- a/t/data/tests/module1.pm
+++ b/t/data/tests/module1.pm
@@ -1,2 +1,3 @@
 use Mojo::Base 'basetest', -signatures;
+my $x = 'module1';
 1;


codecov.yml
```diff
@@ -30,10 +30,10 @@ coverage:
           - mmapi.pm
           - osutils.pm
           - signalblocker.pm
-          - t
           - xt
-        ignore:
-          - t/data/tests
+          - t/*.t
```
okurz, line 34: I don't understand this line change. We *want* to collect the coverage for the tests themselves and ensure 100% so that we know that every test line is actually executed. The other line changes are fine. Does it still work if you revert this line?

okurz (changes requested):

codecov.yml
perlpunk: I don't understand. My change will make sure that all `t/*.t` are covered 100%

perlpunk: I can make a PR with uncovered lines to prove it actually works: #2293

perlpunk: It worked: `codecov/project/fully_covered Failing after 1s — 99.99% (target 100.00%)`

okurz: Oh, you actually removed the line "ignore" , I didn't see that 🤣

okurz (approved):

codecov.yml
perlpunk: Maybe I should have kept the correct sort order, I think then the diff would have been clearer :)
Did that now, also to trigger OBS checks

Martchus (approved):

PR #2295, opened 2023-04-04
[WIP] Fix script_run waiting for escape sequences

dzedro:
`$cmd` can contain escape sequences, mostly new line \n, but on serial console are escape sequences expanded thus do not match, this could be solved with some magic, but I don't think it's needed
We don't wait for `$cmd` and marker in other consoles, no reason to do it in serial console which is more robust AFAIK
Optionally moved "\n" into same type_string with `$marker`, one line less in log

This line is gone with `testapi::type_string($marker . "\n");`
```
[2023-04-03T16:55:06.082436+02:00] [debug] [pid:10170] <<< consoles::serial_screen::type_string(cmd="backend_type_string", json_cmd_token="YAlbgaMk", text="\n")
```
e.g. zypper_lifecycle
before: https://dzedro.suse.cz/tests/934#step/zypper_lifecycle/52 3m 43s
after:
https://dzedro.suse.cz/tests/931#step/zypper_lifecycle/52 47s
https://openqa.suse.de/tests/10854031#step/zypper_lifecycle/52 s390x
https://openqa.suse.de/tests/overview?distri=sle&build=lifecycle s390x

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -126,9 +126,8 @@ sub script_run ($self, $cmd, @args) {
         my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
-            testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
-            testapi::type_string("\n");
+            testapi::type_string("$marker\n");
+            testapi::wait_serial($marker, no_regex => 1, quiet => $args{quiet});
         }
         else {
             testapi::type_string "$marker > /dev/$testapi::serialdev\n";


distribution.pm
```diff
@@ -126,9 +126,8 @@ sub script_run ($self, $cmd, @args) {
         my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
-            testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
-            testapi::type_string("\n");
+            testapi::type_string($marker . "\n");
```
Martchus, line 129:
```suggestion
            testapi::type_string("$marker\n");
```

Martchus (commented):
I guess this should work. Maybe it makes sense to add checks to unit tests. There are likely already tests covering this function.

t/03-testapi.t
```diff
@@ -345,7 +345,12 @@ subtest 'script_run' => sub {
     $fake_exit = 1234;
     is(background_script_run('sleep 10'), '1234', 'background_script_run returns a PID');
     is(background_script_run('sleep 10', output => 'foo'), '1234', 'background_script_run with output returns valid PID');
-};
+
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->redefine(is_serial_terminal => 1);
+    $mock_testapi->redefine(hashed_string => 'XXX');
+    stderr_like { is(assert_script_run('true'), undef, 'nothing happens on success') } qr/wait_serial: (?^u:XXX-\d+-): ok/, 'log';
```
Martchus, line 352: It would be nice if the test messages (`nothing happens on success` and `log`) would be unique within one test-level (here, this sub test). Maybe you can just merge this with the place you've copied these from (as the original test is quite unspecific anyways)?

dzedro: This part does not have coverage, will try to do so.
Screenshot_2023-04-04_15-05-14

Martchus (approved):

t/03-testapi.t
dzedro: @Martchus I'm not able to add working unit test now, I will do it later. PR can stay open, it's not super important.

mimi1vx (approved):

pevik:
@dzedro Could you please retest this after recent svirt serial terminal changes (ssh-virtsh-serial - user-sut-serial). The speedup looks interesting.

mdoucha:
The removed `wait_serial()` call here checks that the command string was correctly received by the SUT. We don't do that on VNC consoles (even though mistyped commands are far more common there) because we'd have to somehow OCR a poorly defined region of a  screenshot.

This PR simply tries to hide a test bug: Multiline commands are not supported on serial consoles. A better solution would be to fix the zypper_lifecycle test and add a sanity check to `script_run()` that'll call `die` if somebody passes command string that contains a newline.

mdoucha (changes requested):
This PR tries to hide a test bug. The real problem is that the zypper lifecycle calls `assert_script_run()` on command strings which contain newlines.

dzedro:
We can have mistyped command also in serial console ? This is not only about zypper_lifecycle, newline is used also in other tests, I guess we will find out when it will die.

mdoucha:
Yes, all types of serial console can mistype commands, though sometimes only as a result of kernel bug or wrong shell settings.

PR #2296, opened 2023-04-12
Drop Leap-15.3 from OBS CI

mimi1vx:
Leap 15.3 is EOL for long time, no need test against it

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,7 +14,7 @@ pull_request_rules:
         # "unresolvable" is not reported in general:
         # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
         # So we need to require the number of tests explicitly:
-        - "#check-success>=14"
+        - "#check-success>=15"
         - "#check-failure=0"
         - "#check-pending=0"
         - linear-history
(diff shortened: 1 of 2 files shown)


perlpunk:
I just wanted to suggest to also reduce the number here: https://github.com/os-autoinst/os-autoinst/blob/master/.mergify.yml#L17
But I saw that we now have 19 checks in this PR. I think that's more than normal because the branch was created in the original repo, and github reports any actions for both push and PR events. So that are 4 extra checks, normally we would have 15 then, that's still more than the 14 that we originally put into the mergify config. So I assume we have added checks after we added that to mergify. So maybe just increase the value to 15?

perlpunk:
I just saw you reduced it to 13. Not sure why.

mimi1vx:
> I just saw you reduced it to 13. Not sure why.

opposite action to time you added 15.4 : D

now so how successful checks are mandatory for mergify?

Martchus (approved):

.mergify.yml
```diff
@@ -14,7 +14,7 @@ pull_request_rules:
         # "unresolvable" is not reported in general:
         # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
         # So we need to require the number of tests explicitly:
-        - "#check-success>=14"
+        - "#check-success>=13"
```
perlpunk, line 17: I would try 15
```suggestion
        - "#check-success>=15"
```

perlpunk (approved):

perlpunk:
This has to be manually merged because of the mergify config change, will do that

PR #2297, opened 2023-04-12
ci: Improve marking uncoverable statements

perlpunk:
It's only possible to find out which statements are covered or not by having a close look at the HTML report of each file, especially at the number in the "stmt" column.
This will also tell you how many statements a line actually has, as it will add the number of calls in an extra row per statement.

Also if a statement is covered and marked as uncovered, it will decrease the overall percentage (look for lines which are marked as uncoverable but have a call number of greater than 0.
So I removed some of the uncoverable markers and added some elsewhere, where one or more statements on a line aren't covered. Now locally I get 100% for all t/*.t files.

Issue: https://progress.opensuse.org/issues/125237

### DIFF
--- a/t/04-check_vars_docu.t
+++ b/t/04-check_vars_docu.t
@@ -60,7 +60,7 @@ sub read_doc () {
                 next unless ($var);
                 $default = '' unless (defined $default);
                 $value = '' unless (defined $value);
-                fail "still missing explanation for backend $backend variable $var" unless $explanation;    # uncoverable statement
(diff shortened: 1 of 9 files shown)


perlpunk:
So as you can see now in the Devel::Cover report https://github.com/os-autoinst/os-autoinst/actions/runs/4680256856/jobs/8291363992?pr=2297#step:3:955 all `t/*.t` files have 100%

Martchus (approved):

PR #2300, opened 2023-04-17
Add workaround to prevent `t/29-backend-generalhw.t` being unstable

Martchus:
* Don't die in `stop_serial_grab` if the process does not exist * In this case the process has most likely already terminated anyways so the post-condition of `stop_serial_grab` is still satisfied. * This avoids the corresponding test to sometimes fail with `Can't kill('-TERM', '2481'): No such process at /opt/backend/generalhw.pm line 178` as in the test setup the process might apparently sometimes not be running anymore (not sure why that would be the case, though).
* Related ticket: https://progress.opensuse.org/issues/127532

### DIFF
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -174,9 +174,11 @@ sub start_serial_grab ($self) {
 }
 
 sub stop_serial_grab ($self, @) {
-    return unless $self->{serialpid};
-    kill("-TERM", $self->{serialpid});
-    return waitpid($self->{serialpid}, 0);
+    return 0 unless $self->{serialpid};
+    eval { kill -TERM => $self->{serialpid} };
+    return waitpid($self->{serialpid}, 0) unless my $error = $@;
+    return -1 if $error =~ qr/No such process/i;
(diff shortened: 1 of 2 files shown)


backend/generalhw.pm
```diff
@@ -174,9 +174,11 @@ sub start_serial_grab ($self) {
 }
 
 sub stop_serial_grab ($self, @) {
-    return unless $self->{serialpid};
-    kill("-TERM", $self->{serialpid});
-    return waitpid($self->{serialpid}, 0);
+    return 0 unless $self->{serialpid};
+    eval { kill('-TERM', $self->{serialpid}) };
```
perlpunk, line 178:
```suggestion
    eval { kill '-TERM => $self->{serialpid} };
```

Martchus: I'd removed the braces but since the quoting around the signal cannot be avoided anyways I'd rather stay away from `=>` for this function.

perlpunk: The quoting can be avoided

okurz (approved):

PR #2303, opened 2023-04-19
Drop setting of TESSDATA_PREFIX in invoke-tests

AdamWill:
It's not appropriate to do this here. This is not the right path on all distros. Forcing this path here makes it impossible to run the tests successfully on e.g. Fedora, where the tesseract data is in /usr/share/tesseract/tessdata (and there is no need to explicitly specify this, as tesseract is built to find it there without any such specification).

### DIFF
--- a/tools/invoke-tests
+++ b/tools/invoke-tests
@@ -52,9 +52,6 @@ fi
 export OS_AUTOINST_BUILD_DIRECTORY=$build_directory
 export OS_AUTOINST_MAKE_TOOL=$make_path
 
-# set TESSDATA_PREFIX for 02-test_ocr.t
-export TESSDATA_PREFIX='/usr/share/tessdata/'
-
 args=()
 # use unbuffer if it was found so e.g. timestamps on OBS builds will be useful
 # note: When executing tests with prove it seems that the output is only flushed after one test has finished unless there's a terminal.


AdamWill:
My first idea here was:

    if [[ -z "$TESSDATA_PREFIX" ]]; then
        # set TESSDATA_PREFIX for 02-test_ocr.t
        export TESSDATA_PREFIX='/usr/share/tessdata/'
    fi

but on second thoughts that still doesn't seem right. tesseract finds its data files just fine on Fedora with no explicit setting of `TESSDATA_PREFIX`, so we shouldn't have to explicitly set it just to make this script not override it.

If this needs to be set to make things work in CI for some reason, it needs to be done somewhere else that is specific to the CI process, not somewhere that gets used whenever the tests are run anywhere.

okurz (changes requested):
I agree. Also we should try to rely less on the invoke tool. I would like to eventually not need it again anymore. But we should only merge this PR if the setting is provided where it's actually still needed

AdamWill:
well, the CI checks passed (except fullstack, where the failure doesn't seem to be anything to do with this), so I don't know where else it would be needed. The RPM spec sets it the same as this file previously did. I did look around but couldn't see anywhere else it would be needed.

AdamWill:
We can tag @mimi1vx , who added this - do you remember why? The commit message just says "Export TESSDATA_PREFIX in check phase of os-autoinst.spec", it does not say why it also adds it to `invoke-tests`.

okurz (approved):
yeah, it looks fine so maybe we don't actually need it in invoke-tests

Martchus:
I've re-triggered the fullstack test as it just failed during the package installation. Let's see whether it passes with this change.

mimi1vx:
@AdamWill because opensuse tumbleweed package uses nonstandard data dir :D and I added TESSDATA_PREFIX to all possible places to be safe

mimi1vx (approved):

PR #2309, opened 2023-04-28
Add s390x support to QEMU backend

LKHN:
These virtual devices added for s390x virtualization:

- `virtio-scsi` as alias of `virtio-scsi-ccw` for SCSI controller
- `virtio-gpu` as alias of `virtio-gpu-ccw` for VirtIO GPU Device
- `virtio-rng` as alias of `virtio-rng-ccw` for VirtIO RNG
- `virtio-tablet` as alias of `virtio-tablet-ccw` for input mouse interaction
- `virtio-keyboard` as alias of `virtio-keyboard-ccw` for keyboard interaction

Disabled audio support since there is no audio virtual device available on this architecture (see `qemu-kvm -device help`)

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -95,7 +95,8 @@ sub configure_controllers ($self, $vars) {
     }
 
     if ($vars->{CDMODEL} eq 'scsi-cd' || $vars->{HDDMODEL} eq 'scsi-hd') {
-        $vars->{SCSICONTROLLER} ||= "virtio-scsi-pci";
+        my $is_s390x = ($vars->{ARCH} // '') eq 's390x';
+        $vars->{SCSICONTROLLER} ||= $is_s390x ? 'virtio-scsi' : 'virtio-scsi-pci';
     }
 
     my $scsi_con = $vars->{SCSICONTROLLER} || 0;
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -573,7 +573,7 @@ sub _set_graphics_backend ($self, $is_arm) {
         $device = "cirrus-vga" if ($vga eq "cirrus");
         $device = "VGA" if ($vga eq "std");
     }
-    my @edids = ("VGA", "virtio-vga", "virtio-gpu-pci", "bochs-display");
+    my @edids = ("VGA", "virtio-vga", "virtio-gpu-pci", "bochs-display", "virtio-gpu");
     if (grep { $device eq $_ } @edids) {
         # these devices support EDID
         $options = ",edid=on,xres=$self->{xres},yres=$self->{yres}";
@@ -705,16 +705,22 @@ sub start_qemu ($self) {
     my $arch = $vars->{ARCH} // '';
     $arch = 'arm' if ($arch =~ /armv6|armv7/);
     my $is_arm = $arch eq 'aarch64' || $arch eq 'arm';
+    my $is_s390x = $arch eq 's390x';
 
     $self->_set_graphics_backend($is_arm);
 
     # misc
     my $arch_supports_boot_order = $vars->{UEFI} ? 0 : 1;    # UEFI/OVMF supports ",bootindex=N", but not "-boot order=X"
     my $use_usb_kbd;
+    my $use_virtio_kbd;
 
     if ($is_arm) {
         $arch_supports_boot_order = 0;
         $use_usb_kbd = 1;
+    }
+    elsif ($is_s390x) {
+        $arch_supports_boot_order = 0;
+        $use_virtio_kbd = 1;
     }
     elsif ($vars->{OFW}) {
         $use_usb_kbd = $self->qemu_params_ofw;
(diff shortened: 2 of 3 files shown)


backend/qemu.pm
```diff
@@ -902,11 +911,12 @@ sub start_qemu ($self) {
         }
 
         unless ($vars->{QEMU_NO_TABLET}) {
-            sp('device', ($vars->{OFW} || $arch eq 'aarch64') ? 'nec-usb-xhci' : 'qemu-xhci');
-            sp('device', 'usb-tablet');
+            sp('device', ($vars->{OFW} || $arch eq 'aarch64') ? 'nec-usb-xhci' : $is_s390x ? 'virtio-tablet' : 'qemu-xhci');
+            unless ($is_s390x) { sp('device', 'usb-tablet'); }
```
Martchus, line 915: This looks rather odd; I would prefer our normal coding style:
```suggestion
            sp('device', 'usb-tablet') unless $is_s390x;
```

backend/qemu.pm
```diff
@@ -805,23 +811,25 @@ sub start_qemu ($self) {
...
     sp('serial', 'chardev:serial0');
 
-    if ($self->requires_audiodev) {
-        my $audiodev = $vars->{QEMU_AUDIODEV} // 'intel-hda';
-        my $audiobackend = $vars->{QEMU_AUDIOBACKEND} // 'none';
-        sp('audiodev', $audiobackend . ',id=snd0');
-        if ("$audiodev" eq "intel-hda") {
-            sp('device', $audiodev);
-            $audiodev = "hda-output";
+    unless ($is_s390x) {
```
Martchus, line 814:
```suggestion
    if (!$is_s390x) {
```

Martchus (changes requested):
I suppose all of this is generally not very problematic because we don't use the QEMU backend for s390x test at SUSE (so it won't break existing use cases). However, tests need to be adapted (to pass again) and supposedly also extended (to cover all newly introduced code paths).

LKHN:
@Martchus Thanks for the review.

I fixed the not passing tests and applied your suggestion. Thanks a lot!

Another issue that I tried to extend the tests with asserting QEMU parameters for s390x. However it fails due the mixed QEMU parameters from the previous tests. How can I generate the QEMU parameters which are use only variables I set for s390x?


Test - https://github.com/LKHN/os-autoinst/commit/8680ff66c6af1cf59499b01cb622c2800d29e766#diff-2c6d71750f92f3631c01e08b2695ed0a936acb082240f62d5c6162ddb5c04ce7
Error - https://github.com/LKHN/os-autoinst/actions/runs/4874753836/jobs/8696080601

Martchus:
Likely by just setting `bmwqemu::vars{ARCH} = 's390x';` before invoking the function you need to test. You can also delete keys/values from that global hash map if they lead to cases you don't want to test.

LKHN:
It looks like, it passes now but no response yet from the codecov -https://app.codecov.io/gh/os-autoinst/os-autoinst/commit/16e5a31424afd5f4dbfa2845c7be237d21aa9ad0

I tried to extend the unit tests to cover the testing of these QEMU options:

When: `ARCH => 's390x', QEMU_VIDEO_DEVICE => 'virtio-gpu', OFW => 0`

- EDID is present for virtio-gpu
- -boot option not present.
- virtio-keyboard is present.
- no audio options present.
- isa-fdc.fdtypeA=none is not present.
- virtio-rng is present.
- virtio-tablet is present.

What is left, is the testing of `virtio-scsi` as `SCSICONTROLLER` Which comes from `OpenQA/Qemu/Proc.pm`.

@Martchus @okurz  Could tell me if this approach of the implementation tests is correct? If yes how can I test the `SCSICONTROLLER` too?

Martchus:
> but no response yet from the codecov

The CI check shows your diff is fully covered. I think that's good enough.

Martchus (approved):

OpenQA/Qemu/Proc.pm
```diff
@@ -95,7 +95,9 @@ sub configure_controllers ($self, $vars) {
     }
 
     if ($vars->{CDMODEL} eq 'scsi-cd' || $vars->{HDDMODEL} eq 'scsi-hd') {
-        $vars->{SCSICONTROLLER} ||= "virtio-scsi-pci";
+        my $arch = $vars->{ARCH} // '';
+        my $is_s390x = $arch eq 's390x';
```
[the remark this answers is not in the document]
LKHN: Updated. Many thanks!

okurz (approved):

LKHN:
@Martchus and @okurz , Thanks for the review, please let me know if we need any further changes.

Martchus:
No further changes are required. The PR hasn't been merged automatically because not all OBS checks have reported back. However, https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2309/os-autoinst shows that all builds were successful. So I'm merging this manually.

PR #2310, opened 2023-04-28
Add support for serial device of s390x in testapi

LKHN:
Use `/dev/ttysclp0` as a serial device for s390x architecture

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -968,6 +968,12 @@ subtest init => sub {
     testapi::init;
     is $testapi::serialdev, 'hvc0', 'init sets serial device for OFW/PPC';
     set_var('OFW', 0);
+    set_var('ARCH', 's390x');
+    set_var('BACKEND', 'qemu');
+    testapi::init;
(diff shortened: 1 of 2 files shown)


Martchus (changes requested):
I couldn't tell whether that would impact our s390x tests for SUSE. That's supposedly something we need to test. Or you further constrain this to be only used with the QEMU backend.

Martchus:
You also need to adapt unit tests.

LKHN:
@Martchus Thanks for the review!

Unfortunately, I don't have any other hardware to test it other backends. So I constrained this to be only used with the QEMU backend. 

The unit test also added.

Martchus (approved):

Martchus:
Oh, I suppose it was you I have just declined a private chat via Matrix. If you still have questions we can chat via Matrix. I would also generally allow a private chat. However, it makes generally more sense to use the openQA room (#openqa:opensuse.org).

LKHN:
Yes, It was me. Okay, I will do. thanks!

PR #2311, opened 2023-04-28
Enable VNC endian conversion when worker is big-endian

LKHN:
When a worker is big-endian (`s390x`), the colors are broken on a little-endian based WebUIs (`x86_64`, `AArch64` and `ppc64le`). Do the endian conversion when `$server_is_big_endian` and `$client_is_big_endian` values returns `true`.

Fixes

openQA_s390x_broken_colors

Tested in these combinations

`x86_64` Worker with `x86_64` WebUI:
```
server_is_big_endian: 0

client_is_big_endian: 0

_do_endian_conversion: 0
```

`s390x` Worker with `s390x` WebUI:

```
server_is_big_endian: 1

client_is_big_endian: 1

_do_endian_conversion: 1
```

`s390x` Worker with `x86_64` WebUI:

```
server_is_big_endian: 1

client_is_big_endian: 1

_do_endian_conversion: 1
```

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -378,7 +378,7 @@ sub _server_initialization ($self) {
     $self->_pixinfo(\%pixinfo);
     $self->_bpp($supported_depths{$self->depth}->{bpp});
     $self->_true_colour($supported_depths{$self->depth}->{true_colour});
-    $self->_do_endian_conversion($self->no_endian_conversion ? 0 : $server_is_big_endian != $client_is_big_endian);
+    $self->_do_endian_conversion($self->no_endian_conversion ? 0 : ($server_is_big_endian && $client_is_big_endian));
 
     if ($name_length) {
         $socket->read(my $name_string, $name_length)


Martchus:
Of course now the question is whether this works with an s390x SUT, an x86 worker and an x86 web UI. Have you tested this? You previous comment unfortunately makes no statements about the SUT's architecture (which may differ with non QEMU backends).

LKHN:
Hi @Martchus, Thanks for the review!

I used this patch in these combinations with QEMU backend:

x86_64 SUT/Worker and x86_64 WebUI:
https://openqa.almalinux.org/group_overview/10

AArch64 SUT/Worker and x86_64 WebUI:
https://openqa.almalinux.org/group_overview/11

ppc64le SUT/Worker and x86_64 WebUI:
https://openqa.almalinux.org/group_overview/12

s390x SUT/Worker and x86_64 WebUI:
https://openqa.almalinux.org/group_overview/13

Martchus:
So what's left of a s390x SUT and the worker and web UI x86_64. Supposedly that's something we need to test ourselves as you apparently don't have such a setup.

consoles/VNC.pm
```diff
@@ -378,7 +378,7 @@ sub _server_initialization ($self) {
     $self->_pixinfo(\%pixinfo);
     $self->_bpp($supported_depths{$self->depth}->{bpp});
     $self->_true_colour($supported_depths{$self->depth}->{true_colour});
-    $self->_do_endian_conversion($self->no_endian_conversion ? 0 : $server_is_big_endian != $client_is_big_endian);
+    $self->_do_endian_conversion(($server_is_big_endian && $client_is_big_endian) ? 1 : 0);
```
okurz, line 381: When you remove the use of "no_endian_conversion" I recommend you also remove that attribute.

Please also extend the explanation in the git commit message a bit to explain the problem with the former approach and how the current approach is better. Right now it kinda says something like "true is true" which doesn't help me that much ;)

okurz (changes requested):

consoles/VNC.pm
Martchus, line 381: And **if** this doesn't break our use case at SUSE, this is simpler written as:
```suggestion
    $self->_do_endian_conversion($server_is_big_endian && $client_is_big_endian);
```

LKHN:
> So what's left is a s390x SUT and the worker and web UI x86_64. Supposedly that's something we need to test ourselves as you apparently don't have such a setup.

@Martchus Sorry, I didn't understand. Don't these tests cover the case you mentioned?

These tests run on a remote s390x worker which connected to x86_64 WebUI
https://openqa.almalinux.org/group_overview/13

consoles/VNC.pm
LKHN: @okurz Thanks for the review! I reverted the change with the `no_endian_conversion` attribute and updated the git message. I hope it's more understandable now.

LKHN: Updated. Thanks a lot!

Martchus:
Not really. The web UI's architecture should not matter. All the web UI does is displaying PNGs that have been created on the worker host. What really matters is what endianness the VNC server provided by the SUT uses and what endianness the worker uses. And the SUT and the worker might have a different endianness. In your case they don't (as the SUT and the worker either both are s390x or x86_64 or whatever). In our case they differ because we still use x86_64 worker hosts to conduct s390x tests (so isotovideo and the VNC code you're changing here run on x86_64) and only the SUT-provided VNC server runs on s390x. So your change might break our case. We need to test whether that's the case and if it is find a different solution.

okurz (approved):

okurz:
The CI job "ci/test" failed now in https://github.com/os-autoinst/os-autoinst/actions/runs/4881705923/jobs/8710922631?pr=2311#step:4:97 with "Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}". Not your fault. Likely a retry works as we have observed this problem in the past being a sporadic problem (more details in https://progress.opensuse.org/issues/128129)

okurz:
I suggest we just merge this and if problems are observed we will revert

Martchus:
Ok, let's merge it manually at a time where we can closely monitor what's happening.

okurz:
Patching our o3 s390x worker rebel manually:

```
systemctl mask --now openqa-auto-update
curl -sSL https://github.com/os-autoinst/os-autoinst/pull/2311.patch | patch -p1 -d /usr/lib/os-autoinst
```

Then

```
openqa-clone-job --within-instance https://openqa.opensuse.org/tests/3263494 _GROUP=0 BUILD= TEST+=okurz-os-autoinst-pr2311
```

1 job has been created:
 - opensuse-Tumbleweed-DVD-s390x-Build20230503-textmode-server@s390x-zVM-vswitch-l2 -> https://openqa.opensuse.org/tests/3265981

okurz:
https://openqa.opensuse.org/tests/3265981 looks good to me, ready to merge.

Martchus (approved):

PR #2315, opened 2023-05-24
Allow a needle to have multiple click points

sdclarke:
Needles will now be able to have multiple click points, provided that each is given an ID. If no ID is specified when calling click_lastmatch, the first click point found will be used

### DIFF
--- a/t/01-test_needle.t
+++ b/t/01-test_needle.t
@@ -434,10 +434,24 @@ subtest 'click point' => sub {
     $needle = needle->new('click-point-center.json');
     is_deeply($needle->{area}->[0]->{click_point}, 'center', 'click point "center" parsed');
 
+    $needle = needle->new('click-point-multiple-ids.json');
+    is_deeply($needle->{area}->[0]->{click_point}, {xpos => 2, ypos => 4, id => "first"}, 'first click point parsed');
+    is_deeply($needle->{area}->[1]->{click_point}, {xpos => 1, ypos => 3, id => "second"}, 'second click point parsed');
+
     like(warning {
             $needle = needle->new('click-point-multiple.json');
-    }, qr/click-point-multiple\.json has more than one area with a click point/, 'warning shown');
-    is_deeply($needle, undef, 'multiple click points not accepted');
+    }, qr/click-point-multiple\.json has more than one area with a click point without assigning IDs to each/, 'warning shown');
+    is_deeply($needle, undef, 'multiple click points without IDs not accepted');
+
+    like(warning {
+            $needle = needle->new('click-point-multiple-mixed-1.json');
+    }, qr/click-point-multiple-mixed-1\.json has more than one area with a click point without assigning IDs to each/, 'warning shown');
+    is_deeply($needle, undef, 'multiple click points without IDs not accepted');
+
+    like(warning {
+            $needle = needle->new('click-point-multiple-mixed-2.json');
+    }, qr/click-point-multiple-mixed-2\.json has more than one area with a click point without assigning IDs to each/, 'warning shown');
+    is_deeply($needle, undef, 'multiple click points without IDs not accepted');
 };
 
 subtest 'workaround property' => sub {
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -321,9 +321,9 @@ subtest 'script_run' => sub {
     is(script_run('false', die_on_timeout => 1, output => 'foo'), '1', 'script_run with no check of success and output, returns exit code');
     is(script_run('false', 0, die_on_timeout => 1), undef, 'script_run with no check of success, returns undef when not waiting');
     $fake_matched = 0;
-    throws_ok { script_run('sleep 13', timeout => 10, die_on_timeout => 1, quiet => 1) } qr/command.*timed out/, 'exception occured on script_run() timeout';
+    throws_ok { script_run('sleep 13', timeout => 10, die_on_timeout => 1, quiet => 1) } qr/command.*timed out/, 'exception occurred on script_run() timeout';
     $testapi::distri->{script_run_die_on_timeout} = 1;
-    throws_ok { script_run('sleep 13', timeout => 10, quiet => 1) } qr/command.*timed out/, 'exception occured on script_run() timeout';
+    throws_ok { script_run('sleep 13', timeout => 10, quiet => 1) } qr/command.*timed out/, 'exception occurred on script_run() timeout';
 
     throws_ok { assert_script_run('sleep 13', timeout => 10, quiet => 1) } qr/command.*timed out/, 'exception occurs on assert_script_run() timeout by default';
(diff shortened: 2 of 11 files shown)


t/01-test_needle.t
```diff
@@ -434,10 +434,24 @@ subtest 'click point' => sub {
     $needle = needle->new('click-point-center.json');
     is_deeply($needle->{area}->[0]->{click_point}, 'center', 'click point "center" parsed');
 
+    $needle = needle->new('click-point-multiple-ids.json');
+    is_deeply($needle->{area}->[0]->{click_point}, {xpos => 2, ypos => 4, id => "first"}, 'first click point patsed');
```
okurz, line 438: patsed?

okurz (changes requested):
I am curious about two points:
1. What are the high-level use cases or benefits that would like to cover?
2.  Have you considered the integration into the openQA webUI needle editor?

t/01-test_needle.t
kalikiana, line 438:
```suggestion
    is_deeply($needle->{area}->[0]->{click_point}, {xpos => 2, ypos => 4, id => "first"}, 'first click point parsed');
```

t/01-test_needle.t
```diff
@@ -434,10 +434,24 @@ subtest 'click point' => sub {
...
+    is_deeply($needle->{area}->[1]->{click_point}, {xpos => 1, ypos => 3, id => "second"}, 'second click point patsed');
```
kalikiana, line 439:
```suggestion
    is_deeply($needle->{area}->[1]->{click_point}, {xpos => 1, ypos => 3, id => "second"}, 'second click point parsed');
```

t/03-testapi.t
```diff
@@ -534,6 +534,24 @@ subtest 'assert_and_click' => sub {
...
     }, 'assert_and_click clicks at the  click point specified as "center"') or diag explain $cmds;
 
+    $cmds = [];
+    @areas = ({x => 50, y => 60, w => 22, h => 20, click_point => {xpos => 5, ypos => 7, id => 'first'}}, {x => 0, y => 0, w => 10, h => 10, click_point => {xpos => 5, ypos => 7, id => 'second'}});
+    ok(assert_and_click('foo', point_id => 'first'));
+    is_deeply($cmds->[1], {
+            cmd => 'backend_mouse_set',
+            x => 55,
+            y => 67,
+    }, 'assert_and_click clicks at the  click point with ID "first"') or diag explain $cmds;
```
kalikiana, line 544:
```suggestion
    }, 'assert_and_click clicks at the click point with ID "first"') or diag explain $cmds;
```

t/03-testapi.t
```diff
@@ -534,6 +534,24 @@ subtest 'assert_and_click' => sub {
...
+
+    $cmds = [];
+    @areas = ({x => 50, y => 60, w => 22, h => 20, click_point => {xpos => 5, ypos => 7, id => 'first'}}, {x => 0, y => 0, w => 10, h => 10, click_point => {xpos => 5, ypos => 7, id => 'second'}});
+    ok(assert_and_click('foo', point_id => 'second'));
+    is_deeply($cmds->[1], {
+            cmd => 'backend_mouse_set',
+            x => 5,
+            y => 7,
+    }, 'assert_and_click clicks at the  click point with ID "second"') or diag explain $cmds;
```
kalikiana, line 553:
```suggestion
    }, 'assert_and_click clicks at the click point with ID "second"') or diag explain $cmds;
```

testapi.pm
```diff
@@ -455,7 +455,9 @@ needle area. If C<$dclick> is set, do double click instead.  C<$mustmatch> can
 be string or C<ARRAYREF> of strings (C<['tag1', 'tag2']>).  C<$button> is by
 default C<'left'>. C<'left'> and C<'right'> is supported. If C<$mousehide> is
 true then always move mouse to the 'hidden' position after clicking to prevent
-to hide the area where user wants to assert/click in second step.
+to hide the area where user wants to assert/click in second step. If
```
kalikiana, line 458:
```suggestion
to hide the area where the user wants to assert/click in the second step. If
```

sdclarke:
>     1. What are the high-level use cases or benefits that would like to cover?
The use case is when there are screens which act as a sort of hub, like a home screen with many different buttons to click to take you to different screens. Instead of requiring a needle of the home screen for each of the different click points, we can have just one and specify which point to click. In theory this should reduce the number of needles required.

>     2. Have you considered the integration into the openQA webUI needle editor?
I haven't considered this yet as I personally don't make much use of the needle editor provided by the webUI, I'll happily take a look into it at some point when I get time though.

testapi.pm
sdclarke: This and the above suggestion both came from copy/pasting existing things within the files, so I'll adjust the places I took them from as well.

Martchus:
I'm trying to think of a concrete use. I guess one could:

1. assert the screen on an overview page/window.
2. open another page/window from there by clicking on one of the click points.
3. do something.
4. close the previously opened page again.
5. assume you're back at 1. and open another page/window from the overview via `click_lastmatch` without asserting the screen again.

Of course this is rather problematic because:

* In step 5. you might not actually be back at 1. if something went wrong.
* You might have done further screen matching in step 3. so you cannot just invoke `click_lastmatch` with a different ID in step 5. because the match from step 1. is not the last match anymore. Maybe that could be improved by returning the click coordinates on the first assertion so the test code can keep them in a local variable and use them later on without relying on `clicki_lastmatch`.

I guess you could use that to test a chess program, e.g. to test the behavior of an invalid move first before doing the actual move. I suppose in such a case it is safe to assume nothing has changed on the chess board. However, to check the behavior of the invalid move you'd likely again need to asset the screen. So the last point I found problematic also applies here.

sdclarke:
>     5. assume you're back at 1. and open another page/window from the overview via `click_lastmatch` without asserting the screen again.

Here instead of assuming and using click_lastmatch, I would assert_and_click again on the original needle, only with a different click point specified.

Currently in our tests we have things such as:
```perl
assert_and_click('home-to-music');
# Do something here...
# Return to the home screen
assert_and_click('home-to-settings');
# etc...
```
Where the PNG associated with the two needles is the same, as they are representing two different click points on the same screen.

Instead of this, being able to have a single PNG image and a single JSON file with all of the necessary information would be ideal, i.e.:
```perl
assert_and_click('home', point_id => 'music');
# Do something here...
# Return to the home screen
assert_and_click('home', point_id => 'settings');
# etc...
```

That's the thinking behind suggesting this change from my side.

Martchus:
Ok, that makes sense to me :-)

okurz (changes requested):
Please address https://github.com/os-autoinst/os-autoinst/pull/2315#issuecomment-1560870659 , rest looks good

Martchus (approved):

PR #2321, opened 2023-06-06
Improve coverage, add tests, ensure all scripts handle invalid options

okurz:
tools/update-deps is a symlink to a script in external/ which is already
excluded from coverage analysis as it comes from an external repository.
tools/update-deps should hence also not be included in our coverage
analysis – especially as its statement coverage is not 100% – and as
tools/update-deps is currently the only file within tools/ that ends up
in coverage data anyway as visible in
https://app.codecov.io/gh/os-autoinst/os-autoinst/tree/master/tools
we should consistently exclude the complete tools directory from
analysis.

### DIFF
--- a/check_needles.pl
+++ b/check_needles.pl
@@ -2,11 +2,24 @@
 
 use Mojo::Base -strict, -signatures;
 use File::Basename;
+use Getopt::Long;
 use needle;
 use cv;
 
 cv::init();
 require tinycv;
+
+Getopt::Long::Configure("no_ignore_case");
+
+my %options;
+
+sub usage ($r) {
+    eval { require Pod::Usage; Pod::Usage::pod2usage($r) };
+    die "cannot display help, install perl(Pod::Usage)\n" if $@;    # uncoverable statement
+}
+
+GetOptions(\%options, 'help|h|?') or usage(1);
+usage(0) if $options{help};
 
 my ($res, $needle, $img);
 my $ndir = $ARGV[0] || ".";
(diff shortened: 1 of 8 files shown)


t/40-check_qemu_oom.t
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env perl
```
Martchus, line 1: Wouldn't it make more sense to add this to `t/44-scripts.t` instead of having a separate file? But I guess it is ok, this test doesn't include much and has not expensive setup.

check_needles.pl
```diff
@@ -8,6 +8,20 @@
 cv::init();
 require tinycv;
 
+use Getopt::Long;
```
Martchus, line 11: I'd prefer to move it up to the other `use …` lines.

Martchus (changes requested):
It looks like some tests fail now.

t/40-check_qemu_oom.t
okurz: t/44-scripts.t is intended to call every script in a loop in the same way, same as we have in openQA. So any specific tests should IMHO be in a separate test unit. In particular for os-autoinst we don't have any expensive database setup so the overhead of a separate file is really small.

isotovideo
```diff
@@ -75,7 +75,11 @@ use OpenQA::Isotovideo::Utils qw(git_rev_parse spawn_debuggers handle_generated_
 
 my %options;
 
+# global exit status
+my $return_code = 1;
```
perlpunk, line 79: You now need to set `$return_code = 0` in `sub version`

okurz:
> It looks like some tests fail now.

fixed

perlpunk (approved):

Martchus (approved):

PR #2322, opened 2023-06-06
Limit the number of test steps/results that can be created

Martchus:
Related ticket: https://progress.opensuse.org/issues/129068


okurz (approved):

Martchus:
This results in a test failure (not incomplete) with

```
[2023-06-06T16:14:52.417984+02:00] [debug] [pid:19304] Maximum allowed test steps (MAX_TEST_STEPS=10000) exceeded, aborting at /hdd/openqa-devel/repos/os-autoinst/basetest.pm line 415.
```

followed by a backtrace.

Not the nicest outcome but I guess acceptable. I'm not even sure whether we should consider this an incomplete because this actually is a test mistake (test produces too many results).

In my local test I've noticed that the limit of 10000 is actually nothing. The web UI can still handle it quite well and it doesn't look *that* huge. In fact, I'm afraid we'd run into that limit in production jobs and would fail those jobs quite needlessly. So I'll do some more local testing to determine a more sensible limit.

okurz:
> This results in a test failure (not incomplete) […]
> Not the nicest outcome but I guess acceptable. I'm not even sure whether we should consider this an incomplete because this actually is a test mistake (test produces too many results).

I would favor "incomplete". When we implemented the "reason" (mostly you did, of course) we decided that incomplete with specific reasons is fine even if the instance admins are not the right target audience but test maintainers. Think of examples like "missing asset", "unable to find schedule file", etc. So can it be made an "incomplete"?

> In my local test I've noticed that the limit of 10000 is actually nothing. The web UI can still handle it quite well and it doesn't look _that_ huge. In fact, I'm afraid we'd run into that limit in production jobs and would fail those jobs quite needlessly. So I'll do some more local testing to determine a more sensible limit.
> 
> My findings so far:
> 
>     * 10 000 is still not problematic
>     * 100 000 is way too much

good point. I would be ok with anything in the range 10k-100k as a default :)

Martchus:
> So can it be made an "incomplete"?

Ok, I suppose that makes sense. And the answer to the question is of course yet. Then a simple "die" doesn't suffice anymore of course.

I'm currently trying 50000 and I guess it is still too big. I guess I'll have to do at least one more test tomorrow.

Martchus:
I've just changed the limit to 50000 which is hopefully high enough to still cover valid production use cases and low enough to now crash the browser for most users (likely depends on the available RAM). Since finding a sensible limit that fits all use cases is likely not possible it is good that this can still be changed via the job setting `MAX_TEST_STEPS`. On that note I'd say it is also supposedly ok to *not* enforce anything on the server-side after all.

kalikiana (approved):

okurz:
I retriggered GHA CI jobs which passed now but we are missing codecov check results even though there is a coverage message comment. Trying rebasing.

@Mergifyio rebase

basetest.pm
```diff
@@ -399,15 +412,21 @@ sub save_test_result ($self) {
     return $result;
 }
 
+sub _increment_test_count ($self, $max = $bmwqemu::vars{MAX_TEST_STEPS} // 50) {
```
perlpunk, line 415: I thought that should be 50000 and not 50? Or am I missing something?

okurz: seems like a problem in rebasing happened here. I will create a fix

okurz: fixed in https://github.com/os-autoinst/os-autoinst/pull/2323/

perlpunk: Fixed by #2323

PR #2326, opened 2023-06-13
Update examples for external video encoder

Martchus:
* Add example for using SVT-1 which seems to be a good alternative to VP9
* Use `-cpu-used 1` in VP9 example as it already provides a considerable speed-up compared to `-cpu-used 0` and produces only slightly larger files
* See https://progress.opensuse.org/issues/129955#note-6


doc/backend_vars.asciidoc
```diff
@@ -17,7 +17,7 @@ _EXIT_AFTER_SCHEDULE;boolean;0;Exit test execution immediately after evaluation
...
 NEEDLES_GIT_REFSPEC;string;;git refspec to checkout within `NEEDLES_DIR`. See `TEST_GIT_REFSPEC` for details.
-EXTERNAL_VIDEO_ENCODER_CMD;string;;Specifies the command line for invoking a custom video encoder. It is supposed to accept a sequence of PPM images via stdin. The placeholder `%OUTPUT_FILE_NAME%` is replaced with the output file path. The output file path is appended if the placeholder is missing. Example: `ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p -c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 0`
+EXTERNAL_VIDEO_ENCODER_CMD;string;;Specifies the command line for invoking a custom video encoder. It is supposed to accept a sequence of PPM images via stdin. The placeholder `%OUTPUT_FILE_NAME%` is replaced with the output file path. The output file path is appended if the placeholder is missing. Examples: `ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p -c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 1`, `ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p -c:v libsvtav1 -preset 10 -crf 35`
```
kalikiana, line 20: I guess the idea is to be able to copy this as-is. Considering the two examples are almost identical, though, maybe it would be better to mention recommended variants explicitly, say `Consider using libsvtav1 for faster encoding and smaller file sizes if it's available for your distribution`.

Martchus: I've been cranking it up a notch in https://github.com/os-autoinst/os-autoinst/pull/2327. I guess after throwing podman into the mix it makes sense to keep those commands separate.

Martchus:
Closing in favor of https://github.com/os-autoinst/os-autoinst/pull/2327 which also updates the documentation.

PR #2328, opened 2023-06-16
Enable external video encoder by default if available

Martchus:
* Enable the use of an external video encoder by default if it
  one is available; the following options are tested and the
  first one wins:
    * Use EXTERNAL_VIDEO_ENCODER_CMD if it is non-empty (as before)
    * Use the built-in encoder if EXTERNAL_VIDEO_ENCODER_CMD is
      present but empty (as before)
    * Use SVT-AV1 if a version of ffmpeg supporting it is in path
    * Use VP9 if a version of ffmpeg supporting it is in path
    * Use the built-in encoder (as before)
* See https://progress.opensuse.org/issues/129955

---

I tested this locally when:
1. ffmpeg is not present in path
2. ffmpeg is present in path but not executable
3. an empty `EXTERNAL_VIDEO_ENCODER_CMD` is configured
4. a non-empty `EXTERNAL_VIDEO_ENCODER_CMD` is configured
5. ffmpeg is present and executable and supports SVT-AV1

In cases 1. to 3. the internal video encoder is used. In 4. and 5. the external video encoder is invoked with expected arguments.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -35,6 +35,7 @@ use OpenQA::NamedIOSelect;
 
 use constant FULL_SCREEN_SEARCH_FREQUENCY => $ENV{OS_AUTOINST_FULL_SCREEN_SEARCH_FREQUENCY} // 5;
 use constant FULL_UPDATE_REQUEST_FREQUENCY => $ENV{OS_AUTOINST_FULL_UPDATE_REQUEST_FREQUENCY} // 5;
+use constant DEFAULT_FFMPEG_CMD => 'ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p';
 
 # should be a singleton - and only useful in backend process
 our $backend;
@@ -342,10 +343,18 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
 }
 
+sub _ffmpeg_banner () { `ffmpeg 2>&1` // '' }
+
+sub _auto_detect_external_video_encoder ($self) {
+    my $ffmpeg_banner = _ffmpeg_banner;
+    return DEFAULT_FFMPEG_CMD . ' -c:v libsvtav1 -crf 50 -preset 7' if $ffmpeg_banner =~ qr/--enable-libsvtav1(\s|$)/;
+    return DEFAULT_FFMPEG_CMD . ' -c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 1' if $ffmpeg_banner =~ qr/--enable-libvpx(\s|$)/;
+}
+
 sub _start_external_video_encoder_if_configured ($self) {
     return 0 if $bmwqemu::vars{NOVIDEO};
 
-    my $cmd = $bmwqemu::vars{EXTERNAL_VIDEO_ENCODER_CMD} or return 0;
+    my $cmd = $bmwqemu::vars{EXTERNAL_VIDEO_ENCODER_CMD} // $self->_auto_detect_external_video_encoder or return 0;
     my $output_file_name = $bmwqemu::vars{EXTERNAL_VIDEO_ENCODER_OUTPUT_FILE_EXTENSION} // 'webm';
     my $output_file_path = "video.$output_file_name";
     $cmd .= " '$output_file_path'" unless $cmd =~ s/%OUTPUT_FILE_NAME%/$output_file_path/;
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -343,7 +343,7 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
 }
 
-sub _ffmpeg_banner () { `ffmpeg 2>&1` // '' }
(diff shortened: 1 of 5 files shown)


backend/baseclass.pm
```diff
@@ -342,10 +343,18 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
 }
 
+sub _ffmpeg_banner () { `ffmpeg 2>&1` // '' }
+
+sub _auto_detect_external_video_encoder ($self) {
+    my $ffmpeg_banner = _ffmpeg_banner;
+    return DEFAULT_FFMPEG_CMD . ' -c:v libsvtav1 -crf 50 -preset 7' if $ffmpeg_banner =~ qr/--enable-libsvtav1(\s|$)/;
+    return DEFAULT_FFMPEG_CMD . ' -c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 1' if $ffmpeg_banner =~ qr/--enable-libvpx(\s|$)/;
```
okurz, line 351:
```suggestion
    return join(' ', DEFAULT_FFMPEG_CMD, $bmwqemu::vars{FFMPEG_ARGS} // '-c:v libsvtav1 -crf 50 -preset 7') if $ffmpeg_banner =~ qr/--enable-libsvtav1(\s|$)/;
    return join(' ', DEFAULT_FFMPEG_CMD, $bmwqemu::vars{FFMPEG_ARGS} // '-c:v libvpx-vp9 -crf 35 -b:v 1500k -cpu-used 1') if $ffmpeg_banner =~ qr/--enable-libvpx(\s|$)/;
```

backend/baseclass.pm
```diff
@@ -35,6 +35,7 @@ use OpenQA::NamedIOSelect;
 
 use constant FULL_SCREEN_SEARCH_FREQUENCY => $ENV{OS_AUTOINST_FULL_SCREEN_SEARCH_FREQUENCY} // 5;
 use constant FULL_UPDATE_REQUEST_FREQUENCY => $ENV{OS_AUTOINST_FULL_UPDATE_REQUEST_FREQUENCY} // 5;
+use constant DEFAULT_FFMPEG_CMD => 'ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p';
```
okurz, line 38: I suggest to replace that constant with a function like

```perl
sub ffmpeg_cmd() { $bmwqemu::vars{FFMPEG_CMD} // 'ffmpeg -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p' }
```

okurz (changes requested):

Martchus:
I'll have to cross check the performance on aarch64 because https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/System-Requirements.md doesn't sound very promising.

backend/baseclass.pm
Martchus: I don't think this is a useful customization point. We already have one via the worker config and adding and additional one that is very redundant is causing more confusion than it helps.

okurz: You mean one can customize the external video encoder command line?

Martchus: Yes, via `EXTERNAL_VIDEO_ENCODER_CMD`.

backend/baseclass.pm
Martchus: I would also not allow this. It makes no sense that both cases would shrink into one case when this variable is specified. If one wanted to customize this then `EXTERNAL_VIDEO_ENCODER_CMD` is the way to go. We could allow `EXTERNAL_VIDEO_ENCODER_FFMPEG_ARGS` in addition so one can just adjust the last part of the command line individually but I would not do this in these special cases. These special cases are really for automatic detection in case one has *not* specified the arguments explicitly and I wouldn't mix the code for adding a customization point into it.

backend/baseclass.pm
okurz: right. So then it's not strictly necessary to make the rest configurable

backend/baseclass.pm
okurz: ok, leave as is

okurz (approved):

Martchus:
I've just wanted to test the performance on aarch64 using a Tumbleweed container but the official Tumbleweed ffmpeg build has no SVT-AV1 support. So I guess we don't have to worry about it at all. (In production we'd fallback to VP9 anyways on all archs as we only use Leap there.)

backend/baseclass.pm
```diff
@@ -342,10 +343,18 @@ sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     $pipe->blocking(!!$bmwqemu::vars{VIDEO_ENCODER_BLOCKING_PIPE});
 }
 
+sub _ffmpeg_banner () { `ffmpeg 2>&1` // '' }
```
perlpunk, line 346:
```suggestion
sub _ffmpeg_banner () { qx{ffmpeg 2>&1} // '' }
```

Martchus: Does this change anything? Why is it preferred? Considering most people know backticks from bash it is likely the easier syntax.

perlpunk: I find backticks not very eye-friendly, and we use `qx` in the majority of the cases in os-autoinst, and it's easier to grep for.

perlpunk (approved):

Martchus:
Let's ignore the missing test coverage of the openvswitch script for now.

PR #2333, opened 2023-06-23
Remove non-functional save_storage_drives from testapi

mimi1vx:
https://progress.opensuse.org/issues/130922

`save_storage_drives` never worked and isn't used by any test
(os-autoinst-distri-*) and with high probability in any other instance
of openQA because nobody reported an issue with it.

As for why it can't work --> `qemu-img` works only with offline images, it
can't work on images used by running or paused VM. Any use of this method
will end with:

~~~
qemu-img: Could not open '/var/lib/openqa/pool/2/raid/hd0-overlay0': Failed to get shared "write" lock
  Is another process using the image [/var/lib/openqa/pool/2/raid/hd0-overlay0]?
~~~

A fix for this is non-trivial and there is minimum users for this
functionality is better to remove this method first to prevent users to use
it and also to remove practically dead code from os-autoinst

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 38;
+our $version = 39;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -488,8 +488,6 @@ sub switch_network ($self, $args) { $self->notimplemented }
 
 sub save_memory_dump ($self, $args) { $self->notimplemented }
 
-sub save_storage_drives ($self, $args) { $self->notimplemented }
-
 ## MAY be overwritten:
 
 # vm's would return
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -322,20 +322,6 @@ sub save_memory_dump ($self, $args) {
     }
 }
 
-sub save_storage_drives ($self, $args) {
-    diag "Attempting to extract disk #$args->{disk}";
-    $self->do_extract_assets(
-        {
-            hdd_num => $args->{disk},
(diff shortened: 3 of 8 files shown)


okurz (changes requested):

Make sure that the git commit message subject line states the *Why* because that goes into the changelog. Think from the changelog reader perspective what you want to read there.

You can not know if this is used by any tests by any user of openQA. I guess you just grepped os-autoinst-distri-opensuse?

I am not convinced we should remove that function. I am sure it has worked in a certain way when it was introduced. Maybe we introduced a regression since then and our unit tests are of course not going deep enough to really cover the functionality.

Did you try to make use of it and try to fix it?

mimi1vx:
Nobody ever reported it’s wrong by design and it can’t work … so premise it isn’t used is with big probability correct

foursixnine:
> You can not know if this is used by any tests by any user of openQA. I guess you just grepped os-autoinst-distri-opensuse?
> 
> I am not convinced we should remove that function. I am sure it has worked in a certain way when it was introduced. Maybe we introduced a regression since then and our unit tests are of course not going deep enough to really cover the functionality.
> 
> Did you try to make use of it and try to fix it?

@mimi1vx how is it failing? Back when I introduced the function, it was meant for developers interested in getting both, a memory dump and a dump of the storage devices available for the SUT... I I have no attachments to the code, however I would suggest deprecating it first and removing it in the future.

Looking at the unit test, it's kind of snake-oily though... but the backend calls should _work_ otherwise we should have found about this long time ago.

mimi1vx:
> @mimi1vx how is it failing? Back when I introduced the function, it was meant for developers interested in getting both, a memory dump and a dump of the storage devices available for the SUT... I I have no attachments to the code, however I would suggest deprecating it first and removing it in the future.
> 

as I wrote in the commit message it can't work -> `qemu-img convert` doesn't work on attached images. And nobody newer used this function --> so nobody found its faulty.

Martchus:
Maybe this is intended to be called when the VM is not in a running state?

okurz (approved):
Ok for me then as well. I will leave it for @foursixnine to decide then how we should proceed

foursixnine:
> Then it really looks like we should remove this function. I'm only wondering why we ever thought this was working (and thus merged it).

The code changed after the Qemu rewrite, in any case  I'm also fine with removing it then too.

foursixnine (approved):

foursixnine:
Hell... I forgot about Mergify! my bad!

PR #2337, opened 2023-07-05
lockapi: add timeout parameter to barrier_wait()

cfconrad:
I try to solve the following problem:
You have a multi-machine Job with multiple test modules. One test-module fail (`Child.TestModule1`) so the barrier `ready1` isn't called.
In this situation, the `Parent` job would wait for barrier `ready1` till `Child` finished or `MAX_JOB_TIME` is reached.
In the given example, `Parent` would wait till MAX_JOB_TIME is reached, because the `Child` job would wait in barrier `setup2` for the `Parent` => **dead-lock**

For me, the best solution to solve this is, _let the job `die()` on timeout_. So fatal test-modules quit the job and in other cases we continue with the next test-module, where a `endless` barrier could sync again.
```
┌───────────────────────────────────────────┐
│Multimachine Testcase                      │
│                                           │
│   ┌───────────────┐  ┌─────────────────┐  │
│   │ Parent        │  │ Child           │  │
│   └───────┬───────┘  └────────┬────────┘  │
│           │                   │           │
│           ▼                   ▼           │
│   ┌───────────────┐  ┌─────────────────┐  │
│   │TestModule1:   │  │TestModule1:     │  │
│   │wait("setup1") │  │wait("setup1)    │  │
│   │wait("ready1") │  │die("some issue")│  │
│   └───────┬───────┘  └────────┬────────┘  │
│           │                   │           │
│           ▼                   ▼           │
│   ┌───────────────┐  ┌─────────────────┐  │
│   │TestModule2:   │  │TestModule2:     │  │
│   │wait("setup2") │  │wait("setup2")   │  │
│   │wait("ready2") │  │wait("ready2")   │  │
│   └───────────────┘  └─────────────────┘  │
│                                           │
└───────────────────────────────────────────┘
```

Verification runs:
 * http://dante.wicked.suse.de/t147 (with patched os-autoinst => good-case)
 * https://openqa.suse.de/tests/11513272 (dead-lock => bad-case)
 
 Try it out:
 ```bash
 host=http://openqa.wicked.suse.de

openqa-cli api --host "$host" -X POST jobs \
ARCH:1=x86_64 ARCH:2=x86_64 DISTRI:1=opensuse DISTRI:2=opensuse FLAVOR:1=CI FLAVOR:2=CI BUILD:1=barrier_test BUILD:2=barrier_test \
BACKEND:1=qemu BACKEND:2=qemu WORKER_CLASS:1=tap,qemu_x86_64 WORKER_CLASS:2=tap,qemu_x86_64 DELAYED_START:1=1 DELAYED_START:2=1 \
'CASEDIR:1=https://github.com/cfconrad/os-autoinst-distri-opensuse#barrier_test' 'CASEDIR:2=https://github.com/cfconrad/os-autoinst-distri-opensuse#barrier_test'  \
MAX_JOB_TIME:1=300 MAX_JOB_TIME:2=300 \
_PARALLEL:2=1 \
IS_PARENT:1=1 \
TEST:1=barrier_parent TEST:2=barrier_child \
SCHEDULE:1="tests/barrier/create,tests/barrier/module1,tests/barrier/module2" SCHEDULE:2="tests/barrier/create,tests/barrier/module1,tests/barrier/module2" | openqa-wait-job --host "$host"
 ```
 
 If it has a chance of merging, let me know and I can add tests...

### DIFF
--- a/lockapi.pm
+++ b/lockapi.pm
@@ -6,6 +6,7 @@ package lockapi;
 
 use Mojo::Base 'Exporter', -signatures;
 use Scalar::Util 'looks_like_number';
+use List::Util qw(min);
 use Time::Seconds;
 our @EXPORT = qw(mutex_create mutex_lock mutex_unlock mutex_try_lock mutex_wait
   barrier_create barrier_wait barrier_try_wait barrier_destroy);
@@ -126,8 +127,9 @@ sub barrier_try_wait ($name, $where = undef, @) {
 }
 
 sub barrier_wait (@args) {
-    my ($name, $where, $check_dead_job) = ref $args[0] eq 'HASH' ? (@{$args[0]}{qw(name where check_dead_job)}) : @args;
+    my ($name, $where, $check_dead_job, $timeout) = ref $args[0] eq 'HASH' ? (@{$args[0]}{qw(name where check_dead_job timeout)}) : @args;
     $check_dead_job = looks_like_number($check_dead_job) && $check_dead_job ? 1 : 0;
+    $timeout = looks_like_number($timeout) && $timeout >= 0 ? $timeout : -1;
 
     bmwqemu::mydie('missing barrier name') unless $name;
     bmwqemu::diag("barrier wait '$name'");
@@ -141,8 +143,17 @@ sub barrier_wait (@args) {
             return 1;
         }
 
-        bmwqemu::diag("barrier '$name' not released, sleeping " . POLL_INTERVAL . ' seconds');    # uncoverable statement
-        sleep POLL_INTERVAL;    # uncoverable statement
+        my $poll_interval = POLL_INTERVAL;
+        if ($timeout != -1) {
+            my $elapsed_time = time - $start;
+            if ($elapsed_time >= $timeout) {
+                bmwqemu::mydie "barrier '$name' timeout after $elapsed_time seconds";
+            } else {
+                $poll_interval = min($timeout - $elapsed_time, $poll_interval);
+            }
+        }
+        bmwqemu::diag("barrier '$name' not released, sleeping " . $poll_interval . ' seconds');
+        sleep $poll_interval;
     }
 }
 
--- a/t/30-mmapi.t
+++ b/t/30-mmapi.t
@@ -7,6 +7,7 @@
 # threads upfront.
 
 use Test::Most;
+use Test::Mock::Time;
 use Mojo::Base -strict, -signatures;
 
 BEGIN {
@@ -117,8 +118,12 @@ $fake_api->post('/barrier' => sub {
 });
 $fake_api->post('/barrier/#name' => sub {
         my ($self) = @_;
+        state $counter = 0;
         my $name = $self->param('name') // '';
+        return $self->render(status => (++$counter % 3 == 0) == 0 ? 200 : 409, text => 'ok') if $name eq 'unblocked_next';
         return $self->render(status => 200, text => 'ok') if $name eq 'unblocked';
+        return $self->render(status => 409, text => 'ok') if $name eq 'blocked';
+        return $self->render(status => 410, text => 'ok') if $name eq 'not-exists';
         return $self->render(status => 200, text => 'ok') if $name eq 'check_dead_job_barrier' && ($self->param('check_dead_job') // '' eq '1');
         return $self->render_mutex;
 });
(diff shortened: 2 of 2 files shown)


Martchus:
Being able to specify a timeout when waiting for something makes generally sense so I suppose this has a chance of merging - especially because it looks like it wouldn't change the default behavior.

Martchus:
I suppose the corresponding `barrier_wait` function in openQA does not need to be adjusted because the timeout is only handled in os-autoinst itself.

t/30-mmapi.t
```diff
@@ -15,6 +15,7 @@ BEGIN {
     $ENV{OS_AUTOINST_MMAPI_RETRY_COUNT} = 1;
     $ENV{OS_AUTOINST_MMAPI_RETRY_INTERVAL} = 0;
     $ENV{OS_AUTOINST_MMAPI_POLL_INTERVAL} = 0;
+    $ENV{OS_AUTOINST_LOCKAPI_POLL_INTERVAL} = 1;
```
Martchus, line 18: It would make sense to use `Test::MockTime` here so the `sleep` doesn't actually sleep. (This module will cause `time` to report the slept time but not actually sleep.)

Of course one could also mock the sleeping manually or just set the interval to zero and the timeout to zero.

lockapi.pm
```diff
@@ -141,6 +141,10 @@ sub barrier_wait (@args) {
             return 1;
         }
 
+        if (looks_like_number($timeout) && time > $start + $timeout) {
+            bmwqemu::mydie "barrier '$name' timeout after " . (time - $start) . "s";
+        }
+
```
okurz, line 147:
```suggestion
        bmwqemu::mydie "barrier '$name' timed out after " . (time - $start) . "s" if looks_like_number($timeout) && time > $start + $timeout;
```

No need for all those additional blank lines or other lines :)

t/30-mmapi.t
okurz: I second that. https://github.com/os-autoinst/os-autoinst/actions/runs/5467368921/jobs/9953579888#step:3:679 from the latest commit on master shows a test runtime of 2s vs. https://github.com/os-autoinst/os-autoinst/actions/runs/5466579991/jobs/9951710094?pr=2337#step:3:679 for your PR. That's a significant and too big increase. I suggest you crosscheck the runtime of t/30-mmapi.t locally and ensure that it doesn't get too high.

okurz (changes requested):

lockapi.pm
cfconrad: If you think this is more readable, okay.

What about `timeout => -1`  maybe we should treat it (or all negative values) as infinitive as well?

t/30-mmapi.t
cfconrad: https://metacpan.org/pod/Test::MockSleep is what we would need, or? I think `Test::MockTime` can set the time value, but we do a sleep which cause the delay. And with
```perl
use Test::MockSleep qw(:with_mocktime);
```
it would be solved.

But I don't see a package for it. How do we install the dependencies for os-autoinst tests?

Martchus: I was under the impression that `use Test::MockTime qw( :all );` would do the trick. The dependency would likely go under https://github.com/os-autoinst/os-autoinst/blob/d13a461849f1d8fe027f64008f18fe48d1c576fa/dependencies.yaml#L76. You'll have to invoke the build system target `pdate-deps` to apply the change to this file in other places.

cfconrad: Ok I think I got it, it's `Test::MockTime` vs `Test::Mock::Time` :facepalm: 

```perl
time perl <<'EOT'
use Mojo::Base -strict;
use Test::Mock::Time;

my ($start) = time;
sleep 2;
my ($end) = time;
say "end:$end - start:$start = " . ($end - $start);
EOT
```

lockapi.pm
cfconrad: I changed the timeout behavior that only positive numbers including 0 are valid. Everything else will be treated as infinitive.

Also, do not wait longer then timeout, if it doesn't match the poll interval.

lockapi.pm
```diff
@@ -141,8 +142,17 @@ sub barrier_wait (@args) {
...
 
-        bmwqemu::diag("barrier '$name' not released, sleeping " . POLL_INTERVAL . ' seconds');    # uncoverable statement
-        sleep POLL_INTERVAL;    # uncoverable statement
+        my $poll_interval = POLL_INTERVAL;
+        if ($timeout != -1) {
+            my $elapsed_time = time - $start;
+            if ($elapsed_time >= $timeout) {
+                bmwqemu::mydie "barrier '$name' timeout after $elapsed_time seconds";
+            } else {
+                $poll_interval = $timeout - $elapsed_time if ($timeout - $elapsed_time < $poll_interval);
```
Martchus, line 151: Just a suggestion:
```suggestion
                $poll_interval = min($timeout - $elapsed_time, $poll_interval);
```
It would require `use List::Util qw(min);`.

Martchus (commented):
It looks generally good.

Martchus (approved):

okurz:
Please fix the typo in the git commit message subject line

cfconrad:
Is something missing?

okurz:
> Is something missing?

Well, one CI check failed. Without that passing this PR will not be merged automatically

okurz (approved):

PR #2342, opened 2023-07-11
Fix script_run sanity checks

mdoucha:
The `script_run()` function checks for mistyped commands on serial terminal but the check does nothing except for slowing down the job when it fails. Make the mistype check failures clearly visible.

In addition, add special error message for command strings containing newline characters because these will cause mistype check failures. Use the newline check even on VNC console so that all invalid usage gets fixed now and switching from VNC to serial console works seamlessly in the future.

Mutually exclusive with #2295.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -127,7 +127,8 @@ sub script_run ($self, $cmd, @args) {
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
+            die 'Command was mistyped'
+              unless testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
             testapi::type_string("\n");
         }
         else {
--- a/distribution.pm
+++ b/distribution.pm
@@ -116,6 +116,7 @@ sub script_run ($self, $cmd, @args) {
             quiet => undef
         }, ['timeout'], @args);
 
+    die "Multiline commands are not supported:\n$cmd;\n" if $cmd =~ m/\n/;
     if (testapi::is_serial_terminal) {
         testapi::wait_serial($self->{serial_term_prompt}, no_regex => 1, quiet => $args{quiet});
     }


mdoucha:
Note: This PR will make some bad code in out tests blow up. That is intentional.

pevik:
@dzedro FYI

Martchus (approved):

distribution.pm
```diff
@@ -116,6 +116,8 @@ sub script_run ($self, $cmd, @args) {
             quiet => undef
         }, ['timeout'], @args);
 
+    die "Multiline commands are not supported:\n$cmd;\n" if $cmd =~ m/\n/;
+
```
okurz, line 120:
```suggestion
    die "Multiline commands are not supported:\n$cmd;\n" if $cmd =~ m/\n/;
```

distribution.pm
```diff
@@ -127,7 +129,8 @@ sub script_run ($self, $cmd, @args) {
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
+            die 'Command was mistyped'
+              unless testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet}, timeout => 5);
```
okurz, line 133: the "5" is a magic number. I suggest to use an overridable constant with an explanatory name and defaulting to 5 if not set from the outside

okurz (changes requested):
Please also bump the version number in https://github.com/os-autoinst/os-autoinst/blob/master/OpenQA/Isotovideo/Interface.pm#L12 as you are introducing a breaking change

distribution.pm
pevik: What is the difference in these two lines?

pevik:
LGTM.

distribution.pm
okurz: The point is that anyone else will wonder what the 5 means. So just by using a properly named constant you can avoid that confusion when reading the code. Likely no one ever would need to override the constant dynamically. This is only about code readability.

distribution.pm
okurz: The difference is the blank line removed. Also I added a comment after that which apparently became lost. So in essence please just remove the redundant blank lines.

okurz:
>  I'm just highlighting things that have been broken for a long time.

I get that. But are you sure that in all cases newlines would really cause problems? Bumping the interface numbers does not harm but can be a hint if the change does have a more visible impact in case of script_run misuses

distribution.pm
mdoucha: Properly named constant such as `SHOULD_MATCH_INSTANTLY` or `DEFAULT_TIMEOUT_IS_WAY_TOO_LONG`? If you're concerned about code readability, then the real problem is Perl. The best I can do here is a code comment.

okurz: if you say this should match instantly would "0" or "1" actually work?

perlpunk: > If you're concerned about code readability, then the real problem is Perl.

Please no language bashing. It doesn't help anyone.
It's easy to write readable code in Perl.
When I'm reading code in a language I'm not very familiar with I also have problems with its readability.

mdoucha: > if you say this should match instantly would "0" or "1" actually work?

"0" will not work for SSH consoles. "1" should work most of the time but it's not robust enough in case of network issues.

kalikiana:
```suggestion
              unless testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet}, timeout => $args{wait_serial_timeout});
```

The reasoning here as far as I understand is a prime case for making it overridable. In unusual cases 1 is not enough but we really don't want the default of 90 seconds either. I'd suggest an argument if not a test setting (but I can't make it a nice looking suggesting here).

mdoucha: In the usual cases, 1 second is more than enough. But I want to leave some safety margin for network glitches and occasional worker overload.

If your worker needs more than 1 second to echo back a string that was just typed even under ideal conditions, you really need to rethink the hardware setup.

distribution.pm
```diff
@@ -127,7 +129,8 @@ sub script_run ($self, $cmd, @args) {
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
+            die 'Command was mistyped'
```
perlpunk, line 132: I'm wondering if a `die` is really what we want.

Normally `script_run` returns undef if the wait_serial at the end returns undef.

It's the decision of the user of the function if they want it to be fatal (usually be using `assert_script_run`) or not.
OTOH a clear distinction that the typing itself didn't work vs. a non-zero return code from the script is a good idea.

If we decide we want this to die here, then maybe `autotest::croak` would make more sense, as that's also used in `_handle_script_run_ret` (called from `assert_script_run`).

distribution.pm
perlpunk: Personally I'm using more blank lines than @okurz does, but in our team I believe the majority doesn't like too much blank lines, so I would suggest to just remove it.

distribution.pm
perlpunk: `croak` usually prints the call stack

Martchus (approved):

distribution.pm
mdoucha: Blank line deleted.

distribution.pm
mdoucha: I've removed the timeout argument. Saving 85 seconds in a rare failure case is not worth this kind of discussion.

pevik (approved):

kalikiana (approved):
We seem to be on the same page now. If this does end up breaking things it can always be reverted.

DimStar77:
https://openqa.opensuse.org/tests/3526645#step/mariadb_odbc/3

seems multiline commands were supported until this change was merged

DimStar77:
I'm  collecting a list of tests failing on this change in https://github.com/os-autoinst/os-autoinst/pull/2342#issuecomment-1691196812 - what is the ETA for these being fixed? After all this change just broke 'working states' of QA runs for Tumbleweed and is now splattering the results with reds. So it's just fair to understand how long this will be and who will be responsible to fix it

DimStar77:
> So let's blow up all the broken code now and fix all of it in one go.

That's a great way - if nobody else would rely on the tool

mdoucha:
> > So let's blow up all the broken code now and fix all of it in one go.
> 
> That's a great way - if nobody else would rely on the tool

Yes, I've warned the Tools team that this PR will break things. It's not my fault they've decided to merge and deploy without coordinating with other teams that this change will require additional work that's hard to do in advance.

michaelgrifalconi:
Now that we have a list of broken things, can we please revert this PR and fix the stuff avoiding an "emergency mode"? 
Test approval queue is already in a bad shape

alvarocarvajald:
Also breaks the support server: https://openqa.suse.de/tests/11909871#step/setup/64

perlpunk:
> Yes, I've warned the Tools team that this PR will break things. It's not my fault they've decided to merge and deploy without coordinating with other teams that this change will require additional work that's hard to do in advance.

I would say that we were simply not aware that it wasn't ready to merge.

mdoucha:
> I would say that we were simply not aware that it wasn't ready to merge.

The PR was ready, it just needed some extra coordination of the post-merge work.

michaelgrifalconi:
I think the problem was not the post-merge work, there should have been **pre-merge** work.
If we know that something will break tests, and we can identify which ones, why not fix the test before merging the breaking change and avoid breaking the update validation process?

IMHO you don't make bad (but working) code blow up in production, you find and fix the bad code, then you make only new bad code blow up when first introduced.

Especially if the revert is not such an immediate thing to do.

perlpunk:
Sometimes the most efficient way to *find* the bad code is to let it break, I agree about that with @mdoucha .
But either the PR was ready to merge or not.
Merging means it will be automatically deployed.
Again, it simply wasn't clear to us that this was not ready to deploy.

mdoucha:
Here are the PRs fixing the remaining failures listed by @DimStar77 above.
Ready for merge:
os-autoinst/os-autoinst-distri-opensuse#17615
os-autoinst/os-autoinst-distri-opensuse#17654
os-autoinst/os-autoinst-distri-opensuse#17655
os-autoinst/os-autoinst-distri-opensuse#17658

Waiting for verification run to finish:
os-autoinst/os-autoinst-distri-opensuse#17657

PR #2343, opened 2023-07-17
Make it possible to run more than one python test module

perlpunk:
Before all functions would be imported to the global namespace.

Issue: https://progress.opensuse.org/issues/132656

I had to rename test.py to pythontest.py because otherwise python was complaining:

    ImportError: Module 'test' is not installed.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -114,12 +114,22 @@ sub loadtest ($script, %args) {
     elsif ($script =~ m/\.py$/) {
         # Adding the include path of os-autoinst into python context
         my $inc = File::Basename::dirname(__FILE__);
-        $code .= "
+        my $script_dir = path(File::Basename::dirname($script_path))->to_abs;
+        $code .= <<~"EOM";
             use base 'basetest';
-            use Mojo::File 'path';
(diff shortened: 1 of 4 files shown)


autotest.pm
```diff
@@ -114,14 +114,29 @@ sub loadtest ($script, %args) {
...
+            for my \$func (qw(@mapped_functions)) {
+                if (grep { \$_ eq \$func } \@{ \$info{functions} }) {
+                    py_bind_func("${name}::\$func", "$name", \$func);
+                }
+            }
+            EOM
         $is_python = 1;
     }
+
+
```
okurz, line 139: Unrelated change

perlpunk:
I added a second test module in our unit tests, and I also run the tests now to make sure the two modules do different things.

perlpunk:
Also I now map all python functions to perl because I realized there are much more functions we call which might be overridden by the test.

perlpunk:
Blocked by #2344

kalikiana (approved):

kalikiana:
@mergifyio rebase

okurz (approved):

PR #2344, opened 2023-07-17
Update Perl::Tidy to 20230701

perlpunk:


kraih (approved):

t/develcover.t
```diff
@@ -0,0 +1,18 @@
...
+    my $res = -1;
+    return pass 'foo' if $res == -1;
+}
+
+subtest 'color output can be configured via the command-line' => sub {
+    isotovideo();
+    pass "ok";
+};
+
+done_testing();
```
kraih, line 18: What are these new tests about? Accidental addition?

perlpunk: oops! accidentally added local tests I was playing with. fixed

okurz (approved):
Looks good but we should give some heads-up time for according downstream test distribution changes to be prepared.

perlpunk:
> Looks good but we should give some heads-up time for according downstream test distribution changes to be prepared.

Ok. But note that other os-autoinst PRs will be blocked on this.
https://github.com/os-autoinst/os-autoinst/pull/2343 is failing because of this, because the used container already has the new perltidy version installed.

okurz:
Yes, sure. I already pinged people and "QE Core" has reacted but I don't think they have prepared a PT yet. Ok, I guess 8h need to suffice ;)

PR #2345, opened 2023-07-25
Revert "Use tempfile TMPDIR option"

perlpunk:
This reverts commit 3766bf348dfd9732cd2dc230b712ed0ed339f811.

The problem seems to be that we are setting $ENV{TMPDIR} to the pool directory in openQA, and when $libvirtauthfilename is created in there, it gets removed for some reason.
https://github.com/os-autoinst/os-autoinst/commit/3766bf348dfd9732cd2dc230b712ed0ed339f811#r122622230

I guess we just have to rely on the fact that `/tmp/` will always exist :)

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -483,7 +483,7 @@ sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 sub define_and_start ($self) {
     my $remote_vmm = "";
     if ($self->vmm_family eq 'vmware') {
-        my ($fh, $libvirtauthfilename) = File::Temp::tempfile('libvirtauth-XXXX', TMPDIR => 1);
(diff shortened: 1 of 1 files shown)


mloviska:
Thanks!

mloviska:
Can you please share when we should expect the changes in o.s.d?

perlpunk:
> Can you please share when we should expect the changes in o.s.d?

Not sure about that, as we also still have issues with the moved o3 machine. @okurz any idea? The last update on osd was July 20

perlpunk:
One problem is gitlab which is not running any job currently, and that's what we need for deployment. :(

okurz:
I retriggered the pipeline after the gitlab instance is up again. You can monitor https://gitlab.suse.de/openqa/osd-deployment/-/pipelines/740680 (SUSE internal) for the progress

perlpunk:
@mloviska deployed!

mloviska:
> @mloviska deployed!

It works, thanks!

https://openqa.suse.de/tests/11661027#step/bootloader_uefi/4

PR #2347, opened 2023-07-26
Add multi-machine setup script

okurz:
Related progress issue: https://progress.opensuse.org/issues/133025


script/os-autoinst-setup-multi-machine
```diff
@@ -0,0 +1,52 @@
...
+for i in $(seq 1 $instances; seq 64 $((64+$instances)); seq 128 $((128+$instances))); do ln -sf ifcfg-tap0 /etc/sysconfig/network/ifcfg-tap$i && echo "OVS_BRIDGE_PORT_DEVICE_$i='tap$i'" >> /etc/sysconfig/network/ifcfg-$bridge; done
+cat > /etc/firewalld/zones/trusted.xml <<EOF
+<?xml version="1.0" encoding="utf-8"?>
+<zone target="ACCEPT">
+  <short>Trusted</short>
+  <description>All network connections are accepted.</description>
+  <service name="isotovideo"/>
+  <interface name="$bridge"/>
+  <interface name="ovs-system"/>
+  <interface name="eth0"/>
```
Martchus, line 37: The interface should be configurable, similar to the bridge.

okurz: done

okurz:
Added commit "Include os-autoinst-setup-multi-machine in doc+package"

okurz:
I don't understand

```
[  133s] found conflict of os-autoinst-4.6.1690870233.228c7a0-lp155.1624.1.x86_64 with os-autoinst-openvswitch-4.6.1690870233.228c7a0-lp155.1624.1.x86_64:
[  133s]   - /usr/bin/os-autoinst-setup-multi-machine
```

anyone has an idea?

CMakeLists.txt
```diff
@@ -227,6 +227,11 @@ foreach (EXTRA_SCRIPT check_qemu_oom dewebsockify vnctest)
         PERMISSIONS ${EXECUTABLE_PERMISSIONS}
     )
 endforeach ()
+install(
+    FILES "script/os-autoinst-setup-multi-machine"
+    DESTINATION "${CMAKE_INSTALL_BINDIR}"
+    PERMISSIONS ${EXECUTABLE_PERMISSIONS}
```
Martchus, line 233:
```suggestion
    PERMISSIONS ${EXECUTABLE_PERMISSIONS}
    COMPONENT "openvswitch" EXCLUDE_FROM_ALL
```

Martchus (changes requested):

Martchus:
Maybe that new script gets pulled into the main package as well via that macro: https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#%perl_gen_filelist

okurz:
> Maybe that new script gets pulled into the main package as well via that macro: https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#%perl_gen_filelist
> 
> I found the use of this macro very confusing/unintuitive. So if it goes in the way I suggest to simply remove the use of `%perl_gen_filelist` and list those files explicitly.

Ok, for now I just removed the explicit mention of the file

Martchus:
But then it will end up in the wrong package.

okurz:
> But then it will end up in the wrong package.

Well, I wouldn't care. It's just a simple shell script :)

Martchus (approved):

script/os-autoinst-setup-multi-machine
```diff
@@ -0,0 +1,80 @@
...
+  <interface name="$ethernet"/>
+  <masquerade/>
+</zone>
+EOF
+    cat > /etc/wicked/scripts/gre_tunnel_preup.sh <<EOF
+#!/bin/sh
+action="\$1"
+bridge="\$2"
+ovs-vsctl set bridge \$bridge stp_enable=true
+# TODO add entries according to your network topology
```
kalikiana, line 65: What is this TODO telling me? Do we need to add things here? Are we expecting users to edit the script before using it? If so, make it fail, so they know to edit it

okurz: Yes, the admin reading this needs to patch up the script afterwards. If it would fail then wicked would likely ignore the failure anyway.

kalikiana (approved):

PR #2350, opened 2023-08-01
consoles: Mention VNC keyboard layout in errors

kalikiana:
Unify the error which is being used in 3 different places.

See: https://progress.opensuse.org/issues/133451

### DIFF
--- a/consoles/VNC.pm
+++ b/consoles/VNC.pm
@@ -588,6 +588,10 @@ sub shift_keys () {
 
 ## use critic
 
+sub die_on_invalid_mapping ($key) {
+    die decode_utf8 "No map for '$key' - layouts other than en-us are not supported\n";
+}
+
 sub init_x11_keymap ($self) {
     return if $self->keymap;
     # create a deep copy - we want to reuse it in other instances
@@ -606,7 +610,7 @@ sub init_x11_keymap ($self) {
     }
     # VNC doesn't use the unshifted values, only prepends a shift key
     for my $key (keys %{shift_keys()}) {
-        die "no map for $key" unless $keymap{$key};
+        die_on_invalid_mapping($key) unless $keymap{$key};
         $keymap{$key} = [$keymap{shift}, $keymap{$key}];
     }
     $self->keymap(\%keymap);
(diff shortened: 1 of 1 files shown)


consoles/VNC.pm
```diff
@@ -587,6 +587,10 @@ sub shift_keys () {
 
 ## use critic
 
+sub _invalid_mapping ($self, $key) {
+    die "No map for '$key' - layouts other than en-us are not supported"
```
okurz, line 591: this can just be a free method

```suggestion
sub die_on_invalid_mapping ($key) {
    die "No map for '$key' - layouts other than en-us are not supported"
```

consoles/VNC.pm
```diff
@@ -605,7 +609,7 @@ sub init_x11_keymap ($self) {
     }
     # VNC doesn't use the unshifted values, only prepends a shift key
     for my $key (keys %{shift_keys()}) {
-        die "no map for $key" unless $keymap{$key};
+        $self->_invalid_mapping($key) unless $keymap{$key};
```
okurz, line 612:
```suggestion
        die_on_invalid_mapping($key) unless $keymap{$key};
```

consoles/VNC.pm
```diff
@@ -627,7 +631,7 @@ sub init_ikvm_keymap ($self) {
     }
     my %map = %{shift_keys()};
     while (my ($key, $shift) = each %map) {
-        die "no map for $key" unless $keymap{$shift};
+        $self->_invalid_mapping($key) unless $keymap{$shift};
```
okurz, line 634:
```suggestion
        die_on_invalid_mapping($key) unless $keymap{$shift};
```

consoles/VNC.pm
```diff
@@ -657,7 +661,7 @@ sub map_and_send_key ($self, $keys, $down_flag, $press_release_delay) {
             next;
         }
         else {
-            die "No map for '$key'";
+            $self->_invalid_mapping($key)
```
okurz, line 664:
```suggestion
            die_on_invalid_mapping($key)
```

okurz (changes requested):

consoles/VNC.pm
kalikiana: I was assuming we'll eventually use the layout in the error message but I don't really care

perlpunk:
That's a large diff. Can you rebase?

kalikiana:
> It doesn't look like this would fix any encoding issues yet but I'm fine with it.

Yes, that's going to be a separate PR because it's not a problem with the VNC backed as such.

Martchus (approved):

perlpunk:
> It doesn't look like this would fix any encoding issues yet but I'm fine with it.

#2351

PR #2354, opened 2023-08-18
Improve setting default zone in os-autoinst-setup-multi-machine

asdil12:


script/os-autoinst-setup-multi-machine
```diff
@@ -19,7 +19,7 @@ configure_firewall() {
     firewall-cmd --permanent --new-service isotovideo
     for i in $(seq 1 $instances); do firewall-cmd --permanent --service=isotovideo --add-port=$((i * 10 + 20003))/tcp ; done
     firewall-cmd --permanent --zone="$zone" --add-service=isotovideo
-    firewall-cmd --set-default-zone="$zone" ||:
+    firewall-cmd --get-default-zone | grep -q "$zone" || firewall-cmd --set-default-zone="$zone"
```
kalikiana, line 22: Can you clarify why this is needed? Why not just always set it?

asdil12: Because if it is already set, the very intelligent firewall-cmd exits with status > 0 and the script crashes as it is run with `-e`.
And this way of setting it would not ignore actual errors.

Martchus (approved):

perlpunk:
The commit message has a typo: `Improve settign default zone in os-autoinst-setup-multi-machine`

kalikiana (approved):

PR #2356, opened 2023-08-24
Locate actual wheels in WHEELS_DIR

kalikiana:
- The wheels need to be stored in a way that they can be modified easily rather than pool which is volatile.
- By default the pool is used as always.
- When cloning the path is explicitly logged.

Fixes: https://progress.opensuse.org/issues/134390


perlpunk:
t/08-autotest.t and t/14-isotovideo.t  failing
https://github.com/os-autoinst/os-autoinst/actions/runs/6022113292/job/16336207154?pr=2356#step:3:930

autotest.pm
```diff
@@ -45,7 +45,8 @@ loadtest is called.
 =cut
 
 sub find_script ($script) {
-    if (defined(my $wheel = (glob Cwd::getcwd . "/*/tests/$script"))) {
+    my $wheels_dir = $bmwqemu::vars{WHEELS_DIR} // Cwd::getcwd . "/*/tests/";
```
Martchus, line 48: Is this variable used anywhere?

autotest.pm
```diff
@@ -45,7 +45,8 @@ loadtest is called.
...
+    if (defined(my $wheel = (glob $bmwqemu::vars{WHEELS_DIR} . "/*/tests/$script"))) {
```
perlpunk, line 49:
```suggestion
    my $wheels_dir = $bmwqemu::vars{WHEELS_DIR} // Cwd::getcwd;
    if (defined(my $wheel = (glob $wheels_dir . "/*/tests/$script"))) {
```

t/14-isotovideo.t
```diff
@@ -130,47 +130,48 @@ subtest 'isotovideo with wheels' => sub {
...
-    path($pool_dir, 'writer', 'tests', 'pen')->make_path->child('ink.pm')->spurt("use Mojo::Base 'basetest'; use Copy::Writer::Content 'write'; sub run {}; 1");
+    path($wheels_dir, 'writer', 'lib', 'Copy', 'Writer')->make_path->child('Content.pm')->spurt("package Copy::Writer::Content; use Mojo::Base 'Exporter'; our \@EXPORT_OK = qw(write); sub write {}; 1");
+    path($wheels_dir, 'writer', 'tests', 'pen')->make_path->child('ink.pm')->spurt("use Mojo::Base 'basetest'; use Copy::Writer::Content 'write'; sub run {}; 1");
     my $log = combined_from { isotovideo(
-            opts => "wheels_dir=$wheels_dir casedir=$data_dir/tests schedule=pen/ink _exit_after_schedule=1") };
+            opts => "wheels_dir=$wheels_dir casedir=$case_dir schedule=pen/ink _exit_after_schedule=1") };
     like $log, qr@Skipping to clone.+copy/writer@, 'already cloned wheel picked up';
     like $log, qr/scheduling ink/, 'module from the wheel scheduled';
-    rmtree "$pool_dir/writer";
+    rmtree "$wheels_dir/writer";
```
perlpunk, line 174:
```suggestion
    rmtree "$wheels_dir/writer";
    unlink "$data_dir/tests/wheels.yaml";
```

autotest.pm
kalikiana: Thanks! I didn't realize I hadn't replaced that correctly

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/6037760508/job/16382651406?pr=2356#step:3:894
```
Cannot read t/data/wheels_dir/writer/lib/Copy/Writer/Content.pm: No such file or directory at /usr/lib/perl5/site_perl/5.38.0/Devel/Cover/Report/Codecovbash.pm line 76.
```
I think `t/data/wheels_dir` needs to be added to this `ignore`:
https://github.com/os-autoinst/os-autoinst/blob/master/tools/invoke-tests#L47

autotest.pm
perlpunk: That's s not exactly what I suggested.
You append `"/*/tests/"` to the cwd, and one line later you append `"/*/tests/$script"` again to that string.
It only doesn't complain because we don't have a test case where the wheel would be in cwd.

autotest.pm
perlpunk: I intentionally left out the `. "/*/tests/"` in the first line for a reason. See my other comment.

perlpunk (changes requested):
I think we need to test the case where the wheel would be located in the pool dir as well. As I wrote in my comment, this would fail with the PR in its current form.

autotest.pm
kalikiana: I added a test that I think should reproduce it. Unfortunately I can't get it to fail with the wrong code in `find_script` 🤔 I'm pushing it anyway, maybe you have an idea

t/14-isotovideo.t
```diff
@@ -130,47 +130,58 @@ subtest 'isotovideo with wheels' => sub {
...
+        chdir $pool_dir;
+        $bmwqemu::vars{WHEELS_DIR} = $dir;
+        $wheels_dir = $dir // $pool_dir;
+
+        subtest "wheels in $wheels_dir" => sub {
+            $specfile->spurt("version: v0.1\nwheels: [copy/writer]");
+            path($wheels_dir, 'writer', 'lib', 'Copy', 'Writer')->make_path->child('Content.pm')->spurt("package Copy::Writer::Content; use Mojo::Base 'Exporter'; our \@EXPORT_OK = qw(write); sub write {}; 1");
+            path($wheels_dir, 'writer', 'tests', 'pen')->make_path->child('ink.pm')->spurt("use Mojo::Base 'basetest'; use Copy::Writer::Content 'write'; sub run {}; 1");
+            my $log = combined_from { isotovideo(
+                    opts => "wheels_dir=$wheels_dir casedir=$case_dir schedule=pen/ink _exit_after_schedule=1") };
```
perlpunk, line 178: Here we always set `wheels_dir=$wheels_dir`, so it never tests the case where it's undef.
`$bmwqemu::vars{WHEELS_DIR}` is set in the test process, but isotovideo is running in its own process

autotest.pm
perlpunk: See https://github.com/os-autoinst/os-autoinst/pull/2356#discussion_r1317005147

autotest.pm
```diff
@@ -45,7 +45,8 @@ loadtest is called.
 =cut
 
 sub find_script ($script) {
-    if (defined(my $wheel = (glob Cwd::getcwd . "/*/tests/$script"))) {
+    my $wheels_dir = $bmwqemu::vars{WHEELS_DIR} // Cwd::getcwd;
+    if (defined(my $wheel = (glob $wheels_dir . "/*/tests/$script"))) {
```
Martchus, line 49: This makes the order of operations more clear by using *less* code:
```suggestion
    if (defined(my $wheel = glob "$wheels_dir/*/tests/$script")) {
```

kalikiana:
Ah! I forgot to add the docs update. Will push shortly.

perlpunk (approved):

doc/backend_vars.asciidoc
```diff
@@ -6,10 +6,10 @@ Supported variables per backend
...
+CASEDIR;string;;Absolute or relative path to the test distribution, which can contain an optional `wheels.yaml`. Specifying the variable is mandatory. Can also be a Git repository URL of a test distribution to checkout with an optional refspec or Git commit hash into the current directory. It tries to follow the definition of https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies (e.g. `https://github.com/os-autoinst/os-autoinst-distri-opensuse.git#feature/test` or `git@github.com:os-autoinst/os-autoinst-distri-opensuse.git`). Note that cloning via SSH is only possible if Git is configured accordingly on the host where isotovideo is executed. Can be combined with `TEST_GIT_REFSPEC` and `NEEDLES_GIT_REFSPEC`. Relative paths are considered relative to the working directory of `isotovideo`.
```
okurz, line 9: I suggest to not mention wheels.yaml in the very first sentence of CASEDIR. Think from perspective of a novice user reading about CASEDIR for the first time. The wheels are an optional feature that are not relevant for many distributions from the start. I suggest to not mention wheels.yaml here in this line at all.

doc/backend_vars.asciidoc
```diff
@@ -6,10 +6,10 @@ Supported variables per backend
...
 PRODUCTDIR;string;;Absolute or relative path to the optional "product directory" which includes the test schedule entry point `main.pm` as well as a `needles` subdirectory with the needles to load. Defaults to `CASEDIR`. Relative paths are considered relative to `CASEDIR`.
 NEEDLES_DIR;string;;Absolute or relative path to the needles subdirectory. Defaults to `needles` within `PRODUCTDIR`. Can also be a Git repository URL, comparable to `CASEDIR`. Relative paths are considered relative to the working directory of `isotovideo`.
-WHEELS_DIR;string;;Absolute or relative path to the wheels subdirectory. Defaults to `CASEDIR`. Relative paths are considered relative to the working directory of `isotovideo`.
+WHEELS_DIR;string;;Absolute or relative path to the wheels subdirectory. Defaults to the working directory of `isotovideo`. Relative paths are considered relative to the working directory of `isotovideo`.
```
okurz, line 12: Also shortly discussed with perlpunk. This variable should explain that the wheels subdirectory is the place where wheels are checked out to automatically. Also s/subdirectory/directory/. And better specify the wheels.yaml from above in this context.

okurz (changes requested):

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst/actions/runs/6015946938/job/16318885617#step:3:647 shows that previously the test took 62s. Now in https://github.com/os-autoinst/os-autoinst/actions/runs/6147257051/job/16678324796?pr=2356#step:3:647 75s. That is a runtime increase that I consider too big to be acceptable.

I created
https://progress.opensuse.org/issues/135521 for t/14-isotovideo.t running into timeouts already in local executions.

perlpunk:
>  I checked again if we can avoid calling isotovideo here - and eventually realized that we can't test that libs from a wheel are loaded any other way.

I'm wondering about that. Maybe with the right mocking it can be done?

kalikiana:
> > I checked again if we can avoid calling isotovideo here - and eventually realized that we can't test that libs from a wheel are loaded any other way.
> 
> I'm wondering about that. Maybe with the right mocking it can be done?

My concern was more to do with being able to reproduce the loading of the libraries. Good news is I was able to reproduce it via load_test_schedule. Bad news is it doesn't in fact add the inc path.

perlpunk:
I played with it and got this working: https://github.com/perlpunk/os-autoinst/commit/cffabc6dfc85c22d6ab9a15e801a926d9ea94cb5
* I had to replace `glob` with `File::Glob::bsd_glob` and don't really know why. The normal `glob` had problems on the pool directory.
* Instead of mocking `checkout_git_repo_and_branch`, which does the actual push to `@INC`, I only mock `clone_git`
* I created two different sets of wheels because once a certain Package::Name is loaded, perl won't load it again (unless you cheat and remove it from `%INC`). I decided to make it distinguishable with the added numeric suffix.

t/14-isotovideo.t
```diff
@@ -128,49 +128,60 @@ subtest 'isotovideo with git refspec specified' => sub {
...
+        $wheels_dir = $dir // $pool_dir;
+        $bmwqemu::vars{CASEDIR} = $case_dir;
+        $bmwqemu::vars{SCHEDULE} = 'pen/ink';
+        $bmwqemu::vars{WHEELS_DIR} = $dir;
+
+        subtest "wheels in $wheels_dir" => sub {
+            path($wheels_dir, 'writer', 'lib', 'Copy', 'Writer')->make_path->child('Content.pm')->spurt("package Copy::Writer::Content; use Mojo::Base 'Exporter'; our \@EXPORT_OK = qw(write); sub write {}; 1");
+            path($wheels_dir, 'writer', 'tests', 'pen')->make_path->child('ink.pm')->spurt("use Mojo::Base 'basetest'; use Copy::Writer::Content 'write'; sub run {}; 1");
+            checkout_wheels($case_dir, $wheels_dir);
+            my $log = lives_ok { combined_from { load_test_schedule } } 'no error';
```
perlpunk, line 179: `lives_ok` does not return a log

kalikiana: This was meant to keep the test plan from breaking when git fails. I guess we don't need it with your approach. The non-working case is finally less annoying and fails instantly!

PR #2361, opened 2023-09-15
script/os-autoinst-setup-multi-machine: Drop CAP_NET_ADMIN on qemu

Vogtinator:
The tap interfaces are owned by the worker user already.


nicksinger:
Are you sure this is not required? I remember some "permission denied" error before setting this. Not sure why, though.

okurz (changes requested):
Please phrase the git commit message subject line in a way that describes the "why" not just the "what". Same as Nick stated the command is there for a reason because that "made it work". How can we verify that this is fine. Would you be able to call this script in a VM for example? At best of course already a full blown openQA test using that internally ;)

Vogtinator:
Well, the commit message is entirely the why...

openqaworker-arm22 and openqaworker22 do not have the capability set and work fine. Both with AppArmor in enforced mode as well.

okurz:
> Well, the commit message is entirely the why...

yes, but not the *subject*. So just try to turn around the two lines :) One reason I am asking for that is because only the subject, i.e. the first line of the commit message ends up in changelogs, is displayed in git blame, etc.

> openqaworker-arm22 and openqaworker22 do not have the capability set and work fine. Both with AppArmor in enforced mode as well.

Ok, that's a good indication. Works for me as reason enough :)

Vogtinator:
> > Well, the commit message is entirely the why...
> 
> yes, but not the _subject_. So just try to turn around the two lines :) One reason I am asking for that is because only the subject, i.e. the first line of the commit message ends up in changelogs, is displayed in git blame, etc.

Yes: The subject is the "what" and the summary/description contains the rationale and details of the change. Doing it the other way around won't work FWICT.

okurz:
ok, how about you change the first line to `script/os-autoinst-setup-multi-machine: Drop not needed setcap`

Vogtinator:
> ok, how about you change the first line to `script/os-autoinst-setup-multi-machine: Drop not needed setcap`

Done (kind of)

okurz (approved):

Martchus (approved):

okurz:
@Mergifyio rebase

okurz:
@Vogtinator please rebase

PR #2364, opened 2023-09-21
IPMI backend incorrectly handle empty return value

Dawei-Pang:
Refer to the issue: https://progress.opensuse.org/issues/136226

There is something wrong during handle the return value of IPC::Run::run.

Refer to https://metacpan.org/pod/IPC::Run#RETURN-VALUES: "run() and finish() return TRUE when all subcommands exit with a 0 result code. This is the opposite of perl's system() command."

In my experiments, $ret = IPC::Run::run(), when `$ret = ""` the ipmitool run into failure, `$ret=1` run success.

I need widely verify this change on bare-metal systems.

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -32,7 +32,7 @@ sub ipmitool ($self, $cmd, %args) {
     my @tries = (1 .. $args{tries});
     for (@tries) {
         $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
-        if ($ret == 0) {
+        if ($ret) {
             $self->dell_sleep;
             last;
         } else {


Julie-CAO:
Thank you for the fix, @Daiwei-Pang. I also found there is a problem with the return code but I did not noticed the incorrect "==" . Good catch.

backend/ipmi.pm
```diff
@@ -32,7 +32,7 @@ sub ipmitool ($self, $cmd, %args) {
     my @tries = (1 .. $args{tries});
     for (@tries) {
         $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
-        if ($ret == 0) {
+        if ($ret eq 0) {
```
Martchus, line 35: > if $ret is empty or undefined, $ret == 0 is true, we need use "eq".

If `$ret` is `undef` the new code will produce a warning. Considering you say `undef` is to be expected this should probably be:

```suggestion
        if (defined $ret && $ret eq 0) {
```

It is also a bit weird to use `eq` here. Can `$ret` really be empty? If not I would use:

```suggestion
        if (defined $ret && $ret == 0) {
```

Dawei-Pang: Thanks @Martchus! I believe double equal can not work as expected.

Please refer to followed sample when $ret is empty string, the code will print "Hello"
```
my $ret = "";

if (defined $ret && $ret == 0) {
    print "Hello";                          #The code can print Hello -> incorrect
}
```

Let us verify `eq` as following
```
my $ret = 0;

if (defined $ret && $ret eq 0) {
    print 'Hello';                           # print Hello -> correct
}
-----------
my $ret = "0";

if (defined $ret && $ret eq 0) {
    print 'Hello';                           # print Hello -> correct
}
-----------
my $ret = "";

if (defined $ret && $ret eq 0) {
    print 'Hello';                          # Not print Hello -> correct
}
```

Martchus: According to https://metacpan.org/pod/IPC::Run#RETURN-VALUES there are just 2 distinctions to make. All sub processed existed with a 0 exit code  (then `$ret` would be a truthy value) or at least one sub processed existed with a non-0 exit code.

So likely this should just be:

```suggestion
        if (!$ret) {
```

There's no use in making a distinction between `0`, `''` and `undef` because all are falsy values and have the same meaning.

Dawei-Pang: During previous debugging, I observed the $ret is empty string, see the cut part from http://10.67.129.4/tests/62560
```
[2023-09-21T15:48:50.141416+08:00] [debug] [pid:959] IPMI ret: 1
[2023-09-21T15:48:54.210818+08:00] [debug] [pid:959] IPMI ret: 1
[2023-09-21T15:48:58.211160+08:00] [debug] [pid:959] IPMI: Chassis Power is off
[2023-09-21T15:48:58.280219+08:00] [debug] [pid:959] IPMI ret:
[2023-09-21T15:49:02.281003+08:00] [info] [pid:959] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
ipmitool -I lanplus -H 10.67.135.18 -U root -P [masked] chassis power on: Set Chassis Power Control to Up/On failed: Command not supported in present state at /usr/lib/os-autoinst/backend/ipmi.pm line 49.
```

`if (!$ret)` still make `if` condition is incorrectly true when $ret is empty string, please try the followed codes. 
```
my $ret = '';

if (!$ret) {
    print "Hello";                     #The code can print Hello -> incorrect and not expected.
}
```
I believe `==` is used for numerical only, `eq` may more widely compatible for '0', 0, -1, 1 and ''.

Feel free to correct with if any is wrong.

Martchus: If ret was an empty string (or rather *some* falsy value logged like an empty string) this means the command returned with a non-zero exit code. This is the exact same case when 0 was returned. So I really don't get why you don't want to treat them the same. But maybe I'm also misreading the documentation of `IPC::Run` (which I don't find very good/explicit).

Dawei-Pang: For ipmitool, returning `empty` string or non-zero value/string means failure, `0` or 0 means successfully.

Julie-CAO:
I am a little confused about the return code of `$ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);`, why I used to get the "$ret=1" but actually the command is successful.

This is a case when I first found there was some problems. but I am not confident that it will still valid because the test run months ago.

```
[2023-01-18T01:38:52.896917-05:00] [debug] IPMI: Chassis Power Control: Down/Off

debug: running 'ipmitool -I lanplus -H 10.67.135.1 [masked] chassis power status' for 1 times
debug: the $ret of ipmi command is 1
debug: the output of ipmi command is 'Chassis Power is on'

debug: running 'ipmitool -I lanplus -H 10.67.135.1 [masked] chassis power status' for 2 times
debug: the $ret of ipmi command is 1
debug: the output of ipmi command is 'Chassis Power is off'

debug: running 'ipmitool -I lanplus -H 10.67.135.1 [masked] chassis power status' for 3 times
debug: the $ret of ipmi command is 1
debug: the output of ipmi command is 'Chassis Power is off'

[2023-01-18T01:39:14.249539-05:00] [debug] IPMI: Chassis Power is off

debug: running 'ipmitool -I lanplus -H 10.67.135.1 [masked] chassis power on' for 1 times
debug: the $ret of ipmi command is 1
debug: the output of ipmi command is 'Chassis Power Control: Up/On'

[2023-01-18T01:39:18.333867-05:00] [debug] IPMI: Chassis Power Control: Up/On

debug: running 'ipmitool -I lanplus -H 10.67.135.1 [masked] chassis power status' for 1 times
debug: the $ret of ipmi command is 
debug: the output of ipmi command is ''
debug: error is 'Error: Unable to establish IPMI v2 / RMCP+ session'

[2023-01-18T01:39:31.451083-05:00] [info] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  ipmitool -I lanplus -H 10.67.135.1 -U admin -P [masked] chassis power status: Error: Unable to establish IPMI v2 / RMCP+ session at /usr/lib/os-autoinst/backend/ipmi.pm line 50.
```

Martchus:
> Why I used to get "$ret=1" but actually the command was successful.

"1" is a truthy value so according to the documentation (https://metacpan.org/pod/IPC::Run#RETURN-VALUES) that means success. So the return value of `IPC::Run`'s functions is not the exist code of the command. So what you observed makes actually sense. (Although this design decisions of `IPC::Run` is questionable. I personally find that `IPC::Run` is a rather bad abstraction but since we're using it here we need to deal with that.)

Dawei-Pang:
> > Why I used to get "$ret=1" but actually the command was successful.
> 
> "1" is a truthy value so according to the documentation (https://metacpan.org/pod/IPC::Run#RETURN-VALUES) that means success. So the return value of `IPC::Run`'s functions is **not** the exit code of the command. So what you observed makes actually sense. (Although this design decision of `IPC::Run` is questionable. I personally find that `IPC::Run` is a rather bad abstraction but since we're using it here we need to deal with that.)

Thanks @Julie-CAO and @Martchus, so I think we just modify `0` to `1` to fix this issue, am I correct?
```
 -       if ($ret == 0) {
 +      if ($ret == 1) {
```

Martchus:
> am I correct?

I don't think so. If you want to test for successful execution it would just be `if ($ret) {` because the documentation states nowhere that the truthy value has to be 1.

Dawei-Pang:
> > am I correct?
> 
> I don't think so. If you want to test for successful execution it would just be `if ($ret) {` because the documentation states nowhere that the truthy value has to be 1.

Thanks! To follow document more exactly, I would like to use `if ($ret == 1) { ` to avoid any ambiguous detection, for example: 
```
my $ret = 2;
if ($ret) {
    print 'Hello';                           # print Hello -> incorrect
}
```
I will run more widely verification on bare-metal machines to make sure the change work as expected.

perlpunk:
The actual exit value of a command run by `run()` can be found in `$?` (to get the actual value, use `$? >> 8`)

backend/ipmi.pm
okurz: Martchus is right. I also see checking for equality of (pseudo-)boolean values as bad style. The correct way is to treat it as boolean so `if ($ret)` and `if (!ret)` or `… unless $ret`, etc.

For a bit of context, this is who other uses of IPC::Run::run look like in os-autoinst+openQA:

```
okurz@linux-28d7:~/local/os-autoinst/os-autoinst 0 (fix/tidy) $ git grep -A 5 'IPC.*run'
OpenQA/Isotovideo/Utils.pm:        IPC::Run::run(\@{$debugging_tools{$tool}}, \$stdin, \$stdout, \$stderr);
OpenQA/Isotovideo/Utils.pm-    }
OpenQA/Isotovideo/Utils.pm-}
OpenQA/Isotovideo/Utils.pm-
OpenQA/Isotovideo/Utils.pm-1;
--
backend/amt.pm:    $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
backend/amt.pm-    chomp $stdout;
backend/amt.pm-    chomp $stderr;
backend/amt.pm-
backend/amt.pm-    die join(' ', @cmd) . ": $stderr" unless ($ret);
backend/amt.pm-    bmwqemu::diag("AMT: $stdout");
--
backend/ipmi.pm:        $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
backend/ipmi.pm-        if ($ret == 0) {
backend/ipmi.pm-            $self->dell_sleep;
backend/ipmi.pm-            last;
backend/ipmi.pm-        } else {
backend/ipmi.pm-            sleep 4;
okurz@linux-28d7:~/local/os-autoinst/os-autoinst 0 (fix/tidy) $ cd ../openQA
okurz@linux-28d7:~/local/os-autoinst/openQA 0 (feature/spew) $ git grep -A 5 'IPC.*run'
lib/OpenQA/Schema/ResultSet/Jobs.pm:        IPC::Run::run(\@cmd, \undef, \$stdout, \$stderr);
lib/OpenQA/Schema/ResultSet/Jobs.pm-        next if $stderr;
lib/OpenQA/Schema/ResultSet/Jobs.pm-        push(@results, map { $_ =~ s/\..*$//; basename $_} split(/\n/, $stdout));
lib/OpenQA/Schema/ResultSet/Jobs.pm-    }
lib/OpenQA/Schema/ResultSet/Jobs.pm-    return \@results;
lib/OpenQA/Schema/ResultSet/Jobs.pm-}
--
lib/OpenQA/Utils.pm:        my $ipc_run_succeeded = IPC::Run::run($cmd, \$stdin, \$stdout, \$stderr);
lib/OpenQA/Utils.pm-        my $return_code = $?;
lib/OpenQA/Utils.pm-        chomp $stderr;
lib/OpenQA/Utils.pm-        if ($ipc_run_succeeded) {
lib/OpenQA/Utils.pm-            OpenQA::Log->can("log_$stdout_level")->($stdout);
lib/OpenQA/Utils.pm-            OpenQA::Log->can("log_$stderr_level")->($stderr);
--
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm:        IPC::Run::run(\@cmd, \undef, \$stdout, \$stderr);
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm-        return $self->render(json => {error => "Grep failed: $stderr"}, status => 400) if $stderr;
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm-
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm-        my $basename = $distri->basename;
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm-        my $last_filename = '';
lib/OpenQA/WebAPI/Controller/API/V1/Search.pm-        my @lines = split("\n", $stdout);
--
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm:    eval { IPC::Run::run(\@cmd, \$stdin, \$stdout, \$error); $exit_code = $?; };
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-    my $error_from_exception = $@;
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-    $helper->unlock($project);
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-    return $job->finish(0) if (!$exit_code);
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-
--
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm:        eval { IPC::Run::run(\@cmd, \$stdin, \$stdout, \$error); $exit_code = $?; };
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-        my $err = $@;
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-        return ($exit_code, $error // $err);
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-    };
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-
lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm-    my ($exit_code, $error) = $helper->for_every_batch($alias, $sub);
```

so in any other place there is a proper check for just true/false evaluation of the returned value or just catching any exceptions and looking into the actual exit code from `$?`

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst/pull/2364#discussion_r1333362548

backend/ipmi.pm
```diff
@@ -32,7 +32,11 @@ sub ipmitool ($self, $cmd, %args) {
     my @tries = (1 .. $args{tries});
     for (@tries) {
         $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
-        if ($ret == 0) {
+        bmwqemu::diag("IPMI: IPC::Run::run return $ret");
+        # Refer to https://metacpan.org/pod/IPC::Run#RETURN-VALUES
+        # "run() and finish() return TRUE when all subcommands exit with a 0 result code."
+        # "This is the opposite of perl's system() command."
+        if ($ret) {
```
Martchus, line 39: Not sure whether we really need to quote the documentation here. If we would do that on any use of that function or even other functions it would clutter our code a lot. So maybe let's avoid setting a bad precedence here. You could still add a short inline comment:

```suggestion
        if ($ret) {  # command succeeded
```

I would also avoid printing the return value literally as it is just a boolean and not really intended for literal printing. To be able to follow whether the dell-sleep path was entered I'd move the logging into the function itself.

Dawei-Pang: Thanks and agree! The code is updated already.

Martchus (approved):

PR #2366, opened 2023-09-21
WIP: IPMI_HW=dell:Enable dell_sleep for each ipmitool run whenever successfully or failure 

Dawei-Pang:
Currently dell_sleep only works when ipmitool successfully, but not in failure case. Enable dell_sleep in failure case can gain more opportunity in retries.

Issue ticket: https://progress.opensuse.org/issues/136232

Verified URL: http://10.67.129.4/tests/62583#live

The MR code depends on https://github.com/os-autoinst/os-autoinst/pull/2364

### DIFF
--- a/backend/ipmi.pm
+++ b/backend/ipmi.pm
@@ -32,8 +32,8 @@ sub ipmitool ($self, $cmd, %args) {
     my @tries = (1 .. $args{tries});
     for (@tries) {
         $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
+        $self->dell_sleep;
         if ($ret == 0) {
-            $self->dell_sleep;
             last;
         } else {
             sleep 4;


backend/ipmi.pm
```diff
@@ -32,8 +32,8 @@ sub ipmitool ($self, $cmd, %args) {
     my @tries = (1 .. $args{tries});
     for (@tries) {
         $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
+        $self->dell_sleep;
```
okurz, line 35: so that would call `dell_sleep` plus additionally the `sleep 4` some lines below? I don't think this is a good idea but just further time waste. If you want to touch this part of code then I suggest to evaluate if both the sleeps can not be replaced by something better. For that I suggest to look into the git log why those choices have been made in the past

okurz (changes requested):

backend/ipmi.pm
Dawei-Pang: Thanks for your comment! 

The code comes from https://github.com/os-autoinst/os-autoinst/pull/1487
The original code have two issues so far I encountered.
1) unable to handle correctly ipmitool return value $ret, see my MR https://github.com/os-autoinst/os-autoinst/pull/2364
2) dell_sleep only on $ret=0, not work when $ret!=0 or $ret="", the goal of retry is to make the testing forward instead of blocked by kinds of ipmi issue, I hope the failure ipmitool command result need delay as well to gain more opportunity to make testing forward.

In my experiments, when running ipmitool command on DELL machine with small interval, "Command not supported in present state" or other error message is often encountered, especially for power on or off system, more additional interval can help to handle kinds of ipmitool error, instead of one by one.

I think our goal is to make testing forward on DELL machine by some reasonable workaround, not to fix kinds of ipmitool issue or firmware issue.

okurz: @os-autoinst/tools-team further thoughts?

kalikiana: Given that `dell_sleep` is really just `sleep 4` I feel like it would be better to increase the retries e.g. 3 by default on dell.

Dawei-Pang: More triers may cause IPMI entering bad/no response status, giving more time to allow IPMI response will help to mitigate unstable situation and move our testing forward.

kalikiana: Right. So you consider sleeping 8 seconds safer. Is there anything in the output that can be inspected? It seems like it's just ignored when retrying.

Dawei-Pang: Here is an example:
http://10.67.129.4/tests/62565, much retries in 4 second interval,  ipmitool reports "Command not supported in present state", verified more delay for DELL can migrate the issue.
```
[2023-09-21T15:51:21.995170+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:25.995541+08:00] [debug] [pid:1337] IPMI: Selftest: passed
[2023-09-21T15:51:26.060053+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:30.123718+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:34.187828+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:38.188171+08:00] [debug] [pid:1337] IPMI: Chassis Power is off
[2023-09-21T15:51:38.255229+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:42.255628+08:00] [debug] [pid:1337] IPMI: Chassis Power Control: Up/On
[2023-09-21T15:51:45.386089+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:52.951143+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:51:57.019500+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:52:01.019866+08:00] [debug] [pid:1337] IPMI: Chassis Power is off
[2023-09-21T15:52:01.092565+08:00] [debug] [pid:1337] IPMI ret: 
[2023-09-21T15:52:05.093131+08:00] [info] [pid:1337] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  ipmitool -I lanplus -H 10.67.135.18 -U root -P [masked] chassis power on: Set Chassis Power Control to Up/On failed: Command not supported in present state at /usr/lib/os-autoinst/backend/ipmi.pm line 49.
[2023-09-21T15:52:05.156129+08:00] [debug] [pid:1337] IPMI ret: 1
[2023-09-21T15:52:09.156477+08:00] [debug] [pid:1337] IPMI: Chassis Power Control: Down/Off
```

okurz: I am not ready to accept such significant delays in code execution just based on individual tests. I strongly suggest to look into identifying the actual real issue and coming up with a more resilient but quicker way without unconditional longer sleeping. If you *really* want to just sleep longer then I would only be convinced by a more convincing statistical investigation, e.g. run experiments without 100+ jobs without/with your change and show the fail ratio and confidence value.

Dawei-Pang: Thanks Oliver! In my recent experiments, https://github.com/os-autoinst/os-autoinst/pull/2364 can probably fix this timing issue.

Since I need more widely verification, make this PR as WIP.

Dawei-Pang:
Close this MR for conflicts

PR #2367, opened 2023-09-25
Fix Perl::Tidy version

perlpunk:


perlpunk:
The decreased coverage might be from #2364 which was merged with failing tests:
```
sub dell_sleep ($self) {
    return unless ($bmwqemu::vars{IPMI_HW} || '') eq 'dell';
    sleep 4;                            # <------------------------------
}
```

PR #2371, opened 2023-10-02
Error on non-existing WHEELS_DIR

perlpunk:
I had a typo in my WHEELS_DIR and was wondering why there was no error.

Issue: https://progress.opensuse.org/issues/130369

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -158,6 +158,7 @@ sub ensure_valid_vars () {
 
     die "CASEDIR variable not set, unknown test case directory" if !defined $vars{CASEDIR};
     die "No scripts in CASEDIR '$vars{CASEDIR}'\n" unless -e $vars{CASEDIR};
+    die "WHEELS_DIR '$vars{WHEELS_DIR}' does not exist" if defined $vars{WHEELS_DIR} && !-d $vars{WHEELS_DIR};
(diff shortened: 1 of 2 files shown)


bmwqemu.pm
```diff
@@ -158,6 +158,9 @@ sub ensure_valid_vars () {
 
     die "CASEDIR variable not set, unknown test case directory" if !defined $vars{CASEDIR};
     die "No scripts in CASEDIR '$vars{CASEDIR}'\n" unless -e $vars{CASEDIR};
+    if (defined $vars{WHEELS_DIR} && !-d $vars{WHEELS_DIR}) {
+        die "WHEELS_DIR '$vars{WHEELS_DIR}' does not exist";
+    }
```
okurz, line 163: I'd suggest to use post-condition to be consistent with the above two lines.

```suggestion
    die "WHEELS_DIR '$vars{WHEELS_DIR}' does not exist" if defined $vars{WHEELS_DIR} && !-d $vars{WHEELS_DIR};
```

okurz (changes requested):

bmwqemu.pm
perlpunk: done

okurz (approved):

Martchus (approved):

PR #2374, opened 2023-10-04
Improve error message about JSON encoding

perlpunk:
If there is an object in `$result`, we are seeing an error message like this in the log which is not helpful to find the problematic code:

    encountered object 'needle=HASH(0x55579f60d460)', but neither
    allow_blessed, convert_blessed nor allow_tags settings are enabled (or
    TO_JSON/FREEZE method missing) at /usr/lib/os-autoinst/bmwqemu.pm line 274.

Seeing the content can lead to the original broken code (in this case a hash was passed to `record_info` which contained a needle object.

Output will look like this now:
```
# Cannot encode input:                                                                                                 
# $result = {                                                                                                                
#               'foo' => 'bar',                                                                                               
#               'baz' => 42,                                                                                                  
#               'object' => bless( {                           
#                                    'this' => 'cannot be encoded'                                                            
#                                  }, 'Foo' )                                                                                 
#             };                                                                                                              
#  at t/12-bmwqemu.t line 168.                                                                                                
```

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -272,7 +272,11 @@ sub mydie ($cause_of_death) {
 # store the obj as json into the given filename
 sub save_json_file ($result, $fn) {
     open(my $fd, ">", "$fn.new");
-    my $json = Cpanel::JSON::XS->new->pretty->canonical->encode($result);
+    my $json = eval { Cpanel::JSON::XS->new->pretty->canonical->encode($result) };
+    if (my $err = $@) {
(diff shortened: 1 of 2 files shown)


okurz (approved):

kraih (approved):

Martchus (approved):
Looks good but according to CI output you need to run tidy.

perlpunk:
Fixed tidy, and also added `$@` to the error message

PR #2375, opened 2023-10-05
ci: Use PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS

perlpunk:
Set to 0 so by default we get fatal warnings.

This makes it easy to quickly set the variable in the workflow settings when
necessary.

Issue: https://progress.opensuse.org/issues/137105

Tested: https://github.com/perlpunk/os-autoinst/actions/runs/6458335529/job/17531841574#step:3:743
```
7: [15:02:30] ./t/24-myjsonrpc-debug.t ................................ ok      856 ms ( 0.00 usr  0.00 sys +  0.78 cusr  0.08 csys =  0.86 CPU)
7: Use of uninitialized value $x in addition (+) at ./t/24-myjsonrpc.t line 32.
7: # Found 1 warnings but allowing them because PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS is set
7: # Got the following unexpected warnings:
7: #   1: Use of uninitialized value $x in addition (+) at ./t/24-myjsonrpc.t line 32.
7: [15:02:31] ./t/24-myjsonrpc.t ...................................... ok      879 ms ( 0.00 usr  0.00 sys +  0.81 cusr  0.07 csys =  0.88 CPU)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@
 name: ci
 # yamllint disable-line rule:truthy
 on: [push, pull_request]
+env:
+  # Set to 1 to temporarily ignore warnings
+  PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS: 0
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/openqa_fullstack.yml
+++ b/.github/workflows/openqa_fullstack.yml
@@ -9,6 +9,9 @@ on:
       stability_test:
         description: 'Set to 1 to fail if any of the RETRY fails'
         default: 0
+env:
+  # Set to 1 to temporarily ignore warnings
+  PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS: 0
 
 jobs:
   fullstack:
--- a/tools/container_run_ci
+++ b/tools/container_run_ci
@@ -61,7 +61,7 @@ if [[ \"$target\" == \"coverage-codecovbash\" ]]; then
 fi
 cd /opt
 ./tools/install-new-deps.sh
-export CI=1 WITH_COVER_OPTIONS=1 CHECK_GIT_STATUS=1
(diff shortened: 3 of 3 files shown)


perlpunk:
I did set the variable for this repo, but somehow the variable is not visible in the pull request, although it's a variable and not a secret:
https://github.com/os-autoinst/os-autoinst/actions/runs/6421078795/job/17434608466?pr=2375#step:3:1

perlpunk:
Huh, repository vars are not visible in pull requests, and according to [this discussion](https://github.com/orgs/community/discussions/44322) that might even be on purpose, although the reasons are unclear, but so far I don't see an official answer.
So I will hardcode it in the workflow file...

Martchus (approved):

perlpunk:
To make it clear, this PR should pass in its current state; warnings shouldn't be fatal. But it doesn't work. So nothing to approve right now :)

perlpunk:
I stumbled upon a new problem.
zypper reports it has perl-Test-Warnings 0.31, but 0.32 would be available:
https://github.com/perlpunk/os-autoinst/actions/runs/6431924960/job/17465748377#step:4:37
```
Repository     : openSUSE-Tumbleweed-Oss
Name           : perl-Test-Warnings
Version        : 0.32.0-1.1
Arch           : noarch
Vendor         : openSUSE
Installed Size : 53.5 KiB
Installed      : Yes
Status         : out-of-date (version 0.031-1.12 installed)
Source package : perl-Test-Warnings-0.32.0-1.1.src
```
That's why the variable doesn't have an effect even when setting it explicitly in tools/container_run_ci.
Not sure how I would convince the test to install the new version.

perlpunk:
> I stumbled upon a new problem. zypper reports it has perl-Test-Warnings 0.31, but 0.32 would be available: https://github.com/perlpunk/os-autoinst/actions/runs/6431924960/job/17465748377#step:4:37

This is because os-autoinst_dev failed to build for the last 6 days.
Waiting for #2377

perlpunk:
Updated PR to set the variable in the workflow file (to 0, but this way it's just a small change to enable it)

baierjan (approved):

okurz (approved):

PR #2377, opened 2023-10-06
Fix os-autoinst_dev build on current Tumbleweed

okurz:
The 'nobody' user is provided by the distribution and trying to add it
now fails the build as visible in OBS.


baierjan (approved):

perlpunk (approved):

okurz:
I will create a ticket for "codecov/project/fully_covered — 99.99% (target 100.00%) "

PR #2384, opened 2023-10-17
Get the python code to snitch it's version

foursixnine:
Ensure there's evidence of the Python version that is embedded in the perl extension

While looking at Inline-Perl with @josegomezr, we realized that there's no information about the python version the extension is compiled with

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -126,6 +126,8 @@ sub loadtest ($script, %args) {
             sys.path.append("$inc")
             sys.path.append("$script_dir")
             import $name
+            #print("Using python version: %s" % sys.version)
+            perl.diag("Using python version: %s" % sys.version)
             END_OF_PYTHON_CODE
             # Bind the python functions to the perl $name package
             my %info = py_study_package("$name");
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -6,7 +6,7 @@ use Mojo::Base -strict, -signatures;
 use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '5';
-use Test::Output qw(stderr_like combined_from);
+use Test::Output qw(stderr_like combined_from output_like combined_like);
 use Test::Exception;
 use Test::Fatal;
 use Test::Warnings qw(:report_warnings warning);
@@ -316,9 +316,13 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
 };
 
 subtest python => sub {
-    stderr_like {
+    my $python_module_regex = qr{scheduling pythontest tests/pythontest};
+    my $python_version_regex = qr{Using python version};
+
+    combined_like {
         lives_ok { autotest::loadtest('tests/pythontest.py') } 'can load test module'
-    } qr{scheduling pythontest tests/pythontest.py}, 'python pythontest module referenced';
+    } qr{(?|scheduling pythontest tests/pythontest|Using python version)}, 'python pythontest module referenced';
+
     %autotest::tests = ();
     loadtest 'pythontest.py';
     loadtest 'morepython.py';
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -316,9 +316,6 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
 };
 
 subtest python => sub {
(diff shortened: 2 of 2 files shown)


t/08-autotest.t
```diff
@@ -316,9 +316,13 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
...
 subtest python => sub {
-    stderr_like {
+    my $python_module_regex = qr{scheduling pythontest tests/pythontest};
+    my $python_version_regex = qr{Using python version};
+
+    combined_like {
         lives_ok { autotest::loadtest('tests/pythontest.py') } 'can load test module'
-    } qr{scheduling pythontest tests/pythontest.py}, 'python pythontest module referenced';
+    } qr{$python_module_regex+$python_version_regex}, 'python pythontest module referenced';
+
```
josegomezr, line 325: why not running `autotest::loadtest('tests/pythontest.py')` for each assertion?

one for testing that `scheduling pythontest ...` exists. one for the python version.

autotest.pm
```diff
@@ -126,6 +126,7 @@ sub loadtest ($script, %args) {
             sys.path.append("$inc")
             sys.path.append("$script_dir")
             import $name
+            print("Using python version: %s" % sys.version)
```
okurz, line 129: IMHO we can use better approaches than an unconditional print. How about using python logging and a log.debug here? I don't see enough value to make that an unconditional print.

okurz (changes requested):

autotest.pm
josegomezr: `log.X` will still mean an unconditional print, python logger by default (and by no other interaction) has the lowest reporting level.

To be able to "skip" that message you'd be willing to do the undoable: alter the global defaults to skip a message print.

Martchus, line 129: This would print the Python version before each module. It would likely be better to print it just once. And why is this needed anyways?

Martchus:
I guess the title/commit message should use "its" and not "it's".

t/08-autotest.t
foursixnine: @josegomezr that was my initial idea, for some reason this isn't captured properly

autotest.pm
foursixnine: @okurz if you suggest a different place to put the Python version, I'm cool with that, I honestly don't care about the unconditional print, with an unconditional print, we'd would clearly mark that X module is running with Python (see my answer to @Martchus )

Also this has to be perl.diag, so it shows up in the autoinst-log.txt

t/08-autotest.t
josegomezr: I'm no regex expert, I understand the regex match you're using is:

`#(?^:(?^:scheduling pythontest tests/pythontest)+(?^:Using python version))#`

against:

```
[2023-10-17T09:26:39.922954Z] [debug] [pid:2061] scheduling pythontest tests/pythontest.py
```

The first part of the regex  `$python_module_regex` matches, but the second it's just immediately after, it's expecting `$python_version_regex` (or `Using python version`) to be immediately after the module name.

I think you're printing two lines instead. See this regex101:

https://regex101.com/r/jsB2gf/1

t/08-autotest.t
```diff
@@ -316,9 +316,13 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
 };
 
 subtest python => sub {
-    stderr_like {
+    my $python_module_regex = qr{scheduling pythontest tests/pythontest};
+    my $python_version_regex = qr{Using python version};
+
```
foursixnine, line 321:
```suggestion
```

t/08-autotest.t
```diff
@@ -316,9 +316,10 @@ subtest 'load test successfully when CASEDIR is a relative path' => sub {
 };
 
 subtest python => sub {
-    stderr_like {
+    combined_like {
         lives_ok { autotest::loadtest('tests/pythontest.py') } 'can load test module'
-    } qr{scheduling pythontest tests/pythontest.py}, 'python pythontest module referenced';
+    } qr{(?|scheduling pythontest tests/pythontest|Using python version)}, 'python pythontest module referenced';
```
perlpunk, line 321: That regex would even match an empty string.
I think what you would want here is:
```suggestion
    } qr{scheduling pythontest tests/pythontest.*Using python version}s, 'python pythontest module referenced';
```

perlpunk:
I'll try to prepare a PR which prints the python version only when loading a python test for the first time

PR #2385, opened 2023-10-17
os-autoinst-setup-multi-machine: Fix missing dir for gre-up-script

asdil12:
Ticket: https://progress.opensuse.org/issues/133025

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -60,6 +60,7 @@ EOF
   <masquerade/>
 </zone>
 EOF
+    mkdir -p /etc/wicked/scripts/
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
Commit message too long. How about

"os-autoinst-setup-multi-machine: Fix missing dir for gre-up-script"

kalikiana (approved):

Martchus (approved):

okurz (approved):

okurz:
After thinking about this again: The directory should come from wicked. If it does not exist then that's simply because wicked is not installed. In our openQA-in-openQA tests we likely have Tumbleweed with NetworkManager, not wicked. Just saying.

PR #2386, opened 2023-10-18
Revert "os-autoinst-setup-multi-machine: Fix missing dir for gre-up-script"

asdil12:
Reverts os-autoinst/os-autoinst#2385

Seems that this error appeared, because the test was scheduled on a NetworkManager installation.
Reverting as this hides errors.

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -60,7 +60,6 @@ EOF
   <masquerade/>
 </zone>
 EOF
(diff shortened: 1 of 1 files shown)


Martchus (approved):
This change makes sense but it would actually be great if the script would either install wicket as needed or print a clear error message if it isn't installed.

perlpunk (approved):

asdil12:
Added check for wicked.

okurz (approved):

PR #2390, opened 2023-10-20
Enforce in tests that Python does not support `run_args`

josegomezr:
run_args do not flow to python through `Inline::Python` properly, the references do not cross the context boundary rendering it unusable.

Updating tests to disregard any argument in the `run` function to reflect this.

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -334,6 +334,36 @@ subtest python => sub {
     } qr/Traceback.*No module named.*thismoduleshouldnotexist.*/s, 'Python traceback logged';
 };
 
+subtest 'python run_args' => sub {
+    %autotest::tests = ();
+    my $targs = OpenQA::Test::RunArgs->new();
+    $targs->{data} = 23;
+
+    autotest::loadtest('tests/pythontest_with_runargs.py', run_args => $targs);
+
+    loadtest 'pythontest_with_runargs.py';
+    my $p1 = $autotest::tests{'tests-pythontest_with_runargs'};
+
+    stderr_like { $p1->runtest } qr{run_args are not supported in python}, 'Expected output from pythontest_with_runargs.py';
+    is $bmwqemu::vars{PY_SUPPORT_FN}, 'ipsum', 'set_var() using a py function works';
+};
+
+subtest 'python bad run_args' => sub {
+    %autotest::tests = ();
+    my $targs = OpenQA::Test::RunArgs->new();
+    $targs->{data} = 23;
+
+    autotest::loadtest('tests/pythontest_with_bad_runargs.py', run_args => $targs);
+
+    loadtest 'pythontest_with_bad_runargs.py';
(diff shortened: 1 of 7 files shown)


josegomezr:
Extra context:

Trying to use `run_args` (like `run_args.has('key')` errors out like this:
```
# runtest: [2023-10-20T14:03:25.627717+02:00] [info] [pid:32037] ::: basetest::runtest: # Test died: Attribute "OpenQA::Test::RunArgs=HASH(0x5568ef5b4a58)" invalid at (eval 276) line 3.
```

`Inline::Python` does not handle properly the references.

t/08-autotest.t
```diff
@@ -334,6 +334,20 @@ subtest python => sub {
...
+    %autotest::tests = ();
+    my $targs = OpenQA::Test::RunArgs->new();
+    $targs->{data} = 23;
+
+    autotest::loadtest('tests/pythontest_with_runargs.py', run_args => $targs);
+
+    loadtest 'pythontest_with_runargs.py';
+    my $p1 = $autotest::tests{'tests-pythontest_with_runargs'};
+
+    stderr_like { $p1->runtest } qr{run_args are not supported in python}, 'Expected output from pythontest_with_runargs.py';
```
kalikiana, line 347: This is checking that the test test prints the message. I don't see how that tells us that all tests using run_args will fail cleanly.

kalikiana: I would expect that the test aborts because run_args was set before even executing the test code.

josegomezr: It does, https://github.com/os-autoinst/os-autoinst/blob/master/basetest.pm#L349 calls `run` with parameters. If the python function does not handle them it should raise an error on the python side.

I can add a test for that too.

josegomezr: ugh, I just saw my mistake :facepalm:

josegomezr: Is there a way to mark a test module as a python test? Or assert that? I can't find anything like that, all tests look pretty equal 🤔

kalikiana: @josegomezr https://github.com/os-autoinst/os-autoinst/blob/master/autotest.pm#L148 seems explicitly aware of calling Python code already?

josegomezr: @kalikiana Thanks for the pointer!

Check 34bb021, it's using now `$is_python` to bail if `run_args` is provided.

And that way `def run(self):` is completely stable :D

kalikiana: Excellent! Thank you for making the changes. Looks awesome!

kalikiana (approved):
Assuming there's no surprises in CI I'm good with this version (I approved the CI checks, which aren't automatically run for new contributors)

foursixnine (approved):
Its good to know that parametrized tests (I think that was the name), can't be used, even better, that there unit tests for it 🗡️

PR #2392, opened 2023-10-23
Fix rsync command when asset name contains a colon

Martchus:
This fixes the rsync command for using worker cache via svirt backend in
the case the asset name contains a colon. Before this change the following
happens:

```
sshpass … rsync … -av 'sle-12-SP5-s390x-:31152:kgraft-…' 'root@…:/var…'
The source and destination cannot both be remote.
rsync error: syntax or usage error (code 1) at main.c(1436) [Receiver=3.2.3]
```

With this change the first part of the asset name is no longer wrongly
treated as host specification:

```
sshpass … rsync … -av '/var/lib/openqa/factory/hdd/sle-12-SP5-s390x-:31152:kgraft-…' 'root@…:/var/…'
Warning: Permanently added …
sending incremental file list
sle-12-SP5-s390x-:31152:kgraft-…

sent …
total size is …
```

---

Based on https://github.com/os-autoinst/os-autoinst/pull/2391 because the PRs would conflict.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -12,6 +12,7 @@ use File::Temp 'tempfile';
 use File::Basename;
 use File::Which;
 use Mojo::DOM;
+use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
 
 use backend::svirt;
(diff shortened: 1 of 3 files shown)


Martchus:
An example job is https://openqa.suse.de/tests/12618328. I reproduced/tested the problem/changes as mentioned in the commit message on worker34.

okurz (approved):

baierjan (approved):

PR #2399, opened 2023-10-25
experiment: Unify perl linting rules

josegomezr:
Testing the perltidy config from os-autoinst/os-autoinst-common#30


OpenQA/Isotovideo/Utils.pm
```diff
@@ -214,13 +214,21 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
             push @toextract, _store_asset($i, $name, $dir);
         }
         if ($bmwqemu::vars{UEFI} && $bmwqemu::vars{PUBLISH_PFLASH_VARS}) {
-            push(@toextract, {pflash_vars => 1,
+            push(
```
foursixnine, line 217: You broke GH syntax coloring!

image

OpenQA/Qemu/BlockDev.pm
```diff
@@ -186,13 +185,8 @@ sub _from_map ($self, $drives, $snap_conf) {
...
       if @rest > 0;
 
-    return $self->driver($this->{driver})
-      ->file($this->{file})
-      ->node_name($this->{node_name})
-      ->size($this->{size})
-      ->needs_creating($this->{needs_creating})
-      ->implicit($this->{implicit})
-      ->snapshot($snap_conf->get_snapshot(sequence => $this->{snapshot}));
+    return $self->driver($this->{driver})->file($this->{file})->node_name($this->{node_name})->size($this->{size})->needs_creating($this->{needs_creating})
```
foursixnine, line 188: this looks particularily ugly from the linter perspective

backend/qemu.pm
```diff
@@ -231,9 +234,7 @@ sub _wait_for_migrate ($self) {
 
     # Avoid race condition where QEMU allows us to start the VM (set state to
     # running) then tries to transition to post-migarte which fails
-    $self->_wait_while_status_is(qr/paused|finish-migrate/,
-        $max_execution_time - $execution_time,
-        'Timed out waiting for migration to finalize');
+    $self->_wait_while_status_is(qr/paused|finish-migrate/, $max_execution_time - $execution_time, 'Timed out waiting for migration to finalize');
```
foursixnine, line 237: I wonder if this (and other changes) are result of the extended character count to 160?

consoles/video_stream.pm
```diff
@@ -83,17 +83,20 @@ sub connect_remote ($self, $args) {
...
             $self->{dv_timings_last_check} = time;
             bmwqemu::diag "Current DV timings: $timings";
-        } else {
+        }
+        else {
             $self->{dv_timings_supported} = 0;
             bmwqemu::diag "DV timings not supported";
         }
-    } else {
+    }
```
foursixnine, line 98: I wonder how did it make this decision...

foursixnine (commented):
There are some questionable decisions by the linter... but I like where this unification is going!

OpenQA/Qemu/BlockDev.pm
josegomezr: I think these changes come due to `-fnl # don't remove new lines` being deleted.

perlpunk: Yeah, I think we should keep `-fnl`

josegomezr: Bringing back `-fnl` yields: no changes :tada:

Martchus (approved):

.perltidyrc
```diff
@@ -2,11 +2,9 @@
 # # See <https://github.com/houseabsolute/perl-code-tidyall/issues/84>.
 --character-encoding=none
 --no-valign
-# 120 characters would be desired but it is not feasible right now
-#-l=120   # 120 characters per line
```
okurz, line 6: We have 160 in the test distribution, right? I don't think we should try to ensure the same level of quality into both low level software as well as the test distribution so I would keep the stricter line length. Maybe it's possible to override settings with multiple config files?

okurz (changes requested):

.perltidyrc
foursixnine: I disagree, 120 or 160 isn't really the end of the world, and I would much rather keep the same standard for everybody, also there is test code (albeit baseclass and company) in os-autoinst at least.

foursixnine (approved):

.perltidyrc
kalikiana: > I disagree, 120 or 160 isn't really the end of the world, and I would much rather keep the same standard

I second that. An extra config file for a tiny difference doesn't really warrant the added overhead.

kalikiana (approved):

.perltidyrc
josegomezr: @okurz there's no support for extra profile files sadly. See the link below:

https://github.com/perltidy/perltidy/blob/master/lib/Perl/Tidy.pm#L4218-L4223

In any case this is just aligning the contents with the PR in `os-autoinst-commons` (removing two comments and adding a space). The config is exactly the same.

foursixnine: @okurz I guess you still have the objection but no further discussion is needed, right?

okurz: Yeah fine by me. Please update the git commit message, not just "Experiment"

PR #2401, opened 2023-11-13
Disable SVIRT_WORKER_CACHE by default

Martchus:
This feature is a nice optimization but has the caveat mentioned in the documentation and might also otherwise behave slightly different leading to issues like https://progress.opensuse.org/issues/138746. So it is likely best to only enable it in cases when it is safe to use.


kalikiana (approved):
Seems sensible.

PR #2402, opened 2023-11-13
backend/qemu: Preserve case when modifying OVMF file names

dm0-:
I've set up a RHEL-based OpenQA environment, and the default OVMF vars file also points to the code file which prevents UEFI VMs from starting.  It can be worked around by explicitly setting the paths, but this change will fix the default behavior by preserving uppercase file names.  It shouldn't affect other distros since the other paths in `@ovmf_locations` don't contain "code" in their directories.  (Feel free to rewrite this if it's too ugly--I don't know if there is a prettier Perl way to preserve case in substitutions.)

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -671,7 +671,7 @@ sub start_qemu ($self) {
     }
     elsif ($vars->{UEFI} && (($vars->{ARCH} // '') eq 'x86_64')) {
         $vars->{UEFI_PFLASH_CODE} //= find_ovmf;
-        $vars->{UEFI_PFLASH_VARS} //= $vars->{UEFI_PFLASH_CODE} =~ s/code/vars/r;
+        $vars->{UEFI_PFLASH_VARS} //= $vars->{UEFI_PFLASH_CODE} =~ s/code/$&=~tr,CcOoDdEe,VvAaRrSs,r/eir;
         die "No UEFI firmware can be found! Please specify UEFI_PFLASH_CODE/UEFI_PFLASH_VARS or BIOS or UEFI_BIOS or install an appropriate package" unless $vars->{UEFI_PFLASH_CODE};
     }
     if ($vars->{UEFI_PFLASH} || $vars->{BIOS}) {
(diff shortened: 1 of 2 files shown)


Martchus:
I guess extending the tests wouldn't hurt. For me the code itself isn't too ugly to merge.

perlpunk:
I'm wondering in which forms the filenames can occur. The code suggests that `.cOdE` would be possible, and if that's the filename, the other one would be `.vArS`.
That looks a bit weird. Is really every combination of upper and lower case possible? If not, maybe the code can be made nicer.

Martchus:
I guess having mixed casing only partially work would be the most unexpected situation from a user's perspective. So I suggest we either keep it as-is or preserve mixed casings always (as this PR does). A middle ground would be rather strange.

dm0-:
I tried adding a test for it.  It's not quite like the others since QEMU doesn't fail to start with an error here, it just hangs.  I also wasn't sure how to deal with the accumulating invalid options from other tests, so I just saved and restored the settings from the beginning.

Martchus (approved):

kalikiana (approved):

mimi1vx:
@dm0-  please tidy test and then can be merged

dm0-:
Okay, sorry I missed that.  The tidy test looks happy now.

backend/qemu.pm
```diff
@@ -671,7 +671,7 @@ sub start_qemu ($self) {
     }
     elsif ($vars->{UEFI} && (($vars->{ARCH} // '') eq 'x86_64')) {
         $vars->{UEFI_PFLASH_CODE} //= find_ovmf;
-        $vars->{UEFI_PFLASH_VARS} //= $vars->{UEFI_PFLASH_CODE} =~ s/code/vars/r;
+        $vars->{UEFI_PFLASH_VARS} //= $vars->{UEFI_PFLASH_CODE} =~ s/code/$&=~tr,CcOoDdEe,VvAaRrSs,r/eir;
```
okurz, line 674: @perlpunk do you have an idea how this could be written prettier?

okurz (changes requested):
Now CI checks fail. Likely due to a problem uploading the report to codecov in https://github.com/os-autoinst/os-autoinst/actions/runs/6944814175/job/18897249939?pr=2402#step:4:98 . Retriggering.

backend/qemu.pm
perlpunk: no :)

perlpunk (approved):

okurz (approved):

PR #2407, opened 2023-11-23
Add support for NetworkManager in setup multi-machine script

baierjan:
PoC support for NetworkManager in our setup multi-machine.

It might work (or at least the test job is not failing): http://dagr.qam.suse.cz/tests/174

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -21,24 +21,6 @@ configure_firewall() {
     firewall-cmd --permanent --zone="$zone" --add-service=isotovideo
     firewall-cmd --get-default-zone | grep -q "$zone" || firewall-cmd --set-default-zone="$zone"
     systemctl reload firewalld
-}
-
-setup_multi_machine_with_networkmanager() {
-    # Restart NM to load ovs plugin
-    systemctl restart NetworkManager
-    systemctl enable --now openvswitch
-    # Delete any previsous connections
-    nmcli con | grep -oP 'ovs-(interface|port|bridge|slave)-[\w-]+' | xargs nmcli con del
-    # Create bridge, port and interface connection
-    nmcli con add type ovs-bridge con.int "$bridge"
-    nmcli con add type ovs-port con.int "$bridge" con.master "$bridge"
-    nmcli con add type ovs-interface con.int "$bridge" con.master "$bridge" ipv4.method manual ipv4.address 10.0.2.2/15 con.zone "$zone"
-    echo "OS_AUTOINST_USE_BRIDGE=$bridge" > /etc/sysconfig/os-autoinst-openvswitch
-    # Create tap interfaces
-    for i in 0 $(seq 1 "$instances"; seq 64 $((64+instances)); seq 128 $((128+instances))); do
-        nmcli con add type ovs-port con.int "tap$i" con.master "$bridge"
-        nmcli con add type tun mode tap owner "$(id -u _openqa-worker)" group "$(getent group nobody | cut -f3 -d:)" con.int "tap$i" master "tap$i"
-    done
     cat > /etc/firewalld/zones/"$zone".xml <<EOF
 <?xml version="1.0" encoding="utf-8"?>
 <zone target="ACCEPT">
@@ -51,7 +33,11 @@ setup_multi_machine_with_networkmanager() {
   <masquerade/>
 </zone>
 EOF
-    cat > /etc/NetworkManager/dispatcher.d/gre_tunnel_preup.sh <<EOF
+}
+
+create_gre_preup_script() {
+    local location=$1
+    cat > "$location" <<EOF
 #!/bin/sh
 action="\$1"
 bridge="\$2"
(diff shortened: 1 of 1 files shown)


script/os-autoinst-setup-multi-machine
```diff
@@ -17,18 +17,59 @@ configure_firewall() {
...
 }
 
-setup_multi_machine() {
+setup_multi_machine_with_networkmanager() {
+    # Restart NM to load ovs plugin
+    systemctl restart NetworkManager
     systemctl enable --now openvswitch
-    ovs-vsctl list-br | grep -q $bridge && ovs-vsctl del-br $bridge
-    ovs-vsctl add-br $bridge
+    # Delete any previsous connections
```
Martchus, line 30:
```suggestion
    # Delete any previous connections
```

script/os-autoinst-setup-multi-machine
```diff
@@ -17,18 +17,59 @@ configure_firewall() {
...
+<zone target="ACCEPT">
+  <short>"${zone^}"</short>
+  <description>All network connections are accepted.</description>
+  <service name="isotovideo"/>
+  <interface name="$bridge"/>
+  <interface name="ovs-system"/>
+  <interface name="$ethernet"/>
+  <masquerade/>
+</zone>
+EOF
```
Martchus, line 53: It would be nice to avoid duplicating the firewalld part as it looks identical in both cases.

script/os-autoinst-setup-multi-machine
```diff
@@ -17,18 +17,59 @@ configure_firewall() {
...
+    cat > /etc/NetworkManager/dispatcher.d/gre_tunnel_preup.sh <<EOF
+#!/bin/sh
+action="\$1"
+bridge="\$2"
+ovs-vsctl set bridge \$bridge stp_enable=true
+# TODO add entries according to your network topology
+#ovs-vsctl --may-exist add-port \$bridge gre1 -- set interface gre1 type=gre options:remote_ip=<IP address of other host>
+EOF
+    chmod +x /etc/NetworkManager/dispatcher.d/gre_tunnel_preup.sh
```
Martchus, line 62: Also this looks identical apart from the location of the file.

script/os-autoinst-setup-multi-machine
```diff
@@ -17,18 +17,59 @@ configure_firewall() {
...
+    systemctl enable os-autoinst-openvswitch
+    systemctl restart openvswitch os-autoinst-openvswitch
```
Martchus, line 64: These two lines could simply be outside of the function.

script/os-autoinst-setup-multi-machine
```diff
@@ -74,14 +115,22 @@ EOF
 }
 
 main() {
-    if ! readlink /etc/systemd/system/network.service | grep -q wicked.service ; then
-        echo "This script will only work with wicked network daemon"
+    network=$(basename -s.service "$(readlink /etc/systemd/system/network.service)")
+    test "$network" = wicked -o "$network" = NetworkManager || {
+        echo "This script will only work with wicked network daemon or NetworkManager"
```
Martchus, line 120:
```suggestion
        echo "This script only works with wicked network daemon or NetworkManager"
```

Martchus (commented):
Looks generally good and you have a test job :-)

I just have a few remarks mainly about code duplication.

baierjan:
I extracted the common parts, now it should be properly deduplicated.

Martchus (approved):
Thanks, looks quite clean now.

okurz (approved):

script/os-autoinst-setup-multi-machine
```diff
@@ -17,18 +17,58 @@ configure_firewall() {
...
+<zone target="ACCEPT">
+  <short>"${zone^}"</short>
+  <description>All network connections are accepted.</description>
+  <service name="isotovideo"/>
+  <interface name="$bridge"/>
+  <interface name="ovs-system"/>
+  <interface name="$ethernet"/>
+  <masquerade/>
+</zone>
+EOF
```
okurz, line 35: Hi @baierjan as @r-richardson pointed out: firewall-cmd would apply changes and then an XML file is written. Isn't this code block overriding all previous changes of the firewall-cmd?

baierjan: Hm, that's a good point, however the code is not new. If I look at the [previous version](https://github.com/os-autoinst/os-autoinst/blob/56484aa7b7a7932993b978d31b8ba4d66384ce8e/script/os-autoinst-setup-multi-machine) before the change, the order is the same
```
    configure_firewall
    setup_multi_machine
```

So either it was already broken before, or the `firewall-cmd --permanent` writes the changes in some other file. But at least the command `firewall-cmd --permanent --zone="$zone" --add-service=isotovideo` looks to me like it correlate with the `<service name="isotovideo"/>` in the xml, so I would say that definitely some of the commands are indeed overridden.

But let me reiterate once more, this was not really changed by the PR, just shuffled around (in the same order) to avoid duplication. It looks like it wasn't like that all the time and actually you introduced that change in https://github.com/os-autoinst/os-autoinst/pull/2347/commits/a900bef06ab9b5ee447519d96619cfebd825ccbc. So it looks like the XML and the firewall-cmd are maybe both not really the same?

PR #2411, opened 2023-12-01
Increase bootindex of network devices if PXEBOOT is enabled

mimi1vx:
https://progress.opensuse.org/issues/151258

As discussed here https://gitlab.com/qemu-project/qemu/-/issues/1995 qemu firmware should skip empty bootdisk and boot from next bootindex device (PXE) and then after reboot use disk with bootindex=0

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -897,7 +897,7 @@ sub start_qemu ($self) {
             else {
                 die "unknown NICTYPE $vars->{NICTYPE}\n";
             }
-            my $bootorder = $vars->{PXEBOOT} ? "bootindex=$i" : '';
+            my $bootorder = $vars->{PXEBOOT} ? "bootindex=" . ($i + 1) : '';
             sp('device', [qv "$vars->{NICMODEL} netdev=qanet$i mac=$nicmac[$i] $bootorder"]);
         }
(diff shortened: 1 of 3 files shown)


okurz (changes requested):
s/Incerase/Increase/, rest looks fine

okurz (approved):

backend/qemu.pm
```diff
@@ -897,7 +897,7 @@ sub start_qemu ($self) {
             else {
                 die "unknown NICTYPE $vars->{NICTYPE}\n";
             }
-            my $bootorder = $vars->{PXEBOOT} ? "bootindex=$i" : '';
+            my $bootorder = $vars->{PXEBOOT} ? "bootindex=@{[$i+1]}" : '';
```
Martchus, line 900: Does this syntax create an array-ref with one element and then just de-references that to be able to use interpolation? Maybe it would make sense to use a simpler syntax:

```suggestion
            my $bootorder = $vars->{PXEBOOT} ? 'bootindex=' . ($i + 1) : '';
```

Martchus (commented):
It would be nice to have some tests.

Martchus (approved):

PR #2413, opened 2023-12-06
Adjust MTU in `os-autoinst-setup-multi-machine`

Martchus:
* See added comment in diff itself for further explanation
* See https://progress.opensuse.org/issues/151612#note-23 and subsequent notes for further context

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -5,6 +5,12 @@ bridge="${bridge:-"br1"}"
 ethernet="${ethernet:-"br0"}"
 zone="${zone:-"trusted"}"
 
+# Set the MTU of the bridge interface according to https://docs.openvswitch.org/en/latest/faq/issues
+# * See https://progress.opensuse.org/issues/151310
+# * Use default of 1460 (instead of 1450 as the FAQ suggests) because 1460 should be low enough but
+#   is still higher than 1458 which openSUSE MM tests configure within the SUT.
+mtu=${mtu:-1460}
+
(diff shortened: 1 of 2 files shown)


mimi1vx (approved):

okurz (approved):

script/os-autoinst-setup-multi-machine
```diff
@@ -68,9 +68,11 @@ setup_multi_machine_with_networkmanager() {
 setup_multi_machine_with_wicked() {
     ovs-vsctl list-br | grep -q "$bridge" && ovs-vsctl del-br "$bridge"
     ovs-vsctl add-br "$bridge"
-    # see https://progress.opensuse.org/issues/151310
-    # The specific value comes from https://docs.openvswitch.org/en/latest/faq/issues/
-    ovs-vsctl set int $bridge mtu_request=1450
+    # Set the MUT of the bridge interface according to https://docs.openvswitch.org/en/latest/faq/issues
```
perlpunk, line 71: ?
```suggestion
    # Set the MTU of the bridge interface according to https://docs.openvswitch.org/en/latest/faq/issues
```

perlpunk (dismissed):
.

Martchus:
@baierjan That makes sense but I think regardless of whether NetworkManager or wicked is used we need to use `ovs-vsctl set …` here. So maybe I can unify the code paths.

Martchus:
The tests fail with "root is not in the sudoers file." which is definitely unrelated. I'll have a look.

baierjan (approved):

PR #2416, opened 2023-12-16
Make qemu-img compress file using multiple threads

marmarek:
qemu-img can compress using up to 4 threads, enable this functionality. In Qubes OS setup, some jobs upload multi-GB disk images, and reducing compress time from 40min (about the same time as the rest of the job) down to 10min makes a huge difference.

This feature is not clearly documented, but I learned about it on QEMU's IRC, and posted findings here:
https://gitlab.com/qemu-project/qemu/-/issues/80#note_1669835297


okurz (approved):
This sounds great. There might be other openQA worker instances running and compressing with multiple threads might interfere however the load might also distribute enough to be fine either way. Let's see.

Martchus (approved):

PR #2417, opened 2023-12-19
Add more video formats/methods

marmarek:
This extends video support mostly for generalhw backend:
 - support https://github.com/pikvm/ustreamer for accessing /dev/video0 (highly optimized for capturing HDMI input on RPi)
 - support Tight JPEG encoding on VNC - which allows using PiKVM's VNC server (it doesn't support uncompressed VNC stream)

TODO:
- [x] add/update tests
- [ ] figure out how to map shared memory (`map_handle`) only once (I need help here, when I tried to save it in `$self->{ustreamer_map}`, I got "this variable is not memory mapped" later on)
- [x] verify/ensure that Tight JPEG is not preferred over lossless encodings


marmarek:
@SergiiDmytruk you might be interested in the last commit here. Unfortunately, based on my tests, compression distorts an image enough that some needles made on lossless version (ppm via ffmpeg, or VNC to qemu) do not match and need to be remade. But should work fine if using exclusively this version (or if accepting that some needles need to be made twice...)

marmarek:
>     * [ ]  verify/ensure that Tight JPEG is not preferred over lossless encodings

Test with Xvnc says it chooses Tight compression (the "Fill" variant, not JPEG one) even if it's placed as the least preferred. The "Fill" variant is lossless, so it could be okay in this case, but since it should choose "raw" (as listed as more preferred), I don't have faith some other VNC implementation won't send lossy compressed updates.
I'll try adding an option to enable JPEG support only when explicitly requested.

marmarek:
> Test with Xvnc says it chooses Tight compression (the "Fill" variant, not JPEG one) even if it's placed as the least preferred. The "Fill" variant is lossless, so it could be okay in this case, but since it should choose "raw" (as listed as more preferred), I don't have faith some other VNC implementation won't send lossy compressed updates.

This behavior is to be expected, specification says:
> RFB servers should only produce raw encoding unless the client specifically asks for some other encoding type.

consoles/VNC.pm
```diff
@@ -921,6 +934,44 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+    }
+
+    my $read_len = 0;
+    while ($read_len < $data_len) {
+        my $len = read($socket, $data, $data_len - $read_len, $read_len);
+        OpenQA::Exception::VNCProtocolError->throw(error => "short read for zrle data $read_len - $data_len") unless $len;
+        $read_len += $len;
+    }
+    diag sprintf("read $data_len in %fs\n", time - $stime) if (time - $stime > 0.1);
+    my $image = tinycv::from_ppm($data);
```
Martchus, line 971: So `from_ppm` can actually do JPEG decoding? Or where does the decoding happen?

marmarek: `imdecode()`, which is more or less `imread()` [supports various formats, including PPM, PNG and JPEG](https://docs.opencv.org/4.8.0/d4/da8/group__imgcodecs.html#ga288b8b3da0892bd651fce07b3bbd3a56)
> The function determines the type of an image by the content, not by the file extension.

Martchus: Ah, so the function name `image_from_ppm` we have chosen for our bindings is actually a bit misleading.

consoles/VNC.pm
```diff
@@ -428,8 +438,12 @@ sub _server_initialization ($self) {
...
     my @encs = grep { $_->{supported} } @encodings;
 
-    # Prefer the higher-numbered encodings
-    @encs = reverse sort { $a->{num} <=> $b->{num} } @encs;
+    # Prefer the higher-numbered encodings, but not Tight
+    @encs = reverse sort {
+        $a->{name} eq 'Tight' ? -1 :
+          $b->{name} eq 'Tight' ? 1 :
+          $a->{num} <=> $b->{num}
+    } @encs;
```
Martchus, line 446: Maybe we should just specify them in the order we want instead of doing the sorting at runtime.

ppmclibs/tinycv_impl.cc
```diff
@@ -617,6 +617,28 @@ VNCInfo* image_vncinfo(bool do_endian_conversion, bool true_colour,
...
+
+    // avoid an exception
+    if (x < 0 || y < 0 || y + height > a->img.rows || x + width > a->img.cols) {
+        std::cerr << "ERROR - fill_pixel: out of range\n"
+                  << std::endl;
+        return;
+    }
+
+    for (int i = y; i < y + height; i++)
+        for (int j = x; j < x + width; j++)
```
Martchus, line 638: Maybe make the casting explicit:
```suggestion
    for (auto i = static_cast<int>(y); i < y + height; i++)
        for (auto j = static_cast<int>(x); j < x + width; j++)
```

ppmclibs/tinycv_impl.cc
```diff
@@ -645,6 +667,47 @@ void image_map_raw_data_rgb555(Image* a, const unsigned char* data)
...
 }
 
+void image_map_raw_data_uyvy(Image* a, const unsigned char* data)
+{
+    for (int y = 0; y < a->img.rows; y++) {
+        for (int x = 0; x < a->img.cols; x += 2) {
+            int u  = data[(y * a->img.cols + x) * 2 + 0];
+            int y1 = data[(y * a->img.cols + x) * 2 + 1];
+            int v  = data[(y * a->img.cols + x) * 2 + 2];
+            int y2 = data[(y * a->img.cols + x) * 2 + 3];
```
Martchus, line 677: It would probably be better to avoid repeating sub-expressions like that:
```suggestion
            int offset = (y * a->img.cols + x) * 2;
            int u  = data[offset + 0];
            int y1 = data[offset + 1];
            int v  = data[offset + 2];
            int y2 = data[offset + 3];
```
(Maybe you can come up with a better variable name than me.)

t/27-consoles-vnc.t
```diff
@@ -243,6 +243,39 @@ subtest 'update framebuffer' => sub {
...
+    ok $c->update_framebuffer, 'truthy return value for successful pixel update';
+    ($blue, $green, $red) = $c->_framebuffer->get_pixel(14, 44);
+    is $blue, 41, 'pixel data updated in framebuffer (blue)';
+    is $green, 37, 'pixel data updated in framebuffer (green)';
+    is $red, 31, 'pixel data updated in framebuffer (red)';
+
+    # Tight encoding, JpegCompression
+    open(my $f, '<', "t/data/frame1.jpeg") or die "OPENING t/data/frame1.jpeg: $!\n";
+    my $jpeg_data = do { local ($/); <$f> };
+    close($f);
```
Martchus, line 264:
```suggestion
    my $jpeg_data = Mojo::File->slurp('t/data/frame1.jpeg');
```

consoles/VNC.pm
```diff
@@ -921,6 +931,63 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+    my $socket = $self->socket;
+
+    my $pi = $self->_pixinfo;
+
+    my $stime = time;
+    $socket->read(my $data, 1)
+      or OpenQA::Exception::VNCProtocolError->throw(error => 'short read for compression control');
+    my ($compression_control) = unpack('C', $data);
+    # FillCompression
+    if (($compression_control & 0xF0) eq 0x80) {
```
Martchus, line 944:
```suggestion
    if (($compression_control & 0xF0) == 0x80) {
```

consoles/VNC.pm
```diff
@@ -921,6 +931,63 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+        my $data;
+        # special case for TPIXEL, otherwise identical to PIXEL
+        if ($self->_true_colour and $self->_bpp eq 32 and $self->depth eq 24) {
```
Martchus, line 947:
```suggestion
        if ($self->_true_colour and $self->_bpp == 32 and $self->depth == 24) {
```

consoles/VNC.pm
```diff
@@ -921,6 +931,63 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+
+    $data = "";
+    my $read_len = 0;
+    while ($read_len < $data_len) {
+        my $len = $socket->read(my $data_tmp, $data_len - $read_len);
+        OpenQA::Exception::VNCProtocolError->throw(error => "short read for jpeg data $read_len - $data_len") unless $len;
+        $read_len += $len;
+        $data .= $data_tmp;
+    }
+    diag sprintf("read $data_len in %fs\n", time - $stime) if (time - $stime > 0.1);
```
Martchus, line 986: I'd avoid repeated calls to `time`:
```suggestion
    my $elapsed = time - $stime;
    diag "read $data_len in $elapsed\n" if $elapsed > 0.1;
```
But maybe that's just a leftover from debugging anyways?

consoles/video_stream.pm
```diff
@@ -110,6 +119,17 @@ sub _get_ffmpeg_cmd ($self, $url) {
     return \@cmd;
 }
 
+sub _get_ustreamer_cmd ($self, $url, $sink_name) {
+    my @cmd = ('ustreamer', '--device', $url, '-f', '5');
+    # do not produce JPEG stream
+    push(@cmd, ('-c', 'NOOP'));
```
Martchus, line 125: Our normal coding style would be to avoid all the brackets:
```suggestion
    push @cmd, '-c', 'NOOP';
```

consoles/video_stream.pm
```diff
@@ -120,15 +140,35 @@ sub connect_remote_video ($self, $url) {
...
+        my $ffmpeg;
+        $self->{ustreamerpid} = open($ffmpeg, '-|', @$cmd)
+          or die "Failed to start ustreamer for video stream at $url";
+        $self->{ustreamer_pipe} = $ffmpeg;
+        my $timeout = 100;
+        while ($timeout and !-f "/dev/shm/$sink_name") {
+            sleep(0.1);
+            $timeout -= 1;
+        }
+        die "ustreamer startup timeout" if !$timeout;
```
Martchus, line 157:
```suggestion
        die "ustreamer startup timeout" if $timeout < 0;
```

consoles/video_stream.pm
```diff
@@ -181,6 +221,83 @@ sub _receive_frame ($self) {
     return $img;
 }
 
+sub _receive_frame_ustreamer ($self) {
+    my $ustreamer = $self->{ustreamer};
+    $ustreamer or die 'ustreamer is not running. Probably your backend instance could not start or died.';
+
+    flock($self->{ustreamer}, Fcntl::LOCK_EX);
```
Martchus, line 228:
```suggestion
    die 'ustreamer is not running. Probably your backend instance could not start or died.' unless my $ustreamer = $self->{ustreamer};

    flock($ustreamer, Fcntl::LOCK_EX);
```

Martchus (commented):
I guess it looks generally acceptable for a draft. (Of course I haven't tested whether it actually works at all.)

consoles/VNC.pm
marmarek: That's copied from ZRLE handling. But I guess it isn't really important.

consoles/video_stream.pm
marmarek: But the loop exits on $timeout == 0, so if anything then <=

consoles/VNC.pm
```diff
@@ -921,6 +931,63 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+              or OpenQA::Exception::VNCProtocolError->throw(error => 'short read for data len');
+            my ($data_len2) = unpack('C', $data);
+            $data_len |= $data_len2 << 14;
+        }
+    }
+
+    $data = "";
+    my $read_len = 0;
+    while ($read_len < $data_len) {
+        my $len = $socket->read(my $data_tmp, $data_len - $read_len);
```
marmarek, line 981: @Martchus I changed this from `read($socket, $data, $data_len - $read_len, $read_len)`, but I'm not entirely happy about the change (especially the `$data .= $data_tmp` below). But I couldn't figure out how to mock the old variant for tests

Martchus: I suppose you could always just wrap the call to the built-in `read(…)` function in another function and mock that.

marmarek:
> OBS: os-autoinst - openSUSE_Leap_15.5/x86_64 

This is:
> Invalid type 'D' in pack at /home/abuild/rpmbuild/BUILD/os-autoinst-4.6.1703098319.6820b92/consoles/video_stream.pm line 276.

Documentation for `D` in Perl 5.26.1 says:
```
D  A float of long-double precision in native format.
     (Long doubles are available only if your system supports
      long double values _and_ if Perl has been compiled to
      support those.  Raises an exception otherwise.
      Note that there are different long double formats.)
```
The note about compile option is gone in later versions. I guess it isn't enabled in Leap 15.5...

Anyway, what can I do here? I do need to write a value that is `long double` in the header.  Write encoding by hand?

consoles/VNC.pm
```diff
@@ -921,6 +936,65 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
     return $res;
 }
 
+# wrapper to make testing easier
+sub _read_socket ($socket, $data, $data_len, $offset) {
+    return read($socket, $data, $data_len, $offset);
```
okurz, line 941:
```suggestion
    return read($socket, $data, $data_len, $offset);  # uncoverable statement
```

consoles/VNC.pm
```diff
@@ -921,6 +936,65 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
...
+}
+
+sub _receive_tight_encoding ($self, $x, $y, $w, $h) {
```
okurz, line 944: As this file consoles/VNC.pm is getting quite big I would prefer if we can split the code. Do you think this new method or some other code could fit elsewhere and then include here?

consoles/video_stream.pm
```diff
@@ -181,6 +221,102 @@ sub _receive_frame ($self) {
...
+
+        my $img;
+        if ($format eq 'JPEG') {
+            # tinycv::from_ppm in fact handles a bunch of formats, including JPEG
+            $img = tinycv::from_ppm(substr($ustreamer_map, 129, $used));
+        } elsif ($format eq 'UYVY') {
+            $img = tinycv::new($width, $height);
+            $img->map_raw_data_uyvy(substr($ustreamer_map, 129, $used));
+        } else {
+            die "Unsupported video format '$format'";
```
okurz, line 310:
```suggestion
            die "Unsupported video format '$format'";  # uncoverable statement
```

consoles/video_stream.pm
```diff
@@ -202,11 +338,19 @@ sub update_framebuffer ($self) {
...
-        $have_recieved_update = 1;
+    if ($self->{ffmpeg}) {
+        while ($self->_receive_frame_ffmpeg()) {
+            $have_recieved_update = 1;
+        }
+    } elsif ($self->{ustreamer}) {
+        # shared-memory interface "discards" older frames implicitly,
+        # no need to loop
+        if ($self->_receive_frame_ustreamer()) {
+            $have_recieved_update = 1;
```
okurz, line 352: the typo was already there in before but I would appreciate if you can fix that in a separate commit in your PR:

```suggestion
            $have_received_update = 1;
```

t/26-video_stream.t
```diff
@@ -94,6 +98,33 @@ subtest 'connect stream' => sub {
...
+    my $expected;
+    eval {
+        $expected = pack("D", $value);
+    };
+    if ($@) {
+        ok 1, "skip, nothing to compare to";
+    } else {
+        my $actual = $console->_pack_long_double($value);
+        is $actual, $expected, "correctly encoded";
+    }
```
okurz, line 127: IMHO this if/else in a test does not make sense. Tests on the top level should be designed to be reproducibly executed all the time so this should be changed.

t/27-consoles-vnc.t
```diff
@@ -243,6 +246,51 @@ subtest 'update framebuffer' => sub {
...
+    $gray_pixel = pack(CCCC => 31, 37, 41, 0);    # dark prime grey
+    my $of_type_tight_with_coordinates_12_42_4_4 = pack(nnnnNC => 12, 42, 4, 4, 7);
+    my $fill_compression = pack("C", 0x80);
+    $s->set_series(mocked_read => $update_message, $one_rectangle, $of_type_tight_with_coordinates_12_42_4_4, $fill_compression, $gray_pixel);
+    $c->_framebuffer(undef)->width(1024)->height(512)->vncinfo($vncinfo);
+    ok $c->update_framebuffer, 'truthy return value for successful pixel update';
+    ($blue, $green, $red) = $c->_framebuffer->get_pixel(14, 44);
+    is $blue, 41, 'pixel data updated in framebuffer (blue)';
+    is $green, 37, 'pixel data updated in framebuffer (green)';
+    is $red, 31, 'pixel data updated in framebuffer (red)';
```
okurz, line 262: Instead of a comment you can define that a as a title for a nested subtest, e.g.

```suggestion
    subtest 'Tight encoding, FillCompression' => sub {
        $c->_do_endian_conversion($machine_is_big_endian);    # assume server is little-endian
        $vncinfo = tinycv::new_vncinfo($c->_do_endian_conversion, $c->_true_colour, $c->_bpp / 8, 255, 0, 255, 8, 255, 16);
        $gray_pixel = pack(CCCC => 31, 37, 41, 0);    # dark prime grey
        my $of_type_tight_with_coordinates_12_42_4_4 = pack(nnnnNC => 12, 42, 4, 4, 7);
        my $fill_compression = pack("C", 0x80);
        $s->set_series(mocked_read => $update_message, $one_rectangle, $of_type_tight_with_coordinates_12_42_4_4, $fill_compression, $gray_pixel);
        $c->_framebuffer(undef)->width(1024)->height(512)->vncinfo($vncinfo);
        ok $c->update_framebuffer, 'truthy return value for successful pixel update';
        ($blue, $green, $red) = $c->_framebuffer->get_pixel(14, 44);
        is $blue, 41, 'pixel data updated in framebuffer (blue)';
        is $green, 37, 'pixel data updated in framebuffer (green)';
        is $red, 31, 'pixel data updated in framebuffer (red)';
    }
```

t/27-consoles-vnc.t
```diff
@@ -243,6 +246,51 @@ subtest 'update framebuffer' => sub {
...
+
+    # Tight encoding, FillCompression, TPIXEL
```
okurz, line 264: same here, same below

okurz (changes requested):

t/26-video_stream.t
marmarek: Should I hardcode "correct" encoding of example value then? The point of this test is testing if it gets the same result as `pack("D", ...)` (but for that I need a perl version that supports that)

marmarek: Uhm, this actually seems to be broken on arm64...

Can I somehow declare that ustreamer is not supported on openSUSE Leap and use `pack("D")`?

consoles/VNC.pm
marmarek: I guess `_receive_*_encoding` could be placed in another (new) file. But that won't be pretty, especially since `_receive_zrle_encoding` and `_receive_ikvm_encoding` use few more fields from the VNC object, so at least those will need to get whole `$self` to keep it sane.

t/26-video_stream.t
okurz: I guess that should be fine

consoles/VNC.pm
```diff
@@ -921,6 +936,65 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
     return $res;
 }
 
+# wrapper to make testing easier
+sub _read_socket ($socket, $data, $data_len, $offset) {
+    return read($socket, $data, $data_len, $offset);    # uncoverable statement
+}
```
okurz, line 942: As I see the message about missing coverage I guess you could further simplify that to

```suggestion
sub _read_socket ($socket, $data, $data_len, $offset) { read($socket, $data, $data_len, $offset) }  # uncoverable statement
```

though you might still need to add another "uncoverable statement" or multiple to make Devel::Cover happy. Or mock the internal "read" and call `_read_socket` in tests anyway

t/26-video_stream.t
marmarek: How to do that, more specifically how to skip those tests on too old Perl? And where to document this limitation?

okurz: Well, I wouldn't skip tests based on a specific version. Better to only conduct those tests that matter for the actual implementation in os-autoinst. So how about leaving out that specific test for the value encoding?

marmarek: The thing is, if I use `pack("D")`, then all ustreamer tests will fail on the older Perl version

marmarek: To be clear: I'd like to simply drop [the last commit](https://github.com/os-autoinst/os-autoinst/pull/2417/commits/333a02671cbcb59e3564de1752592b0b8e3826fd). And then do something to make it not complain on the old Perl version (accepting it will not work there).

marmarek:
Besides splitting `consoles/VNC.pm` (which I'm not sure how to do to not make it more messy), this should be ready for another review.

t/26-video_stream.t
```diff
@@ -115,8 +132,44 @@ subtest 'frames parsing' => sub {
     $received_img = $console->current_screen();
     is $received_img->similarity($img), 1_000_000, "received correct frame";
     $console->disable_video;
+
```
Martchus, line 135:
```suggestion
```

consoles/video_stream.pm
```diff
@@ -181,6 +221,83 @@ sub _receive_frame ($self) {
...
+        #     long double encode_begin_ts;
+        #     long double encode_end_ts;
+        #     // 112
+        #     long double last_client_ts;
+        #     bool        key_requested;
+        #
+        #     // 192
+        #     uint8_t     data[US_MEMSINK_MAX_DATA];
+        # } us_memsink_shared_s;
+
```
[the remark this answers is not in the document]
marmarek: I'd prefer to keep it, with the offset comments added

consoles/video_stream.pm
```diff
@@ -181,6 +221,83 @@ sub _receive_frame ($self) {
...
+        my ($magic, $version, $id, $used, $width, $height, $format, $stride, $online, $key, $gop) =
+          unpack("QLx4QQIIa4ICCxxI", $ustreamer_map);
+        if ($magic != Math::BigInt->new("0xCAFEBABECAFEBABE")) {
```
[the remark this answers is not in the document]
marmarek: I had a warning without using BigInt...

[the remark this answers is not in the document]
Martchus: Maybe @perlpunk or @kraih know how to deal with this in Perl? In my testing Perl was perfectly able to cope with that literal. What is the warning saying?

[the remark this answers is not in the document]
marmarek: `Hexadecimal number > 0xffffffff non-portable at /opt/consoles/video_stream.pm line 268.`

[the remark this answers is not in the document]
Martchus: Maybe then just 14627333968358193854 works? It wouldn't be hexadecimal then.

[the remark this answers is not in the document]
marmarek: This works...

consoles/video_stream.pm
```diff
@@ -181,6 +214,85 @@ sub _receive_frame ($self) {
...
+        # } us_memsink_shared_s;
+
+        my ($magic, $version, $id, $used, $width, $height, $format, $stride, $online, $key, $gop) =
+          unpack("QLx4QQIIa4ICCxxI", $ustreamer_map);
+        # This is US_MEMSINK_MAGIC, but perl considers hex literals over 32bits non-portable
+        if ($magic != 14627333968358193854) {
+            bmwqemu::diag "Invalid ustreamer magic: $magic";
+            return undef;
+        }
+        die "Unsupported Invalid ustreamer version '$version' (only verrsion 4 supported)" if $version != 4;
```
Martchus, line 267:
```suggestion
        die "Unsupported ustreamer version '$version' (only version 4 supported)" if $version != 4;
```

Martchus (changes requested):
Looks good in general and I assume you tested this also manually. Just one further nitpick from my side.

marmarek:
Yup, both ffmpeg and ustreamer versions work.

marmarek:
CI failure looks unrelated to my changes:
```
[2024-01-09T15:41:08.885Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

consoles/video_stream.pm
```diff
@@ -181,6 +214,85 @@ sub _receive_frame ($self) {
...
+        # } us_memsink_shared_s;
+
+        my ($magic, $version, $id, $used, $width, $height, $format, $stride, $online, $key, $gop) =
+          unpack("QLx4QQIIa4ICCxxI", $ustreamer_map);
+        # This is US_MEMSINK_MAGIC, but perl considers hex literals over 32bits non-portable
+        if ($magic != 14627333968358193854) {
+            bmwqemu::diag "Invalid ustreamer magic: $magic";
+            return undef;
+        }
+        die "Unsupported ustreamer version '$version' (only verrsion 4 supported)" if $version != 4;
```
Martchus, line 267:
```suggestion
        die "Unsupported ustreamer version '$version' (only version 4 supported)" if $version != 4;
```

Martchus (changes requested):
There's one typo left. The failure is definitely unrelated. Maybe pushing again with the typo fixed helps as it seems due to codecov being (hopefully only temporarily) unavailable.

consoles/VNC.pm
okurz: Are you planning to look into https://github.com/os-autoinst/os-autoinst/pull/2417#issuecomment-1863126651 still?

Martchus (approved):

consoles/VNC.pm
Martchus: There's now 100 % patch coverage.

okurz (approved):

SergiiDmytruk:
@marmarek I'm pretty sure this is wrong:

https://github.com/os-autoinst/os-autoinst/blob/dac4aeeca0bf32ea95625729fcf81d6e7fa217ec/consoles/VNC.pm#L939-L940

The tests pass but VNC console doesn't work because `$data` in `_receive_tight_encoding()` is never populated with the data.  Replacing `_read_socket()` with `read()` fixes it.  Second parameter should probably be a pointer or whatever it's called in Perl.

Other than that, it works with PiKVM now, so thanks for implementing it!

marmarek:
> Second parameter should probably be a pointer or whatever it's called in Perl.

Oops, I'll try to figure out how it works in Perl...

marmarek:
(hopefully) fixed in #2442

zagagyka:
Hello. Thank you for these changes.
However, when OpenQA connected to PiKVM via VNC, Tight JPEG compression distorts an image and leads to loss of quality.
It turned out that all needles made in the lossless version (VNC in qemu backend) do not match and require reworking.
Is it possible to add support H.264 decoder in the future? (as in https://github.com/TigerVNC/tigervnc/issues/1187)

marmarek:
> However, when OpenQA connected to PiKVM via VNC, Tight JPEG compression distorts an image and leads to loss of quality.

Yes, this is expected for any lossy compression, as noted above:
> Unfortunately, based on my tests, compression distorts an image enough that some needles made on lossless version (ppm via ffmpeg, or VNC to qemu) do not match and need to be remade. But should work fine if using exclusively this version (or if accepting that some needles need to be made twice...)

I don't think H.264 is lossless either (even if has better quality).

Anyway, I don't have use case for H.264 right now, so I won't be working on this anytime soon.

PR #2419, opened 2023-12-20
Report test module failures via exit codes

josegomezr:
- Introduced a `--exit-status-from-test-results` / `-e` flag to `isotovideo`.

  It'll inspect the test results after a full run and compute the exit code from them.

  If:
  - no test results exist: exit code = 100
  - failed test results results exist: exit code = 101 (failed result => anything different than ok) else
  - exit code = 0

- Documented the flags in the perl pod section.

- Unit tested with an empty distribution & a fail module distribution.

Complements [poo#138416](https://progress.opensuse.org/issues/138416)

### DIFF
--- a/isotovideo
+++ b/isotovideo
@@ -31,22 +31,63 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
 
 Show the current program version and test API version
 
+=item B<-e, -?, --exit-status-from-test-results>
+
+isotovideo will exit with status 1 when a test module fails.
+
 =item B<-h, -?, --help>
 
 Show this help.
 
-=head1 TEST PARAMETER
+=back
+
+=head1 TEST PARAMETERS
 
 All additional command line arguments specified in the C<key=value> format are
 parsed as test parameters which take precedence over the settings in the
 vars.json file. Lower case key names are transformed into upper case
 automatically for convenience.
 
+=head1 EXIT CODES
+
+=over 4
+
+=item B<0 - SUCCESS>
+
+isotovideo successfully executed a complete run. In the case of C<--exit-status-from-test-results> flag additionally all test modules passed or softfailed.
+
+=item B<1 - ERROR>
+
+An error ocurred during the test execution related to a test backend failure.
+
+=item B<100 - NO TEST MODULES SCHEDULED>
+
+No test module was scheduled.
+
+This exit code can only be reported when invoked with the
+C<--exit-status-from-test-results> flag.
+
+=item B<101 - TEST MODULES FAILED>
+
+At least one test module did not end with result "ok" or "softfail".
+
+This exit code can only be reported when invoked with the
+C<--exit-status-from-test-results> flag.
+
+=back
+
 =cut
 
 use Mojo::Base -strict, -signatures;
 use autodie ':all';
 no autodie 'kill';
+
+use constant {
+    EXIT_STATUS_OK => 0,
+    EXIT_STATUS_ERR => 1,
+    EXIT_STATUS_ERR_NO_TESTS => 100,
+    EXIT_STATUS_ERR_FROM_TEST_RESULTS => 101,
+};
 
 # Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
 # Details: https://progress.opensuse.org/issues/90371
@@ -72,14 +113,15 @@ Getopt::Long::Configure("no_ignore_case");
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::Runner;
 use OpenQA::Isotovideo::Utils qw(git_rev_parse spawn_debuggers handle_generated_assets);
+use Mojo::File qw(path);
 
 my %options;
 
 # global exit status
-my $return_code = 1;
+my $RETURN_CODE = EXIT_STATUS_ERR;
 
 sub usage ($r) {
-    $return_code = $r;
+    $RETURN_CODE = $r;
     eval { require Pod::Usage; Pod::Usage::pod2usage($r) };
     die "cannot display help, install perl(Pod::Usage)\n" if $@;
 }
@@ -89,13 +131,30 @@ sub _get_version_string () {
     return "Current version is $thisversion [interface v$OpenQA::Isotovideo::Interface::version]";
 }
 
+sub _exit_code_from_test_results () {
+    my @results = glob(path(bmwqemu::result_dir(), "result-*.json"));
+    return EXIT_STATUS_ERR_NO_TESTS if @results == 0;
+
+    my $did_fail = 0;
+    for my $result_file_path (@results) {
+        my $result_file = path($result_file_path);
+        my $test_result = decode_json($result_file->slurp)->{result};
+        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
+        # If a failure (anything different from ok & softfail) is found, keep it.
+        next if $did_fail;
+
+        $did_fail = $test_result !~ m/^(ok|softfail)$/;
+    }
+    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
+}
+
 sub version () {
     print _get_version_string() . "\n";
-    $return_code = 0;
+    $RETURN_CODE = EXIT_STATUS_OK;
     exit 0;
 }
 
-GetOptions(\%options, 'debug|d', 'workdir=s', 'color=s', 'help|h|?', 'version|v') or usage(1);
+GetOptions(\%options, 'debug|d', 'workdir=s', 'color=s', 'help|h|?', 'version|v', 'exit-status-from-test-results|e') or usage(1);
 usage(0) if $options{help};
 version() if $options{version};
 
@@ -137,28 +196,32 @@ eval {
 
     $runner->handle_commands;
 
-    $return_code = 0;
+    $RETURN_CODE = EXIT_STATUS_OK;
 
     # enter the main loop: process messages from autotest, command server and backend
     $runner->run;
+
+    if ($options{'exit-status-from-test-results'}) {
+        $RETURN_CODE = _exit_code_from_test_results();
+    }
 
     # terminate/kill the command server and let it inform its websocket clients before
     $runner->stop_commands('test execution ended');
 
     if ($runner->testfd) {
         # unusual shutdown
-        $return_code = 1;    # uncoverable statement
+        $RETURN_CODE = EXIT_STATUS_ERR;    # uncoverable statement
         CORE::close $runner->testfd;    # uncoverable statement
         $runner->stop_autotest();    # uncoverable statement
     }
 
-    diag 'isotovideo ' . ($return_code ? 'failed' : 'done');
-    my $clean_shutdown = $runner->handle_shutdown(\$return_code);
+    diag 'isotovideo ' . ($RETURN_CODE ? 'failed' : 'done');
+    my $clean_shutdown = $runner->handle_shutdown(\$RETURN_CODE);
 
     # read calculated variables from backend and tests
     bmwqemu::load_vars();
 
-    $return_code = handle_generated_assets($runner->command_handler, $clean_shutdown) unless $return_code;
+    $RETURN_CODE = handle_generated_assets($runner->command_handler, $clean_shutdown) unless $RETURN_CODE;
 };
 if (my $error = $@) {
     log::fctwarn $error, 'main';
@@ -171,8 +234,8 @@ END {
     $runner and $runner->stop_autotest();
 
     # in case of early exit, e.g. help display
-    $return_code //= 0;
+    $RETURN_CODE //= 0;
 
-    print "$$: EXIT $return_code\n";
-    $? = $return_code;
+    print "$$: EXIT $RETURN_CODE\n";
+    $? = $RETURN_CODE;
 }
--- a/t/14-isotovideo.t
+++ b/t/14-isotovideo.t
@@ -7,8 +7,10 @@ use Test::Warnings ':report_warnings';
 use Test::MockModule;
 use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
-use OpenQA::Test::TimeLimit '20';
 use autodie ':all';
+
+use OpenQA::Test::TimeLimit '20';
+
 use IPC::System::Simple qw(system);
 use Test::Output qw(combined_like combined_from stderr_from);
 use File::Basename;
@@ -238,7 +240,47 @@ subtest 'productdir variable relative/absolute' => sub {
     unlike $log, qr/assert_screen_fail_test/, 'assert screen test not scheduled';
 };
 
+subtest 'exit status from test results' => sub {
+    # dummy isotovideo invocations
+    chdir($pool_dir);
+    path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
+    path('vars.json')->remove if -e 'vars.json';
+    path('serial0')->touch;
+
+    subtest 'no test scheduled' => sub {
+        path('testresults/')->remove_tree;
+        my $log = combined_from { isotovideo(
+                default_opts => '--exit-status-from-test-results backend=null',
+                opts => "casedir=$data_dir/empty_test_distribution", exit_code => 100) };
+    };
+
+    subtest 'tests scheduled' => sub {
+        path('testresults/')->remove_tree;
+        my $module_basename = 'failing_module';
+        my $module = "tests/$module_basename";
+        my $log = combined_from { isotovideo(
+                default_opts => '--exit-status-from-test-results backend=null',
+                opts => "casedir=$data_dir/tests schedule=$module", exit_code => 101) };
+
+        like $log, qr/scheduling $module_basename $module\.pm/, 'module scheduled';
+        like $log, qr/Test result \[testresults\/result\-$module_basename\.json\] fail/, 'failed test module report';
+    };
+
+    subtest 'softfailed module' => sub {
+        path('testresults/')->remove_tree;
+        my $module_basename = 'softfail_module';
+        my $module = "tests/$module_basename";
+        my $log = combined_from { isotovideo(
+                default_opts => '--exit-status-from-test-results backend=null',
+                opts => "casedir=$data_dir/tests schedule=$module", exit_code => 0) };
+
+        like $log, qr/scheduling $module_basename $module\.pm/, 'module scheduled';
+        like $log, qr/Test result \[testresults\/result\-$module_basename\.json\] softfail/, 'soft failed test module report';
+    };
+};
+
 subtest 'upload assets on demand even in failed jobs' => sub {
+    # qemu isotovideo invocation
     chdir($pool_dir);
     path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
     path('vars.json')->remove if -e 'vars.json';
@@ -252,6 +294,7 @@ subtest 'upload assets on demand even in failed jobs' => sub {
 };
 
 subtest 'load test success when casedir and productdir are relative path' => sub {
+    # qemu isotovideo invocation
     chdir($pool_dir);
     path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
     path('vars.json')->remove if -e 'vars.json';
@@ -356,6 +399,7 @@ subtest 'publish assets' => sub {
 done_testing();
 
 END {
+    unlink "./serial0" if -e "./serial0";
     rmtree "$Bin/data/tests/product";
     rmtree "$data_dir/wheels_dir/writer";
     rmtree "$pool_dir/writer";
--- a/t/data/empty_test_distribution/main.pm
+++ b/t/data/empty_test_distribution/main.pm
@@ -1,0 +1,10 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Mojo::Base -strict;
+use testapi;
+use autotest;
+
+# intentionally not loading anything...
+
+1;
--- a/t/data/tests/tests/softfail_module.pm
+++ b/t/data/tests/tests/softfail_module.pm
@@ -1,0 +1,11 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Mojo::Base 'basetest', -signatures;
+use testapi;
+
+sub run ($self) {
+    record_soft_failure('failing me softly with this song');
+}
+
+1;


isotovideo
```diff
@@ -31,23 +31,67 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
 
 Show the current program version and test API version
 
+=item B<-e, -?, --exit-status-from-test-results>
+
+isotovideo will exit with code 1 when a test module fails.
```
okurz, line 36: I suggest to stay consistent with the wording and call it "exit status" so

```suggestion
isotovideo will exit with status 1 when a test module fails.
```

isotovideo
```diff
@@ -31,23 +31,67 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
...
+
+=item B<0 - SUCCESS>
+
+Test ran successfully. Bear in mind that if any of the test modules scheduled
+failed the exit code I<will> be 0.
+
+C<isotovideo> can return non-zero exit codes on test module failures with the
+C<--exit-status-from-test-results> flag.
+
+=item B<0 - ERROR>
```
okurz, line 63: Likely you meant

```suggestion
=item B<1 - ERROR>
```

t/data/empty_test_distribution/main.pm
```diff
@@ -0,0 +1,12 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use strict;
+use warnings;
+
```
okurz, line 6:
```suggestion
use Mojo::Base -strict;
```

I guess you wanted to test me :)

Also see all test modules in this repo like
https://github.com/os-autoinst/os-autoinst/blob/master/t/data/tests/bar/module2.pm

t/data/empty_test_distribution/needles/.gitkeep
```diff
@@ -0,0 +1,12 @@
...
+
+use strict;
+use warnings;
+
+use testapi;
+use autotest;
+
+# intentionally not loading anything...
+
+1;
```
okurz, line 12: I guess you copied that in by mistake. Likely .gitkeep should stay empty

t/14-isotovideo.t
```diff
@@ -238,6 +238,30 @@ subtest 'productdir variable relative/absolute' => sub {
...
+    my $module_basename = 'failing_module';
+    my $module = "tests/$module_basename";
+    my $log = combined_from { isotovideo(
+            default_opts => '--exit-status-from-test-results backend=null',
+            opts => "casedir=$data_dir/tests schedule=$module qemu_no_kvm=1 arch=i386 backend=qemu qemu=i386 novideo=1", exit_code => 101) };
+    
+    like $log, qr/scheduling failing_module $module\.pm/, 'module scheduled';
+    like $log, qr/Test result \[testresults\/result\-$module_basename\.json\] fail/, 'module test result parsed';
+};
+
```
okurz, line 264: Please carefully check the runtime of t/14-isotovideo.t as this test module often increased significantly in runtime which is hindering further development.

isotovideo
```diff
@@ -89,13 +134,33 @@ sub _get_version_string () {
...
+        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
+
+        # If a failure is found, keep the failure
+        $did_fail = $test_result eq 'ok' ? 0 : 1 unless $did_fail;
+    }
+
+    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
+}
+
+
```
okurz, line 156:
```suggestion
    return EXIT_STATUS_ERR_NO_TESTS if scalar @results == 0;
    my $did_fail = 0;
    for my $result_file_path (@results) {
        my $result_file = path(Mojo::Path->new($result_file_path)->canonicalize);
        my $test_result = decode_json($result_file->slurp)->{result};
        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
        # If a failure is found, keep the failure
        $did_fail = $test_result eq 'ok' ? 0 : 1 unless $did_fail;
    }
    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
}

```

isotovideo
```diff
@@ -89,13 +134,33 @@ sub _get_version_string () {
...
+    my $did_fail = 0;
+
+    for my $result_file_path (@results) {
+        my $result_file = path(Mojo::Path->new($result_file_path)->canonicalize);
+        my $test_result = decode_json($result_file->slurp)->{result};
+        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
+
+        # If a failure is found, keep the failure
+        $did_fail = $test_result eq 'ok' ? 0 : 1 unless $did_fail;
```
okurz, line 150: so this does not regard the test flags as defined in http://open.qa/docs/#_test_flags meaning that `isotovideo --exit-status-from-test-results` and openQA can come to different conclusions about the overall result if `ignore_failure` is used. Are we ok with that?

okurz (changes requested):

isotovideo
josegomezr: For now I think is reasonable enough.

I can't determine if `ignore_failure` is persisted in the test result. If so then I can make it aware of it and consider that marked test module as ok instead.

okurz: I guess it could be as simple as `decode_json($result_file->slurp)->{test_flags}` so theoretically you could do

```suggestion
        my $test = decode_json($result_file->slurp);
        $did_fail = $test->{result} eq 'ok' && $test->{test_flags} ne 'ignore_failure' ? 0 : 1 unless $did_fail;
```

(untested) pseudo code of course :)

t/14-isotovideo.t
josegomezr: Should be the fastest it can be, it only schedules the module that fails.

Pointers are welcome cuz I did saw it falling randomly locally.

okurz: > Should be the fastest it can be, it only schedules the module that fails.

I suggest to compare runtime before and after your changes and state the numbers here for verification.

> Well, let us know how it fails because we take stability of tests very serious.

isotovideo
josegomezr: inspecting `basetest::save_test_result` the testflags are not persisted in the test result. Only:

```perl
my $result = {
    details => $self->details(),
    result => $self->result(),
    dents => $self->{dents},
    execution_time => $self->{execution_time},
};
$result->{extra_test_results} = $self->{extra_test_results} if $self->{extra_test_results};
```

t/14-isotovideo.t
josegomezr:
```
os-autoinst on  master [!?] via △ v3.20.4 via 🐪 v5.26.1 via 🐍 v2.7.18 on VPN 🇩🇪 
📅2023-12-20🕙19:57:31 ➜ time prove -I. t/14-isotovideo.t 
t/14-isotovideo.t .. ok    
All tests successful.
Files=1, Tests=13,  9 wallclock secs ( 0.02 usr  0.00 sys +  6.10 cusr  4.40 csys = 10.52 CPU)
Result: PASS

real	0m9.279s
user	0m6.171s
sys	0m4.418s
```

vs this branch:
```
os-autoinst on  add_exit_on_failed_test_module [!?] via △ v3.20.4 via 🐪 v5.26.1 via 🐍 v2.7.18 on VPN 🇩🇪 took 17s632ms 
📅2023-12-20🕙19:56:19 ➜ time prove -I. t/14-isotovideo.t 
t/14-isotovideo.t .. ok    
All tests successful.
Files=1, Tests=15, 18 wallclock secs ( 0.02 usr  0.00 sys +  8.45 cusr  8.84 csys = 17.31 CPU)
Result: PASS

real	0m17.554s
user	0m8.558s
sys	0m8.848s
```

I removed one test and got:

```
real	0m13.323s
user	0m7.367s
sys	0m6.805s
```

- every `isotovideo` invocation costs like 4s (?)
- is `use OpenQA::Test::TimeLimit '20';` a limit per-subtest o per-test-file?

isotovideo
```diff
@@ -31,23 +31,67 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
...
+=item B<100 - NO TEST MODULES SCHEDULED>
+
+No test module was scheduled.
+
+This exit code is reported when invoked with the C<--exit-status-from-test-results> flag.
+
+
+=item B<101 - TEST MODULES FAILED>
+
+A test module failed.
```
okurz, line 76: "failed" is too strict here considering all the different results a test module can take, see http://open.qa/docs/#_jobs . You can avoid that with negation:

```suggestion
At least one test module did not end with result "passed" or "softfailed".
```

Adjust the header accordingly. I don't know if we define the possible result states in os-autoinst or if we actually only have that definition in the context of openQA

isotovideo
```diff
@@ -31,23 +31,67 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
...
+
+This exit code is reported when invoked with the C<--exit-status-from-test-results> flag.
```
okurz, line 78:
```suggestion
This exit code can only be reported when invoked with the C<--exit-status-from-test-results> flag.
```

because it *is* only reported under the circumstance of at least one test module not ended with result "passed" or "softfailed".

isotovideo
josegomezr: Current code suggests: `At least one test module did not end with result "ok"` instead.

isotovideo
```diff
@@ -72,11 +115,12 @@ Getopt::Long::Configure("no_ignore_case");
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::Runner;
 use OpenQA::Isotovideo::Utils qw(git_rev_parse spawn_debuggers handle_generated_assets);
+use Mojo::File qw(path);
 
 my %options;
 
 # global exit status
-my $return_code = 1;
+my $return_code = EXIT_STATUS_ERR;
```
kraih, line 123: Standard Perl convention is for global variables to be all uppercase, like `$RETURN_CODE`. It's not a rule we've been very good about following in openQA though.

isotovideo
```diff
@@ -89,13 +133,27 @@ sub _get_version_string () {
     return "Current version is $thisversion [interface v$OpenQA::Isotovideo::Interface::version]";
 }
 
+sub _exit_code_from_test_results () {
+    my @results = glob(path(bmwqemu::result_dir(), "/result-*.json"));
+    return EXIT_STATUS_ERR_NO_TESTS if scalar @results == 0;
```
kraih, line 138: `scalar` is implicit here with the `==` operator.

isotovideo
```diff
@@ -89,13 +133,27 @@ sub _get_version_string () {
...
+    my $did_fail = 0;
+    for my $result_file_path (@results) {
+        my $result_file = path(Mojo::Path->new($result_file_path)->canonicalize);
```
kraih, line 141: I don't believe you really want to use `Mojo::Path` here. That's for canonicalizing HTTP paths, not file system.

kraih: Maybe `path($result_file_path)->realpath`?

isotovideo
josegomezr: Renamed them to `$RETURN_CODE`

isotovideo
josegomezr: I wanted something to remove duplicated `/` on paths, but in latest commit that doesn't happen anymore.

isotovideo
josegomezr: TIL. Changed in 0df7e176

isotovideo
josegomezr: There's now support for `softfail` on test module result :tada:

t/14-isotovideo.t
josegomezr: on https://github.com/josegomezr/os-autoinst/commits/0df7e176 I marked the tests that run a full `isotovideo` with `# full isotovideo invocation` and placed a rough estimation on what `TimeLimit` should be. My math may be off, but so far it was fine in my local with 0 flickering tests.

t/data/empty_test_distribution/needles/.gitkeep
josegomezr: Guilty as charged.

t/14-isotovideo.t
okurz: >     * every `isotovideo` invocation costs like 4s (?)

that sounds likely. That is aggravated when coverage analysis is active so we have to be very conservative regarding extending tests in t/14-isotovideo.t

How about extracting the exit status evaluation into a library for which you can use a more cost-effective unit-test?

>     * is `use OpenQA::Test::TimeLimit '20';` a limit per-subtest o per-test-file?

per test file

josegomezr: Sure, and we agree that still at least one e2e test must run with the new flag to be able to properly test the feature. Right?

okurz: Not if that makes t/14-isotovideo.t run significantly lower because I see a much higher risk that developers turn to inefficient development practices due to too long test execution runtimes rather than the risk of possibly breaking the new feature without being noticed.

josegomezr: How would you cover that extra branch of the code in the main program?

okurz: It wouldn't so that's why you should keep it as simple as possible in the main program and at best do the branching within one line so that we can keep our complete statement coverage, call it cheating, I don't mind ;)

josegomezr: How's that not an inefficient development practice? :)

okurz: Don't you agree that unit tests should each be executable within a couple of seconds? Otherwise people won't call them until they create a PR and CI runs them which means feedback only after about 0,5h and reviewers were already notified so rather wasteful for a failing "unit" test.

josegomezr: > Don't you agree that unit tests should each be executable within a couple of seconds?

That's irrelevant to the test itself. You should properly test all branches when possible. If it's too slow to your taste, the test and the component should be redesigned so it can run faster.

That can be as easy as splitting E2E tests into it's own file so it can run in parallel (`prove --jobs N`) or rewrite the component so it can have more components tested in isolation.

But the approach of "let's avoid writing test cuz it runs slow" is gonna force you to go into bad practices just suffice that artificial constraint.

> Otherwise people won't call them until they create a PR and CI runs them

The contributing guide literally says:

> Every commit is checked by our CI system as soon as you create a pull request but you **should** run the os-autoinst tests locally. Checkout the build instructions for further details.

Nothing prevents a contributor from running the test.

>  and reviewers were already notified so rather wasteful for a failing "unit" test.

People can still see the code and comment over it, is not like a green test will _reveal_ the hidden source code

kalikiana: > If it's too slow to your taste, the test and the component should be redesigned so it can run faster.

If you don't do it here, chances are nobody will. And I say that having complained about it myself since it can be annoying when you just had a lovely working test.

josegomezr: Check the latest commit (940c7868), speed improved significantly (it didn't need a bump on the timer anymore).

`backend=null`  was being overwritten by qemu even if it's not used.

t/14-isotovideo.t
```diff
@@ -238,7 +246,59 @@ subtest 'productdir variable relative/absolute' => sub {
...
+    like $log, qr/scheduling $module_basename $module\.pm/, 'module scheduled';
+    like $log, qr/Test result \[testresults\/result\-$module_basename\.json\] fail/, 'module test result parsed';
+};
+
+subtest 'exit status from test results: softfailed test' => sub {
+    # dummy isotovideo invocation
+    chdir($pool_dir);
+    path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
+    path('vars.json')->remove if -e 'vars.json';
+    # TODO: backend/baseclass.pm raises if serial0 does not exist.
```
kalikiana, line 288: See https://github.com/os-autoinst/os-autoinst/pull/2419/files#r1435031923

t/14-isotovideo.t
```diff
@@ -238,7 +246,59 @@ subtest 'productdir variable relative/absolute' => sub {
     unlike $log, qr/assert_screen_fail_test/, 'assert screen test not scheduled';
 };
 
+subtest 'exit status from test results: no test scheduled' => sub {
+    # dummy isotovideo invocation
+    chdir($pool_dir);
+    path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
+    path('vars.json')->remove if -e 'vars.json';
+    # TODO: backend/baseclass.pm raises if serial0 does not exist.
```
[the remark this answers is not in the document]
josegomezr: No idea what's the impact, I became aware of the consequence with the failing tests using `backed=null`. It does not happen with `backend=qemu` though.

t/14-isotovideo.t
```diff
@@ -238,7 +246,59 @@ subtest 'productdir variable relative/absolute' => sub {
...
+    path('serial0')->touch();
+    my $module_basename = 'soft_failed_test';
+    my $module = "tests/$module_basename";
+    my $log = combined_from { isotovideo(
+            default_opts => '--exit-status-from-test-results backend=null',
+            opts => "casedir=$data_dir/tests schedule=$module", exit_code => 0) };
+
+    like $log, qr/scheduling $module_basename $module\.pm/, 'module scheduled';
+    like $log, qr/Test result \[testresults\/result\-$module_basename\.json\] softfail/, 'module test result parsed';
```
okurz, line 297: There is quite some duplication in the three subtests. Can you please extract the common parts so that the actual difference that matters becomes more clear? Also please ensure that the test message strings are unique so that from the output it's already clear where code is executed from.

t/data/tests/tests/soft_failed_test.pm
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019-2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Mojo::Base 'basetest', -signatures;
+use testapi;
+
+sub run ($) {
```
okurz, line 7:
```suggestion
sub run () {
```

The `$` would actually not be signatures but prototypes(?) which we would like to avoid

t/data/tests/tests/soft_failed_test.pm
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019-2021 SUSE LLC
```
okurz, line 1:
```suggestion
# Copyright SUSE LLC
```

t/data/tests/tests/soft_failed_test.pm
okurz, line 1: Better rename to be consistent and I would avoid the ambiguous term "test" so how about "softfailing_module.pm"?

t/data/tests/tests/soft_failed_test.pm
josegomezr: Noting, I had copied it from the samples in the tests.

t/data/tests/tests/soft_failed_test.pm
josegomezr: Renamed

t/data/tests/tests/soft_failed_test.pm
josegomezr: Applied

t/data/tests/tests/soft_failed_test.pm
josegomezr: It does need the parameter though, so I rewrote it as `run ($self)`

t/14-isotovideo.t
josegomezr: Refactored as a subtest with subtests, it reuses the initial setup.

t/14-isotovideo.t
```diff
@@ -20,6 +19,15 @@ use Mojo::Util qw(scope_guard);
...
 use OpenQA::Isotovideo::CommandHandler;
 
+# Each full isotovideo invocation (not using _exit_after_schedule=1)
+# is expected to take maximum ~6.67 seconds per run.
+#
+# There are 3 invocations, that approxs to ~20 seconds.
+#
+# Update this number if a new test involving a full isotovideo invocation is
+# added.
+use OpenQA::Test::TimeLimit '20';
```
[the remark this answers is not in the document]
josegomezr: Moved back to line 10.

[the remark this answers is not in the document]
josegomezr: Addenda:  

> imports are taking unusually long so I prefer the previous location.

Is that even a thing? I thought imports were *instant*.

t/14-isotovideo.t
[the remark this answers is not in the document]
kalikiana: It's also created in t/23-baseclass.t, with contents even, so I'd say just drop the FIXME. This seems to be expected.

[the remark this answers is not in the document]
kraih: Maybe `null.pm` is missing a method that should be overloaded?

isotovideo
```diff
@@ -137,28 +199,35 @@ eval {
...
-    $return_code = 0;
+    $RETURN_CODE = EXIT_STATUS_OK;
 
     # enter the main loop: process messages from autotest, command server and backend
     $runner->run;
 
+    if ($options{'exit-status-from-test-results'}) {
+        # Ideally: $RETURN_CODE = $runner->command_handler->test_completed == 0;
+        #          should do it, but autotest returns success if the last
+        #          module ran successfully. Rendering this useless...
```
okurz, line 210: I don't think even "ideally" that "test_completed" should yield a return_code. That method can just tell if all modules had been executed, regardless of their individual result. So I guess you should just ditch this comment

t/14-isotovideo.t
```diff
@@ -7,8 +7,17 @@ use Test::Warnings ':report_warnings';
 use Test::MockModule;
 use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
-use OpenQA::Test::TimeLimit '20';
 use autodie ':all';
+
+# Each full isotovideo invocation (not using _exit_after_schedule=1)
+# is expected to take maximum ~6.67 seconds per run.
+#
+# There are 3 invocations, that approxs to ~20 seconds.
```
okurz, line 15: I don't think such specific comments are a good idea. Comments lie sooner or later. Someone will add another invocation and not change this number.

t/14-isotovideo.t
```diff
@@ -7,8 +7,17 @@ use Test::Warnings ':report_warnings';
...
+#
+# Update this number if a new test involving a full isotovideo invocation is
+# added.
```
okurz, line 18: I would say that's also rather obvious and the comment is redundant considering that we use the TimeLimit in most test modules so the same applies to all: "Update the number if tests need to take longer". IMHO no need for this comment

t/14-isotovideo.t
okurz: That's nice and looks clean. The test message strings are still ambiguous though

t/14-isotovideo.t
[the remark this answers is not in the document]
okurz: Nope. Perl modules may execute code on import.

isotovideo
josegomezr: it's not a return code, but a count of tests executed. But it doesn't behave as I'd expected. In any case it's deleted now.

t/14-isotovideo.t
josegomezr: Agree to strongly disagree but check the latest commit where this is removed.

t/14-isotovideo.t
josegomezr: Check latest commit.

t/14-isotovideo.t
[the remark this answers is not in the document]
kalikiana: The TODO is still there. Please remove it.

josegomezr:
There are two ways here:

- do I remove the comment anole?

- do I make the dummy driver create the

serial0

file?

Please advise.

Liv Dywan ***@***.***> schrieb am Do., 4. Jan. 2024, 18:24:

t/14-isotovideo.t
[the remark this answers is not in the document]
josegomezr: p.s.a: reply-by-email doesn't work on Discussion Threads.

https://github.com/os-autoinst/os-autoinst/pull/2419#issuecomment-1877537132

kalikiana:
> There are two ways here: - do I remove the comment anole? - do I make the dummy driver create the serial0 file? Please advise.

Either is fine by me. You're already deleting the file. Feel free to leave that and just get rid of the comment.

kalikiana (approved):

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/7420191234/job/20191101201?pr=2419#step:6:78
`Package 'perl-DBD-Pg-3.17.0' not found.`
will be fixed by https://github.com/os-autoinst/openQA/pull/5418

isotovideo
```diff
@@ -89,13 +134,30 @@ sub _get_version_string () {
...
+
+    my $did_fail = 0;
+    for my $result_file_path (@results) {
+        my $result_file = path($result_file_path);
+        my $test_result = decode_json($result_file->slurp)->{result};
+        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
+        # If a failure (anything different from ok & softfail) is found, keep it.
+        next if $did_fail;
+
+        $did_fail = !!($test_result !~ m/^(ok|softfail)$/);
```
perlpunk, line 149: This looks pretty like you just want to exit the loop if you get one failure, so why not:
```suggestion
        $did_fail = !!($test_result !~ m/^(ok|softfail)$/) and last;
```

josegomezr:
Rhats

josegomezr:
That's a surprising error, why would openqa (a consumer) block a change to os-autoinst?

isotovideo
josegomezr: Hmm, now I'm torn

On one side I'd like to get the message for all test results.

But on the other side the result is determined by the first failure 🤔

perlpunk: Oh I see. I missed that. But you can remove the `!!`

kalikiana:
> That's a surprising error, why would openqa (a consumer) block a change to os-autoinst?

To spot accidental changes in the ABI. If openQA can't process isotovideo result all kinds of problems will occur.

perlpunk:
The workflow could be adjusted here by just stripping the version numbers from ci-packages.txt I guess, so that it would always install the latest.

t/14-isotovideo.t
```diff
@@ -356,6 +399,7 @@ subtest 'publish assets' => sub {
 done_testing();
 
 END {
+    rmtree "./serial0";
```
perlpunk, line 402: I guess this works, but it looks a bit misleading to me as it is a file and not a directory.
```suggestion
    unlink "./serial0";
```

josegomezr:
Applied suggestions & squashed commits.

isotovideo
josegomezr: Applied to [267cb9d6](https://github.com/os-autoinst/os-autoinst/commit/267cb9d6ba265a92f26a32f760ed587b866251f2)

josegomezr:
Ping (?)

isotovideo
```diff
@@ -31,23 +31,67 @@ ANSI_COLORS_DISABLED or NO_COLOR can be set to any value to disable colors.
...
+=head1 EXIT CODES
+
+=over 4
+
+=item B<0 - SUCCESS>
+
+isotovideo successfully executed a complete run. In the case of C<--exit-status-from-test-results> flag additionally all test modules passed or softfailed.
+
+C<isotovideo> can return non-zero exit codes on test module failures with the
+C<--exit-status-from-test-results> flag.
```
Martchus, line 60: Mentioning this flag again seems a bit redundant:
```suggestion
```

Martchus (approved):

isotovideo
josegomezr: Addressed in [5eb5dfb8](https://github.com/os-autoinst/os-autoinst/pull/2419/commits/5eb5dfb82b6cd3bdfeda8794909d8931a71bf526).

okurz (approved):

PR #2424, opened 2023-12-29
Add UEFI boot support for vmware tests

nanzhg:
Due to UEFI becoming the mainstream system boot option for user OS installation, it should also support UEFI boot in VMware tests. From VMware ESXi 7.0 guest installation, the boot option switches from BIOS to UEFI as default, and it will help to workaround bsc#1216655. For VMware guests, the firmware attribute is set to efi when the guest uses UEFI, and it is not set when using BIOS.

The libvirt XML format described as below.

```
Legacy boot (BIOS)
------
  <os>
    <type>hvm</type>
  </os>

UEFI boot
------
  <os firmware='efi'>
    <type>hvm</type>
  </os>
```

- Related ticket: https://progress.opensuse.org/issues/152917
- Related PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/18378
- Verification run:

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -17,7 +17,7 @@ use Mojo::JSON qw(decode_json);
 
 use backend::svirt;
 
-has [qw(instance name vmm_family vmm_type)];
+has [qw(instance name vmm_family vmm_type vmm_firmware)];
 
 sub new ($class, $testapi_console = undef, $args = {}) {
     my $self = $class->SUPER::new($testapi_console, $args);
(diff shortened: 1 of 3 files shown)


okurz (commented):
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages. Especially please make it obvious in the git commit message subject line *why* this change brakes a benefit

nanzhg:
Updated more details in commit message.

Martchus (approved):

kalikiana (approved):

PR #2425, opened 2024-01-03
Make default click times configurable via variables

marmarek:
Make the default changeable globally via a variable, instead of changing
each invocation of mouse_click/assert_and_click etc. Useful if a backend
introduce some latency already (to not cause "long click" by default),
but also when testing with changed mouse settings.


kalikiana (approved):
Looks sensible.

I was pondering if this can be unit-tested in 03-testapi.t but we don't have a good way to verify sleep's I don't think.

mimi1vx (approved):

PR #2428, opened 2024-01-09
os-autoinst-setup-multi-machine: Fix tap group discrepancy

asdil12:
Always use "nogroup" (even for NetworkManager)


kalikiana (approved):
👍🏾

Martchus (approved):

PR #2429, opened 2024-01-09
script_run: base serial buffer size on command length

AdamWill:
If you use script_run at a serial console to run a command more than 4096 characters long, the wait_serial that waits until the whole command has made it across the serial connection will fail, because it uses the default buffer size (4096), so it will never see the full command. This means there'll be a 90 second timeout before the command is actually run, and a red frame in the results.

This solves the problem by always sizing the buffer to the length of the command, plus 128 extra bytes (for the marker string, and some more for safety).

Issue: https://progress.opensuse.org/issues/153304

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -127,7 +127,7 @@ sub script_run ($self, $cmd, @args) {
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
+            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet}, buffer_size => length($cmd) + 128);
             testapi::type_string("\n");
         }
         else {
--- a/t/05-distribution.t
+++ b/t/05-distribution.t
@@ -12,6 +12,8 @@ use Test::Warnings qw(:all :report_warnings);
 use Test::Fatal;
 use Test::MockModule;
 
+my @wait_serial_calls;
+
 subtest 'script_run' => sub {
     require distribution;
     my $d = distribution->new;
@@ -33,6 +35,30 @@ subtest 'script_run' => sub {
     lives_ok sub { $d->script_run('foo\&') }, 'escaped terminator is accepted';
     lives_ok sub { $d->script_run('foo && bar') }, 'AND operator is accepted';
     lives_ok sub { $d->script_run('foo "x&"') }, 'quoted & is accepted';
+    $mock_testapi->redefine(wait_serial => sub {
+            my $regexp = shift;
+            push @wait_serial_calls, {
+                regexp => $regexp,
+                timeout => 90,
+                expect_not_found => 0,
+                quiet => undef,
+                no_regex => 0,
+                buffer_size => undef,
+                record_output => undef,
+                @_
+            };
+    });
+    $mock_testapi->redefine(is_serial_terminal => 1);
+    $d->script_run('short_command');
+    # script_run calls wait_serial three times when on a serial
+    # console, the call we want to check - which actually types the
+    # command - is the second
+    my $cmdcall = $wait_serial_calls[1];
+    is $cmdcall->{buffer_size}, 141, 'appropriate buffer size used for short command';
+    @wait_serial_calls = ();
+    $d->script_run('long_command' x 512);
+    $cmdcall = $wait_serial_calls[1];
+    is $cmdcall->{buffer_size}, 6272, 'appropriate buffer size used for long command';
 };
 
 done_testing;


distribution.pm
```diff
@@ -127,7 +127,8 @@ sub script_run ($self, $cmd, @args) {
         my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
-            testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
+            my $buffer = length($cmd) > 3584 ? length($cmd) + 512 : 4096;
```
okurz, line 130: So the problem you observed is related to the arbitrary buffer limit selection of 4k but now you introduce multiple new magic numbers. This doesn't feel right. Maybe this needs to be done at another location where the 4k originally come from?

okurz (changes requested):

distribution.pm
AdamWill: there isn't really any magic number. 3584 is just 4096 - 512. the logic is, if we need a buffer longer than 4096, just use the command length plus 512 for padding; otherwise explicitly use the previous implicit default (4096).

I can tweak this a bit if you like, e.g.:

`my $buffer = length($cmd) > (4096 - 512) ? length($cmd) + 512 : undef;`

or something like that, if you think it's better? That makes it clearer where 3584 comes from, and goes back to using the implicit default for other cases...

We could also just always use a buffer length that's 512 longer than the length of the command string:

`my $buffer = length($cmd) + 512;`

but that seemed like more change than this option, so I went with this. I'm open to tweaks.

AdamWill: The 4096 comes from `consoles/serial_screen.pm` `read_until`. I didn't bother figuring out what `$pattern` is there. If it's actually the thing we're looking for here, we *could* just do more or less the same thing we're doing here there instead (if the `buffer_size` arg isn't passed, don't just always default to 4096, instead choose a value based on the size of `$pattern`, or something...)

okurz: I meant magic number as defined as "A unique value with unexplained meaning or multiple occurrences which could (preferably) be replaced with a named constant" from https://en.wikipedia.org/wiki/Magic_number_(programming)

okurz: No, sorry, that is not what I meant. Sorry if I could not make myself clear. If I look into the current codebase of os-autoinst I found the following references to the number 4096

```
backend/baseclass.pm:    while (defined(my $bytes_read = $chan->read($buffer, 4096))) {
consoles/serial_screen.pm:                     buffer_size => 4096, record_output => 0, exclude_match => 0,
consoles/serial_screen.pm:    my $buflen = $nargs{buffer_size} || 4096;
consoles/serial_screen.pm:    my $buflen = $nargs{buffer_size} || 4096;
t/23-baseclass.t:        my $expect_output = "FOO$/" x 4096;
t/23-baseclass.t:                $max //= 4096;
```

Disregarding the occurences in t/, in your change are you referring to backend/baseclass.pm or consoles/serial_screen.pm or both?

@Martchus the number was originally introduced by you. Do you remember what the meaning of "4048" in https://github.com/os-autoinst/os-autoinst/pull/1239/files#diff-ee84fe6541d1e2e971e9e0d4bf0187272df8fa4ac878d033b1c8c600325e9f6dR1240 is? In a later commit this was replaced by 4096 so now we have the above occurences. Should we replace all that by a common properly named constant? Or as alternative different constants with different names and if yes, which ones?

Martchus: I think I was just going for a small power of two (and therefore "fixed" the number to be actually a power of two in commit d81c82015). Ideally the value we choose for this buffer size has no impact on the behavior (except that the memory usage and overhead changes). If that's not the case it would ideally make sense to change the logic to cope with the case when the buffer is exceeded.

I guess I'd simply call the constant `SSH_SERIAL_READ_BUFFER_SIZE` because there's really no further meaning to it.

Martchus: But I think this is actually about the buffer size in the `read_until` function (see e.g. https://github.com/os-autoinst/os-autoinst/blob/master/consoles/serial_screen.pm and https://github.com/os-autoinst/os-autoinst/blame/9e9bfd5f329eeb46ac70c348f9300265d0a63d05/consoles/serial_screen.pm#L225) which is different from SSH serial handling "my" buffer size is about.

AdamWill: yes, that's what I thought too, and explained above.

okurz: Sure. My point was just about ensuring that we know what 4096 actually refers and what it means. I have now created
* https://github.com/os-autoinst/os-autoinst/pull/2434

to simply replace the occurences of 4096 with accordingly named constants. You are welcome to either rebase on my PR and exchange the values in your original implementation or do the "length of command string + buffer" approach. Or as third alternative if you prefer you can also leave everything as is and I will rebase based on your commits and use constants for the values you introduced/changed. Let me know what you prefer.

AdamWill: I think I talked myself into the 'always base the buffer size for this call on the length of the command' approach, so I'll tweak this to do that.

AdamWill:
I don't think the failures have anything to do with the PR (okurz's other PR is failing on the same things).

okurz (approved):

okurz:
> I don't think the failures have anything to do with the PR (okurz's other PR is failing on the same things).

Yes, sorry about that. @b10n1k was experimenting with more CI checks. I have reverted the according commits for now and will rebase here 

@mergifyio rebase

PR #2430, opened 2024-01-10
Ensure D-Bus config is effective after installing os-autoinst-openvswitch

asdil12:
Ticket: https://progress.opensuse.org/issues/138302

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -140,6 +140,7 @@ Group:          Development/Tools/Other
 Requires:       openvswitch
 Requires:       openvswitch-switch
 Requires:       os-autoinst
+Requires(post): dbus-1
 
 %description openvswitch
 This package contains openvswitch support for os-autoinst.
@@ -250,6 +251,9 @@ cd %{__builddir}
 
 %post openvswitch
 %service_add_post os-autoinst-openvswitch.service
+if test $1 -eq 1 ; then
+  %{_bindir}/dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig 2>&1 || :
+fi
 
 %preun openvswitch
 %service_del_preun os-autoinst-openvswitch.service


dist/rpm/os-autoinst.spec
```diff
@@ -250,6 +250,9 @@ cd %{__builddir}
 
 %post openvswitch
 %service_add_post os-autoinst-openvswitch.service
+if test $1 -eq 1 ; then
+  %{_bindir}/dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig 2>&1 || :
```
kalikiana, line 254: So considering the conversation in #5419 will this fail in CI? And where is $1 defined?

okurz: > So considering the conversation in #5419 will this fail in CI?

I guess this can't fail in CI because OBS CI checks only check the main package, not the openvswitch subpackage unlike openQA (see e.g. OBS checks in https://github.com/os-autoinst/openQA/pull/5352) showing OBS: openQA:openQA-client-test and OBS: openQA:openQA-test and such.

> And where is $1 defined?

I guess `$1` is defined in the RPM macro and the caller.

okurz (approved):

dist/rpm/os-autoinst.spec
Martchus: Do we need to depend on `dbus-1-tools` for this line to work?

asdil12: yes - I added the `Requires(post)` depdency

asdil12: `$1` is `1` on install and `2` on update.

asdil12: @Martchus I added a requirement to `dbus-1` which requires `dbus-1-tools`

okurz (approved):

Martchus (approved):

PR #2433, opened 2024-01-11
Test OBS CI against SLE_15_all_Backports

b10n1k:


okurz:
@b10n1k please refrain from merging PRs yourself unless in special circumstances. Especially not when CI checks that you have introduced already fail.

PR #2434, opened 2024-01-11
Use named constant for 4k magic values

okurz:
* Use named constant for serial_screen read buffer size
* Use properly named constant for "ssh serial read buffer size"


perlpunk (approved):

backend/baseclass.pm
```diff
@@ -36,6 +36,7 @@ use OpenQA::NamedIOSelect;
 use constant FULL_SCREEN_SEARCH_FREQUENCY => $ENV{OS_AUTOINST_FULL_SCREEN_SEARCH_FREQUENCY} // 5;
 use constant FULL_UPDATE_REQUEST_FREQUENCY => $ENV{OS_AUTOINST_FULL_UPDATE_REQUEST_FREQUENCY} // 5;
 use constant DEFAULT_FFMPEG_CMD => 'ffmpeg -y -hide_banner -nostats -r 24 -f image2pipe -vcodec ppm -i - -pix_fmt yuv420p';
+use constant SSH_SERIAL_READ_BUFFER_SIZE => 4096;
```
b10n1k, line 39: Would that make sense as a ENV variable?

Martchus (approved):

backend/baseclass.pm
okurz: No. I don't think it's even necessary to overwrite this. If it will become necessary then we can still change it

okurz:
@mergifyio rebase

okurz:
"OBS: os-autoinst - SLE_15_SP5_Backports/x86_64 " is not included anymore in master config so we should ignore that result here.

PR #2438, opened 2024-01-12
Add SLE16 Backports OBS CI integration

b10n1k:
This is the only backports working in building in OBS.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -9,6 +9,12 @@ pr:
   - configure_repositories:
       project: devel:openQA:GitHub
       repositories:
+      - name: SLE_15_SP6_Backports
+        paths:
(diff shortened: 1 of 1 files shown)


.obs/workflows.yml
```diff
@@ -9,6 +9,12 @@ pr:
   - configure_repositories:
       project: devel:openQA:GitHub
       repositories:
+      - name: SLE_15_SP6_Backports
+        paths:
+          - target_project: openSUSE:Backports:SLE-15-SP6:Update
+            target_repository: standard
+        architectures:
+          - x86_64
```
perlpunk, line 17: Why not use the same style as in the other places?
```
 architectures: [ x86_64 ]
```

PR #2439, opened 2024-01-12
Add test package spec for sub-packages

baierjan:
Reference: https://progress.opensuse.org/issues/153340

### DIFF
--- a/dist/rpm/os-autoinst-devel-test.spec
+++ b/dist/rpm/os-autoinst-devel-test.spec
@@ -1,0 +1,25 @@
+%define         short_name os-autoinst-devel
+Name:           %{short_name}-test
+Version:        4.6
+Release:        0
+Summary:        Test package for %{short_name}
+License:        GPL-2.0-or-later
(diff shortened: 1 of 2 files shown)


okurz (approved):

dist/rpm/os-autoinst-devel-test.spec
```diff
@@ -0,0 +1,25 @@
...
+Summary:        Test package for %{short_name}
+License:        GPL-2.0-or-later
+BuildRequires:  %{short_name} == %{version}
+ExcludeArch:    %{ix86}
+
+%description
+.
+
+%prep
+# workaround to prevent post/install failing assuming this file for whatever
```
kalikiana, line 14: assuming this file... the sentence seems incomplete? what does this mean?

okurz: just leave it. We had that already in other places :)

Martchus (approved):

PR #2441, opened 2024-01-15
Test the changed version, not the released one

baierjan:
Test sub-package is pulling any version, not the one currently in build.
That makes tests in PR useless. This should remedy the issue.

Reference: https://progress.opensuse.org/issues/153340

### DIFF
--- a/dist/rpm/os-autoinst-openvswitch-test.spec
+++ b/dist/rpm/os-autoinst-openvswitch-test.spec
@@ -5,7 +5,7 @@ Version:        4.6
 Release:        0
 Summary:        test package for %{short_name}
 License:        GPL-2.0-or-later
-BuildRequires:  %{short_name}
+BuildRequires:  %{short_name} == %{version}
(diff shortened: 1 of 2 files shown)


okurz (approved):

dist/rpm/os-autoinst-openvswitch-test.spec
```diff
@@ -5,7 +5,7 @@ Version:        4.6
 Release:        0
 Summary:        test package for %{short_name}
 License:        GPL-2.0-or-later
-BuildRequires:  %{short_name}
+BuildRequires:  %{short_name} == %{version}
```
perlpunk, line 8: For some reason this is commented out in 
https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst-test.spec#L8
```
#BuildRequires:  %{short_name} == %{version}
BuildRequires:  %{short_name}
```
@okurz since you added this - any idea why?

Martchus (approved):

PR #2442, opened 2024-01-17
Fix _read_socket mockup

marmarek:
It needs to pass arguments to read() by refenrece, so local variables
from signature are not appropriate, as those are copies.

This time add a test for it.


SergiiDmytruk:
I can confirm this version of `_read_socket()` works, despite CI not liking it :)

marmarek:
And now test fail because signatures are expected everywhere... How can I pass a scalar by reference while using signatures? IIUC I cannot use `$_` then

marmarek:
Ok, I got a version CI is happy about :)

okurz (approved):

consoles/VNC.pm
```diff
@@ -937,7 +937,9 @@ sub _receive_zrle_encoding ($self, $x, $y, $w, $h) {
 }
 
 # wrapper to make testing easier
-sub _read_socket ($socket, $data, $data_len, $offset) { return read($socket, $data, $data_len, $offset); }    # uncoverable statement
+sub _read_socket ($socket, $data, $data_len, $offset) {
+    return read($socket, ${$data}, $data_len, $offset);
+}
```
Martchus, line 942:
```suggestion
sub _read_socket ($socket, $data, $data_len, $offset) { read($socket, $$data, $data_len, $offset) }
```

Martchus (approved):

marmarek:
codecov upload failed...

PR #2445, opened 2024-01-18
Add yamltidy config

perlpunk:
And change .yamllint config to always indent block sequences

I would like to run yamltidy before committing YAML files, and I think we profit from a common indentation.

### DIFF
--- a/.github/workflows/ci_extended.yml
+++ b/.github/workflows/ci_extended.yml
@@ -7,7 +7,7 @@ on:
       - master
   pull_request:
     paths:
-      - 'container/**'
+      - container/**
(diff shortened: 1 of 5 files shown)


perlpunk:
Currently there is no documentation about style checks in this repo at all.
I can add it to the README.

okurz (approved):
LGTM. How does that align with inherited style checks from os-autoinst-common?

perlpunk:
> LGTM. How does that align with inherited style checks from os-autoinst-common?

Currently they are not inherited.

Martchus (approved):

PR #2448, opened 2024-01-25
Add editorconfig and update files according to shfmt

b10n1k:


Martchus (approved):

tools/container_run_ci
```diff
@@ -28,26 +28,35 @@ set -o pipefail
...
+    -t | --target)
+        target="$2"
+        shift 2
+        ;;
+    --)
+        shift
+        break
+        ;;
+    *) break ;;
+    esac
```
okurz, line 46: can we prevent this IMHO excessive line breaking here?

okurz (changes requested):

tools/container_run_ci
perlpunk: I also don't like it, but I can't find an option for that.
https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd
```
[*.sh]
# like -i=4
indent_style = space
indent_size = 4

# --language-variant
shell_variant      = posix
binary_next_line   = true
# --case-indent
switch_case_indent = true
space_redirects    = true
keep_padding       = true
# --func-next-line
function_next_line = true
```
But I can live with it

okurz (approved):

perlpunk (approved):

okurz:
Next step is to call shfmt automatically as part of CI checks

PR #2450, opened 2024-01-29
Enable shfmt in CI

b10n1k:
https://progress.opensuse.org/issues/153793#change-756619

### DIFF
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -65,6 +65,18 @@ else ()
     message(STATUS "Set PERLCRITIC_PATH to the path of the perlcritic executable to enable Perl syntax/style checks.")
 endif ()
 
+# add test for bash script syntax
+find_program(SH_PATH shfmt)
+if (SH_PATH)
+    add_test(
+        NAME test-local-bash-syntax
+	 COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts" "${SH_PATH}" "${CMAKE_SOURCE_DIR}"
+	 WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+    )
+else ()
+    message(STATUS "Set SH_PATH to the path of the shfmt executable to enable bash script syntax checks.")
+endif ()
+
 # add spell checking for test API documentation
 find_program(PODSPELL_PATH podspell)
 find_program(SPELL_PATH spell)
--- a/tools/check-bash-scripts
+++ b/tools/check-bash-scripts
@@ -1,0 +1,4 @@
+#!/bin/sh -e
+executable=$1
+path=$2
+"${executable:-shfmt}" -d "${path}"


tools/check-bash-scripts
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh -e
+"${1:-shfmt}" -d .
```
okurz, line 2: I don't understand why you create this wrapper and your git commit message in https://github.com/os-autoinst/os-autoinst/pull/2450/commits/d20aa3a8e931bb2ca65030b7043e2b00a0df7b46 does not tell more about that context. Please put more details in the git commit message and remove that "Signed-off-by" line, that's redundant.

For this specific change here I assume you copied "test-local-perl-style" but if you take a look further like in "find_program(SPELL_PATH spell)" how an external command is used

okurz (changes requested):

tools/check-bash-scripts
perlpunk: To me it looks like a copy from the shellcheck check, and every check seems to have a wrapper, so I guess I would have created one myself.

cmake/test-targets.cmake
```diff
@@ -65,6 +65,18 @@ else ()
     message(STATUS "Set PERLCRITIC_PATH to the path of the perlcritic executable to enable Perl syntax/style checks.")
 endif ()
 
+# add test for bash script syntax
+find_program(SH_PATH shfmt)
+if (SH_PATH)
+    add_test(
+        NAME test-local-bash-syntax
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts"
```
perlpunk, line 73: I guess you need to pass the path as an argument:
```suggestion
        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts" "${SH_PATH}"
```

tools/check-bash-scripts
okurz: @Martchus you might be able to give some insight on this design with lots of wrappers here

cmake/test-targets.cmake
b10n1k: i just saw this but the code has already changed. I am gonna resolve it and feel free to comment in the updated commit.

tools/check-bash-scripts
Martchus: I guess the wrappers were because someone wanted a simple shell script that can be invoked separately. I personally don't think it is that useful and probably would invoke shfmt just from CMake (so `ctest` can be used to invoke the test). However, considering we already have one wrapper and we have people who might not be that experienced with CMake I guess creating one more wrapper is not a big deal.

Note that in CMake-based build scripts it is a best practice to make paths of executables overridable; that's why `find_program(SH_PATH shfmt)` is used here and that's why we pass that path to the wrapper script. I guess we shouldn't needlessly deviate from that practice.

By the way, the corresponding CMake code (which is actually the lengthy part) could likely be reduced by using a loop but it would be at the cost of readability.

cmake/test-targets.cmake
```diff
@@ -65,6 +65,17 @@ else ()
     message(STATUS "Set PERLCRITIC_PATH to the path of the perlcritic executable to enable Perl syntax/style checks.")
 endif ()
 
+# add test for bash script syntax
+find_program(SH_PATH shfmt)
+if (SH_PATH)
+    add_test(
+        NAME test-local-bash-syntax
+        COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts ${CMAKE_SOURCE_DIR}"
+    )
```
Martchus, line 74: Please make actually use of `SH_PATH` so users can specify a different path by invoking CMake like `-DSH_PATH=/custom/shfmt`:

```suggestion
    add_test(
        NAME test-local-bash-syntax
        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts" "${SH_PATH}" "${SHELLCHECK_PATH}"
        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
    )
```

Note that `${CMAKE_SOURCE_DIR}` also lacked quoting on shell-level and it is generally a bad practice to assume `sh` exists in CMake (for the sake of portability). So one would usually use https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-a-command-line-tool instead (example: https://github.com/Martchus/syncthingtray/blob/b55480548752fb77f5d244f7ab4dc34cd7d5d9a2/syncthing/CMakeLists.txt#L185). This also allows passing environment variables which would be an alternative way of passing the executable's path if you don't like using the first argument for that.

tools/check-bash-scripts
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
+path=$1
+shfmt -d "${path}"
```
Martchus, line 3: In accordance with my other suggestion:

```suggestion
executable=$1 path=$2
"${executable:-shfmt}" -d "${path}"
```

Martchus (changes requested):

tools/check-bash-scripts
okurz: @Martchus why do we need this wrapper at all?

cmake/test-targets.cmake
perlpunk: Why would you pass `"${SH_PATH}" "${SHELLCHECK_PATH}"` here?
The shellcheck target is some lines above...

Martchus: That was a C&P mistake. I amended my suggestion.

tools/check-bash-scripts
perlpunk: the `-d` option has two effects: 1. it shows the diff 2. the exit code will be non-zero if there is a diff
The directory parameter is always needed.

okurz (approved):

cmake/test-targets.cmake
```diff
@@ -65,6 +65,18 @@ else ()
     message(STATUS "Set PERLCRITIC_PATH to the path of the perlcritic executable to enable Perl syntax/style checks.")
 endif ()
 
+# add test for bash script syntax
+find_program(SH_PATH shfmt)
+if (SH_PATH)
+    add_test(
+        NAME test-local-bash-syntax
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts" "${SH_PATH}" "${SHELLCHECK_PATH}"
```
perlpunk, line 73: SHELLCHECK_PATH?

perlpunk: https://github.com/os-autoinst/os-autoinst/actions/runs/7712505507/job/21020210377?pr=2450#step:3:595
```
6: test-local-bash-syntax
6: Test command: /opt/tools/check-bash-scripts "/usr/bin/shfmt" "/usr/bin/shellcheck"
6: Working Directory: /opt
6: Test timeout computed to be: 10000000
6: /usr/bin/shellcheck:1:12: invalid UTF-8 encoding
6/8 Test #6: test-local-bash-syntax ...........***Failed    0.02 sec
```

Martchus (changes requested):

perlpunk (approved):

Martchus (approved):

PR #2451, opened 2024-01-31
Use config files from os-autoinst-common

b10n1k:
Replace .perlcriticrc and .yamllint with the ones coming from os-autoinst-common.


okurz (approved):

kalikiana (approved):

perlpunk:
btw, I think the tests are only passing here because we use `perlcritic --gentle` which is the same as `perlcritic --severity 5`.

PR #2452, opened 2024-01-31
Ensure Git URL variables are pointing to actual remote despite caching

Martchus:
When using caching via `GIT_CACHE_DIR` the function `git_remote_url` so far returns only a local path and not the actual remote URL anymore. With this change the actual remote URL is looked up so the variables `TEST_GIT_URL` and `NEEDLES_GIT_URL` (which are used by external tooling) are still populated as before.

Related ticket: https://progress.opensuse.org/issues/154156

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -45,14 +45,15 @@ sub calculate_git_hash ($git_repo_dir) {
     return $git_hash;
 }
 
-sub git_remote_url ($git_repo_dir) {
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

okurz (commented):
Ok but is it still visible from variables or autoinst-log.txt that the caching was effective?

Martchus:
It is visible from the logs but not from variables - which I actually consider a good thing (cache = hidden).

okurz (approved):

PR #2454, opened 2024-02-06
Update tidy version to 20240202

b10n1k:


Martchus (approved):

perlpunk:
If the tests are not failing, there is something wrong.
`make update-deps` should be called during the tests and complain that there are changes.
I need to have a look.
I'm also trying to push something to os-autoinst-common, but nothing I'm pushing starts a github workflow. I'm out of ideas right now.

perlpunk:
I know why it doesn't fail :(
https://github.com/os-autoinst/os-autoinst-common/blob/master/tools/update-deps#L41
```
update_cpanfile($modules_by_target) if $cpanfile;
```
The cpanfile now needs to be added on the commandline explicitly. I will do a PR

perlpunk:
#2455

perlpunk (changes requested):
We don't yet have Perl::Tidy 20240202 in openSUSE yet

okurz (changes requested):
Actually we should only ever update when there is a new package in openSUSE Tumbleweed+Leap which is not the case right now. IMHO this should be rejected.

b10n1k:
when i run `make update-deps` i get 
```
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -16,7 +16,7 @@ ENV NAME openQA test environment
 ENV LANG en_US.UTF-8
 
 # AUTODEPS START
-# This part is autogenerated by tools/update-deps from dependencies.yaml
+# This part is autogenerated by ../../../../../../../home/iob/os-autoinst-distri-opensuse/os-autoinst/tools/update-deps from dependencies.yaml```
is that correct to submit?

b10n1k:
> We don't yet have Perl::Tidy 20240202 in openSUSE yet

closing based on that

PR #2456, opened 2024-02-06
Require Module::CPANfile

perlpunk:
tools/tidyall is using it

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -45,6 +45,7 @@ devel_requires:
   perl(Code::TidyAll):
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
+  perl(Module::CPANfile):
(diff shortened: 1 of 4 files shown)


okurz (approved):

Martchus (approved):

dependencies.yaml
```diff
@@ -45,6 +45,7 @@ devel_requires:
   perl(Code::TidyAll):
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
+  perl(Module::CPANfile):
```
foursixnine, line 48: I wonder if we can have a single file where dependencies are updated/generated and we could leave that to a single script for i.e generating the container image on obs, or when building the rpms...

(last I checked, long time ago, all of this was manual)

PR #2460, opened 2024-02-21
Add ipmitool dependencies for openqa-worker-container

nicksinger:
Similar to what we do already for s390 we should also provide dependencies for IPMI SUTs.

See https://progress.opensuse.org/issues/153706


dist/rpm/os-autoinst.spec
```diff
@@ -184,6 +185,15 @@ Requires:       %s390_zvm_requires
...
 Convenience package providing os-autoinst + s390 worker jumphost dependencies.
 
+%package ipmi-deps
+Summary:        Convenience package providing os-autoinst + ipmi jumphost deps
+Group:          Development/Tools/Other
+Requires:       os-autoinst
+Requires:       %ipmi_requires
+
+%description ipmi-deps
+Convenience package providing os-autoinst + s390 worker jumphost dependencies.
```
okurz, line 195:
```suggestion
Convenience package providing os-autoinst + ipmi jumphost dependencies.
```

and add at the end next to s390 a files section.
See
https://github.com/os-autoinst/os-autoinst/pull/2460/files#diff-2bf7294cf30e1f4bda6077c3c165bfbe7b6b67fed2d3afc7e4fe02664ae1f060R321

okurz (changes requested):

dist/rpm/os-autoinst.spec
okurz:
```suggestion
Convenience package providing os-autoinst + ipmi jumphost dependencies.
```

is still needed

Martchus (approved):

okurz (approved):

PR #2462, opened 2024-02-23
t: Split author tests and build test

perlpunk:
Issue: https://progress.opensuse.org/issues/155062

Add test-local-author-perl target and workflow for test-local 

This way we have some of the static tests in its own workflow. Faster and clearer feedback from github.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -287,6 +287,16 @@ Run individual tests by specifying them explicitly:
 make test-perl-testsuite TESTS="t/15-logging.t t/28-signalblocker.t"
 ----
 
+Run perl author tests:
+----
+make test-local-author-perl
+----
+
+Run all author tests:
+----
+make test-local
+----
+
 Notice that the user needs to include the test directory for each test (either t for normal or
 xt for developer-centric tests) when specifying individual tests.
 
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -93,10 +93,15 @@ endif ()
 find_program(PROVE_PATH prove)
 find_program(UNBUFFER_PATH unbuffer)
 if (PROVE_PATH)
-    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
+    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}" t)
     add_test(
         NAME test-perl-testsuite
         COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
+        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+    )
+    add_test(
+        NAME test-local-author-perl
+        COMMAND "${PROVE_PATH}" xt
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
 else ()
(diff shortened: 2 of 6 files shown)


perlpunk:
For some reason in OBS we now have:
```
[  122s] test 4
[  122s]     Start 4: test-author-perl
[  122s] 
[  122s] 4: Test command: /usr/bin/prove "xt"
```
I don't know why. I don't understand cmake :(

README.asciidoc
```diff
@@ -287,6 +287,16 @@ Run individual tests by specifying them explicitly:
...
 ----
 
+Run perl author tests:
+----
+make test-author-perl
+----
+
+Run all other static tests:
+----
+make test-local
```
okurz, line 297: With this the above

```
Run all tests:
----
make check
----

```

wouldn't be true anymore. I suggest we keep one top level target that simply executes all tests

okurz (changes requested):

README.asciidoc
perlpunk: Yes, but I need help to figure out how.
* We only want tests under `t/` in OBS
* We (at least I) want to split tests in github workflows to get better feedback

One possibility:
* Move all static checks to the Makefile
* `make check `only runs the dynamic tests
* Add a `test-all` target in the Makefile that runs the static checks and `make check`

Makefile
```diff
@@ -14,3 +14,7 @@ Makefile: ;
 build/CMakeCache.txt:
 	@mkdir -p ${build}
 	@cmake -B ${build} -S . -G Ninja
+
+.PHONY: test-author-perl
+test-author-perl:
```
Martchus, line 19: I would rather add this target on CMake level so it also works with Ninja and can depend on the build if that's needed.

perlpunk: but that didn't work like expected, like I wrote.
and it doesn't depend on the build, that's the point.
all the static checks don't depend on the build.

but if someone can explain the magic of the check target to me, and how we can make a target that doesn't run all the static checks, then I'm willing to go with cmake

okurz: Add discussed short-term you should define an additional custom target in cmake/test-targets.cmake e.g. "test-author" and also exclude that in "check-pkg-build"

perlpunk: Done

perlpunk:
The new workflow now runs `test-local` for all author tests, and the ci run only runs `check-pkg-build`

cmake/test-targets.cmake
```diff
@@ -93,24 +93,30 @@ endif ()
...
 # add build system targets for invoking specific tests
+add_custom_target(test-local-author-perl COMMAND ${CMAKE_CTEST_COMMAND} -R "test-local-author-perl" WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
 add_custom_target(test-local COMMAND ${CMAKE_CTEST_COMMAND} -R "test-local-.*" WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
 add_custom_target(test-doc COMMAND ${CMAKE_CTEST_COMMAND} -R "test-doc-.*" WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
 add_custom_target(test-installed-files COMMAND ${CMAKE_CTEST_COMMAND} -R "test-installed-files" WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 if (PROVE_PATH)
     add_custom_target(test-perl-testsuite
-        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
-        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+        COMMAND ${CMAKE_CTEST_COMMAND} -R "test-perl-testsuite"
```
perlpunk, line 118: I think this was an oversight, as `invoke-tests` was already specified in the `add_test(est-perl-testsuite ...)` above, so I replaced it.

Martchus: No, this was no oversight. The idea was to avoid the additional "ctest layer" in case we only need to invoke prove anyway. CTest has the same role as prove - it is a test runner (although probably a more generic one). Making CTest invoke prove is only useful when also running non-Perl tests that probably shouldn't/can't run via prove. See https://github.com/os-autoinst/os-autoinst/commit/e3032d6b1cab0090db002fd2258e9443744d8329 for further reasoning.

Martchus (changes requested):
Looks good except for the change regarding the `test-perl-testsuite` target.

cmake/test-targets.cmake
perlpunk: Ok, I reverted it. I hope `INVOKE_TEST_ARGS` will still be set

okurz (approved):

Martchus (approved):

cmake/test-targets.cmake
Martchus: > I hope `INVOKE_TEST_ARGS` will still be set

Yes, the change to `INVOKE_TEST_ARGS` will still apply automatically in both cases.

PR #2465, opened 2024-02-27
video_stream: Fix uninitialized cmd_prefix

marmarek:
Fix this warning:
Use of uninitialized value $cmd_prefix in split at /usr/lib/os-autoinst/consoles/video_stream.pm line 64.

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -61,7 +61,7 @@ sub disable ($self, @) {
 
 # uncoverable statement count:1..5 note:the function is redefined in tests
 sub _v4l2_ctl ($device, $cmd_prefix, $cmd) {
-    my @cmd = split(/ /, $cmd_prefix);    # uncoverable statement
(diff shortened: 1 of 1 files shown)


consoles/video_stream.pm
```diff
@@ -61,6 +61,7 @@ sub disable ($self, @) {
 
 # uncoverable statement count:1..5 note:the function is redefined in tests
 sub _v4l2_ctl ($device, $cmd_prefix, $cmd) {
+    $cmd_prefix //= "";
     my @cmd = split(/ /, $cmd_prefix);    # uncoverable statement
```
Martchus, line 65: Since this is only used once:
```suggestion
    my @cmd = split(/ /, $cmd_prefix // '');    # uncoverable statement
```
We also prefer single quotes unless interpolation is used.

Martchus (approved):

okurz (approved):

PR #2470, opened 2024-03-11
Skip upload assets when value of PUBLISH_* is none

b10n1k:
Do not allow creation of a file when the PUBLISH_HDD_* is none or None.

https://progress.opensuse.org/issues/156394

### DIFF
--- a/OpenQA/Isotovideo/Utils.pm
+++ b/OpenQA/Isotovideo/Utils.pm
@@ -322,6 +322,10 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
             my $name = $bmwqemu::vars{"STORE_HDD_$i"} || undef;
             unless ($name) {
                 $name = $bmwqemu::vars{"PUBLISH_HDD_$i"} || undef;
+                if ($name =~ /none/i) {
+                    bmwqemu::log_call("Asset upload is skipped for PUBLISH_HDD_$i=$name");
+                    next;
+                }
                 $dir = 'assets_public';
             }
             next unless $name;
--- a/t/20-openqa-isotovideo-utils.t
+++ b/t/20-openqa-isotovideo-utils.t
@@ -3,7 +3,7 @@
 use Test::Most;
 use Test::Warnings qw(warning :report_warnings);
 use autodie ':all';
-use Test::Output qw(combined_like stderr_like);
+use Test::Output qw(combined_like combined_from stderr_like);
 use File::Path qw(remove_tree rmtree);
 use Cwd 'abs_path';
 use Mojo::File qw(tempdir path);
@@ -13,8 +13,9 @@ use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '10';
 use OpenQA::Isotovideo::Utils qw(git_rev_parse checkout_git_refspec
+  handle_generated_assets
   git_remote_url load_test_schedule);
-
+use OpenQA::Isotovideo::CommandHandler;
 
 my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
 my $pool_dir = "$dir/pool";
@@ -65,6 +66,19 @@ subtest 'error handling when loading test schedule' => sub {
         $bmwqemu::vars{PRODUCTDIR} = 'not/found';
         throws_ok { load_test_schedule } qr/PRODUCTDIR.*invalid/, 'error logged';
     };
+};
+
+subtest 'prevent upload assets when publish_hdd is none with case-insensitive' => sub {
+    my $command_handler = OpenQA::Isotovideo::CommandHandler->new();
+    my @possible_values = qw(none None NONE);
+    $bmwqemu::vars{BACKEND} = 'qemu';
+    for my $v (@possible_values) {
+        $bmwqemu::vars{PUBLISH_HDD_1} = $v;
+        $command_handler->test_completed(1);
+        my $return_code;
+        my $log = combined_from { $return_code = handle_generated_assets($command_handler, 1) };
+        like $log, qr/Asset upload is skipped for PUBLISH_HDD/, "Upload is skipped when PUBLISH_HDD_1 is $v";
+    }
 };
 
 is_deeply OpenQA::Isotovideo::Utils::_store_asset(0, 'foo.qcow2', 'bar'), {hdd_num => 0, name => 'foo.qcow2', dir => 'bar', format => 'qcow2'}, '_store_asset returns correct parameters';


OpenQA/Isotovideo/Utils.pm
```diff
@@ -321,6 +321,7 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
             my $dir = 'assets_private';
             my $name = $bmwqemu::vars{"STORE_HDD_$i"} || undef;
             unless ($name) {
+                next if $bmwqemu::vars{"PUBLISH_HDD_$i"} =~ /none|None/;
```
Martchus, line 324:
```suggestion
                next if $bmwqemu::vars{"PUBLISH_HDD_$i"} =~ /none/i;
```

But this is probably the wrong place. It would make much more sense to skip the whole `handle_generated_assets` function early.

Martchus (changes requested):

perlpunk:
It would probably be helpful to write a message in the log that uploading is skipped, so people aren't wondering about it

OpenQA/Isotovideo/Utils.pm
perlpunk: I'm wondering why it isn't suffiicient to just allow `none`

b10n1k: Change id done. I took a look around but i am not sure how easy or small will be to move the logic somewhere else. We can discuss it. For now i am thinking to add a test but my system is broken atm and test do not run.

t/14-isotovideo.t
```diff
@@ -291,6 +291,12 @@ subtest 'upload assets on demand even in failed jobs' => sub {
     like $log, qr/qemu-img.*foo.qcow2/, 'requested image is published even though the job failed';
     ok(-e $pool_dir . '/assets_public/foo.qcow2', 'published image exists');
     ok(!-e bmwqemu::STATE_FILE, 'no fatal error recorded') or die path(bmwqemu::STATE_FILE)->slurp;
+
+    subtest 'upload assets when publish_hdd is none' => sub {
+        $log = combined_from { isotovideo(
+                opts => "casedir=$data_dir/tests schedule=$module publish_hdd_1=none qemu_no_kvm=1 arch=i386 backend=qemu qemu=i386 novideo=1", exit_code => 0) };
+        like $log, qr/Asset upload is skipped for PUBLISH_HDD/, 'Upload is NOT skipped when PUBLISH_HDD_1 is none';
```
perlpunk, line 298:
```suggestion
        like $log, qr/Asset upload is skipped for PUBLISH_HDD/, 'Upload is skipped when PUBLISH_HDD_1 is none';
```

OpenQA/Isotovideo/Utils.pm
```diff
@@ -321,6 +321,10 @@ sub handle_generated_assets ($command_handler, $clean_shutdown) {
             my $dir = 'assets_private';
             my $name = $bmwqemu::vars{"STORE_HDD_$i"} || undef;
             unless ($name) {
+                if ($bmwqemu::vars{"PUBLISH_HDD_$i"} =~ /none/i) {
+                    bmwqemu::log_call("Asset upload is skipped for PUBLISH_HDD_$i");
+                    next;
+                }
```
perlpunk, line 327: This can probably be moved below the next line, where we already assign `$name`.
```suggestion
                if ($name =~ /none/i) {
                    bmwqemu::log_call("Asset upload is skipped for PUBLISH_HDD_$i=$name");
                    next;
                }
```

OpenQA/Isotovideo/Utils.pm
perlpunk: > But this is probably the wrong place. It would make much more sense to skip the whole `handle_generated_assets` function early.

Why would you completely skip all asset uploads if only PUBLISH_HDD_x is none?
In that case I think a special variable `NO_ASSET_UPLOAD` would IMHO make more sense.
But you are more experienced with assets, so maybe I'm missing something.

Martchus: You're right. So let's keep it within that loop.

t/14-isotovideo.t
```diff
@@ -291,6 +291,12 @@ subtest 'upload assets on demand even in failed jobs' => sub {
     like $log, qr/qemu-img.*foo.qcow2/, 'requested image is published even though the job failed';
     ok(-e $pool_dir . '/assets_public/foo.qcow2', 'published image exists');
     ok(!-e bmwqemu::STATE_FILE, 'no fatal error recorded') or die path(bmwqemu::STATE_FILE)->slurp;
+
+    subtest 'upload assets when publish_hdd is none' => sub {
```
perlpunk, line 295: I think this subtest inside the `upload assets on demand even in failed jobs` subtest is misleading. This one is not about failed jobs.

Also I think we might be able to test this without doing a whole isotovideo call (which seems to increase the runtime by 10 seconds).
In line 327 there is a subtest about publishing assets, which unit-tests `handle_generated_assets`. Maybe that's a better place.

t/14-isotovideo.t
```diff
@@ -394,6 +394,14 @@ subtest 'publish assets' => sub {
         is $return_code, 0, 'Successful handle_generated_assets' or die path(bmwqemu::STATE_FILE)->slurp;
         ok !-e $publish_asset, 'test.qcow2 does not exist';
     };
+
+    subtest 'prevent upload assets when publish_hdd is none' => sub {
+        $bmwqemu::vars{PUBLISH_HDD_1} = 'none';
```
kalikiana, line 399: There should be another case using `NONE` or `None`

okurz: that would go a bit too far. This is really just a trivial change costing too much already. Just keep as is

okurz (approved):

t/14-isotovideo.t
b10n1k: Done

b10n1k:
@okurz i already updated the test. then i saw your comment. do i need to revert or...?

kalikiana (approved):

okurz (changes requested):
> @okurz i already updated the test. then i saw your comment. do i need to revert or...?

I wanted to prevent that we block this implementation for too long. As you added it already then it's fine as long as the test runtime is not significantly increased. Right now I can't run t/14-isotovideo.t successfully locally as it fails in t/14-isotovideo.t line 288 so I can't check that and in your PR CI checks have failed due to the registry not being available.

t/14-isotovideo.t
```diff
@@ -394,6 +394,17 @@ subtest 'publish assets' => sub {
...
+    subtest 'prevent upload assets when publish_hdd is none with case-insensitive' => sub {
+        my @possible_values = qw(none None NONE);
+        for my $v (@possible_values) {
+            $bmwqemu::vars{PUBLISH_HDD_1} = $v;
+            $command_handler->test_completed(1);
+            my $return_code;
+            my $log = combined_from { $return_code = handle_generated_assets($command_handler, 1) };
+            like $log, qr/Asset upload is skipped for PUBLISH_HDD/, "Upload is skipped when PUBLISH_HDD_1 is $v";
+        }
+    };
```
okurz, line 407: The corresponding test module for the file OpenQA/Isotovideo/Utils.pm is 20-openqa-isotovideo-utils.t . Can you consider moving your test code there?

okurz (changes requested):

t/14-isotovideo.t
perlpunk: then that whole asset subtest should be moved there.
it wouldn't make sense to just move this single test

Martchus (approved):
I guess I wouldn't have tested the different casings and rather added a check to verify that the asset handling is actually skipped (not just checking the log message). However, I think this is good enough.

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/8282581547/job/22663706882?pr=2470#step:4:171
```
5: ::warning file=/__w/os-autoinst/os-autoinst/t/20-openqa-isotovideo-utils.t,line=74,col=20,title=Hash key "BACKEND" with quotes - severity 5::[HashKeyQuotes] Avoid useless quotes
```

t/20-openqa-isotovideo-utils.t
```diff
@@ -67,6 +68,19 @@ subtest 'error handling when loading test schedule' => sub {
...
+    my @possible_values = qw(none None NONE);
+    $bmwqemu::vars{"BACKEND"} = 'qemu';
+    for my $v (@possible_values) {
+        $bmwqemu::vars{PUBLISH_HDD_1} = $v;
+        $command_handler->test_completed(1);
+        my $return_code;
+        my $log = combined_from { $return_code = handle_generated_assets($command_handler, 1) };
+        like $log, qr/Asset upload is skipped for PUBLISH_HDD/, "Upload is skipped when PUBLISH_HDD_1 is $v";
+    }
+};
```
perlpunk, line 82: I think the whole "publish assets" subtest from 14-isotovideo.t  should be moved here together.
It doesn't make sense to have very similar tests in two different test scripts.

kalikiana: To my mind both are not great. I'd say leave it where it is for now - or actually move this and other asset tests into a new file

b10n1k: i agree as i notice but i didnt want to "take it too far". please instruct me how to proceed. or we can merge this and create another one as anyway it is going to be another commit anyway

okurz: nevermind. Just fix the author test failure :)

kalikiana: @b10n1k This:

```
The following tests FAILED:
	  5 - test-local-perl-style (Failed)
```

b10n1k: yea but why? the tidy doesnt do anything

b10n1k: solved

okurz (approved):

perlpunk (approved):
I approve it and will make a followup PR moving the tests together again that belong together
- https://github.com/os-autoinst/os-autoinst/pull/2473

PR #2473, opened 2024-03-15
Move the same kind of tests together

perlpunk:
All `handle_generated_assets` tests should be in one file

Unfortunately tests are failing. Presumable there are things happening in `t/14-isotovideo.t` that made the tests pass, and that things are missing in `t/20-openqa-isotovideo-utils.t`

As I don't have a ticket assigned for that and it was @okurz ' idea to not put the new `handle_generated_assets` to the other 5 `handle_generated_assets` tests where it would belong, I leave it to him to decide if we want to
* keep the current state of similar tests across several test files
* fix the tests
* or actually just keep it simple and move the new test to  `t/14-isotovideo.t` as it was actually fine and working, and noone before actually complained about the existing `handle_generated_assets` unit tests in there.


okurz:
Both
* fix the tests
* or actually just keep it simple and move the new test to t/14-isotovideo.t as it was actually fine and working, and noone before actually complained about the existing handle_generated_assets unit tests in there.

sound acceptable to me. But as t/14-isotovideo.t was in the past often leading to unintended significant runtime increases I suggest "fix the tests"

PR #2474, opened 2024-03-25
Move exit_code_from_test_results to Runner.pm

b10n1k:
`exit_code_from_test_results` is moved to Runner Class in order to be able                                                                                                                                                                   
to speed the tests up for 14-isotovideo.t.                                                                                                                                                                                                   
14-isotovideo.t is faster by ~3 wallclock.    

https://progress.opensuse.org/issues/157339

### DIFF
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -3,9 +3,6 @@
 
 package OpenQA::Isotovideo::Runner;
 use Mojo::Base -base, -signatures;
-# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
-# Details: https://progress.opensuse.org/issues/90371
-use JSON::PP;
 
 use autodie ':all';
 no autodie 'kill';
@@ -13,6 +10,7 @@ use POSIX qw(:sys_wait_h _exit);
 use Mojo::File qw(path);
 use Mojo::UserAgent;
 use IO::Select;
+use JSON::PP;
 use log qw(diag fctwarn);
 use OpenQA::Isotovideo::Utils qw(checkout_git_repo_and_branch checkout_git_refspec checkout_wheels
   load_test_schedule);
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -3,9 +3,14 @@
 
 package OpenQA::Isotovideo::Runner;
 use Mojo::Base -base, -signatures;
+# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
+# Details: https://progress.opensuse.org/issues/90371
+use JSON::PP;
+
 use autodie ':all';
 no autodie 'kill';
 use POSIX qw(:sys_wait_h _exit);
+use Mojo::File qw(path);
 use Mojo::UserAgent;
 use IO::Select;
 use log qw(diag fctwarn);
@@ -28,6 +33,12 @@ has [qw(backend command_handler)];
 
 # the loop status
 has loop => 1;
+
+use constant {
+    EXIT_STATUS_OK => 0,
+    EXIT_STATUS_ERR_NO_TESTS => 100,
+    EXIT_STATUS_ERR_FROM_TEST_RESULTS => 101,
+};
 
 sub run ($self) {
     # now we have everything, give the tests a go
@@ -233,6 +244,23 @@ sub _init_bmwqemu ($, @args) {
             diag("Setting forced test parameter $key -> $2");
         }
     }
+}
+
+sub exit_code_from_test_results ($self) {
+    my @results = glob(path(bmwqemu::result_dir(), "result-*.json"));
+    return EXIT_STATUS_ERR_NO_TESTS if @results == 0;
+
+    my $did_fail = 0;
+    for my $result_file_path (@results) {
+        my $result_file = path($result_file_path);
+        my $test_result = decode_json($result_file->slurp)->{result};
+        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
+        # If a failure (anything different from ok & softfail) is found, keep it.
+        next if $did_fail;
+
+        $did_fail = $test_result !~ m/^(ok|softfail)$/;
+    }
+    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
 }
 
 sub handle_shutdown ($self, $return_code) {
--- a/isotovideo
+++ b/isotovideo
@@ -85,8 +85,6 @@ no autodie 'kill';
 use constant {
     EXIT_STATUS_OK => 0,
     EXIT_STATUS_ERR => 1,
-    EXIT_STATUS_ERR_NO_TESTS => 100,
-    EXIT_STATUS_ERR_FROM_TEST_RESULTS => 101,
 };
 
 # Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
@@ -113,7 +111,6 @@ Getopt::Long::Configure("no_ignore_case");
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::Runner;
 use OpenQA::Isotovideo::Utils qw(git_rev_parse spawn_debuggers handle_generated_assets);
-use Mojo::File qw(path);
 
 my %options;
 
@@ -129,23 +126,6 @@ sub usage ($r) {
 sub _get_version_string () {
     my $thisversion = git_rev_parse(curfile->dirname);
     return "Current version is $thisversion [interface v$OpenQA::Isotovideo::Interface::version]";
-}
-
-sub _exit_code_from_test_results () {
-    my @results = glob(path(bmwqemu::result_dir(), "result-*.json"));
-    return EXIT_STATUS_ERR_NO_TESTS if @results == 0;
-
-    my $did_fail = 0;
-    for my $result_file_path (@results) {
-        my $result_file = path($result_file_path);
-        my $test_result = decode_json($result_file->slurp)->{result};
-        diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
-        # If a failure (anything different from ok & softfail) is found, keep it.
-        next if $did_fail;
-
-        $did_fail = $test_result !~ m/^(ok|softfail)$/;
-    }
-    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
 }
 
 sub version () {
@@ -202,7 +182,7 @@ eval {
     $runner->run;
 
     if ($options{'exit-status-from-test-results'}) {
-        $RETURN_CODE = _exit_code_from_test_results();
+        $RETURN_CODE = $runner->exit_code_from_test_results;
     }
 
     # terminate/kill the command server and let it inform its websocket clients before
--- a/t/19-isotovideo-command-processing.t
+++ b/t/19-isotovideo-command-processing.t
@@ -12,11 +12,14 @@ use Test::Warnings ':report_warnings';
 use Test::Fatal;
 use Mojo::JSON 'encode_json';
 use Mojo::File qw(tempdir path);
+use Mojo::Util qw(scope_guard);
 use OpenQA::Isotovideo::CommandHandler;
 use OpenQA::Isotovideo::Interface;
 use OpenQA::Isotovideo::Runner;
 
 my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
+chdir $dir;
+my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
 
 # declare fake file descriptors
 my $cmd_srv_fd = 0;
@@ -399,7 +402,6 @@ subtest signalhandler => sub {
 };
 subtest 'Check exit_code_from_test_results' => sub {
     my $runner = OpenQA::Isotovideo::Runner->new;
-    chdir($dir);
 
     subtest 'no test scheduled' => sub {
         my $path_mock = Test::MockModule->new('Mojo::File');
@@ -450,8 +452,3 @@ subtest 'shutdown handling' => sub {
 };
 
 done_testing;
-
-END {
-    unlink OpenQA::Isotovideo::CommandHandler::AUTOINST_STATUSFILE;
-    unlink bmwqemu::STATE_FILE;
-}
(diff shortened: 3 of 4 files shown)


Martchus (dismissed):

isotovideo
```diff
@@ -135,17 +135,14 @@ sub _exit_code_from_test_results () {
...
         my $test_result = decode_json($result_file->slurp)->{result};
         diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
-        # If a failure (anything different from ok & softfail) is found, keep it.
-        next if $did_fail;
-
-        $did_fail = $test_result !~ m/^(ok|softfail)$/;
+        # If a failure (anything different from ok & softfail) is found, return immediately the exit status.
+        return EXIT_STATUS_ERR_FROM_TEST_RESULTS if $test_result !~ m/^(ok|softfail)$/;
     }
-    return $did_fail ? EXIT_STATUS_ERR_FROM_TEST_RESULTS : EXIT_STATUS_OK;
```
okurz, line 148: I'd be more careful about this change. Why was it introduced like this originally then to distinguish between the cases instead of immediately returning on failure? Maybe this has an impact on using "ignore_failure" from http://open.qa/docs/#_test_flags ?

okurz (changes requested):

isotovideo
```diff
@@ -135,17 +135,14 @@ sub _exit_code_from_test_results () {
...
     for my $result_file_path (@results) {
         my $result_file = path($result_file_path);
         my $test_result = decode_json($result_file->slurp)->{result};
         diag sprintf("Test result [%s] %s", $result_file->to_string, $test_result);
-        # If a failure (anything different from ok & softfail) is found, keep it.
-        next if $did_fail;
-
-        $did_fail = $test_result !~ m/^(ok|softfail)$/;
+        # If a failure (anything different from ok & softfail) is found, return immediately the exit status.
+        return EXIT_STATUS_ERR_FROM_TEST_RESULTS if $test_result !~ m/^(ok|softfail)$/;
```
perlpunk, line 143: The return wasn't done here so all test results can be printed.

perlpunk (changes requested):

perlpunk:
> This commit intends to do the call faster returning immediately on the first failing result.

If this is just about returning faster because of the unit test runtime - I doubt that will result in a visible difference.

isotovideo
Martchus: Oh, right. So we probably don't want to merge this after all.

OpenQA/Isotovideo/Runner.pm
```diff
@@ -19,16 +19,29 @@ use autotest ();
 use needle ();
 use commands ();
 use distribution ();
+use Mojo::File qw(path);
+
+# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
+# Details: https://progress.opensuse.org/issues/90371
+use JSON::PP;
```
perlpunk, line 26: That shouldn't be moved here. It's something that should happen before all other module use statements.

perlpunk: If it is actually need in this module, then copy that use statement, but leave the original in the isotovideo script

okurz:
guess the PR subject and description don't match your intention anymore?

OpenQA/Isotovideo/Runner.pm
```diff
@@ -49,6 +62,7 @@ sub run ($self) {
         $ch->check_asserted_screen if defined($ch->tags);
     }
     $ch->stop_command_processing;
+    $self->_exit_code_from_test_results();
```
kalikiana, line 65: Do we always want to compute the exit code? If yes, why not return it here?

Otherwise I would have expected that this is called if needed like before

OpenQA/Isotovideo/Runner.pm
kalikiana: @b10n1k Did you see this? To me it looks like you didn't change it.

I recommend always checking and responding to unresolved conversations.

Martchus (changes requested):
The part "in order to decouple the isotovideo from the isotovideo." of the commit message is still weirdly worded. One "isotovideo" should likely be something else. You also should not put an article (like "the") before names (like "isotovideo").

OpenQA/Isotovideo/Runner.pm
b10n1k: i am looking at the second option.

perlpunk:
I'm getting:
```
    # Subtest: tests scheduled
        1..0
        # No tests run!
    not ok 2 - No tests run for subtest "tests scheduled"

    #   Failed test 'No tests run for subtest "tests scheduled"'
    #   at t/19-isotovideo-command-processing.t line 415.
    1..2
    # Looks like you failed 1 test of 2.
not ok 16 - Check results_exit_code

#   Failed test 'Check results_exit_code'
#   at t/19-isotovideo-command-processing.t line 425.
Can't open file "testresults/result-failing_module.json": No such file or directory at t/19-isotovideo-command-processing.t line 411.
```
I suggest to create a tmpdir like in 14-isotovideo.t and `chdir` into it. Then you can also skip the manual removal of that directory.
We generally avoid creating files in the git directory during tests.

OpenQA/Isotovideo/Runner.pm
perlpunk: yes, I also think this is the better option

isotovideo
```diff
@@ -85,14 +85,8 @@ no autodie 'kill';
...
     EXIT_STATUS_OK => 0,
     EXIT_STATUS_ERR => 1,
-    EXIT_STATUS_ERR_NO_TESTS => 100,
-    EXIT_STATUS_ERR_FROM_TEST_RESULTS => 101,
 };
 
-# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
-# Details: https://progress.opensuse.org/issues/90371
-use JSON::PP;
-
```
perlpunk, line 95: These 3 lines need to stay here. It's a bug introduced by Cpanel::JSON::XS (it's creating the same functions as JSON::PP) and fixed in newer versions of JSON::PP, however we are still on a perl version where we don't have the newer JSON::PP.
In the module, just do `use JSON::PP;`

okurz:
I checked the runtime of t/14-isotovideo in my setup and got:
* *before:* 21.0±0.2 s
* *after:* 18.5±0.3 s

which is an improvement :)

isotovideo
kalikiana: As far as I can tell this has been moved without changing how it works in the latest revision.

OpenQA/Isotovideo/Runner.pm
kalikiana: Are we good here then?

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +394,34 @@ subtest signalhandler => sub {
     } qr/isotovideo received signal INT/, 'Signal logged';
     is($last_signal, 'INT', 'Event emitted');
 };
+subtest 'Check exit_code_from_test_results' => sub {
+    my $runner = OpenQA::Isotovideo::Runner->new;
+    use Mojo::File 'path';
+    use Mojo::JSON 'encode_json';
```
Martchus, line 400: Please add `use` statements only at the top of the file (preferably in alphabetical order and merging with existing `use` statements).

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +394,34 @@ subtest signalhandler => sub {
...
+    my %fake_results = ();
```
Martchus, line 401:
```suggestion
    my %fake_results;
```

But probably it makes sense to remove the variable completely, see my other suggestion.

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +394,34 @@ subtest signalhandler => sub {
...
+
+    subtest 'no test scheduled' => sub {
+        my $path_mock = Test::MockModule->new('Mojo::File');
+        $path_mock->redefine('path' => sub { 0 });
```
Martchus, line 405:
```suggestion
        $path_mock->redefine(path => 0);
```

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +394,34 @@ subtest signalhandler => sub {
...
+        is($runner->exit_code_from_test_results(), 100, 'EXIT_STATUS_ERR_NO_TESTS returns if no results');
+    };
+
+    subtest 'tests scheduled' => sub {
+        %fake_results = ("result" => "failing");
```
Martchus, line 410:
```suggestion
        %fake_results = (result => 'failing');
```
Also counts for the other occurrences. But probably it makes sense to remove the variable completely, see my other suggestion.

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +394,34 @@ subtest signalhandler => sub {
...
+        my $json = encode_json(\%fake_results);
```
Martchus, line 411: Unless I'm overlooking something here you can probably remove the `%fake_results;` variable completely:
```suggestion
        my $json = encode_json({result => 'failing'});
```

OpenQA/Isotovideo/Runner.pm
```diff
@@ -19,6 +23,7 @@ use autotest ();
 use needle ();
 use commands ();
 use distribution ();
+use Mojo::File qw(path);
```
Martchus, line 26: Maybe move this up to the other `Mojo::` uses so things are at least a little bit in order/grouped.

OpenQA/Isotovideo/Runner.pm
```diff
@@ -235,6 +246,23 @@ sub _init_bmwqemu ($, @args) {
     }
 }
 
+sub exit_code_from_test_results ($self) {
```
Martchus, line 249: I guess for this particular package it is *ok* that the variable `$self` is not actually used. (So no change requested here. I just find it a bit unusual.)

OpenQA/Isotovideo/Runner.pm
```diff
@@ -3,6 +3,10 @@
 
 package OpenQA::Isotovideo::Runner;
 use Mojo::Base -base, -signatures;
+# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
+# Details: https://progress.opensuse.org/issues/90371
+use JSON::PP;
```
Martchus, line 8: Have you checked whether we really need this here as well now?

Martchus (changes requested):

OpenQA/Isotovideo/Runner.pm
b10n1k: for me it doesnt work if i remove it. So i did copy it and put it back in isotovideo as suggested.

OpenQA/Isotovideo/Runner.pm
b10n1k: yes. it doesnt work without it. See also https://github.com/os-autoinst/os-autoinst/pull/2474#discussion_r1540974446

OpenQA/Isotovideo/Runner.pm
b10n1k: Done

t/19-isotovideo-command-processing.t
b10n1k: Done

t/19-isotovideo-command-processing.t
b10n1k: done

t/19-isotovideo-command-processing.t
b10n1k: sure. i removed it

okurz (approved):

Martchus (approved):

Martchus:
But it looks like style tests are failing so you should probably invoke the test `test-local-perl-style` locally and see what's wrong.

OpenQA/Isotovideo/Runner.pm
```diff
@@ -3,9 +3,14 @@
 
 package OpenQA::Isotovideo::Runner;
 use Mojo::Base -base, -signatures;
+# Avoid "Subroutine JSON::PP::Boolean::(0+ redefined" warnings
+# Details: https://progress.opensuse.org/issues/90371
```
perlpunk, line 7: Please remove the comment here.
You only need to use the module, as I said before.

perlpunk (changes requested):
I get this error:
```
% prove -I . t/19-isotovideo-command-processing.t -v

...

    not ok 2 - No tests run for subtest "tests scheduled"

    #   Failed test 'No tests run for subtest "tests scheduled"'
    #   at t/19-isotovideo-command-processing.t line 412.
    1..2
    # Looks like you failed 1 test of 2.
not ok 16 - Check exit_code_from_test_results

#   Failed test 'Check exit_code_from_test_results'
#   at t/19-isotovideo-command-processing.t line 420.
Can't open file "testresults/result-failed_module.json": No such file or directory at t/19-isotovideo-command-processing.t line 409.
ok 17 - no (unexpected) warnings (via END block)
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 17.
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/17 subtests 
```
Please remove any files that don't belong to git and run the test to see if it fails for you.

Was my previous comment about this ignored?
https://github.com/os-autoinst/os-autoinst/pull/2474#issuecomment-2025573765

perlpunk:
I'm assuming that some test before `t/19-isotovideo-command-processing.t` creates a `testresults` directory. That's why CI is not failing.
And OpenQA::Test::CheckGitStatus doesn't complain about it because it's empty.

t/19-isotovideo-command-processing.t
```diff
@@ -394,6 +395,29 @@ subtest signalhandler => sub {
     } qr/isotovideo received signal INT/, 'Signal logged';
     is($last_signal, 'INT', 'Event emitted');
 };
+subtest 'Check exit_code_from_test_results' => sub {
+    my $runner = OpenQA::Isotovideo::Runner->new;
+
+    subtest 'no test scheduled' => sub {
+        my $path_mock = Test::MockModule->new('Mojo::File');
+        $path_mock->redefine('path' => 0);
```
perlpunk, line 403:
```suggestion
        $path_mock->redefine(path => 0);
```

perlpunk:
There's one thing left:
https://github.com/os-autoinst/os-autoinst/actions/runs/8539855359/job/23395531577?pr=2474#step:3:549
```
3: [2024-04-03T13:52:21.301214Z] [debug] [pid:2382] Test result [testresults/result-failed_module.json] failing
3: [2024-04-03T13:52:21.303509Z] [debug] [pid:2382] Test result [testresults/result-softfailed_module.json] softfail
3: [13:52:22] t/19-isotovideo-command-processing.t ..... ok     1767 ms ( 0.01 usr  0.00 sys +  1.65 cusr  0.12 csys =  1.78 CPU)
```
We can mock `log::diag` to catch this output.

perlpunk:
I also added the mocking of the diag function.

perlpunk (approved):
Ok, no other feedback anymore, so I'll approve

PR #2475, opened 2024-03-26
invoke-tests: Allow PERL5OPT with coverage

kalikiana:
Currently options are always overridden when enabling coverage.

For example:

    PERL5OPT=-d:NYTProf ./tools/invoke-tests --coverage t/14-isotovideo.t


Martchus (approved):

perlpunk (approved):
Note that it is only overriden when we use `--coverage` though.
edit: that's what you wrote, sorry. Was just wanting to say that using NYTProf **and** coverage at the same time is probably not useful

okurz:
I would approve if the PR would be ready for review and not draft

PR #2477, opened 2024-03-28
Fix video_stream console (ustreamer mode) and make debugging such issues easier

marmarek:
Two fixes for the ustreamer mode in the video_stream console.
And also do not ignore error in `select_console()` - that would have saved me several hours of debugging.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -254,7 +254,8 @@ sub load_snapshot ($sname) {
     # On VMware VNC console needs to be re-selected after snapshot revert,
     # so the screen is refreshed. Same with serial console.
     return unless ($command // '') eq 'vmware_fixup';
-    testapi::select_console('sut');
+    my $ret = testapi::select_console('sut');    # uncoverable statement
+    die $ret->{error} if $ret->{error};    # uncoverable statement
     query_isotovideo('backend_stop_serial_grab');
     query_isotovideo('backend_start_serial_grab');
 }
--- a/backend/generalhw.pm
+++ b/backend/generalhw.pm
@@ -93,7 +93,8 @@ sub relogin_vnc ($self) {
             jpeg => $bmwqemu::vars{GENERAL_HW_VNC_JPEG} // 0,
         });
     $vnc->backend($self);
-    $self->select_console({testapi_console => 'sut'});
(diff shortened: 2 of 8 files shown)


marmarek:
```
3:     # No tests run!
3: 
3: #   Failed test 'No tests run for subtest "starting VM"'
3: #   at t/29-backend-s390x.t line 38.
3: Need variable ZVM_HOST at /opt/consoles/s3270.pm line 420.
3: # Tests were run but no plan was declared and done_testing() was not seen.
3: # Looks like your test exited with 255 just after 4.
```

So, this is a failure in another `select_console()` call that was ignored before. Let me mock some more in that test...

okurz:
cool. Please look into test coverage for the two missing lines https://github.com/os-autoinst/os-autoinst/pull/2477#issuecomment-2024317996

Martchus (approved):

marmarek:
Technically, they weren't covered before either (that's the "Misses -1" in summary), but sure, I can add a comment to make coverage happy...

autotest.pm
```diff
@@ -249,12 +249,14 @@ sub make_snapshot ($sname) {
 }
 
 sub load_snapshot ($sname) {
+    # uncoverable subroutine
```
marmarek, line 252: This isn't enough :(

marmarek:
>  codecov/project/fully_covered Failing after 1s — 99.99% (target 100.00%)

What does it complain about? 

I do see this in the other place:

> All modified and coverable lines are covered by tests ✅

Martchus:
`codecov/project/fully_covered` is about a subset of files that need to be fully covered (so we don't regress there). The check fails because of a change in test coverage in the file `backend/baseclass.pm` which is among those files where we expect 100 % test coverage. Your changes don't touch this file but apparently affect its coverage indirectly which shows on Codecov as well: https://app.codecov.io/gh/os-autoinst/os-autoinst/pull/2477/indirect-changes (line 664: `{error => $_};`)

Probably we don't run into this error case anymore because the errors are now handled before. I suppose the easiest workaround would be to mark this line as uncoverable. One could also try to add an explicit test for it.

PR #2478, opened 2024-03-28
Avoid errors in `get_job_autoinst_vars` due to not using FQDNs

Martchus:
* Use FQDN in cross-worker MM-API calls if available
* See https://progress.opensuse.org/issues/158185

### DIFF
--- a/mmapi.pm
+++ b/mmapi.pm
@@ -183,8 +183,9 @@ sub get_job_autoinst_url ($target_id) {
     my $workers = $tx->res->json('/workers') // [];
     for my $worker (@$workers) {
         if ($worker->{jobid} && $target_id == $worker->{jobid} && $worker->{host} && $worker->{instance} && $worker->{properties}{JOBTOKEN}) {
-            my $hostname = $worker->{host};
-            my $token = $worker->{properties}{JOBTOKEN};
+            my $properties = $worker->{properties};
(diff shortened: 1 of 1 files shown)


Martchus:
Yes, and I commented about that in the ticket.

mmapi.pm
```diff
@@ -183,8 +183,9 @@ sub get_job_autoinst_url ($target_id) {
     my $workers = $tx->res->json('/workers') // [];
     for my $worker (@$workers) {
         if ($worker->{jobid} && $target_id == $worker->{jobid} && $worker->{host} && $worker->{instance} && $worker->{properties}{JOBTOKEN}) {
-            my $hostname = $worker->{host};
-            my $token = $worker->{properties}{JOBTOKEN};
+            my $properties = $worker->{properties};
+            my $hostname = $properties->{WORKER_HOSTNAME} // $worker->{host};
```
kalikiana, line 187: Thinking out loud here. Wasn't there a reason to avoid this in the first place?

Not saying it's wrong but I remember long conversations on (not) using FQDN or not being able to expect that it is what we want

Martchus: I think this code simply predates the introduction of `WORKER_HOSTNAME`.

kalikiana (approved):

PR #2479, opened 2024-03-28
setup-multimachine: Retry package installation up to 3 times

kalikiana:
We already use retry in tests which rely on this script so it should be available. Failing that install it anyway.

See: https://progress.opensuse.org/issues/158245

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -16,7 +16,8 @@ ensure_ip_forwarding() {
 }
 
 install_packages() {
-    zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    command -v retry > /dev/null || zypper -n in retry
+    retry -- zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
 }
 
 configure_firewall() {


script/os-autoinst-setup-multi-machine
```diff
@@ -16,7 +16,9 @@ ensure_ip_forwarding() {
 }
 
 install_packages() {
-    zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    for i in $(seq 3); do
+        retry zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs && break
```
perlpunk, line 20: Hmm, the description says that it's *not* using retry :)

okurz: retry is already installed in the environment we are testing: https://openqa.opensuse.org/tests/4049058#step/prepare/5 . And retry is part of an official repository so it's much more likely to be successful compared to a package from a devel repository which is also what failed in https://progress.opensuse.org/issues/158245 , see https://openqa.opensuse.org/tests/4049058#step/openqa_worker/7 , so I suggest to install retry if it isn't installed and then use it

```bash
    command -v retry >/dev/null || zypper -n in retry
    retry zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
```

okurz (changes requested):

script/os-autoinst-setup-multi-machine
kalikiana: Good point. It makes sense to assume retry is installed in the test environment then which it generally is so I'll simply use it.

Martchus (commented):
Do you have a verification run (showing that retry is actually available here)?

kalikiana:
> Do you have a verification run (showing that retry is actually available here)?

Well, https://openqa.opensuse.org/tests/4049058#step/prepare/5 shows that retry is installed and used in the test. Which is why I make that assumption.

script/os-autoinst-setup-multi-machine
perlpunk: I think the suggestion was to install retry if it is not installed. I guess that test is not the only place where this script might be called.

Martchus (approved):

script/os-autoinst-setup-multi-machine
```diff
@@ -16,7 +16,7 @@ ensure_ip_forwarding() {
 }
 
 install_packages() {
-    zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    retry zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
```
okurz, line 19: I doubt you tested this `retry: invalid option -- 'n'`

okurz (changes requested):
`

script/os-autoinst-setup-multi-machine
kalikiana: I didn't like that suggestion. But I also don't care enough to defend my position here so I changed it.

script/os-autoinst-setup-multi-machine
kalikiana: Should not have tested individual lines outside the script 😿

Martchus: It would be nice if that was correct, though. Then `retry` would be in-line with other "prefix"-commands such as `sudo` and `time`.

kalikiana: Not sure what you mean. Those are not used here?

The behavior now is that retry is installed if the command is unavailable. And zypper is retried 3 times if it fails i.e. the default.

okurz (approved):

script/os-autoinst-setup-multi-machine
okurz: I think @Martchus means would be nice if github.com/okurz/retry could behave better as in understanding that the parameter "-n" is not meant for retry but the executed command. If one of you has a good idea how https://github.com/okurz/retry/blob/main/retry#L19 could be changed to do that that would be great.

baierjan: https://github.com/okurz/retry/pull/6 should fix that behavior.

PR #2481, opened 2024-04-08
setup-multimachine: Retry package installation up to 7 times

kalikiana:
Also do exponential retry and wait 30 seconds.

See: https://progress.opensuse.org/issues/158553

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -17,7 +17,7 @@ ensure_ip_forwarding() {
 
 install_packages() {
     command -v retry > /dev/null || zypper -n in retry
-    retry -- zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    retry -e -s 30 -r 7 -- sh -c "zypper ref && zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs"
 }
 
 configure_firewall() {


script/os-autoinst-setup-multi-machine
```diff
@@ -17,7 +17,7 @@ ensure_ip_forwarding() {
 
 install_packages() {
     command -v retry > /dev/null || zypper -n in retry
-    retry -- zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    retry -e -s 30 -r 7 -- zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
```
okurz, line 20: That likely won't prevent the problem. We need to rerun `zypper ref`. We have something similar elsewhere already. I suggest you look up that code and use the same here

kalikiana: See https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/170

Martchus (approved):

perlpunk (approved):

perlpunk:
@mergifyio rebase

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/170#discussion_r1557892156

script/os-autoinst-setup-multi-machine
```diff
@@ -17,7 +17,7 @@ ensure_ip_forwarding() {
 
 install_packages() {
     command -v retry > /dev/null || zypper -n in retry
-    retry -- zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs
+    retry -e -s 30 -r 7 -- sh -c "zypper ref && zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs"
```
okurz, line 20: Again, you need to distinguish between installing with custom OBS development repositories or without. AFAIK os-autoinst-setup-multi-machine never states anything about development repositories so it shouldn't care to do those multi-layer retries unless it's an accepted upstream zypper issue that we want to reference and workaround.

okurz (changes requested):

script/os-autoinst-setup-multi-machine
kalikiana: I'm confused. You literally said `That likely won't prevent the problem. We need to rerun zypper ref` in your previous comment. I guess I'll bring it up on a call

okurz: yes, but only when *installing with custom OBS development repositories*

okurz: yes, but only when *installing with custom OBS development repositories*

okurz (approved):

PR #2482, opened 2024-04-09
Unify size argument for disk creation

grisu48:
There was an inconsistency between _copy_image_to_vm_host and add_disk where one assumed the size to be an int and the other method assumed to be in the format qemu-img expects it (e.g. 20G).

This commit unifies the argument by making it everywhere use the qemu-img format, as it is used widely in the os-autoinst-distri-opensuse code.


consoles/sshVirtsh.pm
```diff
@@ -399,8 +399,9 @@ sub _copy_image_to_vm_host ($self, $args, $vmware_openqa_datastore, $file, $name
     }
     else {
         $file = $basedir . $file;
-        # requested expected value in GBs, if there is any passed as an argument
-        my $size = $args->{size} // 0;
+        # requested expected value in GiBs, if there is any passed as an argument
+        my $size = $args->{size} // '0G';
+        $size =~ tr/.*G$/.*/d;
```
okurz, line 404: so you add a G as default just to remove it again in the second line and then in three lines below the G is added again? How about

```suggestion
        $args->{size} =~ tr/.*G$/.*/d;
        my $size = $args->{size} // 0;
```

or

```suggestion
        my $size = $args->{size} =~ tr/.*G$/.*/dr // 0;
```

okurz (changes requested):

consoles/sshVirtsh.pm
grisu48: I like suggestion two and amended it accordingly.

grisu48: On a second though, I do think that the original approach was better because it shows the expected format of the `$args->{size}` argument:

```
        my $size = $args->{size} // '0G';
        $size =~ tr/.*G$/.*/d;
```
From the first line one can immediately deduce that the expected format of `$args->{size}` is something like `20G`.

The new solution is more compact, but this information is now also kinda difficult to read out of the context. It's encoded in the regex, but hidden in logic, which makes the code less readable.

okurz: well, why would the code later add a 'G' as suffix then?

okurz (approved):

consoles/sshVirtsh.pm
grisu48: This discussion suggests actually to add an explanatory comment, that might resolve this issue. Give me a moment ....

grisu48: @okurz please check again, this is a solution I would be also happy with.

kalikiana (approved):

grisu48:
No idea how to make the Codecov test happy. Three restarts and all fail with the same 404 error.

okurz:
Don't know either but all CI tests are good and basically you only changed one line that was already covered in before so we are good.

Martchus (commented):
I guess this works but it would probably make sense to move it into its own function and add a test.

okurz (approved):

consoles/sshVirtsh.pm
```diff
@@ -399,8 +399,8 @@ sub _copy_image_to_vm_host ($self, $args, $vmware_openqa_datastore, $file, $name
     }
     else {
         $file = $basedir . $file;
-        # requested expected value in GBs, if there is any passed as an argument
-        my $size = $args->{size} // 0;
+        # $args->{size} is expected to be e.g. '20G' but internally we need it as integer
+        my $size = $args->{size} =~ tr/.*G$/.*/dr // 0;
```
perlpunk, line 403: Now we would get a warning if `$args->{size}` is undef.
`Use of uninitialized value in transliteration (tr///)`

perlpunk: I think the `tr` can be simplified:
`tr/G//dr`
`.*` and `$` have no special meaning in transliterations.

grisu48: https://github.com/os-autoinst/os-autoinst/pull/2483

PR #2485, opened 2024-04-10
Adjust dependencies to support build of s390x packages

okurz:
Related progress issue: https://progress.opensuse.org/issues/158455

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -9,8 +9,8 @@
 ---
 targets:
   cpanfile: [                    cover, devel,         main,             lint,           test_base, test, test_version_only                    ]
-  spec:     [ build_base, build,        devel,         main, spellcheck,       yamllint, test_base, test, test_version_only, python_style, ocr ]
-  docker:   [ build_base, build,        devel, docker, main, spellcheck, lint, yamllint, test_base, test,                    python_style, ocr ]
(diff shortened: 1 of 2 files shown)


dependencies.yaml
```diff
@@ -95,14 +95,12 @@ test_base_requires:
...
   qemu: '>= 4.0'
   qemu-tools:
-  qemu-x86:
   procps:
   python3-setuptools:
   xterm-console:
   xorg-x11-Xvnc:
   icewm:
   xterm:
-  ipxe-bootimgs:
```
kalikiana, line 105: You still need to commit an updated `Dockerfile` it seems:

```
9: 
9: #   Failed test 'No changed files after 'make update-deps''
9: #   at xt/27-make-update-deps.t line 25.
9: #  M container/os-autoinst_dev/Dockerfile
9: # Looks like you failed 1 test of 2.
9: xt/27-make-update-deps.t .. 
```

kalikiana: Right. We have `ifarch x86_x86` in `os-autoinst.spec` which might be easier than supporting that in `dependencies.yaml`.

okurz: @Martchus I understood how I can use a similar approach as for "ocr", adapted now

@kalikiana ifarch x86_64 is not the same as ifnarch s390x

Martchus (approved):

kalikiana (approved):

PR #2489, opened 2024-04-17
qemu: Set proper default on s390x

okurz:
'vga' seems to be not supported on s390x.

Related progress issue: https://progress.opensuse.org/issues/158985

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -567,10 +567,10 @@ sub setup_tpm ($self, $arch) {
     }
     sp('chardev', "socket,id=chrtpm,path=$vmsock");
     sp('tpmdev', 'emulator,id=tpm0,chardev=chrtpm');
-    if ($arch eq 'aarch64') {
+    if (($arch // '') eq 'aarch64') {
         sp('device', 'tpm-tis-device,tpmdev=tpm0');
     }
-    elsif ($arch eq 'ppc64le') {
+    elsif (($arch // '') eq 'ppc64le') {
         sp('device', 'tpm-spapr,tpmdev=tpm0');
         sp('device', 'spapr-vscsi,id=scsi9,reg=0x00002000');
     }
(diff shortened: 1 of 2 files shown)


backend/qemu.pm
```diff
@@ -734,28 +747,18 @@ sub start_qemu ($self) {
         # of ports is 32 so enough for 14 workers per host.
         $vars->{VDE_PORT} ||= ($vars->{WORKER_ID} // 0) * 2 + 2;
     }
-
-    # arch discovery
-    my $arch = $vars->{ARCH} // '';
```
Martchus, line 739: Storing the arch just once in a local variable made actually more sense than the repeated hash accesses and the repeated `// ''` operations. (But I don't really care.)

backend/qemu.pm
```diff
@@ -614,6 +624,14 @@ sub _set_graphics_backend ($self, $is_arm) {
     sp('device', "${device}${options}");
 }
 
+sub get_qemu_version($self, $qemubin) {
```
Martchus, line 627: This is not a simple getter (as it actually invokes a sub process) so maybe `determine_…` would make more sense:
```suggestion
sub determine_qemu_version($self, $qemubin) {
```

backend/qemu.pm
```diff
@@ -580,17 +580,16 @@ sub setup_tpm ($self, $arch) {
     }
 }
 
-sub _set_graphics_backend ($self, $is_arm) {
+sub is_arm ($arch = '') { ($arch // '') =~ /armv6|armv7|arm|aarch64/ }
+sub is_ppc ($arch = '') { ($arch // '') =~ /ppc/ }
+sub is_riscv ($arch = '') { ($arch // '') eq 'risv64' }
+sub is_s390x ($arch = '') { ($arch // '') eq 's390x' }
+sub is_x86 ($arch = '') { ($arch // '') eq 'i586' || ($arch // '') eq 'x86_64' }
```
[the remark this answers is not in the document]
okurz: ok, done

backend/qemu.pm
okurz: sure. I re-introduced the arch variable

backend/qemu.pm
```diff
@@ -614,6 +624,14 @@ sub _set_graphics_backend ($self, $is_arm) {
     sp('device', "${device}${options}");
 }
 
+sub get_qemu_version($self, $qemubin) {
+    my $qemu_version = qx{$qemubin -version};
+    $qemu_version =~ /([0-9]+([.][0-9]+)+)/;
+    $qemu_version = $1;
+    $self->{qemu_version} = $qemu_version;
+    bmwqemu::diag "qemu version detected: $self->{qemu_version}";
```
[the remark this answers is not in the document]
okurz: sure, done

backend/qemu.pm
okurz: done

Martchus (approved):
All points I mentioned in my previous review have been addressed and the code looks good - so now only tests need to be adapted.

kalikiana (approved):
Nice

PR #2490, opened 2024-04-18
Add fallback for `PRODUCTDIR` under `CASEDIR/products/DISTRI`

Martchus:
This helps locating the `PRODUCTDIR` if it is under `products`. This is so far done by the openQA worker but it doesn't work if `CASEDIR` comes from Git and would produce wrong results in our setup for the test distribution of openQA itself. Since the Git checkout is only available within os-autoinst to do this lookup here and remove it from openQA via https://github.com/os-autoinst/openQA/pull/5582 for fully Git-based tests.

Related ticket: https://progress.opensuse.org/issues/159168


okurz (commented):
After thinking a lot about this I think it's a good idea to do this. So please extend tests accordingly.

okurz (approved):

baierjan (approved):

PR #2491, opened 2024-04-24
MM setup: Ensure zone of uplink device is set correctly

Martchus:
See https://progress.opensuse.org/issues/159414


Martchus:
I invoked this on aarch64-o3 after reverting my manual change for this from yesterday and it worked. At least for wicked-based system I can confirm that this is all that was needed (as I rebooted aarch64-o3 yesterday after doing this change manually and then ran verification jobs).

I'm not sure whether something similar is required for the NetworkManager code path (but probably not).

kalikiana (approved):

PR #2494, opened 2024-05-02
Prevent cloning from 'file' git URL schemes

asdil12:
This is introduced as a security precaution
to prevent leaking data.


Martchus (approved):

okurz (changes requested):
I don't want to generally prevent that. Maybe we can restrict to certain path prefixes? And don't our tests rely on that?

asdil12:
As later in the test distri we can execute arbitrary code anyway, I think this doesn't actually make sense.

PR #2495, opened 2024-05-06
Update actions versions to latest release

baierjan:
We have some Node.js 16 deprecation warnings in our workflows. Upgrading the actions should solve that.


perlpunk (approved):

kalikiana (approved):
Feels a little ugly but I guess it's how GHA implements it so thank you for adding the docs to explain it.

PR #2499, opened 2024-06-07
Lower GIT_CACHE_DIR_LIMIT used in the tests

AdamWill:
On Fedora, the test fails when this is set to 100; the expected cleanup does not happen. If I drop it to 50, though, the cleanup happens and the test passes. I'm not sure why the difference.

### DIFF
--- a/t/34-git.t
+++ b/t/34-git.t
@@ -191,7 +191,7 @@ subtest 'cloning with caching' => sub {
             ok -d $repo_cache_dir, 'no cleanup has happened yet';
         };
         subtest 'over limit' => sub {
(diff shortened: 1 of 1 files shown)


AdamWill:
I didn't look into this in much detail, just realized that lowering the threshold 'fixes' things, so I'm sending this. If it's really really expected to pass at 100, and the fact that it doesn't indicates a problem, we can look in more detail...

okurz (approved):
I am fine with that. The value selection in before was also arbitrary so I guess changing to 50 is ok.

Martchus (approved):

PR #2500, opened 2024-06-10
ustreamer: restrict to aarch64 and x86_64

AdamWill:
The use of `unpack` in the ustreamer code assumes a 64-bit, little-endian environment. It fails on s390x (because it's big- endian) and i686 (because it's 32-bit). In practice, ustreamer is really only used on aarch64. Let's bail unless we're running on aarch64 or x86_64, and similarly disable the tests outside of those arches. We keep x86_64 for the convenience of testing in CI environments and on developers' local machines.

Related progress issue: https://progress.opensuse.org/issues/161969

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -9,6 +9,7 @@ use Mojo::UserAgent;
 use Mojo::URL;
 use Mojo::Util 'scope_guard';
 
+use Config;
 use List::Util 'max';
 use Time::HiRes qw(usleep clock_gettime CLOCK_MONOTONIC);
 use Fcntl;
@@ -146,6 +147,7 @@ sub connect_remote_video ($self, $url) {
     }
 
     if ($url =~ m^ustreamer://^) {
+        die 'unsupported arch' unless ($Config{archname} =~ /^aarch64|x86_64/);
         my $dev = ($url =~ m^ustreamer://(.*)^)[0];
         my $sink_name = "raw-sink$dev";
         $sink_name =~ s^/^-^g;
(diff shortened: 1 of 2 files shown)


consoles/video_stream.pm
```diff
@@ -146,6 +147,7 @@ sub connect_remote_video ($self, $url) {
     }
 
     if ($url =~ m^ustreamer://^) {
+        die 'unsupported arch' unless ($Config{archname} =~ /^aarch64|x86_64/);
```
kalikiana, line 150: I guess we need the host architecture. Was wondering since we never used this so far.

t/26-video_stream.t
```diff
@@ -151,6 +155,12 @@ subtest 'frame parsing - ustreamer' => sub {
     # ustreamer requires pack("D") support, not availabe in openSUSE Leap 15.5's Perl
     eval { $_ = pack("D", 1.0); };
     plan skip_all => 'packing long double is not supported' if $@;
+    # frame unpacking is only correct on little-endian 64-bit arches, in
+    # practice ustreamer is only used on aarch64, restrict tests to
+    # aarch64 and x86_64 (x86_64 for CI convenience)
+    # see https://progress.opensuse.org/issues/161969
```
kalikiana, line 161: Please also add something like `See: https://progress.opensuse.org/issues/161969` to your PR desc/commit message.

kalikiana (changes requested):

consoles/video_stream.pm
AdamWill: well, don't we need the arch of whatever system does the `unpack`? which is the same process as hits this code, if I'm reading it right.

Martchus (approved):

t/26-video_stream.t
kalikiana: @AdamWill This was the "requested change" I meant to ask for.

okurz (approved):

kalikiana (approved):

PR #2501, opened 2024-06-11
Add configurable secret key hiding support

okurz:
Related progress issue: https://progress.opensuse.org/issues/162083


asdil12 (approved):
looks good so far

okurz:
In https://github.com/os-autoinst/openQA/pull/5691 @Martchus proposed a different name. I will await further reviews and ideas and then adapt this PR accordingly

doc/backend_vars.asciidoc
```diff
@@ -55,6 +55,7 @@ DEFAULT_CLICK_SLEEP;float;0.15;Default single click time in seconds
 DEFAULT_DCLICK_SLEEP;float;0.10;Default double/triple click time in seconds (both press time and interval between clicks)
 GIT_CACHE_DIR;string;;If set enables locally caching Git repositories in the specified directory when handling Git URLs in variables like `CASEDIR` and wheels
 ENABLE_MODERN_PERL_FEATURES;boolean;0;Enables use of modern Perl features in test modules avoiding the need to use e.g. `use Mojo::Base 'basetest', -signatures;` in all test modules. This variable must be set before invoking `autotest::loadtest`. It only applies to the test modules themselves. It does *not* apply to e.g. `main.pm` and other Perl modules used via e.g. `use some::module`.
+_HIDE_SECRETS_REGEX;string;;If set any test variables matching the specified regular expression additional to the default '^_SECRET_' and '_PASSWORD' are excluded from being saved into vars.json or further processing. This is intended for custom test variable keys that are to be treated as secret but can not match the default patterns.
```
m-dati: In `_HIDE_SECRETS_REGEX` it is considered 'regexpr' a string that can be the initial left part of a PARAMETER NAME, that is a **specific expression** and NOT a generic _regular expression_.
Moreover, we're describing a part of a variable NAME and not its content value, but the actual description could lead someone to misunderstandings about matching the value rather than the name.

Therefore I'd suggest to explicitly mention that the new parameter contains a customizable string that is the left part of a `PARAMETER NAME`, used to store secrets.

m-dati: Moreover, it is important to add a comma in `If set, any test ...`, because without it, the sentence could be read referring to test, in `if set any test ...`, instead of the new parameter

okurz: ok, done

gauravpathak:
Is it possible to trigger a validation run on o3 or OSD using the code from this PR? I’m curious if `CASEDIR` can be pointed directly to this branch, or if it’s restricted to the main `os-autoinst-distri-opensuse` repository.

PR #2506, opened 2024-06-21
baseclass: Use exponential backoff for SSH console

kalikiana:
The machine may take time to come up. Let's increase the delay accordingly instead of constantly retrying.

See: https://progress.opensuse.org/issues/162272


backend/baseclass.pm
```diff
@@ -1226,8 +1228,9 @@ sub new_ssh_connection ($self, %args) {
         }
         else {
             bmwqemu::diag "Could not connect to $con_pretty, Retrying after some seconds...";
```
Martchus, line 1230: It would now make sense to state the number of seconds in this message.

doc/backend_vars.asciidoc
```diff
@@ -27,6 +27,7 @@ SCREENSHOTINTERVAL;float;0.5;The interval in seconds at which screenshots are ta
 SSH_COMMAND_TIMEOUT_S;integer;300;Timeout for any SSH based command in SSH based consoles, disabled for a value of 0. Time in seconds.
 SSH_CONNECT_RETRY;integer;5;Maximum retries to connect to SSH based console targets
 SSH_CONNECT_RETRY_INTERVAL;float;10;Interval in seconds between retries to connect to SSH based console targets. Related to SSH_CONNECT_RETRY
+SSH_CONNECT_RETRY_FACTOR;integer;2;Support for exponential backoff. Factor by which the SSH_CONNECT_RETRY_INTERVAL is increased with each failed attempt.
```
Martchus, line 30:
```suggestion
SSH_CONNECT_RETRY_FACTOR;float;2;Support for exponential backoff. Factor by which the SSH_CONNECT_RETRY_INTERVAL is increased with each failed attempt.
```

Martchus (changes requested):

perlpunk (approved):

backend/baseclass.pm
```diff
@@ -1207,8 +1207,10 @@ sub new_ssh_connection ($self, %args) {
     # timeout requires libssh2 >= 1.2.9 so not all versions might have it
     my $ssh = Net::SSH2->new(timeout => ($bmwqemu::vars{SSH_COMMAND_TIMEOUT_S} // 5 * ONE_MINUTE) * 1000);
 
-    # Retry multiple times, in case of the guest is not running yet
-    my $counter = $bmwqemu::vars{SSH_CONNECT_RETRY} // 5;
+    # Retry multiple times in case the guest is not running yet
+    my $counter = $bmwqemu::vars{SSH_CONNECT_RETRY} // 7;
+    my $interval = $bmwqemu::vars{SSH_CONNECT_RETRY_INTERVAL} // 10;
+    my $factor = $bmwqemu::vars{SSH_CONNECT_RETRY_FACTOR} // 2;
```
okurz, line 1213: This combination does not make sense to me. If the "guest is not running yet" then it's more likely to be up the later we try. So rather than exponentially increasing backoff it could be the inverse. To keep our change simple I am reconsidering: How about you keep your change about making the sleep interval more explicit and in a separate commit increase the retry, but don't introduce the factor

okurz (changes requested):

backend/baseclass.pm
kalikiana: During the estimation we discussed that the test fails because the machine takes longer to come up. In these situations we use exponential backoff.

kalikiana: Alternatively there's a suggestion to simply increase retries for this test scenario. Maybe we don't want a generic solution afterall.

okurz: > During the estimation we discussed that the test fails because the machine takes longer to come up. In these situations we use exponential backoff.
Well, we use it in case of network problems. Like if something fails try immediately, if it still does not work then maybe something is unavailable for longer in the network. In some cases we cover over potential outages of hosts. But here it's different because the host we want to reach was triggered for reboot by the very same execution flow. So that's making it more likely that a host is reachable the later you are in the waiting process unlike in other cases where we don't know why we can't reach a remote end.

> Alternatively there's a suggestion to simply increase retries for this test scenario. Maybe we don't want a generic solution afterall.
yes, that's what I suggested to do but in a separate commit. If you prefer you can also just bump the retry and revert all other changes although I find your change to make the sleep period explicit a good idea.

Martchus: Making it exponential is maybe really not so helpful here. In addition to what @okurz said, I think the tries here are also pretty cheap - so it isn't really problematic to do many frequent retries.

kalikiana: Not sure about that separate commit. I assume the scenario-specific retry will need to be done in a job template somewhere.

kalikiana: I guess that would be https://openqa.suse.de/group_overview/131

kalikiana: https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1703

Martchus (commented):
Now the PR title isn't in-line with the commit message and the commit message doesn't follow our style (there's no verb in imperative).

It also looks like it only changes the log message and the code a little bit. Those changes are good but the commit message should be in-line with that.

okurz (approved):

Martchus (approved):

PR #2507, opened 2024-06-21
Restore running of individual tests via test-perl-testsuite

kalikiana:
Passing arguments to tools/invoke-tests overrides TESTS and makes it impossible to run individual tests as we document in the README.md, for example:

    make test-perl-testsuite TESTS="t/23-baseclass.t"

Without this change all tests from t are always executed.

### DIFF
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -93,12 +93,14 @@ endif ()
 find_program(PROVE_PATH prove)
 find_program(UNBUFFER_PATH unbuffer)
 if (PROVE_PATH)
-    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}" t)
+    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
     add_test(
         NAME test-perl-testsuite
         COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
+    set_tests_properties(test-perl-testsuite
+        PROPERTIES ENVIRONMENT "TESTS=t")
     add_test(
         NAME test-local-author-perl
         COMMAND "${PROVE_PATH}" xt


Martchus (approved):
I suppose you tested that.

perlpunk (changes requested):
See my above comment

kalikiana:
> I suppose you tested that.

Yes. I'm also cross-checking with `./tools/inv
oke-tests --unbuffer-tool=''` to check the script itself.

> Without it it would run any `.t` file anywhere in the repo, which is wrong. There must be another way to override this. By default it must only run tests from `t/`.

Apparently [that's prove's default](https://metacpan.org/dist/Test-Harness/view/bin/prove#Default-Test-Directory). Supplying a fallback for TESTS from within invoke-tests seems to address that.

tools/invoke-tests
```diff
@@ -58,6 +58,8 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 [[ "$*" ]] && TESTS="$*"
+# Run all tests from t by default (instead of all .t files)
+[[ -z "$TESTS" ]] && TESTS=t
```
perlpunk, line 62: Hm, the problem is, below in line 81 we have 
```
TESTS=${TESTS:-"-r t xt"}
```

kalikiana: 🤦🏼‍♀️

tools/invoke-tests
```diff
@@ -77,8 +77,8 @@ args=()
 # use unbuffer if it was found so e.g. timestamps on OBS builds will be useful
 # note: When executing tests with prove it seems that the output is only flushed after one test has finished unless there's a terminal.
 [[ $unbuffer_path ]] && [[ $unbuffer_path != 'UNBUFFER_PATH-NOTFOUND' ]] && args+=("$unbuffer_path")
-# split TESTS on white-spaces to allow specifying multiple tests as documented
-TESTS=${TESTS:-"-r t xt"}
+# run tests from t only by default
+TESTS=${TESTS:-"t"}
```
perlpunk, line 81: ok, I guess that will work. The `-r` would be needed when we have subdirectories with tests though.

perlpunk (approved):

tools/invoke-tests
okurz, line 81: I don't understand this part. When you just specify "t" that means we now exclude "xt" tests?

okurz (changes requested):

tools/invoke-tests
kalikiana: Yes. From the default which test-perl-testsuite relies on.

I guess that doesn't need to change, though. I'm passing TESTs= instead.

cmake/test-targets.cmake
```diff
@@ -93,10 +93,10 @@ endif ()
 find_program(PROVE_PATH prove)
 find_program(UNBUFFER_PATH unbuffer)
 if (PROVE_PATH)
-    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}" t)
+    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
     add_test(
         NAME test-perl-testsuite
-        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS} TESTS=t
```
kalikiana, line 99: Seems like this trips up the "test" target somehow. Not sure yet how to fix it. Or maybe I need to adapt `invoke-tests` afterall 🤔

okurz: Well, invoke-tests takes the test file paths as command line arguments directly *but* there is also an *environment* variable TESTS so you would need to use

```suggestion
        COMMAND "TESTS=t ${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
```

if cmake allows to use a variable assignment for an application. Otherwise use

```suggestion
        COMMAND "env TESTS=t ${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
```

and that would work for invoke-tests as the command line parameter if specified takes precedence over an environment variable.
However then you again would harcode the variable TESTS and do not have a possibilty to override that on all so you would not have gained anything. How about just changing the README to recommend `prove -I. t/15-logging.t t/28-signalblocker.t` for the customized command?

tools/invoke-tests
okurz: Still the problem is that now xt is not executed anymore, see https://github.com/os-autoinst/os-autoinst/actions/runs/9679787998/job/26706502005?pr=2507

perlpunk: > Still the problem is that now xt is not executed anymore, see https://github.com/os-autoinst/os-autoinst/actions/runs/9679787998/job/26706502005?pr=2507

That's what we want. the xt folder is covered in the author tests.

okurz: ah, right. But I would very much prefer if the default for invoke-tests is to run all the tests. Why should invoke-tests just run for "t"?

okurz:
@kalikiana please rephrase the git commit message subject line to state the why, not the what.

tools/invoke-tests
perlpunk: The default is to run `-r t xt` now, and this PR doesn't change this (anymore).

okurz: true, not included anymore

cmake/test-targets.cmake
```diff
@@ -93,12 +93,14 @@ endif ()
...
 find_program(UNBUFFER_PATH unbuffer)
 if (PROVE_PATH)
-    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}" t)
+    set(INVOKE_TEST_ARGS --prove-tool "${PROVE_PATH}" --make-tool "${CMAKE_MAKE_PROGRAM}" --unbuffer-tool "${UNBUFFER_PATH}" --build-directory "${CMAKE_CURRENT_BINARY_DIR}")
     add_test(
         NAME test-perl-testsuite
         COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/invoke-tests" ${INVOKE_TEST_ARGS}
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
+    SET_TESTS_PROPERTIES(test-perl-testsuite
```
Martchus, line 102:
```suggestion
    set_test_properties(test-perl-testsuite
```

Martchus (approved):

okurz (approved):

PR #2508, opened 2024-06-24
Refactor: generate python bindings from template

askebm:
Instead of copying code use a template to generate the python bindings


CMakeLists.txt
```diff
@@ -202,6 +204,13 @@ install(
 foreach (FILE README.asciidoc COPYING ${GENERATED_DOC_FILES})
     install(FILES "${FILE}" DESTINATION "${OS_AUTOINST_DOC_DIR}")
 endforeach ()
+foreach (PYTHON_BINDING IN LISTS PYTHON_BINDINGS)
+    configure_file("api.py.in" "${PYTHON_BINDING}.py")
+    install(
+        FILES "${CMAKE_CURRENT_BINARY_DIR}/${PYTHON_BINDING}.py"
+        DESTINATION "${OS_AUTOINST_DATA_DIR}"
+    )
```
Martchus, line 212: It would be nice if - in addition to installing those files - the custom `symlinks` target would create symlinks of those generated `*.py` files into the root of the source checkout. Of course `.gitignore` should be extended to ignore those symlinks. (Search for `# allow symlinking created executables into source directory` in `CMakeLists.txt` for the relevant target and further comments on it. We have this target for convenience but it is not invoked by default to avoid polluting the source directory with build artifacts by default.)

CMakeLists.txt
```diff
@@ -202,6 +204,13 @@ install(
 foreach (FILE README.asciidoc COPYING ${GENERATED_DOC_FILES})
     install(FILES "${FILE}" DESTINATION "${OS_AUTOINST_DOC_DIR}")
 endforeach ()
+foreach (PYTHON_BINDING IN LISTS PYTHON_BINDINGS)
+    configure_file("api.py.in" "${PYTHON_BINDING}.py")
```
Martchus, line 208: We already have many files at the root. Maybe this is a chance to tidy that up a little bit, e.g.:
```suggestion
    configure_file("cmake/templates/api.py.in" "${PYTHON_BINDING}.py")
```
(And of course move the actual file.)

Martchus (commented):
Looks generally good. I have two ideas for improvement (see inline-comments).

CMakeLists.txt
askebm: Great idea :+1:  I've done as you suggested, it has also fixed the tests locally, so now I hope the the github ci also passes  :crossed_fingers:

CMakeLists.txt
askebm: :+1:

Martchus (approved):
Thanks for implementing the symlinking. Seems that it in fact helped with the tests.

okurz (approved):
awesome!

PR #2509, opened 2024-07-05
Refactor isotovideo

okurz:
* isotovideo: Extract 'handle_shutdown' method
* Extract proper isotovideo 'runner::init' method instead of private call

### DIFF
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -245,7 +245,7 @@ sub _init_bmwqemu ($, @args) {
 }
 
 sub init ($self, @args) {
-    $self->_init_bmwqemu(@ARGV);
+    $self->_init_bmwqemu(@args);
     $self->prepare;
     $self->start_autotest;
     $self->create_backend;
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -244,6 +244,13 @@ sub _init_bmwqemu ($, @args) {
     }
 }
 
+sub init ($self, @args) {
+    $self->_init_bmwqemu(@ARGV);
+    $self->prepare;
+    $self->start_autotest;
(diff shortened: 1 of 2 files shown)


OpenQA/Isotovideo/Runner.pm
```diff
@@ -244,6 +244,13 @@ sub _init_bmwqemu ($, @args) {
     }
 }
 
+sub init ($self, @args) {
+    $self->_init_bmwqemu(@ARGV);
+    $self->prepare;
+    $self->start_autotest;
+    $self->create_backend;
+}
+
```
Martchus, line 253: I'd rather just add that to `new` so we don't need to call `new` + `init` separately.

Martchus (changes requested):

OpenQA/Isotovideo/Runner.pm
okurz: ok I tried but failed. Because then at least t/19-isotovideo-command-processing.t fails as it wants to instantiate a clean runner object and not initialize everything. Also "new" is the constructor and IMHO shouldn't do that much if any dynamic evaluation but "init bmwqemu" and "start autotest" is already doing a lot. With that I prefer the separate `init`. WDYT?

OpenQA/Isotovideo/Runner.pm
```diff
@@ -244,6 +244,13 @@ sub _init_bmwqemu ($, @args) {
     }
 }
 
+sub init ($self, @args) {
+    $self->_init_bmwqemu(@ARGV);
```
perlpunk, line 248:
```suggestion
sub init ($self, @args) {
    $self->_init_bmwqemu(@args);
```

okurz: done

perlpunk (approved):

Martchus (approved):

PR #2511, opened 2024-07-17
Improve lib lookup in check_needles.pl

okurz:
Same as in other scripts we can allow check_needles.pl to find libs on
its own as well.

### DIFF
--- a/check_needles.pl
+++ b/check_needles.pl
@@ -3,6 +3,10 @@
 use Mojo::Base -strict, -signatures;
 use File::Basename;
 use Getopt::Long;
+
+use FindBin '$Bin';
+use lib "$Bin";
+
 use needle;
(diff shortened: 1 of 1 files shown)


perlpunk:
I realized the script has no `pod`, so `check_needles.pl -h` outputs nothing. Could you add that as well, in this PR or a new one? I wouldn't know what to write as I haven't used it so far.

okurz:
> I realized the script has no `pod`, so `check_needles.pl -h` outputs nothing. Could you add that as well, in this PR or a new one? I wouldn't know what to write as I haven't used it so far.

I honestly didn't care about the funtionality. I just found that while working on #1978 but after checking the script I guess it's better if we just remove it completely:
* #2512

PR #2514, opened 2024-07-25
video_stream: handle HID command confirmations

marmarek:
When sending commands to stdin of GENERAL_HW_INPUT_CMD, wait for it to
confirm reception/execution. Not waiting may cause some commands
queueing up, which will break timing in some cases (for example, a sleep
between clicks will not work, if both clicks happen to wait in a stdin
pipe one after another). Normally, HID commands are handled quickly, but
recently I've got a case where mouse_move may take a bit more time
(emulating real mouse with only relative movement, by moving it step by
 step to the desired position, and try to not hit pointer acceleration).

This is a change in the protocol used for GENERAL_HW_INPUT_CMD -
previously it didn't need to print anything to stdout, and now it needs
to print "ok" after handling every command. Scripts not updated to this
new protocol will result in a test hang at the first input. Since this
isn't broadly used feature, this API change hopefully is acceptable.

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -14,6 +14,7 @@ use List::Util 'max';
 use Time::HiRes qw(usleep clock_gettime CLOCK_MONOTONIC);
 use Fcntl;
 use File::Map qw(map_handle unmap);
+use IPC::Open2 qw(open2);
 
 use Try::Tiny;
 use bmwqemu;
@@ -55,6 +56,7 @@ sub disable ($self, @) {
     my $ret = $self->disable_video;
     if ($self->{input_pipe}) {
         close($self->{input_pipe});
+        close($self->{input_feedback});
         waitpid($self->{inputpid}, 0);
     }
     return $ret;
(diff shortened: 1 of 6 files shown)


kalikiana:
@mergifyio rebase

Since I'm guessing this would benefit from #2517

marmarek:
```
 - File './repodata/66876b5e579532a10a899bac694d9201888585ea45673444b0db8e4ff77d1bd8-primary.xml.gz' not found on medium 'http://download.opensuse.org/repositories/devel:openQA/15.6'
```
?!

okurz:
> ```
>  - File './repodata/66876b5e579532a10a899bac694d9201888585ea45673444b0db8e4ff77d1bd8-primary.xml.gz' not found on medium 'http://download.opensuse.org/repositories/devel:openQA/15.6'
> ```
> 
> ?!

yeah that's a common problem we are facing and so far the mirror admins and OBS experts couldn't help us to fix it. The best we can do for now is more automatic retries all over. But anyway we would like to see a second approval by another reviewer before merge.

Martchus (approved):

PR #2513, opened 2024-07-17
Add boot sector to disks poo#164018 

foursixnine:

PR #2516, opened 2024-07-26
Add support to configure sector sizes of HDD images

okurz:
Inspired by
* #2513

Created as a result from the joint Co-working session in the SUSE QE Tools workshop 2024-07-26.

@foursixnine your review would be appreciated.

Related progress issue: https://progress.opensuse.org/issues/164018

### DIFF
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -67,6 +67,14 @@ The number of I/O queues of the drive, esp. for NVMe devices
 =cut
 has 'num_queues';
 
+=head3 sector_size
+
+The logical and physical block size of the drive if differing from QEMU
+default.
+
+=cut
+has 'sector_size';
+
 sub new_overlay_id ($self) { $self->last_overlay_id($self->last_overlay_id + 1)->last_overlay_id }
 
 sub node_name ($self) { $self->id . DEVICE_POSTFIX }
@@ -98,6 +106,7 @@ sub gen_cmdline ($self) {
         $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex) if (!$path->id || $path->id eq 'path0');
         $self->_push_ifdef(\@params, 'serial=', $self->serial);
         $self->_push_ifdef(\@params, 'num_queues=', $self->num_queues) if (($self->num_queues // -1) != -1);
+        push(@params, 'logical_block_size=' . $self->sector_size . ',physical_block_size=' . $self->sector_size) if $self->sector_size;
         push(@cmdln, ('-device', join(',', @params)));
     }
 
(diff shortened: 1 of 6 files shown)


doc/backend_vars.asciidoc
```diff
@@ -118,6 +118,7 @@ HDDMODEL_$_;see qemu -device ?;virtio-blk;Storage device for virtualized HDD. Ov
 HDDSIZEGB;integer;10;Creates HDD with specified size in GiB
 HDD_$i;filename;;Filename of HDD image to be used for VM. Up to 9
 HDDNUMQUEUES_$i;integer;-1;see qemu-system-x86_64 -device nvme,help - set the number of queues for HDD_$i
+HDDSECTORSIZE_$i;integer;undef;specifies the physical and logical block size and boot sector size of the HDD image
```
nicksinger, line 121: I don't fully understand the call order in these files but from what I get we introduce one new option to basically control two variables (`logical_block_size` and `physical_block_size`). Is there a reason to abstract these two values behind just one?

OpenQA/Qemu/DriveDevice.pm
```diff
@@ -98,6 +106,7 @@ sub gen_cmdline ($self) {
         $self->_push_ifdef(\@params, 'bootindex=', $self->bootindex) if (!$path->id || $path->id eq 'path0');
         $self->_push_ifdef(\@params, 'serial=', $self->serial);
         $self->_push_ifdef(\@params, 'num_queues=', $self->num_queues) if (($self->num_queues // -1) != -1);
+        push(@params, 'logical_block_size=' . $self->sector_size . ',physical_block_size=' . $self->sector_size) if $self->sector_size;
```
nicksinger, line 109: would it make sense to have even a simple sanity check? E.g. `physical_block_size` >= `logical_block_size`? Or does it totally not matter and both can be freely chosen?

nicksinger (approved):

OpenQA/Qemu/DriveDevice.pm
foursixnine, line 109: I'm not sure this is a good idea, but we can try, physcal block size and logical block size are two different things, although having the same block size seems to be supported:

- https://utcc.utoronto.ca/~cks/space/blog/tech/SCSIBlocksizesDiscovery
- https://www.ibm.com/docs/en/linux-on-systems?topic=prr-valid-physical-block-size-combinations-lvm-physical-volumes

```suggestion
        push(@params, 'logical_block_size=' . $self->sector_size . ',physical_block_size=' . $self->sector_size) if $self->sector_size;
```

foursixnine (approved):

OpenQA/Qemu/DriveDevice.pm
okurz: I think at this level we don't have any restrictions. We simply pass the settings to qemu so let qemu handle any inconsistencies I would say :)

kalikiana: It is very common because it is usually faster - the question would be if we want to test systems where it's not configured the same.

okurz: Similar as in https://github.com/os-autoinst/os-autoinst/pull/2516#discussion_r1692684214 . If you guys say we need two separate settings then we can do that.

I can't spot the actual diff in your suggestion. What do you propose to change there?

kalikiana: I was only confirming matching the size is the common case. If we have no use case for specifying two values I'd say we can always add HDDLOGICALSIZE later

kalikiana (approved):

okurz:
@mergify rebase

PR #2520, opened 2024-07-30
testapi: Keep the positional mode for script_run and alike

okurz:
In 7185ac80 we declared the positional mode of calling functions as
"deprecated" but did not plan to actually remove that mode. By now we
can acknowledge that we can support the positional mode indefinitely but
still discourage its use.

Related progress issue: https://progress.opensuse.org/issues/164541

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -810,7 +810,7 @@ sub is_serial_terminal () {
 
   wait_serial($regex or ARRAYREF of $regexes, [, timeout => $timeout] [, expect_not_found => $expect_not_found] [, %args]);
 
-Deprecated mode
+Positional mode (not suggested)
 
   wait_serial($regex or ARRAYREF of $regexes [, $timeout [, $expect_not_found [, @args ]]]);
 
@@ -892,7 +892,7 @@ sub _handle_script_run_ret {    # no:style:signatures
 
   assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
 
-Deprecated mode
+Positional mode (not suggested)
 
   assert_script_run($cmd [, $timeout [, $fail_message]]);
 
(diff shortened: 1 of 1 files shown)


kalikiana:
> Related progress issue: https://progress.opensuse.org/issues/162035

That ticket is **Use needles from correct ref of NEEDLES_DIR**. How is removing the deprecation related to that? And what is the motivation to not remove this afterall?

okurz:
> > Related progress issue: https://progress.opensuse.org/issues/162035
> 
> That ticket is **Use needles from correct ref of NEEDLES_DIR**. How is removing the deprecation related to that?

True, see https://progress.opensuse.org/issues/164541

> And what is the motivation to not remove this afterall?

Backward compatibility. Actually deprecating means introducing a deprecation warning, informing users, helping them to update deprecated code, wait at least a year, remove the old behavior, get yelled at because people ignore warnings, fix some more code, etc.

kalikiana:
> True, see https://progress.opensuse.org/issues/164541
> 
> > And what is the motivation to not remove this afterall?
> 
> Backward compatibility. Actually deprecating means introducing a deprecation warning, informing users, helping them to update deprecated code, wait at least a year, remove the old behavior, get yelled at because people ignore warnings, fix some more code, etc.

https://progress.opensuse.org/issues/164541#note-11 I am reading the comment very differently.

I guess I don't have a strong opinion. As stated we already deprecated the code and we could move forward. Please make sure everyone is on the same page, though.

okurz:
> https://progress.opensuse.org/issues/164541#note-11 I am reading the comment very differently.

One problem regarding the ticket was that people wrongly assumed we would obsolete the positional mode but the deprecation warning that they received was about the "die_on_timeout" behaviour. I created this PR so that the first part would cause confusions while within https://progress.opensuse.org/issues/164541 the work will continue to accept the "die_on_timeout => 1" behaviour as default.

kalikiana (approved):

Martchus (approved):

PR #2521, opened 2024-07-30
Remove obsolete support for "script_run" with "die_on_timeout=0"

okurz:
The correct and expected way to call script_run is

```
script_run($cmd [, timeout => $timeout] [, output => ''] [, quiet => $quiet])
```

which will die on timeout. The "die on timeout" is the same behaviour as for
other commands like assert_screen and such. As the old behaviour had the
deprecation message in place for more than one year now we should remove
support for the obsolete way.

The idea of `script_run` is to await the execution of a command and return the
result. If one does not care about the result and does not want to wait for the
execution to be done then one can still just use `enter_cmd`. This would only
be a problem if any subsequent command can not handle that the previous one did
not finish. As alternative consider

```
script_run 'timeout $lower_timeout $cmd', timeout => $higher_timeout
```

Another but not suggested way would be

```
eval { script_run($cmd) };
if ($@) { $handle_timeout }
```

Related progress issue: https://progress.opensuse.org/issues/164712

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 41;
+our $version = 42;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -311,18 +311,17 @@ subtest 'script_run' => sub {
     );
     $fake_exit = 0;
     $cmds = [];
-    is(script_run('true', die_on_timeout => 1), '0', 'script_run with no check of success, returns exit code');
+    is(script_run('true'), '0', 'script_run with no check of success, returns exit code');
     like($cmds->[1]->{text}, qr/; echo /);
     $cmds = [];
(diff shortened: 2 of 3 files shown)


kalikiana:
> I'm not against it. We'll have to see what this will break and can revert if necessary.

Following [our own documented best practice](https://progress.opensuse.org/projects/qa/wiki/Tools#Best-practices-for-major-changes) I suggest to pick a date and announce it beforehand.

okurz:
> > I'm not against it. We'll have to see what this will break and can revert if necessary.
> 
> Following [our own documented best practice](https://progress.opensuse.org/projects/qa/wiki/Tools#Best-practices-for-major-changes) I suggest to pick a date and announce it beforehand.

I read that again and I consider all of that done now dragging on over 3 years already. I also updated osado so that most relevant tests are already covered.

Regarding a date I recommend some day after 2024-09-10 when I will merge this.

kalikiana:
> Regarding a date I recommend some day after 2024-09-10 when I will merge this.

That's from today. At least two people seconded that [in the Slack announcement](https://suse.slack.com/archives/C02CANHLANP/p1723798592289769).

Please also fix the ticket reference. This should be [poo#164712](https://progress.opensuse.org/issues/164712).

okurz:
> > Regarding a date I recommend some day after 2024-09-10 when I will merge this.
> […]
> Please also fix the ticket reference. This should be [poo#164712](https://progress.opensuse.org/issues/164712).

Fixed

kalikiana (approved):

pdostal (approved):

perlpunk (approved):

PR #2524, opened 2024-08-04
Virsh VMware image script debugging

m-dati:
The [SL Micro 6.0 Product Increments - Containers](https://openqa.suse.de/group_overview/572) VMware test revealed an issue in the
[failed job](https://openqa.suse.de/tests/15040545#step/bootloader_svirt/16), where images in `hdd/fixed` folder are included in initial file search, but then only images in `hdd` folder are managed.

This PR started with scope of adding also `subfolders` management, now focused to only add debugging of a script output.

ticket: https://progress.opensuse.org/issues/162941#note-14

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -16,7 +16,6 @@ use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
 use Mojo::Util;
 use Carp 'croak';
-
 use backend::svirt;
 
 has [qw(instance name vmm_family vmm_type vmm_firmware)];
@@ -330,8 +329,10 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
     # otherwise copy image from NFS datastore.
     my $nfs_dir = $backingfile ? 'hdd' : 'iso';
     my $vmware_nfs_datastore = $bmwqemu::vars{VMWARE_NFS_DATASTORE} or die 'Need variable VMWARE_NFS_DATASTORE';
+    # cmd debugging activable by setting VMWARE_NFS_DATASTORE_DEBUG=1
+    my $ds_debug = ($bmwqemu::vars{VMWARE_NFS_DATASTORE_DEBUG} // 0) ? "set -x;" : "";
     my $cmd =
-      "if test -e $vmware_openqa_datastore$file_basename; then " .
+      "$ds_debug if test -e $vmware_openqa_datastore$file_basename; then " .
       "while lsof | grep 'cp.*$file_basename'; do " .
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
       'done;' .


m-dati:
@jlausuch @okurz please kind check

m-dati:
VR to be added.

consoles/sshVirtsh.pm
```diff
@@ -339,6 +339,7 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       'done;' .
       'else ' .
       "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
+      "[ $? -ne 0 ] && cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/fixed/$file_basename $vmware_openqa_datastore;" .
```
okurz, line 342: No code so far in os-autoinst needs to do anything with the "fixed" directory which is only something that openQA knows. I would like to keep this design this separated. As also your git commit message does not state the reason or what problem this addresses this shouldn't be accepted as-is.

okurz (dismissed):

consoles/sshVirtsh.pm
m-dati: Sorry, it is not clear to me, but `fixed` is a subdirectory of `hdd/` (directory that is included into the os-autinst design), but do not managed (forgotten ?) into that routine: by design _the image is copied from `hdd`, but in this case it is present in the `hdd/fixed` subfolder instead_.
The commit message may be is too short,but it could be updated.
But also, in the PR header I explained the problem occurred, unless explanation was not clear yet: should I better explain that error?

asmorodskyi: > No code so far in os-autoinst needs to do anything with the "fixed" directory which is only something that openQA knows.

Generally I would agree with you Oli , this is implementation details which should be hidden in `os-autoinst` and test code should not be aware about existence of `fixed`  folder. But also I must say that it is "too late" for such comment :man_shrugging: . Major part of things what happen in  `_copy_image_vmware` should happen in os-autoinst and should be hidden from test code . So there are two options how to proceed further : 
- accept the fact that from now on test code will know about `fixed` folder and increase even more tightly coupling of os-autoinst and `osada`
- move fully or partially logic from `osada` into os-autoinst so test code will be able transparently copy image no matter if it is in `fixed` folder or just in `hdd` folder 

P.S. Please bare with me , my knowledge of this piece of code is result of 5 minutes reading of this PR and comments to it so I might missing some obvious things here . So feel free fully or partially ignore it :smile:

okurz: > Sorry, it is not clear to me, but `fixed` is a subdirectory of `hdd/`, directory included into the os-autinst design, but that subfolder is not managed into that routine: by design _the image is copied from `hdd`, but in this case it is present in the `hdd/fixed` subfolder instead_.

A simple `git grep` reveals that "fixed/" is not used anywhere in os-autoinst so far and hence also not used in any other backend. That's why I wonder what could be the right solution to not hard-code the fixed folder path here.

> The commit message may be is too short,but it could be updated. But also, in the PR header I explained the problem occurred, unless explanation was not clear yet: should I better explain that error and needed changes?

What you put into the description of the pull request could also be in the commit message details. I don't know how you created your pull request but your git commit messages has *only* the subject line while the pull request description has more details. Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage.

I can recommend the tool [hub](https://github.com/github/hub) (zypper in rubygem-hub) for easier PR creation. Also, I myself use a script [git-pr-last](http://github.com/okurz/scripts/tree/master/git-pr-last) to create a PR with proper description for these simple one-commit PRs:
```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
git push $target && git show --no-patch --format=%B | hub pull-request -F -
```

You might want to use the more up-to-date tool https://github.com/cli/cli

m-dati:
Sorry,but, as I explained in the related poo note, it is Never _"later on copied to the vmware machine"_, by error, that this PR would like to correct.

m-dati:
> I do not really see the necessity of this change as for [`fixed`](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/installation/bootloader_svirt.pm#L138) is already being scanned for corresponding image that is needed for the tests and later on copied to the vmware machine.

Sorry, but, as I explained in the related poo note and visibile in the failed test, the error in this routine's bash snippet, and it Never _copy to the vmware machine the fixed/image_ by error, that this PR would like to correct.

mloviska:
> Sorry, but, as I explained in the related poo note and visibile in the [failed test,](https://openqa.suse.de/tests/15040545#step/bootloader_svirt/16) the error is in this routine's bash snippet, and it Never _copies to the vmware machine the fixed/image_, by an error that this PR would like to correct.

Right, I see the problem now. The test module actually finds the image in the correct place, but the information is not passed to `_copy_image_vmware`. I think the right solution would be to move the search subroutine to the backend and then pass the image that should be copied to `_copy_image_vmware`

consoles/sshVirtsh.pm
m-dati: Actual problem is that other than `hdd/` folder, its subfolder (`fixed`) too is actually popolated by some images to test, somewhere not managed yet so this misalignment may cause errors.
So, alike in the last poo's note written, clear rules for files in this last folder should be properly defined, documented and every appropriate code updated.

pdostal:
Btw. I just noticed:
```
pdostal@openqa:~> grep fixed /etc/cron.d/SLES.CRON
*/5 * * * * geekotest (cd /var/lib/openqa/share/factory/repo/ && for i in fixed/SLE-12-{,SP?-}Server-DVD-s390x-GM-DVD1/ ; do ln -sf $i ; done)
*/5 * * * * geekotest (cd /var/lib/openqa/share/factory/repo/ && for i in fixed/SLE-12-SP?-SDK-POOL-{x86_64,aarch64,ppc64le,s390x}-BuildGM-Media1/ ; do ln -sf $i ; done)
*/5 * * * * geekotest (cd /var/lib/openqa/share/factory/repo/ && for i in fixed/SLE-15-*-s390x-GM-*1/ ; do ln -sf $i ; done)
```

consoles/sshVirtsh.pm
okurz: Right. So how about a `find` call to recurse into any subfolder?

m-dati:
Activity restarted: 
I noted that in the `_copy_image_to_vm_host()` routine the full path of the source image (including evnetual subfolder like `fixed`) was present in `@args`, but never used and even lost.

The image basename only was passed as parameter in the call of `_copy_image_vmware()`, but the image path was re-calculated without any subfolder management, therefore images in hdd (or iso) subolders were not copied in the expected destination.

Here I replaced both in `_copy_image_to_vm_host()` and `_copy_image_vmware()` the file_basename input param. with the full path source file, available from previous parent calls, so avoiding the source-folder recalculation and becoming able to manage also images in subdir. like 'fixed/'.

consoles/sshVirtsh.pm
m-dati: Restarting the activities I did an attempt to find a desired image under hdd/ and subfolders: that process consumed resources in cpu and took long time to provide result, being many images present, to skip.
Moreover we already know the original file position, provided by [search_image_on_svirt_host](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/800edc414310ac6bafc982c7221eff4589c78137/tests/installation/bootloader_svirt.pm#L46) hence not needed that `find`.

m-dati:
After some issues affected the test group 572, then fixed see [poo165536](https://progress.opensuse.org/issues/165536), the rerun today were ok and also VMware tests pass ok the `bootloader_svirt module`. 
Consider we that no change was applied to test code, therfore no clear the reason why VMware now pass the point, the original image still is in the `hdd/fixed/` subdir _not managed into that bash snippet_, but that image seems _already present_ into the right destination folder, apparently having no process placed it there, in that test run.

Having I long time analyzed that status, but cause still unclear, I consider useful to undo any change until all clarified. 
But I'd like to add the debug setting -x, to allow inspection of the snippet status, lacking any visible output.

New commit coming soon.

m-dati:
Introduced on-demand debugging, by setting `VMWARE_NFS_DATASTORE_DEBUG=1`.

Martchus (approved):

kalikiana (approved):

PR #2525, opened 2024-08-05
video_stream: make FPS configurable

marmarek:
This is all about balance between compute power on the worker host and
desired latency and video smoothness. Ustreamer is a bit lighter than
ffmpeg, so it can capture more frames (on Raspberry Pi), but then it
puts more stress on the system as more frames are searched for needles
and more frames goes to video encoder. On the other hand, lower FPS may
cause some details to be missed (like quickly dismissed notification).
The optimal value may even differ from test to test, for example based
on the test length and how much backlog for video encoder is acceptable
(amount of available RAM is important in this case too).
Instead of trying to find perfect value that fits all, make it
configurable. Keep the default unchanged.

### DIFF
--- a/t/26-video_stream.t
+++ b/t/26-video_stream.t
@@ -96,6 +96,16 @@ subtest 'connect stream' => sub {
         [('/dev/video0', undef, '--set-dv-bt-timings query')],
         [('/dev/video0', undef, '--get-dv-timings')],
     ], "calls to v4l2-ctl";
+
+    my $cmd = $mock_console->original('_get_ffmpeg_cmd')->($console, 'udp://@:5004');
+    is_deeply $cmd, [
+        'ffmpeg', '-loglevel', 'fatal', '-i', 'udp://@:5004',
+        '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 4, '-'], "correct cmd built for UDP source";
+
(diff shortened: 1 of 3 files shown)


t/26-video_stream.t
```diff
@@ -96,6 +96,16 @@ subtest 'connect stream' => sub {
...
+
+    my $cmd = $mock_console->original('_get_ffmpeg_cmd')->($console, 'udp://@:5004');
+    is_deeply $cmd, [
+        'ffmpeg', '-loglevel', 'fatal', '-i', 'udp://@:5004',
+        '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 4, '-'], "correct cmd built";
+
+    $cmd = $mock_console->original('_get_ffmpeg_cmd')->($console, '/dev/video0?fps=3');
+    is_deeply $cmd, [
+        'ffmpeg', '-loglevel', 'fatal', '-i', '/dev/video0',
+        '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 3, '-'], "correct cmd built";
```
okurz, line 108: I suggest to make those test explanations strings unique. How about

```suggestion
        '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 3, '-'], 'correct cmd built for fps=3';
```

same below

Martchus (approved):

okurz (approved):

PR #2526, opened 2024-08-08
Explain regexp argument for wait_serial

perlpunk:


testapi.pm
```diff
@@ -815,6 +815,7 @@ Positional mode (not suggested)
   wait_serial($regex or ARRAYREF of $regexes [, $timeout [, $expect_not_found [, @args ]]]);
 
 Wait for C<$regex> or anyone of C<$regexes> to appear on serial output.
+A regex object can be created with the C<qr> operator.
```
mpagot, line 818:
```suggestion
A regex object has to be created with the C<qr> operator.
```

mpagot (approved):
LGTM

baierjan (approved):

testapi.pm
```diff
@@ -829,6 +830,10 @@ Returns the string matched or C<undef> if C<$expect_not_found> is false
 Returns C<undef> or (after timeout) the string that I<did _not_ match> if
 C<$expect_not_found> is true. The default timeout is 90 seconds.
 
+Example:
+
+    wait_serial(qr/Password:\s*$/i, timeout => 30);
```
okurz, line 835: I would not encourage a lower than default timeout which is often misunderstood. How about

```suggestion
    wait_serial(qr/Password:\s*$/i, timeout => 120);
```

kalikiana: In just `lib/serial_terminal.pm` there is variations of this with 3, 30 and no specific timeout. And looking at other files in osado it's a mix of 30 and defaults - I'd say do not provide any suggestion here unless you can also explain **why** a timeout is recommended

okurz: How about

```
Examples:
    wait_serial qr/Password:\s*$/i;

    wait_serial(qr/Password:\s*$/i, timeout => 120, expect_not_found => 1);
```

perlpunk: The point of an example is not to leave out additional parameters but to show how to use additional parameters. Just to get a quick glance of how it should look like.
It's what I often miss in documentation.
I now made it two examples. Hope that's better now.

kalikiana: Totally agree with adding examples. This was really the specific case.

kalikiana (approved):

PR #2527, opened 2024-08-12
Fix script/ install

okurz:


dist/rpm/os-autoinst-openvswitch-test.spec
```diff
@@ -19,7 +19,7 @@ touch %{_sourcedir}/%{short_name}
 %build
 # call one of the components but not openqa itself which would need a valid
 # configuration
-/usr/lib/os-autoinst/os-autoinst-openvswitch --help
+/usr/lib/os-autoinst/script/os-autoinst-openvswitch --help
```
kalikiana, line 22: What's the context here? Was there a related change that broke tests? If not, how did you find out this was broken?

kalikiana: I guess this might be related to #1978 - no CI failures there 🤔

okurz: Yes,
* #1978

kalikiana: Apparently I can't view OBS test results for that, or this PR.

PR #2529, opened 2024-08-13
VMWare: add combustion and ignition support

mloviska:
In VMWare's `vmx` we can pass provisioning information for the current VM. The configuration data in case of cloud-init or ignition as well as configuration script, in case of combustion has to encoded in base64 at least. Usually users would compress and encode the content of these files. It would be user-friendly if we can pass only the file name that should be located in osado's `data` folder

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -14,6 +14,8 @@ use File::Which;
 use Mojo::DOM;
 use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
+use Mojo::Util;
+use Carp 'croak';
 
 use backend::svirt;
 
@@ -496,6 +498,20 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $config, $key) {
+    # expand path
+    $config = "$bmwqemu::vars{CASEDIR}/data/$config";
+
+    croak "'$config' either doesn't exist or it is not a file, update the $key variable" unless -f $config;
+
+    my $content = Mojo::File->new($config)->slurp;
+    my $gzip = Mojo::Util::gzip $content;
+    my $encoded_config = Mojo::Util::b64_encode($gzip);
+    $encoded_config =~ s/\R//g;
+
+    return $encoded_config;
+}
+
 sub define_and_start ($self) {
     my $remote_vmm = "";
     if ($self->vmm_family eq 'vmware') {
@@ -542,17 +558,33 @@ __END"
 
         # set default boot delay
         $self->run_cmd(qq{echo 'bios.bootDelay = "10000"' >> $vmx}, domain => 'sshVMwareServer');
+        my $fb_tool = $bmwqemu::vars{GUESTINFO_CONFIG};
 
-        # inject cloud init metadata and userdata required for the image if there are any
-        my $ci_meta = $bmwqemu::vars{CLOUD_INIT_META};
-        my $ci_user = $bmwqemu::vars{CLOUD_INIT_USER};
-        my $ci_encoding = $bmwqemu::vars{CLOUD_INIT_ENCODING};
+        if ($fb_tool && $fb_tool ne 'wizard') {
+            my $encoding = 'gzip+base64';
+            if ($fb_tool =~ /combustion|ignition/) {
+                if ($bmwqemu::vars{GUESTINFO_COMBUSTION}) {
+                    my $conf = $self->_encode_config($bmwqemu::vars{GUESTINFO_COMBUSTION}, 'GUESTINFO_COMBUSTION');
+                    $self->run_cmd(qq{echo 'guestinfo.combustion.script = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                }
+                if ($bmwqemu::vars{GUESTINFO_IGNITION}) {
+                    my $conf = $self->_encode_config($bmwqemu::vars{GUESTINFO_IGNITION}, 'GUESTINFO_IGNITION');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                }
+            } elsif ($fb_tool eq 'cloud-init') {
+                croak "GUESTINFO_CLOUD_INIT is unset, or does not contain user-data and meta-data configs" unless ($bmwqemu::vars{GUESTINFO_CLOUD_INIT});
 
-        if ($ci_meta && $ci_user && $ci_encoding) {
-            $self->run_cmd(qq{echo 'guestinfo.metadata = "$ci_meta"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.metadata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata = "$ci_user"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                my ($conf, $meta) = split(',', $bmwqemu::vars{GUESTINFO_CLOUD_INIT});
+                $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                $self->run_cmd(qq{echo 'guestinfo.metadata.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                $conf = $self->_encode_config($conf, 'GUESTINFO_CLOUD_INIT');
+                $self->run_cmd(qq{echo 'guestinfo.userdata = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                $meta = $self->_encode_config($meta, 'GUESTINFO_CLOUD_INIT');
+                $self->run_cmd(qq{echo 'guestinfo.metadata = "$meta"' >> $vmx}, domain => 'sshVMwareServer');
+            } else {
+                croak 'Unknown provisioning option has been passed through GUESTINFO_CONFIG test variable';
+            }
         }
     }
 
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -149,13 +149,29 @@ subtest 'starting VMware console' => sub {
     ], 'expected commands invoked' or diag explain \@cmds;
 };
 
-subtest 'starting VMware console with Cloud Init' => sub {
+subtest 'test config encoding' => sub {
+    $bmwqemu::vars{GUESTINFO_COMBUSTION} = 'someTestScript';
+    my $test_script_dir = path(File::Temp->newdir('testXXXX'));
+    $bmwqemu::vars{CASEDIR} = $test_script_dir->to_string;
+
+    my $test_script = $test_script_dir->child('data')->make_path;
+    $test_script = $test_script->child($bmwqemu::vars{GUESTINFO_COMBUSTION});
+    my $test_string = 'combustion' x 100;
+    $test_script->spew($test_string);
+
+    my $console_mock = Test::MockModule->new('consoles::sshVirtsh');
+    my $conf_encoded = $svirt_console->_encode_config($test_script->basename, 'GUESTINFO_COMBUSTION');
+    like $conf_encoded, qr/^H4sI.*noAwAA$/, 'config encoded';
+    is_deeply length $conf_encoded < length $test_string, 1, 'Encoded string is shorter';
+    dies_ok { $svirt_console->_encode_config('testingFile') } '_encode_config dies with insufficient amount of arguments passed';
+};
+
+subtest 'starting VMware console with InvalidConfig' => sub {
     $bmwqemu::vars{VMWARE_HOST} = 'h';
     $bmwqemu::vars{VMWARE_USERNAME} = 'u';
     $bmwqemu::vars{VMWARE_PASSWORD} = 'p';
-    $bmwqemu::vars{CLOUD_INIT_META} = 'test@meta';
-    $bmwqemu::vars{CLOUD_INIT_USER} = 'test%user';
-    $bmwqemu::vars{CLOUD_INIT_ENCODING} = 'gzip+base64';
+    $bmwqemu::vars{GUESTINFO_COMBUSTION} = 'someTestScript';
+    $bmwqemu::vars{GUESTINFO_CONFIG} = 'NotValid';
 
     my $chan_mock = Test::MockObject->new->set_true(qw(write send_eof close read2 eof exit_status));
     my $backend_mock = Test::MockModule->new('backend::svirt');
@@ -169,6 +185,29 @@ subtest 'starting VMware console with Cloud Init' => sub {
     $console_mock->redefine(get_ssh_credentials => sub { (hostname => 'foo', username => 'root', password => '123') });
     $tmp_mock->redefine(tempfile => sub { (undef, '/t') });
 
+    dies_ok { $svirt_console->define_and_start } 'Invalid GUESTINFO_CONFIG was passed';
+};
+
+subtest 'starting VMware console with combustion' => sub {
+    $bmwqemu::vars{VMWARE_HOST} = 'h';
+    $bmwqemu::vars{VMWARE_USERNAME} = 'u';
+    $bmwqemu::vars{VMWARE_PASSWORD} = 'p';
+    $bmwqemu::vars{GUESTINFO_COMBUSTION} = 'someTestScript';
+    $bmwqemu::vars{GUESTINFO_CONFIG} = 'combustion';
+
+    my $chan_mock = Test::MockObject->new->set_true(qw(write send_eof close read2 eof exit_status));
+    my $backend_mock = Test::MockModule->new('backend::svirt');
+    my $console_mock = Test::MockModule->new('consoles::sshVirtsh');
+    my $tmp_mock = Test::MockModule->new('File::Temp');
+    my (@cmds, @ssh_cmds);
+    $console_mock->redefine(run_cmd => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $console_mock->redefine(get_cmd_output => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $backend_mock->redefine(run_ssh => sub ($self, $cmd, %args) { push @ssh_cmds, $cmd; (undef, $chan_mock) });
+    $backend_mock->redefine(start_serial_grab => 1);
+    $console_mock->redefine(get_ssh_credentials => sub { (hostname => 'foo', username => 'root', password => '123') });
+    $tmp_mock->redefine(tempfile => sub { (undef, '/t') });
+    $console_mock->redefine(_encode_config => sub ($self, $f, $k) { 'testingCombustion' });
+
     $svirt_console->define_and_start;
     like shift @cmds, qr/cat > \/t <<.*username=u.*password=p.*auth-esx-h/s, 'config written';
     my $s = 'virsh -c esx://u@h/?no_verify=1\\&authfile=/t ';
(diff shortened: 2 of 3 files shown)


consoles/sshVirtsh.pm
```diff
@@ -496,6 +499,33 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $file) {
+    unless (!!$file) {
+        croak "Please provide a configuration file to encode!";
+    }
```
Martchus, line 505:
```suggestion
    croak 'Please provide a configuration file to encode!' unless $file;
```
(There's no need to use the `!!`-syntax here as `$file` would be converted to a boolean either way.)

consoles/sshVirtsh.pm
```diff
@@ -496,6 +499,33 @@ sub resume ($self) {
...
+
+        my $z = IO::Compress::Gzip->new('test', Level => Z_BEST_COMPRESSION);
+        $z->print(<$fh_conf>);
+        $z->close();
+        close $fh_conf;
+
+        open(my $gzip, '<', 'test');
+        local $/ = undef;
+        my $data = <$gzip>;
+        close $gzip;
```
Martchus, line 521:
```suggestion
        my $data = Mojo::File->new($gzip)->slurp;
```

consoles/sshVirtsh.pm
```diff
@@ -14,6 +14,9 @@ use File::Which;
 use Mojo::DOM;
 use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
+use Carp 'croak';
+use MIME::Base64 qw(encode_base64);
+use IO::Compress::Gzip qw(gzip :level);
```
Martchus, line 19: It would probably make sense to use `Mojo::Util:b64_encode` (see https://docs.mojolicious.org/Mojo/Util#b64_encode) and `Mojo::Util::gzip` (see https://docs.mojolicious.org/Mojo/Util#gzip). To read/write files you can use `Mojo::File`. All these functions have the advantage that error handling is less manual and the usage more streamlined. (If your code needs the detailed API these modules provide you can keep using them - but probably not.)

consoles/sshVirtsh.pm
```diff
@@ -542,17 +572,32 @@ __END"
...
-        my $ci_encoding = $bmwqemu::vars{CLOUD_INIT_ENCODING};
-
-        if ($ci_meta && $ci_user && $ci_encoding) {
-            $self->run_cmd(qq{echo 'guestinfo.metadata = "$ci_meta"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.metadata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata = "$ci_user"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
+        my $fb_tool = $bmwqemu::vars{FIRST_BOOT_CONFIG};
+
+        if (!!$fb_tool && $fb_tool ne 'wizard') {
```
Martchus, line 577:
```suggestion
        if ($fb_tool && $fb_tool ne 'wizard') {
```

consoles/sshVirtsh.pm
```diff
@@ -496,6 +499,33 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $file) {
+    unless (!!$file) {
+        croak "Please provide a configuration file to encode!";
```
Martchus, line 504: Maybe it would make sense if this function would get only the key and read `$bmwqemu::vars{$setting_key}`. Then it could tell the user what concrete variable is missing. Otherwise the error message could also just list all relevant variables.

Martchus (changes requested):
This needs more tests considering not all lines are covered.

The newly introduced variables should also be documented.

consoles/sshVirtsh.pm
mloviska: Fixed. I usually do this in order to make sure that the conversion happens. I find it more readable, but that is a matter of personal taste.

consoles/sshVirtsh.pm
mloviska: Added, this works nice, thanks!

consoles/sshVirtsh.pm
mloviska: fixed

consoles/sshVirtsh.pm
mloviska: Adapted, this worked nice too. thanks for the suggestion

consoles/sshVirtsh.pm
mloviska: That crossed my mind too. The problem over here is that in case of cloud-init we need to encode user-data and meta-data. So the function would need to return 2 strings. Also the variable as of now is not single but depends on the configuration tool that we want to use (combustion, ignition, cloud-init). I will keep this open and rethink what would be a better approach

Martchus: You can also just pass the name of the variable as extra argument.

consoles/sshVirtsh.pm
```diff
@@ -14,6 +14,10 @@ use File::Which;
 use Mojo::DOM;
 use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
+use Mojo::Util;
+use Carp 'croak';
+use Data::Dumper;
```
Martchus, line 19: This is probably a leftover from debugging:
```suggestion
```

consoles/sshVirtsh.pm
```diff
@@ -496,6 +498,22 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $config, $key) {
+    unless ($config) {
+        croak "Please provide a configuration file to encode, variable $key has to be set!";
```
kalikiana, line 503: If this isn't easily testable in the unit test you need to mark it as uncoverable
```suggestion
        croak "Please provide a configuration file to encode, variable $key has to be set!"; # uncoverable statement
```

consoles/sshVirtsh.pm
```diff
@@ -542,17 +560,32 @@ __END"
...
+            } elsif ($fb_tool eq 'cloud-init') {
+                my ($conf, $meta) = split(',', $bmwqemu::vars{FIRST_BOOT_CLOUD_INIT});
+                $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                $self->run_cmd(qq{echo 'guestinfo.metadata.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                $conf = $self->_encode_config($conf, 'FIRST_BOOT_CLOUD_INIT');
+                $self->run_cmd(qq{echo 'guestinfo.userdata = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                $meta = $self->_encode_config($meta, 'FIRST_BOOT_CLOUD_INIT');
+                $self->run_cmd(qq{echo 'guestinfo.metadata = "$meta"' >> $vmx}, domain => 'sshVMwareServer');
+            } else {
+                croak 'Unknown provisioning option has been passed through FIRST_BOOT_CONFIG test variable';
```
kalikiana, line 587: If this isn't easily testable in the unit test you need to mark it as uncoverable
```suggestion
                croak 'Unknown provisioning option has been passed through FIRST_BOOT_CONFIG test variable'; # uncoverable statement
```

consoles/sshVirtsh.pm
```diff
@@ -542,17 +560,32 @@ __END"
...
-            $self->run_cmd(qq{echo 'guestinfo.userdata = "$ci_user"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
+        my $fb_tool = $bmwqemu::vars{FIRST_BOOT_CONFIG};
+
+        if ($fb_tool && $fb_tool ne 'wizard') {
+            my $encoding = 'gzip+base64';
+            if ($fb_tool =~ /combustion|ignition/) {
+                if ($bmwqemu::vars{FIRST_BOOT_COMBUSTION}) {
+                    my $conf = $self->_encode_config($bmwqemu::vars{FIRST_BOOT_COMBUSTION}, 'FIRST_BOOT_COMBUSTION');
+                    $self->run_cmd(qq{echo 'guestinfo.combustion.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
```
Vogtinator, line 570: Doesn't exist

consoles/sshVirtsh.pm
```diff
@@ -542,17 +560,32 @@ __END"
...
-            $self->run_cmd(qq{echo 'guestinfo.metadata = "$ci_meta"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.metadata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata = "$ci_user"' >> $vmx}, domain => 'sshVMwareServer');
-            $self->run_cmd(qq{echo 'guestinfo.userdata.encoding = "$ci_encoding"' >> $vmx}, domain => 'sshVMwareServer');
+        my $fb_tool = $bmwqemu::vars{FIRST_BOOT_CONFIG};
+
+        if ($fb_tool && $fb_tool ne 'wizard') {
+            my $encoding = 'gzip+base64';
+            if ($fb_tool =~ /combustion|ignition/) {
+                if ($bmwqemu::vars{FIRST_BOOT_COMBUSTION}) {
```
Vogtinator, line 568: You might have to call this `ASSET_*` so that it gets copied

consoles/sshVirtsh.pm
mloviska: Does it mean user cannot set the encoding or that I have missed `script` after `combustion`?

consoles/sshVirtsh.pm
mloviska: Isn't a clone of http://github.com/os-autoinst/os-autoinst-distri-opensuse/ always present on the worker when a job is being spawned? The idea is basically the same as here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/journal_check.pm#L20

consoles/sshVirtsh.pm
Vogtinator: > Does it mean an user cannot set the encoding

Correct

consoles/sshVirtsh.pm
Vogtinator: Ok, so you want the config inside the test distri and not as an asset?

consoles/sshVirtsh.pm
mloviska: Thanks, removed

consoles/sshVirtsh.pm
mloviska: Yeah, it is already present in `data` and the test itself doesn't need to have the file, just the encoded content

consoles/sshVirtsh.pm
mloviska: So I have tested the code without a parameter, and the function wouldn't be even callable. Hence, I have removed this condition

consoles/sshVirtsh.pm
mloviska: Added a test for it

consoles/sshVirtsh.pm
```diff
@@ -496,6 +498,22 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $config, $key) {
+    # expand path
+    $config = sprintf('%s/data/%s', $bmwqemu::vars{CASEDIR}, $config);
+
+    unless (-e -f $config) {
+        croak "'$config' either doesn't exist or it is not a file, update the $key variable";
+    }
```
Martchus, line 507: I think `-e` is redundant. To fool the coverage analysis (so you don't have to write a test case for this `croak`) you can use a post-unless:
```suggestion
    croak "'$config' either doesn't exist or it is not a file, update the $key variable" unless -f $config;
```

consoles/sshVirtsh.pm
```diff
@@ -496,6 +498,22 @@ sub resume ($self) {
...
+
+    my $content = Mojo::File->new($config)->slurp;
+    my $gzip = Mojo::Util::gzip $content;
+    my $encoded_config = sprintf('%s', Mojo::Util::b64_encode($gzip));
```
Martchus, line 511: `sprintf` isn't doing anything here, right?
```suggestion
    my $encoded_config = Mojo::Util::b64_encode($gzip);
```

consoles/sshVirtsh.pm
```diff
@@ -542,17 +560,31 @@ __END"
...
+                    my $conf = $self->_encode_config($bmwqemu::vars{FIRST_BOOT_COMBUSTION}, 'FIRST_BOOT_COMBUSTION');
+                    $self->run_cmd(qq{echo 'guestinfo.combustion.script = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                }
+                if ($bmwqemu::vars{FIRST_BOOT_IGNITION}) {
+                    my $conf = $self->_encode_config($bmwqemu::vars{FIRST_BOOT_IGNITION}, 'FIRST_BOOT_IGNITION');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                }
+            } elsif ($fb_tool eq 'cloud-init') {
+                my ($conf, $meta) = split(',', $bmwqemu::vars{FIRST_BOOT_CLOUD_INIT});
```
Martchus, line 578: `FIRST_BOOT_CLOUD_INIT` might not be set here or not contain a `,`. The code should not produce Perl warnings in those cases (and instead probably `croak` with a clear error message).

t/22-svirt.t
```diff
@@ -169,6 +185,103 @@ subtest 'starting VMware console with Cloud Init' => sub {
...
+    my $tmp_mock = Test::MockModule->new('File::Temp');
+    my (@cmds, @ssh_cmds);
+    $console_mock->redefine(run_cmd => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $console_mock->redefine(get_cmd_output => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $backend_mock->redefine(run_ssh => sub ($self, $cmd, %args) { push @ssh_cmds, $cmd; (undef, $chan_mock) });
+    $backend_mock->redefine(start_serial_grab => 1);
+    $console_mock->redefine(get_ssh_credentials => sub { (hostname => 'foo', username => 'root', password => '123') });
+    $tmp_mock->redefine(tempfile => sub { (undef, '/t') });
+
+    my ($conf, $meta) = $console_mock->redefine(_encode_config => sub ($self, $f, $k) { 'CI_test_CONF' });
```
Martchus, line 283: Those variables are not used (and supposedly cannot be initialized that way anyway).

```suggestion
    $console_mock->redefine(_encode_config => sub ($self, $f, $k) { 'CI_test_CONF' });
```

I'd also avoid the blank line and move it to the other `$console_mock->redefine` calls.

t/22-svirt.t
```diff
@@ -169,6 +185,103 @@ subtest 'starting VMware console with Cloud Init' => sub {
...
+    my $tmp_mock = Test::MockModule->new('File::Temp');
+    my (@cmds, @ssh_cmds);
+    $console_mock->redefine(run_cmd => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $console_mock->redefine(get_cmd_output => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $backend_mock->redefine(run_ssh => sub ($self, $cmd, %args) { push @ssh_cmds, $cmd; (undef, $chan_mock) });
+    $backend_mock->redefine(start_serial_grab => 1);
+    $console_mock->redefine(get_ssh_credentials => sub { (hostname => 'foo', username => 'root', password => '123') });
+    $tmp_mock->redefine(tempfile => sub { (undef, '/t') });
+
+    $console_mock->redefine(_encode_config => sub ($self, $f, $k) { 'ignitionTEST' });
```
Martchus, line 246: I'd remove the blank line and move it to the other `$console_mock->redefine` calls.

consoles/sshVirtsh.pm
```diff
@@ -496,6 +498,22 @@ sub resume ($self) {
 
 sub get_remote_vmm ($self) { $bmwqemu::vars{VMWARE_REMOTE_VMM} // '' }
 
+sub _encode_config ($self, $config, $key) {
+    # expand path
+    $config = sprintf('%s/data/%s', $bmwqemu::vars{CASEDIR}, $config);
```
Martchus, line 503:
```suggestion
    $config = "$bmwqemu::vars{CASEDIR}/data/$config";
```

Martchus (changes requested):

consoles/sshVirtsh.pm
mloviska: fixed

consoles/sshVirtsh.pm
mloviska: yes, true! thanks

consoles/sshVirtsh.pm
mloviska: I honestly prefer `sprintf` formatting than mangling it altogether with concatenation.

t/22-svirt.t
mloviska: fixed

t/22-svirt.t
mloviska: fixed

mloviska:
> The newly introduced variables should also be documented.

This part was not addressed yet. Frankly, I would rather define the variables in `osado` as people will look for them in the context of test suite variables, hence they should end up in [variables.md](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/variables.md). Nevertheless, I do not object to document them in this repo.

consoles/sshVirtsh.pm
Martchus: Really? For me the interpolation is much more readable because one doesn't have to shuffle around the pieces together in one's head. (With `sprintf` the middle part of the code becomes the start of the string.)

Martchus:
See it from the perspective of the maintainers of this repo and contributors outside the scope of openSUSE: When reading the code we might wonder what the purpose of those variables is. So it is helpful to find the documentation here directly. Otherwise one had to search where this is actually used (and osado is just one openSUSE-specific possibility).

consoles/sshVirtsh.pm
mloviska: I admit, it is a matter of taste. I prefer it as you can clearly see what is the string literal and what is going to be replaced and in which order, and is usually more spacious than a dense long concatenation.

consoles/sshVirtsh.pm
```diff
@@ -542,17 +558,35 @@ __END"
...
+                }
+                if ($bmwqemu::vars{FIRST_BOOT_IGNITION}) {
+                    my $conf = $self->_encode_config($bmwqemu::vars{FIRST_BOOT_IGNITION}, 'FIRST_BOOT_IGNITION');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data.encoding = "$encoding"' >> $vmx}, domain => 'sshVMwareServer');
+                    $self->run_cmd(qq{echo 'guestinfo.ignition.config.data = "$conf"' >> $vmx}, domain => 'sshVMwareServer');
+                }
+            } elsif ($fb_tool eq 'cloud-init') {
+                unless ($bmwqemu::vars{FIRST_BOOT_CLOUD_INIT}) {
+                    croak "FIRST_BOOT_CLOUD_INIT is unset, or does not contain user-data and meta-data configs";
+                }
```
Martchus, line 578: I guess this should be one statement, again for fooling the coverage analysis :-)

mloviska: Hidden, it won't find it anymore :D

mloviska:
> See it from the perspective of the maintainers of this repo and contributors outside the scope of openSUSE: When reading the code we might wonder what the purpose of those variables is. So it is helpful to find the documentation here directly. Otherwise one had to search where this is actually used (and osado is just one openSUSE-specific possibility).

I have documented the variables.

Martchus (approved):

mloviska:
I have just renamed the variables to be more VMWare specific (hopefully).
Please let me know whether you are fine with the changes, and we can merge this one @kalikiana @Vogtinator

consoles/sshVirtsh.pm
kalikiana: I also find interpolation easier to read. Maybe even go a step further:

```suggestion
    my $casedir = $bmwqemu::vars{CASEDIR};
    $config = "$casedir/data/$config";
```

mloviska: Alright, added. Let's move on :)

kalikiana (approved):

PR #2531, opened 2024-08-16
Allow package builds without `Inline::Python`

Martchus:
The dependency is already just recommended by our spec file. However, the code uses it unconditionally. This change removes that. It also allows tests to run without the dependency.

This will also ease submitting packages to SLE because `Inline::Python` and some of its dependencies are not in SLE.

Related ticket: https://progress.opensuse.org/issues/128318

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -8,14 +8,15 @@
 #######################################################
 ---
 targets:
-  cpanfile: [                    cover, devel,         main,             lint,           test_base, test, test_version_only                    ]
-  spec:     [ build_base, build,        devel,         main, spellcheck,       yamllint, test_base, test, test_version_only, python_style, ocr, test_non_s390 ]
-  docker:   [ build_base, build,        devel, docker, main, spellcheck, lint, yamllint, test_base, test,                    python_style, ocr, test_non_s390 ]
+  cpanfile: [ cover, devel, main, lint, test_base, test, test_version_only, python_support ]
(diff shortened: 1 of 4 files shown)


Martchus:
Note that the changes within the spec file are identical to what we already to for `python_style_requires` (because black is missing on SLE).r

autotest.pm
```diff
@@ -100,7 +100,7 @@ sub _debug_python_version () {
     state $python_loaded = 0;
     return if $python_loaded++;
     my $code = <<~'EOM';
-    use Inline::Python;
+    require Inline::Python;
```
perlpunk, line 103: I would say this change is not necessary.
This function is only executed for a `.py` test, and the code is in a string eval.

b10n1k (approved):
I would say LGTM. But needs more eyes. Not very confident when it comes to .spec files

perlpunk (approved):

okurz (commented):
With this change we won't notice as easily if we just loose the package in all cases and actually Python support is quite important for the case is "openQA for SUSE customers". We should still try to have the python dependencies in SLE

Martchus:
I think it is generally beneficial to make this truly optional. Of course we may decide later that it's worth packaging the required dependent chain and then we can just flip the switch.

AdamWill:
In the spec file, this added a define of `python_support_requires` which is never used. It should either get added to one of the _requires , or removed, I think?

Martchus:
https://github.com/os-autoinst/os-autoinst/pull/2572

PR #2532, opened 2024-08-16
spec file: Disable OCR support on SLE and fix condition for Python support

Martchus:
See particular commit messages for details and https://progress.opensuse.org/issues/128318

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -85,10 +85,13 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %define python_style_requires %{nil}
 %endif
-%ifnarch ppc ppc64 ppc64le s390x
-%bcond_without ocr
-%else
+%ifarch ppc ppc64 ppc64le s390x
+%define ocr_not_supported_by_arch 1
+%endif
+%if 0%{?ocr_not_supported_by_arch} || !0%{?is_opensuse}
 %bcond_with ocr
+%else
+%bcond_without ocr
 %endif
 %if %{with ocr}
 # The following line is generated from dependencies.yaml


kalikiana (approved):

perlpunk:
```
[   31s] error: parse error in expression:  0%{ocr_not_supported_by_arch} || !01
[   31s] error:                              ^
[   31s] error: /home/abuild/rpmbuild/SOURCES/os-autoinst.spec:91: bad %if condition:  0%{ocr_not_supported_by_arch} || !01
```

dist/rpm/os-autoinst.spec
```diff
@@ -68,7 +68,7 @@ Source0:        %{name}-%{version}.tar.xz
 %bcond_with black
 %endif
 # SLE is missing Python support requirements
-%if !0%{?is_opensuse}
+%if 0%{?is_opensuse}
```
kalikiana, line 71: Ah. I guess you need to get rid of the ? here 🤔 
```suggestion
%if 0%{is_opensuse}
```

baierjan: Nope `0%{?is_opensuse}` is correct syntax: https://en.opensuse.org/openSUSE:Specfile_guidelines#Conditionals

dist/rpm/os-autoinst.spec
```diff
@@ -85,10 +85,13 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %define python_style_requires %{nil}
 %endif
-%ifnarch ppc ppc64 ppc64le s390x
-%bcond_without ocr
-%else
+%ifarch ppc ppc64 ppc64le s390x
+%define ocr_not_supported_by_arch 1
+%endif
+%if 0%{ocr_not_supported_by_arch} || !0%{?is_opensuse}
```
baierjan, line 91: From https://rpm-software-management.github.io/rpm/manual/macros.html

```suggestion
%ifarch ppc ppc64 ppc64le s390x
%define ocr_not_supported_by_arch 1
%endif
%if 0%{?ocr_not_supported_by_arch} || !0%{?is_opensuse}
```

Or, looking at https://github.com/rpm-software-management/rpm/blob/master/docs/manual/conditionalbuilds.md#creating-build-conditionals you might probably use something like 

```suggestion
%ifarch ppc ppc64 ppc64le s390x
%bcond ocr_not_supported_by_arch 1
%endif
%if %{with ocr_not_supported_by_arch} || !0%{?is_opensuse}
```

_Note: this is syntax suggestion only, I did not try to understand the full logic workflow to optimize the conditionals._

PR #2543, opened 2024-09-09
Fix pod for backend_get_wait_still_screen_on_here_doc_input

perlpunk:


testapi.pm
```diff
@@ -2267,10 +2267,13 @@ https://progress.opensuse.org/issues/60566 which is that typing too fast into
...
 starting to write the script into the here document.
+
 =cut
+
 sub backend_get_wait_still_screen_on_here_doc_input () {
     state $ret;
     $ret = query_isotovideo('backend_get_wait_still_screen_on_here_doc_input', {}) unless defined($ret);
     return get_var(_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT => $ret);
 }
+
```
kalikiana, line 2278: So is this actually required? Or a change in a dependency? I'm wondering why the docs target didn't look broken when I tried to reproduce it. I didn't see any warnings or errors either.

perlpunk: No, this line is not required.

But an empty line after `=cut` is required for some pod renderers and always has been. And it's a good practice.

Martchus (approved):

PR #2544, opened 2024-09-10
Remove quotes from URL in the log strings

b10n1k:
The problem which the PR tries to address is the extra characters in the linking of the URL variables in the logs. This appears to occur because of a condition in Anser.linkify[0].
Beyond the whitespace condition, there appears another issue when a string has one URL next to another.

[0] https://github.com/IonicaBizau/anser/tree/master?tab=readme-ov-file#anserlinkifytxt


OpenQA/Isotovideo/Utils.pm
```diff
@@ -101,7 +101,7 @@ sub _open_cache_index ($root_cache_dir, $index_file) {
 
 sub _determine_size ($dir, $handle_output) {
     $handle_output->($?, my $du = qx{du -s "$dir"});
-    die "Unable to determine size of Git directory under '$dir': du returned '$du'\n" unless $du =~ /(\d+).*/;
+    die "Unable to determine size of Git directory under $dir : du returned '$du'\n" unless $du =~ /(\d+).*/;
```
Martchus, line 104: The directory shouldn't be problematic so I guess we shouldn't change this line. The same counts for the other occurrences of `$dir`.

OpenQA/Isotovideo/Utils.pm
```diff
@@ -307,7 +307,7 @@ sub checkout_git_refspec ($dir, $refspec_variable) {
     }
     my $hash = calculate_git_hash($dir);
     my $url = git_remote_url($dir);
-    bmwqemu::diag "git url in '$dir': $url";
+    bmwqemu::diag "git url in '$dir': $url ";
```
Martchus, line 310:
```suggestion
    bmwqemu::diag "git url in '$dir': $url";
```

Martchus (changes requested):
Maybe we can fix Anser (or workaround the bug in Answer) instead of trying to be very conservative in what kind of URLs we log.

OpenQA/Isotovideo/Utils.pm
b10n1k: it needs a literal whitespace. with out it adds `<span>`

OpenQA/Isotovideo/Utils.pm
b10n1k: As I see there are cases where `$dir` is actual a URL. I dont know if the name can change tho

b10n1k:
> Maybe we can fix Anser (or workaround the bug in Answer) instead of trying to be very conservative in what kind of URLs we log.

yea upstream fix is the first thought but I guess it is mentioned in the docs for a reason and it is a small thing to do here. it is about how we phrase the strings.

perlpunk:
The `linkify` code is really simple and just looks for non-space characters:
```
txt.replace(/(https?:\/\/[^\s]+)/gm, str => `<a href="${str}">${str}</a>`);
```
That's why the single quote and also `</span>` is included.
The author welcomes a PR to fix that:
https://github.com/IonicaBizau/anser/issues/63

Of course, matching a URL really correct can result in a big regex:
https://metacpan.org/dist/Regexp-Common/source/lib/Regexp/Common/URI/http.pm
https://metacpan.org/release/ABIGAIL/Regexp-Common-2024080801/source/lib/Regexp/Common/URI/RFC2396.pm

I would prefer to fix Anser. A quick workaround would be to write our own slightly more sophisticated `linkify` which excludes things like quotes and angle brackets.

OpenQA/Isotovideo/Utils.pm
perlpunk: Do you have an example where `$dir` is a URL? The variable name sounds wrong in that case.

Martchus: I'm pretty sure this function expects a normal path. If `$dir` was a URL this function would not work correctly and should not be called in the first place. (I don't think this is the case, though. Probably you can really just omit this change.)

okurz:
and what about https://linkify.js.org/ as alternative

Martchus:
Good idea but will have to see how well it ties in with anser which we still need for dealing with escape sequence.

Martchus:
By the way, is this related to a progress ticket?

perlpunk:
I created https://github.com/os-autoinst/openQA/pull/5920 to fix linkify.

perlpunk (commented):
Can you change the PR to change all single quotes around URLs to double quotes? Then we should be fine after https://github.com/os-autoinst/openQA/pull/5920

perlpunk:
btw, IMHO quoting variables in logging statements is often a good thing because it can reveal wrong input or trailing spaces more easily.

b10n1k:
https://progress.opensuse.org/issues/166676

b10n1k:
> https://progress.opensuse.org/issues/166676

@Martchus ^^. I should have put it in your comment above

b10n1k:
> Can you change the PR to change all single quotes around URLs to double quotes? Then we should be fine after [os-autoinst/openQA#5920](https://github.com/os-autoinst/openQA/pull/5920)

Done on c7edd9d3

Martchus (approved):

OpenQA/Isotovideo/Utils.pm
```diff
@@ -195,11 +195,11 @@ sub clone_git ($local_path, $clone_url, $clone_depth, $branch, $dir, $dir_variab
         $clone_depth *= 2;
         @out = qx[git -C $local_path fetch --progress --depth=$clone_depth 2>&1];
         $handle_output->($?, @out);
-        die "Could not find '$branch' in complete history in cloned Git repository '$dir'" if grep /remote: Total 0/, @out;
+        die "Could not find '$branch' in complete history in cloned Git repository \"$dir\"" if grep /remote: Total 0/, @out;
     }
     @out = qx{git -C $local_path checkout $branch};
     bmwqemu::diag "@out" if @out;
-    die "Unable to checkout branch '$branch' in cloned Git repository '$dir'" unless $? == 0;
+    die "Unable to checkout branch '$branch' in cloned Git repository $dir" unless $? == 0;
```
baierjan, line 202:
```suggestion
    die "Unable to checkout branch '$branch' in cloned Git repository \"$dir\"" unless $? == 0;
```

or is there a reason for exception?

perlpunk (commented):
Just a note: I personally prefer `qq/ ... "$url" ... /` to avoid having to backslash.

The commit message I don't find very easy to understand:
```
The problem which the PR tries to address is the extra characters in the
linking of the URL variables in the logs. This appears to occur because of a
condition in Anser.linkify[0].
```
Think about reading this message when you go through the git log as a new team member in one year from now.
I think just stating that unescaped single quotes are valid parts of URIs is enough explanation.

```
Beyond the whitespace condition, there appears another issue when a string has
one URL next to another.
```
Where do we have URLs next to each other and why is that a specific problem?
```
Solution provided on os-autoinst#2544 and
this is a adaptable change to that commit
```
You are referring to the very same PR here?

OpenQA/Isotovideo/Utils.pm
b10n1k: No exception. I was running jobs to test @perlpunk PR. I will fix

b10n1k: Done https://github.com/os-autoinst/os-autoinst/commit/ea085328ccc8838cf86cc8062810ac1da49a1676

b10n1k:
> > The commit message I don't find very easy to understand:
> 
> I had the same thought. I haven't said anything because I think it is an improvement over previous commit messages (from previous PRs) but yes, it is basically way too wordy and also weirdly worded¹.
> 
> ¹ @b10n1k I'm not a native English speaker as well. However, I think the word "condition" is not really not used that way in English. Rather say "because of a limitation in …" or "because of the the whitespace problem …" in such cases.

Removed

perlpunk (commented):
line 79 was missed:
```
sub _fetch_new_refs ($clone_url, $cache_dir, $branch_arg, $handle_output) {
    bmwqemu::fctinfo "Updating Git cache for '$clone_url' under '$cache_dir'";
```

PR #2545, opened 2024-09-12
Exclude sub packages with many additional dependencies on SLE

Martchus:
As stated in https://build.suse.de/request/show/343857#comment-4023273 not all sub packages are installable on SLE. This only concerns the `devel` and `deps` sub packages which are really optional. So it makes sense to avoid submitting a huge number of Haskell some utility packages to SLE and exclude those packages instead.

Related ticket: https://progress.opensuse.org/issues/165683

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -31,6 +31,14 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %define opencv_require pkgconfig(opencv)
 %endif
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
I would like to resolve the open points in https://github.com/os-autoinst/openQA/pull/5923 first

Martchus:
Maybe we can keep the `-deps` packages, see https://progress.opensuse.org/issues/165683#note-29.

okurz (approved):
Either way the change is fine

kalikiana (approved):

Issue #2549, opened 2024-09-25
isotovideo frequently crashing in the signal handler stuff (especially on aarch64)


AdamWill:
Filing a new ticket for this as discussing it in https://github.com/os-autoinst/os-autoinst/pull/1640 was getting confusing.

On the Fedora deployments currently, I'm seeing frequent crashes in isotovideo, in the signal handler stuff that is trying to prevent crashes caused by the perl signal handler trying to handle signals from opencv threads. This is happening especially often on aarch64, but it's also happening on x86_64. It also seems to be causing workers to get into a broken state, as the qemu process from the crashed job hangs around and blocks things. I'm having to kill tons of qemu processes and restart worker processes, or just restart the affected systems periodically, or else I lose all the workers.

The backtraces I get on aarch64 and x86_64 are slightly different. [x86_64 backtrace](https://github.com/user-attachments/files/17136994/x64bt.txt), [aarch64 backtrace](https://github.com/user-attachments/files/17137256/a64bt.txt) . aarch64 shows the signal as 15 and crashes at `(PL_signals & PERL_SIGNALS_UNSAFE_FLAG))`, x86_64 shows the signal as optimized out and crashes at `sig == SIGFPE ||`.

@Vogtinator asked in the earlier discussion if I could "try to get its signal masks", but I honestly have no idea how to do that. I'm a Python coder, damnit. :D If anyone wants to walk me through it, I can try.

Martchus:
> I'm having to kill tons of qemu processes and restart worker processes, or just restart the affected systems periodically, or else I lose all the workers.

Normally the openQA worker is supposed to SIGKILL the whole process group after a timeout. That this is not working is of course yet another issue (which we saw recently as well but so far only on one of our workers).

---

This indeed sounds like the signal mask is not set as it should be on all OpenCV threads. Since OpenCV is spawning the threads on its own and by the help of yet another abstraction library (e.g. TBB) so this isn't straight forward. The current approach is to set the signal mask of the main thread, execute some code that will hopefully fully spawn OpenCV's thread pool and then set back the signal mask of the main process. ~~Maybe our code was not able to fully spawn OpenCV's thread pool so some processes did not inherit the temporarily changed signal mask from the main process.~~ Considering `sig == SIGPE` the problem is more likely that we'll have to take care of this signal as well (which we currently don't do).

Vogtinator:
[...]

PR #2551, opened 2024-09-30
Extend error handling and tests of signal blocker

Martchus:
This may help to debug problems like https://github.com/os-autoinst/os-autoinst/issues/2549.

Not sure whether it makes sense to merge (all of) this. It might be helpful to hotpatch an affected worker before we merge this.

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -333,7 +333,7 @@ private:
  */
 int opencv_default_thread_count()
 {
-    return std::max(1, std::min(cv::getNumThreads(), cv::getNumberOfCPUs() - 1));
+    return std::max(1, std::min(cv::getNumThreads(), cv::getNumberOfCPUs()));
 }
 
 /*!
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -3,12 +3,16 @@
 
 #include <cerrno>
 #include <condition_variable>
+#include <csignal>
 #include <functional>
 #include <mutex>
 #include <exception>
 #include <iostream>
+#include <cerrno>
+#include <cstdlib>
 #include <cstdint>
 #include <cstdio>
+#include <signal.h>
 #include <sys/time.h>
 #include <byteswap.h>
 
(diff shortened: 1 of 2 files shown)


Martchus:
I removed the

```
my @threads = split('\n', `grep ^SigBlk /proc/$$/task/*/status`);
my $expected_mask = sprintf('0%x', (1 << SIGCHLD - 1) | (1 << SIGTERM - 1));
like shift @threads, qr/0000000000000000/, 'signal mask of first thread is cleared';
like $_, qr/$expected_mask/, 'signal mask of other threads blocks SIGCHLD/SIGTERM' for @threads;
```

part because it is maybe a bit too specific for a unit test.

Not sure how useful this whole PR still is. It would definitely help to uncover cases in production where the signal blocker isn't working at all. However, following https://github.com/os-autoinst/os-autoinst/issues/2549 it seems that only a small number of threads are affected and we need to find out where they come from.

ppmclibs/tinycv_impl.cc
```diff
@@ -364,11 +368,34 @@ void create_opencv_threads(int thread_count)
...
+        // exit with an error if the signal mask does not contain at least SIGCHLD and SIGTERM
+        if (sigmask_get != 0) {
+            std::cerr << "Unable to get sigprocmask of OpenCV thread: " << std::strerror(errno) << '\n';
+            std::exit(EXIT_FAILURE);
+        }
+        const auto is_sigchld_blocked = sigismember(&current_set, SIGCHLD);
+        const auto is_sigterm_blocked = sigismember(&current_set, SIGTERM);
+        if (is_sigchld_blocked != 1 || is_sigterm_blocked != 1) {
+            if (is_sigchld_blocked == -1 || is_sigterm_blocked == -1) {
+                auto lock = std::unique_lock<std::mutex>(m);
```
baierjan, line 392: I guess this it the lock declared on https://github.com/os-autoinst/os-autoinst/pull/2551/files#diff-56ffbf3709dd0405661f7f4e7e2fafcf718a5f766d54f92c31f59c40590048c8L366?

Can you please explain quickly to me this assignment? It is not locking/unlocking the mutex here, or is it?

baierjan (changes requested):

ppmclibs/tinycv_impl.cc
Martchus: > It is not locking/unlocking the mutex here, or is it?

It *is* locking/unlocking the mutex here using c'tor/d'tor magic. It does that so the log messages are not mixed up.

baierjan: Ah, right. That makes sense. It wasn't clear from the documentation on the first read (as it states the lock is optional, but I see it now). Thanks!

baierjan (approved):

Martchus:
By the way, I added the "not-ready" label because the discussion on https://github.com/os-autoinst/os-autoinst/issues/2549 is still ongoing and removing the `-1` might be questionable.

PR #2552, opened 2024-10-10
Add xen-pv USBPV hub

mloviska:
Add USBPV functionality when starting a xen-pv instance.

- ticket: https://progress.opensuse.org/issues/138410


t/22-svirt.t
```diff
@@ -499,6 +499,14 @@ sub svirt_xml_validate ($svirt, %args) {
...
         is($driver_node->getAttribute('cache'), $args{driver}->{cache}, 'cache attribute of <driver> is correct');
     }
+
+    if($args{usb}) {
+        #<controller type="usb" index="0" ports="8"/>
+        my $target = $doc->findnodes('domain/devices/controller')->shift();
+        my $driver_node = $target->parentNode->findnodes('controller')->shift;
+        is($driver_node->getAttribute('ports'), 8, 'no. of ports <controller> is correct');
+        is($driver_node->getAttribute('type'), 'usb', 'type attribute of <controller> is correct');
+    }
```
Martchus:
```suggestion
    subtest '<controller type="usb" index="0" ports="8"/>' => sub {
        my $target = $doc->findnodes('domain/devices/controller')->shift();
        my $driver_node = $target->parentNode->findnodes('controller')->shift;
        is($driver_node->getAttribute('ports'), 8, 'no. of ports <controller> is correct');
        is($driver_node->getAttribute('type'), 'usb', 'type attribute of <controller> is correct');
    } if $args{usb};
```

Martchus (commented):
I can't tell whether this actually works but the code itself looks good. I only have one style suggestion and it also looks like you need to run tidy.

mloviska:
> I can't tell whether this actually works but the code itself looks good. I only have one style suggestion and it also looks like you need to run tidy.

[/dev/bus/usb](http://kepler.suse.cz/tests/24054#step/podman_privileged_mode/28) exists in xen-pv and is accessible by a privilege container. Obscure way how to prove that it works :D

xml snip
```
      </console>
      <controller type="usb" index="0" ports="8"/>
      <graphics type="vnc" port="5964" autoport="no" listen="0.0.0.0" sharePolicy="force-shared" passwd="nots3cr3t">
```

t/22-svirt.t
mloviska: Thanks Marius, you have inspired me to move this code as a subtest for xen-pv creation.

mimi1vx (approved):

consoles/sshVirtsh.pm
```diff
@@ -229,6 +229,17 @@ sub add_pty ($self, $args) {
     return;
 }
 
+sub add_usb_hub ($self) {
+    my $doc = $self->{domainxml};
+    my $devices = $self->{devices_element};
+
+    my $controller = $doc->createElement('controller');
+    $controller->setAttribute(type => 'usb');
+    $controller->setAttribute(index => '0');
```
grisu48, line 238: What happens when this method is called multiple times? Would it work? If not, perhaps it would be a consideration to make it safe to call multiple times by storing this index as variable and incrementing it each time.

grisu48 (commented):
One minor suggestion

Martchus (approved):

consoles/sshVirtsh.pm
mloviska: xen PV VM start to fail when it is called twice. I would leave it as this since we dont need multiple usb hubs and the parameters can be added later

```
      </disk>
      <controller type='usb' index='0' ports='8'/>
      <controller type='usb' index='0' ports='8'/>
      <controller type='xenbus' index='0'/>
      <interface type='network'>
```

okurz (approved):

PR #2553, opened 2024-10-21
Cover upload_logs in testapi.pm

asdil12:
Ticket: https://progress.opensuse.org/issues/130420

PR #2554, opened 2024-10-22
testapi: Slightly simplify upload_logs

okurz:
* #2553

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -10,6 +10,7 @@ use Test::Mock::Time;
 use Test::MockObject;
 use File::Temp;
 use Mojo::File qw(path);
+use Mojo::Util qw(b64_encode);
 use Test::Output qw(stderr_like stderr_unlike);
 use Test::Fatal;
 use Test::Warnings qw(:all :report_warnings);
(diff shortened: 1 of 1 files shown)


Martchus (approved):

testapi.pm
```diff
@@ -2189,12 +2189,8 @@ sub upload_asset ($file, $public = undef, $nocheck = undef) {
...
     my $basename = basename($file);
     $cmd .= autoinst_url("/upload_asset/$basename");
-    if ($nocheck) {
-        type_string("$cmd\n");
-    }
-    else {
-        return assert_script_run($cmd);
-    }
+    return assert_script_run($cmd) unless $nocheck;
+    type_string("$cmd\n");
```
kalikiana, line 2193: Please get #2553 done first as that will add the test coverage for this

kalikiana (changes requested):

testapi.pm
okurz: The idea was that with this change adding test coverage is easier. That's what we discussed today in the daily

perlpunk (commented):
Subject talks about `upload_logs`, but change is in `upload_asset`

testapi.pm
kalikiana: Why not just add it to the other branch? Making more untested changes seems counter-intuitive to me

okurz: Well, either we merge here first and @asdil12 rebases or they pick my commit into their branch

okurz:
@mergifyio rebase

PR #2556, opened 2024-10-28
Add dependencies of `tools/check-shellcheck` explicitly, fix CI checks

Martchus:
The author tests are currently failing with:

```
 4: /__w/os-autoinst/os-autoinst/tools/check-shellcheck: line 3: file: command not found
4: No files specified.
4:
4: Usage: shellcheck [OPTIONS...] FILES...
```

This is because `file` is no longer implicitly installed in the development container. This change adds dependencies of the `check-shellcheck` script explicitly to avoid running into errors like this.

### DIFF
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -29,6 +29,7 @@ RUN zypper in -y -C \
        aspell-spell \
        cmake \
        cpio \
+       file \
        gcc-c++ \
        git-core \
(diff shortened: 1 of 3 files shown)


Martchus:
This will not be directly effective. We'll have to merge it despite the checks not passing and I'll have to re-trigger services of the container build on OBS.

kalikiana (approved):

perlpunk (approved):

PR #2557, opened 2024-10-28
Cover `testapi::_handle_found_needle` in unit tests

Martchus:
Related ticket: https://progress.opensuse.org/issues/130420

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -10,6 +10,7 @@ use Test::Mock::Time;
 use Test::MockObject;
 use File::Temp;
 use Mojo::File qw(path);
+use Mojo::Util qw(b64_encode);
 use Test::Output qw(stderr_like stderr_unlike);
 use Test::Fatal;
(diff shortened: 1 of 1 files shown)


t/03-testapi.t
```diff
@@ -421,7 +440,7 @@ subtest 'check_assert_screen' => sub {
         is_deeply($autotest::current_test->{details}, [
                 {
                     result => 'unk',
-                    screenshot => 'basetest-13.png',
+                    screenshot => 'basetest-14.png',
```
kalikiana, line 443: How is this different now? I don't see any changes in assets or jobs 🤔

Martchus: It is different because the newly added test inserts a test detail so subsequent test details will have a higher number.

kalikiana: Ah. I totally missed that it writes the result to disk, too. Makes sense!

kalikiana (approved):

perlpunk (approved):

PR #2562, opened 2024-11-04
Require ffmpeg during tests

baierjan:
Reference: https://progress.opensuse.org/issues/166664

### DIFF
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -29,6 +29,7 @@ RUN zypper in -y -C \
        aspell-spell \
        cmake \
        cpio \
+       ffmpeg \
        file \
        gcc-c++ \
        git-core \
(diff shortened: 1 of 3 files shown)


okurz (changes requested):
I don't understand the reason. Also I would be careful with such approach because skipping tests can lead to unnoticed test degradation.

baierjan:
> I don't understand the reason. Also I would be careful with such approach because skipping tests can lead to unnoticed test degradation.

`backend::baseclass::_ffmpeg_banner` is a direct call `ffmpeg -version` which only make sense if `ffmpeg` is installed

baierjan:
I guess as an alternative we could force installation of ffmpeg in our testing environment (it is not there as it is only a recommended dependency). Not sure what else we can do here, what do you think?

okurz:
> I guess as an alternative we could force installation of ffmpeg in our testing environment (it is not there as it is only a recommended dependency). Not sure what else we can do here, what do you think?

Yes, forcing the installation for test environments makes sense to me

baierjan:
So I guess that could be as simple as adding to the test requirements?

okurz (approved):

Martchus (approved):

Issue #2560, opened 2024-11-01
Feature request: [accessibility] A new command in the audio section: wait_for_silence


rclobus:
For Debian we have a speech variant of the installer. It will run a screen reader during the installation steps.

It is possible to capture the audio, but it will take a lot longer to speak all text than to display it. Therefore using 'wait_still_screen' is good enough to verify that all text has been spoken. A new command like 'wait_for_silence' with 2 arguments: timeout and silenttime would help in capturing and diagnosing a11y issues.

See https://openqa.debian.net/tests/319072#step/bootwalk_3/5 for a basic test that starts the audio capture, but currently does not do any comparisons.

ssssam:
This would be useful for the GNOME OS openQA tests as well.

b10n1k:
out of curiosity, have you tried `assert_recorded_sound` or `check_recorded_sound` in any way?

rclobus:
> out of curiosity, have you tried `assert_recorded_sound` or `check_recorded_sound` in any way?

I have turned the recordings on in our Debian production server. At the moment it is not really checking the content, because the recording is way too short.

See some recordings:
* amd64, live image: https://openqa.debian.net/tests/323944 -> bootwalk_3 and bootwalk_4:4:N

okurz:
We discussed this (with you) during the SUSE QE Tools workshop session 2024-11-15 and came up with potential alternative approaches, e.g. just explicitly create needle match areas covering the white part of the recording visualizations which should be a good check for actual silence. This can be done for both cases when silence is part of expected behaviour as well as when recordings end prematurely or there is no audio output at all as part of  erroneous behaviour. In that case a "soft-fail" needle can be created (just a click to the according checkbox).

rclobus:
@okurz I've done some more experiments.

This is what I found out so far:
* snd2png does not use all available pixels -> https://github.com/os-autoinst/os-autoinst/pull/2571
* snd2png only shows the first bit of the audio recording (exactly 10.24s, corresponding to 1024 chunks of 10ms each)
* snd2png automatically strips the silence at the beginning and the end, so I cannot create a match region for the silence at the end (for recordings longer than 10.24s)
* The output from espeakup in the installer differs from run to run, the pause between utterances differs, which makes a match hard (this issue is out-of-scope for openQA)
[...]

PR #2571, opened 2024-11-18
snd2png: Clamping fix

rclobus:
When there is silence at the beginning of the recording, the image would not use all 1024 pixels for the FFT graph of the audio recording, which results in white columns at the end

### DIFF
--- a/snd2png/snd2png.cpp
+++ b/snd2png/snd2png.cpp
@@ -176,8 +176,8 @@ int main(int argc, char* argv[])
     int freqs = height / scale_factor;
     Mat grayscaleMat(height, 1024, CV_8U, Scalar(255));
 
-    if (last_non_silence - first_non_silence >= grayscaleMat.cols)
-        last_non_silence = grayscaleMat.cols - first_non_silence - 1;
+    if (last_non_silence - first_non_silence > grayscaleMat.cols)
+        last_non_silence = grayscaleMat.cols + first_non_silence;
 
     fprintf(stderr, "silences: %d %d\n", first_non_silence, last_non_silence);
     for (int i = 1; i < freqs; ++i) {


rclobus:
An example is here: https://openqa.debian.net/tests/325312#step/bootwalk_4/8

rclobus:
A side question (which might result in another change request): Since the rendering is truncated after 1024 samples, would a scaling in the X-direction make sense? The recording mentioned here is about 5 minutes, and only the first bit is shown in the image.

okurz:
> A side question (which might result in another change request): Since the rendering is truncated after 1024 samples, would a scaling in the X-direction make sense? The recording mentioned here is about 5 minutes, and only the first bit is shown in the image.

Scaling would likely mean that needles would often not match anymore. And audio clips should still be uniquely identifiable from 1k samples in general

rclobus:
> > A side question (which might result in another change request): Since the rendering is truncated after 1024 samples, would a scaling in the X-direction make sense? The recording mentioned here is about 5 minutes, and only the first bit is shown in the image.
> 
> Scaling would likely mean that needles would often not match anymore. And audio clips should still be uniquely identifiable from 1k samples in general

Agreed, scaling would make matching a lot harder. Currently the audio recordings up to 10s240ms can be matched with the needles. If scaling were to be implemented (I could create a MR for that), it should only be applied to longer recordings.

Let's not keep this bugfix from being merged while discussing this side-track. The feature request at https://github.com/os-autoinst/os-autoinst/issues/2560 could be used to further discuss the audio matching strategies.

snd2png/snd2png.cpp
```diff
@@ -176,8 +176,8 @@ int main(int argc, char* argv[])
     int freqs = height / scale_factor;
     Mat grayscaleMat(height, 1024, CV_8U, Scalar(255));
 
-    if (last_non_silence - first_non_silence >= grayscaleMat.cols)
-        last_non_silence = grayscaleMat.cols - first_non_silence - 1;
+    if (last_non_silence - first_non_silence > grayscaleMat.cols)
+        last_non_silence = grayscaleMat.cols + first_non_silence;
```
okurz, line 180: hm, the original code was introduced by @coolo in 4e08b45e in 2015. There are no tests so I guess the best way to verify is with existing openQA tests. For this I will trigger a test based on openSUSE tests using audio comparison.

okurz (changes requested):
I will run a test for comparison. Please wrap your git commit message details at 80 characters per line.

snd2png/snd2png.cpp
rclobus: You could use this audio file (a live example from Debian):
https://openqa.debian.net/tests/325775/file/bootwalk_2:6:0-captured.wav

Then call 'snd2png my.wav my.png'

Alternatively, to enforce a segfault, you could use the 'aplay-captured.wav' file, but use e.g. audacity to add at least 1025 samples of silence at the beginning.

rclobus: git commit is now rewrapped.

okurz (approved):
Verified with https://openqa.opensuse.org/tests/4655815#step/firefox_audio/8

rclobus:
> Verified with https://openqa.opensuse.org/tests/4655815#step/firefox_audio/8

Note that this file is shorter than 10.24s and without silence at the beginning. To verify this change, you'll need silence at the beginning of the recording and at length of least 10.24s in total.

Try this recording 15s silence, 15s DMTF tones, 15s silence:
* https://www.rclobus.nl/snd2png.wav
* https://www.rclobus.nl/snd2png.png
```
snd2png snd2png.wav snd2png.png
snd2png: 1 channels, samplerate 44100 Hz, 1965953 frames (44.58 seconds)
snd2png: 881 frequency bins
spectogram samples: 4456
max amplitude: 90.210837
silences: 1457 2481
```

Martchus:
I guess the idea behind the verification run @okurz did was to see whether the change breaks our existing test case. It doesn't break it which is good. Presumably this will require re-needling in the actually relevant cases, though. So I'm a bit hesitant to approve this presumably non-backwards compatible change.

okurz:
> I guess the idea behind the verification run @okurz did was to see whether the change breaks our existing test case. It doesn't break it which is good. Presumably this will require re-needling in the actually relevant cases, though. So I'm a bit hesitant to approve this presumably non-backwards compatible change.

Which cases would reneedling if already our Firefox-Audio works without needing a change? I'd say we can accept it

Martchus:
Cases that fulfill the criteria mentioned in https://github.com/os-autoinst/os-autoinst/pull/2571#issuecomment-2489361984.

Martchus (approved):

PR #2577, opened 2024-12-02
basetest: Simplify record_screenfail to increase coverage

okurz:
Related progress issue: https://progress.opensuse.org/issues/167950

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -225,19 +225,13 @@ sub record_screenfail ($self, %args) {
     my $status = $args{result} || 'fail';
     my $overall = $args{overall};    # whether and how to set global test result
     my $frame = $args{frame};
-
-    my $candidates;
-    for my $cand (@{$needles || []}) {
-        push @$candidates, $self->_serialize_match($cand);
-    }
-
     my $result = {
         screenshot => $self->next_resultname('png'),
         result => $status,
         frametime => _framenumber_to_timerange($frame),
     };
-
-    $result->{needles} = $candidates if $candidates;
+    my @candidates = map { $self->_serialize_match($_) } @{$needles || []};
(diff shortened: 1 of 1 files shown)


basetest.pm
```diff
@@ -225,18 +225,12 @@ sub record_screenfail ($self, %args) {
...
-        push @$candidates, $self->_serialize_match($cand);
-    }
-
     my $result = {
         screenshot => $self->next_resultname('png'),
         result => $status,
         frametime => _framenumber_to_timerange($frame),
     };
-
+    my $candidates = map { $self->_serialize_match($_) } @{$needles || []};
```
perlpunk, line 233:
```suggestion
    my $candidates = [map { $self->_serialize_match($_) } @{$needles || []}];
```

basetest.pm
```diff
@@ -225,18 +225,12 @@ sub record_screenfail ($self, %args) {
...
-        push @$candidates, $self->_serialize_match($cand);
-    }
-
     my $result = {
         screenshot => $self->next_resultname('png'),
         result => $status,
         frametime => _framenumber_to_timerange($frame),
     };
-
+    my $candidates = [map { $self->_serialize_match($_) } @{$needles || []}];
```
perlpunk, line 233: See my amendment to my previous comment.
that would always make `$candidates` true.
Maybe use an array instead.

```suggestion
    my @candidates = map { $self->_serialize_match($_) } @{$needles || []};
```
and then later `if @candidates`

perlpunk (approved):

basetest.pm
okurz: sounds good. I thought of that as well. Unfortunately I couldn't find a good unit test covering the functionality properly. Let's see if all CI tests are happy.

r-richardson (approved):

PR #2578, opened 2024-12-02
Add test coverage to myjsonrpc

Wabri:
Implementing missing tests for myjsonrpc:

- [myjsonrpc:43-44](https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/myjsonrpc.pm#L43) => done
- [myjsonrpc:104-105](https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/myjsonrpc.pm#L104) => set as uncover
- [myjsonrpc:111-112](https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/myjsonrpc.pm#L111) => done

Issue: https://progress.opensuse.org/issues/167935

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -14,9 +14,20 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+# hash for keeping state
+our $sockets;
+
 sub _syswrite($to_fd, $json) { syswrite($to_fd, $json) }
 
 sub is_debug () { DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC} }
+
+sub handle_read_error ($fd) {
+    # throw an error except can_read has been interrupted
+    my $error = $!;
+    confess "ERROR: unable to wait for JSON reply: $error\n" unless $!{EINTR};
+    # try again if can_read's underlying system call has been interrupted as suggested by the perlipc documentation
+    bmwqemu::diag("read_json($fd): can_read's underlying system call has been interrupted, trying again\n") if is_debug;    # uncoverable statement
+}
 
 sub send_json ($to_fd, $cmd) {
     # allow regular expressions to be automatically converted into
@@ -43,9 +54,6 @@ sub send_json ($to_fd, $cmd) {
     }
     return $cmdcopy{json_cmd_token};
 }
-
-# hash for keeping state
-our $sockets;
 
 # utility function
 sub read_json ($socket, $cmd_token = undef, $multi = undef) {
@@ -91,16 +99,10 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
 
         # wait for next read
 
-        until (my @res = $s->can_read) {
-            # throw an error except can_read has been interrupted
-            my $error = $!;
-            confess "ERROR: unable to wait for JSON reply: $error\n" unless $!{EINTR};
-            # try again if can_read's underlying system call has been interrupted as suggested by the perlipc documentation
-            bmwqemu::diag("read_json($fd): can_read's underlying system call has been interrupted, trying again\n") if is_debug();    # uncoverable statement
-        }
+        handle_read_error($fd) until (my @res = $s->can_read);
 
         my $qbuffer;
-        if (!sysread($socket, $qbuffer, READ_BUFFER)) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(), return }
+        if (!sysread($socket, $qbuffer, READ_BUFFER)) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(); return }
         $cjx->incr_parse($qbuffer);
     }
 
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -96,7 +96,7 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
             my $error = $!;
             confess "ERROR: unable to wait for JSON reply: $error\n" unless $!{EINTR};
             # try again if can_read's underlying system call has been interrupted as suggested by the perlipc documentation
-            bmwqemu::diag("read_json($fd): can_read's underlying system call has been interrupted, trying again\n") if is_debug();
+            bmwqemu::diag("read_json($fd): can_read's underlying system call has been interrupted, trying again\n") if is_debug();    # uncoverable statement
         }
 
         my $qbuffer;
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -14,6 +14,10 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+sub _syswrite($to_fd, $json) {
+    return syswrite($to_fd, "$json");
+}
+
 sub send_json ($to_fd, $cmd) {
     # allow regular expressions to be automatically converted into
     # strings, using the Regex::TO_JSON function as defined at the end
@@ -37,9 +41,9 @@ sub send_json ($to_fd, $cmd) {
     }
     $json .= "\n";
 
-    confess 'myjsonprc: called on undefined file descriptor' unless defined $to_fd;
-    my $wb = syswrite($to_fd, "$json");
-    if (!$wb || $wb != length($json)) {
+    confess 'myjsonrpc: called on undefined file descriptor' unless defined $to_fd;
+    my $written_bytes = _syswrite($to_fd, "$json");
+    if (!$written_bytes || $written_bytes != length($json)) {
         die('myjsonrpc: remote end terminated connection, stopping') if !DEBUG_JSON && $! =~ qr/Broken pipe/;
         confess "syswrite failed: $!";
     }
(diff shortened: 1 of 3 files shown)


t/24-myjsonrpc.t
```diff
@@ -3,6 +3,9 @@
 # Copyright 2019-2020 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+BEGIN {
+    $ENV{PERL_MYJSONRPC_DEBUG} = 0;
+}
```
perlpunk, line 8: We do have `t/24-myjsonrpc-debug.t` for that already.

myjsonrpc.pm
```diff
@@ -38,10 +38,10 @@ sub send_json ($to_fd, $cmd) {
     $json .= "\n";
 
     confess 'myjsonprc: called on undefined file descriptor' unless defined $to_fd;
-    my $wb = syswrite($to_fd, "$json");
-    if (!$wb || $wb != length($json)) {
-        die('myjsonrpc: remote end terminated connection, stopping') if !DEBUG_JSON && $! =~ qr/Broken pipe/;
-        confess "syswrite failed: $!";
+    my $wb = syswrite($to_fd, "$json"); # wb = write buffer
+    if (!$wb || $wb != length($json)) { # is true when wb is empty or when the lenght of wb is not the entire json lenght
```
perlpunk, line 42: Rather than commenting stuff we should aim for making the code clearer, for example by using descriptive variable names.
In this case it seemed to cause a misunderstanding. `$wb` is not the buffer, but the return value of `syswrite`, which is `Returns the number of bytes actually written`.
With that I hope we can go without a comment. `!$number` is basically short for `$number == 0` in perl:
```suggestion
    my $written_bytes = syswrite($to_fd, "$json");
    if (!$written_bytes || $written_bytes != length($json)) {
```

myjsonrpc.pm
```diff
@@ -38,10 +38,10 @@ sub send_json ($to_fd, $cmd) {
...
+        die('myjsonrpc: remote end terminated connection, stopping') if !DEBUG_JSON && $! =~ qr/Broken pipe/; # terminate the script if we are not debugging and the syserror is Broken pipe 
+        confess "syswrite failed: $!"; # write to output the error
```
perlpunk, line 44: That's commenting obvious things.
Of course you might argue that it is not that obvious for a perl newbie, but then we would end up with thousands of comments in our code :)

t/24-myjsonrpc.t
Wabri: I set to 0 not to 1

myjsonrpc.pm
Wabri: As a draft this is not a ready pr, I'll remove the comment while refactoring :+1: 

I thought that `!$wb` was used to store the output, thanks for the hint :pray: I'll rename the variable as you suggest :+1:

myjsonrpc.pm
Wabri: Yes, I've never use perl so this helps me remember what I've read. As I mention above I'll remove every comments on my code that is not strictly necessary.

Martchus (commented):
The commit "[Update readability on send_json function](https://github.com/os-autoinst/os-autoinst/pull/2578/commits/70cb075540fdc561b106d543cdeee669d1115445) should probably be squashed with the commit that initially changed that function.

The rest looks good so far although https://github.com/os-autoinst/os-autoinst/blob/master/t/24-myjsonrpc-debug.t might be the better place to add those tests to.

t/24-myjsonrpc.t
```diff
@@ -56,11 +59,28 @@ sub magic_close () {
...
     like($warnings[0], qr{received magic close});
 };
 
+subtest 'send_json dies when buffer is empty and pipe is broken' => sub {
+    local *CORE::GLOBAL::syswrite = sub {
+        $! = 'broken pipe';
+        return 0;
+    };
+    dies_ok { send_json($child, undef) };
+};
```
perlpunk, line 74: You can't easily override builtin functions in a different package like that.
But I think we can test this without overriding, by passing a filehandle that is not for writing:
```suggestion
subtest 'syswrite error handling' => sub {
    my @warnings = warnings {
        throws_ok {
            myjsonrpc::send_json(\*STDIN, {cmd => 'something'})
        } qr/syswrite failed: Bad file descriptor/, 'send_json dies when syswrite fails';
    };
    like $warnings[0], qr/Filehandle STDIN opened only for input/, 'got expected warning';
};

```

t/24-myjsonrpc.t
```diff
@@ -56,11 +59,28 @@ sub magic_close () {
...
+
+subtest 'send_json succeeds when buffer is empty and pipe is NOT broken' => sub {
```
perlpunk, line 76: The successful case is already tested, so we don't need that

t/24-myjsonrpc.t
Wabri: Miss that and yes, I've done several tests around mocking the syswrite buildin function, but with no results.

To solve it I've extracted from the send_json function the syswrite into another function in order to mock that and it works, I've updated the code right now!

t/24-myjsonrpc.t
Wabri: Yes, I've already removed with my latest commits!

t/24-myjsonrpc.t
Wabri: As you seen in my latest commit:

```perl
subtest 'send_json dies when buffer is empty and pipe is broken' => sub {
    my $myjsonrpc_mock = Test::MockModule->new('myjsonrpc');
    $myjsonrpc_mock->redefine(get_written_bytes => sub ($to_fd, $json) {
            return 0;
    });
    dies_ok { myjsonrpc::send_json($child, $send1) } 'myjsonrpc: remote end terminated connection, stopping';
};
```

with get_written_bytes that is a function in myjsonrpc.pm that simply bind the syswrite buildin function:

```perl
sub get_written_bytes($to_fd, $json) {
    return syswrite($to_fd, "$json");
}
```

It is not elegant, what do you think?

myjsonrpc.pm
```diff
@@ -14,6 +14,10 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+sub get_written_bytes($to_fd, $json) {
+    return syswrite($to_fd, "$json");
+}
```
okurz, line 19:
```suggestion
sub _syswrite($to_fd, $json) { syswrite($to_fd, "$json") }
```

Martchus: And also avoid the string interpolation:

```suggestion
sub _syswrite($to_fd, $json) { syswrite($to_fd, $json) }
```

We could actually skip the argument processing here to make it even simpler and more lightweight:

```suggestion
sub _syswrite(@_) { syswrite(@_) }
```

okurz:
what about the "not-ready" label?

okurz (changes requested):
Please also add the file myjsonrpc.pm in codecov in the "fully_covered" section

myjsonrpc.pm
```diff
@@ -95,22 +98,11 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
...
+
+        wait_read($fd) until (my @res = $s->can_read);
 
         my $qbuffer;
-        my $bytes = sysread($socket, $qbuffer, READ_BUFFER);
-        if (!$bytes) {
-            bmwqemu::fctwarn("sysread failed: $!") if DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC};
-            return;
-        }
+        if (!sysread($socket, $qbuffer, READ_BUFFER)) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(), return }
```
perlpunk, line 105: You can use a semicolon here:
```suggestion
        if (!sysread($socket, $qbuffer, READ_BUFFER)) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(); return }
```

perlpunk (changes requested):

perlpunk (commented):
It's ok to keep commits separate if they are about different things, but I think in those 7 commits you rename and rearrange things multiple times, so I suggest to squash some commits.

perlpunk:
Also please rebase on current master. The coverage report shows several unrelated coverage changes.
It's a good idea to fetch origin and rebase a PR whenever you push new changes, daily.

myjsonrpc.pm
```diff
@@ -14,6 +14,21 @@ use bmwqemu ();
...
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+# hash for keeping state
+our $sockets;
+
+sub _syswrite($to_fd, $json) { syswrite($to_fd, $json) }
+
+sub is_debug () { DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC} }
+
+sub wait_read ($fd) {
```
Martchus, line 24: The function is not waiting for anything:
```suggestion
sub handle_read_error ($fd) {
```
(Of course the function `can_read` is also badly named because that one *does* the waiting and is not called `wait_…`.)

Martchus (commented):
Looks good, I only have a small nitpick on one of the function names.

Wabri:
Done rebase and solve all the code review

myjsonrpc.pm
perlpunk: I still see a comma

Wabri: Try now!

okurz (approved):

Martchus (approved):

perlpunk (approved):

myjsonrpc.pm
```diff
@@ -14,6 +14,21 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+# hash for keeping state
+our $sockets;
```
perlpunk, line 18: This doesn't need to be a global package var

```suggestion
my $sockets;
```
* Fixed in #2960

PR #2582, opened 2024-12-06
Add logging and error handling

perlpunk:
We don't check if the eval failed.

Also adding some more logging.

Issue: https://progress.opensuse.org/issues/166445

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -275,8 +275,10 @@ sub run_all () {
     }
     eval {
         bmwqemu::save_vars(no_secret => 1);
+        bmwqemu::diag("Sending tests_done");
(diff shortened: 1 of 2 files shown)


r-richardson (approved):

baierjan (approved):

autotest.pm
```diff
@@ -275,8 +275,10 @@ sub run_all () {
     }
     eval {
         bmwqemu::save_vars(no_secret => 1);
+        bmwqemu::diag("Sending tests_done");
```
okurz, line 278: this triggered now unhandled output in test output, e.g. see https://github.com/os-autoinst/os-autoinst/actions/runs/12260500123/job/34205258725#step:3:526

I opened https://progress.opensuse.org/issues/174094 . Should be easy to cover.

PR #2584, opened 2024-12-07
Extend sshVirtsh console test coverage

b10n1k:
Introduced a subtest to verify `add_input`, `add_interface` and `add_pty`. 
Some statements marked as uncovered

https://progress.opensuse.org/issues/168370

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -440,7 +440,7 @@ sub _handle_disk_type ($vmm_family, $cdrom, $dev_id) {
     return ("hd$dev_id", 'ide') if $vmm_family eq 'vmware';
     return ("hd$dev_id", 'ide') if $cdrom && $vmm_family eq 'kvm';
     return ("vd$dev_id", 'virtio') if $vmm_family eq 'kvm';
-    return (undef, undef);
+    return (undef, undef);    # uncoverable statement
 }
 
 sub _bootorder_elem ($doc, $bootorder) {
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -96,6 +96,7 @@ sub _is_xml ($actual_xml_data, $expected_xml_file_path) {
 
 subtest 'XML config for VNC and serial console' => sub {
     $svirt_console->_init_xml();
+    $testapi::password = 'secret';
     $svirt_console->add_vnc({port => 5901});
     $svirt_console->add_pty({target_port => SERIAL_CONSOLE_DEFAULT_PORT});
     $svirt_console->add_pty({pty_dev => SERIAL_TERMINAL_DEFAULT_DEVICE, pty_dev_type => 'pty', target_port => SERIAL_TERMINAL_DEFAULT_PORT});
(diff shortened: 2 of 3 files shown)


okurz:
s/converage/coverage/

consoles/sshVirtsh.pm
```diff
@@ -488,6 +488,7 @@ sub add_disk ($self, $args) {
     return;
 }
 
+# uncoverable block
```
okurz, line 491: What for? Shouldn't be so hard to just cover some variable settings. Just call it in tests.

okurz (dismissed):

consoles/sshVirtsh.pm
b10n1k: hard maybe not. but it is not essential IMO. What do you really get from those test?

perlpunk: We should cover everything, as long as it's possible and feasible.
Covering code in tests ensures that the code doesn't break.
If we start discussing about which functions are too simple to be broken, we're going down a rabbit hole, and should rather use the time to just cover the code.

Code coverage does not guarantee 100% that code will work for any case, but it is *much* better than uncovered code.

t/22-svirt.t
```diff
@@ -109,6 +109,68 @@ subtest 's390x specifics' => sub {
...
+    my $virsh = consoles::sshVirtsh::virsh();
+    is $virsh, 'virsh my_vmm', 'correct output from virsh()';
+};
+
+subtest 'check suspend() method' => sub {
+    my $console_mock = Test::MockModule->new('consoles::sshVirtsh');
+    my @cmds;
+    $console_mock->redefine(run_cmd => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $svirt_console->suspend();
+    is_deeply @cmds, 'virsh my_vmm suspend openQA-SUT-1', 'correct command from suspend()';
```
perlpunk, line 157: If you pass an array here, and it would have more than one item, the other items would serve as parameters for the `is_deeply` function.
either:
```suggestion
    is $cmds[0], 'virsh my_vmm suspend openQA-SUT-1', 'correct command from suspend()';
```
or
```suggestion
    is_deeply \@cmds, ['virsh my_vmm suspend openQA-SUT-1'], 'correct command from suspend()';
```

t/22-svirt.t
```diff
@@ -109,6 +109,68 @@ subtest 's390x specifics' => sub {
...
+};
+
+subtest 'check resume() method' => sub {
+    my $console_mock = Test::MockModule->new('consoles::sshVirtsh');
+    my @cmds;
+    $console_mock->redefine(run_cmd => sub ($self, $cmd, %args) { push @cmds, $cmd; 0 });
+    $svirt_console->resume();
+    is_deeply @cmds, 'virsh my_vmm resume openQA-SUT-1', 'correct command from suspend()';
```
perlpunk, line 165: Same as above

perlpunk (changes requested):

Martchus (approved):

perlpunk (approved):

PR #2585, opened 2024-12-11
Ensure full statement coverage in consoles::sshVirtsh

b10n1k:
Mark remaining uncovered lines to uncoverable and add the console to the codecov.yml as it should measure 100% coverage with that latest commit.

https://progress.opensuse.org/issues/168370

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -144,8 +144,8 @@ sub _init_xml ($self, $args = {}) {
         }
         if (!$bmwqemu::vars{BIOS}) {
             # We know this won't go well.
-            my $virsh_hostname = $bmwqemu::vars{VIRSH_HOSTNAME} // '';
-            die "No UEFI firmware can be found on hypervisor '$virsh_hostname'. Please specify BIOS or UEFI_BIOS or install an appropriate package.";
+            my $virsh_hostname = $bmwqemu::vars{VIRSH_HOSTNAME} // '';    # uncoverable statement
+            die "No UEFI firmware can be found on hypervisor '$virsh_hostname'. Please specify BIOS or UEFI_BIOS or install an appropriate package."; # uncoverable statement
         }
     }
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -144,8 +144,10 @@ sub _init_xml ($self, $args = {}) {
         }
         if (!$bmwqemu::vars{BIOS}) {
             # We know this won't go well.
+            # uncoverable block
```
perlpunk, line 147: Where did you find this? This isn't listed as a way to mark uncoverable lines.
https://metacpan.org/pod/Devel::Cover#UNCOVERABLE-CRITERIA

Wabri: I think is possible to use `# uncoverable branch true` to skip the entire if

b10n1k: checking. 
@Wabri `# uncoverable branch true` didnt work(coverage didnt change). I try with `false`. Maybe it needs both?! give me some time

perlpunk: Some background: Devel::Cover supports statement and branch coverage.
If we used that for our reports, our coverage percentage would actually drop. We are only using line coverage right now, mainly because that's the only thing that codecov supports. (And I guess supporting branch coverage as a provider for code coverage reporting for multiple languages would be too much effort)

So `# uncoverable branch` is only useful if we actually enable branch coverage.
It's necessary in all cases to add `# uncoverable statement` to each line.
And btw you can just use `make coverage TESTS=t/...` to try it out locally and see the Devel::Cover report. There's no need to wait for codecov every time.

b10n1k: I followed the `# uncoverable statement` approach as per @perlpunk's last comment. please review again

perlpunk (approved):

okurz (approved):

PR #2586, opened 2024-12-11
Slightly simplify myjsonrpc

okurz:
https://progress.opensuse.org/issues/167935

Potential further step: https://github.com/okurz/os-autoinst/tree/feature/simplify_extract_loop

Related to
* #2578

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -14,6 +14,8 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
+sub is_debug () { DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC} }
+
 sub send_json ($to_fd, $cmd) {
     # allow regular expressions to be automatically converted into
     # strings, using the Regex::TO_JSON function as defined at the end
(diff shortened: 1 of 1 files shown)


myjsonrpc.pm
```diff
@@ -74,10 +70,7 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
         # remember the trailing text
         if ($hash) {
             $sockets->{$fd} = $cjx->incr_text();
-            if (DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC}) {
-                my $token = $hash->{json_cmd_token} // 'no-token';
-                bmwqemu::diag("read_json($fd) json_cmd_token=$token");
-            }
+            bmwqemu::diag("read_json($fd) json_cmd_token=" . $hash->{json_cmd_token} // 'no-token') if is_debug;
```
Martchus, line 73:
```suggestion
            bmwqemu::diag("read_json($fd) json_cmd_token=" . ($hash->{json_cmd_token} // 'no-token')) if is_debug;
```

myjsonrpc.pm
```diff
@@ -95,22 +88,16 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
...
+            bmwqemu::diag("read_json($fd): can_read's underlying system call has been interrupted, trying again\n") if is_debug;
         }
 
         my $qbuffer;
-        my $bytes = sysread($socket, $qbuffer, READ_BUFFER);
-        if (!$bytes) {
-            bmwqemu::fctwarn("sysread failed: $!") if DEBUG_JSON || $bmwqemu::vars{DEBUG_JSON_RPC};
-            return;
-        }
+        my $bytes = sysread($socket, $qbuffer, READ_BUFFER) or do { bmwqemu::fctwarn("sysread failed: $!") if is_debug and return undef };
```
Martchus, line 100: This is really hard to read. It is really unclear what the left-hand-side of the `and` evaluates to if `is_debug` is false and it also looks like `is_debug and return undef` is all just part of the `if` condition although the `and return undef` part is actually not (meant) to be part of it.

I think the following will work (provided `bmwqemu::fctwarn` returns undef as well):
```suggestion
        my $bytes = sysread($socket, $qbuffer, READ_BUFFER) or return (is_debug ? bmwqemu::fctwarn("sysread failed: $!") : undef;
```

Martchus (changes requested):

myjsonrpc.pm
perlpunk: But if we have a `do` block there's no reason to just use multiple statements separated by `;`.

perlpunk: I don't have a problem with the original code, and putting it in one line is also just cheating with code coverage, and explicitly marking those lines as uncoverable would make it more clear that this **is** actually uncovered.

Wabri:
I've cherry-picked during my work on #2578 and than mixed with the suggestions given by @Martchus => https://github.com/os-autoinst/os-autoinst/pull/2578/commits/fb19b3c34dfec2dc4d46b622e701f77c333da5c7

PR #2590, opened 2024-12-13
Cover lockapi::mutex_wait

Wabri:
Add lockapi::mutex_wait test in t/30-mmapi.t

https://progress.opensuse.org/issues/167938

PR #2593, opened 2024-12-16
 Remove _typing_limit subroutine because it's never used 

Wabri:
https://progress.opensuse.org/issues/174271

### DIFF
--- a/consoles/vnc_base.pm
+++ b/consoles/vnc_base.pm
@@ -66,8 +66,6 @@ sub current_screen ($self) {
     return $self->{vnc}->_framebuffer;
 }
 
-sub _typing_limit () { $bmwqemu::vars{VNC_TYPING_LIMIT} // VNC_TYPING_LIMIT_DEFAULT || 1 }
-
 sub send_key_event ($self, $key, $press_release_delay) {
     $self->{vnc}->map_and_send_key($key, undef, $press_release_delay);
 }
--- a/codecov.yml
+++ b/codecov.yml
@@ -28,6 +28,7 @@ coverage:
           - consoles/sshVirtsh.pm
           - consoles/sshIucvconn.pm
           - consoles/virtio_terminal.pm
+          - consoles/vnc_base.pm
           - myjsonrpc.pm
           - lockapi.pm
           - basetest.pm


consoles/vnc_base.pm
```diff
@@ -66,7 +66,9 @@ sub current_screen ($self) {
     return $self->{vnc}->_framebuffer;
 }
 
-sub _typing_limit () { $bmwqemu::vars{VNC_TYPING_LIMIT} // VNC_TYPING_LIMIT_DEFAULT || 1 }
+sub _typing_limit () {
```
perlpunk, line 69: It seems this subroutine is never used anywhere. Maybe we should remove it instead.

Wabri: Also the #2590 subroutine was never used, but we decided to keep that. I've not strong opinion whether keep it or not.

perlpunk: `_typing_limit` has a leading underscore, so by convention it is a subroutine that should only be called from that package. So if we don't use it in that package, IMHO we should remove it. Or alternatively use it in the places where we currently use `VNC_TYPING_LIMIT_DEFAULT`.

perlpunk: `mutex_wait` is actually used: https://github.com/search?q=repo%3Aos-autoinst%2Fos-autoinst-distri-opensuse%20mutex_wait&type=code

Martchus: Yes, just because it isn't used within os-autoinst itself it doesn't mean it isn't used.

okurz: The removal of `_typing_limit` was overlooked in 007c1837. Please remove the function

okurz (changes requested):

consoles/vnc_base.pm
Wabri: Ok, done it!

perlpunk (approved):

perlpunk:
The PR title still needs to be updated

kalikiana (approved):

Wabri:
I've fixed the typo on the commit and updated the title!

okurz (approved):

PR #2601, opened 2024-12-24
Modify loadtestdir in autotest for more code coverage

gauravpathak:


okurz (approved):

t/08-autotest.t
```diff
@@ -449,9 +449,14 @@ subtest make_snapshot => sub {
...
-        autotest::loadtestdir('test-snapshot');
-    } qr/does not exist!/, 'test died as expected';
+    use lib qw(t/data/tests);
+    $bmwqemu::vars{CASEDIR} = 't/data/tests';
+    stderr_like {
+        like warning {
+            autotest::loadtestdir('tests');
+        }, qr/ARRAY/, 'script found';
+    } qr/scheduling/, 'loadscript success';
+
```
Wabri, line 459: Just to be sure that loadtestdir have load the directory we should check if the package variable `tests` contains one of the script in the directory, for example `boot.pm` the loadtest subroutine should create a `tests-boot` element
```suggestion
ok exists $autotest::tests{'tests-boot'}, 'boot.pm loaded';
```

If you want to check the content of the `tests` variable there is [`Data::Dumper`](https://metacpan.org/pod/Data::Dumper) module that can be use to translate data structures into string.

Wabri (changes requested):

autotest.pm
```diff
@@ -281,7 +281,7 @@ sub run_all () {
 }
 
 sub handle_sigterm ($sig) {    # uncoverable statement
```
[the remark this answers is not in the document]
okurz: That's not enough. One still needs to add uncoverable statement to every statement

Wabri (approved):

Wabri:
Please rebase!

kalikiana:
@mergifyio rebase

t/08-autotest.t
```diff
@@ -449,9 +449,14 @@ subtest make_snapshot => sub {
 };
 
 subtest loadtestdir => sub {
-    throws_ok {
-        autotest::loadtestdir('test-snapshot');
-    } qr/does not exist!/, 'test died as expected';
+    use lib 't/data/tests';
```
perlpunk, line 452: This is something that shouldn't be necessary. autotest should add the necessary include paths itself.
It seems to me `loadtestdir` hasn't been used for many years, not in our code and not in osado.
The way it works produces warnings, and below you are just kind of ignoring the warnings by comparing it to `ARRAY`, which is just a stringification of an array ref.
The warning is
```
loadtest needs a script below t/data/tests - t/data/tests/tests/assert_screen.pm is not 
```
That's because of the way `loadtestdir` passes the tests to `loadtest`.

I think we should rather fix `loadtestdir` instead, or remove it.

perlpunk:
What's the corresponding issue for this PR?

t/08-autotest.t
gauravpathak: So, Should we remove `loadtestdir` completely?

perlpunk: https://github.com/search?q=loadtestdir+language%3APerl&type=code&l=Perl
maybe not...
The problematic part seems to be that the `glob()` in `loadtestdir` can return a part of the path that is already in CASEDIR, and then `find_script` will complain about `(!-f join('/', $casedir, $script)`.

gauravpathak, line 452: I made a few changes. First, I attempted to check for the warning `loadtest needs a script below` instead of using `qr/ARRAY/`. However, I was still getting the following output:
```
# Subtest: loadtestdir
    not ok 1 - script found

    #   Failed test 'script found'
    #   at t/08-autotest.t line 456.
    #                   'ARRAY(0x55c13921d970)'
    #     doesn't match '(?^u:loadtest needs a script below)'
    ok 2 - loadscript success
    ok 3 - boot.pm loaded
    1..3
    # Looks like you failed 1 test of 3.
```

Additionally, to avoid using `use lib 't/data/tests';`, I had to set the casedir as follows:
`$bmwqemu::vars{CASEDIR} = Cwd::getcwd();`

I also provided the path to the tests directory when calling `loadtestdir`. Here’s the updated test code snippet:

```
subtest loadtestdir => sub {
    $bmwqemu::vars{CASEDIR} = Cwd::getcwd();
    stderr_like {
        like warning {
            autotest::loadtestdir('t/data/tests/tests');
        }, qr/loadtest needs a script below/, 'script found';
    } qr/scheduling/, 'loadscript success';
    ok exists $autotest::tests{'tests-boot'}, 'boot.pm loaded';
};
```
Could you please confirm if we can use the above test code snippet for `loadtestdir`?

gauravpathak: Tried some more changes, the catch is in the way casedir is set. If I use below code where I use relative path of current dir as casedir, then it doesn't throw the warning

```
subtest loadtestdir => sub {
    $bmwqemu::vars{CASEDIR} = '.';
    stderr_like {
        autotest::loadtestdir('t/data/tests/tests');
    } qr/scheduling/, 'loadscript success';
    ok exists $autotest::tests{'tests-boot'}, 'boot.pm loaded';
};
```

PR #2603, opened 2024-12-24
Cover code of os-autoinst path needle.pm fully

Wabri:
https://progress.opensuse.org/issues/167932


needle.pm
```diff
@@ -42,6 +42,8 @@ sub new ($classname, $jsonfile) {
 
     my $self = {};
 
+    # warn "Looking for JSON file: $jsonfile\n";
+
```
perlpunk, line 46:
```suggestion
```

t/misc_needles/test_tag1.json
```diff
@@ -0,0 +1,20 @@
+{
```
perlpunk, line 1: Could those needles be a bit more minimalized, e.g. remove things like clickpoint and other stuff that's not important for the actual test?

perlpunk (changes requested):
It's ok to have separate commits to make several changes more comprehensible, but this is a bit too detailed, and also
* there are commits updating previous code, those should be squashed
* some commits have a title that conflicts with the actual diff

needle.pm
```diff
@@ -156,10 +138,7 @@ sub unregister ($self, $reason = undef) {
 sub register ($self) {
     my %check_dups;
     for my $g (@{$self->{tags}}) {
-        if ($check_dups{$g}) {
-            bmwqemu::diag("$self->{name} contains $g twice");
-            next;
-        }
+        bmwqemu::diag("$self->{name} contains $g twice"), next if ($check_dups{$g});
```
perlpunk, line 141: Have you tried to cover this?
If something is not too difficult to cover, we should do it. The ideal goal is to have 100% coverage, not to appear every line as covered although it really isn't.
I know I disagree with @okurz here but I think we shouln't overuse this one-line trick.

Wabri:
Fix incoming :+1: 

image

needle.pm
Martchus: And I must also say that using the comma operator like this is probably not a good idea as it makes the code less readable (and is it even correct?).

needle.pm
```diff
@@ -125,26 +127,6 @@ sub new ($classname, $jsonfile) {
     return $self;
 }
 
-sub save ($self, $fn = $self->{file}) {
```
Martchus, line 128: Are you removing this function because it is unused? I'm wondering whether this function is really completely useless (e.g. it might be useful during development/debugging) but I'm also not against removing it (as the function is not very well written anyway).

Martchus (changes requested):
Looks like the PNG images you're adding here actually contain JSON data. That's confusing. The the PNG images doesn't need to be valid you can just add empty files or a minimal PNG or a symlink to some other PNG we already have in this repo.

needle.pm
perlpunk: See https://progress.opensuse.org/issues/167932#note-11 where I investigated if the function is used.
Note that it already got a bug 5 years ago.

When you say "useful during development/debugging", what do you mean by that? During openQA development, or during developing an test openQA test? When and how would you use it?

Martchus: During development of os-autoinst itself to dump a needle for debuging purposes. But as I said, I'm fine with removing the function.

needle.pm
```diff
@@ -119,32 +119,12 @@ sub new ($classname, $jsonfile) {
     my $png = $self->{png} || $self->{name} . ".png";
 
     $self->{png} = path(dirname($jsonfile), $png)->to_string;
-    warn "Can't find $self->{png}" and return unless -s $self->{png};
+    warn "Can't find $self->{png}" and return unless -e $self->{png};
```
Wabri, line 122:
```
    -e: Checks if the file exists.
    -s: Checks if the file is non-empty.
```

Since the warning is for file not found I've updated with a check of the existing file

Wabri (commented):
@Martchus I've done a copy paste of https://github.com/os-autoinst/os-autoinst/blob/master/t/misc_needles/click-point.png so it is already a symlink, but anyway it points to the json file

t/01-test_needle.t
```diff
@@ -37,6 +40,56 @@ throws_ok(
...
+    }, qr/broken json.*broken\.json/, 'warning shown for broken JSON file');
+};
+
+subtest 'handle tags duplicates' => sub {
+    my $sandbox = tempdir(CLEANUP => 1);
+    needle::set_needles_dir($sandbox);
+    my $tag_json_path = path($sandbox, 'tag.json');
+    my $filename = path($sandbox, 'tag.png');
+
+    sysopen(my $fh, $filename, O_WRONLY | O_CREAT, 0777) or die "Cannot create file '$filename': $!"; close($fh);
```
perlpunk, line 72: Why are you using sysopen here and not just a normal open?

Wabri: No real reason, I've just found out that method to create a new file. Do you think something like this could be better?
```suggestion
    open(my $fh, '>', $tag_png_path) or die "Cannot open file '$tag_png_path': $!"; close($fh);
```

perlpunk: Or just use
```
$filename->spew('');
```
You are using that for the json file, so why not for the png file as well

needle.pm
perlpunk: That changes the behaviour, and maybe it is good to check for a non-empty file though. Maybe the warning should be updated to reflect that instead.

t/01-test_needle.t
Wabri: I thought that works only for json encoding, but the subroutine takes string so make sense to just use also to create file

t/01-test_needle.t
```diff
@@ -37,6 +39,56 @@ throws_ok(
...
+};
+
+subtest 'handle tags duplicates' => sub {
+    my $sandbox = tempdir(CLEANUP => 1);
+    needle::set_needles_dir($sandbox);
+    my $tag_json_path = path($sandbox, 'tag.json');
+    my $tag_png_path = path($sandbox, 'tag.png');
+
+    open my $fh, '>', $tag_png_path or die $!; close($fh);
+    path($tag_json_path)->spew('{"area": [{"x" : 123, "y" : 456}],"tags": ["tag1", "tag1"]}');
```
perlpunk, line 72:
```suggestion
    path($sandbox, 'tag.png')->spew('');

    $tag_json_path->spew('{"area": [{"x" : 123, "y" : 456}],"tags": ["tag1", "tag1"]}');
```

t/01-test_needle.t
```diff
@@ -37,6 +39,56 @@ throws_ok(
...
+    combined_like { $needle = needle->new($tag_json_path->basename) } qr/\[debug\].*tag contains tag1 twice/, 'tag contains tag1 twice';
+    is($needle->has_tag('tag1'), 1, "tag found");
+};
+
+subtest 'handle invalid click point' => sub {
+    my $sandbox = tempdir(CLEANUP => 1);
+    needle::set_needles_dir($sandbox);
+    my $invalid_click_point_json_path = path($sandbox, 'invalid-click-point.json');
+
+    path($invalid_click_point_json_path)->spew('{"area": [{"click_point": "invalid"}]}');
```
perlpunk, line 84:
```suggestion
    $invalid_click_point_json_path->spew('{"area": [{"click_point": "invalid"}]}');
```

t/01-test_needle.t
```diff
@@ -37,6 +39,56 @@ throws_ok(
...
+
+    like(warning {
+            my $needle = needle->new($invalid_click_point_json_path->basename);
+            is($needle, undef, 'needle object not created with invalid click point');
```
perlpunk, line 88:
```suggestion
            is $needle, undef, 'needle object not created with invalid click point';
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+    my $tag3 = 'tag3';
+    needle::set_needles_dir($misc_needles_dir);
+
+    needle->new('test_tag1.json');
+    needle->new('test_tag2.json');
+    needle->new('test_tag3.json');
+
+    for my $n (needle::all()) {
+        $n->register();
+    }
```
perlpunk, line 560:
```suggestion
    $_->register() for needle::all();
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+
+    my $result = needle::tags($tag1);
+    print "@$result\n";
+    is(scalar @$result, 2, "two needles found for tag1");
```
perlpunk, line 564:
```suggestion
    is scalar @$result, 2, "two needles found for tag1";
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+    $result = needle::tags($tag2);
+    is(scalar @$result, 2, "two needles found for tag2");
```
perlpunk, line 566:
```suggestion
    is scalar @$result, 2, "two needles found for tag2";
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+    $result = needle::tags($tag3);
+    is(scalar @$result, 1, "one needle found for tag3");
```
perlpunk, line 568:
```suggestion
    is scalar @$result, 1, "one needle found for tag3";
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+    $result = needle::tags('nonexistent');
+    is(scalar @$result, 0, "no needles found for nonexistent tag");
```
perlpunk, line 570:
```suggestion
    is scalar @$result, 0, "no needles found for nonexistent tag";
```

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
 
+subtest 'test tags method' => sub {
+    my $tag1 = 'tag1';
+    my $tag2 = 'tag2';
+    my $tag3 = 'tag3';
+    needle::set_needles_dir($misc_needles_dir);
+
+    needle->new('test_tag1.json');
+    needle->new('test_tag2.json');
+    needle->new('test_tag3.json');
```
perlpunk, line 556:
```suggestion
    needle->new($_) for qw(test_tag1.json test_tag2.json test_tag3.json);
```

t/01-test_needle.t
```diff
@@ -37,6 +39,56 @@ throws_ok(
...
+    } qr/Needle $invalid_json_path is not under needle directory $misc_needles_dir/,
+      'throws error when needle JSON file is not under needle directory';
+};
+
+subtest 'handle broken JSON file' => sub {
+    my $sandbox = tempdir(CLEANUP => 1);
+    needle::set_needles_dir($sandbox);
+    my $broken_json_path = path($sandbox, 'broken.json');
+
+    path($broken_json_path)->spew('{ "tags": ["test');
```
perlpunk, line 57:
```suggestion
    $broken_json_path->spew('{ "tags": ["test');
```

perlpunk (commented):
Added some more simplifications

needle.pm
Wabri: Using -s will break the handle tags duplicates, but I'll populate the tag.png with random string!

Martchus:
> @Martchus I've done a copy paste of https://github.com/os-autoinst/os-autoinst/blob/master/t/misc_needles/click-point.png so it is already a symlink, but anyway it points to the json file

Ah, then the view on GitHub fooled me. I guess as a symlink it is kind of ok (and I introduced this kind of symlink). So you can keep it.

Martchus (approved):

t/01-test_needle.t
```diff
@@ -491,4 +545,29 @@ subtest 'clarify error message when needles directory does not exist' => sub {
...
+    needle->new('test_tag1.json');
+    needle->new('test_tag2.json');
+    needle->new('test_tag3.json');
+
+    for my $n (needle::all()) {
+        $n->register();
+    }
+
+    my $result = needle::tags($tag1);
+    print "@$result\n";
```
perlpunk, line 563:
```suggestion
```

perlpunk (approved):

Wabri:
I didn't understand why code coverage fail even if the coverage is 100%, I've rebased with master to check if is something related to it

okurz (approved):
Looks like codecov has some old data or is incorporating the pre-state for fully covered? You can try to exclude the codecov.yml changes from this PR and we can try to bring it in in another PR

needle.pm
perlpunk: I would just use `YAML::PP::DumpFile('file.yaml', $needle)` if I wanted to debug the content into a file...

okurz:
Now https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/needle.pm shows what looks still missing even though that looks different than the patch report in this PR

Wabri:
Why it didn't show previously? neither locally nor remotely? Working on it!

PR #2606, opened 2024-12-27
Remove redundant newline in "GOT GO" message

okurz:


Martchus (approved):

autotest.pm
```diff
@@ -312,10 +312,9 @@ sub start_process () {
 
             $0 = "$0: autotest";
             my $line = <$isotovideo>;
-            if (!$line) {
-                _exit(0);    # uncoverable statement
-            }
-            bmwqemu::diag("GOT $line\n");
+            _exit(0) unless $line;    # uncoverable statement
+            chomp($line);
```
Wabri, line 316: Why chomp is necessary?

okurz: Because $line itself can contain a new line as received from the backend. So this commit removes *at least* one newline, often even two ;)

Wabri (approved):

PR #2550, opened 2024-09-27
Enable multifd support

ggardet:
From qemu 9.1.0, the "compress" migration capability has been removed, but multifd migration is able to do compression and can be used instead.

https://progress.opensuse.org/issues/167218

PR #2607, opened 2025-01-02
Enable multifd support enhanced with fully coverage

Wabri:
This pr grants the fully coverage for #2550

### Differences

- set the non-coverable states correctly in qemu.pm
- update 18-backend-qemu.t tests by removing unreachable states

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -251,6 +251,7 @@ sub _migrate_to_file ($self, %args) {
 
     # Internally compressed dumps can't be opened by crash. They need to be
     # fed back into QEMU as an incoming migration.
+    # uncoverable branch false
     if ($migration_use_multifd) {
         $self->set_migrate_capability('multifd', 1);
         $self->set_migrate_capability('mapped-ram', 1);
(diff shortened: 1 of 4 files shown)


okurz (approved):
very nice!

okurz:
merged as part of https://github.com/os-autoinst/os-autoinst/pull/2608

AdamWill:
Um. Is it nice? It just marked a bunch of stuff as uncoverable (which just makes number go up, it doesn't solve the problem that we're not covering them!) and made an incorrect statement: "removed else statements because being a test we know a priori what the state of the variables is" is not true, because the state of the variables in question depends on the version of qemu installed on the system running the tests. Those changes will cause the tests to fail if run on older qemu versions.

Even on its own terms that commit seems incomplete, because if you drop one half of the conditional, why not just remove the conditional statement entirely? Instead it now *looks* like the test ought to work for older qemu (because we still have the `if ($backend->{qemu_version} ge version->declare(9.1))` bits) but in fact it will not.

I appreciate the effort but I just don't think this is a good resolution.

PR #2608, opened 2025-01-03
Feature/167218/multifd support

okurz:
same as  https://github.com/os-autoinst/os-autoinst/pull/2607 but w/o https://github.com/os-autoinst/os-autoinst/pull/2607/commits/8afaeb19de20bca112433075673c9d2a3fc5cd9f

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -251,6 +251,7 @@ sub _migrate_to_file ($self, %args) {
 
     # Internally compressed dumps can't be opened by crash. They need to be
     # fed back into QEMU as an incoming migration.
+    # uncoverable branch false
     if ($migration_use_multifd) {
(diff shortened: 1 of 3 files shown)


AdamWill:
Per my comment in https://github.com/os-autoinst/os-autoinst/pull/2607#issuecomment-2569545906 I don't think this is really a good resolution. I just confirmed that if you build os-autoinst with this PR on Fedora 40 - which has qemu 8.2 - the tests fail: https://kojipkgs.fedoraproject.org/work/tasks/6062/127516062/build.log . If you build it without "Remove unreachable states in tests 18-backend-qemu", they pass.

AdamWill:
I'm trying to work up a proper fix for this now - we actually already mock out the backend to some extent so it shouldn't be hard, unless I hit something unexpected.

PR #2609, opened 2025-01-03
Properly fix coverage of qemu codepaths that depend on version (see https://progress.opensuse.org/issues/167218 )

AdamWill:
This reverts @Wabri 's changes from https://github.com/os-autoinst/os-autoinst/pull/2608 ,which I think were wrong, and instead addresses the missing coverage by mocking the qemu version discovery code so we can always test both paths regardless of what qemu is installed.


AdamWill:
waiting for codecov...

PR #2610, opened 2025-01-06
Add subtests for commands.pm

gauravpathak:
Related progress ticket: https://progress.opensuse.org/issues/167947

### DIFF
--- a/commands.pm
+++ b/commands.pm
@@ -11,10 +11,11 @@ require IPC::System::Simple;
 use Try::Tiny;
 use Socket;
 use POSIX '_exit', 'strftime';
-use myjsonrpc;
-use bmwqemu;
 use Mojo::JSON 'to_json';
 use Mojo::File 'path';
+use myjsonrpc;
+use bmwqemu;
+use testapi ();
 
 BEGIN {
     $ENV{MOJO_REACTOR} = 'Mojo::Reactor::Poll';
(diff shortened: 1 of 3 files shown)


t/07-commands.t
```diff
@@ -203,6 +209,20 @@ subtest 'upload api' => sub {
     };
 };
 
+subtest 'get_temp_file from worker' => sub {
+    $t->get_ok("$base_url/$job/files/non-existing-file.txt")->status_is(500);
```
perlpunk, line 213: A non existing file should actually result in a 404 Not Found, not a 500 Internal Server Error.

perlpunk (changes requested):

t/07-commands.t
gauravpathak: I agree, but I am getting 500 as response.

gauravpathak: I tried putting a `bmwqemu::diag()` statement at https://github.com/os-autoinst/os-autoinst/blob/master/commands.pm#L198 but I didn't get any output.
If I put any `bmwqemu::diag()` statement before `my $path = testapi::hashed_string($relpath);`, I get the output but putting any diag statement after that line doesn't seem to work for me

perlpunk: I actually couldn't get any output directly from `get_temp_file`; but I created a file in /tmp for debugging and put an eval around the `testapi::hashed_string()` call, printed `$@` into that file and got:
```
Undefined subroutine &testapi::hashed_string ...
```
The problem is that commands.pm doesn't load the testapi module, although it uses a function from it.
That only works because when starting isotovideo it loads both commands.pm and testapi.pm. But in the unit test it is missing.
I suggest to add `use testapi ();` in commands.pm.

commands.pm
```diff
@@ -11,10 +11,11 @@ require IPC::System::Simple;
 use Try::Tiny;
 use Socket;
 use POSIX '_exit', 'strftime';
-use myjsonrpc;
-use bmwqemu;
 use Mojo::JSON 'to_json';
 use Mojo::File 'path';
+use myjsonrpc;
+use bmwqemu;
+use testapi();
```
okurz, line 18:
```suggestion
use testapi ();
```

okurz (changes requested):

perlpunk (approved):

Wabri (approved):

perlpunk:
Squash commits please

okurz (approved):

PR #2614, opened 2025-01-08
Apply code review suggestions to loadtestdir subtest

gauravpathak:
Related progress ticket: https://progress.opensuse.org/issues/167926

- remove "use lib 't/data/tests';" from within loadtestdir subtest subroutine
- use relative path for casedir instead of path to test modules directory

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -449,12 +449,9 @@ subtest make_snapshot => sub {
 };
 
 subtest loadtestdir => sub {
-    use lib 't/data/tests';
-    $bmwqemu::vars{CASEDIR} = 't/data/tests';
+    $bmwqemu::vars{CASEDIR} = '.';
     stderr_like {
-        like warning {
-            autotest::loadtestdir('tests');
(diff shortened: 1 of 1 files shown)


t/08-autotest.t
```diff
@@ -449,12 +449,9 @@ subtest make_snapshot => sub {
 };
 
 subtest loadtestdir => sub {
-    use lib 't/data/tests';
-    $bmwqemu::vars{CASEDIR} = 't/data/tests';
+    $bmwqemu::vars{CASEDIR} = '.';
```
perlpunk, line 452: that might work to avoid the warning in this test, but it's not a real world scenario.
I think loadtestdir needs to be fixed.

gauravpathak: It's confusing to me.  :confused:
I am closing the PR and let's file a Progress ticket for improving `loadtestdir()` subroutine

perlpunk: I created https://progress.opensuse.org/issues/175099

PR #2615, opened 2025-01-09
autotest/loadtest: Update test fullname along with name when duplicated

rbmarliere:
Currently, if the developer choose to take control of the test and pause at a duplicate, the job will not pause. The issue is that [_handle_command_set_current_test](https://github.com/os-autoinst/os-autoinst/blob/master/OpenQA/Isotovideo/CommandHandler.pm#L282)  in OpenQA/Isotovideo/CommandHandler.pm needs the correct $full_test_name, which might not be accurate anymore. Fix this by also updating the test "fullname" in loadtest.

Test runs (with some extra debugging prints):
- Without patch:
Pause scheduled: https://rmarliere-openqa.qe.prg2.suse.org/tests/1134/logfile?filename=autoinst-log.txt#line-3624
Fullname mismatch impeding pause: https://rmarliere-openqa.qe.prg2.suse.org/tests/1134/logfile?filename=autoinst-log.txt#line-6969
- With patch:
Pause scheduled: https://rmarliere-openqa.qe.prg2.suse.org/tests/1135/logfile?filename=autoinst-log.txt#line-102
Execution paused: https://rmarliere-openqa.qe.prg2.suse.org/tests/1135/logfile?filename=autoinst-log.txt#line-7059

NOTE: fullname is being sent by the [openQA UI](https://github.com/os-autoinst/openQA/blob/master/assets/javascripts/running.js#L811).

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -177,7 +177,9 @@ sub loadtest ($script, %args) {
 
     my $nr = '';
     while (exists $tests{$fullname . $nr}) {
-        $test->{name} = join "#", $name, ++$nr;
+        my $new_name = join "#", $name, ++$nr;
+        $test->{name} = $new_name;
+        $test->{fullname} = "$category-$new_name";
     }
     $test->{name} = $args{name} if $args{name};
 
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -96,8 +96,9 @@ is(keys %autotest::tests, 2, 'two tests have been scheduled');
 loadtest 'start', 'rescheduling same step later';
 is(keys %autotest::tests, 3, 'three steps have been scheduled (one twice)') || diag explain %autotest::tests;
 is($autotest::tests{'tests-start1'}->{name}, 'start#1', 'handle duplicate tests');
+is($autotest::tests{'tests-start1'}->{fullname}, 'tests-start#1', 'duplicate test has correct fullname');
 is($autotest::tests{'tests-start1'}->{$_}, $autotest::tests{'tests-start'}->{$_}, "duplicate tests point to the same $_")
-  for qw(script fullname category class);
+  for qw(script category class);
 
 like warning {
     stderr_like { autotest::run_all } qr/Sending tests_done/, 'tests_done sent';


autotest.pm
```diff
@@ -177,7 +177,9 @@ sub loadtest ($script, %args) {
 
     my $nr = '';
     while (exists $tests{$fullname . $nr}) {
-        $test->{name} = join "#", $name, ++$nr;
+        my $new_name = join "#", $name, ++$nr;
+        $test->{name} = $new_name;
+        $test->{fullname} = "$category-$new_name";
```
perlpunk, line 182: I guess the `$fullname` variable should also be updated, as it is used 3 lines below

perlpunk (commented):
The test is failing because it explicitly tests for the same fullname:
```
is($autotest::tests{'tests-start1'}->{$_}, $autotest::tests{'tests-start'}->{$_}, "duplicate tests point to the same $_")
  for qw(script fullname category class);                                       
```
but I guess we can move `fullname` out of this. Any other opinions?

autotest.pm
rbmarliere: `$tests{$fullname . $nr} = $test;`

Indeed. It is missing the '#' character.

Martchus:
I suppose we can leave it out. The change looks good in general and judging by your tests I suppose it really does the right thing.

autotest.pm
rbmarliere: Actually, I don't think it's the case. The loop statement `while (exists $tests{$fullname . $nr}) {` already expects that pattern. It's not clear to me where it `$tests` come from so I'll keep it as is.

Martchus: I was also confused about this suggestion when reviewing this change earlier and decided to not write anything to avoid writing something wrong :-)

So take the following with a grain of salt: You don't need to set `$fullname` here because considering how it is used below and in the loop condition it shouldn't contain `$nr`. So the change can stay as-is now that you have adapted tests.

Martchus (approved):

perlpunk:
codecov upload failed; I retriggered

t/08-autotest.t
```diff
@@ -97,7 +97,7 @@ loadtest 'start', 'rescheduling same step later';
 is(keys %autotest::tests, 3, 'three steps have been scheduled (one twice)') || diag explain %autotest::tests;
 is($autotest::tests{'tests-start1'}->{name}, 'start#1', 'handle duplicate tests');
 is($autotest::tests{'tests-start1'}->{$_}, $autotest::tests{'tests-start'}->{$_}, "duplicate tests point to the same $_")
-  for qw(script fullname category class);
+  for qw(script category class);
```
perlpunk, line 100: I suggest to add a new test for fullname though, instead of just leaving it out.

perlpunk (approved):

PR #2621, opened 2025-01-16
Update loadtestdir subroutine with correct paths to loadtest()

Wabri:
https://progress.opensuse.org/issues/175099

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -463,7 +463,7 @@ sub loadtestdir ($dir) {
     $dir =~ s/^\Q$bmwqemu::vars{CASEDIR}\E\/?//;    # legacy where absolute path is specified
     $dir = join('/', $bmwqemu::vars{CASEDIR}, $dir);    # always load from casedir
     die "'$dir' does not exist!\n" unless -d $dir;
-    loadtest($_) for (glob "$dir/*.pm");
+    loadtest($_) for map { s{^\Q$bmwqemu::vars{CASEDIR}/}{}r } (glob "$dir/*.pm");
 }
 
 # This is called if the framework loaded a VM snapshot. All consoles
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -450,13 +450,10 @@ subtest make_snapshot => sub {
 };
 
 subtest loadtestdir => sub {
-    use lib 't/data/tests';
     $bmwqemu::vars{CASEDIR} = 't/data/tests';
     stderr_like {
-        like warning {
-            autotest::loadtestdir('tests');
-        }, qr/ARRAY/, 'script found';
-    } qr/scheduling/, 'loadscript success';
+        autotest::loadtestdir('tests');
+    } qr/debug.*scheduling/, 'loadtestdir is scheduling successfully';
     ok exists $autotest::tests{'tests-boot'}, 'boot.pm loaded';
 };
 


autotest.pm
```diff
@@ -463,7 +463,7 @@ sub loadtestdir ($dir) {
     $dir =~ s/^\Q$bmwqemu::vars{CASEDIR}\E\/?//;    # legacy where absolute path is specified
     $dir = join('/', $bmwqemu::vars{CASEDIR}, $dir);    # always load from casedir
     die "'$dir' does not exist!\n" unless -d $dir;
-    loadtest($_) for (glob "$dir/*.pm");
+    loadtest($_) for map { s{^\Q$bmwqemu::vars{CASEDIR}/}{}r } (glob "$dir/*.pm");
```
okurz, line 466: So two lines above dir is added and here you remove it again?

okurz (approved):

autotest.pm
perlpunk: Your new change cannot work. @okurz was not right.
`glob` needs to get the full path since we cannot rely that the current working directory is that directory.

Tests are failing, and I wonder if you thought about updating the tests according to the acceptance criteria?
The test is still doing a `use lib`, which it shouldn't have to do, and the test checks that a warning is created, while the fix should ensure that there is no warning anymore.

r-richardson: Maybe if you create the relative path after "globbing" ? Something like
```
$dir = join('/', $bmwqemu::vars{CASEDIR}, $dir);    # always load from casedir
die "'$dir' does not exist!\n" unless -d $dir;
for my $testfile (glob "$dir/*.pm") {
    (my $relfile = $testfile) =~ s{^\Q$casedir/}{};
    loadtest($relfile);
}
```

perlpunk: That for loop is basically the same as `loadtest($_) for map { s{^\Q$bmwqemu::vars{CASEDIR}/}{}r } (glob "$dir/*.pm");`

perlpunk:
basically what we need is the following to work:
```perl
    $bmwqemu::vars{CASEDIR} = '/some/path';
    autotest::loadtestdir('tests'); # given that CASEDIR has a subdirectory 'tests' with *.pm files in it
```

autotest.pm
r-richardson: ah i guess you're right, sorry

okurz (approved):

t/08-autotest.t
```diff
@@ -450,7 +450,6 @@ subtest make_snapshot => sub {
 };
 
 subtest loadtestdir => sub {
-    use lib 't/data/tests';
     $bmwqemu::vars{CASEDIR} = 't/data/tests';
     stderr_like {
         like warning {
```
perlpunk, line 455: Do you know what this test does?
It calls the `warning` function from Test::Warnings, which returns an arrayref if there is more than one warning, and if you compare that to a string, it will produce `ARRAY(0x...)`, which the person writing this test thought is a good check.
But it only checks if there is a warning, not even the content of the warning.

And the ticket says that the new code should not produce a warning anymore. You should modify the test accordingly.
Just run it without the `like warning { }` and see what output you get.

perlpunk (changes requested):

perlpunk (approved):

b10n1k (approved):

PR #2626, opened 2025-01-21
Add more code coverage for video_stream.pm

gauravpathak:


b10n1k:
`request_screen_update` looks still uncovered. whats the plan here?

gauravpathak:
> `request_screen_update` looks still uncovered. whats the plan here?

It is still WIP

consoles/video_stream.pm
```diff
@@ -218,7 +218,7 @@ sub _receive_frame_ffmpeg ($self) {
 
     # support P6 only
     if (!($header =~ m/^(P6\n(\d+) (\d+)\n(\d+)\n)/)) {
-        die "Invalid PPM header: $header";
+        die "Invalid PPM header: $header";    # uncoverable statement
```
Martchus, line 221: I think we are able to cover this by mocking the startup of ffmpeg and having it read some garbage data instead.

gauravpathak: Ok, let me try to mock ffmpeg startup

t/26-video_stream.t
```diff
@@ -291,6 +298,23 @@ subtest 'v4l2 resolution' => sub {
...
+
+    @v4l2_ctl_calls = ();
+
+    # changed resolution
+    %v4l2_ctl_results = (
+        '--query-dv-timings' => '0',
+        '--get-dv-timings' => '1024x768p60',
+        '--set-dv-bt-timings query' => 'BT timings set',
+    );
+    $console->{dv_timings_last_check} = time - 4;
```
b10n1k, line 310: I think calling `connect_remote` should covert this. `connect_remote` is called in other subtask. Can it also invoked here or is there any reason is not used?

consoles/video_stream.pm
gauravpathak: I managed to make it die without mocking ffmpeg. I just supplied a png file instead of ppm file as another extra test

t/26-video_stream.t
gauravpathak: Hi @b10n1k 
I think `connect_remote` cannot cover this line: https://app.codecov.io/gh/os-autoinst/os-autoinst/blob/master/consoles%2Fvideo_stream.pm#L389
The above addition is covering this.

gauravpathak: Let me update the comment to make it a bit more clear

consoles/video_stream.pm
Martchus: Even better :-)

okurz (changes requested):
Regarding the commit message: Those are not "codecov unit tests". codecov is only a web service we use to visualize the coverage.

Martchus (changes requested):

t/26-video_stream.t
gauravpathak: Basically, the above line is just used to make the condition true written here https://github.com/os-autoinst/os-autoinst/blob/master/consoles/video_stream.pm#L380
Maybe we can also set `$console->{dv_timings_last_check}` to a known constant value such as:
`$console->{dv_timings_last_check} = $console->DV_TIMINGS_CHECK_INTERVAL;` to make the `if` condition true.

t/26-video_stream.t
```diff
@@ -106,6 +106,13 @@ subtest 'connect stream' => sub {
     is_deeply $cmd, [
         'ffmpeg', '-loglevel', 'fatal', '-i', '/dev/video0',
         '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 3, '-'], "correct cmd built for fps=3";
+
+    $mock_console->redefine(update_framebuffer => sub ($self) { $self->{_last_update_received} = 200; return 0; });
+    $console->request_screen_update();
+    is $console->{dv_timings_supported}, 1, "dv timings are supported";
+    ok($console->{dv_timings_last_check} >= time, "dv_timings_last_check ok");
```
okurz, line 113:
```suggestion
    is $console->{dv_timings_supported}, 1, 'dv timings are supported';
    ok $console->{dv_timings_last_check} >= time, 'dv_timings_last_check ok';
```

t/26-video_stream.t
```diff
@@ -166,6 +173,14 @@ subtest 'frames parsing' => sub {
     my $received_update = $console->update_framebuffer();
     is $received_update, 0, "detected incomplete frame";
     $console->disable_video;
+
+    # now invalid frame - not in PPM format to test exception
+    $mock_video_source = $data_dir . "accept-ssh-host-key.png";
```
okurz, line 178:
```suggestion
    $mock_video_source = $data_dir . 'accept-ssh-host-key.png';
```

okurz (changes requested):

okurz (approved):

t/26-video_stream.t
```diff
@@ -266,7 +281,7 @@ subtest 'v4l2 resolution' => sub {
     %v4l2_ctl_results = (
         '--query-dv-timings' => '640x480p60',
     );
-    $console->{dv_timings_last_check} = time - 4;
+    $console->{dv_timings_last_check} = $console->DV_TIMINGS_CHECK_INTERVAL;
```
Martchus, line 284: I would avoid invoking constants like this. It works but is not really how they are intended to be used.

```suggestion
    $console->{dv_timings_last_check} = consoles::video_stream::DV_TIMINGS_CHECK_INTERVAL;
```

The same counts for the other occurrence.

t/26-video_stream.t
```diff
@@ -106,6 +106,13 @@ subtest 'connect stream' => sub {
     is_deeply $cmd, [
         'ffmpeg', '-loglevel', 'fatal', '-i', '/dev/video0',
         '-vcodec', 'ppm', '-f', 'rawvideo', '-r', 3, '-'], "correct cmd built for fps=3";
+
+    $mock_console->redefine(update_framebuffer => sub ($self) { $self->{_last_update_received} = 200; return 0; });
+    $console->request_screen_update();
+    is $console->{dv_timings_supported}, 1, 'dv timings are supported';
+    ok $console->{dv_timings_last_check} >= time, 'dv_timings_last_check ok';
```
Martchus, line 113: There's https://perldoc.perl.org/Test::More#cmp_ok which gives a better error message in case the check fails.

The use of `>=` makes only sense when recording `time` before running the test, e.g. put `my $start_time = time;` before `$console->request_screen_update();` and compare against that. Otherwise, `time` might still be ahead of `dv_timings_last_check` if there's a big enough delay.

```suggestion
    ok $console->{dv_timings_last_check} >= time, 'dv_timings_last_check ok';
```

Martchus (changes requested):

okurz (approved):

t/26-video_stream.t
Martchus: I now see that my suggestion isn't changing anything. I guess I clicked on the button to add a suggestion but forgot to remove it after deciding not to go into concrete coding.

Can you just implement what I have written in text?

gauravpathak: I made the changes earlier itself https://github.com/os-autoinst/os-autoinst/pull/2626/files#diff-1320e7b8296971afbba67346b76e74f7b0cfbc5613a102f10406b36c9855beefR113

It's not visible here.

Martchus: Ah, right. But what I wrote in the 2nd paragraph hasn't been implemented yet. And if you think it doesn't make any sense what I wrote you can of course also say that :-)

gauravpathak: I have updated the code

perlpunk (approved):

Martchus (approved):

PR #2627, opened 2025-01-21
Call SYNC_ASSETS_HOOK before downloading fresh needles too

marmarek:
This is useful to fetch new needles that might be created via developer
mode.
See SYNC_ASSETS_HOOK documentation in the OpenQA repo.

This PR extends https://github.com/os-autoinst/openQA/pull/6131

### DIFF
--- a/OpenQA/Isotovideo/NeedleDownloader.pm
+++ b/OpenQA/Isotovideo/NeedleDownloader.pm
@@ -124,6 +124,11 @@ sub download ($self) { $self->_download_file($_) for (@{$self->files_to_download
 # downloads missing needles considering $new_needles
 # (see t/21-needle-downloader.t for an example of $new_needles)
 sub download_missing_needles ($self, $new_needles) {
+    if ($new_needles && $bmwqemu::vars{SYNC_ASSETS_HOOK}) {
+        bmwqemu::diag("Running SYNC_ASSETS_HOOK");
+        my $ret = system($bmwqemu::vars{SYNC_ASSETS_HOOK});
+        return if ($ret == (32 << 8));
+    }
     $self->add_relevant_downloads($new_needles);
     $self->download();
 }
--- a/t/21-needle-downloader.t
+++ b/t/21-needle-downloader.t
@@ -149,6 +149,40 @@ subtest '_download_file' => sub {
     like $stderr, qr{internal error occurred when downloading.*oops}, 'internal error was logged';
 };
 
+subtest 'sync assets hook used at the right time' => sub {
+    $downloader = OpenQA::Isotovideo::NeedleDownloader->new;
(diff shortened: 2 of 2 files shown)


Martchus (commented):
> This is useful to fetch new needles that might be created via developer
> mode.

The downloader (where you add this hook) is actually supposed to do exactly that. I assume this isn't working for you, though - which raises the question why? How did you pause the test? The needle syncing is unfortunately not done in all cases but it would perhaps make sense to change that.

marmarek:
In this case `/var/lib/openqa/share` is a network share mounted read-only, so worker can't write there directly.
See https://github.com/os-autoinst/openQA/pull/6131/files#diff-71734bf4d91118f92743c7d0db9218c4f24729410e020110884606b5ad6a3878

Martchus:
I suppose it works in our case because when using the cache service one still has a writable needle directory. Of course it would be possible to let the downloader write the needles somewhere else. I'll have to look into the code again tomorrow.

marmarek:
Hi, did you have time to look at it?

OpenQA/Isotovideo/NeedleDownloader.pm
```diff
@@ -124,6 +124,10 @@ sub download ($self) { $self->_download_file($_) for (@{$self->files_to_download
 # downloads missing needles considering $new_needles
 # (see t/21-needle-downloader.t for an example of $new_needles)
 sub download_missing_needles ($self, $new_needles) {
+    if ($new_needles && $bmwqemu::vars{SYNC_ASSETS_HOOK}) {
+        bmwqemu::diag("Running SYNC_ASSETS_HOOK");
+        system($bmwqemu::vars{SYNC_ASSETS_HOOK});
+    }
```
Martchus, line 130: If a hook script is specified the "normal" needle download should probably be skipped as it would not work anyway. Maybe the skipping could depend on the exit code of the hook script.

When skipping the download you could maybe also avoid the moving via `Test::MockObject::Extends` in the test code.

Martchus (commented):
No, but good that you remind me.

So I had a look now. It is true that the needle downloader just downloads into the needle directory which will indeed not work in your case (because it is read-only).

I guess we can merge this. I have only one remark (see the inline comment).

OpenQA/Isotovideo/NeedleDownloader.pm
marmarek: I've done it by some non-default exit code (32) as needles download might still be useful for example with alternative needles dir (openqa-clone-custom-git-refspec?) or when just "factory" dir (but not "tests") is covered by this sync hook.

okurz (approved):

Martchus (approved):

PR #2629, opened 2025-01-27
Avoid potential endless loop in case of SSH read timeouts in `check_ssh_serial`

Martchus:
The `read2` call might time out (the default timeout is 5 minutes) or otherwise fail. So far this leads to an endless loop until `eof` is reached which might not happen in a timely manner, e.g. because the command being executed is stuck. Then the whole backend would get stuck and the test might run into `MAX_JOB_TIME`, see
https://progress.opensuse.org/issues/176076.

With this change the timeout on SSH operations is effective in case the invoked command does not produce any output as tests would now die with `Time out waiting for data (-9 LIBSSH2_ERROR_TIMEOUT)` when a command does not return any new output for the timeout.

Since the default timeout is only 5 minutes this might disrupt long operations such as copying big files over network if those operations don't produce any output while running. ~~So maybe it should be possible to set the timeout on command-level (not included by this change).~~ So this PR adds an option to specify the timeout (see 2nd commit for details).

This change also provokes more immediate errors in other cases where `read2` fails.

---

I tested this with by cloning https://openqa.suse.de/tests/16551116 locally and just put an unconditional `die` where this change might cause a `die`. The result looks like this:

grafik

### DIFF
--- a/OpenQA/Isotovideo/Interface.pm
+++ b/OpenQA/Isotovideo/Interface.pm
@@ -9,7 +9,7 @@ use Mojo::Base -strict, -signatures;
 # -> increment on every change of such APIs
 # -> never move that variable to another place (when refactoring)
 #    because it may be accessed by the tests itself
-our $version = 42;
+our $version = 43;
 
 # major version of the (web socket) API relevant to the developer mode
 # -> increment when making non-backward compatible changes to that API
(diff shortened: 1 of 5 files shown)


okurz (changes requested):

Martchus:
I added a unit test and I also added a `timeout` parameter because now that the timeout is actually effective it should be easy to change it for individual commands. This is of course still a breaking change.

r-richardson:
LGTM

okurz (commented):
Please bump the version number in OpenQA::Isotovideo::Interface

Martchus:
done

perlpunk (approved):

okurz (approved):

PR #2634, opened 2025-01-30
Remove time comparison with dv_timings_last_check

gauravpathak:
Related: https://progress.opensuse.org/issues/174253
- Do not compare values returned from `time` call of different function calls to avoid test comparison runtime failure


okurz (approved):

Martchus (approved):
Alternatively you could try to mock time, there's a Perl module for that.

gauravpathak:
> Alternatively you could try to mock time, there's a Perl module for that.

I will try it in a new Pull Request

PR #2640, opened 2025-01-31
Make it obvious when autotest receives SIGTERM in any case

Martchus:
So far we silently exit when receiving SIGTERM when no test module is running (e.g. autotest is between modules). Then it is not obvious why the test terminated prematurely. With this change we always get the relevant log line.

Related ticket: https://progress.opensuse.org/issues/174259

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -283,8 +283,9 @@ sub run_all () {
 }
 
 sub handle_sigterm ($sig) {    # uncoverable statement
+    my $info = $current_test ? ', saving results of current test before exiting' : '';    # uncoverable statement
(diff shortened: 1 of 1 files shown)


baierjan (approved):

okurz (changes requested):
I like the intention but I see many people deeply confused by the overly verbose output in autoinst-log.txt, i.e. they are unable to find relevant error messages because there are so many lines even for "clean shutdowns". How about concatenating the strings and outputting the final string after the if? Then it would be post-save but maybe the saving process isn't that critical that we would definitely need a message before.

okurz (approved):

PR #2641, opened 2025-01-31
Fix race condition when autotest terminates

Martchus:
When autotest terminates because all tests are done it sends the message `tests_done` to isotovideo to indicate a normal exit. If isotovideo does not get this message it concludes autotest has exited prematurely and considers the test incomplete.

This makes generally sense. However, autotest exists immediately after sending the `tests_done` message. Then isotovideo handles that exit by setting its loop condition to false (like the it does when any of its persistent child processes exists). This means it is not guaranteed that the I/O loop of isotovideo still reads the `tests_done`. In case it does not the test would be wrongfully considered incomplete.

This is happening from time to time in production openQA-in-openQA tests, see https://progress.opensuse.org/issues/174259#note-24.

This change should fix the issue by letting autotest wait until isotovideo responds to the `tests_done` message. It is actually normal for autotest to wait for isotovideo to respond as this happens on every query_isotovideo invocation. So it shouldn't cause any problems to use the same approach also to communicate termination.

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -291,6 +291,7 @@ sub _handle_command_set_current_test ($self, $response, @) {
 sub _handle_command_tests_done ($self, $response, @) {
     $self->test_died($response->{died});
     $self->test_completed($response->{completed});
+    $self->_respond_ok;
     $self->emit(tests_done => $response);
     $self->current_test_name('');
     $self->status('finished');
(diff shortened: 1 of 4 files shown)


okurz (commented):
+1

Martchus:
I tested this locally and normal test still pass. I also tested cancellation and that still worked as expected with no weird log messages showing up.

okurz (approved):

perlpunk (approved):

PR #2644, opened 2025-02-05
WIP Gather JSON coverage data for each individual test

perlpunk:
Proof of concept so far.
Will link ticket...

### DIFF
--- a/tools/container_run_ci
+++ b/tools/container_run_ci
@@ -71,7 +71,10 @@ fi
 cd /opt
 ./tools/install-new-deps.sh
 export CI=1 WITH_COVER_OPTIONS=1 CHECK_GIT_STATUS=1 PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS=$PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B \"$build_dir\"
-cmake --build \"$build_dir\" --verbose --target check-pkg-build
-cmake --build \"$build_dir\" --verbose --target $target
-cover -report html_minimal \"$build_dir\"/cover_db"
+make
+./tools/cover-by-test.sh"
+
+#cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B \"$build_dir\"
+#cmake --build \"$build_dir\" --verbose --target check-pkg-build
+#cmake --build \"$build_dir\" --verbose --target $target
+#cover -report html_minimal \"$build_dir\"/cover_db
--- a/tools/cover-by-test.pl
+++ b/tools/cover-by-test.pl
@@ -1,0 +1,50 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use v5.24;
+use experimental qw(signatures);
+use JSON::PP qw(decode_json encode_json);
+use autodie;
+use File::Path qw(make_path);
+use File::Basename qw(dirname);
+use Data::Dumper;
+
+my ($test, $report, $dir) = @ARGV;
+
+my $coverage = read_json($report);
+
+my $by_file = $coverage->{coverage};
+
+for my $key (sort keys %$by_file) {
+    say "=== $key";
+    my $lines = $by_file->{$key};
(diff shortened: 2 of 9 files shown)


tools/container_run_ci
```diff
@@ -71,7 +71,10 @@ fi
 cd /opt
 ./tools/install-new-deps.sh
 export CI=1 WITH_COVER_OPTIONS=1 CHECK_GIT_STATUS=1 PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS=$PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B \"$build_dir\"
-cmake --build \"$build_dir\" --verbose --target check-pkg-build
-cmake --build \"$build_dir\" --verbose --target $target
-cover -report html_minimal \"$build_dir\"/cover_db"
+make
```
perlpunk, line 74: I couldn't get this running with cmake, as this already runs all tests with `prove`, but for this we need to run the shell script instead.
So we need a way to run cmake but without running tests, just building

tools/cover-by-test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -x
+
+for testfile in $(find t -name "*.t"); do
```
gauravpathak, line 5: I am not sure if we follow shell check for our shell scripts, but using find like this can be problematic if find encounters whitespace or glob patterns in filename, if we do not allow whitespace or special characters in file names then it is fine
Refer: https://www.shellcheck.net/wiki/SC2044

Martchus: I guess it doesn't matter here but I'd also like it this wouldn't break on white-spaces.

tools/cover-by-test.pl
```diff
@@ -0,0 +1,50 @@
...
+    my $json = do { local $/; <$fh> };
+    close $fh;
+    return decode_json $json;
+}
+sub write_json ($file, $data) {
+    make_path dirname $file;
+    open my $fh, '>', $file;
+    print $fh encode_json $data;
+    close $fh;
+}
```
Martchus, line 49: It would make sense to use Mojo::File here to turn these into one-liners.

tools/container_run_ci
```diff
@@ -71,7 +71,10 @@ fi
 cd /opt
 ./tools/install-new-deps.sh
 export CI=1 WITH_COVER_OPTIONS=1 CHECK_GIT_STATUS=1 PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS=$PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS
-cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -S . -B \"$build_dir\"
```
Martchus, line 74: If we just use `make` here we should remove the `build_dir` variable at the top. Not sure about the `target` variable. We might still need it.

Note that your newly introduce script could also have its own CMake target so we wouldn't need to change this code (and could just invoke this target like others).

tools/cover-by-test.sh
baierjan, line 5: The easiest way around in this case is probably:
```suggestion
shopt -s globstar nullglob
for testfile in t/**/*.t; do
```

tools/container_run_ci
```diff
@@ -71,7 +71,10 @@ fi
...
-cmake --build \"$build_dir\" --verbose --target check-pkg-build
-cmake --build \"$build_dir\" --verbose --target $target
-cover -report html_minimal \"$build_dir\"/cover_db"
+make
+./tools/cover-by-test.sh"
```
Wabri, line 75:
```suggestion
./tools/cover-by-test.sh
```

tools/cover-by-test.sh
Wabri: Maybe something like this could also work in `sh`:

```
find t -name "*.t" -print0 | while IFS= read -r -d '' testfile
do
    echo "===== $testfile"
    rm -rf cover_db
    ./tools/invoke-tests --coverage "$testfile"
    cover -report codecovbash
    perl tools/cover-by-test.pl "$testfile" cover_db/codecov.json cover-by-test
done
```

PR #2646, opened 2025-02-05
[IGNORE] test power off output in CI

b10n1k:
https://progress.opensuse.org/issues/176319


t/data/tests/main.pm
```diff
@@ -39,11 +39,11 @@ autotest::loadtest "tests/assert_screen.pm";
...
-    autotest::loadtest "tests/assert_screen_fail_test.pm";
-    autotest::loadtest "tests/reload_needles.pm";
-    autotest::loadtest "tests/modify_and_upload_file.pm";
+    #autotest::loadtest "tests/select_console_fail_test.pm";
+    #autotest::loadtest "tests/select_ssh_console_fail_test.pm";
+    #autotest::loadtest "tests/assert_screen_fail_test.pm";
+    #autotest::loadtest "tests/reload_needles.pm";
+    #autotest::loadtest "tests/modify_and_upload_file.pm";
 }
 autotest::loadtest "tests/shutdown.pm";
```
okurz, line 48: It should suffice to run with `INTEGRATION_TEST=1` to skip the whole block

Wabri: We got a success on the @b10n1k local machine with this tests removed, we are trying if is related to this changes or something else.

PR #2647, opened 2025-02-05
Remove Test::Fatal dependency

perlpunk:
We switched everything to Test::Exception (Test::Most) in #2643


Martchus (approved):

okurz (approved):

okurz:
Good. I plan to do my previous change and also the removal of that dependency for openQA as well

PR #2649, opened 2025-02-06
Enable is_shutdown to be used after "power('off')" on qemu

okurz:
"power('off')" for the qemu backend sends the QMP command "quit" which
causes the whole qemu stack to shutdown closing the QMP pipe from the
remote end which previously triggered a "Broken pipe" exception whenever
"is_shutdown" is called after the power command. This commit explicitly
handles this situation by returning the "shutdown" status correctly.
Other combinations of testapi commands if used in undefined operations
might still fail

Tested manually using t/99-full-stack.t by using "power('off')" instead
of "sudo powerdown" in t/data/tests/tests/shutdown.pm

Related progress issue: https://progress.opensuse.org/issues/176319

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -41,6 +41,7 @@ sub new ($class) {
     $self->{pidfilename} = 'qemu.pid';
     $self->{proc} = OpenQA::Qemu::Proc->new();
     $self->{proc}->_process->pidfile($self->{pidfilename});
+    $self->{expected_shutdown} = 0;
     return $self;
 }
 
@@ -74,6 +75,11 @@ sub power ($self, $args) {
         reset => 'system_reset',
         off => 'quit',
     );
+    if ($args->{action} eq 'off') {
+        $self->{expected_shutdown} = 1;
+        $self->disable_consoles();
+    }
+    bmwqemu::diag("POWER: action: $args->{action}, expected_shutdown: $self->{expected_shutdown}");
     $self->handle_qmp_command({execute => $action_to_cmd{$args->{action}}});
 }
 
@@ -1120,8 +1126,18 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
         bmwqemu::fctinfo("Skipping the following qmp_command because QEMU_ONLY_EXEC is enabled:\n$line");
         return undef;
     }
-    my $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line);
-    die "handle_qmp_command: syswrite failed $!" unless ($wb == length($line));
+    local $SIG{__DIE__} = undef;    # prevent custom die handler so that a broken pipe can be handled
+    my $wb;
+    try { $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line) }
+    catch ($e) {
+        if ($line =~ qr/query-status/ && $self->{expected_shutdown} && $e =~ qr/(Broken pipe|Bad file descriptor)/) {
+            return {return => {status => 'shutdown'}};
+        }
+        # these two lines are actually covered via t/99-full-stack-stop.t but the coverage doesn't detect it
+        bmwqemu::fctwarn("The following qmp_command failed because qemu was explicitly stopped from test code via power('off'):\n$line") if $self->{expected_shutdown}; # uncoverable statement
+        die "handle_qmp_command: Unexpected error: $e";    # uncoverable statement
+    }
+    die "handle_qmp_command: syswrite failed $!" unless $wb && $wb == length($line);
 
     my $hash;
     do {
@@ -1157,7 +1173,8 @@ sub read_qemupipe ($self) {
     return $bytes;
 }
 
-sub close_pipes ($self) {
+sub close_pipes ($self, $closeall = 0) {
+    # if closeall is 1, close not only pipes to qemu but also jsonrpc connection to autotest
     $self->do_stop_vm() if $self->{started};
 
     if (my $qemu_pipe = $self->{qemupipe}) {
@@ -1175,7 +1192,7 @@ sub close_pipes ($self) {
         $self->{qmpsocket} = undef;
     }
 
-    $self->SUPER::close_pipes() unless exists $self->{stop_only_qemu} && $self->{stop_only_qemu};
+    $self->SUPER::close_pipes() if $closeall || !((exists $self->{stop_only_qemu} && $self->{stop_only_qemu}) || $self->{expected_shutdown});
 }
 
 sub is_shutdown ($self, @) {
--- a/t/99-full-stack-stop.t
+++ b/t/99-full-stack-stop.t
@@ -1,0 +1,62 @@
+#!/usr/bin/perl
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+
+use Test::Most;
+use Mojo::Base -strict, -signatures;
+
+use FindBin '$Bin';
+use lib "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '450';
+use Test::Warnings ':report_warnings';
+use File::Basename;
+use Cwd 'abs_path';
+use Mojo::JSON 'decode_json';
+use Mojo::File qw(path tempdir);
+use Mojo::Util qw(scope_guard);
+
+my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
+my $toplevel_dir = "$Bin/..";
+my $data_dir = "$Bin/data/";
+my $pool_dir = "$dir/pool/";
+mkdir $pool_dir;
+
+note("data dir: $data_dir");
+note("pool dir: $pool_dir");
+
+chdir($pool_dir);
+my $cleanup = scope_guard sub { chdir $Bin; undef $dir };
+
+my $casedir = path($data_dir, 'tests');
+path('vars.json')->spew(<<EOV);
+{
+   "ARCH" : "i386",
+   "BACKEND" : "qemu",
+   "QEMU" : "i386",
+   "QEMU_NO_TABLET" : "1",
+   "QEMU_NO_FDC_SET" : "1",
+   "CASEDIR" : "$casedir",
+   "ISO" : "$data_dir/Core-7.2.iso",
+   "CDMODEL" : "ide-cd",
+   "HDDMODEL" : "ide-hd",
+   "VERSION" : "1",
+   "SSH_CONNECT_RETRY"  : "2",
+   "SSH_CONNECT_RETRY_INTERVAL"  : ".001",
+   "NAME" : "00001-1-i386@32bit",
+   "INTEGRATION_TESTS" : "1",
+   "CMD_AFTER_STOP" : "1",
+}
+EOV
+# create screenshots
+path('live_log')->touch;
+system("cd $toplevel_dir && perl $toplevel_dir/isotovideo --workdir $pool_dir -d 2>&1 | tee $pool_dir/autoinst-log.txt");
+my $log = path('autoinst-log.txt')->slurp;
+my $version = -e "$toplevel_dir/.git" ? qr/[a-f0-9]+/ : 'UNKNOWN';
+like $log, qr/Current version is $version [interface v[0-9]+]/, 'version read from git';
+like $log, qr/\d*: EXIT 1/, 'test execution failed';
+unlike $log, qr/Tests died:/, 'Tests did not fail within modules' or diag "autoinst-log.txt: $log";
+unlike $log, qr/warn.*qemu-system.*terminating/, 'No warning about expected termination';
+like $log, qr/qemu was explicitly stopped from test code.*system_reset/s, 'Warning about QMP cmd after qemu stopped';
+
+done_testing();
--- a/t/data/tests/tests/shutdown.pm
+++ b/t/data/tests/tests/shutdown.pm
@@ -5,8 +5,9 @@ use Mojo::Base 'basetest', -signatures;
 use testapi;
 
 sub run ($) {
-    enter_cmd 'sudo poweroff';
+    power 'off';
     assert_shutdown(get_var('INTEGRATION_TESTS') ? 90 : undef);
+    power 'reset' if get_var('CMD_AFTER_STOP');
 }
 
 sub test_flags ($) { {fatal => 1} }
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -453,7 +453,7 @@ sub stop_vm ($self, @) {
     }
     $self->_stop_video_encoder();
     $self->close_ssh_connections();
-    $self->close_pipes();    # does not return
+    $self->close_pipes(1);    # does not return
     return;
 }
 
@@ -588,7 +588,7 @@ sub wait_still_screen ($self, $args) {
     return {postponed => 1};
 }
 
-sub close_pipes ($self) {
+sub close_pipes ($self, $closeall = 0) {
     if ($self->{cmdpipe}) {
         close($self->{cmdpipe}) || die "close $!\n";
         $self->{cmdpipe} = undef;


perlpunk (changes requested):

backend/qemu.pm
```diff
@@ -1127,8 +1127,10 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
         bmwqemu::fctinfo("Skipping the following qmp_command because QEMU_ONLY_EXEC is enabled:\n$line");
         return undef;
     }
-    my $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line);
-    die "handle_qmp_command: syswrite failed $!" unless ($wb == length($line));
+    $SIG{__DIE__} = undef;    # prevent custom die handler so that a broken pipe can be handled
```
Martchus, line 1130: Don't we need to restore that handler again after leaving this function? This could be done using https://docs.mojolicious.org/Mojo/Util#scope_guard. Maybe it would also work to use `local $SIG{__DIE__} = undef;`.

backend/qemu.pm
```diff
@@ -1127,8 +1127,10 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
...
+    my $wb = eval { defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line) };
+    return {return => {status => 'shutdown'}} if $@ && $@ =~ qr/Broken pipe/;
```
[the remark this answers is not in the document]
okurz: I guess we should rethrow the exception then?

backend/qemu.pm
perlpunk: yeah, i was also going to comment on that line

backend/qemu.pm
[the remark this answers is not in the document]
perlpunk: yes

perlpunk (approved):

kalikiana (changes requested):
> Tested manually using t/99-full-stack.t by using "power('off')" instead of "sudo powerdown" in t/data/tests/tests/shutdown.pm

I don't see this change here and would feel more confident reviewing this if it was covered.

okurz:
> > Tested manually using t/99-full-stack.t by using "power('off')" instead of "sudo powerdown" in t/data/tests/tests/shutdown.pm
> 
> I don't see this change here and would feel more confident reviewing this if it was covered.

You mean to change the code to actually use `power('off')` in the full stack test and not type anymore? I will try

backend/qemu.pm
```diff
@@ -1127,8 +1127,11 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
         bmwqemu::fctinfo("Skipping the following qmp_command because QEMU_ONLY_EXEC is enabled:\n$line");
         return undef;
     }
-    my $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line);
-    die "handle_qmp_command: syswrite failed $!" unless ($wb == length($line));
+    local $SIG{__DIE__} = undef;    # prevent custom die handler so that a broken pipe can be handled
+    my $wb = eval { defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line) };
+    return {return => {status => 'shutdown'}} if $@ && $@ =~ qr/Broken pipe/;
```
Martchus, line 1132: The only problem I see is that encountering a broken pipe does not necessarily mean an expected shutdown happened. So if we call `handle_qml_command` to invoke some random QMP command like we do in many places we will continue the execution as if nothing bad happened in case the pipe broke. This would lead to wrong and maybe not so easy to debug behavior later on.

It would make more sense to me if `power 'off'` would set some flag somewhere and only if this flag is set we would consider broken pipes to be expected and only then we'd ever enter this special error handling branch.

I would also save `$@` and `$!` in variables because it is always a source for error when those are used not directly after the function call that's checked.

Martchus (changes requested):

backend/qemu.pm
```diff
@@ -1127,8 +1127,11 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
...
+    die "handle_qmp_command: Unexpected error: $@" if $@;
+    die "handle_qmp_command: syswrite failed $!" unless $wb && $wb == length($line);
```
perlpunk, line 1134: > I would also save `$@` and `$!` in variables

This is how it could look like with Feature::Compat::Try:
```suggestion
    use Feature::Compat::Try;
    my $wb = do {
        try { defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line) }
        catch ($e) { return {return => {status => 'shutdown'}} if $e =~ qr/Broken pipe/; die $e }
    };
    die "handle_qmp_command: syswrite failed $!" unless $wb && $wb == length($line);
```

okurz: looks nice. I am preparing some changes to bring in Feature::Compat::Try and get rid of some older suboptimal code. I plan to continue here afterwards so that we have a good style for catching exception going forward before I introduce more as necessary for this PR.

okurz: ->
* #2650

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -119,6 +119,7 @@ sub _send_to_cmd_srv ($self, $data) { myjsonrpc::send_json($self->cmd_srv_fd, $d
 sub _send_to_backend ($self, $data) { myjsonrpc::send_json($self->backend_fd, $data) }
 
 sub send_to_backend_requester ($self, $data) {
+    return undef unless $self->backend_requester;
     myjsonrpc::send_json($self->backend_requester, $data);
     $self->backend_requester(undef);
```
b10n1k, line 124: this looks like it could be just this

```
myjsonrpc::send_json($self->backend_requester, $data)  unless $self->backend_requester;
$self->backend_requester(undef);
```

okurz: I guess you mean `myjsonrpc::send_json($self->backend_requester, $data) if $self->backend_requester`. But then we don't need to set it to undef when it's already undef so I think my proposal is cleaner

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -119,6 +119,7 @@ sub _send_to_cmd_srv ($self, $data) { myjsonrpc::send_json($self->cmd_srv_fd, $d
 sub _send_to_backend ($self, $data) { myjsonrpc::send_json($self->backend_fd, $data) }
 
 sub send_to_backend_requester ($self, $data) {
+    return undef unless $self->backend_requester;
     myjsonrpc::send_json($self->backend_requester, $data);
```
Martchus, line 123:
```suggestion
    return undef unless my $requester = $self->backend_requester;
    myjsonrpc::send_json($requester, $data);
```

okurz: Do I really need a local variable just to access an attribute within the current object?

Martchus: This is faster than accessing it twice (invoking the function that does the attribute lookup twice; there's no inlining) and no additional lines of code are required. So it is in my opinion a win (even if not a big one).

OpenQA/Isotovideo/CommandHandler.pm
b10n1k: yes the correct syntax is with `if`.
not sure I understatnd the logic. `myjsonrpc::send_json($requester, $data);` should be executed only if `backend_requester` is not _undef_. Correct me but `myjsonrpc::send_json($self->backend_requester, $data) if $self->backend_requester` would return _undef_ if condition is false anyway. likely you wont even need the `$self->backend_requester(undef);`. Even sure I am not sure about that.

Martchus:
Just to recap what's still missing: The case when autotest takes long enough to terminate to still receive SIGTERM from isotovideo is not handled correctly. So far this change would lead to an incomplete test in this case (unless @okurz has already fixed this since our last session). Maybe autotest should simply avoid sending SIGTERM if it passed a power off command. Otherwise autotest could consider the SIGTERM "expected" in case it previously send a power off command.

okurz:
Thanks for reminding about the open points. I did not have the opportunity to work on this since our last session

asdil12:
Updated the PR:
- Don't let the qemu backend close the JSONRPC socket when closing the QMP socket (because qemu exited)
-  Still close the JSONRPC socket when triggered by the backend baseclass
- Close the consoles on poweroff
- Simulate `shutdown` status on `query-status` qmp cmd if qemu is not running anymore

t/data/tests/tests/shutdown.pm
```diff
@@ -5,8 +5,13 @@ use Mojo::Base 'basetest', -signatures;
 use testapi;
 
 sub run ($) {
-    enter_cmd 'sudo poweroff';
+    power 'off';
     assert_shutdown(get_var('INTEGRATION_TESTS') ? 90 : undef);
+    sleep 8; # TODO this causes autotest to take a little bit longer causing: [2025-02-17T13:05:51.891245+01:00] [warn] [pid:25301] !!! OpenQA::Isotovideo::Runner::_read_response: THERE IS NOTHING TO READ 17 4 3
```
okurz, line 10: Given the much quicker execution without this I see this as something temporary you can try out to experiment with but not anything acceptable for merging

okurz:
It seems you now changed the code back to where we were before with the TODO still open:
> TODO we need to find a way to give clear feedback to test writers that one must not execute any more commands after power off *and* handle the case when autotest just takes a little bit longer

asdil12:
The feedback is here - I don't know how I could make it any more clear:

https://github.com/os-autoinst/os-autoinst/pull/2649/files#diff-675dc99664e3cb2e63629d86dd587727a186671073760a12af76d2d974a692a0R1136

okurz:
> The feedback is here - I don't know how I could make it any more clear:
> 
> https://github.com/os-autoinst/os-autoinst/pull/2649/files#diff-675dc99664e3cb2e63629d86dd587727a186671073760a12af76d2d974a692a0R1136

I see. That's actually pretty good. Good enough for me :) Please just handle https://github.com/os-autoinst/os-autoinst/pull/2649#discussion_r1991846452 or hand over to me again

backend/qemu.pm
```diff
@@ -74,6 +75,11 @@ sub power ($self, $args) {
         reset => 'system_reset',
         off => 'quit',
     );
+    if ($args->{action} eq 'off') {
+        $self->{expected_shutdown} = 1;
+        $self->disable_consoles();
+    }
+    bmwqemu::fctinfo("POWER: action: " . $args->{action} . ", expected_shutdown: " . $self->{expected_shutdown});
```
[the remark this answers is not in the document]
asdil12: I actually switched it from warning to info with the intention to keep it.

OpenQA/Isotovideo/CommandHandler.pm
Martchus: It would be still nice to change this.

backend/qemu.pm
```diff
@@ -1175,7 +1190,7 @@ sub close_pipes ($self) {
         $self->{qmpsocket} = undef;
     }
 
-    $self->SUPER::close_pipes() unless exists $self->{stop_only_qemu} && $self->{stop_only_qemu};
+    $self->SUPER::close_pipes() unless !$closeall && ((exists $self->{stop_only_qemu} && $self->{stop_only_qemu}) || $self->{expected_shutdown});
```
Martchus, line 1193: This would reduce the number of negations by one:
```suggestion
    $self->SUPER::close_pipes() if $closeall || !((exists $self->{stop_only_qemu} && $self->{stop_only_qemu}) || $self->{expected_shutdown});
```
(I hope I got it right.)

---

Maybe it would also make sense to pass `$closeall` so in case the base class implementation would actually do something with the parameter it would actually work:

```suggestion
    $self->SUPER::close_pipes($closeall) if $closeall || !((exists $self->{stop_only_qemu} && $self->{stop_only_qemu}) || $self->{expected_shutdown});
```

backend/qemu.pm
```diff
@@ -1120,8 +1126,17 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
         bmwqemu::fctinfo("Skipping the following qmp_command because QEMU_ONLY_EXEC is enabled:\n$line");
         return undef;
     }
-    my $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line);
-    die "handle_qmp_command: syswrite failed $!" unless ($wb == length($line));
+    local $SIG{__DIE__} = undef;    # prevent custom die handler so that a broken pipe can be handled
+    my $wb;
+    try { $wb = defined $optargs{send_fd} ? tinycv::send_with_fd($sk, $line, $optargs{send_fd}) : syswrite($sk, $line) }
+    catch ($e) {
+        if ($line =~ qr/query-status/ && $self->{expected_shutdown} && ($e =~ qr/Broken pipe/ || $e =~ qr/Bad file descriptor/)) {
```
Martchus, line 1133:
```suggestion
        if ($line =~ qr/query-status/ && $self->{expected_shutdown} && $e =~ qr/(Broken pipe|Bad file descriptor)/) {
```

backend/qemu.pm
```diff
@@ -1120,8 +1126,17 @@ sub handle_qmp_command ($self, $cmd, %optargs) {
...
+            return {return => {status => 'shutdown'}};
+        }
+        bmwqemu::fctwarn("The following qmp_command failed because qemu was explicitly stopped from test code via power('off'):\n$line") if $self->{expected_shutdown};
+        die "handle_qmp_command: Unexpected error: $e";
```
Martchus, line 1137:
```suggestion
        die "handle_qmp_command: unexpected error: $e";
```

backend/qemu.pm
Martchus, line 82:
```suggestion
    bmwqemu::fctinfo("POWER: action: $args->{action}, expected_shutdown: $self->{expected_shutdown}");
```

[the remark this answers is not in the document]
okurz: This is how it looks

```
# [2025-03-12T17:22:46.274509+01:00] [debug] [pid:20968] <<< testapi::power(action="off")
# POWER: action: off, expected_shutdown: 1 at /home/okurz/local/os-autoinst/os-autoinst/script/../backend/qemu.pm line 79.
# [2025-03-12T17:22:46.277436+01:00] [debug] [pid:20982] EVENT {"data":{"guest":false,"reason":"host-qmp-quit"},"event":"SHUTDOWN","timestamp":{"microseconds":276749,"seconds":1741796566}}
```

so looks a little bit "debuggy". I would prefer `bmwqemu::diag` but ok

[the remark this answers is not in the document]
asdil12: done

okurz (commented):
I don't think it's a good idea to add another complete test run just to test the reset. How long does it take?

Also please create a PR based on your item fork

t/99-full-stack-stop.t
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/perl
+# Copyright 2017-2021 SUSE LLC
```
Martchus, line 2:
```suggestion
# Copyright SUSE LLC
```

Martchus (approved):

asdil12:
> I don't think it's a good idea to add another complete test run just to test the reset. How long does it take?

12 seconds

We can't reuse the normal full stack test as it would fail otherwise. The whole sense of it is to check that a single `power('off')` is gracefully handeled.

asdil12:
> Also please create a PR based on your item fork

?

okurz:
> > Also please create a PR based on your item fork
> 
> What do you mean?

Sorry. Please create a PR based on your own fork

asdil12:
> Sorry. Please create a PR based on your own fork

Sure: https://github.com/os-autoinst/os-autoinst/pull/2676 (wonder why, though)

okurz:
> > I don't think it's a good idea to add another complete test run just to test the reset. How long does it take?
> 
> 12 seconds
> 
> We can't reuse the normal full stack test as it would fail otherwise. The whole sense of it is to check that a single `power('off')` is gracefully handled.

But then it's hardly a good "full stack test", right? Our full stack test should just do `power('off')` instead of typing the shutdown. Why call "reset"? Maybe move that to a separate test, e.g. in t/18-backend-qemu.t

asdil12:
I didn't call anything reset. I just execute a command (`power('reset')`) that triggers a QMP command after `power('off')` in order to make the coverage happy (which isn't happy anyways because it does perl things and doesn't detect the coverage even though I even check for the line output in the logs, so I still had to mark it uncoverable).

I don't see any value in moving code around - that would be a waste of time. Let's just merge it.
Also please comment on my PR rather than here.

okurz:
All included with https://github.com/os-autoinst/os-autoinst/pull/2676#event-16746919354

PR #2651, opened 2025-02-07
t: Consistently use Test::Exception method instead of manual eval checks

okurz:
Related progress issue: https://progress.opensuse.org/issues/176475

PR #2652, opened 2025-02-07
Use Feature::Compat::Try in place of Try::Tiny

okurz:
After:
* #2651
* #2653

Related progress issue: https://progress.opensuse.org/issues/176475


backend/baseclass.pm
```diff
@@ -657,12 +657,14 @@ sub select_console ($self, $args) {
...
-    catch {
-        {error => $_};
+    my $activated = do {
+        try {
+            local $SIG{__DIE__} = 'DEFAULT';
+            $selected_console->select;
+        }
+        catch ($e) {
+            {error => $e};
+        }
```
perlpunk, line 667: It looks like this could be simplified by replacing the
```
return $activated if ref($activated);
```
below with this:
```suggestion
        catch ($e) { return {error => $e} }
```
The subtle difference between implicitly returning from the block and explicitly `return`ing from the sub might be hard to get at first.
But this case is an example where it comes in handy.

okurz:
Needs #2653 first for os-autoinst dependencies within the openQA full-stack test

t/07-commands.t
```diff
@@ -225,7 +225,7 @@ subtest 'current_script test' => sub {
 
 kill TERM => $spid;
 waitpid($spid, 0);
-combined_like { eval { $cserver->stop() } } qr/commands process exited/, 'commands server stopped';
+combined_like { $cserver->stop() } qr/commands process exited/, 'commands server stopped';
```
perlpunk, line 228: ?

perlpunk:
Oh no, I commented on it and now realized i'ts a PR with 6 commits... :(

I will stop reviewing until I see a PR with one commit

okurz:
> Oh no, I commented on it and now realized i'ts a PR with 6 commits... :(
> 
> I will stop reviewing until I see a PR with one commit

Then I recommend you start with
* #2654

okurz:
> > Oh no, I commented on it and now realized i'ts a PR with 6 commits... :(
> > I will stop reviewing until I see a PR with one commit

Sorry if something upset you. There are multiple changes that accumulated and in the end I want to have all included. As you said you prefer multiple PRs so I created multiple ones. And I followed all points from https://progress.opensuse.org/projects/qa/wiki/Tools#Collaboration-best-practices regarding dependant PRs for easier review.

> Then I recommend you start with
> 
>     * [t: Capture SYNC_ASSETS_HOOK debug output #2654](https://github.com/os-autoinst/os-autoinst/pull/2654)

t/07-commands.t
okurz: This never needed eval

okurz:
Who can help me understand https://github.com/os-autoinst/os-autoinst/actions/runs/13240349733/job/36962037922?pr=2652#step:4:314 ? Can't reproduce that locally with `prove -I. xt/06-pod-coverage.t`

perlpunk:
> Who can help me understand https://github.com/os-autoinst/os-autoinst/actions/runs/13240349733/job/36962037922?pr=2652#step:4:314 ? Can't reproduce that locally with `prove -I. xt/06-pod-coverage.t`

See #2660 
It can't compile testapi, probably because Feature::Compat::Try is still missing in that container. Unfortunately Pod::Coverage doesn't reveal the full error message

Martchus:
I re-triggered services on https://build.opensuse.org/package/show/devel:openQA/os-autoinst_dev but I think this has already been done automatically. So I'm not sure why the dependency is still missing.

perlpunk (approved):

perlpunk:
https://registry.opensuse.org/ says the amd64 container has been built yesterday morning, while the aarch64 has been built a couple of hours ago...

okurz:
Finally https://build.opensuse.org/package/show/devel:openQA/os-autoinst_dev is succeeded. Let's see if retrigger helps.

Wabri (approved):

PR #2653, opened 2025-02-07
Add dependency on Feature::Compat::Try

okurz:
After:
* #2651

Related progress issue: https://progress.opensuse.org/issues/176475


t/33-vagrant.t
```diff
@@ -190,8 +190,7 @@ subtest 'backend creation dies when VAGRANT_ASSETDIR cannot be opened' => sub {
     $backend_vars->{VAGRANT_ASSETDIR} = $fakedir;
     $backend_vars->{VAGRANT_BOX} = "/foobar";
 
-    eval { backend::vagrant->new(); };
-    like($@, qr/Could not opendir $fakedir/, 'opendir error message is present in the output');
+    throws_ok { backend::vagrant->new() } qr/Could not opendir $fakedir/, 'opendir error message is present in the output';
```
b10n1k, line 193: `Could not opendir` seems like it meant to be `Could not open dir`?! which if this is the case i would like to request a better message

okurz: I honestly didn't even read the string. If you care then please investigate what is meant here in detail and propose a separate PR

perlpunk (approved):

baierjan (approved):

Martchus (approved):

t/33-vagrant.t
perlpunk: `opendir` is a perl function

PR #2645, opened 2025-02-05
Add unit test for code coverage of consoles/s3270.pm

gauravpathak:
Related: https://progress.opensuse.org/issues/168364

PR #2656, opened 2025-02-10
Extension to #2645 with some ideas

okurz:
* t: Add coverage for consoles::s3270::ensure_screen_update
* Extract while loop body method in consoles::s3270
* t: Add coverage for consoles::s3270 clear_buffer
* fixupgit log Simplify
* consoles: Mark uncoverable error condition in s3270
* Simplify consoles::s3270
* Add subtest for more subroutines inside s3270.pm
* Apply suggestions from code review
* Change OpenQA::Test::TimeLimit to 5 and add Test::Mock::Time
* Apply suggestions from code review
* Add unit test for code coverage of consoles/s3270.pm


consoles/s3270.pm
```diff
@@ -144,100 +230,11 @@ sub expect_3270 ($self, %arg) {
...
-        {
-            confess "expect_3270: timed out.\n" . "  waiting for ${\Dumper \%arg}\n" . "  last output:\n" . Dumper($result);
-        }
-    }
-
+    my $do_run = 1;
     # tracing output
+    while ($do_run) {
+        $do_run = $self->_handle_expect_3270_cycle($result, time(), %arg);
+    }
```
Martchus, line 237: Without the loop body the this comment is now lacking context so I'd just remove it. You can now also avoid the variable and loop body.

```suggestion
    while ($self->_handle_expect_3270_cycle($result, time(), %arg));
```

okurz: Well, my Perl says `while (...);` is not valid syntax but at least we can shorten it to `while (...) {};`

Martchus: That is a disappointment :-)
We can go for `{}` then.

Martchus (approved):

PR #2657, opened 2025-02-10
Extend timeout for rsync commands done by the svirt backend to 15 min

Martchus:
After 0d09ec97e6c84068e2f9911c1d0229f991484a15 the timeout of 5 minutes is actually effective. We now see that some jobs are running into that timeout, see https://progress.opensuse.org/issues/176868.

It makes sense to set a higher timeout for these asset downloads because rsync would not produce any output during the possibly long-running download. We could use `--progress` but it would probably be too much having all these status updates in our log files. So this change uses `--timeout` instead so rsync will itself time out. It also sets the timeout of `run_cmd` accordingly (with one minute headroom).

This change also sets the timeout in case `SVIRT_WORKER_CACHE` is used to enforce it consistently regardless of that setting.

The timeout itself is configurable via `SVIRT_ASSET_DOWNLOAD_TIMEOUT_M`.

---

I haven't done a full test run yet but the change is hopefully simple enough so unit tests are enough.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -381,17 +381,18 @@ sub _copy_nvram_vmware ($self, $name, $vmware_openqa_datastore, $vmware_disk_pat
 sub _system (@cmd) { system @cmd }    # uncoverable statement
 
 sub _copy_image_else ($self, $file, $file_basename, $basedir) {
-    my $timeout_s = ONE_MINUTE * ($bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_TIMEOUT_M} // 15);
+    my $download_timeout_s = ONE_MINUTE * ($bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_TIMEOUT_M} // 15);
+    my $inactivity_timeout_s = ONE_MINUTE * ($bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_INACTIVITY_TIMEOUT_M} // 2.5);
 
     # utilize asset possibly cached by openQA worker, otherwise sync locally on svirt host (usually relying on NFS mount)
     if (($bmwqemu::vars{SVIRT_WORKER_CACHE} // 0) && -e $file_basename && defined which 'rsync') {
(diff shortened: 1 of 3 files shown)


consoles/sshVirtsh.pm
```diff
@@ -380,15 +381,17 @@ sub _copy_nvram_vmware ($self, $name, $vmware_openqa_datastore, $vmware_disk_pat
...
     if (($bmwqemu::vars{SVIRT_WORKER_CACHE} // 0) && -e $file_basename && defined which 'rsync') {
         my %c = $self->get_ssh_credentials;
         my $abs = path($file_basename)->to_abs;    # pass abs path so it can contain a colon
         bmwqemu::diag "Syncing '$file_basename' directly from worker host to $c{hostname}";
-        _system("sshpass -p '$c{password}' rsync -e 'ssh -o StrictHostKeyChecking=no' -av '$abs' '$c{username}\@$c{hostname}:$basedir/$file_basename'");
+        _system("sshpass -p '$c{password}' rsync -e 'ssh -o StrictHostKeyChecking=no' --timeout='$timeout_s' -av '$abs' '$c{username}\@$c{hostname}:$basedir/$file_basename'");
     }
     else {
-        $self->run_cmd("rsync -av '$file' '$basedir/$file_basename'") && die 'rsync failed';
+        $self->run_cmd("rsync --timeout='$timeout_s' -av '$file' '$basedir/$file_basename'", timeout => $timeout_s + ONE_MINUTE) && die 'rsync failed';
```
perlpunk, line 394: I think for the rsync --timeout we can actually use a much lower timeout, because this is only triggered when no I/O is happening for that time

nicksinger:
JFYI: There is also `--info=progress2` to just print the total progress. `--stats` might be also interesting to only print total statistics *after* transfer (to spot slow connections)

okurz (approved):

Martchus:
@nicksinger I added `--stats`. I guess `--info=progress2` wouldn't make much of a difference because we only have one file here.

okurz (approved):

PR #2659, opened 2025-02-10
Ensure a clean swtpm directory

r-richardson:
- Increase reliability by always cleaning up existing temporary directory containing swtpm-sock
- Remove handling for dealing with existing socket
- Fix expected error message within t/18-backend-qemu.t

With this we do not have to rely on the cleanup process anymore, creating the socket files fresh everytime to increase reliability.


okurz:
s/containinig/containing/

okurz (changes requested):
Please provide more detail why this is the right approach

Martchus:
This simplifies things (which is quite obvious from the diff) which is already a plus on its own.

I also think that it is generally a good approach not to expect that possibly persistent directories like these are in a usable/clean state. (The alternative would be to improve the cleanup but the cleanup can always go wrong so it would never be really reliable.)

Martchus (approved):

r-richardson:
> Please provide more detail why this is the right approach
yes, as marius wrote i thought this would be more reliable, while also making it simpler at the same time

okurz (approved):

PR #2660, opened 2025-02-10
t: Add reason for undefined Pod::Coverage result

perlpunk:
If it returns undef, `why_unrated` will print the reason, for example

    requiring 'testapi' failed

See also https://github.com/os-autoinst/os-autoinst/pull/2652#issuecomment-2648243512


perlpunk:
Hm, that is failing on master, too...

okurz:
+1

okurz (approved):

PR #2661, opened 2025-02-11
Add support for newline seperator in SCHEDULE variable

nicksinger:


perlpunk (commented):
I would rephrase it as "Split SCHEDULE variable on line breaks as well".
With the split you are not removing only trailing line breaks

t/20-openqa-isotovideo-utils.t
```diff
@@ -52,7 +52,7 @@ subtest 'error handling when loading test schedule' => sub {
     subtest 'unable to load test module' => sub {
         $base_state->remove;
         my $module = 'foo/bar';
-        $bmwqemu::vars{SCHEDULE} = $module;
+        $bmwqemu::vars{SCHEDULE} = $module."\n";
```
nicksinger, line 55: I cannot come up with a way to properly test that this only includes one single file because apparently this subtest only has access to `$state` which doesn't provide much insight:

```JSON
{
  'msg' => 'unable to load foo/bar.pm, check the log for the cause (e.g. syntax error)',
  'component' => 'tests'
};
```

unfortunately this is also not an implicit test because the error message stays the same for two files defined in `$module`. Maybe somebody else can help out here?

okurz: Well, how about TDD, i.e. revert your change, add add a failing test, add implementation back to fix it

okurz (approved):

r-richardson:
LGTM (besides that the PR title is now wrong)

t/20-openqa-isotovideo-utils.t
nicksinger: Yeah I tried this already. The thing is: I don't know on what to fail - the needed information of _how many_ tests have been loaded is not available in this subtest here and I couldn't find a way looking at existing tests how I can get this information.

Martchus (approved):

PR #2664, opened 2025-02-13
Fix log colors

perlpunk:
The problem was that when calling `fctinfo` for example, but the log level would not include info, it would print the `color('yellow')`, but not the `color('reset')`.

The result is that the prefix

    2025-02-13T15:50:34.660152Z] [debug] [pid:2866]

itself is not colored anymore, so I had to adapt the tests.

Issue: https://progress.opensuse.org/issues/177024

### DIFF
--- a/log.pm
+++ b/log.pm
@@ -32,36 +32,31 @@ sub log_format_callback ($time, $level, @items) {
 
 sub diag (@args) {
     confess "missing input" unless @args;
-    logger->append(color('white'));
-    $args[-1] .= color('reset');
+    $args[-1] = color('white') . ($args[-1] // '') . color('reset');
     logger->debug(@args);
     return;
 }
 
 sub fctres ($text, $fname = undef) {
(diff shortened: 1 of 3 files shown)


baierjan (approved):

b10n1k (approved):

Martchus (approved):

perlpunk:
This now has the problem that the prefix on the logline will not be colored, only the message itself.
I think there is no easy way around this.
Mojo::Log itself supports colors, but very limited and not configurable.

perlpunk:
I adapted the tests to that new behaviour, but we have to decide if we want this.
Colored log lines don't stand out that much anymore the prefix itself is not colored.

okurz:
> Colored log lines don't stand out that much anymore the prefix itself is not colored.

hm, ok

okurz (approved):

baierjan (approved):

PR #2667, opened 2025-02-21
perltidy: Use new --pack-operator-types option

perlpunk:
It was added in version 20250214.

See perltidy/perltidy#171

We temporarily used `---freeze-newlines` to disable wrapping chained
method calls.

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -12,6 +12,4 @@
 -bt=2    # no spaces around []
 -sbt=2   # no spaces around {}
 -sct     # stack closing tokens )}
-# Prevent excessive line-wrapping
-# See https://github.com/perltidy/perltidy/issues/171
---freeze-newlines
+--pack-operator-types='->' # allow chained method calls on one line
--- a/cpanfile
+++ b/cpanfile
@@ -101,7 +101,7 @@ on 'develop' => sub {
     requires 'Code::TidyAll';
     requires 'Devel::Cover';
     requires 'Module::CPANfile';
-    requires 'Perl::Tidy', '== 20250105.0.0';
+    requires 'Perl::Tidy', '== 20250214.0.0';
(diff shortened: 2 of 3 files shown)


perlpunk:
I think the container still has to be updated after https://github.com/os-autoinst/openQA/pull/6196
Otherwise I can't explain the test failure

baierjan:
> I think the container still has to be updated after [os-autoinst/openQA#6196](https://github.com/os-autoinst/openQA/pull/6196) Otherwise I can't explain the test failure

The test are still running with the container build from master and with perltidy version from there, right? So you need to bump the version first and use the new feature in another PR after?

perlpunk:
> The test are still running with the container build from master and with perltidy version from there, right? So you need to bump the version first and use the new feature in another PR after?

I think that shouldn't be necessary in general, as we don't install a specific version in the Dockerfile.
But it seems it's needed to trigger a new container build...

baierjan:
> > The test are still running with the container build from master and with perltidy version from there, right? So you need to bump the version first and use the new feature in another PR after?
> 
> I think that shouldn't be necessary in general, as we don't install a specific version in the Dockerfile. But it seems it's needed to trigger a new container build...

Yeah right, no fixed version; then disregard my comment. So triggering the service manually as you just did could be the best workaround; although it should already happen automatically on dependency change (in this case the Perl-Tidy package).

The build log from the last build (from `Feb 20 11:55:01`) show `perl-Perl-Tidy-20250105.0.0-1.1.noarch` because the new package was accepted into Factory just 22 hours ago and the build service was not able to build a new image yet (I guess it is still working on some dependencies in the chain).

okurz (approved):

baierjan (approved):

Martchus (approved):

PR #2674, opened 2025-03-11
testapi: allow keeping mouse in place after clicking

marmarek:
Sometimes it's useful to keep the mouse cursor in place after clicking,
to avoid spurious hover events between actions. This is especially
useful when emulated mouse is really a mouse (with relative movement)
instead of a tablet device (with instant teleport).
On the API side, there is already mousehide arg, accepting 0 and 1
values (or true/false). Logically, it should be mousehide=>0, but that
has other meaning already. Implement this as mousehide=>-1 instead.


marmarek:
Looks like there is some issue with CI infra:
> Retrieving repository 'openSUSE-Tumbleweed-Non-Oss' metadata [....error]
Repository 'openSUSE-Tumbleweed-Non-Oss' is invalid.
Warning: Skipping repository 'openSUSE-Tumbleweed-Non-Oss' because of the above error.

and:
> Location 'http://cdn.opensuse.org/distribution/leap/15.6/repo/oss/noarch/perl-PPIx-Utils-0.003-bp156.1.1.noarch.rpm' is temporarily unaccessible.

b10n1k (approved):

PR #2676, opened 2025-03-13
Enable is_shutdown to be used after "power('off')" on qemu

asdil12:
"power('off')" for the qemu backend sends the QMP command "quit" which
causes the whole qemu stack to shutdown closing the QMP pipe from the
remote end which previously triggered a "Broken pipe" exception whenever
"is_shutdown" is called after the power command. This commit explicitly
handles this situation by returning the "shutdown" status correctly.
Other combinations of testapi commands if used in undefined operations
might still fail

Tested using t/99-full-stack.t by using "power('off')" instead of "sudo
powerdown" in t/data/tests/tests/shutdown.pm

Related progress issue: https://progress.opensuse.org/issues/176319

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -453,7 +453,7 @@ sub stop_vm ($self, @) {
     }
     $self->_stop_video_encoder();
     $self->close_ssh_connections();
-    $self->close_pipes();    # does not return
+    $self->close_pipes(1);    # does not return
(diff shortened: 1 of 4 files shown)


Martchus (approved):

okurz (approved):

okurz:
Perfect :)

PR #2681, opened 2025-03-18
Avoid leftover rsync processes on svirt host

Martchus:
Alternative to https://github.com/os-autoinst/os-autoinst/pull/2682.

---

* Disable retries introduced by c30cd0ec again as this potentially leaves leftover rsync processes slowing everything down
* See https://progress.opensuse.org/issues/178324 and https://progress.opensuse.org/issues/179077

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -676,7 +676,7 @@ sub run_cmd ($self, $cmd, %args) {
 }
 
 sub run_cmd_retrying_on_timeouts ($self, $command, @args) {
-    my $attempts = $bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_ATTEMPTS} // 3;
+    my $attempts = $bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_ATTEMPTS} // 1;
     for (my $attempt = 1;;) {
         try {
             return $self->run_cmd($command, @args);
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -451,6 +451,7 @@ subtest 'SSH usage in console::sshVirtsh' => sub {
     };
 
     subtest 'running command with retry on ssh timeout' => sub {
+        local $bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_ATTEMPTS} = 3;
         %ssh_expect = ();
         $run_ssh_cmd_return = 42;
         $fake_timeouts = 2;


okurz:
I suggest you first try that out locally over some days

Martchus:
With "locally" you mean on OSD workers? Yes, we can also apply this mitigation/workaround just there considering that probably nobody but use uses this backend anyway.

perlpunk (approved):

consoles/sshVirtsh.pm
```diff
@@ -676,7 +676,7 @@ sub run_cmd ($self, $cmd, %args) {
 }
 
 sub run_cmd_retrying_on_timeouts ($self, $command, @args) {
-    my $attempts = $bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_ATTEMPTS} // 3;
+    my $attempts = $bmwqemu::vars{SVIRT_ASSET_DOWNLOAD_ATTEMPTS} // 1;
```
okurz, line 679: but this disables retrying which we put in on purpose. What will be the effect of this then?

okurz (changes requested):

consoles/sshVirtsh.pm
Martchus: It means we're back to the problem where jobs fail after running into the already generous timeout. It didn't happen that often and I haven't seen these kinds of failures in the last 2 days (but only checked a small number of jobs).

okurz: But shouldn't we have retries to cover short service outages?

Martchus: I guess retries *with backoff* would make sense to cover service outages. The retries implemented here have no backoff so if there's an outage for e.g. 10 seconds the retry here would not be of much use and just provoke the connection error 3 times. Well, actually it wouldn't because the retry done here is only done on timeouts anyway.

I really only introduced the retry very recently to improve the case when this runs into a timeout as this was the problem reported in the ticket I was working on. (Hence also the function name `…_on_timeouts`.)

okurz: ok. I am still not quite sure that this is the right approach to have retries supported in code but not used by default anywhere but you are the expert :)

okurz (approved):

consoles/sshVirtsh.pm
Martchus: I don't want to remove the retry completely because I might be able to make it actually work at some point, see https://github.com/os-autoinst/os-autoinst/pull/2682. That PR needs further experimentation, though.

PR #2684, opened 2025-03-19
Consistently use try/catch instead of 'if ' everywhere

okurz:
Corresponding to https://github.com/os-autoinst/openQA/pull/6303

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -151,8 +151,8 @@ sub loadtest ($script, %args) {
     eval $code;
     if (my $err = $@) {
         if ($is_python) {
-            eval "use Inline Python => 'sys.stderr.flush()';";
-            bmwqemu::fctwarn("Unable to flush Python's stderr, error message from Python might be missing: $@") if $@;    # uncoverable statement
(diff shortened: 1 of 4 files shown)


autotest.pm
```diff
@@ -151,8 +151,8 @@ sub loadtest ($script, %args) {
     eval $code;
     if (my $err = $@) {
         if ($is_python) {
-            eval "use Inline Python => 'sys.stderr.flush()';";
-            bmwqemu::fctwarn("Unable to flush Python's stderr, error message from Python might be missing: $@") if $@;    # uncoverable statement
+            try { use Inline Python => 'sys.stderr.flush()'; }
```
perlpunk, line 154: That's a string eval and needs to stay

perlpunk: Or:
```suggestion
            try { require Inline; import Inline Python => 'sys.stderr.flush()'; }
```

okurz: you are the best! Updated

perlpunk (approved):

b10n1k (approved):

PR #2686, opened 2025-03-25
Better error when syswrite fails in myjsonrpc

asdil12:
Ticket: https://progress.opensuse.org/issues/176319

### DIFF
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -50,7 +50,7 @@ sub send_json ($to_fd, $cmd) {
     my $written_bytes = _syswrite($to_fd, $json);
     if (!$written_bytes || $written_bytes != length($json)) {
         die('myjsonrpc: remote end terminated connection, stopping') if !DEBUG_JSON && $! =~ qr/Broken pipe/;
-        confess "syswrite failed: $!";
+        confess sprintf "syswrite failed: err: '%s'; written_bytes: %d/%d; JSON: '%s'", $!, $written_bytes, length($json), $json;
     }
     return $cmdcopy{json_cmd_token};
 }


myjsonrpc.pm
```diff
@@ -50,7 +50,7 @@ sub send_json ($to_fd, $cmd) {
     my $written_bytes = _syswrite($to_fd, $json);
     if (!$written_bytes || $written_bytes != length($json)) {
         die('myjsonrpc: remote end terminated connection, stopping') if !DEBUG_JSON && $! =~ qr/Broken pipe/;
-        confess "syswrite failed: $!";
+        confess "syswrite failed: err: '$!'; written_bytes: $written_bytes/" . length($json) . "; JSON: $json";
```
b10n1k, line 53: First I see that the json is displayed above in https://github.com/os-autoinst/os-autoinst/blob/b0d66c3907e8c6ee61e80cb6b081db38448d789c/myjsonrpc.pm#L46 if debug. So here feels a bit redundant . 
Then I would say that `written_bytes: $written_bytes/" . length($json)` is also belong in that `bmwqemu::diag`. 
but my main objection would be the misuse of confess, as the main purpose is to print a stack trace, right? So even if you do not like to add the _details_ in the `bmwqemu::diag` in line 46, I would say to add the error message that you want in another `bmwqemu::diag`.

okurz (approved):

myjsonrpc.pm
asdil12: I don't see why we shouldn't add proper information at the top of the stack trace.
If it would be diag, it would be harder to find.
I add the json, because I can't reproduce an issue locally and the test in the CI doesn't run with json-debug mode.

perlpunk: In line 46 the diag is only called if `is_debug()`, so normally not.

The only thing I would improve here is using sprintf to avoid that `" . ... . "`
```suggestion
        confess sprintf "syswrite failed: err: '%s'; written_bytes: %d/%d; JSON: '%s'", $!, $written_bytes, length($json), $json;
```

asdil12: Applied Tina's suggestion

perlpunk (approved):

Martchus (approved):

Issue #2679, opened 2025-03-16
Memory issues (openqa.debian.net)


rclobus:
Hello,

on our Debian instance, we regularly see memory corruption error messages (e.g. `double free or corruption (!prev)`, `free(): invalid pointer`)

Do you have a suggestion which program is causing these? (And how to fix them?)

Some samples are:
https://openqa.debian.net/tests/345903/logfile?filename=autoinst-log.txt#line-1664
https://openqa.debian.net/tests/370340/logfile?filename=autoinst-log.txt#line-5213
https://openqa.debian.net/tests/370428/logfile?filename=autoinst-log.txt#line-151
https://openqa.debian.net/tests/370429/logfile?filename=autoinst-log.txt#line-3826
https://openqa.debian.net/tests/370434/logfile?filename=autoinst-log.txt#line-718
https://openqa.debian.net/tests/370435/logfile?filename=autoinst-log.txt#line-599
https://openqa.debian.net/tests/370892/logfile?filename=autoinst-log.txt#line-150
https://openqa.debian.net/tests/371101/logfile?filename=autoinst-log.txt#line-154
https://openqa.debian.net/tests/371222/logfile?filename=autoinst-log.txt#line-365
https://openqa.debian.net/tests/371253/logfile?filename=autoinst-log.txt#line-189
https://openqa.debian.net/tests/371555/logfile?filename=autoinst-log.txt#line-2021
https://openqa.debian.net/tests/371557/logfile?filename=autoinst-log.txt#line-1059
https://openqa.debian.net/tests/371558/logfile?filename=autoinst-log.txt#line-986

Martchus:
> Do you have a suggestion which program is causing these? (And how to fix them?)

Figuring out what program causes this would be the first step, indeed. I guess one could try running this with Valgrind but it is a bit complicated because we're talking about tree of processes here.

rclobus:
I got my first stacktrace of the main thread:

```
gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at ./nptl/pthread_kill.c:44
#1  0x00007f8aa63cbf4f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007f8aa637cfb2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
[...]

PR #2693, opened 2025-04-05
Fix buffer overflow in isotovideo

rclobus:
When the ZRLE compressed image has a larger resolution than the current image, resize the image to prevent a buffer overflow.

Fixes https://github.com/os-autoinst/os-autoinst/issues/2679


okurz:
CI checks are failing due to an issue with registry.opensuse.org. I reported that as https://progress.opensuse.org/issues/180059. So not related to your changes.

okurz (approved):

PR #2694, opened 2025-04-07
New script: os-autoinst-generate-needle-previews

rclobus:
Generate .svg files with the needle information as an overlay


rclobus:
This is the script I mentioned in the openQA testing monthly call from 2025-04-06

It generates an svg file that shows the needles and their type. When applied to the whole needles directory, it helps to quickly preview the needles.

An alternative invocation is:
`grep "DESKTOP-gnome" -l products/debian/needles/*.json | generate_needle_previews`

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+ls -1 "*.json" | generate_needle_previews
+
+=cut
+
+use v5.10;
+use strict;
+use warnings;
+use autodie;
+use Image::PNG;
+use JSON qw( decode_json );
```
perlpunk, line 22:
```suggestion
use JSON::PP qw( decode_json );
```
JSON::PP is part of perl core, JSON.pm is not

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+    say '  <defs';
+    say '     id="defs6" />';
+    say '  <g';
+    say '     id="g8">';
+    say '    <image';
+    printf "       width=\"%d\"\n",  $png->width;
+    printf "       height=\"%d\"\n", $png->height;
+    say '       preserveAspectRatio="none"';
+    printf "       xlink:href=\"%s\"\n", $filename_without_extension . '.png';
+    say '       id="needle_bitmap" />';
```
perlpunk, line 65:
```suggestion
    my $width = $png->width;
    my $height = $png->height;
    say <<~"EOM";
    <?xml version="1.0" encoding="UTF-8" standalone="no"?>
    <!-- Created with generate_needle_previews.pl -->
    <svg version="1.1"
       id="svg2" width="$width" height="$height"
       viewBox="0 0 $width $height"
       xmlns:xlink="http://www.w3.org/1999/xlink"
       xmlns="http://www.w3.org/2000/svg"
       xmlns:svg="http://www.w3.org/2000/svg">
      <defs
         id="defs6" />
      <g
         id="g8">
        <image width="$svg_width" height="$svg_height"
           preserveAspectRatio="none"
           xlink:href="$filename_without_extension.png"
           id="needle_bitmap" />
    EOM
```

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+    foreach my $area (@areas) {
+        say '    <rect';
+        if ($area->{'type'} eq 'match') {
```
perlpunk, line 68: Here and elsewhere: alphanumeric keys don't need quotes
```suggestion
        if ($area->{type} eq 'match') {
```

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+        printf "       y=\"%d\" />",   $area->{'ypos'};
+    }
+    say '  </g>';
+    say '</svg>';
+    select STDOUT;
+    close $FILE;
+}
+
+# Read a list of files to process
+while (<STDIN>) {
```
perlpunk, line 88: Maybe we can make that optional and only process STDIN if the user specifies `-`? Other opinions.

In any case, this needs a `--help` output, as `t/44-scripts.t` checks for that.
You can probably just copy from another simple script in our directory.
https://github.com/os-autoinst/os-autoinst/blob/master/script/check_qemu_oom#L20-L28

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+cd products/XXX/needles
+ls -1 "*.json" | generate_needle_previews
+
+=cut
+
+use v5.10;
+use strict;
+use warnings;
+use autodie;
+use Image::PNG;
```
perlpunk, line 21: So we don't have Image::PNG in our dependencies.
How should we handle that?

Martchus: We could add an optional dependency. It isn't in TW so I wouldn't make it a hard dependency.

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+use JSON qw( decode_json );
+use Data::Dumper;
+
+sub process_file {
+    my ($filename) = @_;
```
Martchus, line 26: It would be great to use signatures in new code:
```suggestion
sub process_file ($filename) {
```

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+    my $FILE;
+
+    chomp $filename;
+    my $filename_without_extension = substr($filename, 0, length($filename) - 5);
```
Martchus, line 30:
```suggestion
    my $filename_without_extension = substr($filename, 0, -5);
```

script/generate_needle_previews
Martchus, line 22:
```suggestion
use Mojo::JSON qw(decode_json);
```

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+    printf "Working on %s\n", $filename_without_extension;
+    my $png = Image::PNG->new();
+    $png->read($filename_without_extension . '.png');
+
+    my $json_text = do {
+        open(my $json_fh, '<:encoding(UTF-8)', $filename_without_extension . '.json');
+        local $/;
+        <$json_fh>;
+    };
+    my $data  = decode_json $json_text;
```
Martchus, line 40: With `use Mojo::File qw(path);`:
```suggestion
    my $data  = decode_json(path("$filename_without_extension.json")->slurp);
```

script/generate_needle_previews
```diff
@@ -0,0 +1,91 @@
...
+    say '   xmlns:svg="http://www.w3.org/2000/svg">';
+    say '  <defs';
+    say '     id="defs6" />';
+    say '  <g';
+    say '     id="g8">';
+    say '    <image';
+    printf "       width=\"%d\"\n",  $png->width;
+    printf "       height=\"%d\"\n", $png->height;
+    say '       preserveAspectRatio="none"';
+    printf "       xlink:href=\"%s\"\n", $filename_without_extension . '.png';
```
Martchus, line 64: This needs escaping.

Martchus (changes requested):

okurz (changes requested):
@rclobus thanks for sharing that. If you want to go "all-in" and have that merged as part of the os-autoinst repository certain style rules as outlined by @perlpunk and @Martchus would need to be fulfilled. If you are willing to implement all them that would be greatly appreciated. However if you fill like this is going to be too much effort feel free to submit to os-autoinst/scripts where the rules are bit more "relaxed", e.g. see https://github.com/os-autoinst/scripts/blob/master/openqa-clone-and-monitor-job-from-pr. Or just share with people individually as "gist" outside the repo

perlpunk:
It seems you branched this off in 2024. That's why 34-git.t is failing in OBS. It was fixed meanwhile in master.
In general it's a good idea to rebase regularly.

CMakeLists.txt
```diff
@@ -206,6 +206,11 @@ install(
     DESTINATION "${CMAKE_INSTALL_BINDIR}"
     PERMISSIONS ${EXECUTABLE_PERMISSIONS}
 )
+install(
+    FILES "${CMAKE_CURRENT_SOURCE_DIR}/script/generate_needle_previews"
```
Martchus, line 210: It would be nicer if the name of the script would follow our pattern for new scripts:
```suggestion
    FILES "${CMAKE_CURRENT_SOURCE_DIR}/script/os-autoinst-generate-needle-previews"
```

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+
+sub process_file ($filename) {
+    my $SVGFILE;
+
+    chomp $filename;
+    my $pp = Mojo::File->new($filename);
+    my $name = $pp->basename(".json");
+    printf "Working on %s\n", $name;
+    my $pngpath = path($pp->dirname, "${name}.png");
+    my $output = qx/file --brief ${pngpath}/;
```
Martchus, line 37:
```suggestion
    my $output = qx/file --brief "${pngpath}"/;
```

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+
+    my $data  = decode_json($pp->slurp);
+    my @areas = @{$data->{area}};
+
+    open $SVGFILE, '>', path($pp->dirname, "${name}.svg");
+    select $SVGFILE;
+
+    my $URIname = url_escape(${name});
+    # The generated file should be identical to the svg file as saved with inkscape as 'Standard SVG'
+    printf <<~"EOM";
```
perlpunk, line 52: If you are not replacing anything in the string, just use print
```suggestion
    print <<~"EOM";
```

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+    my $pngpath = path($pp->dirname, "${name}.png");
+    my $output = qx/file --brief ${pngpath}/;
+    my ($width, $height) = $output =~ /^PNG image data, (\d+) x (\d+)/;
+    if (!$width or !$height) {
+        say 'E: Could not determine image size';
+        return
+    }
+
+    my $data  = decode_json($pp->slurp);
+    my @areas = @{$data->{area}};
```
Martchus, line 45: There's no need to make a copy:
```suggestion
    my $areas = $data->{area};
```
With this you of course need to change the code below from `@areas` to `@$areas`. Maybe you can also just replace `@areas` below with `@{$data->{area}}` and remove this line completely.

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+
+    my $URIname = url_escape(${name});
+    # The generated file should be identical to the svg file as saved with inkscape as 'Standard SVG'
+    printf <<~"EOM";
+       <?xml version="1.0" encoding="UTF-8" standalone="no"?>
+       <!-- Created with generate_needle_previews -->
+
+       <svg
+          version="1.1"
+          id="svg2"
```
Martchus, line 58:
```suggestion
```

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+          width="${width}"
+          height="${height}"
+          viewBox="0 0 ${width} ${height}"
+          xmlns:xlink="http://www.w3.org/1999/xlink"
+          xmlns="http://www.w3.org/2000/svg"
+          xmlns:svg="http://www.w3.org/2000/svg">
+         <defs
+            id="defs6" />
```
Martchus, line 66:
```suggestion
```

script/generate_needle_previews
```diff
@@ -0,0 +1,132 @@
...
+         <image
+            width="${width}"
+            height="${height}"
+            preserveAspectRatio="none"
+            xlink:href="${URIname}.png"
+            id="needle_bitmap"
+            x="0"
+            y="0" />
+       EOM
+    my $area_counter = 1;
```
Martchus, line 76: Do we need all these IDs? If not this counter could probably be removed.

Martchus (changes requested):

script/generate_needle_previews
rclobus: Yes, we need different IDs in the SVG file. I've made the SVG files identical to the output of inkscape.

script/generate_needle_previews
rclobus: I don't see a difference here. Are you suggesting to change the indentation? I've made the indentation of the SVG file identical to the output of inkscape (which has an indentation of 2 per level, and 3 for the attributes in the tag)

script/generate_needle_previews
rclobus: Ah thanks, there could be spaces or other mean stuff in the path name.

rclobus:
I'll later update the t/41 test. I assumed that the png files were images, but they are actually symbolic links. I'll generate suitable test files later.

script/generate_needle_previews
r-richardson: i guess these two lines seem to not serve any purpose, the diff suggests to remove them entirely

script/generate_needle_previews
perlpunk: I don't know how this is called, but if the path is really untrusted and can contain anything (e.g. also double quotes), then this is also not safe.

t/41-os-autoinst-generate-needle-preview.t
```diff
@@ -0,0 +1,25 @@
...
+
+use FindBin '$Bin';
+use lib "$FindBin::Bin/lib", "$Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '5';
+
+sub check_os_autoinst_generate_needle_preview_default_needle ($commandline) {
+    system($commandline);
+    is $?, 0, 'Command executed without error code';
+    is compare("$Bin/data/os-autoinst-generate-needle-preview.svg", "$Bin/data/os-autoinst-generate-needle-preview.ref.svg"), 0, 'Equal to reference';
+    unlink("$Bin/data/os-autoinst-generate-needle-preview.svg");
```
perlpunk, line 18: We generally avoid creating files in the local git checkout. If the test fails and aborts early for some reason, the file is leftover, and our git directories fill up with stuff we don't know where it is coming from.

If you look at other tests in os-autoinst, most of them have this:
```
my $dir = tempdir("/tmp/$FindBin::Script-XXXX");
chdir $dir;
```
And then everything is happening in that directory.

I know that others in my team care less about it, so I'm just saying.

script/generate_needle_previews
Martchus: No, I suggest you remove this tag completely.

rclobus: I did the following:
* I removed the complete `defs` tag and the `id` attribute from the `svg` tag
* The files could be opened by inkscape
* I saved the file again in inkscape
* The `defs` tag was reinserted, now with `id="defs1"` (which makes more sense than 'defs6')
* The `svg` tag has the id reinserted, now with `id="svg1"` (also makes more sense than 'svg2')

My goal was to create a file that could be opened and saved in inkscape, and still have the same content.
The SVG file definition might allow for a few less tags or attributes, but that wasn't my reference.

Martchus (approved):

AdamWill:
Doesn't this need to be added to the RPM spec?

And isn't it odd that this is installed to `/usr/bin` while other scripts, e.g. the perennially-useful `crop.py`, are installed to `/usr/lib/os-autoinst/script`?

PR #2695, opened 2025-04-09
Remove unused variables and exports from testapi

perlpunk:
It seems the code using `%cmd` was already removed in 75678433628cbb876ba8efcd22021c50c093b0ef
and `%vars`in
33f52cd18cbd6db7b89a176a99b6d4118abb83f2

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -29,7 +29,7 @@ use Time::Seconds;
 require bmwqemu;
 use constant OPENQA_LIBPATH => '/usr/share/openqa/lib';
 
-our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
+our @EXPORT = qw($realname $username $password $serialdev
 
   get_var get_required_var check_var set_var get_var_array check_var_array autoinst_url
 
@@ -62,8 +62,6 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   save_tmp_file get_test_data
 );
 our @EXPORT_OK = qw(is_serial_terminal);
-
-our %cmd;
 
 our $distri;
 


testapi.pm
```diff
@@ -60,11 +60,10 @@ our @EXPORT = qw($realname $username $password $serialdev %cmd %vars
   diag hashed_string
 
   save_tmp_file get_test_data
+
```
b10n1k, line 63: I guess this can be reverted, no?

perlpunk: done. it seems that github is a bit slow and still didn't receive my push

b10n1k (approved):

perlpunk:
For some reason github has problems with this PR. I'll try another force push

r-richardson (approved):

DimStar77:
https://openqa.opensuse.org/tests/4983982 and many more:

```
  Global symbol "%cmd" requires explicit package name (did you forget to declare "my %cmd"?) at microos/lib/utils.pm line 1409.
  Type of arg 1 to testapi::wait_screen_change must be block or sub {} (not reference constructor) at microos/lib/utils.pm line 1880, near "};"
```

okurz:
> https://openqa.opensuse.org/tests/4983982 and many more:
> 
> ```
>   Global symbol "%cmd" requires explicit package name (did you forget to declare "my %cmd"?) at microos/lib/utils.pm line 1409.
>   Type of arg 1 to testapi::wait_screen_change must be block or sub {} (not reference constructor) at microos/lib/utils.pm line 1880, near "};"
> ```

Looking into it. I prepared a PR https://github.com/os-autoinst/os-autoinst/pull/2696 but haven't merged it yet. Looking into a better solution in osado. But it was bad to do changes to testapi on what is exported and not bumping the interface number.

okurz:
@perlpunk I wonder about https://github.com/os-autoinst/os-autoinst/pull/2695#issuecomment-2789625997 . You crossed off the checkmark but did not bump the version number. I guess anything in perl "EXPORT" should be considered part of the API so any change should be seen as potentially breaking warranting a version number, isn't it?

perlpunk:
Very sorry for that.

I saw that two empty variables are exported that are never used or filled anywhere, there are no unit tests and they are not documented.
I really thought they aren't used because it makes no sense, so I didn't see the need for a version update.
I even checked osado, but apparently not good enough, as I can find it now.

Regarding `%cmd`: I see that in `lib/susedistribution.pm` they are importing `%cmd` and then later use the fully qualified version as `%testapi::cmd = ...`.
So people fill that has with stuff, but os-autoinst has **nothing** to do with it, it just provides a storage for this. Like a stash I recently suggested, but this time as a global var which is very bad.

This should be an API function, if the tests really need a variable to store stuff. And it should be documented. I currently don't completely understand why it is needed and why it can't just be a hash in the module itself.

That's really bad design.

okurz:
> So people fill that hash with stuff, but os-autoinst has **nothing** to do with it, it just provides a storage for this. Like a stash I recently suggested, but this time as a global var which is very bad.
> 
> This should be an API function, if the tests really need a place to store stuff. And it should be documented. I currently don't completely understand why it is needed and why it can't just be a hash in the module itself.
> 
> That's really bad design.

I think the original commit that you identified is also at the time when such code was moved into the test distribution, in particular lib/susedistribution.com and hardly anyone ever touches this.

PR #2696, opened 2025-04-10
Revert "Remove unused variables and exports from testapi"

okurz:
Reverts os-autoinst/os-autoinst#2695 due to failures showing up on openqa.opensuse.org, see https://github.com/os-autoinst/os-autoinst/pull/2695#issuecomment-2791751789


okurz:
I consider the necessary change to osado for `%cmd` too big and likely other test distributions are affected the same. But `%vars` might be ok so updated the commit accordingly.

baierjan (approved):

okurz:
As apparently package caches on o3 workers are empty force-merging now to expedite the deployment

PR #2698, opened 2025-04-23
WIP: NOT MERGE

Wabri:


cmake/test-targets.cmake
```diff
@@ -90,10 +90,11 @@ else ()
 endif ()
 
 # add targets for invoking Perl test suite
-find_program(PROVE_PATH prove)
+set(PROVE_PATH "prove")
+find_program(PROVE_PATH_FOUND NAMES ${PROVE_PATH})
```
Martchus, line 94:
```suggestion
find_program(PROVE_PATH prove)
```
Then it should be possible to just check ´if (PROVE_PATH)`.

Martchus (changes requested):

PR #2699, opened 2025-04-23
Changes to video_stream to work on Orange Pi 5 Plus

marmarek:
I'm trying to get a generalhw worker with higher resolution than 1024x768.
Raspberry Pi 4/5 is not enough due to CSI2 interface not having enough
bandwidth for 1920x1080@60Hz, and even that is too low for some systems (I have
one that requires at least 1920x1200@60Hz). Orange Pi 5 Plus seems to be better
fit, but requires few changes. I haven't got it fully working yet (as in: no
full test passed yet), but with this change openQA can at least capture HDMI
input.

There may be some subsequent related PRs, possibly in the input handling (resolution for the USB tablet).

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -89,12 +89,14 @@ sub _v4l2_ctl ($device, $cmd_prefix, $cmd) {
 sub connect_remote ($self, $args) {
     $self->{_last_update_received} = 0;
 
-    if ($args->{url} =~ m/^\/dev\/video/) {
+    if ($args->{url} =~ m/^(ustreamer:\/\/)?(\/dev\/video\d+)/) {
         if ($args->{edid}) {
-            my $ret = _v4l2_ctl($args->{url}, $args->{video_cmd_prefix}, "--set-edid $args->{edid}");
+            my $ret = _v4l2_ctl($2, $args->{video_cmd_prefix}, "--set-edid $args->{edid}");
             die "Failed to set EDID" unless defined $ret;
         }
+    }
 
+    if ($args->{url} =~ m/^\/dev\/video/) {
         my $timings = _v4l2_ctl($args->{url}, $args->{video_cmd_prefix}, '--get-dv-timings');
         if ($timings) {
             if ($timings ne "0x0pnan") {
--- a/t/26-video_stream.t
+++ b/t/26-video_stream.t
@@ -126,6 +126,16 @@ subtest 'connect stream ustreamer' => sub {
     is $console->{dv_timings_supported}, 0, "correctly skipping DV timing";
     is_deeply \@v4l2_ctl_calls, [], "calls to v4l2-ctl";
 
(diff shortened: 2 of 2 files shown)


consoles/video_stream.pm
```diff
@@ -121,8 +123,10 @@ sub connect_remote ($self, $args) {
 }
 
 sub _get_ffmpeg_cmd ($self, $url) {
-    my $fps = $1 if ($url =~ s/\?fps=([0-9]+)//);
+    my $fps = $1 if ($url =~ s/[\?&]fps=([0-9]+)//);
+    my $format = $1 if ($url =~ s/[\?&]format=([A-Z0-9]+)//);
     $fps //= 4;
+    die "ffmpeg url does not support format=" if $format;
```
Martchus, line 129: So this code just reads the format to error out if it was specified?

I also find variable definitions with assignment within a postfix-if a bit hard to read. I suppose the variables will just be undef but still defined in case the condition is false.

consoles/video_stream.pm
```diff
@@ -131,11 +135,13 @@ sub _get_ffmpeg_cmd ($self, $url) {
 }
 
 sub _get_ustreamer_cmd ($self, $url, $sink_name) {
-    my $fps = $1 if ($url =~ s/\?fps=([0-9]+)//);
+    my $fps = $1 if ($url =~ s/[\?&]fps=([0-9]+)//);
+    my $format = $1 if ($url =~ s/[\?&]format=([A-Z0-9]+)//);
```
Martchus, line 139: It would be nice to avoid duplicating this code here.

consoles/video_stream.pm
marmarek: I find debugging issues due to an option being (sometimes) ignored hard, so I added explicit refusal of mismatching option. Would you prefer it to be ignored instead, or maybe just logged with a warning?

Martchus: No, you can keep the code as-is. I was just wondering whether this behavior was intentional.

Martchus (approved):

marmarek:
Is the CI failure related to my changes? I see:
```
    #   Failed test 'results of test 7'
    #   at t/full-stack.t line 416.
    # unable to read /tmp/IX70gLfVuq/full-stack.d/openqa/testresults/../pool/1/autoinst-log.txt: Can't open file "/tmp/IX70gLfVuq/full-stack.d/openqa/testresults/../pool/1/autoinst-log.txt": No such file or directory at t/full-stack.t line 158.
```
which looks unrelated, but I'm not sure

Martchus:
It looks also unrelated to me. I'll re-trigger the GitHub action so we'll see.

okurz (approved):

marmarek:
Using a gray image for color decoding test was not a smart idea... https://openqa.qubes-os.org/tests/137851#step/install_startup/13 it was supposed to be blue...

PR #2700, opened 2025-04-28
Add support for lua test modules

asdil12:
https://progress.opensuse.org/issues/162998

```
mkdir tmp
cd tmp
isotovideo -d CASEDIR=../t/data/tests/ NOVIDEO=1
```

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -109,6 +109,75 @@ sub _debug_python_version () {
     bmwqemu::diag "Using python version " . $debug;
 }
 
+
+=head2 _lua_use
+
+    use("testapi")
+    use("some::perl::module", {"some_function", "another_function"})
+
+This function is the importer for loading Perl modules from Lua test modules.
+It imports everything that a Perl module exports into the global Lua scope.
+Optionally a table listing the desired functions can be passed.
+
+Imported hashes, scalars and arrays are copied as Inline:Lua doesn't support
+references for them.
+
+=cut
+
+sub _lua_use ($perlmod, $import = undef) {
+    eval "use $perlmod ()";
+    if ($@) { die "Could not load '$perlmod': $@" }
+    my %exports;
+    no strict 'refs';
+    my @export = @{"${perlmod}::EXPORT"};
+    my @export_ok = @{"${perlmod}::EXPORT_OK"};
+    @exports{@export, @export_ok} = ();
+    my @to_import;
+    if ($import) {
+        for my $symbol (@$import) {
+            if (exists $exports{$symbol}) {
+                push @to_import, $symbol;
+            }
+            else {
+                warn "'$symbol' is not exported by '$perlmod'";
+            }
+        }
+    }
+    else {
+        @to_import = @export;
+    }
+    foreach my $symbol (@to_import) {
+        if ($symbol =~ s/^\$//) { lua_set($symbol, ${"${perlmod}::${symbol}"}) }
+        elsif ($symbol =~ s/^\%//) { lua_set($symbol, \%{"${perlmod}::${symbol}"}) }
+        elsif ($symbol =~ s/^\@//) { lua_set($symbol, \@{"${perlmod}::${symbol}"}) }
+        else { lua_set($symbol, \&{"${perlmod}::${symbol}"}) }
+    }
+}
+
+sub _load_lua () {
+    state $lua_loaded = 0;
+    return if $lua_loaded++;
+    my $code = <<~'EOM';
+    use Inline Lua => <<EOLUA;
+    function lua_eval(s)
+        return load("return " .. s)()
+    end
+    function lua_set(var, value)
+        _G[var] = value
+        --print("lua_set", var, value)
+    end
+    function lua_append_path(path)
+        package.path = path .. ";" .. package.path
+        --print(package.path)
+    end
+    EOLUA
+    return lua_eval('_VERSION', 0);
+    EOM
+    my $debug = eval $code;
+    bmwqemu::diag "Using Lua version $debug";
+    lua_set("use", \&_lua_use);
+}
+
 sub loadtest ($script, %args) {
     no utf8;    # Inline Python fails on utf8, so let's exclude it here
     my $casedir = $bmwqemu::vars{CASEDIR};
@@ -122,8 +191,9 @@ sub loadtest ($script, %args) {
     $code .= "use lib '$casedir/lib';";
     my $basename = dirname($script_path);
     $code .= "use lib '$basename';";
-    die "Unsupported file extension for '$script'" unless $script =~ /\.p[my]/;
+    die "Unsupported file extension for '$script'" unless $script =~ /\.(p[my]|lua)/;
     my $is_python = 0;
+    my $is_lua = 0;
     if ($script =~ m/\.pm$/) {
         $code .= "require '$script_path';";
     }
@@ -148,6 +218,33 @@ sub loadtest ($script, %args) {
             }
             EOM
         $is_python = 1;
+    }
+    elsif ($script =~ m/\.lua$/) {
+        _load_lua();    # this loads (actually defines) lua_eval()
+        my $script_dir = path($script_path)->dirname->to_abs;
+        $code .= sprintf <<~'EOM', $script_dir, $script_path;
+            use base 'basetest';
+
+            use autotest;
+            autotest::_load_lua();
+            autotest::lua_append_path("%s/../lib/?.lua");
+            autotest::lua_eval('dofile("%s")');
+
+            sub run($self) {
+                # not passing $self as blessed objects generate a warning
+                #autotest::lua_set("p1", $self);
+                return autotest::lua_eval('run()')
+            }
+
+            sub test_flags($self) {
+                return autotest::lua_eval('_G["test_flags"] and test_flags() or {dummy = 1}')
+            }
+
+            sub post_fail_hook($self) {
+                return autotest::lua_eval('_G["post_fail_hook"] and post_fail_hook() or 1')
+            }
+            EOM
+        $is_lua = 1;
     }
     eval $code;
     if (my $err = $@) {
@@ -204,8 +301,8 @@ our $activated_consoles = [];
 our $last_milestone_console;
 
 sub parse_test_path ($script_path) {
-    unless ($script_path =~ m,(\w+)/([^/]+)\.p[my]$,) {
-        die "loadtest: script path '$script_path' does not match required pattern \\w.+/[^/]+.p[my]\n";
+    unless ($script_path =~ m,(\w+)/([^/]+)\.(p[my]|lua)$,) {
+        die "loadtest: script path '$script_path' does not match required pattern \\w.+/[^/]+.(p[my]|lua)\n";
     }
     my $category = $1;
     my $name = $2;
@@ -469,7 +566,7 @@ sub loadtestdir ($dir) {
     $dir =~ s/^\Q$bmwqemu::vars{CASEDIR}\E\/?//;    # legacy where absolute path is specified
     $dir = join('/', $bmwqemu::vars{CASEDIR}, $dir);    # always load from casedir
     die "'$dir' does not exist!\n" unless -d $dir;
-    loadtest($_) for map { s{^\Q$bmwqemu::vars{CASEDIR}/}{}r } (glob "$dir/*.pm");
+    loadtest($_) for map { s{^\Q$bmwqemu::vars{CASEDIR}/}{}r } (glob "$dir/*.{pm,py,lua}");
 }
 
 # This is called if the framework loaded a VM snapshot. All consoles
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -97,6 +97,7 @@ RUN zypper in -y -C \
        'perl(IPC::Open3)' \
        'perl(IPC::Run::Debug)' \
        'perl(IPC::System::Simple)' \
+       'perl(Inline::Lua)' \
        'perl(Inline::Python)' \
        'perl(JSON::Validator)' \
        'perl(List::MoreUtils)' \
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -75,17 +75,25 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %bcond_with black
 %endif
-# SLE is missing Python support requirements
+# SLE is missing Python and Lua support requirements
 %if 0%{?is_opensuse}
 %bcond_without python_support
+%bcond_without lua_support
 %else
 %bcond_with python_support
+%bcond_with lua_support
 %endif
 %if %{with python_support}
 # The following line is generated from dependencies.yaml
 %define python_support_requires perl(Inline::Python)
 %else
 %define python_support_requires %{nil}
+%endif
+%if %{with lua_support}
+# The following line is generated from dependencies.yaml
+%define lua_support_requires perl(Inline::Lua)
+%else
+%define lua_support_requires %{nil}
 %endif
 %if %{with black}
 # The following line is generated from dependencies.yaml
@@ -118,7 +126,7 @@ Source0:        %{name}-%{version}.tar.xz
 # The following line is generated from dependencies.yaml
 %define test_version_only_requires perl(Mojo::IOLoop::ReadWriteProcess) >= 0.28
 # The following line is generated from dependencies.yaml
-%define test_requires %build_requires %ocr_requires %python_support_requires %spellcheck_requires %test_base_requires %test_non_s390_requires %yamllint_requires ffmpeg python3-Pillow-tk
+%define test_requires %build_requires %lua_support_requires %ocr_requires %python_support_requires %spellcheck_requires %test_base_requires %test_non_s390_requires %yamllint_requires ffmpeg python3-Pillow-tk
 # The following line is generated from dependencies.yaml
 %define devel_requires %python_style_requires %test_requires ShellCheck file perl(Code::TidyAll) perl(Devel::Cover) perl(Module::CPANfile) perl(Perl::Tidy) perl(Template::Toolkit) sed shfmt
 %define s390_zvm_requires /usr/bin/xkbcomp /usr/bin/Xvnc x3270 icewm xterm xterm-console xdotool fonts-config mkfontdir mkfontscale openssh-clients
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -17,6 +17,8 @@ use autotest;
 use bmwqemu;
 use OpenQA::Test::RunArgs;
 
+my $has_lua = eval { require Inline::Lua };
+
 $bmwqemu::vars{CASEDIR} = File::Basename::dirname($0) . '/fake';
 
 throws_ok { autotest::runalltests } qr/ERROR: no tests loaded/, 'runalltests needs tests loaded first';
@@ -28,8 +30,8 @@ like warning {
   'loadtest outputs on stderr';
 
 sub loadtest ($test, $msg = "loadtest($test)") {
-    my $filename = $test =~ /\.p[my]$/ ? $test : $test . '.pm';
-    $test =~ s/\.p[my]//;
+    my $filename = $test =~ /\.(p[my]|lua)$/ ? $test : $test . '.pm';
+    $test =~ s/\.(p[my]|lua)//;
     stderr_like { autotest::loadtest "tests/$filename" } qr@scheduling $test#?[0-9]* tests/$test|$test already scheduled@, $msg;
 }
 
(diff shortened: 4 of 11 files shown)


autotest.pm
```diff
@@ -148,6 +204,27 @@ sub loadtest ($script, %args) {
             EOM
         $is_python = 1;
     }
+    elsif ($script =~ m/\.lua$/) {
+        _debug_lua_version(); # this loads (actually defines) lua_eval()
+        $code .= <<~"EOM";
```
perlpunk, line 209: Maybe
```suggestion
        $code .= sprintf <<~'EOM', $script_path;
```
then we don't need those backslashes

asdil12: What backslashes and what does the sprintf here do and why?

autotest.pm
```diff
@@ -148,6 +204,27 @@ sub loadtest ($script, %args) {
...
+
+            sub run(\$self) {
+                use Data::Dumper;
+                print("HALLO\\n");
+                print(Dumper(\$self));
+                autotest::lua_set("p1", \$self);
+                print("WELT\\n");
+                return autotest::lua_eval('run(p1)')
+            }
+            EOM
```
perlpunk, line 224:
```suggestion
        $code .= sprintf <<~'EOM', $script_path;
            use base 'basetest';

            use autotest;
            autotest::_debug_lua_version();
            autotest::lua_eval('dofile("%s")');

            sub run($self) {
                use Data::Dumper;
                print("HALLO\n");
                print(Dumper($self));
                autotest::lua_set("p1", $self);
                print("WELT\n");
                return autotest::lua_eval('run(p1)')
            }
            EOM
```

autotest.pm
perlpunk: I made a complete suggestion now below

autotest.pm
```diff
@@ -22,7 +22,7 @@ use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use Mojo::File qw(path);
 use File::Glob qw(bsd_glob);
 
-our @EXPORT_OK = qw(loadtest $selected_console $last_milestone_console query_isotovideo);
+our @EXPORT_OK = qw(loadtest $selected_console $last_milestone_console query_isotovideo _openqa_require_lua);
```
Martchus, line 25: If the function is exported it should probably not start with an underscore as this is usually only used for module-internal functions.

autotest.pm
```diff
@@ -108,6 +108,65 @@ sub _debug_python_version () {
     bmwqemu::diag "Using python version " . $debug;
 }
 
+
+=head2 _openqa_require_lua
+
+    use("testapi");
+    use("some::perl::module", {"some_function", "another_function"});
+
+This function is the importer for loading perl modules from Lua test modules.
```
Martchus, line 117:
```suggestion
This function is the importer for loading Perl modules from Lua test modules.
```

autotest.pm
```diff
@@ -108,6 +108,65 @@ sub _debug_python_version () {
...
+It imports everything that a perl module exports into the global Lua scope.
```
Martchus, line 118:
```suggestion
It imports everything that a Perl module exports into the global Lua scope.
```

autotest.pm
```diff
@@ -108,6 +108,65 @@ sub _debug_python_version () {
...
+            $fn =~ s/\$//;
+        }
+        elsif ($fn =~ m/^\%/) {
+            $pref = '\%';
+            $fn =~ s/\%//;
+        }
+        elsif ($fn =~ m/^\@/) {
+            $pref = '\@';
+            $fn =~ s/\@//;
+        }
```
Martchus, line 141: It would be nice to have a comment explaining why this code section does what it does.

autotest.pm
```diff
@@ -108,6 +108,65 @@ sub _debug_python_version () {
...
+    end
+    function lua_append_path(path)
+        package.path = path .. ";" .. package.path
+        print(package.path)
+    end
+    EOLUA
+    return lua_eval('_VERSION', 0);
+    EOM
+    my $debug = eval $code;
+    bmwqemu::diag "Using lua version " . $debug;
```
Martchus, line 166:
```suggestion
    bmwqemu::diag "Using Lua version $debug";
```

autotest.pm
```diff
@@ -148,6 +208,33 @@ sub loadtest ($script, %args) {
             EOM
         $is_python = 1;
     }
+    elsif ($script =~ m/\.lua$/) {
+        _debug_lua_version();    # this loads (actually defines) lua_eval()
```
Martchus, line 212: This function should have a better name so the comment isn't required.

autotest.pm
```diff
@@ -148,6 +208,33 @@ sub loadtest ($script, %args) {
...
+        my $script_dir = path(File::Basename::dirname($script_path))->to_abs;
```
Martchus, line 213: Probably the following works as well:
```suggestion
        my $script_dir = path($script_path)->dirname->to_abs;
```

dist/rpm/os-autoinst.spec
```diff
@@ -87,6 +87,11 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %define python_support_requires %{nil}
 %endif
+# The following line is generated from dependencies.yaml
+%define lua_support_requires perl(Inline::Lua)
+%else
+%define lua_support_requires %{nil}
+%endif
```
Martchus, line 94: This seems to miss an `$if %{with lua_support}` and of course also a preceding `%bcond_…` macro. I think this should be done like we already conditionally enable Python support.

t/data/tests/tests/boot.pm
```diff
@@ -1,19 +0,0 @@
...
-    send_key 'ret';
-
-    # set timeout to 10 minutes so we can't miss the situation when we're waiting for the assert_screen to timeout
-    # (test uses 'Skip timeout' so this won't actually delay the test execution)
-    assert_screen 'on_prompt', timeout => get_var('TESTING_ASSERT_SCREEN_TIMEOUT') ? 600 : 90;
-}
-
-sub test_flags ($) { {} }
-
-1;
```
Martchus, line 19: Don't we still need this when we want to run tests when lua support is disabled?

t/data/tests/main.pm
```diff
@@ -34,7 +34,7 @@ if (!$integration_tests && eval { require Inline::Python }) {
     autotest::loadtest "tests/pre_boot.py";
 }
 
-autotest::loadtest "tests/boot.pm";
```
Martchus, line 37: The Lua version should probably be loaded only if Lua support is enabled (when `Inline::Lua` is available).

t/data/tests/tests/boot.lua
```diff
@@ -0,0 +1,24 @@
+use("testapi")
+use("testlib")
+use("testlib", {"testfunc2"})
+local luatestlib = require 'luatestlib' -- load luatestlib.lua from ../lib/ dir
+
+function run(self)
+    testfunc1()
+    testfunc2()
+    luatestlib.testfunc3()
```
Martchus, line 9: We should probably check whether these functions are called correctly, e.g. check their return values.

t/data/tests/lib/testlib.pm
```diff
@@ -0,0 +1,18 @@
+package testlib;
```
Martchus, line 1: Maybe add a comment that the purpose of this module is to be imported by other languages to test cross-language importing.

autotest.pm
```diff
@@ -108,6 +108,65 @@ sub _debug_python_version () {
...
+            $pref = '\%';
+            $fn =~ s/\%//;
+        }
+        elsif ($fn =~ m/^\@/) {
+            $pref = '\@';
+            $fn =~ s/\@//;
+        }
+        lua_set($fn, eval("${pref}${perlmod}::${fn}"));
+    }
+}
```
Wabri, line 144: Could it make sense to replace the if-else into a given-when statement?

t/data/tests/tests/boot.pm
asdil12: I just wouldn't support that.

autotest.pm
asdil12: Seems to be deprecated:
```
given is deprecated at /home/dheidler/openqa/repos/os-autoinst/autotest.pm line 129.
when is deprecated at /home/dheidler/openqa/repos/os-autoinst/autotest.pm line 130.
when is deprecated at /home/dheidler/openqa/repos/os-autoinst/autotest.pm line 131.
when is deprecated at /home/dheidler/openqa/repos/os-autoinst/autotest.pm line 132.
```

Martchus (approved):

okurz (changes requested):
Missing patch coverage, failing checks, rest is fine

autotest.pm
```diff
@@ -109,6 +109,58 @@ sub _debug_python_version () {
     bmwqemu::diag "Using python version " . $debug;
 }
 
+
+=head2 _lua_use
+
+    use("testapi");
+    use("some::perl::module", {"some_function", "another_function"});
+
+This function is the importer for loading Perl modules from Lua test modules.
```
Martchus, line 118:
```suggestion
This function loads Perl modules from Lua test modules.
```

t/08-autotest.t
```diff
@@ -486,6 +486,22 @@ subtest 'start_process' => sub {
...
 
+subtest 'lua_use' => sub {
+    my $lua_vars = {};
+    $mock_autotest->redefine('lua_set' => sub ($k, $v) { $lua_vars->{$k} = $v; });
+    autotest::_lua_use('testapi');
+    is $lua_vars->{realname}, 'Bernhard M. Wiedemann', 'Check importing strings';
+    is ref($lua_vars->{assert_script_run}), 'CODE', 'Check importing functions';
+
+    use lib './t/data/tests/lib';
+    use testlib;
```
Martchus, line 497: These lines are still executed at compile time before the preceding lines of code are executed. Hence I suggest you move them at the top where they belong to avoid giving a false impression of when the lines are executed.

okurz (commented):
LGTM, @perlpunk WDYT?

t/08-autotest.t
```diff
@@ -486,6 +486,22 @@ subtest 'start_process' => sub {
...
+    is ref($lua_vars->{assert_script_run}), 'CODE', 'Check importing functions';
+
+    autotest::_lua_use('testlib');
+    is $lua_vars->{testfunc1}(), 42, 'Exported function is imported';
+    is $lua_vars->{testfunc2}, undef;
+    autotest::_lua_use('testlib', ['testfunc2']);
+    is $lua_vars->{testfunc2}(), 43, 'Explicit imports of non-exported functions';
+    is_deeply $lua_vars->{testarray}, [1, 2, 3], 'Import Array';
+    is_deeply $lua_vars->{testhash}, {foo => 'bar'}, 'Import Hash';
+};
```
perlpunk, line 503: I ran the tests and noticed that the `pre_boot_lua` module is never tested.

```
$autotest::tests{'tests-pre_boot_lua'}->runtests
```
seems to work basically, if I comment out the `check_screen` and `send_key`.

autotest.pm
```diff
@@ -109,6 +109,58 @@ sub _debug_python_version () {
...
+sub _lua_use ($perlmod, $import = undef) {
+    my $inf = (defined $import) ? " qw(@$import)" : '';
+    eval("use ${perlmod}${inf}");
+    my @me = (defined $import) ? @{$import} : eval("\@{${perlmod}::EXPORT}");
+    foreach my $fn (@me) {
+        my $pref = '\&';    # import functions
+        if ($fn =~ m/^\$/) { $pref = '$'; $fn =~ s/\$//; }    # import scalars
+        elsif ($fn =~ m/^\%/) { $pref = '\%'; $fn =~ s/\%//; }    # import hashes
+        elsif ($fn =~ m/^\@/) { $pref = '\@'; $fn =~ s/\@//; }    # import arrays
+        lua_set($fn, eval("${pref}${perlmod}::${fn}"));
```
perlpunk, line 136: It should be noted that imported variables won't work as expected.
If you import a variable in perl, then the imported variable is the same as the original one, meaning that if you assign a new value in one place, it's visible in the other place.

Here only the initial value will be set in the lua variables.

If we want to provide access to testapi variables, we should probably rather write wrapper functions, e.g. for `$realname`, `$user` and `$password`.

autotest.pm
```diff
@@ -149,6 +202,33 @@ sub loadtest ($script, %args) {
...
+        $code .= sprintf <<~'EOM', $script_dir, $script_path;
+            use base 'basetest';
+
+            use autotest;
+            autotest::_load_lua();
+            autotest::lua_append_path("%s/../lib/?.lua");
+            autotest::lua_eval('dofile("%s")');
+
+            sub run($self) {
+                # not passing $self as blessed objects generate a warning
```
perlpunk, line 217: I can imagine a workaround for passing the basetest object.
And since the basetest is practically a singleton at the time of the test run, we could provide a proxy class in lua which calls perl functions to access the original perl object, stored in a global variable.

perlpunk (changes requested):
See comments

autotest.pm
asdil12: The idea is to eventually get rid of those vars alltogether.
But at least the initial state is available to at least provide something.

t/08-autotest.t
asdil12: Okay - the `08-autotest.t` test should now run a lua module.

autotest.pm
```diff
@@ -109,6 +109,58 @@ sub _debug_python_version () {
...
+
+Imported hashes, scalars and arrays are copied as Inline:Lua doesn't support
+references for them.
+
+=cut
+
+sub _lua_use ($perlmod, $import = undef) {
+    my $inf = (defined $import) ? " qw(@$import)" : '';
+    eval("use ${perlmod}${inf}");
+    my @me = (defined $import) ? @{$import} : eval("\@{${perlmod}::EXPORT}");
```
perlpunk, line 130: This is not checking if a function is actually exported. It either must be in @EXPORT_OK or `@EXPORT`.
Also the variable name could be improved :)
Here is my suggestion to improve `_lua_use` to avoid the below `eval`, to check imports, and to die if the perl module couldn't be loaded.
it also tests if imported arrays and hashes work.
```diff
diff --git a/autotest.pm b/autotest.pm
index a5a2fd05..a40f1800 100644
--- a/autotest.pm
+++ b/autotest.pm
@@ -125,15 +125,32 @@ references for them.
 =cut
 
 sub _lua_use ($perlmod, $import = undef) {
-    my $inf = (defined $import) ? " qw(@$import)" : '';
-    eval("use ${perlmod}${inf}");
-    my @me = (defined $import) ? @{$import} : eval("\@{${perlmod}::EXPORT}");
-    foreach my $fn (@me) {
-        my $pref = '\&';    # import functions
-        if ($fn =~ m/^\$/) { $pref = '$'; $fn =~ s/\$//; }    # import scalars
-        elsif ($fn =~ m/^\%/) { $pref = '\%'; $fn =~ s/\%//; }    # import hashes
-        elsif ($fn =~ m/^\@/) { $pref = '\@'; $fn =~ s/\@//; }    # import arrays
-        lua_set($fn, eval("${pref}${perlmod}::${fn}"));
+    eval "use $perlmod ()";
+    if ($@) { die "Could not load '$perlmod': $@" }
+    my %exports;
+    no strict 'refs';
+    my @export = @{"${perlmod}::EXPORT"};
+    my @export_ok = @{"${perlmod}::EXPORT_OK"};
+    @exports{@export, @export_ok} = ();
+    my @to_import;
+    if ($import) {
+        for my $symbol (@$import) {
+            if (exists $exports{$symbol}) {
+                push @to_import, $symbol;
+            }
+            else {
+                warn "'$symbol' is not exported by '$perlmod'";
+            }
+        }
+    }
+    else {
+        @to_import = @export;
+    }
+    foreach my $symbol (@to_import) {
+        if ($symbol =~ s/^\$//) { lua_set($symbol, ${"${perlmod}::${symbol}"}) }
+        elsif ($symbol =~ s/^\%//) { lua_set($symbol, \%{"${perlmod}::${symbol}"}) }
+        elsif ($symbol =~ s/^\@//) { lua_set($symbol, \@{"${perlmod}::${symbol}"}) }
+        else { lua_set($symbol, \&{"${perlmod}::${symbol}"}) }
     }
 }
 
diff --git a/t/08-autotest.t b/t/08-autotest.t
index 9a1136b2..2637d47c 100755
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -507,7 +507,10 @@ subtest 'lua_runtest' => sub {
     plan skip_all => 'Inline::Lua is not available' unless $has_lua;
 
     my $luatest = $autotest::tests{'tests-unittest_lua'};
-    combined_like { $luatest->runtest() } qr{testfunc1\ntestfunc2\ntestfunc3}, 'output from lua as expected';
+    my $out = combined_from { $luatest->runtest() };
+    like $out, qr{testfunc1\ntestfunc2\ntestfunc3}, 'function calls work';
+    like $out, qr{testarray:\t1,2,3}, 'arrays work';
+    like $out, qr{foo = bar}, 'hashes work';
 };
 
 
diff --git a/t/data/tests/lib/testlib.pm b/t/data/tests/lib/testlib.pm
index 136f78cc..54b657cd 100644
--- a/t/data/tests/lib/testlib.pm
+++ b/t/data/tests/lib/testlib.pm
@@ -5,6 +5,7 @@ package testlib;
 use Mojo::Base 'Exporter', -signatures;
 
 our @EXPORT = qw(testfunc1 @testarray %testhash);
+our @EXPORT_OK = qw(testfunc2);
 
 our @testarray = (1, 2, 3);
 our %testhash = (foo => 'bar');
diff --git a/t/data/tests/tests/unittest_lua.lua b/t/data/tests/tests/unittest_lua.lua
index ae550784..82dc3cd9 100644
--- a/t/data/tests/tests/unittest_lua.lua
+++ b/t/data/tests/tests/unittest_lua.lua
@@ -2,11 +2,16 @@
 
 use("testapi")
 use("testlib")
-use("testlib", {"testfunc2"})
+use("testlib", {"testfunc2","@testarray","%testhash"})
 require 'luatestlib' -- load luatestlib.lua from ../lib/ dir
 
 function run(self)
     assert(testfunc1() == 42)
     assert(testfunc2() == 43)
     assert(luatestlib.testfunc3() == 44)
+    print("testarray:", table.concat(testarray,","))
+    print("testhash:", table.concat(testhash,","))
+    for k,v in pairs(testhash) do
+        print(k.." = "..v)
+    end
 end
```

asdil12: Thanks - that seems to work fine. Applied.

t/08-autotest.t
```diff
@@ -451,7 +451,7 @@ subtest loadtestdir => sub {
     stderr_like {
         autotest::loadtestdir('tests');
     } qr/debug.*scheduling/, 'loadtestdir is scheduling successfully';
-    ok exists $autotest::tests{'tests-boot'}, 'boot.pm loaded';
```
Martchus, line 454: Are we losing test coverage for Python here?

t/data/tests/tests/pre_boot_lua.lua
```diff
@@ -0,0 +1,23 @@
...
+    -- More sophisticated example call
+    --assert_screen('on_prompt', 'timeout', get_var('TESTING_ASSERT_SCREEN_TIMEOUT') and 600 or 90)
+end
+
+--function test_flags(self)
+--    return {fatal = 1}
+--end
+
+--function post_fail_hook(self)
+--end
```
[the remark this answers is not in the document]
asdil12: Sure - they document how to set test flags.

t/08-autotest.t
asdil12: Why for python?

Martchus (approved):

okurz (approved):

container/os-autoinst_dev/Dockerfile
```diff
@@ -97,6 +97,7 @@ RUN zypper in -y -C \
        'perl(IPC::Open3)' \
        'perl(IPC::Run::Debug)' \
        'perl(IPC::System::Simple)' \
+       'perl(Inline::Lua)' \
```
perlpunk, line 100: Will this work if perl-Inline-Lua is only in devel:languages:perl at this point?

okurz (approved):

okurz:
@perlpunk ready now?

container/os-autoinst_dev/Dockerfile
aylen384: I created a comment

perlpunk: The comment is not according to the description in https://build.opensuse.org/project/show/devel:openQA:Leap:15.6

aylen384: Updated the comment

perlpunk: I fixed the factory link with:
```
% osc changedevelrequest openSUSE:Factory perl-Inline-Lua devel:languages:perl perl-Inline-Lua -m "assign a devel project"
1282063
% osc linkpac openSUSE:Factory perl-Inline-Lua devel:languages:perl perl-Inline-Lua -f
Creating _link... Done.
```
https://build.opensuse.org/requests/1282063

PR #2701, opened 2025-05-06
Wrap prove to prevent unhandled output

Wabri:
Reference: https://progress.opensuse.org/issues/181358


Martchus:
```
/opt/tools/invoke-tests: line 87: /home/runner/work/os-autoinst/os-autoinst/tools/prove_wrapper: No such file or directory
```

Looks like the use of `$PWD` isn't correct here (but `$(pwd)` likely wouldn't have worked, too). Maybe you can just completely hardcode the path which would probably have to be `/opt/tools/prove_wrapper`. This would be fine at this level and maybe even the clearest solution.

Wabri:
I'm not completely happy by put an hardcoded path, but I think we can fix that later.

Martchus:
The same script where you hardcode the path now is also already hardcoding the `/opt` part/prefix. You could put this prefix into a variable to avoid repeating it - although I don't find repeating it bad.

And it works now - except that it has actually found unhandled output.

tools/invoke-tests
```diff
@@ -36,6 +36,9 @@ while [[ $# -gt 0 ]]; do
             ;;
         --prove-tool)
             prove_path=$2
+            echo "-----------------------------"
+            echo "Using prove tool: $prove_path"
+            echo "-----------------------------"
```
Martchus, line 41: This can probably be removed again and CMake logs this anyway via ` 3: Test command: /opt/tools/invoke-tests "--prove-tool" "/opt/tools/prove_wrapper" "--make-tool" "/usr/bin/ninja" "--unbuffer-tool" "UNBUFFER_PATH-NOTFOUND" "--build-directory" "/opt/build"`.

Wabri: Yep, I didn't want to be permanent, but I was using only for debugging purpose :+1:

t/18-backend-qemu.t
```diff
@@ -89,11 +89,11 @@ subtest 'using Open vSwitch D-Bus service' => sub {
 
 my $fake_qmp_answer;
 $backend_mock->redefine(handle_qmp_command => sub { push @{$called{handle_qmp_command}}, $_[1]; $fake_qmp_answer });
-$backend->power({action => 'off'});
+combined_like { $backend->power({action => 'off'}); } qr/[debug]/s, 'Debug message logged for power off';
```
b10n1k, line 92: I think it would be good idea to check part of the debug message! not sure how can you be sure that it is the expected bebug message. I guess is the only one in the `power` but it make sense to check for something more specific. like `POWER.*off` or `expected_shutdown`. wdyt?

tools/prove_wrapper
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+sed -E '
+  s/^\[[0-9]{2}:[0-9]{2}:[0-9]{2}\][[:space:]]*//
+  /^All tests successful\./,$d
+' "$OUTPUT" > "$OUTPUT.processed"
+
+UNHANDLED=$(grep -vE '^t\/.*([0-9]{2}-|)([[:alpha:]]|_|-)+\.t \.+' "$OUTPUT.processed" || true)
+
+echo "---"
```
b10n1k, line 24: is that intentional? for debugging? I am not sure.

Wabri: The PR is on WIP so every strange echo or output print is indeed for debugging reasons. I'll remove everything before set the pr ready for reviews

t/18-backend-qemu.t
Wabri: You are right, this is the unhandled output:

```
3: [2025-05-13T15:48:59.157378Z] [debug] [pid:2653] POWER: action: off, expected_shutdown: 1
3: [2025-05-13T15:48:59.157950Z] [debug] [pid:2653] POWER: action: acpi, expected_shutdown: 1
3: [2025-05-13T15:49:01.013982Z] [debug] [pid:2653] POWER: action: off, expected_shutdown: 1
3: [15:49:02] t/18-backend-qemu.t ......................... ok     4538 ms ( 0.04 usr  0.00 sys +  3.72 cusr  0.81 csys =  4.57 CPU)
```

I'll add a proper string :)

tools/invoke-tests
```diff
@@ -64,6 +64,9 @@ db="$build_directory/cover_db"
 
 # set Perl module include path and coverage options
 export PERL5LIB="$source_directory:$source_directory/ppmclibs/blib/lib:$source_directory/ppmclibs/blib/arch/auto/tinycv:$PERL5LIB"
+if [[ "${GITHUB_ACTIONS}" == "true" ]]; then
+    cpanm Devel::Cover
```
Martchus, line 68: I would avoid installing dependencies at this level. We should rather ensure the required package is available in the container. I'm wondering why this is suddenly required, though.

Wabri: In order to let the unhandled output be clear we need to update the Devel::Cover module and I'm aware that is not the proper way to solve the issue, but is a mitigation I've apply. As I said during the daily I'll open a new ticket to proper handle this because is not in the scope of this.

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Static checks, unit- and integration tests
+      - name: Static checks, unit and integration tests
```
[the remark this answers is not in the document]
Wabri: I've miss the plural, but still for me `unit-` not make sense and prefer the suggestion you gave here.

tools/prove_wrapper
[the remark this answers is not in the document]
okurz: I just want to avoid duplication. This is not in conflict with the ACs of the ticket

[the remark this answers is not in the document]
Wabri: I've done a temporary change in order to wait the pr merged in the common repository https://github.com/os-autoinst/os-autoinst-common/pull/57

Wabri:
I've done a temporary change in order to wait the pr merged in the common repository https://github.com/os-autoinst/os-autoinst-common/pull/57

Wabri:
Since the `prove_wrapper` script is being use in both os-autoinst and openqa repository I'm moving that into the common repository using this pr os-autoinst/os-autoinst-common#57 as said in https://github.com/os-autoinst/os-autoinst/pull/2701#discussion_r2089145960

---

I'll update the pr once the pr is merged!

tools/invoke-tests
perlpunk: Ok, so now we have a comment hidden in in a PR review and a ticket in future. How are we going to be reminded of this? :)
The ticket should be added in the code, and it should be at least in our Next queue.

kalikiana:
```suggestion
    # https://progress.opensuse.org/issues/182402
    cpanm Devel::Cover
```

okurz: https://progress.opensuse.org/issues/182402 is now "High" and in our backlog

Wabri: Updated with the reference to the ticket

PR #2703, opened 2025-05-15
Another round of changes for 4K resolution on OrangePi5+

marmarek:
Here comes initial position scaling idea I shared before. And few fixups to ustreamer call, including RGB format handling fix (relate to hardware/driver bug https://github.com/raspberrypi/linux/issues/6068). Turns out using that hardware for the frame format reference wasn't the best idea... This time I checked with hexeditor if the bytes order is correct in the test.


okurz (approved):

marmarek:
CI not happy:
> Package 'perl-YAML-LibYAML-0.903.0' not found.

marmarek:
> Here comes initial position scaling idea I shared before.

BTW, the forced occasional full screen search (every 5 frames?) takes a lot more time with 4K frames. In my setup, normal search is more or less instant, while the full search triggers a warning like this one: `check_asserted_screen took 2.56 seconds for 3 candidate needles`. Would it make sense to add also a max margin to search? A global one? Or even per-needle?

ppmclibs/tinycv_impl.cc
```diff
@@ -187,10 +187,13 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
     }
 
     // Optimization -- Search close to the original area working with ROI
-    int scene_x = std::max(0, int(x - margin));
-    int scene_y = std::max(0, int(y - margin));
-    int scene_bottom_x = std::min(scene->img.cols, int(x + width + margin));
-    int scene_bottom_y = std::min(scene->img.rows, int(y + height + margin));
+    // Scale the possition if object and scene have different sizes
+    int scaled_x = x * scene->img.cols / object->img.cols;
+    int scaled_y = y * scene->img.rows / object->img.rows;
```
Martchus, line 192: In theory it would make sense to avoid the implicit narrowing conversion here:
```suggestion
    long scaled_x = x * scene->img.cols / object->img.cols;
    long scaled_y = y * scene->img.rows / object->img.rows;
```
(In practice it likely doesn't matter because we'll run into an implicit narrowing conversion later anyway and there are also many explicit narrowing conversions.)

Martchus (approved):
I re-triggered the fullstack test. The change itself looks good.

PR #2706, opened 2025-05-17
Try to increase pipe size for external video encoder

marmarek:
With 1024x768 frame, a single frame takes 2.3MB, which is bigger than
the default pipe size (64KB). The need to write in a couple iterations
(with waiting until the encoder retrieves the data in between) has little
performance impact in that case, and queing those frames takes limited
amout of memory. But with 3840x2160 frame, the difference is a lot
bigger (24MB > 64KB) and the impact is also significantly bigger. Slower
sending frames for encoding leads to a much higher memory consumption
and in some cases to the worker beeing killed by OOM-killer.

The default value for /proc/sys/fs/pipe-max-size is 1MB which is not
enough to increase even for 1024x768 frames. With 3840x2160 frames,
/proc/sys/fs/pipe-user-pages-soft limit can be reached too (64MB by
default, but applied to all pipes of a user combined). So, without
increasing global limits, increasing pipe size will fail.
Try to increase the pipe size, but when it fails, just emit a warning
(with suggested solution) instead of crashing.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -343,6 +343,16 @@ sub check_select_rate ($buckets, $wait_time_limit, $hits_limit, $id, $time) {
 sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     my $pid = open($self->{$pipe_name}, '|-', @cmd);
     my $pipe = $self->{$pipe_name};
+    # Try to make pipe big enough to fit full frame at once, instead of sending
+    # it in several chunks. Limit it to custom resolutions, as with the default
+    # the impact of sending it in several chunks is not too bad, and default
+    # Linux's pipe-max-size is too small for 2.3MB size (so it wouldn't work
+    # without raising the limit anyway).
+    if ($bmwqemu::vars{XRES} && $bmwqemu::vars{YRES}) {
+        my $pipe_size = $bmwqemu::vars{XRES} * $bmwqemu::vars{YRES} * 3;
+        no autodie;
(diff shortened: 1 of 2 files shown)


backend/baseclass.pm
```diff
@@ -343,6 +343,11 @@ sub check_select_rate ($buckets, $wait_time_limit, $hits_limit, $id, $time) {
 sub _invoke_video_encoder ($self, $pipe_name, $display_name, @cmd) {
     my $pid = open($self->{$pipe_name}, '|-', @cmd);
     my $pipe = $self->{$pipe_name};
+    if ($bmwqemu::vars{XRES} && $bmwqemu::vars{YRES}) {
```
okurz, line 346: Could you put a part of your explanation from the commit message in here as a comment to explain the reasoning for the code?

okurz (changes requested):

backend/baseclass.pm
marmarek: Sure. I also wonder if it's a good idea to try increasing the pipe size even for default 1024x768? Right now I intentionally limited it to non-default resolutions only (to not emit the warning in most cases), but maybe it makes sense in any case? I haven't tried if improvement on 1024x768 would be noticeable at all, but in theory it may help too.

okurz: I would be very much interested if you can do any benchmarking showing any influence?

marmarek: It looks like I don't have system slow enough for this to be a bottleneck with 1024x768 resolution. On my RPi setup both with and without this change the `Passing remaining frames to the video encoder` stage takes about 3s.

okurz (approved):

kalikiana (approved):
Nice!

PR #2711, opened 2025-05-23
Apply macro to support upcoming opensuse/sle 16 build

b10n1k:
As per
https://en.opensuse.org/openSUSE:Packaging_for_Leap#RPM_Distro_Version_Macros this commit substitutes sle_version to suse_version macro because the sle_version does not exist in versions with code 16 like Leap 16.0. `suse_version` seems to provide backwards compatibility with Code 15 so it is as simple as that. Alternative it would have to check whether sle_version is not defined and adjust the conditions.


perlpunk:
Seems like the https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2711 is missing completely.

asdil12 (approved):

okurz:
This for sure needs package-level verification

dist/rpm/os-autoinst.spec
```diff
@@ -25,14 +25,14 @@ Group:          Development/Tools/Other
 Url:            https://github.com/os-autoinst/os-autoinst
 Source0:        %{name}-%{version}.tar.xz
 %{perl_requires}
-%if 0%{?suse_version} > 1500 || 0%{?sle_version} >= 150400
+%if 0%{?suse_version} >= 1540
```
baierjan, line 28: Leap 15.4 is out of support, maybe we can simplify this condition?

dist/rpm/os-autoinst.spec
```diff
@@ -59,7 +59,7 @@ Source0:        %{name}-%{version}.tar.xz
 %else
 %define spellcheck_requires %{nil}
 %endif
-%if 0%{?sle_version} < 150200 && !0%{?is_opensuse}
+%if 0%{?suse_version} < 1520 && !0%{?is_opensuse}
```
baierjan, line 62: 15-SP2 should be already out of support, maybe we can get rid of this block entirely?

baierjan (commented):
Maybe this is a good opportunity to clean-up the spec file a little and drop support for products which reached EoL?

dist/rpm/os-autoinst.spec
```diff
@@ -25,14 +25,9 @@ Group:          Development/Tools/Other
...
 %{perl_requires}
-%if 0%{?suse_version} > 1500 || 0%{?sle_version} >= 150400
-# openSUSE Tumbleweed and Leap 15.4
 %define opencv_require pkgconfig(opencv4)
-%else
-%define opencv_require pkgconfig(opencv)
-%endif
 # exclude additional sub packages that would pull in a lot of extra dependencies on SLE
-%if 0%{?sle_version} && !0%{?is_opensuse}
+%if 0%{?suse_version} && !0%{?is_opensuse}
```
baierjan, line 30: I am not really sure about this condition. `sle_version` is not defined on Tumbleweed, `suse_version` is. Please double check if this condition does what we are expecting.

baierjan (commented):
One clarification, otherwise good.

dist/rpm/os-autoinst.spec
asdil12:
```
[dheidler@midas ~]$ podman run --rm -it opensuse/leap:15.6 rpm --eval '%if 0%{?suse_version} && !0%{?is_opensuse}'
%if 01500 && !0
[dheidler@midas ~]$ podman run --rm -it opensuse/tumbleweed rpm --eval '%if 0%{?suse_version} && !0%{?is_opensuse}'
%if 01699 && !01

baierjan: Yeah, the condition is TRUE for Leap 15.6 which I am not sure if is the desired state (considering the comment mentions excluding some packages on SLE).

b10n1k: this conversion is quite straight forward IMO. the only concern is the confusing `is_opensuse` which I am told _that %is_opensuse is not defined because that macro is defined via Project Config in the build service.` i guess that's why might return 0, even so it is represented as `1` in 
https://en.opensuse.org/openSUSE:Packaging_for_Leap#RPM_Distro_Version_Macros

b10n1k:
> Still have some concerns, I hope it was tested.

@baierjan I checked the builds, current and old. I can suggest to sit together to check a few things if you need to, to unblock it

baierjan:
> > Still have some concerns, I hope it was tested.
> 
> @baierjan I checked the builds, current and old. I can suggest to sit together to check a few things if you need to, to unblock it

As you can already see, I am no longer blocking this PR 😉

b10n1k:
> > > Still have some concerns, I hope it was tested.
> > 
> > 
> > @baierjan I checked the builds, current and old. I can suggest to sit together to check a few things if you need to, to unblock it
> 
> As you can already see, I am no longer blocking this PR 😉

I see but it would be nice to review it together and have your approval

b10n1k:
> This for sure needs package-level verification

Maybe I misunderstand but there are results in https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2711/os-autoinst and as far as I checked (and others did too i think) it does not break for the projects before Code 16. the only thing I can think of is build it and install it local to see if it works

okurz:
> > This for sure needs package-level verification
> 
> Maybe I misunderstand but there are results in https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2711/os-autoinst and as far as I checked (and others did too i think) it does not break for the projects before Code 16. the only thing I can think of is build it and install it local to see if it works

Where is the verification for 16 then. And there are unresolvables in that project. Did you check them?

PR #2712, opened 2025-05-24
Add wrappers for testapi variables

perlpunk:
Issue: https://progress.opensuse.org/issues/182708


testapi.pm
```diff
@@ -2250,4 +2257,59 @@ sub backend_get_wait_still_screen_on_here_doc_input () {
     return get_var(_WAIT_STILL_SCREEN_ON_HERE_DOC_INPUT => $ret);
 }
 
+=head2 get_username, set_username
+
+    my $un = get_username;
+    set_username($new_name);
+
+Wrapper around the exported C<$username> variable. Direct user of the variable
```
okurz, line 2265: "Direct user"?

okurz (changes requested):

testapi.pm
perlpunk: I meant "direct use". But I removed that now.
Not sure if/when we want to deprecate it at all.

t/03-testapi.t
```diff
@@ -1286,6 +1286,14 @@ subtest 'ensure_installed test' => sub {
     ensure_installed(['openQA', 'os-autoinst-devel']);
 };
 
+subtest 'retrieve and change various global variables' => sub {
+    my $new = 'new value';
+    for my $func (qw(username realname password serialdev)) {
+        testapi->can("set_$func")->($new);
+        is(testapi->can("get_$func")->(), $new, "get/set $func works");
```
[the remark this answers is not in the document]
perlpunk: Could be. I added extra tests.

perlpunk:
Needs
* https://github.com/os-autoinst/openQA/pull/6480

Martchus (approved):

testapi.pm
```diff
@@ -2250,4 +2257,56 @@ sub backend_get_wait_still_screen_on_here_doc_input () {
...
+    my $old = get_serialdev;
+    set_serialdev($new);
+
+Wrapper around the exported C<$serialdev> variable.
+
+=cut
+
+sub get_serialdev () { $serialdev }
+
+sub set_serialdev ($new) { $serialdev = $new }
```
alvarocarvajald, line 2310: This broke one of the pipelines in os-autoinst-distri-opensuse: https://github.com/os-autoinst/os-autoinst-distri-opensuse/actions/runs/16780759661/job/47518464960

But not sure what would be the proper way to fix between renaming the method here, or in `lib/proxymode.pm`, or avoid importing everything from testapi in `lib/proxymode.pm` or filtering the error in the compile check

perlpunk: ok, in `lib/proxymode.pm` I see:
```
sub set_serialdev {                                                             
    $serialdev = $SLAVE_SERIALDEV;                                              
}     
```
so basically this function is doing what `testapi::set_serialdev` is now doing, so it can be removed...

alvarocarvajald: Oh, right. I'll wait until tomorrow to see if the Virtualization Squad responds, otherwise I'll submit a fix. Thanks!

PR #2717, opened 2025-06-07
Check for available space before copying file

realcharmer:

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -384,7 +384,14 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
       'done;' .
       'else ' .
+      "REQUIRED_SPACE=\$(du -b /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename | awk '{print \$1}'); " .
+      "AVAILABLE_SPACE=\$(df -B1 $vmware_openqa_datastore | awk 'NR==2 {print \$4}'); " .
+      "if [ \"\$REQUIRED_SPACE\" -gt \"\$AVAILABLE_SPACE\" ]; then " .
+      "echo 'Error: Not enough space on target datastore $vmware_openqa_datastore to copy $file_basename.'; " .
+      "exit 1; " .
+      "else " .
       "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
+      "fi; " .
       'fi;';
     my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer');
     die "Can't copy VMware image $file_basename" if $retval;


consoles/sshVirtsh.pm
```diff
@@ -384,7 +384,14 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
       'done;' .
       'else ' .
+      "REQUIRED_SPACE=\$(du -b /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename | awk '{print \$1}'); " .
+      "AVAILABLE_SPACE=\$(df -B1 $vmware_openqa_datastore | awk 'NR==2 {print \$4}'); " .
+      "if [ \"\$REQUIRED_SPACE\" -gt \"\$AVAILABLE_SPACE\" ]; then " .
```
d3flex: it should be better wirh `-qe` . I am also thinking if it has to have a margin. if the `$REQUIRED_SPACE` take all the space after the operation, the disk will not have space to "breath", right?

realcharmer: I assume you mean `-eq`? That wouldn't make sense.

d3flex:
```arg1 OP arg2
              OP  is one of -eq, -ne, -lt, -le, -gt, or -ge.  These arithmetic binary operators return true if arg1 is equal to, not equal to, less than, less than or equal to, greater than, or greater than or equal to arg2, respectively.  Arg1 and arg2 may be positive
              or negative integers.  When used with the [[ command, Arg1 and Arg2 are evaluated as arithmetic expressions (see ARITHMETIC EVALUATION above). ``` from the man. I meant `greater or equal` op

Comment (realcharmer) [reply consoles/sshVirtsh.pm]: I see, you meant `-ge`, that would make sense when used here.
Regarding the margin, that could be easily implemented with one extra var, but I am not sure that's something that should be handled here. Also what would a good margin be?

Comment (d3flex) [reply consoles/sshVirtsh.pm]: I found this https://thetechylife.com/how-much-space-should-you-leave-on-an-ssd/
not sure if you want to think it over. it is suggested a ~20%. I will leave it to others for taking the final decision about the handling

Comment (realcharmer) [reply consoles/sshVirtsh.pm]: 20% is there to extend the SSD life, irrelevant for the filesystem itself, or the OS. I don't think we should handle this at all. What we are considering here is enough space for the asset, not infrastructure details.

File: consoles/sshVirtsh.pm [hunk 2]
```diff
@@ -384,7 +384,14 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
       'done;' .
       'else ' .
+      "REQUIRED_SPACE=\$(du -b /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename | awk '{print \$1}'); " .
```
Wabri:
```suggestion
      "REQUIRED_SPACE=\$(du -b \"/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename\" | awk '{print \$1}'); " .
```

consoles/sshVirtsh.pm
```diff
@@ -384,7 +384,14 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
...
+      "AVAILABLE_SPACE=\$(df -B1 $vmware_openqa_datastore | awk 'NR==2 {print \$4}'); " .
```
Wabri:
```suggestion
      "AVAILABLE_SPACE=\$(df -B1 \"$vmware_openqa_datastore\" | awk 'NR==2 {print \$4}'); " .
```

Wabri (changes requested):
Variables could contain spaces!

Wabri (approved):

b10n1k:
just one question. did you test it somehow?

consoles/sshVirtsh.pm
```diff
@@ -384,7 +384,14 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
       'done;' .
       'else ' .
+      "REQUIRED_SPACE=\$(du -b \"/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename\" | awk '{print \$1}'); " .
+      "AVAILABLE_SPACE=\$(df -B1 \"$vmware_openqa_datastore\" | awk 'NR==2 {print \$4}'); " .
+      "if [ \"\$REQUIRED_SPACE\" -ge \"\$AVAILABLE_SPACE\" ]; then " .
```
perlpunk: In perl, you can avoid `\"` by using the [qq operator](https://perldoc.perl.org/perlop#Quote-Like-Operators):
```suggestion
      qq{REQUIRED_SPACE=\$(du -b "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename" | awk '{print \$1}'); } .
      qq{AVAILABLE_SPACE=\$(df -B1 "$vmware_openqa_datastore" | awk 'NR==2 {print \$4}'); } .
      qq{if [ "\$REQUIRED_SPACE" -ge "\$AVAILABLE_SPACE" ]; then } .
```

That whole command is getting quite unreadable though.

It seems that the whole command has to be in one line, although I'm not sure if it really does have to, but that could be achieved by using a heredoc and then removing the linebreaks.

Martchus: Or you move the whole script into separate file and pass the paths via CLI arguments or environment variables. Then the script is testable outside the scope of an openQA test. We would then of course need to read this file here to pass its lines as we used to (plus some upfront variable assignments to pass the paths).

consoles/sshVirtsh.pm
```diff
@@ -382,10 +382,17 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       "$ds_debug if test -e $vmware_openqa_datastore$file_basename; then " .
       "while lsof | grep 'cp.*$file_basename'; do " .
       "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
-      'done;' .
-      'else ' .
+      "done;" .
+      "else " .
+      qq{REQUIRED_SPACE=\$(du -b "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename" | awk '{print \$1}'); } .
```
okurz:
```suggestion
      qq{size=\$(stat --printf="%s" "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename"} .
```

consoles/sshVirtsh.pm
```diff
@@ -382,10 +382,17 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
...
+      qq{AVAILABLE_SPACE=\$(df -B1 "$vmware_openqa_datastore" | awk 'NR==2 {print \$4}'); } .
```
okurz:
```suggestion
      qq{available_space=\$(df --output=avail "$vmware_openqa_datastore" | tail -1} .
```

okurz (changes requested):

consoles/sshVirtsh.pm
realcharmer: Wouldn't there be a difference in output? I think `--output=avail` would return the number of blocks, whereas `-B1` gives the number of bytes. We are getting bytes from the file in the line above.

consoles/sshVirtsh.pm
```diff
@@ -381,11 +381,18 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
...
-      'done;' .
-      'else ' .
-      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
-      'fi;';
+      "echo 'File $file_basename is being copied by other process, sleeping for 60 seconds'; sleep 60; " .
+      "done; " .
+      "else " .
+      qq{available_space=\$(df -B1 "$vmware_openqa_datastore" | awk 'NR==2 {print \$4}'); } .
+      qq{size=\$(stat --printf="%s" "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename"); } .
+      qq{if [ "\$size" -ge "\$available_space" ]; then } .
```
perlpunk:
```suggestion
      q{if [ "$size" -ge "$available_space" ]; then } .
```

PR #2719, opened 2025-06-12
Extend logging in OVS service to debug its occasional unresponsiveness

Martchus:
Related ticket: https://progress.opensuse.org/issues/183833

### DIFF
--- a/script/os-autoinst-openvswitch
+++ b/script/os-autoinst-openvswitch
@@ -155,6 +155,7 @@ sub _cmd (@args) {
 
     # We need open3 because otherwise STDERR is captured by systemd.
     # In such way we collect the error and send it back in the dbus call as well.
+    say "Running command: @args";
     my $ovs_vsctl_pid = open3($wtr, $rdr, $err, @args);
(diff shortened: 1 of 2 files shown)


perlpunk:
There is something weird going on with coverage. It's not just this line actually.
The whole `_cmd` subroutine isn't covered anymore with this change (and other code), and it has to do with mocking it in the last subtest somehow.

perlpunk:
btw, I found it confusing that `$mock_main` is called ~$mock_main`and not `$mock_ovs`.
In the first subtest we have
```
my $mock_main = Test::MockModule->new('main');
```
and in the second:
```
my $mock_main = Test::MockModule->new('OVS', no_auto => 1);
```
Maybe you can fix the naming as part of this PR as well?

perlpunk:
I also don't see why it is necessary. Maybe the mocking confuses Devel::Cover in certain situations, e.g. when the package to be mocked is actually part of the main script, or something like that.

perlpunk (approved):

PR #2722, opened 2025-06-16
generalhw: fix is_shutdown signature

marmarek:


Martchus (approved):

perlpunk (commented):
I'm wondering where it is actually called with more than one parameter...

PR #2725, opened 2025-07-02
Autodetect default ethernet dev/br in os-autoinst-setup-multi-machine

aylen384:


perlpunk:
Maybe we can set the defaults later after the `usage()` call

Martchus:
> or just use awk?

No, I find the solution with `jq` much cleaner. Just add it as dependency, it is not a big deal to depend on it.

okurz (approved):
According to `isc se` we don't have a problem with the additional dependencies (which I didn't expect)

script/os-autoinst-setup-multi-machine
```diff
@@ -2,7 +2,8 @@
 set -euo pipefail
 instances="${instances:-20}"
 bridge="${bridge:-"br1"}"
-ethernet="${ethernet:-"br0"}"
+default_ethernet=$(ip -json route show default | jq -r '.[]|.dev')
```
perlpunk, line 5:
```suggestion
default_ethernet=$(ip -json route show default | jq -r '.[] | .dev')
```

perlpunk (approved):

PR #2728, opened 2025-07-10
autotest: properly handle Inline::Lua not being available

AdamWill:
We always try to load a directory containing perl, python and lua tests, which will fail if Inline::Lua or Inline::Python are missing. Instead of skipping the test entirely, let's split the directory into three and test loading the ones we have the bits for.

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -503,7 +503,7 @@ subtest 'start_process' => sub {
 
 subtest 'lua_use' => sub {
     my $lua_vars = {};
-    $mock_autotest->redefine('lua_set' => sub ($k, $v) { $lua_vars->{$k} = $v; });
+    $mock_autotest->mock('lua_set' => sub ($k, $v) { $lua_vars->{$k} = $v; });
     autotest::_lua_use('testapi');
     is $lua_vars->{realname}, 'Bernhard M. Wiedemann', 'Check importing strings';
     is ref($lua_vars->{assert_script_run}), 'CODE', 'Check importing functions';


Martchus:
Skipping tests with only a warning seems dangerous. It could lead to passing test jobs that should be failures simply because a dependency is missing on the worker.

autotest.pm
```diff
@@ -221,6 +222,10 @@ sub loadtest ($script, %args) {
     }
     elsif ($script =~ m/\.lua$/) {
         _load_lua();    # this loads (actually defines) lua_eval()
+        unless (defined &lua_set) {
+            bmwqemu::fctwarn("Inline::Lua not present, cannot load $script");
```
kalikiana: So if I read this correctly tests will just be ignored if Lua is missing. That would make it very dangerous to rely on test coverage in Lua.

kalikiana (changes requested):

AdamWill:
yeah, I was kinda rethinking this after submitting it. the alternative I guess is to have an alternate version of the tests directory with no lua tests in it, and test *that* if Inline::Lua is not available.

okurz:
> yeah, I was kinda rethinking this after submitting it. the alternative I guess is to have an alternate version of the tests directory with no lua tests in it, and test _that_ if Inline::Lua is not available.

Just in comparison: How is the situation for perl-Inline-Python? To me I assume that's rather similar, isn't it?

b10n1k:
I would say it is easier to skip the test for now, based on the comments this PR gets so far

AdamWill:
fullstack failure isn't my fault, looks like some package versions changed or something.

Martchus (approved):

kalikiana (approved):

PR #2731, opened 2025-07-29
Multiple minor backend code improvements

okurz:
See individual commits

Related progress issue: https://progress.opensuse.org/issues/186380

### DIFF
--- a/backend/svirt.pm
+++ b/backend/svirt.pm
@@ -223,6 +223,18 @@ sub start_serial_grab ($self, $name) {
     $ssh->blocking(0);
 }
 
+# Hyper-V does not support serial console export via TCP, just
+# windows named pipes (e.g. \\.\pipe\mypipe). Such a named pipe
+# has to be enabled by a namedpipe-to-TCP on HYPERV_SERVER application.
+sub serial_console_cmd_hyperv ($, $, $port) { 'socat - TCP4:' . $bmwqemu::vars{HYPERV_SERVER} . ':' . $port . ',crnl' }
+
+# libvirt esx driver does not support `virsh console', so
+# we have to connect to VM's serial port via TCP which is
+# provided by ESXi server.
+sub serial_console_cmd_vmware ($, $, $port) { 'socat - TCP4:' . $bmwqemu::vars{VMWARE_HOST} . ':' . $port . ',crnl' }
+
+sub serial_console_cmd_svirt ($name, $devname, $port) { "virsh console $name $devname$port" }
+
 =head2 open_serial_console_via_ssh
 
   ($ssh, $chan) = open_serial_console_via_ssh($name[, port => ''][, devname => ''])
@@ -239,40 +251,25 @@ C<$args{devname}> used device name
 
 sub open_serial_console_via_ssh ($self, $name, %args) {
     bmwqemu::log_call(name => $name, %args);
-    my ($chan, $cmd, $cmd_full, $ret, $ssh, $stderr, $stdout);
     my $port = $args{port} // '';
     my $devname = $args{devname} // '';
     my $marker = "CONSOLE_EXIT_" . $bmwqemu::vars{JOBTOKEN} or die 'Need variable JOBTOKEN' . ":";
     my $log = $self->serial_terminal_log_file();
     my $max_tries = 10;
-
-    if (_is_vmware) {
-        # libvirt esx driver does not support `virsh console', so
-        # we have to connect to VM's serial port via TCP which is
-        # provided by ESXi server.
-        $cmd = 'socat - TCP4:' . $bmwqemu::vars{VMWARE_HOST} . ':' . $port . ',crnl';
-    }
-    elsif (_is_hyperv) {
-        # Hyper-V does not support serial console export via TCP, just
-        # windows named pipes (e.g. \\.\pipe\mypipe). Such a named pipe
-        # has to be enabled by a namedpipe-to-TCP on HYPERV_SERVER application.
-        $cmd = 'socat - TCP4:' . $bmwqemu::vars{HYPERV_SERVER} . ':' . $port . ',crnl';
-    }
-    else {
-        $cmd = "virsh console $name $devname$port";
-    }
-
-    $cmd_full = "script -f $log -c '$cmd; echo \"$marker \$?\"'";
+    my $cmd = _is_hyperv ? serial_console_cmd_hyperv($name, $devname, $port) :
+      _is_vmware ? serial_console_cmd_vmware($name, $devname, $port) :
+      serial_console_cmd_svirt($name, $devname, $port);
+    my $cmd_full = "script -f $log -c '$cmd; echo \"$marker \$?\"'";
     bmwqemu::diag("Starting SSH connection to connect to libvirt domain '$name' (cmd: '$cmd'), full cmd: '$cmd_full'");
 
-    ($ssh, $chan) = $self->run_ssh($cmd_full, blocking => 0);
+    my ($ssh, $chan) = $self->run_ssh($cmd_full, blocking => 0);
     usleep(500) while ($self->run_ssh_cmd("test -e $log") != 0 && $max_tries-- > 0);
     $self->die("Command 'script' did not create logfile $log") if ($max_tries < 1);
     $self->{need_delete_log} = 1;
 
-    $ret = $self->run_ssh_cmd("grep -q '^$marker' $log");
+    my $ret = $self->run_ssh_cmd("grep -q '^$marker' $log");
     if (!$ret) {
-        (undef, $stdout, undef) = $self->run_ssh_cmd("cat $log", wantarray => 1);
+        (undef, my $stdout, undef) = $self->run_ssh_cmd("cat $log", wantarray => 1);
         $self->die("problem with virsh: cmd: '$cmd', output of script wrapper: '$stdout')");
     }
 


backend/svirt.pm
```diff
@@ -284,11 +284,9 @@ sub delete_log ($self) {
 
 # Intent to use CORE::GLOBAL::die, that does not have $self.
 sub die ($self, $err = '') {
-    if ($self->{need_delete_log}) {
-        bmwqemu::fctwarn("error, cleanup logs before die");
-        $self->delete_log();
-    }
-    die $err;
+    die $err unless $self->{need_delete_log};
```
Martchus: This function previously always died. Now this is conditional. This change in behavior is not in accordance with the commit message.

okurz:
As far as I could I ran local manual verification runs using a hyperv host (OSD worker34:11).

okurz:
review plz

backend/svirt.pm
```diff
@@ -111,17 +111,16 @@ sub can_handle ($self, $args) {
...
-    my $rsp;
-    if (_is_hyperv) {
-        $rsp = $self->run_ssh_cmd("powershell -Command \"if (\$(Get-VM -VMName $vmname \| Where-Object {\$_.state -eq 'Off'})) { exit 1 } else { exit 0 }\"");
-    }
-    else {
-        my $libvirt_connector = $bmwqemu::vars{VMWARE_REMOTE_VMM} // '';
-        $rsp = $self->run_ssh_cmd("! virsh $libvirt_connector dominfo $vmname | grep -w 'shut off'");
-    }
-    return $rsp;
+    return $self->run_ssh_cmd(_is_hyperv ? is_shutdown_ssh_cmd_hyperv($vmname) : is_shutdown_ssh_cmd_svirt($vmname));
```
d3flex: I guess no need to duplicate `_ssh_cmd`. how `is_shutdown_hyperv` or `is_shutdown_hyperv_cmd` sound to you?

backend/svirt.pm
```diff
@@ -40,6 +40,8 @@ sub new ($class) {
     return $self;
 }
 
+sub vmname ($self) { $self->console('svirt')->name }
```
d3flex, line 43: 👍

backend/svirt.pm
d3flex: Actually, you already have a _convention_ for `save_snapshot`

backend/svirt.pm
Wabri, line 43: Why not create a `our` variable instead? with that you can avoid call console every time.

backend/baseclass.pm
```diff
@@ -1224,26 +1224,24 @@ sub new_ssh_connection ($self, %args) {
...
+            my @e = $ssh->error;
+            bmwqemu::diag "Could not connect to $con_pretty: $e[2]. Retrying up to $counter more times after sleeping ${interval}s";
             sleep($interval);
             $counter--;
             next;
         }
+        if (!$args{use_ssh_agent} && defined($args{password})) {
+            $ssh->auth(username => $args{username}, password => $args{password});
+        }
+        else {
```
Wabri:
```suggestion
        } else {
```

backend/svirt.pm
```diff
@@ -61,23 +63,29 @@ sub do_start_vm ($self, @) {
...
+    $self->run_ssh_cmd(qq($ps "\$ProgressPreference='SilentlyContinue'; Remove-VM -Force -VMName $vmname"));
+}
+
+sub do_stop_vm_svirt ($self) {
+    my $vmname = $self->vmname;
+    my $virsh = 'virsh';
+    $virsh .= ' ' . $bmwqemu::vars{VMWARE_REMOTE_VMM} if $bmwqemu::vars{VMWARE_REMOTE_VMM};
+    $self->run_ssh_cmd("$virsh destroy $vmname");
+    $self->run_ssh_cmd("$virsh undefine --snapshots-metadata $vmname");
+
```
Wabri:
```suggestion
```

backend/baseclass.pm
```diff
@@ -1224,26 +1224,24 @@ sub new_ssh_connection ($self, %args) {
...
-        else {
-            bmwqemu::diag "Could not connect to $con_pretty. Retrying up to $counter more times after sleeping ${interval}s";
+        if (!$ssh->connect($args{hostname}, $args{port})) {
+            my @e = $ssh->error;
+            bmwqemu::diag "Could not connect to $con_pretty: $e[2]. Retrying up to $counter more times after sleeping ${interval}s";
             sleep($interval);
             $counter--;
             next;
         }
+        if (!$args{use_ssh_agent} && defined($args{password})) {
```
Wabri:
```suggestion
$args{use_ssh_agent} && defined($args{password})
    ? $ssh->auth(username => $args{username}, password => $args{password})
    : $ssh->auth_agent($args{username});        
```

like you have done in line 88

backend/svirt.pm
```diff
@@ -237,40 +252,25 @@ C<$args{devname}> used device name
...
-        $cmd = 'socat - TCP4:' . $bmwqemu::vars{HYPERV_SERVER} . ':' . $port . ',crnl';
-    }
-    else {
-        $cmd = "virsh console $name $devname$port";
-    }
-
-    $cmd_full = "script -f $log -c '$cmd; echo \"$marker \$?\"'";
+    my $cmd = _is_hyperv ? serial_console_cmd_hyperv($name, $devname, $port) :
+      _is_vmware ? serial_console_cmd_vmware($name, $devname, $port) :
+      serial_console_cmd_svirt($name, $devname, $port);
```
Wabri, line 261: This is unreadable logically for me, maybe better something like this?

```
my $cmd = '';
if (_is_hyperv) {
    $cmd = serial_console_cmd_hyperv($name, $devname, $port)
} else {
    $cmd = _is_vmware 
      ? serial_console_cmd_vmware($name, $devname, $port) 
      : serial_console_cmd_svirt($name, $devname, $port);
}
```

backend/svirt.pm
okurz: So I guess just `is_shutdown_cmd_hyperv`? I can do that. Did that

perlpunk:
The ticket in the title is not clickable. Why not add it to the description and have a better title describing what this is about?

backend/svirt.pm
```diff
@@ -107,41 +115,40 @@ sub can_handle ($self, $args) {
     $args->{function} eq 'snapshots' && _vmm_family =~ qr/kvm|hyperv|vmware/ ? {ret => 1} : undef;
 }
 
+sub is_shutdown_cmd_hyperv ($vmname) { "powershell -Command \"if (\$(Get-VM -VMName $vmname \| Where-Object {\$_.state -eq 'Off'})) { exit 1 } else { exit 0 }\"" }
```
perlpunk, line 117:
```suggestion
sub is_shutdown_cmd_hyperv ($vmname) { qq#powershell -Command "if (\$(Get-VM -VMName $vmname | Where-Object {\$_.state -eq 'Off'})) { exit 1 } else { exit 0 }"# }
```
or even better with sprintf:

```suggestion
sub is_shutdown_cmd_hyperv ($vmname) { sprintf 'powershell -Command "if ($(Get-VM -VMName %s | Where-Object {$_.state -eq 'Off'})) { exit 1 } else { exit 0 }"', $vmname }
```

Wabri (changes requested):

backend/baseclass.pm
okurz: that's not compliant with our Perl tidy change. The newline is added automatically. I would also prefer your style :)

backend/svirt.pm
okurz: > This is unreadable logically for me, maybe better something like this?
> 
> ```
> my $cmd = '';
> if (_is_hyperv) {
>     $cmd = serial_console_cmd_hyperv($name, $devname, $port)
> } else {
>     $cmd = _is_vmware 
>       ? serial_console_cmd_vmware($name, $devname, $port) 
>       : serial_console_cmd_svirt($name, $devname, $port);
> }
> ```

That would be mixing two concepts of using if/else and the ternary operator. In the end we want to assign a value to `$cmd` so I think the ternary operator is a good choice to show that purpose. We could consider a helper function with early-returns like this

```perl
sub serial_console_cmd ($name, $devname, $port) {
    return serial_console_cmd_hyperv($name, $devname, $port) if _is_hyperv;
    return serial_console_cmd_vmware($name, $devname, $port) if _is_vmware;
    return serial_console_cmd_svirt($name, $devname, $port);
}
```

but likely better is the long term approach I planned to use proper object oriented programming. So this is just a stepping stone anyway.

perlpunk: For me it reads naturally. Is it X, then do A. Is it Y, then do B. Otherwise do C

backend/svirt.pm
okurz: I don't consider the sprintf more readable. Also I didn't change this command string, just moved it around. Can we keep it as-is?

backend/svirt.pm
Wabri: Ok, let's leave as it is :+1: for me it's confusing, but I don't like ternary operator in general so that's my problem :grin:

backend/svirt.pm
perlpunk: package global variables should be avoided unless really necessary

perlpunk (approved):
I don't like the `" ...\"..."` stuff, but ok

backend/svirt.pm
Wabri: The fact is that if we call this function multiple times we access `console` multiple time, I don't know how impactfull can be in the long term. I don't know if it make sense to optimize since I don't know how much time it takes or how many resources it takes, I guess not much, but when I see something like this I worry about this kind of things.

perlpunk: Let's check then:
https://github.com/os-autoinst/os-autoinst/blob/master/backend/baseclass.pm#L753
```
sub console ($self, $testapi_console) {
    my $ret = $testapi::distri->{consoles}->{$testapi_console};
    carp "console $testapi_console does not exist" unless $ret;
    return $ret;
}
```
So this is actually just returning something stored in a hash already, and `->name` is an accessor, so it's also fast and should not worry us.

Martchus (approved):

backend/baseclass.pm
okurz: I guess you mean like in https://github.com/os-autoinst/os-autoinst/pull/2731/files#diff-1b51f8d18647e5233cfce8985c22261af0973de2cccd7d16100983f235eed49dR87 ? Normally I would use the ternary operator mostly for conditional assignments like `my $var = is_foo ? 'foo' : 'bar';`. I used the ternary operator for a function call in https://github.com/os-autoinst/os-autoinst/pull/2731/files#diff-1b51f8d18647e5233cfce8985c22261af0973de2cccd7d16100983f235eed49dR87 in an unusual way as the code would most likely be much better with using proper object inheritance but I am not there yet. The code here I don't plan to change in next steps when separating svirt+hyperv+vmware as it's not any different for the different backends. I would favor to keep the code as-is unless you still favor your suggestion after my explanation. Then I will happily use that :)

backend/svirt.pm
Wabri: Good!

PR #2734, opened 2025-08-05
Use a monotonically increasing marker to fix poo#186834

ricardobranco777:
Introduce an internal hashed_string_counter function to use a monotonically increasing marker to avoid ugly hacks like removing `/tmp/script*` when switching from the root serial terminal to the user serial terminal.  Currently this is needed when running the same command as root and rootless because there's an obvious hash collision when the commands are the same.

Related ticket: https://progress.opensuse.org/issues/186834

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -84,6 +84,11 @@ sub become_root ($self) {
     testapi::enter_cmd('cd /tmp');
 }
 
+sub hashed_string_counter ($text) {
+    state $count = 0;
+    return $text . "." . $count++;
+}
+
 =head2 script_run
 
   script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet])
@@ -123,7 +128,7 @@ sub script_run ($self, $cmd, @args) {
         die "Terminator '&' found in script_run call. script_run can not check script success. Use 'background_script_run' instead."
           if $cmd =~ qr/(?<!\\)&$/;
         my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
-        my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
+        my $marker = "; echo " . hashed_string_counter($str) . "-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
             testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet}, buffer_size => length($cmd) + 128);
@@ -165,7 +170,7 @@ sub background_script_run ($self, $cmd, %args) {
     $cmd = "( $cmd )";
     testapi::type_string $cmd;
     my $str = testapi::hashed_string("SR" . $cmd);
-    my $marker = "& echo $str-\$!-" . ($args{output} ? "Comment: $args{output}" : '');
+    my $marker = "& echo " . hashed_string_counter($str) . "-\$!-" . ($args{output} ? "Comment: $args{output}" : '');
     if (testapi::is_serial_terminal) {
         testapi::type_string($marker);
         testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
@@ -193,7 +198,7 @@ sub script_sudo ($self, $prog, $wait = 10) {
     my $str;
     if ($wait > 0) {
         $str = testapi::hashed_string("SS$prog$wait");
-        $prog = "$prog; echo $str > /dev/$testapi::serialdev";
+        $prog = "$prog; echo " . hashed_string_counter($str) . " > /dev/$testapi::serialdev";
     }
     testapi::type_string "sudo $prog\n";
     if (testapi::check_screen "sudo-passwordprompt", 3) {
@@ -238,7 +243,8 @@ sub script_output ($self, $script, @args) {
             type_command => length($script) < 80,
         }, ['timeout'], @args);
 
-    my $marker = testapi::hashed_string("SO$script");
+    my $str = testapi::hashed_string("SO$script");
+    my $marker = hashed_string_counter($str);
     my $script_path = "/tmp/script$marker.sh";
 
     # prevent use of network for offline installations
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -571,7 +571,7 @@ subtest 'upload_logs' => sub {
                 cmd => 'backend_type_string'
             },
             {
-                text => '; echo XXX-$?-',
+                text => '; echo XXX.15-$?-',
                 cmd => 'backend_type_string'
             },
             {
@@ -587,7 +587,7 @@ subtest 'upload_logs' => sub {
                 cmd => 'backend_type_string'
             },
             {
-                text => '; echo XXX-$?-',
+                text => '; echo XXX.16-$?-',
                 cmd => 'backend_type_string'
             },
             {
@@ -603,7 +603,7 @@ subtest 'script_sudo' => sub {
     script_sudo "rm /boot/grub/menu.lst";
     is_deeply($cmds, [
             {
-                text => "sudo rm /boot/grub/menu.lst; echo XXX > /dev/null\n",
+                text => "sudo rm /boot/grub/menu.lst; echo XXX.17 > /dev/null\n",
                 cmd => 'backend_type_string'
             },
             {
@@ -1294,6 +1294,22 @@ subtest 'retrieve and change various global variables' => sub {
         $setter->($new);
         is $getter->(), $new, "get/set $func works";
     }
+};
+
+subtest 'hashed_string_counter increments correctly' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->redefine(hashed_string => sub ($text) { "XXX$text" });
+
+    my ($h0, $n0) = split /\./, distribution::hashed_string_counter("foo");
+    my ($h1, $n1) = split /\./, distribution::hashed_string_counter("foo");
+    my ($h2, $n2) = split /\./, distribution::hashed_string_counter("bar");
+
+    is $h0, 'foo', 'first call hash correct';
+    is $h1, 'foo', 'second call hash correct';
+    is $h2, 'bar', 'third call hash correct';
+
+    is $n1, $n0 + 1, 'counter incremented for second call';
+    is $n2, $n1 + 1, 'counter incremented for third call';
 };
 
 done_testing;


ricardobranco777:
> ```
> fvogt@fvogt-thinkpad:~> perl -e "use Time::HiRes 'gettimeofday'; print(gettimeofday())"
> 1754397195561418
> ```
> 
> This is very long - what about about a counter that starts from 1 and increments on each use?

Do you think we'll need some prefix to guarantee uniqueness?

Vogtinator:
> > This is very long - what about about a counter that starts from 1 and increments on each use?
> 
> Do you think we'll need some prefix to guarantee uniqueness?

Which kind of prefix? What would it conflict with?

Vogtinator:
> This would need lots of VR's

Yup

> and also I think it's outside the scope of this PR

Also yup

> and my Perl skills are limited lol.

Same...

distribution.pm
```diff
@@ -238,8 +239,8 @@ sub script_output ($self, $script, @args) {
             type_command => length($script) < 80,
         }, ['timeout'], @args);
 
-    my $marker = testapi::hashed_string("SO$script");
```
d3flex, line 241: What if you use the `C<$count>` of the hashed_string in the test modules to have different hashes for user and root?

d3flex (changes requested):
I think the `hashed_string` should remain. Instead you can use the `$count` parameter (somehow). But I am not sure what's the way to go without do some tests myself. Sorry

distribution.pm
ricardobranco777: Please check forced-push.

Vogtinator: Would work, but not always: If there's a collision (which is not that unlikely with five base64 characters of data), it'll end up with permission denied again.

With just the counter there will never be any.

perlpunk: Yeah, the `$count` should be part of the marker and not being hashed itself

distribution.pm
```diff
@@ -238,8 +239,9 @@ sub script_output ($self, $script, @args) {
...
+    my $marker = testapi::hashed_string("SO$script$count");
```
perlpunk:
```suggestion
    my $marker = testapi::hashed_string("SO$script") . "-$count";
```
or maybe
```suggestion
    my $marker = testapi::hashed_string("SO$script") . "_$count";
```

distribution.pm
okurz: Be aware that `testapi::hashed_string` already has a "count" parameter but what it does is configure the amount of characters from the hash to keep in mind.

ricardobranco777: There will always be collisions in our case because we're running some tests as root and then rootless in some modules, so the commands will be the same.  What if I append or prepend the UID to the script passed to `hashed_string`?

distribution.pm
```diff
@@ -238,8 +239,9 @@ sub script_output ($self, $script, @args) {
             type_command => length($script) < 80,
         }, ['timeout'], @args);
 
-    my $marker = testapi::hashed_string("SO$script");
+    my $marker = testapi::hashed_string("SO$script") . $count;
     my $script_path = "/tmp/script$marker.sh";
+    $count++;
```
Martchus:
```suggestion
    my $marker = testapi::hashed_string("SO$script") . $count++;
    my $script_path = "/tmp/script$marker.sh";
```

You might also want to move `state $count = 0;` directly before `my $marker …` so everything that belongs together is actually together.

And if you have everything together you can also split the code easily into a separate function, e.g. `make_script_output_marker`. Then you can also test this function separately in unit tests.

Martchus (changes requested):

distribution.pm
ricardobranco777: I moved the logic to another function.

distribution.pm
```diff
@@ -122,7 +131,7 @@ sub script_run ($self, $cmd, @args) {
     if ($args{timeout} > 0) {
         die "Terminator '&' found in script_run call. script_run can not check script success. Use 'background_script_run' instead."
           if $cmd =~ qr/(?<!\\)&$/;
-        my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
+        my $str = hashed_string_counter("SR" . $cmd . $args{timeout});
```
Martchus:
```suggestion
        my $str = hashed_string_counter("SR$cmd$args{timeout}");
```

distribution.pm
```diff
@@ -84,6 +84,15 @@ sub become_root ($self) {
...
 }
 
+sub hashed_string_counter ($text) {
+    state $count = 0;
+
+    my $marker = testapi::hashed_string($text) . $count;
+    $count++;
+
+    return $marker;
+}
```
Martchus:
```suggestion
sub hashed_string_counter ($text) {
    state $count = 0;
    return testapi::hashed_string($text) . $count++;
}
```

distribution.pm
ricardobranco777: This one I'd leave as is to keep a consistent style in the module.

ricardobranco777:
I can't follow the CI failures and would appreciate any help or even a takeover given that I'll be on vacation Thursday & Friday.

d3flex:
> I can't follow the CI failures and would appreciate any help or even a takeover given that I'll be on vacation Thursday & Friday.

I am not completely sure what was the problem with yr approach, but it seems to behave strange in some cases like the upload_logs. Anyways... I got confused, so I started from scratch to come up with this, which seems to work. feel free to use it or inspired to fix what you already have or whatever, but it accomplishes the same with some tweaks. 

```diff --git a/distribution.pm b/distribution.pm
index b3f315bd..7caf81a6 100644
--- a/distribution.pm
+++ b/distribution.pm
@@ -84,6 +84,11 @@ sub become_root ($self) {
     testapi::enter_cmd('cd /tmp');
 }
 
+sub hashed_string_counter ($text) {
+     state $count = 0;
+     return $text . "." . $count++;
+ }
+
 =head2 script_run
 
   script_run($cmd [, timeout => $timeout] [, output => $output] [,quiet => $quiet])
@@ -123,7 +128,7 @@ sub script_run ($self, $cmd, @args) {
         die "Terminator '&' found in script_run call. script_run can not check script success. Use 'background_script_run' instead."
           if $cmd =~ qr/(?<!\\)&$/;
         my $str = testapi::hashed_string("SR" . $cmd . $args{timeout});
-        my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
+        my $marker = "; echo " . hashed_string_counter($str) . "-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
         if (testapi::is_serial_terminal) {
             testapi::type_string($marker);
             testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet}, buffer_size => length($cmd) + 128);
@@ -165,7 +170,7 @@ sub background_script_run ($self, $cmd, %args) {
     $cmd = "( $cmd )";
     testapi::type_string $cmd;
     my $str = testapi::hashed_string("SR" . $cmd);
-    my $marker = "& echo $str-\$!-" . ($args{output} ? "Comment: $args{output}" : '');
+    my $marker = "& echo " . hashed_string_counter($str) . "-\$!-" . ($args{output} ? "Comment: $args{output}" : '');
     if (testapi::is_serial_terminal) {
         testapi::type_string($marker);
         testapi::wait_serial($cmd . $marker, no_regex => 1, quiet => $args{quiet});
@@ -193,7 +198,7 @@ sub script_sudo ($self, $prog, $wait = 10) {
     my $str;
     if ($wait > 0) {
         $str = testapi::hashed_string("SS$prog$wait");
-        $prog = "$prog; echo $str > /dev/$testapi::serialdev";
+        $prog = "$prog; echo " . hashed_string_counter($str) . " > /dev/$testapi::serialdev";
     }
     testapi::type_string "sudo $prog\n";
     if (testapi::check_screen "sudo-passwordprompt", 3) {
diff --git a/t/03-testapi.t b/t/03-testapi.t
index 237bd25a..5a4e1d9c 100755
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -571,7 +571,7 @@ subtest 'upload_logs' => sub {
                 cmd => 'backend_type_string'
             },
             {
-                text => '; echo XXX-$?-',
+                text => '; echo XXX.15-$?-',
                 cmd => 'backend_type_string'
             },
             {
@@ -587,7 +587,7 @@ subtest 'upload_logs' => sub {
                 cmd => 'backend_type_string'
             },
             {
-                text => '; echo XXX-$?-',
+                text => '; echo XXX.16-$?-',
                 cmd => 'backend_type_string'
             },
             {
@@ -603,7 +603,7 @@ subtest 'script_sudo' => sub {
     script_sudo "rm /boot/grub/menu.lst";
     is_deeply($cmds, [
             {
-                text => "sudo rm /boot/grub/menu.lst; echo XXX > /dev/null\n",
+                text => "sudo rm /boot/grub/menu.lst; echo XXX.17 > /dev/null\n",
                 cmd => 'backend_type_string'
             },
             {
@@ -1296,6 +1296,22 @@ subtest 'retrieve and change various global variables' => sub {
     }
 };
 
+subtest 'hashed_string_counter increments correctly' => sub {
+    my $mock_testapi = Test::MockModule->new('testapi');
+    $mock_testapi->redefine(hashed_string => sub ($text) { "XXX$text" });
+
+    my ($h0, $n0) = split /\./, distribution::hashed_string_counter("foo");
+    my ($h1, $n1) = split /\./, distribution::hashed_string_counter("foo");
+    my ($h2, $n2) = split /\./, distribution::hashed_string_counter("bar");
+
+    is $h0, 'foo', 'first call hash correct';
+    is $h1, 'foo', 'second call hash correct';
+    is $h2, 'bar', 'third call hash correct';
+
+    is $n1, $n0 + 1, 'counter incremented for second call';
+    is $n2, $n1 + 1, 'counter incremented for third call';
+};
+
 done_testing;
 
 END {
```

ricardobranco777:
> I am not completely sure what was the problem with yr approach, but it seems to behave strange in some cases like the upload_logs. Anyways... I got confused, so I started from scratch to come up with this, which seems to work. feel free to use it or inspired to fix what you already have or whatever, but it accomplishes the same with some tweaks.

I've just applied it in full.  Let's check.  Thanks!

perlpunk:
Now the fullstack test is failing.
There is one error which is apparently not related to your change:
https://github.com/os-autoinst/os-autoinst/actions/runs/17201530264/job/48792938627?pr=2734#step:3:917
```
3: # [2025-08-25T07:07:12.100976Z] [info] [pid:3113] ::: basetest::runtest: # Test died: Too many arguments for subroutine 'select_ssh_console_fail_test::run' (got 1; expected 0) at /opt/script/../basetest.pm line 346.
```
I guess this never popped up because the inner test is supposed to fail anyway.
I will take care of that.

The actual failure is https://github.com/os-autoinst/os-autoinst/actions/runs/17201530264/job/48792938627?pr=2734#step:3:1061
```
3: #     doesn't match '(?^u:do not wait_still_screen)'
```
which is supposed to come from 
https://github.com/os-autoinst/os-autoinst/blob/master/t/data/tests/tests/typing.pm#L38
The typing test starts at:
https://github.com/os-autoinst/os-autoinst/actions/runs/17201530264/job/48792938627?pr=2734#step:3:657
and there's
```
3: # [2025-08-25T07:03:11.408083Z] [info] [pid:3113] ::: basetest::runtest: # Test died: command 'cat /proc/cpuinfo' timed out at /opt/script/../autotest.pm line 515.
```
which could be related to the change

perlpunk:
```
testapi::type_string(string="; echo X6MQS.0-\$?-
...
testapi::wait_serial: (?^u:X6MQS-\d+-): fail
```
so there is a mismatch, the expected string has no counter

okurz:
Good. Let's have
* #2741

merged first so that we can trust the tests

perlpunk:
@ricardobranco777  The change is now missing at line 246 of distribution.pm

distribution.pm
```diff
@@ -238,7 +243,7 @@ sub script_output ($self, $script, @args) {
             type_command => length($script) < 80,
         }, ['timeout'], @args);
 
-    my $marker = testapi::hashed_string("SO$script");
+    my $marker = testapi::hashed_string_counter("SO$script");
```
perlpunk:
```suggestion
    my $str = testapi::hashed_string("SO$script");
    my $marker = hashed_string_counter($str);
```

PR #2735, opened 2025-08-07
sshVirtsh provide_image_vmware_in_ds: fix file checks and add more info about images

m-dati:
In `provide_image_vmware_in_ds` image elaboration script
the syntax to check if a file ends by .xz, was ok in unit-tests 'bash', but not in vmware 'sh': apply common syntax valid in all cases.

Add in script: prints for steps identification and files names display; useful to trace the files flow and in debugging cases too. It helps to clarify doubtful cases where image should have been decompressed and detected, like in:
https://openqa.suse.de/tests/18692763/logfile?filename=autoinst-log.txt#line-334

That last case skipped the decompression due to the above glitch on check-syntax.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -350,20 +350,23 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     # Use the standard folder for an input file without full path
     my $file_origin = ($input_file eq $basefile) ? "$base_dir/$vmware_nfs_datastore/$nfs_dir/$basefile" : $input_file;
     # check image is present
+    # Note that on ESXi vmware hosts, /bin/sh points to 'busybox', based on underlying system.
     my $cmd = <<~"EOF";
     $debug
+    IN=$input_file
     if test -e "$dest_image"; then
         echo "Waiting while $input_file is loading:"
         while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [ \${IN##*.} = 'xz' ]; then 
         if [ -e "$dest_image.xz" ] || cp "$file_origin" "$vmware_openqa_datastore"; then
             xz --decompress --keep "$dest_image.xz"
         fi
     else
         cp "$file_origin" "$vmware_openqa_datastore"
     fi
+    echo "Done: origin:" $file_origin* " ; dest.:" $dest_image*
     EOF
 
     my $ret = $self->run_cmd($cmd, domain => 'sshVMwareServer');


consoles/sshVirtsh.pm
```diff
@@ -352,18 +352,20 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
     my $cmd = <<~"EOF";
     $debug
+    IN=$input_file
     if test -e "$dest_image"; then
         echo "Waiting while $input_file is loading:"
         while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [[ \${IN: -3} == '.xz' ]]; then
```
mloviska: It would be better to stick to POSIX shell syntax. I believe `[[ ]]` is not POSIX, but I might be wrong

m-dati: this check I verified, now works fine both as on bash as on a vmware sh machine: should I use `test ...` ?

m-dati: Done. But CI reacts differently. Fixing it

m-dati: CI pass ok

m-dati:
Commit message updated: PR ready.

consoles/sshVirtsh.pm
```diff
@@ -352,18 +352,20 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
     my $cmd = <<~"EOF";
     $debug
+    IN=$input_file
     if test -e "$dest_image"; then
         echo "Waiting while $input_file is loading:"
         while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [ \${IN: -3} == '.xz' ]; then
```
d3flex: Can i suggest 
```suggestion
    elif [ \${IN##*.} == 'xz' ]; then 
```

This will extract everything after the dot. 
I also not sure why need another variable.
I would rather keep the `$input_file` so at the end, would you mind to do


```suggestion
    elif [ \${input_file##*.} == 'xz' ]; then 
```

ricardobranco777: This is also a Bashism.  Please test in POSIX shell.  You can experiment using the Debian docker image like this:

```
$ podman run --rm -it docker.io/library/debian sh
# x="a.xz"
# echo "${x: -3}"
sh: 2: Bad substitution
```

ricardobranco777: This is POSIX shell and this works.

m-dati: Kind premise, before to appy it here I checked the syntax of the code on a TW bash and then also on a specific vmware_host, so the solution here provided is the result of long analysis and attempts, finally passed. Note also that vmware machines use a specific shell and `/bin/sh` points to `busybox`.

So, about the variable, sorry, it won't work, the `$input_file` is a "perl" variable and its content is substituted into the final bash script, until EOF, becomiing a fixed string; similarly, all the other `$xxx` variables.

I **had to** add an `IN` shell variable, because the syntax on strings like `${A ...}` doesn't apply to (resulting) fixed string like `${"abcd": -3}.

Your suggestion on `IN##` is welcome, even if the other check code on vmware host shell (strangely) also worked.

m-dati: Thanks, but the same it worked perfectly on the `esxi` machine, configured for our vmware tests. Anyway please see  https://github.com/os-autoinst/os-autoinst/pull/2735#discussion_r2262559252

ricardobranco777: It does indeed work on Busybox shell:

```
$ podman run --rm -it docker.io/library/busybox sh
/ # x="a.xz"
/ # echo "${x: -3}"
.xz
```

But then let's add a comment that we're depending on a non-standard feature in Busybox sh.  Also let's use `=` with `[` instead of `==` because `==` is used for pattern matching.

consoles/sshVirtsh.pm
```diff
@@ -352,18 +352,20 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     # check image is present
     my $cmd = <<~"EOF";
     $debug
+    IN=$input_file
```
ricardobranco777, line 356: `IN` is very generic.

```suggestion
    input_file="$input_file"
```

consoles/sshVirtsh.pm
```diff
@@ -352,18 +352,20 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
     if test -e "$dest_image"; then
         echo "Waiting while $input_file is loading:"
         while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [ \${IN##*.} == 'xz' ]; then
```
ricardobranco777:
```suggestion
    elif [ \${input_file##*.} = 'xz' ]; then 
```

m-dati: I will change == with =, Thanks

consoles/sshVirtsh.pm
m-dati: Thanks, but sorry this doesn't add more utility to the script, I prefer here to use a shorter variable that even simplifies the global view.

consoles/sshVirtsh.pm
ricardobranco777: > Thanks, but sorry this doesn't add more utility to the script, I prefer here to use a shorter variable that even simplifies the global view.

It's not an utilitarian suggestion but for readability.

m-dati: equal sign changed and comment added

consoles/sshVirtsh.pm
```diff
@@ -350,20 +350,23 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
     my $cmd = <<~"EOF";
     $debug
+    IN=$input_file
     if test -e "$dest_image"; then
         echo "Waiting while $input_file is loading:"
         while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [ \${IN##*.} = 'xz' ]; then
```
ricardobranco777, line 362: Both @d3flex and me suggested to use `input_file` instead, or at least something more descriptive than rather just `IN`, which could mean literally anything, like an input stream, stdin, etc. being an English adverb or preposition. It's not clear at all that this is an input file unless you read the assignment above, but each line should be self-descriptible.  We should avoid this kind of abbreviations that don't save that much space.

https://dictionary.cambridge.org/dictionary/english/in

PD: I also think double-quotes wouldn't harm.

```suggestion
    elif [ "\${input_file##*.}" = 'xz' ]; then 
```

consoles/sshVirtsh.pm
mloviska: In POSIX you need to use `=` instead of `==` 

https://github.com/dylanaraps/pure-sh-bible?tab=readme-ov-file#variable-comparisons

consoles/sshVirtsh.pm
mloviska: It should be double quoted

consoles/sshVirtsh.pm
mloviska: I also like Ricardo's point

consoles/sshVirtsh.pm
```diff
@@ -350,20 +350,23 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     # Use the standard folder for an input file without full path
     my $file_origin = ($input_file eq $basefile) ? "$base_dir/$vmware_nfs_datastore/$nfs_dir/$basefile" : $input_file;
     # check image is present
+    # Note that on ESXi vmware hosts, /bin/sh points to 'busybox', based on underlying system.
```
mloviska, line 353: How do we know this? Can you please clarify?

consoles/sshVirtsh.pm
ricardobranco777: > Thanks, but sorry this doesn't add more utility to the script, I prefer here to use a shorter variable that even simplifies the global view.

I see no reason to change the variable name.  Let's avoid this kind of abbreviations in the form of 2 characters.  It doesn't simplify anything and just obscures the intention of the variable.

consoles/sshVirtsh.pm
ricardobranco777: Threads requesting changes shouldn't be put into "resolved" status without actually addressing the issue or leaving a comment when it's an obvious matter.

consoles/sshVirtsh.pm
m-dati: Login on esxi7, then:
```
which sh
ls -l /bin/sh
```

consoles/sshVirtsh.pm
m-dati: https://github.com/os-autoinst/os-autoinst/pull/2735#discussion_r2262661164

consoles/sshVirtsh.pm
m-dati: The main scope of this PR is to resolve a syntax issue and allow the .xz image to be decompressed and proceed with the boot. The variable name is out of scope here, being the current script update already working as it is now and not affected by the variable name, minor change.
Anyway that bash variable change can be done in any (other) moment.

ricardobranco777: > The main scope of this PR is to resolve a syntax issue and allow the .xz image to be decompressed and proceed with the boot. The variable name is out of scope here, being the current script update already working and it is now and not affected by the variable name, minor change. Anyway that bash variable change can be done in any (other) moment.

You're changing the variable name from a readable one to an unreadable one when all that's required is to create a Bash variable that mirrors the Perl variable, and this can be done in an elegant way and you're choosing an unreadable name.

I can't agree to commit code that doesn't meet the minimum quality standards for backend code.

consoles/sshVirtsh.pm
ricardobranco777: > Login on esxi7, then:
> 
> ```
> which sh
> ls -l /bin/sh
> ```

Can you please provide the output of such commands and tell why it proves your point?

ricardobranco777: The correct way to verify this would be something like this:

```
$ podman run --rm -it busybox sh
/ # ls -l /bin/sh
-rwxr-xr-x  405 root     root       1029688 Sep 26  2024 /bin/sh
/ # ls -l /bin/busybox 
-rwxr-xr-x  405 root     root       1029688 Sep 26  2024 /bin/busybox
```

mloviska:
```
[root@esxi7:~] readlink /bin/sh
/usr/lib/vmware/busybox/bin/busybox
```

It is a link, but anyway, POSIX shell should be the same

consoles/sshVirtsh.pm
mloviska: >  I would rather keep the $input_file so at the end, would you mind to do

I also agree with this point. Please, let's change the PR accordingly and we can merge it. This does not need to hang over here anymore

m-dati:
I don’t oppose to the shell variable renaming, but I clearly explained (https://github.com/os-autoinst/os-autoinst/pull/2735#discussion_r2263031662) that the variable name now and here is a minor problem and I wanted before of all to achieve the real PR target.

In this PR, the fix required the creation of a new shell variable, of any preferred name, to host the text and allow to detect the rightmost ‘.xz’ string from the filename, to trigger the file decompression, allowing then the ‘vmdk’ image to successful boot. The shell script is embedded into the perl code and not a file per-se.

But since Ricardo is blocking the PR merge, unless his request for variable renaming is executed, as if it were more important than the bug fix itself that should allow the image to boot and testing to proceed, also supported by others, this PR has become useless.

So, apologizing with the colleague of the other team here also involved, that already approved the PR, I close this PR.

PR #2736, opened 2025-08-09
VMWare: Use POSIX shell on ESXi host that uses Busybox

ricardobranco777:
Use POSIX shell on ESXi host that uses Busybox.  Also apply consistent use of `[` and readability fixes.

Open question: Do we really need to keep the compressed image?

Supersedes https://github.com/os-autoinst/os-autoinst/pull/2735

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -350,20 +350,23 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     # Use the standard folder for an input file without full path
     my $file_origin = ($input_file eq $basefile) ? "$base_dir/$vmware_nfs_datastore/$nfs_dir/$basefile" : $input_file;
     # check image is present
+    # Note: This script must be in POSIX shell as ESXi uses busybox for /bin/sh
     my $cmd = <<~"EOF";
     $debug
-    if test -e "$dest_image"; then
+    input_file="$input_file"
+    if [ -e "$dest_image" ]; then
         echo "Waiting while $input_file is loading:"
-        while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
+        while ps -v | grep -E "cp .*$baseimage|xz .*$basefile" | grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -350,20 +350,23 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
-    if test -e "$dest_image"; then
+    input_file="$input_file"
+    if [ -e "$dest_image" ]; then
         echo "Waiting while $input_file is loading:"
-        while ps -v | grep -E "cp .*$baseimage|xz .*$basefile"|grep -v grep
+        while ps -v | grep -E "cp .*$baseimage|xz .*$basefile" | grep -v grep
             do sleep 5; done
         echo "VMware image $dest_image ready"
-    elif [[ "$input_file" == *.xz ]]; then 
+    elif [ "\${input_file##*.}" = xz ]; then
```
d3flex: not a big thing but lets quote the `xz`. other than that is fine

ricardobranco777: > not a big thing but lets quote the `xz`. other than that is fine

Fixed.

m-dati:
Final offense on https://github.com/os-autoinst/os-autoinst/pull/2735#issuecomment-3170650619

m-dati:
An insult to https://github.com/os-autoinst/os-autoinst/pull/2735#issuecomment-3170634991

ricardobranco777:
> An insult to [#2735 (comment)](https://github.com/os-autoinst/os-autoinst/pull/2735#issuecomment-3170634991)

You closed the PR so I had to open a new one. Please don't spam this PR with irrelevant links.

mloviska:
> Open question: Do we really need to keep the compressed image?

I cannot recall, why we do it. But it does not hurt and in case whatever happens to the original uncompressed image, we can simply repeat the step. Otherwise, the backend would need to download the compressed file again from osd via NFS.

ricardobranco777:
> What could be changed is `| grep -v grep` part.

We should keep it because this part is non-deterministic and we need to filter the main `grep` when it appears.

d3flex (approved):

perlpunk (approved):

PR #2738, opened 2025-08-20
Cleanup qcow images before perform a copy from local storage

d3flex:
A simple approach to perform automatic cleanup of kvm assets produced by the s390x kvm backend (svirt) in os-autoinst is to remove the existing one if it is already existed.

When the `add_disk` is invoked and the test settings provides HDD_* the `add_disk` is called passing the HDD_* variable with the path which obtain from the openQA location. if the image is not found the process dies. At least in the zkvm_add_disk. I guess this can leave old images in the hypervisor, which eventually should be cleaned up from cleanup-openqa-assets cronjob.

When `create` is passed in the `add_disk` the $file construct an .img file (openQA-SUT-x.img usually). This is the case when there is no HDD_*

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -504,6 +504,14 @@ sub _bootorder_elem ($doc, $bootorder) {
     return $elem;
 }
 
+sub _clean_storage($self, $basedir, $args) {
+    if ($args->{file} && ($args->{cdrom} || $args->{backingfile})) {
+        my $existing_img = path . join($basedir, basename($args->{file}));
+        $self->run_cmd(qq#rm $existing_img#) if $self->run_cmd(qq#test -e "$existing_img"#);
+        bmwqemu::diag qq/$existing_img deleted./ unless $@;
+    }
+}
+
 sub add_disk ($self, $args) {
     my $cdrom = $args->{cdrom};
     my $name = $self->name;
@@ -511,6 +519,7 @@ sub add_disk ($self, $args) {
     my $basedir = '/var/lib/libvirt/images/';
     my $vmware_datastore = $bmwqemu::vars{VMWARE_DATASTORE} // '';
     my $vmware_openqa_datastore = "/vmfs/volumes/$vmware_datastore/openQA/";
+    _clean_storage($self, $basedir, $args);
     if ($args->{create}) {
         $file = $self->_create_disk($args, $vmware_openqa_datastore, $file, $name, $basedir);
     }
--- a/t/22-svirt.t
+++ b/t/22-svirt.t
@@ -668,7 +668,7 @@ subtest 'Method consoles::sshVirtsh::add_disk()' => sub {
             my $exp_filename = 'foo_file_' . $svirt->name . '_thinfile.vmdk';
             set_var(VMWARE_NFS_DATASTORE => 'nfs');
             $svirt->add_disk({backingfile => 1, size => '77G', dev_id => $dev_id, file => $filename});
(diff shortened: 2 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -515,6 +515,11 @@ sub add_disk ($self, $args) {
         $file = $self->_create_disk($args, $vmware_openqa_datastore, $file, $name, $basedir);
     }
     else {
+        # When the s390-kvm boots copies HDD_. cleanup old qcow image before proceed
+        if ($args->{file}){
+        $self->run_cmd(qq{rm ${basedir}/$args->{file}}) if -e "${basedir}/$args->{file}";
+        bmwqemu::diag qq/$args->{file} deleted from ${basedir}/ unless $@;
+        }
```
Wabri:
```suggestion
        if ($args->{file}) {
            $self->run_cmd(qq{rm ${basedir}/$args->{file}}) if -e "${basedir}/$args->{file}";
            bmwqemu::diag qq/$args->{file} deleted from ${basedir}/ unless $@;
        }
```

Wabri (changes requested):

consoles/sshVirtsh.pm
```diff
@@ -515,6 +515,11 @@ sub add_disk ($self, $args) {
         $file = $self->_create_disk($args, $vmware_openqa_datastore, $file, $name, $basedir);
     }
     else {
+        # When the s390-kvm boots copies HDD_. cleanup old qcow image before proceed
+        if ($args->{file}) {
+            $self->run_cmd(qq{rm ${basedir}/$args->{file}}) if -e "${basedir}/$args->{file}";
```
perlpunk: Will `-e $path` actually work? I thought the file is possibly not on the same host?

perlpunk: Using `qq{}` is nice, but it can be confusing to the reader when curly braces are part of the content, so I would vote for a different delimiter.

```suggestion
            $self->run_cmd(qq#rm ${basedir}/$args->{file}#) if -e "${basedir}/$args->{file}";
```

consoles/sshVirtsh.pm
```diff
@@ -515,6 +515,11 @@ sub add_disk ($self, $args) {
         $file = $self->_create_disk($args, $vmware_openqa_datastore, $file, $name, $basedir);
     }
     else {
+        # When the s390-kvm boots copies HDD_. cleanup old qcow image before proceed
+        if ($args->{file} && $args->{cdrom} || $args->{backingfile}) {
```
okurz: why the cdrom when above you say HDD?

okurz (changes requested):
Please clarify in your git commit message subject why you want to do this. Now it just says "cleanup before perform a copy" which I would counter with "just copy directly to the target, no need for cleanup"

consoles/sshVirtsh.pm
d3flex: Definetely i need to make this clear. I am getting the understanding bit by bit as I go. 
A quick answer is because of `($cdrom || $backingfile)` in `_copy_image_to_vm_host` but... Consider everything wrong for now until I am sure that the change actually removes the expected files from the expected destination without defects. I will mark the PR as draft for now

Martchus (commented):
I don't think we should just always delete the existing image. I guess this is just a draft to get things going anyway.

When extending this it would perhaps make sense to create a bigger code snippet on the Perl side instead of playing ping-pong between the machines with individual SSH commands.

PR #2739, opened 2025-08-21
Check for available space before copying file

d3flex:

Continue from 4a2c5e88b20b03481a60247ebf231b241a1ab565

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -384,11 +384,18 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
     my $cmd =
       "$ds_debug if test -e $vmware_openqa_datastore$file_basename; then " .
       "while lsof | grep 'cp.*$file_basename'; do " .
-      "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
-      'done;' .
-      'else ' .
-      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
-      'fi;';
+      "echo 'File $file_basename is being copied by other process, sleeping for 60 seconds'; sleep 60; " .
+      "done; " .
+      "else " .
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -384,11 +384,18 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
...
-      'else ' .
-      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
-      'fi;';
+      "echo 'File $file_basename is being copied by other process, sleeping for 60 seconds'; sleep 60; " .
+      "done; " .
+      "else " .
+      qq{available_space=\$(df -B1 --output=avail "$vmware_openqa_datastore" | tail -n 1); } .
+      qq{size=\$(stat --printf="%s" "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename"); } .
+      q{if [ "$size" -gt "$available_space" ]; then } .
+      "echo 'Error: Not enough space on target datastore $vmware_openqa_datastore to copy $file_basename.'; " .
```
kalikiana, line 393: How about handling the deletion of old images as part of this?

1. Check for available space
2. Delete old images if needed
3. Only fail if there still isn't enough space

And maybe move this into a script and run it with `flock`, otherwise the check for available space is racy.

consoles/sshVirtsh.pm
```diff
@@ -384,11 +384,18 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
...
+      "exit 1; " .
+      "else " .
+      qq{cp "/vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename" "$vmware_openqa_datastore"; } .
+      "fi; " .
+      "fi;";
```
Martchus, line 398: It is probably time to move this snipped into a separate file - or at least change the quoting style to something that's easier to read, e.g. a Heredoc.

PR #2740, opened 2025-08-25
baseclass: Fix missing error details on failed SSH connection attempts

okurz:
Already since in 631d0f7a the SSHConnectionError->throw was introduced
there was no useful content available for the error variable "$@" so
this variable can not have useful content causing error messages as
observed like

```
Error connecting to root@my.machine:  at …
```

mind the double space after ":". 

The message for the above example then will look like:

```
Error connecting to <root@my.machine>: All authentication methods failed
```

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -1224,24 +1224,22 @@ sub new_ssh_connection ($self, %args) {
     my $con_pretty = "$args{username}\@$args{hostname}";
     $con_pretty .= ":$args{port}" unless $args{port} == 22;
     while ($counter > 0) {
-        if ($ssh->connect($args{hostname}, $args{port})) {
-
-            if (!$args{use_ssh_agent} && defined($args{password})) {
-                $ssh->auth(username => $args{username}, password => $args{password});
-            }
-            else {
-                # this relies on agent to be set up correctly
-                $ssh->auth_agent($args{username});
-            }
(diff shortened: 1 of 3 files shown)


backend/baseclass.pm
```diff
@@ -1224,26 +1224,24 @@ sub new_ssh_connection ($self, %args) {
...
+        }
+        else {
+            # this relies on agent to be set up correctly
+            $ssh->auth_agent($args{username});
+        }
+        bmwqemu::diag "SSH connection to $con_pretty established" if $ssh->auth_ok;
+        last;
     }
-    OpenQA::Exception::SSHConnectionError->throw(error => "Error connecting to <$con_pretty>: $@") unless $ssh->auth_ok;
+    $ssh->die_with_error("Error connecting to <$con_pretty>") unless $ssh->auth_ok;
```
Martchus: Is changing the exception type changing the error handling behavior? I suppose it isn't super important anyway as we use `die_with_error` in many other places and it presumably does the right thing. Actually, maybe we should remove `SSHConnectionError` entirely then as it isn't really used anywhere else I think.

okurz: Hm, good point. Will check

okurz: I now kept the custom exception in place

okurz:
Fixed CI failures

PR #2742, opened 2025-08-26
Refactor backend::baseclass

okurz:
* backend: Simplify json_fails collection with map
* backend: Simplify check_socket with early-return
* backend: Simplify do_capture by moving loop one level up
* backend: Extract method for do_capture loop body
* backend: Use more member variables in baseclass
* backend: Extract function 'matching_needles_and_tags' in baseclass

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -994,17 +994,10 @@ sub _failed_screens_to_json ($self) {
             push(@$failed_screens, $final_mismatch) if ($sim < 50);
         }
     }
-
-    my @json_fails;
-    for my $l (@$failed_screens) {
-        my ($img, $failed_candidates, $testtime, $similarity, $frame) = @$l;
-        my $h = {
-            candidates => $failed_candidates,
-            image => encode_base64($img->ppm_data),
-            frame => $frame,
-        };
-        push(@json_fails, $h);
-    }
+    my @json_fails = map {
+        my ($img, $failed_candidates, $testtime, $similarity, $frame) = @$_;
+        {candidates => $failed_candidates, image => encode_base64($img->ppm_data), frame => $frame}
+    } @$failed_screens;
 
     # free memory
     $self->assert_screen_fails([]);


Martchus (approved):

backend/baseclass.pm
```diff
@@ -923,14 +923,7 @@ sub similiarity_to_reference ($self, $args) {
...
 
-sub set_tags_to_assert ($self, $args) {
-    my $mustmatch = $args->{mustmatch};
-    my $timeout = $args->{timeout} // $bmwqemu::default_timeout;
-
-    # keep only the most recently used images (https://progress.opensuse.org/issues/15438)
-    needle::clean_image_cache();
-
-    # get the array reference to all matching needles
+sub matching_needles_and_tags ($mustmatch) {
```
nicksinger: Personally I use adjectives for properties of objects and verbs for functions like this here. How about calling this function `find_needles_with_tags`?

nicksinger (commented):
Just a minor nitpick - otherwise LGTM given the tests all pass :) I'd appreciate another review by our Perl experts

backend/baseclass.pm
```diff
@@ -311,7 +311,11 @@ $self->{cmdpipe} is closed, whichever occurs first.
 sub run_capture_loop ($self, $timeout = undef) {
     my $starttime = gettimeofday;
     $self->last_screenshot($starttime) unless $self->last_screenshot;
-    try { $self->do_capture($timeout, $starttime) }
+    try {
+        # Time slot buckets
+        my $buckets = {};
```
perlpunk, line 312: I don't understand the reason for this change

edit: I'm reviewing commit by commit. I guess it's necessary for the following commits

backend/baseclass.pm
```diff
@@ -320,7 +310,7 @@ sub run_capture_loop ($self, $timeout = undef) {
     try {
         # Time slot buckets
         my $buckets = {};
-        $self->do_capture($buckets, $timeout, $starttime);
+        while ($self->do_capture($buckets, $timeout, $starttime)) { }
```
perlpunk:
```suggestion
        1 while $self->do_capture($buckets, $timeout, $starttime);
```

backend/baseclass.pm
okurz: If you prefer I can squash some commits. I pushed the commits in the order of safe steps I executed keeping tests passing in each step

backend/baseclass.pm
okurz: done

backend/baseclass.pm
```diff
@@ -923,14 +923,7 @@ sub similiarity_to_reference ($self, $args) {
...
 
-sub set_tags_to_assert ($self, $args) {
-    my $mustmatch = $args->{mustmatch};
-    my $timeout = $args->{timeout} // $bmwqemu::default_timeout;
-
-    # keep only the most recently used images (https://progress.opensuse.org/issues/15438)
-    needle::clean_image_cache();
-
-    # get the array reference to all matching needles
+sub find_needles_with_tags ($mustmatch) {
```
d3flex, line 923: I am nor sure if we follow the convention for private functions. if that so, `s/find_needles_with_tags/_find_needles_with_tags/`

backend/baseclass.pm
```diff
@@ -189,11 +191,9 @@ sub _check_for_still_screen ($self, $now) {
     return 1;
 }
 
-sub do_capture ($self, $timeout = undef, $starttime = undef) {
-    # Time slot buckets
-    my $buckets = {};
-    my $wait_time_limit = $bmwqemu::vars{_CHKSEL_RATE_WAIT_TIME} // 30;
-    my $hits_limit = $bmwqemu::vars{_CHKSEL_RATE_HITS} // 30_000;
+sub do_capture ($self, $buckets = undef, $timeout = undef, $starttime = undef) {
```
d3flex: I think it makes more sense to initialize `$buckets` with an empty `{}`.  Looking the whole function and the old implementation, it doesnt look to expect undef. and it is also clear what type is expected.

backend/baseclass.pm
okurz: Good question. I think we don't export any functions so I guess we don't need to do and maybe shouldn't that for free functions and reserve that for methods

backend/baseclass.pm
okurz: Good point. Creating a hashref by default will not do the right think. I think it's a better idea to make `$buckets` a mandatory argument.

PR #2743, opened 2025-09-04
Add variable to indicate IPv6 availability to tests

nicksinger:
As we cannot ensure IPv6 availability on all hosts connected to o3 and other openQA instances, we can only signal it to the test. The test can then check for this variable and notify the user about a needed change (e.g. add a worker-class requirement to the test, skip the test with a warning, etc).

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -867,6 +867,10 @@ sub start_qemu ($self) {
         }
     }
 
+    my $v6checkhost = $vars->{IPV6_CHECKHOST} // 'opensuse.org';
+    my ($e, $out) = osutils::run('ping', '-6', '-c', 1, '-w', 1, '-W', 1, $v6checkhost);
+    $vars->{IPV6_AVAILABLE} = $e ? 0 : 1;
+
     bmwqemu::save_vars();    # update variables
 
     mkpath($basedir);
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -871,6 +871,10 @@ sub start_qemu ($self) {
         }
     }
 
+    my $v6checkhost = $vars->{IPV6_CHECKHOST} // 'opensuse.org';
+    my ($e, $out) = osutils::run('ping', '-6', '-c', 1, '-w', 1, '-W', 1, $v6checkhost);
+    $vars->{IPV6_AVAILABLE} = $e ? 0 : 1;
+
     bmwqemu::save_vars();    # update variables
 
     mkpath($basedir);
--- a/doc/backend_vars.md
+++ b/doc/backend_vars.md
@@ -216,6 +216,8 @@ Supported variables per backend
 | QEMU_VIDEO_DEVICE | string | virtio-gpu-pci on ARM, VGA otherwise | Video device to use with VM (using -device, not -vga). Can have options appended e.g. "virtio-gpu-gl,edid=on", but it is better to set QEMU_VIDEO_DEVICE_OPTIONS. See qemu docs and https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ for valid choices |
 | QEMU_VIDEO_DEVICE_OPTIONS | string | none | Additional options for QEMU_VIDEO_DEVICE (comma-separated). Will be appended after automatically-generated resolution setting options on devices that support EDID |
 | SAVE_STORAGE_TIMEOUT | integer | 900 | Timeout for saving one storage volume in `save_storage` test API function. |
+| IPV6_CHECKHOST | string | opensuse.org | Host to ping (ICMP) to determine if IPv6 is working on that worker |
+| IPV6_AVAILABLE | integer |  | Read-only variable which indicates if a ping via IPv6 to `IPV6_CHECKHOST` ran successfuly shortly before the test started running |
 
 ## SVIRT backend
 
--- a/doc/backend_vars.md
+++ b/doc/backend_vars.md
@@ -222,6 +222,8 @@ Supported variables per backend
 | QEMU_VIDEO_DEVICE | string | virtio-gpu-pci on ARM, VGA otherwise | Video device to use with VM (using -device, not -vga). Can have options appended e.g. "virtio-gpu-gl,edid=on", but it is better to set QEMU_VIDEO_DEVICE_OPTIONS. See qemu docs and https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ for valid choices |
 | QEMU_VIDEO_DEVICE_OPTIONS | string | none | Additional options for QEMU_VIDEO_DEVICE (comma-separated). Will be appended after automatically-generated resolution setting options on devices that support EDID |
 | SAVE_STORAGE_TIMEOUT | integer | 900 | Timeout for saving one storage volume in `save_storage` test API function. |
+| IPV6_CHECKHOST | string | opensuse.org | Host to ping (ICMP) to determine if IPv6 is working on that worker |
+| IPV6_AVAILABLE | integer |  | Read-only variable which indicates if a ping via IPv6 to `IPV6_CHECKHOST` ran successfuly shortly before the test started running |
 
 ## SVIRT backend
 


nicksinger:
Hm, I think this fails because there is no `ping`-binary in the fullstack test environment. What would be the easiest to add it there? 

I also considered using a tcp-client from IO::Socket::INET but this seems rather complex for such a simple operation. Opinions?

backend/qemu.pm
```diff
@@ -866,6 +866,9 @@ sub start_qemu ($self) {
         }
     }
 
+    my $v6checkhost = $vars->{IPV6CHECKHOST} // 'opensuse.org';
+    $vars->{IPV6AVAILABLE} = !runcmd('ping', '-6', '-c', 1, '-w', 1, '-W', 1, $v6checkhost);
```
perlpunk: That might end up as a boolean `true` or `false` in the vars.json because in perl versions >= 5.38 that `!` will automatically convert the result to a boolean (in older versions booleans kind of existed, but were not preserved, so a JSON module would only see a 1 or '').
So this might be fine, I just wanted to add this, because the behaviour would change between current Tumbleweed and Leap 15.6.

nicksinger: I've adjusted it to use a ternary operator which makes it a little more explicit and IMHO even more readable.

perlpunk: I actually liked it, very perlish :)
Just wanted to mention it

okurz (changes requested):
Please put your PR details into the commit message

doc/backend_vars.asciidoc
```diff
@@ -51,6 +51,8 @@ VIDEO_ENCODER_BLOCKING_PIPE;boolean;0;Whether the pipe for writing data to the v
 DEFAULT_CLICK_SLEEP;float;0.15;Default single click time in seconds
 DEFAULT_DCLICK_SLEEP;float;0.10;Default double/triple click time in seconds (both press time and interval between clicks)
 GIT_CACHE_DIR;string;;If set enables locally caching Git repositories in the specified directory when handling Git URLs in variables like `CASEDIR` and wheels
+IPV6CHECKHOST;string;opensuse.org;Host to ping (ICMP) to determine if IPv6 is working on that worker
+IPV6AVAILABLE;integer;;Read-only variable which indicates if a ping via IPv6 to `IPV6CHECKHOST` ran successfuly shortly before the test started running
```
kalikiana: Just a little nitpick, but `IPV6_` with an underscore would fit the naming style better I think

backend/qemu.pm
```diff
@@ -866,6 +866,9 @@ sub start_qemu ($self) {
         }
     }
 
+    my $v6checkhost = $vars->{IPV6CHECKHOST} // 'opensuse.org';
+    $vars->{IPV6AVAILABLE} = runcmd('ping', '-6', '-c', 1, '-w', 1, '-W', 1, $v6checkhost) ? 0 : 1;
```
Martchus: Doing this within the QEMU backend means the variable will only be available when using that specific backend. This is not ideal and also not in-line with adding the documentation under `.COMMON backend`.

I would also avoid running the command and setting the variable unconditionally before each test run as we don't always read the variable (and in fact don't read it at all within the scope of os-autoinst itself). So it would make more sense to add a test API function, e.g. `check_ipv6_available` that can be called as needed.

Maybe `check_ipv6_internet_connectivity` would also be a better name as this check will fail if IPv6 is available but but only for e.g. local connections.

Martchus (changes requested):

perlpunk:
Seemingly unrelated errors in 34-git.t that I can't reproduce locally, although I have the same git version 2.51.0 https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2743/os-autoinst/openSUSE_Tumbleweed/x86_64
```
[   96s] 3: # Subtest: failing clone
[   96s] 3:     not ok 1 - Error message when trying to clone wrong git hash
[   96s] 3: 
[   96s] 3:     #   Failed test 'Error message when trying to clone wrong git hash'
[   96s] 3:     #   at t/34-git.t line 67.
[   96s] 3:     # expecting: Regexp ((?^u:Could not find 'abcdef' in complete history in cloned Git repository "file\:\/\/\/tmp\/34\-git\.t\-h6m_\/tmpgitrepo\#abcdef"))
[   96s] 3:     # found: Unable to clone Git repository "file:///tmp/34-git.t-h6m_/tmpgitrepo#abcdef" specified via CASEDIR (see log for details) at /home/abuild/rpmbuild/BUILD/os-autoinst-5.1757079794.ccf4f05-build/os-autoinst-5.1757079794.ccf4f05/OpenQA/Isotovideo/Utils.pm line 164.
[   96s] 3:     not ok 2 - manual git fetch for revspec was attempted
[   96s] 3: 
[   96s] 3:     #   Failed test 'manual git fetch for revspec was attempted'
... and more errors
```

backend/qemu.pm
nicksinger: I adjusted the docs. Apparently I missed the correct table.

Regarding an test API change; I would like to avoid this. I understand that you mainly see variables as inputs for the backend but in this case I would argue that it should be fine to use them as output from the backend for the test (merely to provide capability information). What do others think?

Martchus: Ah, so you're setting this variable to check its value later via `vars.json`. I guess this makes sense then.

But are you sure you only want this for QEMU-based tests?

doc/backend_vars.asciidoc
```diff
@@ -211,6 +211,8 @@ WORKER_HOSTNAME;string;undef;Worker hostname
 QEMU_VIDEO_DEVICE;string;virtio-gpu-pci on ARM, VGA otherwise;Video device to use with VM (using -device, not -vga). Can have options appended e.g. "virtio-gpu-gl,edid=on", but it is better to set QEMU_VIDEO_DEVICE_OPTIONS. See qemu docs and https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ for valid choices
 QEMU_VIDEO_DEVICE_OPTIONS;string;none;Additional options for QEMU_VIDEO_DEVICE (comma-separated). Will be appended after automatically-generated resolution setting options on devices that support EDID
 SAVE_STORAGE_TIMEOUT;integer;900;Timeout for saving one storage volume in `save_storage` test API function.
+IPV6_CHECKHOST;string;opensuse.org;Host to ping (ICMP) to determine if IPv6 is working on that worker
+IPV6_AVAILABLE;integer;;Read-only variable which indicates if a ping via IPv6 to `IPV6CHECKHOST` ran successfuly shortly before the test started running
```
d3flex: I guess `IPV6CHECKHOST` should be `IPV6_CHECKHOST`?!

backend/qemu.pm
nicksinger: > Ah, so you're setting this variable to check its value later via `vars.json`

I intend to fetch this bit of information from within a running test with `get_var` - not sure if you consider this as »check its value later via `vars.json`« ;)

> But are you sure you only want this for QEMU-based tests?

Yes, for now. I looked into the baseclass (`backend::virt`) but it is used for a lot of other architectures where this check would not work or possible even produce wrong results (e.g. `pvm_hmc` and `spvm`). But you're right, it would also be feasible on other virtualized environments.

okurz (approved):

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/17611367599/job/50033871288?pr=2743#step:3:475
```
#   Undefined subroutine &backend::qemu::run called at /opt/script/../backend/qemu.pm line 872.
```

backend/qemu.pm
```diff
@@ -27,6 +27,7 @@ use Data::Dumper;
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use OpenQA::Qemu::Proc;
 use Socket;
+use osutils 'run';
```
perlpunk: Tests are failing.
The problem here is that you are importing a function into the namespace that is already expected to be in that namespace, e.g. backend::driver is calling `$self->{backend}->run(...)`.

That can be avoided by not importing it and using the fully qualified package name:
```suggestion
use osutils ();
# later
my (...) = osutils::run(...);
```

d3flex: btw the `osutils` is already used in L24, so this line shouldnt be needed at all

nicksinger: Thanks! Guess that would have taken me some time to understand :) I currently also try to run these tests locally (just 2 failing left)

perlpunk: You are still importing it (in the latest version of this PR).
Do not add it to the `use osutils` line and use the package name:
```
my (...) = osutils::run(...);
```

nicksinger: yes sorry, this was to verify my local testing environment produces the same result as the CI. Anyhow, changed now with https://github.com/os-autoinst/os-autoinst/commit/9bf1ae48879fa0be6ee2d4928c78734ad8608af8 - it still fails locally but maybe my setup is still wrong. So pushing again…

gauravpathak:
Why this PR is still open?
From the conversation, it seems that all review comments were resolved.

Martchus (approved):
There are conflicts to be resolved, though.

gauravpathak (approved):

kalikiana:
@mergifyio rebase

nicksinger:
I rebased myself and moved the docs from the deprecated asciidoc into the new md-file. I unfortunately have no idea how to fix the OBS-checks and ci/test tells me to "please update backend_vars" (which [I did…](https://github.com/os-autoinst/os-autoinst/pull/2743/files#diff-17bf552be43252119a8c8d70ba4db7f92d092048af55ca87771794409d9ee93eR220))

doc/backend_vars.md
```diff
@@ -216,6 +216,8 @@ Supported variables per backend
 | QEMU_VIDEO_DEVICE | string | virtio-gpu-pci on ARM, VGA otherwise | Video device to use with VM (using -device, not -vga). Can have options appended e.g. "virtio-gpu-gl,edid=on", but it is better to set QEMU_VIDEO_DEVICE_OPTIONS. See qemu docs and https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ for valid choices |
 | QEMU_VIDEO_DEVICE_OPTIONS | string | none | Additional options for QEMU_VIDEO_DEVICE (comma-separated). Will be appended after automatically-generated resolution setting options on devices that support EDID |
 | SAVE_STORAGE_TIMEOUT | integer | 900 | Timeout for saving one storage volume in `save_storage` test API function. |
+| IPV6_CHECKHOST | string | opensuse.org | Host to ping (ICMP) to determine if IPv6 is working on that worker |
+| IPV6_AVAILABLE | integer || Read-only variable which indicates if a ping via IPv6 to `IPV6_CHECKHOST` ran successfuly shortly before the test started running |
```
perlpunk:
```suggestion
| IPV6_AVAILABLE | integer |  | Read-only variable which indicates if a ping via IPv6 to `IPV6_CHECKHOST` ran successfuly shortly before the test started running |
```
This makes `t/04-check_vars_docu.t` pass for me locally

perlpunk (commented):
I can make `t/18-backend-qemu.t` pass locally by increasing the time limit:
```
use OpenQA::Test::TimeLimit '15';
```
but maybe check why it takes longer now

doc/backend_vars.md
nicksinger: oh, good finding, thanks!

nicksinger:
> but maybe check why it takes longer now

I mean, it makes sense that it take 1s longer in case of a timeout (which I assume happens in GH-actions) but 10s seems excessive indeed - I will check if ping maybe takes longer then expected.

PR #2750, opened 2025-09-16
testapi: allow passing max_interval through (assert_)script_run

AdamWill:
Sometimes we might want to slow-type assert_script_run or script_run. Currently that's not really possible. This makes it possible by letting the max_interval arg be passed all the way through to testapi::type_string via distribution::script_run.


okurz (approved):
If you don't need a version bump to check that you use this functionality only in any *new* version I am ok without a version bump as is backward compatible.

distribution.pm
```diff
@@ -112,25 +112,26 @@ sub script_run ($self, $cmd, @args) {
         {
             timeout => $bmwqemu::default_timeout,
             output => '',
-            quiet => undef
+            quiet => undef,
+            max_interval => testapi::DEFAULT_MAX_INTERVAL
```
kalikiana, line 116: Would be good to also document the new option

Martchus (approved):

distribution.pm
AdamWill: true, good point. will do.

AdamWill: [Doc PR](https://github.com/os-autoinst/os-autoinst/pull/2752)

PR #2752, opened 2025-09-18
Document max_interval (and quiet) in (assert_)script_run

AdamWill:
As requested by @kalikiana , this documents max_interval in assert_script_run and script_run (both testapi and distribution). We also document quiet in the testapi subs, as that wasn't documented there either.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -909,7 +909,7 @@ sub _handle_script_run_ret {    # no:style:signatures
 
 =head2 assert_script_run
 
-  assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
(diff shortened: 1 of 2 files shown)


testapi.pm
```diff
@@ -921,6 +921,10 @@ Use C<script_run> instead if C<$cmd> may fail.
 
 C<$fail_message> is returned in the die message if specified.
 
+C<$quiet> and C<$max_interval> are passed through to C<< $distri->script_run >>. In
```
d3flex, line 924: s/C<< $distri->script_run >>/C<$distri->script_run>/

d3flex (changes requested):

perlpunk (approved):

testapi.pm
perlpunk: The double `<<>>` are needed because of the inner `>`

https://perldoc.perl.org/perlpod

d3flex (approved):

PR #2753, opened 2025-09-22
Add nftables function to os-autoinst-setup-multi-machine

aylen384:


script/os-autoinst-setup-multi-machine
```diff
@@ -157,7 +196,7 @@ main() {
     }
     ensure_ip_forwarding
     install_packages
-    configure_firewall
+    configure_nftables
```
Martchus: This leaves `configure_firewall` as unused function. Maybe there should be a switch to select which firewall to use?

okurz: Yes, for now I would prefer to keep the optional support for wicked/networkmanager/firewalld/nftables in. Eventually we can remove obsolete parts

Martchus (approved):

PR #2754, opened 2025-09-22
Improve os-autoinst-setup-multi-machine

okurz:
* os-autoinst-setup-multi-machine: Only call zypper when necessary
* os-autoinst-setup-multi-machine: Only check network interface if not supplied

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -40,7 +40,8 @@ ensure_ip_forwarding() {
 
 install_packages() {
     command -v retry > /dev/null || zypper -n in retry
-    retry -e -s 30 -r 7 -- sh -c "zypper ref && zypper -n in openvswitch os-autoinst-openvswitch firewalld libcap-progs"
+    pkgs=(openvswitch os-autoinst-openvswitch firewalld libcap-progs)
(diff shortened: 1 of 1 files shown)


script/os-autoinst-setup-multi-machine
```diff
@@ -145,6 +145,13 @@ configure_openvswitch() {
     systemctl restart openvswitch os-autoinst-openvswitch
 }
 
+determine_ethernet_interface() {
+    if [[ -n "$ethernet" ]]; then
+        return
+    fi
+    ethernet=$(ip -json route show default | jq -r '.[]|.dev')
```
Martchus, line 152: This could be a one-liner:
```suggestion
    if [[ -n "$ethernet" ]]; then
        return
    fi
    ethernet=${ethernet:-$(ip -json route show default | jq -r '.[]|.dev')}
```
(Not sure whether it'll be easier to read for anyone not that familiar with Bash.)

kalikiana (approved):

script/os-autoinst-setup-multi-machine
okurz: I guess you meant to replace the if-part as well?

okurz: I created https://github.com/os-autoinst/os-autoinst/pull/2755 for this

PR #2758, opened 2025-09-29
Don't overwrite firewall xml

aylen384:
Ticket: https://progress.opensuse.org/issues/189144

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -45,29 +45,31 @@ install_packages() {
 }
 
 configure_firewall() {
+    echo "Now configuring firewalld..."
     systemctl enable --now firewalld
     default_zone=$(firewall-cmd --get-default-zone)
+    echo "- default_zone=$default_zone"
+    echo "- Creating isotovideo firewall service"
     firewall-cmd --permanent --get-services | grep -q isotovideo && firewall-cmd --permanent --delete-service=isotovideo
     firewall-cmd --permanent --new-service isotovideo
     for i in $(seq 1 "$instances"); do firewall-cmd --permanent --service=isotovideo --add-port=$((i * 10 + 20003))/tcp; done
     firewall-cmd --permanent --zone="$zone" --add-service=isotovideo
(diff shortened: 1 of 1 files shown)


script/os-autoinst-setup-multi-machine
```diff
@@ -45,29 +45,31 @@ install_packages() {
 }
 
 configure_firewall() {
+    echo "Now configuring firewalld..."
     systemctl enable --now firewalld
     default_zone=$(firewall-cmd --get-default-zone)
+    echo "- default_zone=$default_zone"
+    echo "- Creating isotovideo firewall service"
```
kalikiana, line 52: Is this and other logging meant to be for testing? I would have thought this should be a draft in that case, and not be merged?

Martchus: Those log messages seem generally useful to me.

aylen384: No in general because it otherwise prints 20x "success" to the console and the user can only guess what that means

script/os-autoinst-setup-multi-machine
```diff
@@ -45,29 +45,31 @@ install_packages() {
...
     firewall-cmd --permanent --get-services | grep -q isotovideo && firewall-cmd --permanent --delete-service=isotovideo
     firewall-cmd --permanent --new-service isotovideo
     for i in $(seq 1 "$instances"); do firewall-cmd --permanent --service=isotovideo --add-port=$((i * 10 + 20003))/tcp; done
     firewall-cmd --permanent --zone="$zone" --add-service=isotovideo
     if [[ $default_zone != "$zone" ]]; then
+        echo "- setting new default_zone=$zone"
```
kalikiana:
```suggestion
        echo "- Setting new default_zone=$zone"
```

script/os-autoinst-setup-multi-machine
kalikiana: I guess I wouldn't include the first and last one, but no super strong opinion

kalikiana (approved):

PR #2796, opened 2026-01-15
Fix snapshot overlay mechanism to avoid duplication

d3flex:
When reverting to snapshots in TESTDEBUG mode, qemu would fail with "Could not open backing file" because it tries to create an overlay image using the same filename as its backing file.

Additionally, the last_overlay_id counter was not persisted to qemu_state.json, causing it to reset to 0 after each run. This led to duplicate overlay filenames when creating new snapshots (e.g., creating overlay1 twice with different snapshot sequence numbers), further breaking, likely, the overlay chain structure.

The fix involves three changes:
(1) Remove the loop from
BlockDev::gen_unlink_list so it only returns the specific overlay being recreated, not its backing files.
(2) Add overlay(undef) in BlockDevConf::revert_to_snapshot to break the forward chain link and prevent deleted overlays from appearing in serialized state, resulting in persistent overlays.
(3) Add last_overlay_id to DriveDevice's `_to_map` and `_from_map` serialization to preserve the overlay counter across VM restarts, ensuring unique overlay filenames. Meaning each block in qemu_state.json creates a new incremental overlay associated with each snapshot

related-issue: https://progress.opensuse.org/issues/193366

### DIFF
--- a/OpenQA/Qemu/BlockDev.pm
+++ b/OpenQA/Qemu/BlockDev.pm
@@ -168,8 +168,6 @@ this will return it in an array so that the caller can unlink it.
 
 sub gen_unlink_list ($self) {
     return () unless $self->needs_creating;
-    return ($self->file, $self->backing_file->gen_unlink_list())
-      if defined $self->backing_file;
     return ($self->file);
 }
 
--- a/OpenQA/Qemu/DriveDevice.pm
+++ b/OpenQA/Qemu/DriveDevice.pm
@@ -151,7 +151,8 @@ sub _to_map ($self) {
         bootindex => $self->bootindex,
         serial => $self->serial,
         id => $self->id,
-        num_queues => $self->num_queues};
+        num_queues => $self->num_queues,
+        last_overlay_id => $self->last_overlay_id};
 }
 
 sub _from_map ($self, $map, $cont_conf, $snap_conf) {
@@ -166,7 +167,8 @@ sub _from_map ($self, $map, $cont_conf, $snap_conf) {
       ->bootindex($map->{bootindex})
       ->serial($map->{serial})
       ->id($map->{id})
-      ->num_queues($map->{num_queues});
+      ->num_queues($map->{num_queues})
+      ->last_overlay_id($map->{last_overlay_id} // 0);
 }
 
 sub CARP_TRACE ($self) { 'OpenQA::Qemu::DriveDevice(' . ($self->id || '') . ')' }
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -303,8 +303,7 @@ $bdc = $proc->blockdev_conf;
 
 $ss = $ssc->revert_to_snapshot('snapshot 1');
 is($ss->sequence, 1, 'Returned snapshot sequence number');
-$bdc->for_each_drive(sub {
-        my $drive = shift;
+$bdc->for_each_drive(sub ($drive) {
         my $unlinks = $bdc->revert_to_snapshot($drive, $ss);
         is(scalar(@$unlinks), 9, 'Correct number of overlay files need unlinking for ' . $drive->id);
 });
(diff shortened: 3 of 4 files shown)


OpenQA/Qemu/BlockDev.pm
```diff
@@ -168,8 +168,6 @@ this will return it in an array so that the caller can unlink it.
 
 sub gen_unlink_list ($self) {
```
Martchus, line 169: I *guess* your change makes generally sense.

Note that this function was introduced in "The Great QEMU Rewrite of 2018" (see c670720d452b536eed3544987f116fe01ee0a52f) so I'm nevertheless not quite sure what its general intention is and whether some of its usages may rely on taking the backing file into account.

OpenQA/Qemu/DriveDevice.pm
```diff
@@ -160,13 +161,16 @@ sub _from_map ($self, $map, $cont_conf, $snap_conf) {
...
+
     return $self->drive($drive)
       ->model($map->{model})
       ->paths(\@paths)
       ->bootindex($map->{bootindex})
       ->serial($map->{serial})
       ->id($map->{id})
-      ->num_queues($map->{num_queues});
+      ->num_queues($map->{num_queues})
+      ->last_overlay_id($restored_overlay_id);
```
Martchus:
```suggestion
    return $self->drive($drive)
      ->model($map->{model})
      ->paths(\@paths)
      ->bootindex($map->{bootindex})
      ->serial($map->{serial})
      ->id($map->{id})
      ->num_queues($map->{num_queues})
      ->last_overlay_id($map->{last_overlay_id} // 0);
```

Or even just:

```suggestion
    return $self->drive($drive)
      ->model($map->{model})
      ->paths(\@paths)
      ->bootindex($map->{bootindex})
      ->serial($map->{serial})
      ->id($map->{id})
      ->num_queues($map->{num_queues})
      ->last_overlay_id($map->{last_overlay_id} // 0);
```

Martchus (commented):
I guess this makes generally sense so far.

OpenQA/Qemu/BlockDev.pm
Martchus: Looks like this is only called in `init_blockdev_images` for an upfront initialization and when restoring snapshots. So your change will probably not disturb anything.

t/18-qemu.t
```diff
@@ -593,4 +595,64 @@ subtest revert_to_snapshot => sub {
...
+        SCSICONTROLLER => 'virtio-scsi-device');
+
+    $proc = qemu_proc('-static-args', \%vars);
+    $ssc = $proc->snapshot_conf;
+    $bdc = $proc->blockdev_conf;
+
+    for my $i (1 .. 3) {
+        my $ss = $ssc->add_snapshot("snapshot$i");
+        $bdc->for_each_drive(sub {
+                my $drv = shift;
```
Martchus: Use signatures in new code. This also counts for the other functions (except sub tests).

okurz (changes requested):
https://github.com/os-autoinst/os-autoinst/pull/2796#pullrequestreview-3686219968

t/18-qemu.t
d3flex: Done. I will add another commit for the rest as well. I hope it is fine

okurz (approved):

t/18-qemu.t
```diff
@@ -593,4 +591,59 @@ subtest revert_to_snapshot => sub {
...
+
+        });
+    }
+    $bdc->mark_all_created();
+
+    my %overlay_files;
+    subtest 'check overlay chain' => sub {
+        $bdc->for_each_drive(sub ($drive) {
+                $drive->for_each_overlay(sub {
+                        my $overlay = shift;
```
Martchus: This function is still not using signatures.

Martchus (approved):

PR #2808, opened 2026-02-21
test: move author tests to xt/ and optimize dependencies

okurz:
Consolidate developer-only "author tests" (linting, commit message checks,
spellchecking) into the xt/ directory, following Perl conventions.
This ensures a cleaner set of core unit tests and reduces the mandatory
dependency list for running basic tests.

- Moved gitlint, yamllint, and spellcheck from test_requires to
  devel_requires in dependencies.yaml.
- Removed redundant test-local-* CTest targets from
  cmake/test-targets.cmake.
- Created Perl test wrappers in xt/ for gitlint, yamllint, python style,
  shellcheck, perl style, bash syntax, and spellcheck.
- Updated .github/workflows/commit-message-checker.yml to use the new
  xt/ test wrapper.

### DIFF
--- a/cmake/test-targets.cmake
+++ b/cmake/test-targets.cmake
@@ -21,16 +21,17 @@ add_test(
     WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"
 )
 
+# add targets for invoking Perl test suite
+find_program(PROVE_PATH prove)
+
 # add test for YAML syntax
 find_program(YAMLLINT_PATH yamllint)
 if (YAMLLINT_PATH)
     add_test(
         NAME test-local-yaml-syntax
-        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-yaml-syntax" "${YAMLLINT_PATH}"
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/tidyall" --select "**/*.{yml,yaml}" --check
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
-else ()
-    message(STATUS "Set YAMLLINT_PATH to the path of the yamllint executable to enable YAML syntax checks.")
 endif ()
 
 # add test for python code style
@@ -38,7 +39,7 @@ find_program(RUFF_PATH ruff)
 if (RUFF_PATH)
     add_test(
         NAME test-local-python-style
-        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-python-style" "${RUFF_PATH}"
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/tidyall" --select "**/*.py" --check
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
 else ()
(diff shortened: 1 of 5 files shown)


perlpunk (approved):

perlpunk (changes requested):
Where do I see the actual author tests running?

https://github.com/os-autoinst/os-autoinst/actions/runs/22365197325/job/64729196403?pr=2808#step:4:51
says 
```
No tests were found!!!
```

Martchus (commented):
And I thought CMake was verbose. Now we have a separate file for each check and each of those files is basically the same and just invoking another script. Well, I guess I'm not against it…

okurz:
> And I thought CMake was verbose. Now we have a separate file for each check and each of those files is basically the same and just invoking another script. Well, I guess I'm not against it…

I have alternatives in mind:
1. Keep it as is
2. Put all into one test file
3. Call the tools as "TidyAll" plugins: I already tried it out and it's basically defining commands and parameters in the TidyAll config file
4. Write any other kind of plugins for common linting thingies

cmake/test-targets.cmake
```diff
@@ -21,24 +21,25 @@ add_test(
     WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"
 )
 
+# add targets for invoking Perl test suite
+find_program(PROVE_PATH prove)
+
 # add test for YAML syntax
 find_program(YAMLLINT_PATH yamllint)
 if (YAMLLINT_PATH)
```
Martchus, line 29: Considering that now tidyall handles locating the executable we might want to avoid the `find_program` call and checking here.

This change of course also means that one now can no longer use `-DYAMLLINT_PATH=/some/path` to override the path. Not a big deal but this actually speaks more for not locating the executable at all on CMake level.

cmake/test-targets.cmake
```diff
@@ -95,51 +96,41 @@ find_program(SHELLCHECK_PATH shellcheck)
...
     add_test(
         NAME test-local-bash-syntax
-	 COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/check-bash-scripts" "${SH_PATH}" "${CMAKE_SOURCE_DIR}"
-	 WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tools/tidyall" --plugins "GenericTransformer / shfmt" -a --check
+        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
-else ()
-    message(STATUS "Set SH_PATH to the path of the shfmt executable to enable bash script syntax checks.")
 endif ()
```
Martchus, line 112: Having all these lines *and* your new tidyall config is a bit much. Maybe you can just use a loop on the CMake-side. (The exact names of the tests on CMake level don't really matter so you could just call them like the tool. Then you could simply loop over the tool names.)

Martchus (changes requested):

okurz:
Multiple improvements for better, separate author tests in xt/ have been extracted from this PR and are merged. I am not convinced about the rest and it's problematic hence closing.

PR #2819, opened 2026-03-07
Stabilize all tests for parallel execution

okurz:
After 10+ years of starting this I am finally confident to have proven stability of running os-autoinst tests in parallel. Verified with 100+ (!) local runs.

Latest evaluation with 60 runs:
count-fail-ratio: Run: 60. Fails: 0. Fail ratio 0±0%. No fails, computed failure probability < 5.00%
mean runtime: 152502±737.78 ms


t/data/tests/tests/softfail_module.pm
```diff
@@ -4,7 +4,7 @@
 use base 'basetest';
 use testapi;
 
-sub run ($self) {
+sub run ($) {
```
perlpunk: The explanation for this change is wrong.

> The 'softfail_module.pm' test data used 'sub run ()'

It used `sub run ($self)`, not `sub run ()`.

This is a signature, if signatures are enabled.

`sub run($)` is a prototype without signatures, and a signature with an unnamed argument if signatures are enabled.

How did you get the error message about an illegal prototype?
Since we decided we want to add `use Mojo::Base -signatures` on top dynamically, we cannot actually compile test modules standalone.

perlpunk (changes requested):

t/12-bmwqemu.t
```diff
@@ -19,8 +19,11 @@ use Mojo::File qw(path);
 use Mojo::JSON qw(decode_json);
 use Cpanel::JSON::XS ();
 use Test::Warnings qw(warning :report_warnings);
+use File::chdir;
 
 my $toplevel_dir = abs_path(dirname(__FILE__) . '/..');
+my $tmpdir = tempdir(CLEANUP => 1);
```
perlpunk: Please use our common tempdir practice as already used in other tests:
```
my $tmpdir = tempdir("/tmp/$FindBin::Script-XXXX");
```
`CLEANUP` is not necessary.

t/10-virtio_terminal.t
```diff
@@ -38,13 +39,17 @@ sub prepare_pipes ($socket_path, $write_buffer = undef) {
...
         };
-        alarm ONE_MINUTE;
+        alarm 5 * ONE_MINUTE;
         my $fd_r = IO::Handle->new();
         my $fd_w = IO::Handle->new();
-        open($fd_r, "<", $pipe_in)
-          or die "Can't open in pipe for writing $!";
-        open($fd_w, ">", $pipe_out)
-          or die "Can't open out pipe for reading $!";
+        until (open($fd_r, "<", $pipe_in)) {
```
perlpunk: I don't understand why this is necessary.

t/data/tests/tests/softfail_module.pm
okurz: but why do many other test modules use `sub run ($) {` as well? I think I discovered this when trying to move perlcritic checks to tidyall. we can leave this commit out if it's wrong or questionable.

t/10-virtio_terminal.t
okurz: I encountered problems executing this test when executed in parallel with others repeatedly. I can look into finding better statistics first to understand why/how this would be exactly needed

t/12-bmwqemu.t
okurz: done

t/data/tests/tests/softfail_module.pm
okurz: anyway, removed

t/10-virtio_terminal.t
okurz: removed

t/data/tests/tests/softfail_module.pm
perlpunk: For historical reasons? :)
The `($)` prototype was actually never checked at all, because prototypes are only checked when called as a function, but not when called as a method.
Ad it is wrong in cases when there is an additional argument, e.g.
```
$self->run($self->{run_args});
```
I think we could just remove all prototypes in test modules under `t/`.

t/14-isotovideo.t
```diff
@@ -36,6 +36,8 @@ $ENV{OS_AUTOINST_GIT_RETRY_COUNT} = 0;
 sub isotovideo (%args) {
     $args{default_opts} //= 'backend=null';
     $args{opts} //= '';
+    my $vnc_port = 90 + ($$ % 100);
```
Martchus, line 41: And interesting way to pick a port. It isn't completely bullet-proof but probably good enough.

Martchus (approved):

t/14-isotovideo.t
okurz: I know that we talked other more elaborated approaches for openQA tests years ago :)

perlpunk (changes requested):
Apparently (see test failures) there are still a lot of tests modifying or adding files in the git checkout. For parallelization they all have to be rewritten to use a tempdir.

Martchus (approved):

okurz:
Blocked by
* https://github.com/os-autoinst/os-autoinst-common/pull/91

kalikiana (approved):

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/25692114867/job/75430991297?pr=2819#step:3:480
```
4: [19:24:20] t/01-test_needle.t .......................... 
4: Dubious, test returned 1 (wstat 256, 0x100)
4: All 90 subtests passed 
```

perlpunk:
At some point, https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2819/os-autoinst/openSUSE_Tumbleweed/x86_64 failed, the check is failing, although the build is passing right now.
Look at the job history to see that it was failing:
https://build.opensuse.org/packages/os-autoinst/job_history/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2819/openSUSE_Tumbleweed/x86_64

So I assume it is still racy.

perlpunk:
Tests are still failing

perlpunk (changes requested):
Tests are failing

okurz:
After:
* #3010

okurz:
after longer time again ready after I could fix missing statement coverage and fix code according to recent perlcritic rule additions

PR #2827, opened 2026-03-16
feat: Prefer `4m` version of OVMF firmware image

Martchus:
The version without `4m` has been removed in the latest version `202602` of the `qemu-ovmf-x86_64` package as provided by TW. So we should probably use the `4m` version instead.

For the best consistency, this change makes the `4m` version the most preferred version. Since the `4m` files are also present in `qemu-ovmf-x86_64` version `202502` as provided by Leap 16 this means we will use the `4m` version for most test runs with this commit.

Related ticket: https://progress.opensuse.org/issues/197930

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -45,6 +45,7 @@ our $screenshotpath = 'qemuscreenshot';
 # the kraxel.org nightly packages, third is Fedora's edk2-ovmf package,
 # fourth is Debian's ovmf package.
 our @ovmf_locations = (
+    '/usr/share/qemu/ovmf-x86_64-ms-4m-code.bin',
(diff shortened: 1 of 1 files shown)


okurz (approved):

d3flex (approved):

bmwqemu.pm
```diff
@@ -45,6 +45,7 @@ our $screenshotpath = 'qemuscreenshot';
 # the kraxel.org nightly packages, third is Fedora's edk2-ovmf package,
 # fourth is Debian's ovmf package.
 our @ovmf_locations = (
+    '/usr/share/qemu/ovmf-x86_64-ms-4m-code.bin',
```
rclobus, line 48: Please update the comment above. The count is now shifted by one

Martchus: https://github.com/os-autoinst/os-autoinst/pull/2985

PR #2849, opened 2026-03-26
feat: fail explicitly if always_rollback is not supported

okurz:
Motivation:
Always rollback is silently ignored on backends that do not support
snapshots. This can lead to unexpected test results when the test author
expects a clean state.

Design Choices:
Introduce a new feature switch FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED to
optionally enforce snapshot support when always_rollback is used.

Benefits:
Provides better feedback to test authors and ensures that test
requirements are met by the backend.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -536,6 +536,9 @@ sub runalltests () {
     for (my $testindex = 0; $testindex <= $#testorder; $testindex++) {
         my $t = $testorder[$testindex];
         my $flags = $t->test_flags();
+        if ($flags->{always_rollback} && !$snapshots_supported && $bmwqemu::vars{FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED}) {
+            die "always_rollback requested but snapshots are not supported by the backend\n";
+        }
         my $fullname = $t->{fullname};
 
         if (!$vmloaded && $fullname eq $firsttest) {
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -232,6 +232,15 @@ subtest 'test always_rollback flag' => sub {
         is $reverts_done, 0, 'no snapshots loaded after fatal failure';
         is $snapshots_made, 0, 'no snapshots made after fatal failure';
     };
+    snapshot_subtest 'fails if snapshots are not supported and FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED is set' => sub {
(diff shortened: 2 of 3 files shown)


okurz:
@Vogtinator

Martchus (approved):

t/08-autotest.t
```diff
@@ -580,10 +589,10 @@ subtest 'lua_use' => sub {
     is ref($lua_vars->{assert_script_run}), 'CODE', 'Check importing functions';
 
     autotest::_lua_use('testlib');
-    is $lua_vars->{testfunc1}(), 42, 'Exported function is imported';
+    stdout_like { is $lua_vars->{testfunc1}(), 42, 'Exported function is imported' } qr/testfunc1/, 'testfunc1 outputs to stdout';
```
d3flex: these changes seems irrelevant. Can be removed?

okurz: yes, they are not related. Reverted

autotest.pm
```diff
@@ -536,6 +536,9 @@ sub runalltests () {
     for (my $testindex = 0; $testindex <= $#testorder; $testindex++) {
         my $t = $testorder[$testindex];
         my $flags = $t->test_flags();
+        if ($flags->{always_rollback} && !$snapshots_supported && $bmwqemu::vars{FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED}) {
+            die "always_rollback requested but snapshots are not supported by the backend\n";
```
kalikiana, line 540:
```suggestion
            die "always_rollback requested but snapshots are not supported by the backend (this is an error because FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED is set)\n";
```

d3flex: I will approve. I guess you can open a PR for that if you think it is worth it

d3flex (approved):

Vogtinator:
> @Vogtinator

I'm a bit late, but +1!

https://progress.opensuse.org/issues/201633 (and the INVALID product bug) is another one that would've been prevented by this.

Is there a plan or tracking somewhere when this feature switch will be flipped?

okurz:
> Is there a plan or tracking somewhere when this feature switch will be flipped?

There were no plans yet. But your referenced ticket is a good motivation. I will propose a PR.

Done: https://github.com/os-autoinst/os-autoinst/pull/2948

PR #2850, opened 2026-03-26
feat: implement retry test flag

okurz:
Motivation:
Allow test modules to be retried automatically upon failure when milestones
are available. This is useful for dealing with flaky tests or environmental
issues.

Design Choices:
- Added a new `retry` test flag that specifies the number of additional
  attempts.
- Modified `runalltests` in `autotest.pm` to include a retry loop.
- Retries are only attempted if snapshots are supported and a milestone
  exists.
- The test object is reset for each retry attempt while preserving `details`
  for debugging.
- Rollback to the last milestone and console reset are performed before each
  retry.

Benefits:
Increased test suite robustness by handling transient failures automatically.
Provides better integration with existing snapshot/rollback features.

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -373,6 +373,7 @@ sub set_current_test ($test) {
           {
             name => $current_test->{name},
             full_name => $current_test->{fullname},
+            attempt => $current_test->{attempt} // 0,
           }
         : {});
 }
@@ -562,6 +563,7 @@ sub runalltests () {
         my $error;
 
         while (1) {
+            $t->{attempt} = $attempt;
             bmwqemu::modstate "starting $name $t->{script}" . ($attempt ? " (retry $attempt/$retry_count)" : '');
             $t->start();
 
--- a/autotest.pm
+++ b/autotest.pm
@@ -557,17 +557,38 @@ sub runalltests () {
         }
 
         my $name = $t->{name};
-        bmwqemu::modstate "starting $name $t->{script}";
-        $t->start();
-
-        # avoid erasing the good vm snapshot
-        if ($snapshots_supported && (($bmwqemu::vars{SKIPTO} || '') ne $fullname) && $bmwqemu::vars{MAKETESTSNAPSHOTS}) {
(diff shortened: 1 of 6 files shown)


autotest.pm
```diff
@@ -551,17 +551,38 @@ sub runalltests () {
...
+                make_snapshot($t->{fullname});
+            }
+
+            $error = undef;
+            try { $t->runtest }
+            catch ($e) { $error = $e }
+
+            last unless $error;
+            last if $attempt >= $retry_count;
+            last unless $snapshots_supported && $last_milestone && !$flags->{no_rollback};
```
perlpunk:
```suggestion
            last if !$snapshots_supported || !$last_milestone || $flags->{no_rollback};
```

d3flex (dismissed):
first this should be marked as `breaking_changes`. there are many factors and implications under consideration, and tests will be hard to cover all

Then this will have serious performance implications (it needs at least a cap flag to prevent 100 retries) and likely it will make debugging harder. for example `make_snapshot($t->{fullname})` runs on every retry, and redundantly overwriting the snapshot which destroying the previous failing VM state

autotest.pm
```diff
@@ -551,17 +551,38 @@ sub runalltests () {
         }
 
         my $name = $t->{name};
-        bmwqemu::modstate "starting $name $t->{script}";
-        $t->start();
+        my $retry_count = $flags->{retry} // 0;
+        my $attempt = 0;
```
perlpunk, line 575: Possibly the attempt number should be saved in the test module, the same as the module name and step number, so that we will be able to link autoinst-log file lines to module steps (which we don't do yet, but it's planned).

okurz: ok, done. Saved as "attempt", defaulting to 0.

Martchus (approved):

Martchus (approved):

perlpunk:
This is still a draft and @Vogtinator hasn't commented yet

PR #2855, opened 2026-04-01
refactor: move Perl modules and OpenQA to lib/ directory

okurz:
Motivation:
Usually in Perl projects `.pm` files should be in a `lib/` subfolder.
Multiple `.pm` files and the `OpenQA` directory were at the top level
of the repository. They need to be moved to adhere to standard Perl
project structure.

Design Choices:
- Moved all top-level `.pm` files and the `OpenQA` directory to `lib/`.
- Created backward compatibility symlinks at the root to avoid breaking
  existing users, scripts, or integrations.
- Updated `@INC` and `PERL5LIB` in relevant scripts to include `lib/`
  and find modules directly from there.
- Ensure CMake configuration handles the moved Perl files appropriately.

Benefits:
- Better adherence to standard Perl project structure.
- Cleaner top-level directory layout.


okurz:
was closed due to rebase gone wrong, reopening

PR #2856, opened 2026-04-03
fix(distribution): make PRETTY_SERIAL_MARKER reboot-safe

okurz:
Motivation:
When PRETTY_SERIAL_MARKER=1 is used, os-autoinst relies on PROMPT_COMMAND being
set in the SUT shell. After a reboot or a new console login, this environment
variable is lost, but os-autoinst assumes it is still there due to internal
caching of the detection result. This leads to command timeouts as the expected
serial markers are never printed.

Design Choices:
- Implemented console_selected hook in Distribution to reset the hook status
  whenever a console is selected (e.g. after a login).
- Added a crosscheck in _detect_serial_marker_capability that verifies if the
  hook is still active (using a quick newline check for Level 3) before
  re-installing it.
- Modified install_serial_marker_hook to add the PROMPT_COMMAND to
  ~/.bashrc and ~/.profile, ensuring it persists across reboots and new
  login shells for the same user.

Verified with an adapted "openQA-in-openQA" test scenario that reboots
in the middle of the execution with console commands evaluated before
and after reboot.

Screenshot of two purposely failing commands without visible serial
markers that trigger the post_fail_hook:
mpv-shot0001

Complete video:
[video.webm](https://github.com/user-attachments/assets/6d99c3d8-4f60-435d-a8b3-a6d5a65fd461)

Benefits:
- Pretty serial markers remain functional after reboot

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -466,8 +466,7 @@ Returns:
 
 sub _detect_serial_marker_capability ($self) {
     my $console = testapi::current_console() // 'sut';
-    if ($self->{_serial_marker_level}->{$console}) {
-        my $level = $self->{_serial_marker_level}->{$console};
+    if (my $level = $self->{_serial_marker_level}->{$console}) {
         return $level if $level < 2 || $self->{_serial_marker_hook_installed}->{$console};
 
         $self->install_serial_marker_hook($level);
@@ -477,10 +476,7 @@ sub _detect_serial_marker_capability ($self) {
     my $level = 1;
     my $pretty = testapi::get_var('PRETTY_SERIAL_MARKER');
     my $serial_term = testapi::is_serial_terminal();
-    if (!$pretty || $serial_term) {
-        $self->{_serial_marker_level}->{$console} = $level;
-        return $level;
-    }
+    return $self->{_serial_marker_level}->{$console} = $level if !$pretty || $serial_term;
 
     testapi::type_string "echo \"BASH:\$BASH_VERSION:\" > /dev/$testapi::serialdev\n";
     my $out = testapi::wait_serial(qr/BASH:([^:]*):/, 10);
@@ -498,8 +494,7 @@ sub _detect_serial_marker_capability ($self) {
         bmwqemu::log_call("serial_marker: console '$console' Level 1 detected (fallback)");
         return 1;
     }
-    $self->{_serial_marker_level}->{$console} = $level;
-    return $level;
+    return $self->{_serial_marker_level}->{$console} = $level;
 }
 
 1;
--- a/distribution.pm
+++ b/distribution.pm
@@ -466,23 +466,26 @@ sub _detect_serial_marker_capability ($self) {
     my $level = 1;
     my $pretty = testapi::get_var('PRETTY_SERIAL_MARKER');
     my $serial_term = testapi::is_serial_terminal();
-    if ($pretty && !$serial_term) {
-        testapi::type_string "echo \"BASH:\$BASH_VERSION:\" > /dev/$testapi::serialdev\n";
-        my $out = testapi::wait_serial(qr/BASH:([^:]*):/, 10);
-        if ($out && $out =~ /BASH:([3-9]|\d{2,})/) {
-            $level = 2;
-            # Check if bash and history features are available to use pretty serial markers
-            testapi::type_string "type fc && set -o | grep -q 'history.*on' && echo \"FC:OK:\" > /dev/$testapi::serialdev\n";
-            if (testapi::wait_serial(qr/FC:OK:/, 10)) {
-                $level = 3;
-            }
-            $self->install_serial_marker_hook($level);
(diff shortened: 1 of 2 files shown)


gauravpathak (approved):

distribution.pm
```diff
@@ -461,28 +466,37 @@ Returns:
...
-            bmwqemu::log_call("serial_marker: console '$console' Level 1 detected (fallback)");
-            return 1;
+    if (!$pretty || $serial_term) {
+        $self->{_serial_marker_level}->{$console} = $level;
+        return $level;
+    }
+
+    testapi::type_string "echo \"BASH:\$BASH_VERSION:\" > /dev/$testapi::serialdev\n";
+    my $out = testapi::wait_serial(qr/BASH:([^:]*):/, 10);
+    if ($out && $out =~ /BASH:([3-9]|\d{2,})/) {
```
perlpunk: I don't see that `$1` is used, so it's good practice to not capture:

```suggestion
    if ($out && $out =~ /BASH:(?:[3-9]|\d{2,})/) {
```

distribution.pm
```diff
@@ -461,28 +466,37 @@ Returns:
 
 sub _detect_serial_marker_capability ($self) {
     my $console = testapi::current_console() // 'sut';
-    return $self->{_serial_marker_level}->{$console} if $self->{_serial_marker_level}->{$console};
+    if ($self->{_serial_marker_level}->{$console}) {
+        my $level = $self->{_serial_marker_level}->{$console};
```
Martchus:
```suggestion
    if (my $level = $self->{_serial_marker_level}->{$console}) {
```

distribution.pm
```diff
@@ -445,6 +445,11 @@ sub install_serial_marker_hook ($self, $level) {
         $pc = "PROMPT_COMMAND='if [ -n \"\$__OA_MARK\" ]; then echo \"\${__OA_MARK}-\$?-\" > $dev; unset __OA_MARK; fi'";
     }
     testapi::type_string "$pc\n";
+    my $marker_match = $level == 3 ? 'OA:DONE' : '__OA_MARK';
+    my $hook_cmd = "for f in ~/.bashrc ~/.profile; do grep -q '$marker_match' \"\$f\" 2>/dev/null || cat <<'EOF' >> \"\$f\"\n$pc\nEOF\ndone\n";
+    testapi::type_string $hook_cmd;
+    my $console = testapi::current_console() // 'sut';
+    $self->{_serial_marker_hook_installed}->{$console} = 1;
```
Martchus, line 452: Where is this reset (except in tests)?

distribution.pm
```diff
@@ -461,28 +466,37 @@ Returns:
...
-            $self->install_serial_marker_hook($level);
-            bmwqemu::log_call("serial_marker: console '$console' Level $level detected");
-        }
-        else {
-            bmwqemu::log_call("serial_marker: console '$console' Level 1 detected (fallback)");
-            return 1;
+    if (!$pretty || $serial_term) {
+        $self->{_serial_marker_level}->{$console} = $level;
+        return $level;
+    }
```
Martchus:
```suggestion
    return $self->{_serial_marker_level}->{$console} = $level if !$pretty || $serial_term;
```

You could probably do the same in other cases, too.

distribution.pm
okurz: nowhere in os-autoinst but can be used in test distribution code in special cases similar to already existing "special cases" like some custom handling of reboots or remote cloud or ssh connections

distribution.pm

distribution.pm

distribution.pm

okurz:
incorporated review suggestions from @perlpunk and @Martchus

distribution.pm
Martchus: But wouldn't that mean modifying a private and undocumented variable from test code?

okurz: yes, it would so I don't recommend it. So what do you propose?

Martchus: If test distribution code needs to unset this to handle "special cases" (even though it is not recommended), the variable shouldn't be private or there should be be some public function to reset it.

If this is just for internal tracking it is probably good as-is. I've just read the code again and I suppose this distinguishes the case when the `_serial_marker_hook_installed` is undef and when it is 1. I still wondering: Wouldn't we need to set this variable back to a falsy value when a reboot is made? Otherwise, how is this working?

okurz: > Wouldn't we need to set this variable back to a falsy value when a reboot is made? Otherwise, how is this working?

No, "installed" here means that we add the functionality into dotfiles so it's persistent within one test over reboots. Only in case of doing full reinstalls or removing the home dir or other elaborate cases we would need to reset. I wouldn't "encourage" using that from test distributions by making a public function, at least not yet when we don't know the use cases when this would be needed exactly. But for the sake of "completeness" we can still do, if you prefer

Martchus (approved):

PR #2857, opened 2026-04-04
feat(llm): integrate LLM analysis into isotovideo

okurz:
Motivation:
We can enable a LLM failure analysis to run automatically at the end of
a test run after all assets are generated and results are finalized.

Design Choices:
- Inject the analysis call inside `handle_shutdown()` right after
  `handle_generated_assets`.
- Use `require` instead of `use` to avoid unnecessary loading when the
  feature is disabled.

Verified manually with an isotovideo run, e.g.

```
isotovideo -d schedule=tests/boot,tests/assert_screen llm_failure_analysis=1
```

using the existing os-autoinst test modules with local adaptions. For
invalid code like an additional test module line
`invalid_function_called();` the LLM investigation using a local
llama.cpp server running the model
"unsloth/gemma-4-E4B-it-GGUF:UD-IQ2_M" the analysis needs 40s and
concludes with

```
The tests failed because the `assert_screen` test encountered an
internal error (`Undefined subroutine`) during execution, indicating a
bug within the test code itself.
To prevent these failures, the test suite needs to be corrected or
updated.
This is likely a test infrastructure problem (false positive) rather
than a product regression.
```

For a non-existant needle tag the analysis revealed

```
1. The tests failed because the test backend was unable to find a
     matching screen state (`on_prompt_does_not_exist`) within the
     allotted timeout, causing the assertion to time out.
2. To prevent these failures, the test environment needs to be robustly
     configured to handle the absence of expected screen states, or the
     test case needs to be modified to handle the timeout scenario
     gracefully.
3. This is likely a test infrastructure problem (false positive) because
     the failure is related to the test runner's inability to confirm the
     expected state, rather than a functional bug in the software being
     tested.
```

Benefits:
- When enabled captures the complete execution context for analysis.

Related issues:
* https://progress.opensuse.org/issues/198056

### DIFF
--- a/OpenQA/Isotovideo/LLMAnalysis.pm
+++ b/OpenQA/Isotovideo/LLMAnalysis.pm
@@ -1,0 +1,130 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Isotovideo::LLMAnalysis;
+use Mojo::Base -signatures;
+use bmwqemu;
+use Feature::Compat::Try;
+use IPC::Run ();
+use Mojo::UserAgent;
+use Mojo::JSON qw(decode_json);
+use Mojo::File qw(path);
+use Text::ParseWords;
+
+use constant MAX_PAYLOAD_SIZE => 8000;
+
+sub _get_file_tail ($filename, $result_dir, $max_lines) {
+    my $file = $filename;
+    $file = "$result_dir/$filename" unless -e $file;
+    return '' unless -e $file;
+    my $tail = qx(tail -n $max_lines \Q$file\E 2>/dev/null) || '';
+    chomp $tail;
+    return $tail;
+}
+
+sub gather_context ($result_dir) {
+    my @failed_tests;
+    for my $res_file (glob "$result_dir/result-*.json") {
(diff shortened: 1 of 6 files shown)


gauravpathak (approved):

Martchus (approved):

OpenQA/Isotovideo/LLMAnalysis.pm
```diff
@@ -0,0 +1,130 @@
...
+    my $arch = $context->{arch};
+    my $failed_tests = $context->{failed_tests};
+    my $log_tail = $context->{log_tail};
+    my $serial_tail = $context->{serial_tail};
+
+    # Truncate inputs to preserve instructions at the end of the prompt
+    $log_tail = substr $log_tail, -MAX_PAYLOAD_SIZE if length($log_tail) > MAX_PAYLOAD_SIZE;
+    $serial_tail = substr $serial_tail, -MAX_PAYLOAD_SIZE if length($serial_tail) > MAX_PAYLOAD_SIZE;
+
+    my $prompt = <<"EOF";
```
perlpunk, line 61: Please use indented here documents.

okurz: https://github.com/os-autoinst/os-autoinst/pull/2862

PR #2863, opened 2026-04-10
fix: prevent deprecation warning about "spurt"

okurz:


kalikiana (approved):

perlpunk (approved):
Where did you see those warnings?

okurz:
> Where did you see those warnings?

Surprisingly only in production in https://openqa.opensuse.org/tests/5840051/logfile?filename=autoinst-log.txt#line-9542 when creating a test with LLM_FAILURE_ANALYSIS=1

PR #2865, opened 2026-04-10
[HELP NEEDED] feat: improve LLM failure analysis visibility

okurz:
Motivation:
LLM-based failure analysis was only saved to a file that was often
inaccessible in openQA. It should be directly visible in the logs
and as a test step.

Design Choices:
- Modified LLMAnalysis.pm to output the analysis via bmwqemu::diag.
- Created a standalone test module result (result-llm_failure_analysis.json)
  to ensure the analysis is uploaded and displayed in the OpenQA UI.
- Updated tests to verify diagnostic output and JSON file creation.
- Replaced wide ellipsis character with three dots to avoid TAP encoding
  issues in some environments.

Benefits:
- Immediate feedback in the test log.
- Native integration with the OpenQA web interface as a result step.

### DIFF
--- a/OpenQA/Isotovideo/LLMAnalysis.pm
+++ b/OpenQA/Isotovideo/LLMAnalysis.pm
@@ -125,6 +125,16 @@ sub run ($result_dir) {
     }
     path("$result_dir/llm-failure-analysis.txt")->spew($output);
     bmwqemu::diag("LLM Analysis:\n$output\nSaved to $result_dir/llm-failure-analysis.txt");
+    my $analysis_result = {
+        name => 'llm_failure_analysis',
+        result => 'ok',
+        details => [{
+                title => 'LLM Failure Analysis',
+                result => 'none',
+                text => 'llm-failure-analysis.txt',
+        }],
+    };
+    bmwqemu::save_json_file($analysis_result, "$result_dir/result-llm_failure_analysis.json");
 }
 
 1;
--- a/script/isotovideo
+++ b/script/isotovideo
@@ -155,9 +155,7 @@ sub handle_shutdown () {
         $runner->stop_autotest();    # uncoverable statement
     }
     my $clean_shutdown = $runner->handle_shutdown(\$RETURN_CODE);
-    bmwqemu::load_vars();    # read calculated variables from backend and tests
     $RETURN_CODE = handle_generated_assets($runner->command_handler, $clean_shutdown) unless $RETURN_CODE;
-    OpenQA::Isotovideo::LLMAnalysis::run(bmwqemu::result_dir()) if $bmwqemu::vars{LLM_FAILURE_ANALYSIS};
     diag 'isotovideo completed handle_shutdown: ' . ($RETURN_CODE ? 'failed' : 'done');
 }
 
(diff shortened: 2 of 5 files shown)


Martchus (approved):

OpenQA/Isotovideo/LLMAnalysis.pm
```diff
@@ -124,7 +124,17 @@ sub run ($result_dir) {
         $output = query_llm_api($prompt, $url, $model);
     }
     path("$result_dir/llm-failure-analysis.txt")->spew($output);
-    bmwqemu::diag("LLM Analysis complete. Saved to $result_dir/llm-failure-analysis.txt");
+    bmwqemu::diag("LLM Analysis:\n$output\nSaved to $result_dir/llm-failure-analysis.txt");
```
okurz, line 127: This kinda works, see https://openqa.opensuse.org/tests/5841471/logfile?filename=autoinst-log.txt#line-1937 which shows

```
[2026-04-10T19:26:33.228631Z] [debug] [pid:2140872] LLM Analysis:
  The tests failed because the system experienced kernel instability, evidenced by multiple "BUG: unable to handle page fault" errors, which led to a timeout during the test's state monitoring (`testapi::wait_serial`). To prevent these failures, the underlying kernel bug must be resolved, as the system instability is the root cause. This is likely a product regression, as the test failure is directly tied to kernel crashes, rather than a flaw in the test infrastructure itself.
  Saved to testresults/llm-failure-analysis.txt
```

but the text is kinda hidden in autoinst-log.txt. For a separate text file upload we would need to change openQA to explicitly look at a file, right?

However, with the following block I was trying to make the text show up in a test step. How can I make that show up in a text box on the job details? ariel:/var/lib/openqa/testresults/05841/05841471-opensuse-Tumbleweed-agama-installer-x86_64-Build20260409-okurz-llm-extra_tests_gnome-okurz-llm@uefi-3G does not show any "result-llm_failure_analysis.json".

Martchus: > For a separate text file upload we would need to change openQA to explicitly look at a file, right?

No - I think the openQA worker will upload all files under `ulogs`. So just put it there.

> How can I make that show up in a text box on the job details?

I would also have to check out the os-autoinst code. The file probably needs to be referenced somewhere. The JSON files *probably* correspond to test modules which *probably* need to be part of the test order.

d3flex (commented):
Sorry because you are excited about that and I am kind of late to say that this is not so good idea IMO. it is waste of resources running for each failing which usually when it occurs we see the failures in other test too. and this will going to run for each of them. Which because there is not proper information the respond will take time and the result is poor - just to include an suggestion or indication.

okurz:
> Sorry because you are excited about that and I am kind of late to say that this is not so good idea IMO. it is waste of resources running for each failing which usually when it occurs we see the failures in other test too. and this will going to run for each of them. Which because there is not proper information the respond will take time and the result is poor - just to include an suggestion or indication.

Running or not running an LLM investigation has nothing to do with this PR changing how the output is handled. Please move that discussion elsewhere. I recommend you read https://progress.opensuse.org/issues/199907

PR #2867, opened 2026-04-13
feat: output LLM analysis with diag

okurz:
Related to
* #2865


Martchus:
The commit  message is not very meaningful. I suppose this should be something like `feat: Avoid printing full LLM analysis with diag`.

perlpunk (approved):

okurz:
> The commit  message is not very meaningful. I suppose this should be something like `feat: Avoid printing full LLM analysis with diag`.

No, it actually *should* print the full analysis, see the `$output` in

```
    bmwqemu::diag("LLM Analysis:\n$output\nSaved to $result_dir/llm-failure-analysis.txt");
```

The whole analysis is 2-3 sentences

Martchus (approved):

PR #2869, opened 2026-04-13
fix(manpages): Fix the output of pod2man

rclobus:
The output of pod2man can be used by lexgrog, which is used by e.g. apropos.

And this will remove 2 lintian warning for Debian :-)

### DIFF
--- a/script/isotovideo
+++ b/script/isotovideo
@@ -4,9 +4,13 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
 
+=head1 NAME
+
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

okurz (approved):

perlpunk:
WIP still seems to be broken.
There is an issue, but it's closed for some reason: https://github.com/wip/app/issues/951

edit: ah, github took a long time to load the last comment. Apparently this time it's github itself

perlpunk:
All mergify rules are met, merging manually

PR #2870, opened 2026-04-13
fix(container): add missing diffutils package to dev Dockerfile

d3flex:
install-new-deps.sh was failing with "diff: command not found" at line 28 because the diffutils package was not installed in the container image.

### DIFF
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -29,6 +29,7 @@ RUN zypper in -y -C \
        aspell-spell \
        cmake \
        cpio \
+       diffutils \
        ffmpeg \
        file \
        gcc-c++ \
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -109,6 +109,7 @@ docker_requires:
   shadow:
   sudo:
   which:
+  diffutils:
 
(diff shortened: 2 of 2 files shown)


d3flex:
https://github.com/os-autoinst/os-autoinst/actions/runs/24330868302/job/71036211870?pr=2867#step:3:251

Martchus (changes requested):
Looks like you have to do this change via `dependencies.yaml`:

```
#   Failed test 'No changed files after 'make update-deps''
#   at xt/27-make-update-deps.t line 25.
#  M container/os-autoinst_dev/Dockerfile
# Looks like you failed 1 test of 2.
```

Martchus (approved):

perlpunk (approved):
I think we should also fix the script to actually fail if the `diff` call fails. I guess you spotted this only by chance? The workflow where this happened suceeded.
Should we create a ticket?

d3flex:
> I think we should also fix the script to actually fail if the `diff` call fails. I guess you spotted this only by chance? The workflow where this happened suceeded.

yes, it was by chance. let me try to address this in the same ticket/separate commit/PR today. Otherwise I will file a ticket tomorrow

PR #2872, opened 2026-04-13
feat: abort test with incomplete if requested HDD size > threshold

okurz:
Motivation:
As observed in #199058, openQA jobs can completely exhaust available storage
on workers. We should check if the requested HDDSIZEGB values fit in the
available space minus a buffer (e.g., 60% threshold) when starting jobs in
os-autoinst and abort early with "incomplete" status if the check fails to
prevent worker node failure.

Design Choices:
- Added a disk space check in `bmwqemu::ensure_valid_vars` that calculates the
  total requested HDD size across all disks.
- Requested size is summed from `HDDSIZEGB` and `HDDSIZEGB_$i` for all disks
  (defaulting to 10 GiB if not specified).
- Extracted `MIB` and `GIB` constants in `bmwqemu.pm` for clearer disk size
  calculations.
- Used `df -PB1 .` to retrieve available disk space in bytes for the current
  directory where disks will be created.
- Implemented a configurable threshold (default 80%): if the total requested
  disk space (in bytes) exceeds the threshold of the available space, the job
  is aborted.
- On failure, `bmwqemu::serialize_state` is called with `result => 'incomplete'`
  and a descriptive message, followed by `die`.

Benefits:
Prevents worker storage exhaustion by proactively aborting jobs that request
excessive disk space. This improves worker stability, maintainability, and
provides clear feedback to users when a job cannot be accommodated.

Related issue: https://progress.opensuse.org/issues/199061

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -73,7 +73,12 @@ sub result_dir () { 'testresults' }
     *init_logger = \&log::init_logger;
 }
 
-use constant STATE_FILE => 'base_state.json';
+use constant {
+    STATE_FILE => 'base_state.json',
+    MIB => 1024 * 1024,
+    GIB => 1024 * 1024 * 1024,
+    STORAGE_KEEP_FREE_RATIO => $ENV{OS_AUTOINST_STORAGE_KEEP_FREE_RATIO} // .2,
+};
 
 # Write a JSON representation of the process termination to disk
 sub serialize_state (%state) {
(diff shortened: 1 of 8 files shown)


bmwqemu.pm
```diff
@@ -150,6 +154,12 @@ sub _check_publish_vars () {
     return 1;
 }
 
+sub _get_available_bytes ($path) {
+    my $df_output = `df -PB1 "$path"`;
```
perlpunk:
```suggestion
    my $df_output = qx{df -PB1 "$path"};
```

perlpunk (approved):

bmwqemu.pm
```diff
@@ -150,6 +154,12 @@ sub _check_publish_vars () {
     return 1;
 }
 
+sub _get_available_bytes ($path) {
+    my $df_output = qx{df -PB1 "$path"};
```
Martchus: I think this needs error handling (despite autodie). Otherwise `$df_output` will be `undef` in the error case and we get a warning about trying to match `undef`. Considering Perl will print `Can't exec …` anyway we can maybe just return `undef` in any error case here (when `$df_output` is `undef`) which the calling function already seems to handle.

bmwqemu.pm
```diff
@@ -160,6 +170,27 @@ sub ensure_valid_vars () {
     die 'CASEDIR variable not set, unknown test case directory' if !defined $vars{CASEDIR};
     die "No scripts in CASEDIR '$vars{CASEDIR}'\n" unless -e $vars{CASEDIR};
     die "WHEELS_DIR '$vars{WHEELS_DIR}' does not exist" if defined $vars{WHEELS_DIR} && !-d $vars{WHEELS_DIR};
+
+    # Abort if total HDD size exceeds threshold of available space
+    my $threshold = $vars{DISK_SPACE_THRESHOLD} // 0.8;
```
Martchus: In my recent PRs you were suggesting I should use "storage" instead of "disk". If you care about this than you should be consistent :-)

bmwqemu.pm
```diff
@@ -160,6 +170,27 @@ sub ensure_valid_vars () {
...
+    my $total_hdd_size_gb = 0;
+    for my $i (1 .. $numdisks) {
+        my $size = $vars{"HDDSIZEGB_$i"} // $vars{HDDSIZEGB} // 10;
+        $total_hdd_size_gb += $size;
+    }
+
+    if ($total_hdd_size_gb > 0) {
+        my $available = _get_available_bytes('.');
+        my $requested_bytes = $total_hdd_size_gb * GIB;
+        if (defined $available && $requested_bytes > $available * $threshold) {
```
Martchus: Maybe it would make sense to print a warning if `!defined $available`. (Just a suggestion.)

Martchus (changes requested):

bmwqemu.pm
perlpunk: Maybe "disk" makes more sense here because it's simply about "generic" disk space and not about a certain storage location for results or archive?

okurz: No, I agree with @Martchus , "storage" is a better generic term. "disc" is the physical form used in partially obsolete magnetic storage on mechanical spinning disks :)

bmwqemu.pm

bmwqemu.pm

bmwqemu.pm

bmwqemu.pm
```diff
@@ -150,16 +155,52 @@ sub _check_publish_vars () {
...
+    }
+    return undef unless $total_hdd_size_gb > 0;
+    my ($total_storage, $available) = _get_storage_stats('.');
+    if (!defined $total_storage) {
+        warn "Could not determine available storage space\n";
+        return undef;
+    }
+    my $requested_bytes = $total_hdd_size_gb * GIB;
+    my $min_free_bytes = $total_storage * $keep_free;
+    return undef unless $requested_bytes > $available - $min_free_bytes;
```
okurz, line 184: @perlpunk algorithm fixed

perlpunk (approved):

Martchus (approved):

PR #2864, opened 2026-04-10
test: Check for unhandled output by default

Martchus:
Related ticket: https://progress.opensuse.org/issues/182111

PR #2875, opened 2026-04-15
chore(mergify): Update required number of successful checks

Martchus:
Judging by
https://github.com/os-autoinst/os-autoinst/pull/2864#issuecomment-4251046241 we should have 29 successful job now.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,7 +14,7 @@ pull_request_rules:
           # "unresolvable" is not reported in general:
           # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
           # So we need to require the number of tests explicitly:
-          - "#check-success>=15"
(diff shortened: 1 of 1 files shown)


okurz:
We don't need to specify the exact number, just ensure that all groups of CI checks are triggered. As soon as pending checks show up the `#check-pending=0` rules catches those. I suggest to not bump the number *that* high so that for example we don't have a problem again when Leap 15.6 is EOF and according OBS checks are disabled

perlpunk:
> As soon as pending checks show up the `#check-pending=0` rules catches those.

Just that OBS checks are never reported as pending

perlpunk (approved):

d3flex (approved):

kalikiana:
@mergifyio rebase

okurz:
I would have preferred a lower number to not have confusing situations again when we remove 15.6 but fine

PR #2877, opened 2026-04-16
ci: Ensure diff is available in install-new-deps.sh

d3flex:
Addressing /pull/2870#pullrequestreview-4120107401

Link: https://github.com/os-autoinst/os-autoinst/pull/2870

### DIFF
--- a/tools/install-new-deps.sh
+++ b/tools/install-new-deps.sh
@@ -25,6 +25,10 @@ $sudo zypper --no-refresh install -y -C "${DEPS[@]}"
 listdeps > $NEWDEPS
 
 echo "Checking updated packages"
+command -v diff &> /dev/null || {
+    echo "ERROR: diff is not installed"
+    exit 1
+}
 if diff $OLDDEPS $NEWDEPS > $DIFFDEPS; then
     echo "NO DIFF"
 else


perlpunk (commented):
looks good, just reformat and then rebase on #2870 when merged

tools/install-new-deps.sh
```diff
@@ -25,6 +25,10 @@ $sudo zypper --no-refresh install -y -C "${DEPS[@]}"
 listdeps > $NEWDEPS
 
 echo "Checking updated packages"
+command -v diff &>/dev/null || {
+    echo "ERROR: diff is not installed"
+    exit 1
+}
 if diff $OLDDEPS $NEWDEPS > $DIFFDEPS; then
```
okurz, line 32: I don't see what's the benefit here. Wouldn't this fail in the next line with "command not found" anyway?

okurz (changes requested):

tools/install-new-deps.sh
perlpunk: No, it wouldn't, and it didn't in the past, that's why we never had a failing workflow.
it's because of the `if diff ...`

okurz (approved):

perlpunk (approved):

PR #2878, opened 2026-04-16
feat: include executed command in step details and serial markers

okurz:
Motivation:
Command outputs in the serial log (autoinst-log.txt) and step details are
currently difficult to link back to the originating test API command.
Including the command string in both places improves visibility and
traceability for both human reviewers and automated analysis tools.

Design Choices:
1. Modified distribution.pm to inject an OA:START marker before command
   execution and include the command in the OA:DONE marker using bash's
   fc -ln -1.
2. Extended wait_serial and record_serialresult to propagate the
   executed command.
3. Injected the command string into the generated .txt detail files in
   basetest.pm, ensuring visibility in the openQA Web UI without
   requiring frontend changes.

Example content of test result step text output:

```
Command: ip a del 10.0.2.2/15 dev br1
wait_serial expected: qr/OA:DONE-(\d+)-/u
Result:
OA:DONE-0-
```

Benefits:
- Clear command boundaries in the serial log.
- Executed commands are immediately visible in the openQA step details.
- No changes required to the openQA web frontend.

Related issue: https://progress.opensuse.org/issues/198872

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -429,12 +429,14 @@ sub record_resultfile ($self, $title, $output, %nargs) {
     $self->write_resultfile($filename, $output);
 }
 
-sub record_serialresult ($self, $ref, $res, $string = undef) {
+sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $string //= '';
     # take screenshot for documentation (screenshot does not represent fail itself)
     $self->take_screenshot() unless (testapi::is_serial_terminal);
 
-    my $output = "# wait_serial expected: $ref\n";
+    my $output = '';
+    $output .= "# Command: $args{command}\n" if defined $args{command};
+    $output .= "# wait_serial expected: $ref\n";
     $output .= "# Result:\n";
     $output .= "$string\n";
     $self->record_resultfile('wait_serial', $output, result => $res);
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -1334,6 +1334,16 @@ subtest 'retrieve and change various global variables' => sub {
     }
 };
 
+subtest 'wait_serial record_command' => sub {
+    my $mock_test = Test::MockModule->new('basetest');
+    my $recorded_cmd;
+    $mock_test->redefine(record_serialresult => sub ($self, $ref, $res, $string, %args) {
+            $recorded_cmd = $args{command};
+    });
+    wait_serial('regex', record_command => 'my_command');
+    is($recorded_cmd, 'my_command', 'command passed to record_serialresult');
+};
+
 done_testing;
 
 END {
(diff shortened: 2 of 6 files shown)


kalikiana (approved):

t/03-testapi.t
```diff
@@ -1334,6 +1334,17 @@ subtest 'retrieve and change various global variables' => sub {
     }
 };
 
+subtest 'wait_serial record_command' => sub {
+    my $mock_test = Test::MockModule->new('basetest');
+    my $recorded_cmd;
+    $mock_test->redefine(record_serialresult => sub {
+            my ($self, $ref, $res, $string, %args) = @_;
```
Martchus: Use signatures in new code.

Martchus (commented):
The ticket was probably about doing this in a more structured way but I guess this change is already an improvement.

t/03-testapi.t
okurz: done

okurz:
Incorporated suggestion from @Martchus to use signatures in new code. See https://github.com/os-autoinst/os-autoinst/pull/2885 to consistently enforce signatures in this project also in this in-place sub-definitions

t/17-basetest.t
```diff
@@ -518,4 +518,18 @@ subtest search_for_expected_serial_failures => sub {
     is($basetest->{result}, 'successfully called function', 'search for expected serial failures is working');
 };
 
+subtest record_serialresult_with_command => sub {
+    my $basetest = basetest->new();
+    $basetest->{name} = 'test';
+    my $recorded_output;
+    my $mock_basetest_local = Test::MockModule->new('basetest');
+    $mock_basetest_local->redefine(record_resultfile => sub {
+            my ($self, $title, $output, %nargs) = @_;
```
d3flex: must be part of the sub as well

okurz: done

basetest.pm
```diff
@@ -429,12 +429,14 @@ sub record_resultfile ($self, $title, $output, %nargs) {
...
-sub record_serialresult ($self, $ref, $res, $string = undef) {
+sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $string //= '';
     # take screenshot for documentation (screenshot does not represent fail itself)
     $self->take_screenshot() unless (testapi::is_serial_terminal);
 
-    my $output = "# wait_serial expected: $ref\n";
+    my $output = '';
+    $output .= "# Command: $args{command}\n" if defined $args{command};
+    $output .= "# wait_serial expected: $ref\n";
```
d3flex, line 439: Can we name the `$ref` -> `$matcher`, `$serial_match` or anything more declarative? It will make more easy to understand what the signature expects. other than that looks good

Martchus (approved):

basetest.pm
okurz: Do you just mean to rename the variable within code or chaneg something in the rendered output?

PR #2880, opened 2026-04-16
fix: allow disabling disk space check by setting ratio to 0

okurz:
Motivation:
The recently introduced disk space check (commit bb08a482) aborts jobs if the
requested HDD size exceeds the available space, even if the keep-free ratio is
set to 0. This prevents jobs using sparse disk images from running on workers
where the total requested size exceeds physical capacity, which is a valid use
case.

Design Choices:
- Modified bmwqemu::_abort_if_storage_limit_exceeded to return early if
  STORAGE_KEEP_FREE_RATIO is set to 0 or less.
- Updated documentation to reflect that 0 disables the check.
- Added a test case to t/12-bmwqemu.t to verify that setting the ratio to 0
  allows jobs to proceed regardless of available space.

Benefits:
- Restores the ability to run jobs with large (sparse) disk images when
  requested by the user.
- Provides a clear way to disable the safety check if needed.

Related issue: https://progress.opensuse.org/issues/199061

PR #2881, opened 2026-04-17
feat(snd2png debugviewer): Move to /usr/lib

rclobus:
Move this helper program out of the public namespace for executables


rclobus:
I have similar code available to debugviewer as well. I'll post that when this PR progresses (and if you are interested).

I'm currently updating the packaging of os-autoinst for Debian with @phil-hands and noticed that Debian would like to have a man page for this tools (which has no real use outside of the os-autoinst toolchain). So instead of writing a man page I shifted the binary out of `/usr/bin` :-)

When this is in, it will also need an update of the AppArmor rules of openQA (from `/usr/bin/snd2png` to `/usr/lib/os-autoinst/snd2png`)

rclobus:
The openQA Fullstack CI test fails to a full HDD?
```
# Reason: died: Not enough storage for requested HDDSIZEGB (requested 10 GiB, available 14 GiB, total 71 GiB, keep-free 20%)
```

okurz:
> The openQA Fullstack CI test fails to a full HDD?
> 
> ```
> # Reason: died: Not enough storage for requested HDDSIZEGB (requested 10 GiB, available 14 GiB, total 71 GiB, keep-free 20%)
> ```

this might need #2880

okurz (commented):
What's the motivation?

dist/rpm/os-autoinst.spec
```diff
@@ -326,6 +326,7 @@ fi
 %defattr(-,root,root)
 %{_docdir}/os-autoinst
 %dir %{_prefix}/lib/os-autoinst
+%{_prefix}/lib/os-autoinst/snd2png
```
Martchus, line 330: I think this should be `/usr/libexec` for Tumbleweed at this point. It makes probably most sense to use `%_libexecdir` instead of hardcoding that path.

(also see https://en.opensuse.org/openSUSE:Specfile_guidelines#Libexecdir)

snd2png/CMakeLists.txt
```diff
@@ -6,4 +6,4 @@ target_link_libraries(${PROJECT_NAME} PRIVATE opencv_core opencv_imgcodecs)
 target_compile_options(${PROJECT_NAME} PRIVATE ${PRIVATE_COMPILE_OPTIONS})
 target_use_pkg_config_module(${PROJECT_NAME} "fftw3")
 target_use_pkg_config_module(${PROJECT_NAME} "sndfile")
-install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
+install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION "${OS_AUTOINST_DATA_DIR}")
```
Martchus, line 9: This should probably use `LIBEXECDIR` from [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) which we might already use elsewhere.

We probably need to find some way to expose `LIBEXECDIR` to Perl so the Perl code knows where to find its executables.

rclobus:
> What's the motivation?

See https://github.com/os-autoinst/os-autoinst/pull/2881#issuecomment-4268167425:
The Debian package has a lintian (one of our QA tools) warning that the man page is missing.

I could have written a man page for this tools, but actually `snd2png` is not intended for general use, and does not need to be available in `/usr/bin`. I'm of this opinion, because actually only 10240ms gets converted to the png file (excluding leading silence) and not the whole file (which I would have expected of a general-purpose tool).

There is already on compiler binary program in /usr/lib/os-autoinst: videoencoder, so I placed snd2png there as well.

Martchus (commented):
There's room for improvement, check out the inline comments.

However, we already do the same for `videoencoder` so I guess hardcoding this to use the data directory and `%{_prefix}/lib/os-autoinst` is not making things worse.

dist/rpm/os-autoinst.spec
rclobus: I copied-and-pasted from the `videoencoder` example, so if you want to use `/usr/libexec`, I would suggest to start with `videoencoder` first... (or if you are happy with this PR, I'll send a PR for debugviewer too, and then all three can be migrated in one PR afterwards)

Martchus: I don't mind either way. As mentioned, your PR doesn't make things worse considering we already do this for `videoencoder` so we can merge it as-is (and improve the whole situation later). If you want to improve `videoencoder` and then install the other executables in the same way I'd appreciate it, though.

rclobus: I've added a second commit to move `debugviewer` as well (that's easier than you having to resolve the merge conflict, see !2883

okurz (approved):

Martchus (approved):

DimStar77:
Might be responsible for all the failures on snapshot 0420

```
Test died: "/usr/lib/os-autoinst/snd2png testresults/firefox_audio-001-captured.wav firefox_audio-001-captured.wav.png" failed to start: "Permission denied" at /usr/lib/perl5/vendor_perl/5.42.0/IPC/System/Simple.pm line 545, <$isotovideo> line 1.
```

E.g https://openqa.opensuse.org/tests/5863501#step/firefox_audio/9

okurz:
> Might be responsible for all the failures on snapshot 0420
> 
> ```
> Test died: "/usr/lib/os-autoinst/snd2png testresults/firefox_audio-001-captured.wav firefox_audio-001-captured.wav.png" failed to start: "Permission denied" at /usr/lib/perl5/vendor_perl/5.42.0/IPC/System/Simple.pm line 545, <$isotovideo> line 1.
> ```
> 
> E.g https://openqa.opensuse.org/tests/5863501#step/firefox_audio/9

Needs apparmor adaption

rclobus:
Wouldn't it be nicer to adjust the AppArmor rules in the os-autoinst package instead of the openqa package?

okurz:
> Wouldn't it be nicer to adjust the AppArmor rules in the os-autoinst package instead of the openqa package?

yes, it would :) Wanna try to submit adjusted apparmor to os-autoinst? But also we need to keep in mind that the default solution for openSUSE/SLE is SELinux now so apparmor will likely be faded out in favor of SELinux.

rclobus:
> Wanna try to submit adjusted apparmor to os-autoinst?

If you as upstream are migrating from AppArmor to SELinux, as downstream we should too.
Debian can switch to SELinux too and stop providing the AppArmor rules. Where are the SELinux rules defined? I could not find them in the os-autoinst code.

okurz:
> > Wanna try to submit adjusted apparmor to os-autoinst?
> 
> If you as upstream are migrating from AppArmor to SELinux, as downstream we should too. Debian can switch to SELinux too and stop providing the AppArmor rules. Where are the SELinux rules defined? I could not find them in the os-autoinst code.

There are none, yet

PR #2882, opened 2026-04-17
docs:  explain assert_screen semantics with multiple tags

kalikiana:
Otherwise what happens with consecutive needles is ambiguous.

### DIFF
--- a/testapi.pm
+++ b/testapi.pm
@@ -397,6 +397,9 @@ very suitable for checking performance expectations. Under the normal
 circumstance of the screen being shown this does not imply a longer waiting
 time as the method returns as soon as a successful needle match occurred.
 
+Note when using multiple tags the method returns on the first match and an
+additional C<assert_screen> may be needed for consecutive screens.
+
 Specify C<$no_wait> to run the screen check as fast as possible that is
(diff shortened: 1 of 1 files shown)


testapi.pm
```diff
@@ -397,6 +397,9 @@ very suitable for checking performance expectations. Under the normal
 circumstance of the screen being shown this does not imply a longer waiting
 time as the method returns as soon as a successful needle match occurred.
 
+Note when using multiple tags only the first match is considered and an
+additional C<assert_screen> may be needed for consecutive screens.
```
okurz, line 401: I guess we should clarify what "is considered" refers to.

okurz (changes requested):

testapi.pm
kalikiana: Maybe

```suggestion
Note when using multiple tags the method returns on the first match and an
additional C<assert_screen> may be needed for consecutive screens.
```

Martchus (approved):

okurz (approved):

perlpunk (approved):

perlpunk:
@mergifyio update

perlpunk:
@mergifyio rebase

PR #2883, opened 2026-04-17
feat(debugviewer): Move to /usr/lib

rclobus:
Move this helper program out of the public namespace for executables


rclobus:
Similar to [!2881](https://github.com/os-autoinst/os-autoinst/pull/2881), this is debugviewer being moved to /usr/lib.
 
Aborting, as it will be much easier to provide 2 commits, otherwise this will cause a (trivially to solve) merge conflict for `dist/rpm/os-autoinst.spec`

PR #2887, opened 2026-04-21
feat: add absolute storage threshold to prevent over-conservative aborts

okurz:
Motivation:
Jobs requesting a small amount of storage (e.g., 20 GiB) were being aborted on
workers with large disks (e.g., 938 GiB) because they failed the 20% relative
free space check, even though 189 GiB was available. This was overly
conservative for workers with large storage.

Design Choices:
1. Introduced STORAGE_KEEP_FREE_GB (defaulting to 50 GiB) as an absolute minimum
   free space threshold.
2. Updated the storage check logic to only abort if BOTH the relative
   (STORAGE_KEEP_FREE_RATIO) and absolute (STORAGE_KEEP_FREE_GB) thresholds
   would be violated by the job.

Benefits:
- Small jobs can run on workers with large disks as long as enough absolute
  space remains.
- Maintains safety for workers with limited storage.
- Improved test maintainability and readability.

Related issue: https://progress.opensuse.org/issues/200006

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -78,6 +78,7 @@ use constant {
     MIB => 1024 * 1024,
     GIB => 1024 * 1024 * 1024,
     STORAGE_KEEP_FREE_RATIO => $ENV{OS_AUTOINST_STORAGE_KEEP_FREE_RATIO} // .2,
+    STORAGE_KEEP_FREE_GB => $ENV{OS_AUTOINST_STORAGE_KEEP_FREE_GB} // 50,
 };
 
 # Write a JSON representation of the process termination to disk
@@ -182,7 +183,10 @@ sub _abort_if_storage_limit_exceeded () {
     return warn "Could not determine available storage space\n" unless defined $total_storage;
     my $requested_bytes = $total_hdd_size_gb * GIB;
     my $min_free_bytes = $total_storage * $keep_free;
-    return undef if $requested_bytes <= $available - $min_free_bytes;
+    my $keep_free_gb = $vars{STORAGE_KEEP_FREE_GB} // STORAGE_KEEP_FREE_GB;
+    my $relative_exceeded = $requested_bytes > $available - $min_free_bytes;
+    my $absolute_exceeded = $keep_free_gb > 0 && $requested_bytes > $available - ($keep_free_gb * GIB);
+    return undef unless $relative_exceeded && $absolute_exceeded;
     my $msg = sprintf 'Not enough storage for requested HDDSIZEGB (requested %d GiB, available %d GiB, total %d GiB, keep-free %d%%)',
       $total_hdd_size_gb, int($available / GIB), int($total_storage / GIB), int($keep_free * 100);
     serialize_state(result => 'incomplete', msg => $msg);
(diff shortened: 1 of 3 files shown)


doc/backend_vars.md
```diff
@@ -62,6 +62,7 @@ Supported variables per backend
 | _HIDE_SECRETS_REGEX | string |  | If set, any test variables whose **NAME** (key) matches the specified regular expression, in addition to the default '^_SECRET_' and '_PASSWORD', are excluded from being saved into vars.json or further processing. For example, to hide all variables starting with 'SCC_REGCODE', use '^SCC_REGCODE'. |
 | STORAGE_KEEP_FREE_RATIO | float | 0.2 | Ratio of total storage space to keep free (e.g. 0.2 for 20%). If the total requested HDD size exceeds the available space while keeping this ratio of total storage size free, the job is aborted early with "incomplete" status. Set to 0 to disable this check. |
 | FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED | boolean | 0 | Fail explicitly if a test module is scheduled with the `always_rollback` flag but snapshots are not supported by the backend. |
+| STORAGE_KEEP_FREE_ABSOLUTE | integer | 50 | Absolute free space threshold in GiB. If the total requested HDD size is below this value, the job can run even on workers with limited storage (only the relative check applies). Set to 0 to disable the absolute threshold. |
```
Martchus: I'd add it directly after `STORAGE_KEEP_FREE_RATIO` so these related settings are together.

And shouldn't it be "*even though* the relative check applies"? And "*even though* `STORAGE_KEEP_FREE_RATIO` is exceeded" would maybe be even clearer.

bmwqemu.pm
```diff
@@ -182,7 +183,9 @@ sub _abort_if_storage_limit_exceeded () {
     return warn "Could not determine available storage space\n" unless defined $total_storage;
     my $requested_bytes = $total_hdd_size_gb * GIB;
     my $min_free_bytes = $total_storage * $keep_free;
-    return undef unless $requested_bytes > $available - $min_free_bytes;
+    my $relative_exceeded = $requested_bytes > $available - $min_free_bytes;
+    my $absolute_exceeded = $total_hdd_size_gb > STORAGE_KEEP_FREE_ABSOLUTE;
```
Martchus: The documentation says:

> Set to 0 to disable the absolute threshold.

So shouldn't 0 be checked?

```suggestion
    my $absolute_exceeded = STORAGE_KEEP_FREE_ABSOLUTE > 0 && $total_hdd_size_gb > STORAGE_KEEP_FREE_ABSOLUTE;
```

Martchus: The documentation makes it look like it is a test variable but `STORAGE_KEEP_FREE_ABSOLUTE` is just initialized from env.

Martchus (changes requested):

bmwqemu.pm
okurz: fixed

okurz: fixed

doc/backend_vars.md
okurz: fixed. And also I renamed the variable to `STORAGE_KEEP_FREE_GB`. I also updated the comment to explain that this value is for space kept free.

okurz:
Included suggestions from @Martchus, renamed the variable with a `_GB` suffix, fixed a logic bug and extended and reworked tests to cover the relevant scenarios combining both the relative and the absolute threshold.

Martchus (approved):

bmwqemu.pm
```diff
@@ -182,7 +183,10 @@ sub _abort_if_storage_limit_exceeded () {
     return warn "Could not determine available storage space\n" unless defined $total_storage;
     my $requested_bytes = $total_hdd_size_gb * GIB;
     my $min_free_bytes = $total_storage * $keep_free;
-    return undef if $requested_bytes <= $available - $min_free_bytes;
+    my $keep_free_gb = $vars{STORAGE_KEEP_FREE_GB} // STORAGE_KEEP_FREE_GB;
+    my $relative_exceeded = $requested_bytes > $available - $min_free_bytes;
+    my $absolute_exceeded = $keep_free_gb > 0 && $requested_bytes > $available - ($keep_free_gb * GIB);
+    return undef unless $relative_exceeded && $absolute_exceeded;
```
d3flex, line 189: if the capacity is still smaller than the HDDSIZEGB the jobs should be picked up despite the STORAGE_KEEP_FREE_RATIO. IMO the logical and here stiil causes the so-called conservative aborts. but i am not sure I see a strong reason to not be more flexible.

okurz: The intention here is that os-autoinst must be prevented from filling up the complete storage, as we need to leave a buffer for anything else running on the worker machine.

Right now, the default absolute limit is 50 GiB. If the available space drops below 50 GiB, no test is executed, even if the requested storage (HDDSIZEGB) isn't that large. We need this conservative absolute threshold also because os-autoinst saves other artifacts during a test run (like logs, videos, asset downloads, memory dumps, etc.) which are not accounted for in the QEMU disk size.

By enforcing this minimum free space regardless of the requested HDD size, we ensure the worker machine doesn't accidentally run out of storage and impact system stability or other processes. Therefore, we should prefer to keep these "conservative aborts" rather than being more flexible at the risk of completely filling the disk.

d3flex (approved):

PR #2891, opened 2026-04-22
feat(qemu): improve port conflict detection and VNC error diagnostics                                                                

d3flex:
To handle better the situations where the some network ports are already in use                                                      
and provide clearier feedback on conflict, the commit introduces two-fold checks                                                     
in start_qemu(). For the input from user (hostfwd) is checked before QEMU starts                                                     
and abort immediately. Where for VNC the check takes place after QEMU, in order                                                      
to get the error which it returns. So instead of errors like                                                                         
"QEMU terminated before QMP connection could be established" or                                                                      
"unexpected end of data", users now get the ss output:                                                                               
Port 39100 (hostfwd port) is already in use:                                                                                         
LISTEN 0 128 0.0.0.0:39100 0.0.0.0:* users:(("nc",pid=12345,fd=3))                                                                   
                                                                                                                                     
The port check uses connect instead of bind to work around localhost resolving                                                       
differently for IPv4/IPv6.                                                                                                           
Finally, at least for hostfwd the reason in openQA reflects the cause in an                                                          
incomplete job result. What is not address is the assignment to an open port for                                                     
any of the cases above. That has to be a different story as there should be                                                          
alignments between the test distributions and openQA side.

Issue: https://progress.opensuse.org/issues/198428

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -11,6 +11,7 @@ use File::Path 'mkpath';
 use File::Which;
 use Time::HiRes qw(sleep gettimeofday);
 use Time::Seconds;
+use IO::Socket::UNIX 'SOCK_STREAM';
 use IO::Handle;
 use POSIX qw(strftime :sys_wait_h mkfifo);
 use Mojo::File 'path';
@@ -922,6 +923,7 @@ sub start_qemu ($self) {
 
         for (my $i = 0; $i < $num_networks; $i++) {
             if ($vars->{NICTYPE} eq 'user') {
+                _assert_port_availability($_, 'hostfwd') for _extract_hostfwd_ports($vars->{NICTYPE_USER_OPTIONS} // '');
                 my $nictype_user_options = $vars->{NICTYPE_USER_OPTIONS} ? ',' . $vars->{NICTYPE_USER_OPTIONS} : '';
                 $nictype_user_options .= ",smb=${\(dirname($basedir))}" if ($vars->{QEMU_ENABLE_SMBD});
                 sp('netdev', [qv "user id=qanet$i$nictype_user_options"]);
@@ -1058,20 +1060,23 @@ sub start_qemu ($self) {
     $self->{qmpsocket} = $self->{proc}->connect_qmp();
     my $init = myjsonrpc::read_json($self->{qmpsocket});
     my $hash = $self->handle_qmp_command({execute => 'qmp_capabilities'});
-
+    my $vncport = 5900 + $bmwqemu::vars{VNC};
     my $vnc = $testapi::distri->add_console(
         'sut',
         'vnc-base',
         {
             hostname => 'localhost',
             connect_timeout => 3,
-            port => 5900 + $bmwqemu::vars{VNC},
+            port => $vncport,
             description => q{QEMU's VNC}});
 
     $vnc->backend($self);
     my $ret = $self->select_console({testapi_console => 'sut'});
-    die $ret->{error} if $ret->{error};
 
+    if ($ret->{error}) {
+        bmwqemu::fctinfo("VNC port details:\n" . _port_details($vncport));
+        die $ret->{error};
+    }
     if ($vars->{NICTYPE} eq 'tap') {
         $self->{allocated_networks} = $num_networks;
         $self->{allocated_tap_devices} = \@tapdev;
@@ -1221,6 +1226,29 @@ sub check_socket ($self, $fh, $write = undef) {
         return 1;
     }
     return $self->SUPER::check_socket($fh);
+}
+
+sub _extract_hostfwd_ports ($args) { return $args =~ /hostfwd=(?:tcp|udp)::(\d+)-/g; }
+
+sub _port_details ($port) {
+    my @ss_output = qx{ss -tlnp 2>/dev/null};
+    my @port_info = grep { /\b$port\b/ } @ss_output;
+    return @port_info ? join('', @port_info) : 'no details found';
+}
+
+sub _assert_port_availability ($port, $service) {
+    bmwqemu::fctinfo("checking $port port availability");
+    my $sock = IO::Socket::IP->new(
+        PeerAddr => 'localhost',
+        PeerPort => $port,
+        Type => SOCK_STREAM,
+        Timeout => 1,
+    );
+    if ($sock) {
+        $sock->close;
+        die "Port $port ($service) is already in use\n" . _port_details($port);
+    }
+    return 0;
 }
 
 sub freeze_vm ($self, @) {
--- a/t/18-backend-qemu.t
+++ b/t/18-backend-qemu.t
@@ -814,6 +814,39 @@ subtest 'special cases when starting QEMU' => sub {
     };
 };
 
+subtest 'extract hostfwd ports from NICTYPE_USER_OPTIONS' => sub {
+    is_deeply [backend::qemu::_extract_hostfwd_ports('hostfwd=tcp::39100-:39100')], [39100], 'single tcp hostfwd';
+    is_deeply [backend::qemu::_extract_hostfwd_ports('hostfwd=tcp::39100-:39100,hostfwd=tcp::5555-:22')],
+      [39100, 5555], 'multiple tcp hostfwd';
+    is_deeply [backend::qemu::_extract_hostfwd_ports('hostfwd=udp::1234-:1234')], [1234], 'udp hostfwd';
+    is_deeply [backend::qemu::_extract_hostfwd_ports('restrict=on')], [], 'no hostfwd options';
+    is_deeply [backend::qemu::_extract_hostfwd_ports('')], [], 'empty string';
+};
+
+subtest 'port availability checks' => sub {
+    my %initial_vars = %bmwqemu::vars;
+    my $sock_mock = Test::MockModule->new('IO::Socket::IP');
+    subtest VNC => sub {
+        combined_like { lives_ok { backend::qemu::_assert_port_availability(5991, 'test') } 'free port passes' }
+        qr/checking 5991 port availability/, 'logs port post-qemu check for VNC';
+        $backend_mock->redefine(select_console => sub { {error => 'VNC connection failed'} });
+        combined_like {
+            throws_ok { $backend->start_qemu } qr/VNC connection failed/, 'dies on VNC error'
+        } qr/VNC port details/, 'logs port details on VNC failure';
+        $backend_mock->redefine(select_console => undef);
+    };
+    subtest hostfwd => sub {
+        $bmwqemu::vars{NICTYPE} = 'user';
+        $bmwqemu::vars{NICTYPE_USER_OPTIONS} = 'hostfwd=tcp::39100-:39100';
+        combined_like { lives_ok { backend::qemu::_assert_port_availability(39100, 'test') } 'free port passes' }
+        qr/checking 39100 port availability/, 'logs port pre-qemu check with hostfwd';
+        $sock_mock->redefine(new => sub { Test::MockObject->new->set_true('close') });
+        combined_like { throws_ok { $backend->start_qemu } qr/Port 39100 \(hostfwd\) is already in use/, 'dies on hostfwd port conflict' }
+          qr/qemu version/si, 'expected logs before hostfwd port check';
+    };
+    %bmwqemu::vars = %initial_vars;
+};
+
 subtest 'special cases when handling QMP command' => sub {
     my $create_virtio_console_fifo_called;
     # uncoverable statement count:2


Martchus (commented):
`s/where the some network ports/where some network ports/`
`s/clearier/clearer/`
`s/What is not address is /What is not addressed is /`

Possibly there are more mistakes. You could try to let an LLM fix the spelling.

---

Looks generally good. I guess the alternative would be to check for "Address already in use" in the log output *after* we've ran into such an error and print the `ss` output then. Checking this beforehand is probably ok in this case, too.

backend/qemu.pm
```diff
@@ -1224,6 +1228,30 @@ sub check_socket ($self, $fh, $write = undef) {
     return $self->SUPER::check_socket($fh);
 }
 
+sub _extract_hostfwd_ports ($args) {
+    my @ports;
+    push @ports, $1 while $args =~ /hostfwd=(?:tcp|udp)::(\d+)-/g;
```
okurz: Looks like you should be able to use `grep { ... }` here

backend/qemu.pm
```diff
@@ -1224,6 +1228,30 @@ sub check_socket ($self, $fh, $write = undef) {
...
+    return @ports;
+}
+
+sub _assert_port_available ($port, $service) {
+    my $sock = IO::Socket::INET->new(
+        LocalAddr => '0.0.0.0',
```
okurz: Does this have to be IPv4? How about "localhost"?

backend/qemu.pm
```diff
@@ -1024,6 +1026,8 @@ sub start_qemu ($self) {
 
         if ($vars->{VNC}) {
             my $vncport = $vars->{VNC} !~ /:/ ? ":$vars->{VNC}" : $vars->{VNC};
+            my ($display) = $vncport =~ /:(\d+)/;
+            _assert_port_available(5900 + $display, "VNC display $vncport") if defined $display;
```
okurz: This approach can be racy as the port might be used between this check and when we actually start qemu later. 
The ticket state that either the situation is prevented or the blocking port is shown *after* qemu failed with its rather specific error message that it has already.
Previously qemu would state an explicit error about hostfwd. Now you write "Port $port ($service) is already in use:\n$details\n" but with that we loose the context why isotovideo even needs this port.
Can you explain why you chose to check before and not after? Unless there is a good reason please let qemu fall, catch the error and add the details, as defined in the ticket

okurz (changes requested):

backend/qemu.pm
d3flex: Admittedly, I didnt think about the race. But I found it very unlike situation. The worker wouldnt start another job without terminate the previous, so how likely is to have the same port open (at least from the worker). 

> Previously qemu would state an explicit error about hostfwd. Now you write "Port $port ($service) is already in use:\n$details\n" but with that we loose the context why isotovideo even needs this port.

This confused me a lot. I checked a few references and the error looks like that:
from https://progress.opensuse.org/issues/198428
```bash
[2026-03-18T07:59:30.316975Z] [debug] [pid:1030500] Waiting for 1 attempts
[2026-03-18T07:59:30.317256Z] [info] [pid:1030500] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  QEMU terminated before QMP connection could be established. Check for errors below
[2026-03-18T07:59:30.317586Z] [info] [pid:1030500] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[2026-03-18T07:59:30.318862Z] [debug] [pid:1030500] Passing remaining frames to the video encoder
[2026-03-18T07:59:30.322224Z] [debug] [pid:1030500] Waiting for video encoder to finalize the video
[2026-03-18T07:59:30.322297Z] [debug] [pid:1030500] The built-in video encoder (pid 1030627) terminated
[2026-03-18T07:59:30.323425Z] [debug] [pid:1030500] QEMU: QEMU emulator version 10.0.8 (SUSE Linux 16)
[2026-03-18T07:59:30.323488Z] [debug] [pid:1030500] QEMU: Copyright (c) 2003-2025 Fabrice Bellard and the QEMU Project developers
[2026-03-18T07:59:30.323560Z] [warn] [pid:1030500] !!! : qemu-system-x86_64: -vnc :110,share=force-shared: Failed to find an available port: Address already in use
[2026-03-18T07:59:30.324583Z] [debug] [pid:1030500] sending magic and exit
[2026-03-18T07:59:30.324984Z] [debug] [pid:1030299] received magic close
[2026-03-18T07:59:30.334694Z] [debug] [pid:1030299] backend process exited: 0
[2026-03-18T07:59:30.435487Z] [warn] [pid:1030299] !!! main: failed to start VM at /usr/lib/os-autoinst/backend/driver.pm line 104.
  	backend::driver::start_vm(backend::driver=HASH(0x564c9278dc20)) called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Backend.pm line 18
  	OpenQA::Isotovideo::Backend::new("OpenQA::Isotovideo::Backend") called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Runner.pm line 110
  	OpenQA::Isotovideo::Runner::create_backend(OpenQA::Isotovideo::Runner=HASH(0x564c8cc8b110)) called at /usr/lib/os-autoinst/OpenQA/Isotovideo/Runner.pm line 252
  	OpenQA::Isotovideo::Runner::init(OpenQA::Isotovideo::Runner=HASH(0x564c8cc8b110)) called at /usr/bin/isotovideo line 184
```
and the reason is `Reason: backend died: QEMU exited unexpectedly, see log for details`

which context exactly is missing? Now the reason shows `Reason: backend died: Port 5991 (VNC display :91) is already in use:`

and the os-autoinst:

```bash
[2026-04-24T16:09:08.920149+02:00] [info] [pid:1982221] ::: backend::baseclass::die_handler: Backend process died, backend errors are reported below in the following lines:
  Port 5991 (VNC display :91) is already in use:
  LISTEN 0      1                          0.0.0.0:5991       0.0.0.0:*                                                
[2026-04-24T16:09:08.921197+02:00] [info] [pid:1982221] ::: OpenQA::Qemu::Proc::save_state: Saving QEMU state to qemu_state.json
[2026-04-24T16:09:08.922600+02:00] [debug] [pid:1982221] Passing remaining frames to the video encoder
[image2pipe @ 0x5628cd89ccc0] Could not find codec parameters for stream 0 (Video: ppm, none): unspecified size
Consider increasing the value for the 'analyzeduration' (0) and 'probesize' (5000000) options
Input #0, image2pipe, from 'fd:':
  Duration: N/A, bitrate: N/A
  Stream #0:0: Video: ppm, none, 24 fps, 24 tbr, 24 tbn
Output #0, webm, to 'video.webm':
[out#0/webm @ 0x5628cd8b2780] Output file does not contain any stream
Error opening output file video.webm.
Error opening output files: Invalid argument
[2026-04-24T16:09:08.939572+02:00] [debug] [pid:1982221] Waiting for video encoder to finalize the video
[2026-04-24T16:09:08.939755+02:00] [debug] [pid:1982221] The built-in video encoder (pid 1982240) terminated
[2026-04-24T16:09:08.939826+02:00] [debug] [pid:1982221] The external video encoder (pid 1982239) terminated
[2026-04-24T16:09:08.940897+02:00] [debug] [pid:1982221] sending magic and exit
[2026-04-24T16:09:08.941447+02:00] [debug] [pid:1982144] received magic close
[2026-04-24T16:09:08.951718+02:00] [debug] [pid:1982144] backend process exited: 0
[2026-04-24T16:09:09.053205+02:00] [warn] [pid:1982144] !!! main: failed to start VM at /home/iob/openqadev/os-autoinst/script/../backend/driver.pm line 104.
  	backend::driver::start_vm(backend::driver=HASH(0x55e45a5fa4d8)) called at /home/iob/openqadev/os-autoinst/script/../OpenQA/Isotovideo/Backend.pm line 18
  	OpenQA::Isotovideo::Backend::new("OpenQA::Isotovideo::Backend") called at /home/iob/openqadev/os-autoinst/script/../OpenQA/Isotovideo/Runner.pm line 110
  	OpenQA::Isotovideo::Runner::create_backend(OpenQA::Isotovideo::Runner=HASH(0x55e452290848)) called at /home/iob/openqadev/os-autoinst/script/../OpenQA/Isotovideo/Runner.pm line 252
  	OpenQA::Isotovideo::Runner::init(OpenQA::Isotovideo::Runner=HASH(0x55e452290848)) called at /home/iob/openqadev/os-autoinst/script/isotovideo line 190
```

which firstly satisfies the AC1 -> openQA tests should not fail with Address Already in use or Could not set up host forwarding rule error or providing very clear feedback which non-os-autoinst process is blocking

How would you like to proceed?

backend/qemu.pm
d3flex: why localhost? shouldnt that check all the interfaces? however I think you have a point because this would not see ipv6, right?

backend/qemu.pm
d3flex: grep? no..

okurz: Why not?

backend/qemu.pm
okurz: yes, 0.0.0.0 is IPv4 only

backend/qemu.pm
okurz: > which context exactly is missing?

Well, you showed the example where it is visible that qemu tries hostfwd and hence tries to access a port. With your changes we just know that the port is not available. If anyone would see that message they would not know why os-autoinst checks that. So either you try to carefully replicate what qemu would tell you, e.g. "Requested port $port via qemu parameter $param' is occupied by process $process, aborting", or let qemu fail as before, detect that and only add the additional output from `ss`

> How would you like to proceed?

I prefer the second: Let qemu fail, just add additional information instead of trying to be "smart" with much more Perl code

okurz (approved):

Martchus (changes requested):
This contains lots of disabled code and I had to unresolve a thread from @okurz again because it hasn't been resolved yet. It looks like you've accidentally pushed a wrong commit now.

okurz (changes requested):
^^

okurz (approved):

Martchus:
I guess this is generally acceptable. I mentioned that the probably better alternative would be to check this afterwards and also @okurz stated that. Have you at least looked into this?

I must also say that the exclusion of the VNC port isn't nice. If I had to decide whether do this check only for one type of port I would have chosen the VNC port because that port is used by almost all tests so almost all tests would benefit from that. With this change only tests that use `NICTYPE_USER_OPTIONS` benefit.

Note that the PR title is no longer in-line with what's actually proposed for merging as the title still states that it checks the VNC port. The PR description also still contains the typos I mentioned before.

okurz (changes requested):
Points from conversation are still open. Description says something about VPN. Didn't find a reason why we should assert before, how that helps

Martchus:
> Description says something about VPN.

Oh, right - I misread "VNC". But VPN makes even less sense.

t/18-backend-qemu.t
```diff
@@ -814,6 +814,41 @@ subtest 'special cases when starting QEMU' => sub {
...
+subtest 'port availability checks' => sub {
+    my %initial_vars = %bmwqemu::vars;
+    my $sock_mock = Test::MockModule->new('IO::Socket::IP');
+    subtest VNC => sub {
+        combined_like { lives_ok { backend::qemu::_assert_port_available(5991, 'test') } 'free port passes' }
+        qr/checking 5991 port availability/, 'logs port post-qemu check for VNC';
+        $backend_mock->redefine(select_console => sub { {error => 'VNC connection failed'} });
+        combined_like {
+            throws_ok { $backend->start_qemu } qr/VNC connection failed/, 'dies on VNC error'
+        } qr/VNC port details/, 'logs port details on VNC failure';
```
Martchus, line 835: How do the details look like? I'm afraid it'll be multi-line data within `[…]` which wouldn't look nice.

backend/qemu.pm
```diff
@@ -1223,6 +1228,29 @@ sub check_socket ($self, $fh, $write = undef) {
...
 
+sub _extract_hostfwd_ports ($args) { return $args =~ /hostfwd=(?:tcp|udp)::(\d+)-/g; }
+
+sub _port_details ($port) {
+    my $ss_output = qx{ss -tlnp 2>/dev/null};
+    my @port_info = grep { /\b$port\b/ } split /\n/, $ss_output;
+    return @port_info ? join("\n", @port_info) : 'no details found';
+}
+
+sub _assert_port_available ($port, $service) {
```
Martchus: In accordance with the log message that already uses the correct form:
```suggestion
sub _assert_port_availability ($port, $service) {
```

Martchus (changes requested):

t/18-backend-qemu.t
d3flex: image

we just add the ss output. the rest should be as it is

Martchus: The multi-line output of `ss` doesn't look good within the `[…]` as expected. It would look better it it was printed like this:

```
… VNC port details:
  LISTEN …
  LISTEN …
```

backend/qemu.pm
```diff
@@ -1223,6 +1228,29 @@ sub check_socket ($self, $fh, $write = undef) {
     return $self->SUPER::check_socket($fh);
 }
 
+sub _extract_hostfwd_ports ($args) { return $args =~ /hostfwd=(?:tcp|udp)::(\d+)-/g; }
+
+sub _port_details ($port) {
+    my $ss_output = qx{ss -tlnp 2>/dev/null};
+    my @port_info = grep { /\b$port\b/ } split /\n/, $ss_output;
+    return @port_info ? join("\n", @port_info) : 'no details found';
```
perlpunk:
```suggestion
    my @ss_output = qx{ss -tlnp 2>/dev/null};
    my @port_info = grep { /\b$port\b/ } @ss_output;
    return @port_info ? join('', @port_info) : 'no details found';
```

t/18-backend-qemu.t
```diff
@@ -814,6 +814,41 @@ subtest 'special cases when starting QEMU' => sub {
...
+        $bmwqemu::vars{NICTYPE_USER_OPTIONS} = 'hostfwd=tcp::39100-:39100';
+        combined_like { lives_ok { backend::qemu::_assert_port_availability(39100, 'test') } 'free port passes' }
+        qr/checking 39100 port availability/, 'logs port pre-qemu check with hostfwd';
+        $sock_mock->redefine(new => sub { Test::MockObject->new->set_true('close') });
+        combined_like { throws_ok { $backend->start_qemu } qr/Port 39100 \(hostfwd\) is already in use/, 'dies on hostfwd port conflict' }
+          qr/qemu version/si, 'expected logs before hostfwd port check';
+    };
+    %bmwqemu::vars = %initial_vars;
+
+
```
perlpunk:
```suggestion
```

okurz (approved):

backend/qemu.pm
okurz: So I understand that the approach of this PR is still not what was suggested to handle failures but rather do a pre-check. Anyway, let's see how it goes this way.

perlpunk (approved):

backend/qemu.pm
```diff
@@ -1058,20 +1060,23 @@ sub start_qemu ($self) {
...
-            port => 5900 + $bmwqemu::vars{VNC},
+            port => $vncport,
             description => q{QEMU's VNC}});
 
     $vnc->backend($self);
     my $ret = $self->select_console({testapi_console => 'sut'});
-    die $ret->{error} if $ret->{error};
 
+    if ($ret->{error}) {
+        bmwqemu::fctinfo("VNC port details \n" . _port_details($vncport));
```
Martchus:
```suggestion
        bmwqemu::fctinfo("VNC port details:\n" . _port_details($vncport));
```

Martchus (changes requested):

Martchus (approved):

PR #2894, opened 2026-04-22
style: Enforce perlcritic policy RequireQuotedHeredocTerminator

perlpunk:
Issue: https://progress.opensuse.org/issues/155191

ValuesAndExpressions::RequireQuotedHeredocTerminator


okurz (commented):
this looks unusual and hard to remember. I would suggest to exclude this rule

perlpunk:
> this looks unusual and hard to remember. I would suggest to exclude this rule

I switched to indented heredocs in the same change. I can revert that. The rule does not require that.

Maybe we can discuss the use of indented heredocs in another PR. We use it already elsewhere.

perlpunk:
I reduced the change to only add the quotes enforced by the policy.

okurz (approved):

d3flex (approved):

PR #2895, opened 2026-04-22
feat: set job to incomplete when perl package is not in @INC

d3flex:
Use the error in the try/catch to trace which reports the `Can't locate.. in \@INC` and update the JSON representation of the process to disk via `serialize_state`. Note that the function makes use of STATE_FILE. As a result the WebUI will render the msg as reason and it will mark the corresponding tests as incomplete.

At that point, as corresponding failures are considered issues originated by both worker and test distris. A follow up task could separate the origin in order to tackle each of the categories differently.

Issue: https://progress.opensuse.org/issues/197183

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -301,7 +301,11 @@ sub loadtest ($script, %args) {
         }
         my $msg = "error on $script: $err";
         bmwqemu::fctwarn($msg);
-        bmwqemu::serialize_state(component => 'tests', msg => "unable to load $script, check the log for the cause (e.g. syntax error)");
+        my ($is_missing_module) = $err =~ /(Can't locate .+ in \@INC)/;
+        my ($state_msg) = $is_missing_module
+          ? "Missing Perl module while loading $script: $is_missing_module"
+          : "unable to load $script, check the log for the cause (e.g. syntax error)";
+        bmwqemu::serialize_state(component => 'tests', msg => $state_msg, result => 'incomplete');
         die $msg;
     }
     $test = $name->new($category);
--- a/basetest.pm
+++ b/basetest.pm
@@ -334,6 +334,14 @@ sub runtest ($self) {
         if (!@{$self->{details}} || ($self->{details}->[-1]->{result} || '') ne 'fail') {
             $self->take_screenshot();
         }
+        if (!$internal && $e =~ /Can't locate .+ in \@INC/) {
+            my $msg = "# Test died with missing dependency: $e";
+            bmwqemu::fctinfo($msg);
+            $self->record_resultfile('Failed', $msg, result => 'fail');
+            $self->{fatal_failure} = 1;
+            bmwqemu::serialize_state(component => 'tests', msg => "Missing Perl module: $e", result => 'incomplete');
+            $died = 1;
+        }
         # show a text result with the die message unless the die was internally generated
         if (!$internal) {
             my $msg = "# Test died: $e";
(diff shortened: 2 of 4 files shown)


autotest.pm
```diff
@@ -301,7 +301,10 @@ sub loadtest ($script, %args) {
         }
         my $msg = "error on $script: $err";
         bmwqemu::fctwarn($msg);
-        bmwqemu::serialize_state(component => 'tests', msg => "unable to load $script, check the log for the cause (e.g. syntax error)");
+        my $state_msg = $err =~ /Can't locate .+ in \@INC/
+          ? "Missing Perl module while loading $script: $err"
+          : "unable to load $script, check the log for the cause (e.g. syntax error)";
+        bmwqemu::serialize_state(component => 'tests', msg => $state_msg, result => 'incomplete');
```
Martchus, line 308: This sets the result to incomplete regardless of whether it is a `Can't locate …` error. It might not make a difference here but to be sure we should probably only pass `result => 'incomplete'` if it is a `Can't locate …` error.

basetest.pm
```diff
@@ -334,6 +334,14 @@ sub runtest ($self) {
         if (!@{$self->{details}} || ($self->{details}->[-1]->{result} || '') ne 'fail') {
             $self->take_screenshot();
         }
+        if (!$internal && $e =~ /Can't locate .+ in \@INC/) {
```
Martchus, line 337: The regex/check and formatting of the error message could be extracted into a helper and not repeated here again.

t/17-basetest.t
```diff
@@ -282,10 +282,32 @@ subtest 'number of test results is limited' => sub {
...
+    ok -e bmwqemu::STATE_FILE, 'state file was written';
+    my $state = decode_json(path(bmwqemu::STATE_FILE)->slurp);
+    is $state->{result}, 'incomplete', 'result is incomplete for missing Perl module';
+    # like $state->{msg}, qr/Can't locate/, 'message mentions the missing module';
+    ok $basetest->{fatal_failure}, 'failure is fatal';
+    path(bmwqemu::STATE_FILE)->remove if -e bmwqemu::STATE_FILE;
+};
 
+delete $bmwqemu::vars{MAX_TEST_STEPS};
+# done_testing;
```
perlpunk: Leftover?

basetest.pm
d3flex: I had the same thought but I came to think that it is simpler to keep it as is and not overengineer the approach.

autotest.pm
d3flex: I remove the `msg` the webui will display `Reason: died: terminated prematurely, see log output for details`. Which is more general reason than ` Missing Perl module while loading tests/installation/welcome.pm....`. 
or I didnt get what do you mean by making sure?!

okurz (approved):

Martchus (approved):
Let's hope the slight change that affects also other types of errors mentioned in https://github.com/os-autoinst/os-autoinst/pull/2895/changes#r3125215173 won't change anything we don't want.

t/17-basetest.t
```diff
@@ -282,6 +282,28 @@ subtest 'number of test results is limited' => sub {
...
+    local $bmwqemu::vars{MAX_TEST_STEPS} = 100;
+    my $basetest = bless {details => [], name => 'foo', fullname => 'foo', category => 'category1'}, 'basetest';
+    my $logs = combined_from { dies_ok { $basetest->runtest } 'runtest dies due to missing module' };
+    like $logs, qr/Can't locate.*\@INC/, 'catch expected error message';
+
+    # Verify state file has incomplete result
+    ok -e bmwqemu::STATE_FILE, 'state file was written';
+    my $state = decode_json(path(bmwqemu::STATE_FILE)->slurp);
+    is $state->{result}, 'incomplete', 'result is incomplete for missing Perl module';
+    # like $state->{msg}, qr/Can't locate/, 'message mentions the missing module';
```
perlpunk, line 304: Leftover?

PR #2897, opened 2026-04-24
test: Remove CoverageWorkaround and check coverage diff

perlpunk:
Just as a demo for a potential Devel::Cover bugreport


perlpunk:
The coverage report shows that code in two `try` / `catch` blocks doesn't count as lines to cover any more:
coverage-workaround-codecov-missing-lines
So they are not reported as missing, but as the same as empty or comment lines, while with the CoverageWorkaround, they were actually covered.

perlpunk:
It also seems to be completely random. I just added one line and pushed again, and now many more indirect changes are reported.

perlpunk:
I reported this here now: https://github.com/pjcj/Devel--Cover/issues/466
Maybe @pjcj can have a look during the workshop

perlpunk:
The fix will be in the next Devel::Cover release

PR #2901, opened 2026-04-24
chore(mergify): prevent the unstable work-in-progress check to block us

okurz:


perlpunk (approved):
I assume the fact that the WIP check is pending when it's broken will still stop us, but let's try

okurz:
> I assume the fact that the WIP check is pending when it's broken will still stop us, but let's try

As a funny coincidence here WIP just directly "succeeded" :)

Martchus (approved):

okurz:
> I assume the fact that the WIP check is pending when it's broken will still stop us, but let's try

To clarify my intentions: Yes, mergify mayb still block on "pending checks" but looking into the number additionally is IMHO not necessary as I would expect WIP to show up immediately or not at all or stay pending.

Need to merge manually as mergify config was changed.

PR #2903, opened 2026-04-27
fix: Disable duplicate address detection in NetworkManager for br1

gauravpathak:
Ticket: https://progress.opensuse.org/issues/199022

By default, NetworkManager performs (DAD) Duplicate Address Detection via ARP before it actually binds an IPv4 address to an interface.  
This commit disables DAD in NM for ovs-bridge-br1 and ovs-interface-br1 by setting dad-timeout to zero for ipv4 configuration.


okurz:
How is this different to before?

gauravpathak:
> How is this different to before?

`ipv6.dad-timeout` was causing the issue, so I only used `ipv4.dad-timeout`.

Martchus (approved):

okurz (approved):

PR #2906, opened 2026-04-27
feat: support generating UEFI vars from JSON via virt-fw-vars

okurz:
Motivation:
Avoid the need for custom OVMF builds with pre-enrolled certificates or specific
default resolutions. This allows using official OVMF packages while still
supporting custom Secure Boot certificates and settings.

Design Choices:
Updated OpenQA::Qemu::Proc to detect .json extension in UEFI_PFLASH_VARS.
When detected, it uses virt-fw-vars to generate a variables store on the fly.
The generated file is stored in the QEMU instance's block device base directory.

Benefits:
Increased flexibility for UEFI tests and reduced maintenance burden for custom
firmware packages.

Related issue: https://progress.opensuse.org/issues/200156


okurz:
@Vogtinator this is my idea how we could pass a JSON file with "FW vars" to an openQA job. But, what to put into that JSON file to replicate what your current custom ovmf files do? Can you help with that?

Vogtinator:
> @Vogtinator this is my idea how we could pass a JSON file with "FW vars" to an openQA job. But, what to put into that JSON file to replicate what your current custom ovmf files do? Can you help with that?

You can convert the current vars to json using e.g. `virt-fw-vars --output-json /dev/stdout -i /usr/share/qemu/aavmf-aarch64-unsupported+ms-vars.bin`, or create the json "from scratch" by applying the modifications to json instead of the vars volume in place, e.g. with `virt-fw-vars --secure-boot --enroll-cert openSUSE-Factory-Staging.crt --no-microsoft --add-db a0baa8a3-041d-48a8-bc87-c36d121b5e3d openSUSE-Factory-Staging.crt --output-json /dev/stdout --set-json /dev/stdin < platformconfig.json`

Martchus:
PR for os-autoinst to create the JSON on its own, similar to what is done so far in the spec file: https://github.com/os-autoinst/os-autoinst/pull/2954

okurz:
https://github.com/os-autoinst/os-autoinst/pull/2954

PR #2908, opened 2026-04-28
fix: Clean up the last VM disks to avoid disk image creation failure

nanzhg:
Guest start up failed because VM disk images were not cleaned up from the last running. To fix it, just removing all related disks to ensure the new disk images can be created properly.
https://openqa.suse.de/tests/22052008#step/bootloader_svirt/34

- Related ticket: https://progress.opensuse.org/issues/194843#note-12
- Verification run: http://10.146.15.126/tests/14

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -329,8 +329,9 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
         my $cmd =
           "( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk \'/$name/ { print \$1 }\');" .
           'if [ $vmid ]; then ' .
-          'vim-cmd vmsvc/power.off $vmid;' .
+          'vim-cmd vmsvc/power.off $vmid; sleep 3;' .
           'vim-cmd vmsvc/destroy $vmid;' .
+          "else rm -f $name*;" .
           'fi;' .
           "vmkfstools -v1 -U $vmware_disk_path;" .
           "vmkfstools -v1 -c $size --diskformat thin $vmware_disk_path; sleep 10 ) 2>&1";
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -325,12 +325,14 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
     if ($self->vmm_family eq 'vmware') {
         my $vmware_disk_path = $vmware_openqa_datastore . $file;
         # Power VM off, delete it's disk image, and create it again.
-        # Than wait for some time for the VM to *really* turn off.
+        # Poll the power state until the VM is *really* turned off.
         my $cmd =
           "( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk \'/$name/ { print \$1 }\');" .
           'if [ $vmid ]; then ' .
           'vim-cmd vmsvc/power.off $vmid;' .
+          'for i in $(seq 1 10); do vim-cmd vmsvc/power.getstate $vmid | grep -q "Powered off" && break; sleep 1; done;' .
           'vim-cmd vmsvc/destroy $vmid;' .
+          "else rm -f $name*;" .
           'fi;' .
           "vmkfstools -v1 -U $vmware_disk_path;" .
           "vmkfstools -v1 -c $size --diskformat thin $vmware_disk_path; sleep 10 ) 2>&1";


okurz (approved):

consoles/sshVirtsh.pm
```diff
@@ -329,8 +329,8 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
         my $cmd =
           "( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk \'/$name/ { print \$1 }\');" .
           'if [ $vmid ]; then ' .
-          'vim-cmd vmsvc/power.off $vmid;' .
-          'vim-cmd vmsvc/destroy $vmid;' .
+          'vim-cmd vmsvc/power.off $vmid && vim-cmd vmsvc/destroy $vmid;' .
```
Martchus: This change introduces command chaining via `&&` so the 2nd command is only executed if the 1st command was successful. Is that intentional?

Martchus (changes requested):
Not only the PR description but also the commit message should contain an explanation and the ticket reference.

consoles/sshVirtsh.pm
RoyCai7: && means: only destroy if power-off succeeds
; means: always try destroy

consoles/sshVirtsh.pm
```diff
@@ -329,8 +329,8 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
...
+          "else rm -f $name*;" .
```
RoyCai7, line 335: A safer approach is to delete only the expected disk

consoles/sshVirtsh.pm
nanzhg: Actually, there are two cases.
For case 1, VM is in 'Powered off', executing commands with `;` by sequence without any impact to remove the disk.
For case 2, VM is still in 'Powered on' by the previous job failed, I'm worried that if the first command "power.off" is not fully complete, the second command "destroy" might cause an error like "device or resource busy".

So maybe a few seconds sleep `sleep 3` between these two commands will solve both cases if we keep using  `;` .

consoles/sshVirtsh.pm
nanzhg: Based on the guest name, it will delete all related disk files generated during the guest creation process., e.g.
```
openQA-SUT-8-1.scoreboard
openQA-SUT-8-2.scoreboard
openQA-SUT-8-Snapshot1.vmsn
openQA-SUT-8.scoreboard
openQA-SUT-8.vmsd
openQA-SUT-8.vmx
openQA-SUT-8b-000001-sesparse.vmdk
openQA-SUT-8b-000001.vmdk
openQA-SUT-8b-flat.vmdk
openQA-SUT-8b.vmdk
```

consoles/sshVirtsh.pm
nanzhg: Yes, I know.

Martchus: And considering that there's a difference this raises the question whether the change was intended.

nanzhg: Yes, just want to ensure that the power state change is complete.

consoles/sshVirtsh.pm
```diff
@@ -330,7 +330,9 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
           "( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk \'/$name/ { print \$1 }\');" .
           'if [ $vmid ]; then ' .
           'vim-cmd vmsvc/power.off $vmid;' .
+          'sleep 3;' .
           'vim-cmd vmsvc/destroy $vmid;' .
+          "else rm -f $name*;" .
```
waynechen55, line 335: Any race condition will be introduced by forcible removal ?

consoles/sshVirtsh.pm
alice-suse: In my last time's debug for vmware issue, I notice some people backup VM files for debug purpose in the directory which can fall into name pattern "$name*". That's not encouraged indeed , but it happens. How about doing more strict remove -- only the NECESSARY files that openqa uses to boot up openQA-SUT-[N] (the disk files, the configuration files )?

consoles/sshVirtsh.pm
nanzhg: It should not happen, it's unique name controlled by `VIRSH_INSTANCE` while multiple jobs running in parallel.

consoles/sshVirtsh.pm
nanzhg: I think it is necessary to clean up all VM name related files **before creating a new guest**. We'd ever hit such problem, such as a boot issue (referring a wrong nvram from the last run). Meanwhile, I have a test to run command `rm -f $name*`, it won't remove any directory.
```
[root@orion:/vmfs/volumes/671e5654-72978702-d7b1-ac1f6b4729e4/openQA] ll openQA-SUT-8*
-rw-r--r--    1 root     root             0 Apr 29 04:51 openQA-SUT-8.vmdk
-rw-r--r--    1 root     root             0 Apr 29 04:51 openQA-SUT-8.vmx

openQA-SUT-8:
total 1152
drwxr-xr-x    1 root     root         69632 Apr 29 04:51 .
drwxr-xr-x    1 root     root        139264 Apr 29 04:51 ..
[root@orion:/vmfs/volumes/671e5654-72978702-d7b1-ac1f6b4729e4/openQA] rm -f openQA-SUT-8*
rm: 'openQA-SUT-8' is a directory
```

consoles/sshVirtsh.pm
```diff
@@ -329,8 +329,9 @@ sub _create_disk ($self, $args, $vmware_openqa_datastore, $file, $name, $basedir
         my $cmd =
           "( set -x; vmid=\$(vim-cmd vmsvc/getallvms | awk \'/$name/ { print \$1 }\');" .
           'if [ $vmid ]; then ' .
-          'vim-cmd vmsvc/power.off $vmid;' .
+          'vim-cmd vmsvc/power.off $vmid; sleep 3;' .
```
Martchus: So there's no better way to wait for this except `sleep 3`?

nanzhg: Another way is to check the VM power state `vim-cmd vmsvc/power.getstate $vmid` instead of `sleep`, and determine whether the `destroy` command should be executed. But implementing this requires more code, and will make this statement more complex. I think `sleep` is a simple way to achieve this, and it has already been used in some places within this module.

waynechen55: If you want to introduce ```sleep```, why not split this long command into two ?

consoles/sshVirtsh.pm
waynechen55: I think more specific removal is better for this case because if certain VM fails to boot up due to pre-existing disk then only this disk should be removed according to my understanding. Could you help explain why you need to remove them all ?

consoles/sshVirtsh.pm
nanzhg: Updated.

okurz: Maybe that wasn't clear but Martchus (and me) want to ensure that there is *no* sleep in the code.

You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

We are waiting because some condition changes some time ... keyword(s) is/are *some time*. It is important to know the exact condition that the code should wait for and check for that instead of delaying program execution for an arbitrary amount of time. For openQA tests in general just try to do what you as a human user would also do. You would not look on your watch waiting for exactly 100s before you look at the screen again, right? :)

Also see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

okurz (changes requested):

Martchus:
Nothing has changed; there's still the `sleep 3` and also an unanswered question by @waynechen55. (I'm only writing this comment as I was asked to re-review.)

Martchus (changes requested):

consoles/sshVirtsh.pm
nicksinger: how about using the following:
```bash
export timeout=3; timeout -t ${timeout} sh -c "while ! \$(vim-cmd vmsvc/power.getstate $vmid | grep -i on > /dev/null); do sleep 0.1; done"
```

we do something similar for e.g. the HMC ppc backend:  https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/bootloader_pvm.pm#L250

consoles/sshVirtsh.pm
nanzhg: I really can't think of a reason why we didn't clean up those VM disks along with related configuration files generated by the previous job & test VM. Especially it often causes VM creation failures and "ghost" VM which needs to be removed manually via the ESXi dashboard.

consoles/sshVirtsh.pm
nanzhg: Thank you for the suggestion! I've updated code to check the power state in a loop.

nanzhg:
> Nothing has changed; there's still the `sleep 3` and also an unanswered question by @waynechen55. (I'm only writing this comment as I was asked to re-review.)

Sorry for delay. Please kindly review again my latest commit. Thank you!

nanzhg:
VR: http://10.146.15.126/tests/14/logfile?filename=autoinst-log.txt

> [2026-06-02T17:25:16.294600+08:00] [debug] [pid:27701] [run_ssh_cmd(( set -x; vmid=$(vim-cmd vmsvc/getallvms | awk '/openQA-SUT-12/ { print $1 }');if [ $vmid ]; then vim-cmd vmsvc/power.off $vmid;for i in $(seq 1 10); do vim-cmd vmsvc/power.getstate $vmid | grep -q "Powered off" && break; sleep 1; done;vim-cmd vmsvc/destroy $vmid;fi;vmkfstools -v1 -U /vmfs/volumes/datastore1/openQA/openQA-SUT-12b.vmdk;vmkfstools -v1 -c 20G --diskformat thin /vmfs/volumes/datastore1/openQA/openQA-SUT-12b.vmdk; sleep 10 ) 2>&1)] exit-code: 0

Martchus (approved):

okurz (approved):

PR #2909, opened 2026-04-28
fix: Avoid restarting openvswitch/NM after OVS bridge setup

d3flex:
configure_openvswitch was restarting both openvswitch and
os-autoinst-openvswitch, which destroyed the bridges NM just created.
Commit https://github.com/d3flex/os-autoinst/commit/3232b807a6bf0eb0428ddf8592a0e6bcc5342ecb tried to fix this by adding a NM restart at the end of
setup_multi_machine_with_networkmanager, but that introduced a race
condition where br1 is torn down and tests fail before it comes back.

The root cause is that the openvswitch restart in configure_openvswitch
is redundant — cleanup_openvswitch_db already provides a clean, running
OVS instance. Remove both the unnecessary openvswitch restart and the
NM restart that was trying to compensate for it.

### DIFF
--- a/script/os-autoinst-setup-multi-machine
+++ b/script/os-autoinst-setup-multi-machine
@@ -177,7 +177,6 @@ setup_multi_machine_with_networkmanager() {
         nmcli con add type tun mode tap owner "$(id -u _openqa-worker)" group "$(getent group nogroup | cut -f3 -d:)" con.int "tap$i" master "tap$i"
     done
     create_gre_preup_script /etc/NetworkManager/dispatcher.d/gre_tunnel_preup.sh
-    systemctl restart NetworkManager
 }
 
 setup_multi_machine_with_wicked() {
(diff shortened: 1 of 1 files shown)


okurz (approved):

Martchus (approved):

Martchus:
Did you test this?

d3flex:
test recovered and passed locally

gauravpathak (approved):

gauravpathak:
The WIP check is blocking the auto merge via mergify.

Martchus:
I guess that version is also good. It would be good if you updated the PR description and commit message when changing the approach of a fix like this.

d3flex:
> I guess that version is also good. It would be good if you updated the PR description and commit message when changing the approach of a fix like this.

the commit was updated. but i think you are talking about the github - makes sense, just lazy

okurz:
@d3flex as you added "first-in-queue", where is the queue?

PR #2910, opened 2026-04-28
feat: disable storage check in CI environments

okurz:
Motivation:
The storage check introduced in bb08a482 protects shared openQA workers from
disk exhaustion. However, ephemeral CI environments (like GitHub Actions) often
have limited disk space but are specifically provisioned for single test runs
where this protection is unnecessary and causes false-positive failures.

Design Choices:
- Modified `_abort_if_storage_limit_exceeded` in `bmwqemu.pm` to return early if
  standard CI environment variables (`CI` or `GITHUB_ACTIONS`) are set.

Benefits:
- Prevents false-positive test failures in GitHub Actions and other CI systems.
- Reduces the need for manual configuration in CI workflows.


alvarocarvajald (approved):
LGTM

ricardobranco777 (approved):

bmwqemu.pm
```diff
@@ -170,6 +170,7 @@ sub default_numdisks ($v = \%vars) {
 }
 
 sub _abort_if_storage_limit_exceeded () {
+    return undef if $ENV{CI} || $ENV{GITHUB_ACTIONS};
```
d3flex, line 173: should those variables be added to doc/backend_vars.md?

doc/backend_vars.md
```diff
@@ -60,8 +60,8 @@ Supported variables per backend
...
-| STORAGE_KEEP_FREE_RATIO | float | 0.2 | Ratio of total storage space to keep free (e.g. 0.2 for 20%). If the total requested HDD size exceeds the available space while keeping this ratio of total storage size free, the job is aborted early with "incomplete" status. Set to 0 to disable this check. |
-| STORAGE_KEEP_FREE_GB | integer | 50 | Absolute free space threshold in GiB. If the total requested HDD size leaves less than this amount of free space, the job can only run if the relative check (`STORAGE_KEEP_FREE_RATIO`) is also satisfied. Set to 0 to disable the absolute threshold. |
+| STORAGE_KEEP_FREE_RATIO | float | 0.2 | Ratio of total storage space to keep free (e.g. 0.2 for 20%). If the total requested HDD size exceeds the available space while keeping this ratio of total storage size free, the job is aborted early with "incomplete" status. Set to 0 to disable this check. Automatically disabled in CI environments (if `CI` or `GITHUB_ACTIONS` is set). |
```
foursixnine: I think being specific here matters

```suggestion
| STORAGE_KEEP_FREE_RATIO | float | 0.2 | Ratio of total storage space to keep free (e.g. 0.2 for 20%). If the total requested HDD size exceeds the available space while keeping this ratio of total storage size free, the job is aborted early with "incomplete" status. Set to 0 to disable this check. Automatically disabled in CI environments (if `CI` or `GITHUB_ACTIONS` is set in the environment). |
```

bmwqemu.pm
okurz: No, those are not os-autoinst variables. They are commonly set in CI environments

doc/backend_vars.md
okurz: done. Consistently for both variables, of course

foursixnine (approved):

perlpunk (approved):

okurz:
All checks fulfilled except for "WIP". Ok, I understood the problem now. I removed "WIP" from the org but in the settings of the individual projects we have set that check as required. Fixed.

PR #2911, opened 2026-04-29
fix(test): handle D-Bus AccessDenied in Open vSwitch test

d3flex:
On machines where the Open vSwitch D-Bus service is running but the user lacks permission, the error is "AccessDenied" rather than "name not provided/activatable" or "Failed to connect". This caused the Open vSwitch D-Bus subtest in t/18-backend-qemu.t to fail locally while still passing in CI


okurz (approved):
It would be better if the test does not even try to reach a system service

Martchus (approved):

PR #2913, opened 2026-04-29
feat: add predefined QEMU hardware profiles support

okurz:
Motivation:
Allow test jobs to select named sets of QEMU arguments (profiles) using a job
setting QEMU_HWPROFILE. This helps catch regressions that only occur on specific
hardware configurations, such as complex PCIe topologies. The initial
implementation used hardcoded profile names, which was refactored to use
capability-based guards for better maintainability.

Design Choices:
- Introduced a new module OpenQA::Qemu::HWProfiles to store named profiles and
  refactored it to eliminate data duplication using shared references.
- Implemented apply_hwprofile in backend::qemu to handle profile selection,
  using a functional pipeline (map/grep/split) and capability-based guards
  (e.g. 'graphics', 'rng', 'usb') to prevent device duplication.
- Added a _profile_has helper to backend::qemu to safely check for capabilities
  instead of checking for specific profile names.
- Stripped leading dashes from custom profile arguments to match
  OpenQA::Qemu::Proc::static_param expectations.
- Documented the new variable in doc/backend_vars.md.

Benefits:
- Facilitates testing with different hardware configurations and simplifies
  integration of complex QEMU command line arguments.
- Decouples backend logic from specific hardware profile data, improving
  legibility and reducing potential for logic errors.
- Provides a flexible way to switch between standard and advanced
  virtualization defaults with zero code duplication in profile definitions.

Related issue: https://progress.opensuse.org/issues/69700

### DIFF
--- a/OpenQA/Qemu/HWProfiles.pm
+++ b/OpenQA/Qemu/HWProfiles.pm
@@ -1,0 +1,34 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Qemu::HWProfiles;
+use Mojo::Base -signatures;
+use Mojo::File qw(path);
+use YAML::PP;
+
+my $serial = [
+    ['chardev', 'ringbuf,id=serial0,logfile=serial0,logappend=on'],
+    ['serial', 'chardev:serial0'],
+];
+
+our %profiles = (
+    default => {
+        default => $serial,
+        provides => [qw(serial)],
+    },
+);
+
+my $external_profiles;
+
+sub get_profile ($name, $arch) {
+    if (!$profiles{$name}) {
+        $external_profiles //= YAML::PP->new->load_file(path(__FILE__)->sibling('hw_profiles.yaml'));
+        return undef unless $external_profiles->{$name};
+        my $p = $external_profiles->{$name};
+        return {args => ($p->{$arch} // $p->{default} // []), provides => ($p->{provides} // [])};
+    }
+    my $p = $profiles{$name};
+    return {args => ($p->{$arch} // $p->{default} // []), provides => ($p->{provides} // [])};
+}
+
+1;
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -21,6 +21,7 @@ use Fcntl;
 use Net::DBus;
 use bmwqemu qw(diag);
 require IPC::System::Simple;
+use OpenQA::Qemu::HWProfiles;
 use osutils qw(find_bin qv run_diag runcmd port_details);
 use List::Util qw(first max);
 use Data::Dumper;
@@ -682,6 +683,33 @@ sub determine_qemu_version ($self, $qemubin) {
     $self->{qemu_version} = $qemu_version;
 }
 
+sub apply_hwprofile ($self) {
+    my $vars = \%bmwqemu::vars;
+    my $name = $vars->{QEMU_HWPROFILE} // 'default';
+    if ($name =~ /\/|\.txt$/) {
+        my $path = $name =~ /^\// ? $name : "$vars->{CASEDIR}/$name";
+        bmwqemu::diag("Loading custom QEMU hardware profile from $path");
+        my @commands = map { [$_->[0] =~ s/^-+//r, @$_[1 .. $#$_]] }
+          map { [split ' '] }
+          grep { /\S/ }
+          map { s/#.*//r }
+          split /\n/, path($path)->slurp('UTF-8');
+        $self->{proc}->static_param(@$_) for @commands;
+        $self->{hwprofile} = {provides => []};
+    }
+    else {
+        my $arch = $vars->{ARCH} // '';
+        $self->{hwprofile} = OpenQA::Qemu::HWProfiles::get_profile($name, $arch);
+        die "QEMU hardware profile '$name' not found for architecture '$arch'" unless $self->{hwprofile};
+        bmwqemu::diag("Applying QEMU hardware profile: $name");
+        $self->{proc}->static_param(@$_) for @{$self->{hwprofile}->{args}};
+    }
+}
+
+sub _profile_has ($self, $cap) {
+    return grep { $_ eq $cap } @{$self->{hwprofile}->{provides} // []};
+}
+
 sub start_qemu ($self) {
     my $vars = \%bmwqemu::vars;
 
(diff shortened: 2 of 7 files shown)


OpenQA/Qemu/HWProfiles.pm
```diff
@@ -0,0 +1,44 @@
...
+            ['device', 'virtio-rng-pci,rng=rng0,id=rng0,bus=root_port7,addr=0x0'],
+            ['object', 'rng-random,filename=/dev/urandom,id=rng0'],
+        ],
+        provides => [qw(serial pci usb graphics balloon rng input)],
+    },
+);
+
+sub get_profile ($name, $arch) {
+    my $p = $profiles{$name} or return undef;
+    return {args => ($p->{$arch} // $p->{default} // []), provides => ($p->{provides} // [])};
```
mdoucha, line 31: It'd be nice to have the built-in profiles defined in a YAML/JSON file. Only the `default` profile should be defined in Perl code. Otherwise I like this implementation.

backend/qemu.pm
```diff
@@ -681,6 +682,38 @@ sub determine_qemu_version ($self, $qemubin) {
     $self->{qemu_version} = $qemu_version;
 }
 
+sub apply_hwprofile ($self) {
+    my $vars = \%bmwqemu::vars;
+    my $name = $vars->{QEMU_HWPROFILE} // 'default';
+    if ($name =~ /\/|\.txt$/) {
+        my $path = $name =~ /^\// ? $name : "$vars->{CASEDIR}/$name";
+        die "Profile $path not found" unless -e $path;
```
Martchus: The `slurp` call will handle non-existing files so this check can be removed.

backend/qemu.pm
```diff
@@ -681,6 +682,38 @@ sub determine_qemu_version ($self, $qemubin) {
...
+        bmwqemu::diag("Loading custom QEMU hardware profile from $path");
+        my @commands = map { [$_->[0] =~ s/^-+//r, @$_[1 .. $#$_]] }
+          map { [split ' '] }
+          grep { /\S/ }
+          map { s/#.*//r }
+          split /\n/, path($path)->slurp;
```
Martchus: Considering this is text and not just passed as-is this should probably be decoded:

```suggestion
          split /\n/, path($path)->slurp('UTF-8');
```

Maybe it would make sense to have a test that contains non-ASCII characters to be sure.

doc/backend_vars.md
```diff
@@ -168,6 +168,7 @@ Supported variables per backend
 | QEMU_COMPRESS_QCOW2 | boolean | 1 | compress qcow2 images intended for upload |
 | QEMU_IMG_CREATE_TRIES | integer | 3 | Define number of tries for qemu-img commands |
 | QEMU_HUGE_PAGES_PATH | string | undef | Define a path to use huge pages (e.g. /dev/hugepages/) |
+| QEMU_HWPROFILE | string | undef | Name of a predefined hardware profile (e.g. `virt-manager-defaults`) or path to a file containing QEMU arguments. |
```
Martchus:
```suggestion
| QEMU_HWPROFILE | string | undef | Name of a predefined hardware profile (e.g. `virt-manager-defaults`) or path to a file containing QEMU arguments.|
```

It would be good to know where to find the list of predefined hardware profiles. It would also be good to know that the QEMU arguments read from the file are new-line separated and support comments via `#`.

OpenQA/Qemu/HWProfiles.pm
```diff
@@ -0,0 +1,44 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Qemu::HWProfiles;
+use Mojo::Base -base, -signatures;
```
Martchus: This doesn't seem to use `-base`.

t/30-qemu-hwprofiles.t
```diff
@@ -0,0 +1,106 @@
...
+my $backend_mock = Test::MockModule->new('backend::qemu', no_auto => 1);
+$backend_mock->redefine(handle_qmp_command => undef);
+$backend_mock->redefine(determine_qemu_version => sub ($self, @) { $self->{qemu_version} = '9.0' });
+
+my $distri = Test::MockModule->new('distribution');
+$distri->redefine(add_console => sub {
+        my $ret = Test::MockObject->new();
+        $ret->set_true('backend');
+        return $ret;
+});
```
Martchus:
```suggestion
$distri->redefine(add_console => Test::MockObject->new>set_true('backend'));
```

Martchus (changes requested):
Looks generally good. I only have a few nitpicks.

OpenQA/Qemu/HWProfiles.pm
okurz: done, see https://github.com/okurz/os-autoinst/blob/506699916baf719d8bb17390edc1684013f6a47d/OpenQA/Qemu/hw_profiles.yaml

backend/qemu.pm

backend/qemu.pm
okurz: done both

OpenQA/Qemu/HWProfiles.pm
okurz: done

doc/backend_vars.md
okurz: done.

t/30-qemu-hwprofiles.t
okurz: done

Martchus (approved):

OpenQA/Qemu/HWProfiles.pm
```diff
@@ -0,0 +1,34 @@
...
+        provides => [qw(serial)],
+    },
+);
+
+my $external_profiles;
+
+sub get_profile ($name, $arch) {
+    if (!$profiles{$name}) {
+        $external_profiles //= YAML::PP->new->load_file(path(__FILE__)->sibling('hw_profiles.yaml'));
+        return undef unless $external_profiles->{$name};
```
mdoucha, line 26: Wouldn't it be better to terminate the job if the requested HW profile doesn't exist?

okurz: true. better be explicit. But I would keep the `return undef` in HWProfiles.pm and handle the not found repo in backend/qemu.pm

mdoucha: It's safer to permit missing profiles with `eval { $profile = get_profile($foo, $bar); }`. But whatever.

mdoucha (approved):

PR #2914, opened 2026-04-30
fix: re-install serial marker hook after console reset

okurz:
Motivation:
Pretty serial markers rely on shell hooks (PROMPT_COMMAND) installed in the
SUT. When the OS changes (e.g. after installation), these hooks are lost,
but os-autoinst kept using its cached knowledge of the hook being installed.

Design Choices:
Added reset_console_cache to distribution.pm and hooked it into
testapi::select_console. When a console is re-activated (e.g. after a
reboot or explicit reset), the cached serial marker state is cleared,
forcing re-detection and re-installation of the hook on the next command.

Benefits:
Ensures pretty serial markers work correctly across reboots and OS
transitions without manual intervention in test scripts.

Manual verification:
* https://openqa.suse.de/tests/22089177#step/system_prepare/7

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -15,6 +15,7 @@ sub new ($class, @) {
     $self->{serial_failures} = [];
     $self->{autoinst_failures} = [];
     $self->{_serial_marker_level} = {};
+    $self->{_serial_marker_hook_installed} = {};
 
 =head2 serial_term_prompt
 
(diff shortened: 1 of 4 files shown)


asmorodskyi (approved):
LGTM

distribution.pm
```diff
@@ -452,6 +452,11 @@ sub install_serial_marker_hook ($self, $level) {
     $self->{_serial_marker_hook_installed}->{$console} = 1;
 }
 
+sub reset_console_cache ($self, $console) {
+    delete $self->{_serial_marker_level}->{$console};
```
d3flex, line 457: I am not sure but why not *reset* in the initial value? the initialization is `$self->{_serial_marker_level} = {};` and this likely delete the `_serial_marker_level` from $self, no?

okurz: The delete statement here is intentional and idiomatic Perl: it removes only the cache entry for the specific $console being reset, not the entire hash. The 'initial value' for a console is effectively 'nothing cached,' which delete achieves. 

Regarding the initialization: I've updated the new constructor to explicitly initialize $self->{_serial_marker_hook_installed} = {} so the structure is consistent from the start.

d3flex (approved):

PR #2915, opened 2026-04-30
fix: support pretty_serial_markers in script_output regex

okurz:
Motivation:
When `pretty_serial_markers` is enabled, the `PROMPT_COMMAND` hook prints the
executed command to the serial console before `OA:START`. In `script_output`, a
temporary script is created with the `$marker` embedded in its name, which
causes the `$marker` to be echoed multiple times. The old regex greedily matched
the first occurrence, causing unintended prefixes like `.sh\nOA:START\n` to be
captured in the expected output.

Design Choices:
The regex in `distribution.pm` was updated to require the starting `$marker` to
be on its own line (or at the start of the string) since `echo $marker` prints
exactly the marker followed by a newline. The inner capture group was also made
non-greedy (`.*?`) to prevent capturing unintended trailing content if multiple
`SCRIPT_FINISHED` markers somehow appear.

Benefits:
This change unambiguously identifies the actual execution output boundaries,
fixing `glibc_locale` test failures and ensuring robust output extraction across
the entire test suite when the `pretty_serial_markers` feature is enabled.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -353,7 +353,7 @@ sub script_output ($self, $script, @args) {
     }
 
     # and the markers including internal exit catcher
-    my $out = $output =~ /$marker(?<expected_output>.+)SCRIPT_FINISHED$marker-\d+-/s ? $+ : '';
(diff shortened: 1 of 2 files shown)


d3flex (approved):

asmorodskyi (approved):
LGTM

PR #2916, opened 2026-04-30
feat: Improve message when aborting if there is not enough storage

Martchus:
With the current message one gets e.g. `requested 25 GiB, available 30 GiB, …` which is confusing because `25 < 30`. This commit changes the message to e.g. `requested 25 GiB, available keeping 20 % free: 22 GiB, generally available 30 GiB, …` where `22 < 25` which makes more sense.

Related ticket: https://progress.opensuse.org/issues/199790

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -185,11 +185,12 @@ sub _abort_if_storage_limit_exceeded () {
     my $requested_bytes = $total_hdd_size_gb * GIB;
     my $min_free_bytes = $total_storage * $keep_free;
     my $keep_free_gb = $vars{STORAGE_KEEP_FREE_GB} // STORAGE_KEEP_FREE_GB;
-    my $relative_exceeded = $requested_bytes > $available - $min_free_bytes;
+    my $relative_available = $available - $min_free_bytes;
+    my $relative_exceeded = $requested_bytes > $relative_available;
     my $absolute_exceeded = $keep_free_gb > 0 && $requested_bytes > $available - ($keep_free_gb * GIB);
     return undef unless $relative_exceeded && $absolute_exceeded;
-    my $msg = sprintf 'Not enough storage for requested HDDSIZEGB (requested %d GiB, available %d GiB, total %d GiB, keep-free %d%%)',
-      $total_hdd_size_gb, int($available / GIB), int($total_storage / GIB), int($keep_free * 100);
+    my $msg = sprintf 'Not enough storage for requested HDDSIZEGB (requested %d GiB, available keeping %d%% free: %d GiB, generally available %d GiB, total %d GiB)',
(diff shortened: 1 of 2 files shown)


okurz (approved):

bmwqemu.pm
```diff
@@ -185,11 +185,12 @@ sub _abort_if_storage_limit_exceeded () {
...
     my $keep_free_gb = $vars{STORAGE_KEEP_FREE_GB} // STORAGE_KEEP_FREE_GB;
-    my $relative_exceeded = $requested_bytes > $available - $min_free_bytes;
+    my $relative_available = $available - $min_free_bytes;
+    my $relative_exceeded = $requested_bytes > $relative_available;
     my $absolute_exceeded = $keep_free_gb > 0 && $requested_bytes > $available - ($keep_free_gb * GIB);
     return undef unless $relative_exceeded && $absolute_exceeded;
-    my $msg = sprintf 'Not enough storage for requested HDDSIZEGB (requested %d GiB, available %d GiB, total %d GiB, keep-free %d%%)',
-      $total_hdd_size_gb, int($available / GIB), int($total_storage / GIB), int($keep_free * 100);
+    my $msg = sprintf 'Not enough storage for requested HDDSIZEGB (requested %d GiB, available keeping %d%% free: %d GiB, generally available %d GiB, total %d GiB)',
+      $total_hdd_size_gb, int($keep_free * 100), $relative_available > 0 ? int($relative_available / GIB) : 0, int($available / GIB), int($total_storage / GIB);
```
perlpunk, line 193: It could probably still be confusing in some cases, because `STORAGE_KEEP_FREE_GB` isn't mentioned at all in the error message.

Martchus: The error message is already quite long. I'm not sure whether it makes sense to mention `STORAGE_KEEP_FREE_GB` which would make it even longer.

t/12-bmwqemu.t
```diff
@@ -191,7 +191,11 @@ subtest 'abort on low disk space' => sub {
         {hdd_size_gb => 10, total_storage => 100, avail_storage => 100, expect => 'pass',
             desc => 'succeed if requested HDDSIZEGB is well within available space'},
         {hdd_size_gb => 60, total_storage => 5, avail_storage => 5, expect => 'fail', extra_vars => {STORAGE_KEEP_FREE_RATIO => 0.9},
-            desc => 'abort if requested HDDSIZEGB exceeds custom threshold'},
+            desc => 'abort if requested HDDSIZEGB exceeds available storage (keeping certain ratio free) which is already negative anyway',
+            expected_stats => 'requested 60 GiB, available keeping 90% free: 0 GiB, generally available 5 GiB, total 5 GiB'},
```
d3flex, line 195: here it could be the variables in place of the hardcoded values.

Martchus: That's a test so I'd prefer to keep it simple. The way I put it makes it perfectly clear what is checked.

perlpunk (approved):

PR #2917, opened 2026-04-30
fix: echo json_cmd_token in isotovideo responses

okurz:
Motivation:
Recent changes in autotest.pm (commit d9a2b23c) introduced token checking
for the tests_done command to avoid a race condition during shutdown.
However, CommandHandler.pm was not echoing the token back, causing
myjsonrpc.pm to throw an error because the received token did not match the
sent token.

Design Choices:
- Centralized token injection using a new _add_token helper method. This
  ensures zero code duplication across handlers and backend calls.
- Adopted request-scoped token management using 'local' in process_command.
  This avoids persistent object-level state for synchronous calls, reducing the
  risk of race conditions.
- Added explicit storage for tokens associated with deferred responses
  (backend_requester_token, postponed_token). This ensures the correct token
  is restored even when responses are delayed or interleaved.
- Implemented _add_token in a functional style, returning a shallow copy of the
  payload instead of mutating in-place, avoiding side-effects.
- Added an acknowledgement response (_respond_ok) to the resume_test_execution
  command to ensure the requester receives a proper protocol answer.

Benefits:
- Fixes the protocol mismatch error and prevents hangs during test shutdown.
- Robust handling of tokens across synchronous, deferred, and interleaved
  commands.
- Improved maintainability through centralized logic and safe state management.
- Verified with new tests covering auto-injection and interleaved commands.

Related issue: https://openqa.suse.de/tests/22078675

### DIFF
--- a/OpenQA/Isotovideo/CommandHandler.pm
+++ b/OpenQA/Isotovideo/CommandHandler.pm
@@ -17,6 +17,7 @@ use constant AUTOINST_STATUSFILE => 'autoinst-status.json';
 
 # io handles for sending data to command server and backend
 has [qw(test_fd cmd_srv_fd backend_fd backend_out_fd answer_fd)] => undef;
+has [qw(current_command_token backend_requester_token postponed_token)];
 
 # the name of the current test (full name includes category prefix, eg. installation-)
 has [qw(current_test_name current_test_full_name)];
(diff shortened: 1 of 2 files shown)


Martchus (approved):

OpenQA/Isotovideo/CommandHandler.pm
```diff
@@ -285,13 +307,13 @@ sub _handle_command_set_current_test ($self, $response, @) {
...
     $self->update_status_file;
-    $self->_respond_ok_or_postpone_if_paused;
+    $self->_respond_ok_or_postpone_if_paused();
 }
 
 sub _handle_command_tests_done ($self, $response, @) {
     $self->test_died($response->{died});
     $self->test_completed($response->{completed});
-    $self->_respond_ok;
+    $self->_respond_ok();
```
kalikiana, line 316: This was introduced in https://github.com/os-autoinst/os-autoinst/commit/d9a2b23c1a162797ad5d31dcbbd1bfd36f2d5fdf not here

Clarifying for myself because of unrelated style changes.

kalikiana (approved):
I guess we can't cover `questions and answers not in the right order` explicitly. Given that the tests make sense.

PR #2919, opened 2026-04-30
fix: restore serial marker hook after switching users

okurz:
Motivation
When a test switches users on the same GUI console (e.g., using `su -`
via `become_root`), the new shell environment no longer has the serial
marker hook (PROMPT_COMMAND) installed if it was only installed for the
previous user. This causes subsequent `script_run` calls to time out
waiting for the `OA:DONE` marker.

Design Choices
We add an `invalidate_serial_marker_hook` method to `distribution.pm`
and call it at the end of `become_root` within `distribution.pm`. This
forces the backend to reinstall the hook on the next `script_run` for the
new user, ensuring the synchronization markers are properly emitted.

As part of this change, we also refactored the marker reset API for
clarity:
- Renamed `reset_console_cache` to `reset_serial_marker` (full reset of
  capabilities and hook).
- Introduced `invalidate_serial_marker_hook` (invalidation of the hook
  while preserving detected capabilities).

Benefits
Ensures reliable command synchronization after user switching while
avoiding redundant re-detection overhead in `script_sudo`. The codebase
maintainability is improved through clearer naming and detailed POD
documentation of the marker state management.

Manual verification:
https://openqa.opensuse.org/tests/5890577

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -85,6 +85,7 @@ sub become_root ($self) {
     testapi::enter_cmd('test $(id -u) -eq 0 && echo "imroot" > /dev/' . $testapi::serialdev, 0);
     testapi::wait_serial('imroot') || die 'Root prompt not there';
     testapi::enter_cmd('cd /tmp');
+    $self->invalidate_serial_marker_hook();
 }
 
 =head 2 disable_key_repeat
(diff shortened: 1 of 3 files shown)


distribution.pm
```diff
@@ -453,8 +454,42 @@ sub install_serial_marker_hook ($self, $level) {
...
+level and the hook installation status, forcing a complete re-detection and
+re-installation on the next command.
+
+Useful when the console environment has fundamentally changed, such as after
+a reboot or when switching to a different OS/shell.
+
+=cut
+
+sub reset_serial_marker ($self, $console = undef) {
+    $console //= testapi::current_console() // 'sut';
```
d3flex: I would like to understand whether the `// 'sut'` is needed. when the no console is given and if not `$autotest::selected_console` why should set console in 'sut'?

d3flex (changes requested):
Some concerns about the fallback to 'sut' with the `reset_serial_marker` and `invalidate_serial_marker_hook`. What happens if this removed and the method is called? i guess that when they are called and console is undef during the test, this would be unexpected behavior and it should fail or something.

distribution.pm
okurz: We can leave out this fallback. I reworked this and added an explicit return if no valid console was passed

kalikiana (approved):

distribution.pm
kalikiana: I prefer the explicit return as well.

d3flex (approved):

distribution.pm
okurz: ok, this turned out to be a bad idea because that means that pretty serial markers *only* work when tests explicitly select consoles which does not necessarily happen. Will need to work on this: https://github.com/os-autoinst/os-autoinst/pull/2981

PR #2920, opened 2026-05-08
fix: ensure serial buffer is cleared during backend_clear_serial_buffer

okurz:
Motivation:
When using PRETTY_SERIAL_MARKER=1, tests rely on the OA:DONE markers for
synchronization. However, after a VM resumes from a snapshot, unconsumed
serial data from the previous state might be flushed into the buffer.
The existing backend_clear_serial_buffer command only updated the file
offset but did not clear the live carry_buffer in the serial_screen
object. This might have led to false positive matches on stale markers,
especially causing first-after-snapshot tests like zypper_lr to fail.

Design Choices:
1. Add a clear_buffer method to consoles::serial_screen to drain the
   non-blocking socket and empty the internal carry_buffer.
2. Update backend::baseclass::clear_serial_buffer to invoke the new
   clear_buffer method on the active console. This ensures that the live
   interactive state is fully reset alongside the file offset.

Benefits:
Fixes a race condition during snapshot restores when using advanced
serial markers. Tests are now more robust against stale data lingering
in the serial communication pipe, ensuring reliable command-to-result
synchronization.

Manual verification:

```
openqa-clone-job --skip-chained-deps --within-instance https://openqa.opensuse.org/t5916451 _GROUP=0 {TEST,BUILD}+=-okurz-pretty PRETTY_SERIAL_MARKER=1 WORKER_CLASS=openqaworker24
```

[opensuse-Tumbleweed-DVD-x86_64-Build20260510-gnome@64bit](https://openqa.opensuse.org/tests/5917783/badge)

### DIFF
--- a/consoles/serial_screen.pm
+++ b/consoles/serial_screen.pm
@@ -293,6 +293,20 @@ sub peak ($self, %nargs) {
     return $self->{carry_buffer};
 }
 
+=head2 clear_buffer
+
+Drain any pending data from the non-blocking socket and empty the internal carry buffer.
+
+=cut
+
+sub clear_buffer ($self) {
+    my $buf;
+    while (sysread $self->{fd_read}, $buf, SOCKET_READ_BUFFER_SIZE) {
+        # Drain the non-blocking socket
+    }
+    $self->{carry_buffer} = '';
+}
+
 sub current_screen ($self) { 0 }
 
 sub request_screen_update ($self, @) { }
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -843,6 +843,9 @@ previous test's serial output. Call this before you start doing something new
 =cut
 
 sub clear_serial_buffer ($self, @) {
+    if (my $screen = $self->{current_screen}) {
+        $screen->clear_buffer() if $screen->can('clear_buffer');
+    }
     $self->{serial_offset} = -s $self->{serialfile};
     return $self->{serial_offset};
 }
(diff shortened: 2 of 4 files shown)


Martchus (approved):

Martchus:
Of course tests are missing but the code/description kind of makes sense.

okurz:
> Of course tests are missing but the code/description kind of makes sense.

tests added, completed coverage

consoles/serial_screen.pm
```diff
@@ -293,6 +293,20 @@ sub peak ($self, %nargs) {
...
 
+=head2 clear_buffer
+
+Drain any pending data from the non-blocking socket and empty the internal carry buffer.
+
+=cut
+
+sub clear_buffer ($self) {
+    my $buf;
+    while (sysread $self->{fd_read}, $buf, 4096) {
```
d3flex: there is a SOCKET_READ_BUFFER_SIZE?!

okurz:
implemented @d3flex one suggestion

okurz:
@Martchus I think we need to adjust the "openQA Fullstack" test in os-autoinst now after your changes. Please see https://github.com/os-autoinst/os-autoinst/actions/runs/25764813313/job/75674694256?pr=2920#step:6:1

Martchus:
Ok, I'll have a look and will come up with a PR.
EDIT: PR: https://github.com/os-autoinst/os-autoinst/pull/2925

okurz:
> Ok, I'll have a look and will come up with a PR. EDIT: PR: #2925

thx. It's merged. rebased my PR on top.

d3flex (approved):

PR #2923, opened 2026-05-11
feat: add optimized helpers for per-module PRETTY_SERIAL_MARKER control

okurz:
Motivation:
Some test modules or specific code sections are known to be incompatible
with advanced serial markers. Testers need a convenient way to disable
pretty markers for these specific scopes and have them automatically
restored afterwards.

Design Choices:
1. Add set_pretty_serial_marker($value) to distribution.pm. This method
   updates the test variable and immediately resets the cached marker
   state. It also types "unset PROMPT_COMMAND" in the SUT when disabling
   to avoid serial buffer pollution.
2. Add pretty_serial_marker_guard($value) to distribution.pm. This
   returns a RAII scope_guard that restores the previous value.
3. Track hook persistence in the SUT config files via
   _serial_marker_hook_persistent to avoid redundant typing during
   state transitions.

Benefits:
Provides test authors with a robust, clean, and automatic way to handle
problematic scenarios with pretty serial markers. Using the RAII guard
prevents accidental state leaks, and the optimizations ensure minimal
log clutter and maximum reliability.

Before:
* #2928

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -4,6 +4,7 @@
 
 package distribution;
 use Mojo::Base -strict, -signatures;
+use Mojo::Util 'scope_guard';
 
 use testapi ();
 use log 'fctwarn';
@@ -16,6 +17,8 @@ sub new ($class, @) {
     $self->{autoinst_failures} = [];
     $self->{_serial_marker_level} = {};
     $self->{_serial_marker_hook_installed} = {};
+    $self->{_serial_marker_hook_persistent} = {};
+    $self->{_pretty_serial_marker} = undef;
 
 =head2 serial_term_prompt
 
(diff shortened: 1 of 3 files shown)


distribution.pm
```diff
@@ -496,6 +509,53 @@ sub invalidate_serial_marker_hook ($self, $console = undef) {
...
+
+    set_pretty_serial_marker($value)
+
+Set C<PRETTY_SERIAL_MARKER> to C<$value> and reset the cached marker state.
+
+=cut
+
+sub set_pretty_serial_marker ($self, $value) {
+    my $old_value = testapi::get_var('PRETTY_SERIAL_MARKER');
+    return if defined $old_value && $old_value == $value;
```
Martchus: Those markers aren't always just numbers, right?
```suggestion
    return if defined $old_value && $old_value eq $value;
```

Martchus (approved):

distribution.pm
```diff
@@ -496,6 +502,53 @@ sub invalidate_serial_marker_hook ($self, $console = undef) {
...
+Set C<PRETTY_SERIAL_MARKER> to C<$value> and reset the cached marker state.
+
+=cut
+
+sub set_pretty_serial_marker ($self, $value) {
+    my $old_value = testapi::get_var('PRETTY_SERIAL_MARKER');
+    return if defined $old_value && $old_value eq $value;
+
+    bmwqemu::log_call(value => $value);
+    testapi::set_var('PRETTY_SERIAL_MARKER', $value);
```
perlpunk: As shortly discussed in a meeting a few minutes ago:
This `PRETTY_SERIAL_MARKER` is documented as a job setting, to set the initial value.
I'm questioning whether we should save the current state of it in the vars hash, though.

Of course it's kind of convenient, and as it is initially set there, it seems the obvious thing to do.

The value that it has at the end will end up in `vars.json`, not the initial one.

I'm wondering if we should use a lexical variable `my $PRETTY_SERIAL_MARKER` in testapi.pm or distribution.pm to hold the current value instead, and not use the vars hash, just because it's convenient to store something there which is is needed globally.

Also anything that can be seen and set via `get/set_var` might be an invitation to be used / changed directly instead of the method `set_pretty_serial_marker` here.

As we have this method already, the implementation on our side should not matter for the test code.

perlpunk (approved):

PR #2924, opened 2026-05-12
fix: uniquely namespace test modules and optimize scheduling

okurz:
Motivation:
When two test modules have the same filename but are in different folders, they
overwrite each other because they were loaded into the same Perl package name
(derived from the filename). This led to confusing test results, "Subroutine
run redefined" warnings, and incorrect test execution.

Design Choices:
- Derived unique Perl package names (classes) by incorporating the category
  (path) into the package name (e.g., tests::jeos::image_checks).
- Enhanced Python test loading using importlib.util to explicitly manage
  sys.modules caching, ensuring independent namespaces for Python scripts with
  colliding filenames. Proper escaping added to prevent code injection.
- Refactored _should_schedule to use a functional style with a dedicated
  _test_identifiers helper, supporting include/exclude rules against base
  names, full classes, or fullnames.
- Optimized performance by caching EXCLUDE_MODULES and INCLUDE_MODULES parsing
  using state variables and optimizing the EXIT_AFTER_MODULE lookup loop.
- Refined naming logic to functionally define properties while preserving
  uniqueness for duplicate modules (#nr suffix).
- Integrated comprehensive reproduction test cases for both Perl and Python
  module collisions in t/08-autotest.t.

Benefits:
- Prevents silent module overwriting and hard-to-debug "redefined" warnings.
- Improves overall test reliability and engine robustness.
- Enhances maintainability through cleaner, functional code patterns without
  unnecessary duplication.
- Gains minor performance improvements in large test suites via caching.
- Maintains full backward compatibility with existing openQA configurations.

Related issue: https://progress.opensuse.org/issues/200886

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -179,9 +179,11 @@ sub _load_lua () {
     lua_set('use', \&_lua_use);
 }
 
-sub _make_test_code_to_eval ($script_path, $script, $name, $is_python) {
+sub _esc_py ($str) { $str =~ s/(["\\])/\\$1/gr }
+
+sub _make_test_code_to_eval ($script_path, $script, $package, $is_python) {
     my $casedir = $bmwqemu::vars{CASEDIR};
-    my $code = "package $name;";
+    my $code = "package $package;";
     my $module_code;
     if ($bmwqemu::vars{ENABLE_MODERN_PERL_FEATURES}) {
         $code .= 'use Mojo::Base -strict, -signatures;';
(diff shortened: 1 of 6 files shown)


Martchus:
Generally a good idea. This is definitely not a new problem as I remember that I also tried to fix that in the past. I'm wondering what happened to my previous PR. Maybe I just got stuck on failing tests - or did something speak against that?

kalikiana:
> Generally a good idea. This is definitely not a new problem as I remember that I also tried to fix that in the past. I'm wondering what happened to my previous PR. Maybe I just got stuck on failing tests - or did something speak against that?

@Martchus What is "that"? Should we avoid going with the approach proposed here?

okurz:
If we would merge this then that would allow that different test modules with the same basename would overwrite results and screenshots. We should obviously not do that hence we should focus on only explicit error feedback.

PR #2925, opened 2026-05-13
ci: Fix fullstack test after removal of `tools/ci/build_cache.sh`

Martchus:
* Switch to using devel container which is also used in openQA now
* Streamline user setup with what we do now in openQA
* Avoid `sudo npm install`; use `Makefile` target instead to benefit from recent improvements

Related ticket: https://progress.opensuse.org/issues/200609

### DIFF
--- a/.github/workflows/openqa_fullstack.yml
+++ b/.github/workflows/openqa_fullstack.yml
@@ -12,9 +12,7 @@ on:
 env:
   # Set to 1 to temporarily ignore warnings
   PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS: 0
-  # Default user in the used container image from
-  # https://github.com/os-autoinst/openQA/blob/master/container/devel%3AopenQA%3Aci/base/Dockerfile
-  NORMAL_USER: squamata
+  TEST_USER: testuser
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

.github/workflows/openqa_fullstack.yml
```diff
@@ -12,32 +12,28 @@ on:
...
-      # See https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user
-      options: --user root
+      image: registry.opensuse.org/devel/openqa/containers/opensuse/openqa_devel:latest
     steps:
       - uses: actions/checkout@v6
+      - name: Create additional user to avoid initdb error "cannot be run as root"
+        run: useradd "$TEST_USER"
       - name: Make project folder user-writable
-        run: chown -R $NORMAL_USER ../../os-autoinst
+        run: chown -R $TEST_USER ../../os-autoinst
```
d3flex, line 29: I wonder if this can be a ENV variable. but I guess it doesnt need to change taking that it was like that before

d3flex (approved):

PR #2926, opened 2026-05-15
Revert "fix: ensure serial buffer is cleared during backend_clear_serial_buffer"

okurz:
Reverts os-autoinst/os-autoinst#2920
as https://openqa.opensuse.org/tests/5928392
opensuse-Tumbleweed-JeOS-for-kvm-and-xen-x86_64-jeos-ltp-containers@uefi_virtio-2G from
https://openqa.opensuse.org/tests/latest?arch=x86_64&distri=opensuse&flavor=JeOS-for-kvm-and-xen&machine=uefi_virtio-2G&test=jeos-ltp-containers&version=Tumbleweed
show a problem.

the issue is seen at https://openqa.opensuse.org/tests/5928392#step/install_ltp/1
isotovideo waits for `#` to identift the prompt on the serial console which the buffer clearing probably already consumed.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -843,9 +843,6 @@ previous test's serial output. Call this before you start doing something new
 =cut
 
 sub clear_serial_buffer ($self, @) {
-    if (my $screen = $self->{current_screen}) {
-        $screen->clear_buffer() if $screen->can('clear_buffer');
-    }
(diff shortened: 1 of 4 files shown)


pevik:
IMHO reason to revert is worth to write into the revert commit itself (for offline reading + not loosing the history on fork or migrating from github in the future).

PR #2928, opened 2026-05-17
perf: use sourcing for efficient pretty serial marker activation

okurz:
Motivation:
Previously, the PRETTY_SERIAL_MARKER hook was both appended to config
files and then manually re-typed in the live shell. This still incurred
significant VNC typing overhead despite previous optimizations.

Design Choices:
Updated install_serial_marker_hook to use shell sourcing (. ~/.bashrc)
to activate the hook in the current session after appending it to the
user's profile. This removes the need to re-type the 150+ character
logic entirely.

Benefits:
Drastically reduces VNC keystrokes, improving reliability on slow
connections and speeding up test initialization.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -156,13 +156,13 @@ sub script_run ($self, $cmd, @args) {
           if $cmd =~ qr/(?<!\\)&$/;
 
         my $level = $self->_detect_serial_marker_capability();
-        my $guard;
+        my $skip_pretty = 0;
         # Automatically protect against manual serial redirections corrupting pretty markers
         if ($level > 1 && $cmd =~ m{(?:>|>>|\btee)\s+(?:-a\s+)?/dev/\Q$testapi::serialdev\E\b}) {
             bmwqemu::diag('Temporarily disabling PRETTY_SERIAL_MARKER to prevent corruption with serial terminal redirection');
             bmwqemu::diag("Manual redirection to /dev/$testapi::serialdev is deprecated and might conflict with advanced serial markers. Use script_output() or use script_run() without the quiet parameter instead.");
-            $guard = $self->pretty_serial_marker_guard(0);
-            $level = $self->_detect_serial_marker_capability();
+            $level = 1;
+            $skip_pretty = 1;
         }
         my ($str, $wait_pattern);
         if ($level == 3) {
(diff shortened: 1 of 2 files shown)


Martchus:
Did you make at least one verification run?

distribution.pm
```diff
@@ -448,24 +450,23 @@ markers to serial.
...
     else {
-        $pc = "PROMPT_COMMAND='if [ -n \"\$__OA_MARK\" ]; then echo \"\${__OA_MARK}-\$?-\" > $dev; unset __OA_MARK; fi; echo \"OA:START\" > $dev'";
+        $func = "__oa_prompt() { r=\$?; if [ -n \"\$OA_NO_MARKER\" ]; then unset OA_NO_MARKER; elif [ -n \"\$__OA_MARK\" ]; then echo \"\${__OA_MARK}-\$r-\" > $dev; unset __OA_MARK; fi; echo \"OA:START\" > $dev; }";
     }
-    testapi::type_string "$pc\n";
+    my $pc = 'PROMPT_COMMAND=__oa_prompt';
+
+    # Consolidate installation and persistence into a single typed line to minimize VNC overhead.
+    # Sourcing ~/.bashrc activates the hook in the current shell without re-typing the entire logic.
+    testapi::type_string "grep -q __oa_prompt ~/.bashrc 2>/dev/null || { echo '$func; $pc' | tee -a ~/.bashrc >> ~/.profile; }; . ~/.bashrc\n";
```
Martchus: This syntax is a bit hard to read. Is this adding the line also to `~/.profile` to cover other shells than Bash? Maybe the point of modifying `~/.profile` should be mentioned in the comment. Maybe it is sufficient to modify only `~/.profile`.

distribution.pm
```diff
@@ -448,24 +450,23 @@ markers to serial.
 
 sub install_serial_marker_hook ($self, $level) {
     return if $level < 2;
-    my $pc;
     my $dev = "/dev/$testapi::serialdev";
+    my $func;
     if ($level == 3) {
-        $pc = "PROMPT_COMMAND='ret=\$?; cmd=\$(fc -ln -1 2>/dev/null); printf \"OA:DONE-%04x-%d-%s\\nOA:START\\n\" \$RANDOM \$ret \"\${cmd#\${cmd%%[![:space:]]*}}\" > $dev'";
+        $func = "__oa_prompt() { r=\$?; if [ -n \"\$OA_NO_MARKER\" ]; then unset OA_NO_MARKER; else c=\$(fc -ln -1 2>/dev/null); printf \"OA:DONE-%04x-%d-%s\\nOA:START\\n\" \$RANDOM \$r \"\${c#\${c%%[![:space:]]*}}\" > $dev; fi; }";
```
Martchus: It would maybe make sense to use different quotes on Perl-level to avoid the escaping.

Martchus (commented):
It looks generally good but the syntax alone makes this a non-trivial change. So a verification run would be good.

okurz:
> Did you make at least one verification run?

Actually 315 :)

* [opensuse-Tumbleweed-DVD-x86_64-Build20260517-extra_tests_on_kde@uefi](https://openqa.opensuse.org/tests/5939377)
* [20260517-okurz-pretty](https://openqa.opensuse.org/tests/overview?distri=opensuse&build=20260517-okurz-pretty&version=Tumbleweed&result=passed)

I will incorporate your suggestions and test again the latest state

* [0260520-okurz-pretty-w25-pr2928](https://openqa.opensuse.org/tests/overview?build=20260520-okurz-pretty-w25-pr2928&version=Tumbleweed&distri=opensuse&result=passed)

still looks good

distribution.pm
okurz: ok. trying to improve the syntax I learned something, tee can append to multiple files. So we can just do `tee -a ~/.bashrc ~/.profile >/dev/null`. Why both? To cover both login and interactive shells.

distribution.pm

Wabri (approved):

distribution.pm
```diff
@@ -449,23 +449,24 @@ markers to serial.
 
 sub install_serial_marker_hook ($self, $level) {
     return if $level < 2;
-    my $pc;
     my $dev = "/dev/$testapi::serialdev";
+    my $func;
     if ($level == 3) {
-        $pc = "PROMPT_COMMAND='ret=\$?; cmd=\$(fc -ln -1 2>/dev/null); printf \"OA:DONE-%04x-%d-%s\\nOA:START\\n\" \$RANDOM \$ret \"\${cmd#\${cmd%%[![:space:]]*}}\" > $dev'";
+        $func = "__oa_prompt() { ret=\$?; cmd=\$(fc -ln -1 2>/dev/null); printf \"OA:DONE-%04x-%d-%s\\nOA:START\\n\" \$RANDOM \$ret \"\${cmd#\${cmd%%[![:space:]]*}}\" > $dev; }";
```
nicksinger: btw; what does `oa` stand for?

nicksinger (approved):

distribution.pm
Martchus: probably for **o**s-**a**utoinst

PR #2931, opened 2026-05-20
chore(AGENTS.md): point to tools/tidy directly

okurz:


AGENTS.md
```diff
@@ -12,8 +12,9 @@ test scripts.
 
 ## Conventions
 
-- Code style: Run `tools/tidyall --all` (or `tools/tidyall --git` for changed
-  files only).
+- Code style: Run `tools/tidy` (or `tools/tidyall --all`, or `tools/tidyall --git`
```
perlpunk, line 15: I think we want to get rid of `tools/tidy` at some point.
I think it's still there because os-autoinst-distri-opensuse is still using this.
What's your reason to include it here?

Martchus (approved):

perlpunk (changes requested):
See my comment above.
As we want to get rid of tools/tidy we shouldn't add it here.

okurz:
> See my comment above. As we want to get rid of tools/tidy we shouldn't add it here.

I don't understand. As soon as we remove that tool, when that ever happens, then we can also synchronously adjust the agents file

perlpunk:
> > See my comment above. As we want to get rid of tools/tidy we shouldn't add it here.
> 
> I don't understand. As soon as we remove that tool, when that ever happens, then we can also synchronously adjust the agents file

But you still haven't answered my question. Why do you want to mention it at all?
It was replaced by tools/tidyall.

AGENTS.md
okurz: > I think we want to get rid of `tools/tidy` at some point.

ok, maybe you are referring to https://progress.opensuse.org/issues/162095 . Well, yes, we can do that.

 I think it's still there because os-autoinst-distri-opensuse is still using this. What's your reason to include it here?

Mostly because `tools/tidy` by default not needing any arguments "does the right thing". In openQA we have a make target. Maybe we can convince @Martchus to have that in os-autoinst as well :)

okurz:
> But you still haven't answered my question. Why do you want to mention it at all?

Answered in https://github.com/os-autoinst/os-autoinst/pull/2931#discussion_r3334960087

> It was replaced by tools/tidyall.

Not true. tools/tidy is still there and merely calls tidyall internally. But I will look into this.

AGENTS.md
Martchus: We already have `tidy-cpp` and `tidy-python` on CMake-level so you could add `tidy-perl` there as well.

okurz: right -> https://github.com/os-autoinst/os-autoinst/pull/2950

PR #2932, opened 2026-05-20
fix: fail explicitly on test module basename collisions

okurz:
Motivation:
When two test modules have the same filename but are in different folders, they
would overwrite each other's results and screenshots in openQA if allowed to
co-exist. To prevent confusing test results and incorrect artifact association,
we should fail explicitly during scheduling if such a collision is detected.

Design Choices:
- Introduced a tracking mechanism for scheduled basenames in autotest.pm.
- In loadtest, verify that any scheduled basename originates from the same
  file path. If the same basename is loaded from a different directory, the
  test execution now aborts with an explicit error message.
- Updated test suite to expect and verify the explicit failure on basename
  collisions instead of checking for co-existence.

Benefits:
- Provides clear and immediate feedback to the user when a naming collision
  occurs, preventing silent artifact overwrites.
- Enhances engine robustness by ensuring basename uniqueness across the
  schedule.

Related issue: https://progress.opensuse.org/issues/200886

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -198,6 +198,7 @@ subtest 'test always_rollback flag' => sub {
     snapshot_subtest 'always_run flag ensures execution after fatal failure' => sub {
         local %autotest::tests = ();
         local @autotest::testorder = ();
+        local %autotest::scheduled_basenames = ();
         local $autotest::last_milestone = undef;
         local $autotest::last_milestone_active_consoles = [];
         local $autotest::activated_consoles = [];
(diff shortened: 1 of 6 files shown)


t/08-autotest.t
```diff
@@ -567,6 +574,32 @@ subtest 'test skipping tests' => sub {
     stderr_like { autotest::runalltests } qr/skipping/, 'Skipping Test Run';
 };
 
+subtest 'modules with same filename' => sub {
+    local %autotest::tests = ();
+    local @autotest::testorder = ();
+    local %autotest::scheduled_basenames = ();
+    local $bmwqemu::vars{CASEDIR} = "$Bin/data/collision";
+    stderr_like { autotest::loadtest('tests/dir1/collision.pm') } qr/scheduling collision tests\/dir1\/collision\.pm/, 'dir1/collision.pm scheduled';
```
perlpunk:
```suggestion
    stderr_like { autotest::loadtest('tests/dir1/collision.pm') } qr{scheduling collision tests/dir1/collision\.pm}, 'dir1/collision.pm scheduled';
```
same for the tests below

Martchus (approved):

t/08-autotest.t
okurz: done

perlpunk (approved):

PR #2934, opened 2026-05-20
ci: Remove package builds for Leap 15.6 as it reached EOL

Martchus:
Related ticket: https://progress.opensuse.org/issues/200814

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -20,11 +20,6 @@ pr:
               - target_project: openSUSE:Factory
                 target_repository: snapshot
             architectures: [ x86_64 ]
-          - name: openSUSE_Leap_15.6
-            paths:
-              - target_project: devel:openQA:Leap:15.6
-                target_repository: openSUSE_Leap_15.6
-            architectures: [ x86_64 ]
           - name: '16.0'
             paths:
               - target_project: devel:openQA:Leap:16.0
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -9,9 +9,9 @@ pr:
     - configure_repositories:
         project: devel:openQA:GitHub
         repositories:
-          - name: SLE_15_SP6_Backports
+          - name: SLE_15_SP7_Backports
             paths:
-              - target_project: openSUSE:Backports:SLE-15-SP6:Update
+              - target_project: openSUSE:Backports:SLE-15-SP7:Update
                 target_repository: standard
             architectures:
               - x86_64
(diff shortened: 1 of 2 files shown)


.obs/workflows.yml
okurz: likely also this is EOL but should be replaced by 15-SP7

okurz (approved):

.obs/workflows.yml
```diff
@@ -9,9 +9,9 @@ pr:
     - configure_repositories:
         project: devel:openQA:GitHub
         repositories:
-          - name: SLE_15_SP6_Backports
+          - name: SLE_15_SP7_Backports
```
perlpunk, line 12: Will this work?
In https://build.opensuse.org/package/show/devel:openQA/os-autoinst I can see that 15.7 is unresolvable:
> nothing provides perl(Inline::Lua)

Martchus: I'm wondering why this worked for SP6 then. Same as for openQA with the MCP module.

I changed the repo path from `openSUSE:Backports:SLE-15-SP7/standard` to `openSUSE:Backports:SLE-15-SP7:Update/standard` in accordance to what we currently use in repos like https://build.opensuse.org/repositories/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2934 for SP6. Maybe that helps.

Martchus: That helped with the openQA package but not with os-autoinst. (Although the openQA package is now blocked again.)

Martchus: The problem with `perl(Inline::Lua)` might be fixed by the commit I've just pushed.

okurz: You can enable 15.7 manually in the meta of https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2934 to check. Also consider to put the fix into a separate PR first

okurz (approved):

.obs/workflows.yml
perlpunk: I don't see this resolved but as @okurz approved I guess I'm missing something that you discussed elsewhere, so I'll also just approve and we'll see

perlpunk (approved):

.obs/workflows.yml
okurz: I see that https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2934 has both SP6+SP7 enabled and the results are the same so I guess it's ok

PR #2936, opened 2026-05-22
Require perl(Syntax::Keyword::Try) for tests

AdamWill:
60ef460ff1cd0a1da7021a7a89115ce8d48e30c2 added a use of Syntax::Keyword::Try in t/48-testmodules-style.t , but did not add it to the test requirements.

### DIFF
--- a/container/os-autoinst_dev/Dockerfile
+++ b/container/os-autoinst_dev/Dockerfile
@@ -141,6 +141,7 @@ RUN zypper in -y -C \
        'perl(Scalar::Util)' \
        'perl(Socket)' \
        'perl(Socket::MsgHdr)' \
+       'perl(Syntax::Keyword::Try)' \
        'perl(Template::Toolkit)' \
        'perl(Term::ANSIColor)' \
        'perl(Test::CheckGitStatus)' \
--- a/cpanfile
+++ b/cpanfile
@@ -85,6 +85,7 @@ on 'test' => sub {
     requires 'Perl::Critic::Policy';
     requires 'Perl::Critic::Utils';
     requires 'Pod::Coverage';
+    requires 'Syntax::Keyword::Try';
     requires 'Test::Compile';
     requires 'Test::Mock::Time';
     requires 'Test::MockModule';
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -149,6 +149,7 @@ test_base_requires:
   perl(Devel::Cover):
   perl(FindBin):
   perl(Pod::Coverage):
+  perl(Syntax::Keyword::Try):
   perl(Test::MockModule):
   perl(Test::MockObject):
   perl(Test::MockRandom):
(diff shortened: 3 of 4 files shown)


AdamWill:
@perlpunk FYI

perlpunk:
Hm, this is weird.

When I added that, I actually had confused `Syntax::Keyword::Try` with `Feature::Compat::Try`. (The latter is simply a convenient module around `Syntax::Keyword::Try` that will use the core feature automatically in perl 5.40, so that's what we want).

But:
* `Syntax::Keyword::Try`  is a dependency of `Feature::Compat::Try`
* `common_requires` lists `Feature::Compat::Try`
* `test_requires` lists `%common_requires`

So I'm wondering why this could lead to a problem for you...

I will replace the usage of `Syntax::Keyword::Try` with `Feature::Compat::Try` nevertheless...

perlpunk:
* https://github.com/os-autoinst/os-autoinst/pull/2937
* https://github.com/os-autoinst/openQA/pull/7472

So I assume this PR can be closed?

Martchus:
I'd also say we can close this in favor of your other PRs.

AdamWill:
> So I'm wondering why this could lead to a problem for you...

In Fedora, perl-Feature-Compat-Try [does not require perl(Syntax::Keyword::Try)](https://koji.fedoraproject.org/koji/rpminfo?rpmID=45727010). Possibly that's a bug?

AdamWill:
And sure, closing this is fine by me.

perlpunk:
> In Fedora, perl-Feature-Compat-Try [does not require perl(Syntax::Keyword::Try)](https://koji.fedoraproject.org/koji/rpminfo?rpmID=45727010). Possibly that's a bug?

Yep, that seems wrong:
https://metacpan.org/release/PEVANS/Feature-Compat-Try-0.05/source/META.json#L24

perlpunk:
However, since in perl 5.40 it will just use the builtin `try` feature, it's probably kind of correct that it doesn't require Syntax::Keyword::Try on platforms with perl >= 5.40, so maybe that's why it was removed from the spec?

perlpunk:
> However, since in perl 5.40 it will just use the builtin `try` feature, it's probably kind of correct that it doesn't require Syntax::Keyword::Try on platforms with perl >= 5.40, so maybe that's why it was removed from the spec?

right (and perl 5.36 actually)
https://metacpan.org/release/PEVANS/Feature-Compat-Try-0.05/source/Build.PL#L20
```
my $HAVE_FEATURE_TRY = do {
   # Core's use feature 'try' only supports 'finally' since 5.35.8
   $] >= 5.035008
};

printf "Using %s\n",
   $HAVE_FEATURE_TRY ? "core's feature 'try'" : "Syntax::Keyword::Try";

my $build = Module::Build->new(
   module_name => 'Feature::Compat::Try',
   requires => {
      'perl' => '5.014',
      $HAVE_FEATURE_TRY ? %REQUIRES_PERL : %REQUIRES_SKT,
   },
```

AdamWill:
I guess that's it, yeah. In Fedora perl deps are auto-generated, so either the generator is super smart and realizes we do not need that dep, or it's super dumb and can't handle deps like that so it just defaults to not including it...and either way turns out to be correct. :D

PR #2938, opened 2026-05-24
feat: mouse_drag uses the clickpoints

rclobus:
Instead of using the center, use the clickpoint as defined inside the needle

### DIFF
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -1042,30 +1042,46 @@ subtest '_calculate_clickpoint' => sub {
     my %fake_needle_area = (x => 100, y => 100, w => 50, h => 40);
     my %fake_click_point = (xpos => 20, ypos => 10);
 
-    # Everything is provided.
-    my ($x, $y) = testapi::_calculate_clickpoint(\%fake_needle, \%fake_needle_area, \%fake_click_point);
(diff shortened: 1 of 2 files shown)


rclobus:
The Fullstack CI jobs failed, but that appears to be independent of my changes.

Because the drag start and end points can not always be in the center of the needle (e.g. because the matching of the area is slightly more complex), it is easier to use the needle editor to determine the mouse coordinated than by providing them explicitly in the code.

In Debian we recently started tests with tablet-like behaviour, which involves more dragging (swiping).
A quick look at os-autoinst-distri-opensuse shows that you don't use `mouse_drag` (yet)

perlpunk:
> The Fullstack CI jobs failed, but that appears to be independent of my changes.

Yeah, it will be fixed as soon as a patched module arrives in Tumbleweed & in the container: https://progress.opensuse.org/issues/201441

Martchus:
Looks generally good.

It would be nice if the commit message would be more detailed. The commit message title should also be in imperative, e.g. "feat: Support click points in `mouse_drag`".

okurz:
@mergifyio rebase

okurz (approved):

Martchus (approved):

PR #2939, opened 2026-05-25
video_stream: accept ustreamer version 8

marmarek:
There is no change in the structure layout since version 7, only max
frame size increased.

### DIFF
--- a/consoles/video_stream.pm
+++ b/consoles/video_stream.pm
@@ -330,12 +330,12 @@ sub _receive_frame_ustreamer ($self) {
             $client_clock_offset = 112;
             $data_offset = 129;
             $meta_offset = 32;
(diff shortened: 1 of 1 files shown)


consoles/video_stream.pm
Martchus: This message should probably be updated as well.

Martchus (approved):

okurz (approved):

marmarek:
The CI failure looks unrelated to my change...

Martchus:
Yes, we'll have to wait until the Minion packaging is fixed in TW. (I could of course do some linking on OBS to fix our container downstream but I hope that the problem is fixed with the next TW snapshot.)

perlpunk:
I retriggered the fullstack test and it suceeded

PR #2946, opened 2026-05-27
test: enable pretty serial markers in full-stack test

okurz:
Motivation:
We need automated end-to-end verification that the pretty serial marker
logic functions correctly in a real QEMU environment.

Design Choices:
- Enabled PRETTY_SERIAL_MARKER in the integration test environment.
- Removed the manual bash alias workaround in typing.pm since real
  bash is now available on the remastered Core ISO.

Benefits:
- Ensures the full stack (isotovideo, distribution, backend) correctly
  coordinates for Level 2/3 marker detection and usage.
- Increases the realism of the full-stack test suite.

### DIFF
--- a/t/data/remaster_core_iso.sh
+++ b/t/data/remaster_core_iso.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+set -e
+
+VANILLA_ISO=$1
+OUTPUT_ISO=$2
+
+if [[ -z $VANILLA_ISO || -z $OUTPUT_ISO ]]; then
+    echo "Usage: $0 <vanilla_iso> <output_iso>"
+    exit 1
+fi
+
+WORK_DIR=$(mktemp -d)
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+# Unpack ISO
+mkdir -p "$WORK_DIR/iso"
+mounted=0
+if mount -o loop,ro "$VANILLA_ISO" "$WORK_DIR/iso"; then
+    mounted=1
+else
+    # If mount fails (e.g. no root), try extracting with xorriso.
+    echo "Failed to mount ISO. Trying extraction with xorriso..."
+    if command -v xorriso > /dev/null 2>&1; then
+        xorriso -osirrox on -indev "$VANILLA_ISO" -extract / "$WORK_DIR/iso_content"
+    else
+        echo "xorriso not found. Cannot extract ISO without root/mount."
+        exit 1
+    fi
+fi
+
+if [[ $mounted -eq 1 ]]; then
+    shopt -s nullglob
+    iso_files=("$WORK_DIR/iso"/*)
+    if [[ ${#iso_files[@]} -gt 0 ]]; then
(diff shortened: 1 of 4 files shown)


t/data/remaster_core_iso.sh
```diff
@@ -0,0 +1,72 @@
...
+    exit 1
+fi
+
+WORK_DIR=$(mktemp -d)
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+# Unpack ISO
+mkdir -p "$WORK_DIR/iso"
+mount -o loop,ro "$VANILLA_ISO" "$WORK_DIR/iso" || {
+    # If mount fails (e.g. no root), try extracting with 7z or similar if available,
```
Martchus: I see no mentioning of 7z except in this comment. So it is probably outdated and should be removed.

t/data/remaster_core_iso.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Failed to mount ISO. Trying extraction with xorriso..."
+    if which xorriso > /dev/null 2>&1; then
+        xorriso -osirrox on -indev "$VANILLA_ISO" -extract / "$WORK_DIR/iso_content"
+    else
+        echo "xorriso not found. Cannot extract ISO without root/mount."
+        exit 1
+    fi
+}
+
+if [ -d "$WORK_DIR/iso" ] && [ -n "$(ls -A "$WORK_DIR/iso")" ]; then
```
Martchus: It is probably nicer to just use nullglob instead of `ls`. And besides, we probably also want to unmount `/iso` if it was empty.

t/data/remaster_core_iso.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+set -e
+
+VANILLA_ISO=$1
+OUTPUT_ISO=$2
+
+if [ -z "$VANILLA_ISO" ] || [ -z "$OUTPUT_ISO" ]; then
```
Martchus: This is Bash so I suggest you use `[[` and no quotes for the first operand:


```suggestion
if [[ -z $VANILLA_ISO ]] || [[ -z $OUTPUT_ISO ]]; then
```

The same applies to the other occurrences.

Martchus (approved):

Martchus (changes requested):
See previous inline comments.

t/data/remaster_core_iso.sh
okurz: done

t/data/remaster_core_iso.sh
okurz: done

t/data/remaster_core_iso.sh
okurz: done

okurz:
Needs https://github.com/os-autoinst/openQA/pull/7495 (merged)

EDIT: Amended and rebased

perlpunk (approved):

Martchus (approved):

PR #2948, opened 2026-05-27
feat!: fail by default if always_rollback is not supported

okurz:
Motivation:
Previously, if a test module requested `always_rollback` but snapshots
were not supported by the backend, it would only warn. This can lead to
unexpected behavior where snapshots are assumed to be taken but are
actually skipped.

Design Choices:
Changed the default value of `FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED` from
0 to 1. Incremented the interface version in
`OpenQA/Isotovideo/Interface.pm` to reflect this change in behavior.
Updated documentation and tests accordingly. Used `local` in tests to
ensure clean state.

Benefits:
Ensures that test developers are explicitly aware when a requested
rollback mechanism is not available, preventing silent failures of the
rollback logic.

BREAKING CHANGE: test modules with `always_rollback` executed in an
environment where snapshots are not supported fail now explicitly. Adapt
the test schedule accordingly (or temporarily mute the behaviour with
FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED=0).

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -23,6 +23,8 @@ use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use Mojo::File qw(path);
 use File::Glob qw(bsd_glob);
 
+use constant FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED => 1;
+
 our @EXPORT_OK = qw(loadtest $selected_console $last_milestone_console query_isotovideo);
 
 our %tests;    # scheduled or run tests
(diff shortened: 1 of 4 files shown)


Martchus (approved):

autotest.pm
```diff
@@ -540,7 +540,7 @@ sub runalltests () {
     for (my $testindex = 0; $testindex <= $#testorder; $testindex++) {
         my $t = $testorder[$testindex];
         my $flags = $t->test_flags();
-        if ($flags->{always_rollback} && !$snapshots_supported && $bmwqemu::vars{FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED}) {
+        if ($flags->{always_rollback} && !$snapshots_supported && ($bmwqemu::vars{FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED} // 1)) {
```
perlpunk: Maybe in this case it makes sense to introduce a constant to have a clearly visible default on top of the file:
```
use constant FAIL_ON_ALWAYS_ROLLBACK_NOT_SUPPORTED => 1;
```

okurz: For a non-boolean value I would have done that already. But I can also do that for this Boolean, sure

perlpunk (approved):

PR #2951, opened 2026-06-01
feat: integrate C++ style checks into CI

okurz:
Motivation:
C++ files were found to be untidied because there was no automated check
ensuring they follow the project's style. Integrating this check into
the test suite prevents style regressions.

Design Choices:
- Created tools/check-cpp-style to verify C++ formatting using clang-format.
- Added test-local-cpp-style to cmake/test-targets.cmake to include it in
  the standard test suite.
- Refactored tidy-cpp in CMakeLists.txt to use the new check script with
  the --fix flag for consistency.
- Tidied snd2png/snd2png.cpp which was previously missed.

Benefits:
- Automated enforcement of C++ code style in CI.
- Unified file discovery logic between checking and tidying.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,7 @@ find_program(CLANG_FORMAT_PATH clang-format)
 add_custom_target(
     tidy-cpp
     COMMENT "Formatting C++ source code"
-    COMMAND sh -c "'${CLANG_FORMAT_PATH}" "--style=WebKit" "-i" "$$(find" "." "-name" "\*.h" "-o" "-name" "\*.cc" "-o" "-name" "\*.cpp)'"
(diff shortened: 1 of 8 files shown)


perlpunk (approved):

Martchus (approved):
Ok; just be aware that this means contributors need the same version of LLVM/Clang as we use in the CI and if that's not available in their distribution compile it themselves or use our TW-based CI container. (I'm just saying this because getting a specific version of LLVM/Clang is not as easy as with Perl Tidy.)

Martchus:
By the way, there's also `cmake-format`. Unfortunately, [the best search result on OBS](https://build.opensuse.org/search?search_text=cmake-format) is still my home project and when I tried to submit the package it wasn't accepted due to nitpicking. *Probably* it is not worth anyway. I'm using it in my repos so from that experience I can tell that the results are only mixed (sometimes better than before but sometimes also worse).

PR #2952, opened 2026-06-01
fix: script_run - type command and marker together on serial console

AdamWill:
When doing script_run on a serial console, we "type" the command and then "type" the marker string in separate calls. Sometimes, somehow, extraneous characters seem to be added between the two in the output echoed back over the serial line, breaking the subsequent check. Typing them in one call should avoid this.

Related ticket: https://progress.opensuse.org/issues/201516

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -181,8 +181,7 @@ sub script_run ($self, $cmd, @args) {
             my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
             my $final_cmd = $skip_pretty ? "OA_NO_MARKER=1; $cmd" : $cmd;
             if (testapi::is_serial_terminal) {
-                testapi::type_string "$final_cmd", max_interval => $args{max_interval};
-                testapi::type_string $marker, max_interval => $args{max_interval};
+                testapi::type_string "$final_cmd$marker", max_interval => $args{max_interval};
                 testapi::wait_serial($final_cmd . $marker, no_regex => 1, quiet => $args{quiet}, buffer_size => (length $final_cmd) + 128, internal_marker => 1)
                   or _handle_cmd_typing_error($final_cmd, \%args);
                 testapi::type_string "\n", max_interval => $args{max_interval};
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -382,12 +382,7 @@ subtest 'script_run' => sub {
     is(assert_script_run('true', max_interval => 1), undef, 'nothing happens on success (slow typing)');
     is_deeply($cmds, [
             {
-                text => 'true',
-                cmd => 'backend_type_string',
-                max_interval => 1
-            },
-            {
-                text => '; echo XXX-$?-',
+                text => 'true; echo XXX-$?-',
                 cmd => 'backend_type_string',
                 max_interval => 1
             },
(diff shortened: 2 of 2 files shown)


distribution.pm
```diff
@@ -181,8 +181,7 @@ sub script_run ($self, $cmd, @args) {
             my $marker = "; echo $str-\$?-" . ($args{output} ? "Comment: $args{output}" : '');
             my $final_cmd = $skip_pretty ? "OA_NO_MARKER=1; $cmd" : $cmd;
             if (testapi::is_serial_terminal) {
-                testapi::type_string "$final_cmd", max_interval => $args{max_interval};
-                testapi::type_string $marker, max_interval => $args{max_interval};
+                testapi::type_string "$final_cmd$marker", max_interval => $args{max_interval};
```
okurz, line 184: With this change I see two drawbacks:
1. The "command" and marker are appearing in the same log-line so there is no explicit, unique log line for the actual command anymore
2. This would be inconsistent with the VNC console typing in lines R190-R191

So given those both drawbacks I am not convinced this is a good idea.

okurz (changes requested):

distribution.pm
AdamWill: It would be trivial to change the VNC console typing, but it seems unnecessary because these are really different codepaths. `type_string` is a misnomer on serial consoles; nothing is being "typed", we're just writing strings to a socket. On a VNC-ish console, the two approaches would be effectively identical: whichever way you write it, what actually *happens* is a series of individual keypress VNC events. On the serial console they are not; you get one `syswrite` per `type_string` call.

Point 1 is fair, but frankly, at this point I've sunk over a week into this freaking bug, and I am entirely blocked on using current os-autoinst in Fedora because of it. I don't want us to be marooned on an ancient os-autoinst forever and I don't want to maintain a downstream patch forever, but I also don't want to waste another week submarining into qemu code or whatever the heck turns out to be actually injecting these additional characters somehow. So I don't see another practical way to fix this. If you reject this I will just have to carry it downstream forever or until somebody else fixes whatever the heck the real bug is (and I notice it's been fixed somehow).

AdamWill:
Test failure appears to be an infra issue.

distribution.pm
okurz: nah, it's fine. I will accept it. Just wanted to bring this up in case you see a better alternative.

okurz (approved):

okurz:
> Test failure appears to be an infra issue.

retriggered the CI run

distribution.pm
AdamWill: Yeah, I do accept it's slightly worse from a logging POV, but I don't really see a better alternative. The alternative I can see is "adjust the check to be more forgiving" but that feels messier and more complex to me.

I'll maybe adjust this to add a comment explaining why we type them together.

AdamWill: well, hmm. I suppose another option would be to do just `testapi::wait_serial($marker...` - i.e. just wait on the marker, not the whole combined string? The actual *purpose* of this wait is to wait until we're sure the VM has received and "processed" what we "typed", right? We don't really need to check that it got all the characters?

okurz: true. Sounds sensible but don't know what further effect that would bring. I assume our current unit test coverage for those specifics of virtio-serial-terminals is lacking in that regard. Also be aware that soon I want to switch to `PRETTY_SERIAL_MARKER=1` being default meaning we wouldn't type the serial marker but instead let the bash prompt_command forward it separately. Anyway, I approved this PR, we can try it out and see how to handle further parts later based on monitoring results in production.

Martchus (approved):

distribution.pm
AdamWill: Well, AFAICS the serial console path is kinda "exempted" from `PRETTY_SERIAL_MARKER` anyway:
```
    my $level = 1;
    my $pretty = $self->get_pretty_serial_marker();
    my $serial_term = testapi::is_serial_terminal();
    return $self->{_serial_marker_level}->{$console} = $level if !$pretty || $serial_term;
```
i.e. `_detect_serial_marker_capability` returns level 1 for serial consoles always, which means we go down the non-pretty-marker path in `script_run`. This is presumably why, when I tried enabling pretty serial markers as a workaround for this bug, it did nothing.

I have to say that, just looking at it superficially (the bits of it I've encountered while looking at this bug, plus the number of commits it's taking to implement), the 'pretty serial marker' stuff comes off as potentially worryingly over-complex to me, and I especially do not like that it modifies the SUT config. But maybe if I take a closer look at it I'll like it.

AdamWill:
@Martchus oh, I was kinda meaning to come to a conclusion on the above discussion before merging. Oh, well, we can always send followups. Thanks for the review.

PR #2954, opened 2026-06-02
feat: Support configuring UEFI firmware dynamically

Martchus:
Related ticket: https://progress.opensuse.org/issues/200156

---

Note tested yet for real, hence still a draft.


Martchus:
I tested at least the resolution change locally. I was able to clone a job that failed using the default vars on a screen match in GRUB due to the wrong resolution. Cloning the job again with `UEFI_PFLASH_RES=800x600` made a difference and let the job go passed this point. I found this job as it was mentioned in https://progress.opensuse.org/issues/200075. (I still couldn't fully run this job on my local machine because it used `%REPO_MIRROR_HOST%` which I haven't configured yet.)

I'll still have to see what job to clone to test enrolling certificates.

okurz:
How does this relate to https://github.com/os-autoinst/os-autoinst/pull/2906?

Martchus:
I'm not sure where to find a job that fails without special certificate. However, I've just cloned any job with `UEFI_PFLASH_CERTS='/home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/suse-unsupported.crt;/home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/openSUSE-Factory-Staging.crt'` and then checked the resulting vars (in the pool directory) with `virt-fw-vars -i ovmf-x86_64-ms-4m-vars-adjusted.bin --extract-certs`. I've got the following:

```
/usr/lib/python3.13/site-packages/virt/firmware/efi/certs.py:10: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import resource_filename
INFO: reading raw edk2 varstore from ovmf-x86_64-ms-4m-vars-adjusted.bin
INFO: var store range: 0x64 -> 0x40000
INFO: writing: KEK-a0baa8a3-041d-48a8-bc87-c36d121b5e3d-openSUSEFactoryStagingOBSProject.pem
INFO: writing: KEK-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationKEKCA2011.pem
INFO: writing: KEK-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationKEK2KCA2023.pem
INFO: writing: PK-a0baa8a3-041d-48a8-bc87-c36d121b5e3d-openSUSEFactoryStagingOBSProject.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftWindowsProductionPCA2011.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-WindowsUEFICA2023.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationUEFICA2011.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftUEFICA2023.pem
```

So  the FactoryStaging certificate is there but SUSELinuxEnterpriseSecureBootCA was lost compared to the case:

```
virt-fw-vars -i /usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin --extract-certs
/usr/lib/python3.13/site-packages/virt/firmware/efi/certs.py:10: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import resource_filename
INFO: reading raw edk2 varstore from /usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin
INFO: var store range: 0x64 -> 0x40000
INFO: writing: KEK-a0baa8a3-041d-48a8-bc87-c36d121b5e3d-SUSELinuxEnterpriseSecureBootCA.pem
INFO: writing: KEK-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationKEKCA2011.pem
INFO: writing: KEK-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationKEK2KCA2023.pem
INFO: writing: PK-a0baa8a3-041d-48a8-bc87-c36d121b5e3d-SUSELinuxEnterpriseSecureBootCA.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftWindowsProductionPCA2011.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-WindowsUEFICA2023.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftCorporationUEFICA2011.pem
INFO: writing: db-77fa9abd-0359-4d32-bd60-28f4e78f784b-MicrosoftUEFICA2023.pem
```

Maybe it isn't possible to specify multiple `--enroll-cert` args. That was the invocation:

```
running `virt-fw-vars -i /usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin -o /home/martchus/dev/openqa/openqa/pool/1/ovmf-x86_64-ms-4m-vars-adjusted.bin --enroll-cert /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/suse-unsupported.crt --enroll-cert /home/martchus/dev/osc2/home:favogt:stagingovmf/ovmf/openSUSE-Factory-Staging.crt --set-json /home/martchus/dev/openqa/openqa/pool/1/resolution-fw-conf.json
```

So I'll probably still have to tweak that. The resolution was still applied, though. (So the combination of `--enroll-cert` and `--set-json` seems to be ok.)

EDIT: Looks like one can only specify one cert at a time. Even when using multiple invocations the certs overwrite each other, though. Maybe we should just limit this to one cert.

---

> How does this relate to https://github.com/os-autoinst/os-autoinst/pull/2906?

That PR allows passing a JSON file but you raised this question yourself: "hat to put into that JSON file? This PR answers that by letting os-autoinst generate that JSON file. (I doubt users want to concatenate two 32-bit little-endian numbers and convert that to hex just to specify the resolution. So it is better if os-autoinst handles that.)

It also allows enrolling certificates which doesn't seem to be covered by https://github.com/os-autoinst/os-autoinst/pull/2906.

---

I was also thinking that thinking that `UEFI_PFLASH_RES=800x600` should be the default as all tests seem to expect that. However, that would mean adjusting lots of unit tests.

Vogtinator:
> EDIT: Looks like one can only specify one cert at a time. Even when using multiple invocations the certs overwrite each other, though. Maybe we should just limit this to one cert.

I suspect `--enroll-cert` does not work, but probably `--add-db` will. The former probably sets the PK, of which there can only be one.

> I was also thinking that thinking that UEFI_PFLASH_RES=800x600 should be the default as all tests seem to expect that. However, that would mean adjusting lots of unit tests.

IMO using 1024x768 would be even better, as that's the native resolution in openQA so far. May require some reneedling though.

Martchus:
> I suspect --enroll-cert does not work, but probably --add-db will. The former probably sets the PK, of which there can only be one.

Yes, that's also what Gemini suggested. So I changed it to use `--add-db` for additional certs.

> IMO using 1024x768 would be even better, as that's the native resolution in openQA so far. May require some reneedling though.

Yes, that would be better in the long run. I'll probably avoid setting a default for now until we've settled on something.

I have also added an option to enable and disable secure boot because that's also done in the spec file explicitly and it can't hurt if we can now turn it on and off via a test variable explicitly as well.

I've also tested whether the openQA asset system can be used to supply the certs and this works without any additional changes. So one can just copy the certs to `other` (or `other/fixed`) and start a test with e.g. `UEFI_PFLASH_CERTS='suse-unsupported.crt;openSUSE-Factory-Staging.crt' ASSET_1=suse-unsupported.crt ASSET_2=openSUSE-Factory-Staging.crt`. The worker will make the certs available in the pool directory. I also tested cloning such a job and openqa-clone-job would download the certs as expected.

OpenQA/Qemu/Proc.pm
```diff
@@ -240,6 +240,56 @@ sub configure_blockdevs ($self, $bootfrom, $basedir, $vars) {
...
+    return () unless defined $enable_or_disable_secure_boot;
+    return (
+        $enable_or_disable_secure_boot ? '--set-true' : '--set-false' => 'SecureBootEnable',
+        !$enable_or_disable_secure_boot ? '--set-true' : '--set-false' => 'CustomMode',
+    );
+}
+
+=head3 _uuid
+
+Generate random UUID for use with `virt-fw-vars`.
```
Vogtinator: I don't particularly like using randomness in tests.

IMO just use the one the spec file used or generate a stable one for os-autoinst.

Martchus: One can add more than one additional cert so just one UUID wouldn't be enough. Maybe I can generate them using pseudo random numbers and a stable seed.

Vogtinator: Is it not possible to use the same UUID multiple times? The UUID is just to represent who added it, it's not the UUID of the entry or anything like that.

OpenQA/Qemu/Proc.pm
```diff
@@ -240,6 +240,57 @@ sub configure_blockdevs ($self, $bootfrom, $basedir, $vars) {
...
+}
+
+=head3 _uuid
+
+Generates a new UUID for use with `virt-fw-vars`. Expects a well-known UUID as "namespace"
+to be able to produce consistent UUIDs.
+
+=cut
+
+sub _uuid ($ns, $name) { (runcmd_out('uuidgen', '--sha1', '--namespace', $ns, '--name', $name))[1] }
```
Vogtinator: No need for this complexity.

The original thread is unfortunately shown as "outdated" now

Martchus: Ok, I'm using just a single UUID now. It looks like at least `virt-fw-vars` has nothing against that.

okurz (approved):

doc/backend_vars.md
```diff
@@ -202,6 +202,9 @@ Supported variables per backend
 | UEFI_SECURE_BOOT | boolean | 1 | Whether the auto-selection of `UEFI_PFLASH_CODE` and `UEFI_PFLASH_VARS` should use firmware with SecureBoot enabled |
 | UEFI_PFLASH_CODE | string |  | Specify the file name of the UEFI firmware code which will be loaded onto a read-only PFLASH drive |
 | UEFI_PFLASH_VARS | string |  | Specify the file name which contains the UEFI firmware variables which will be loaded onto a mutable PFLASH drive |
+| UEFI_PFLASH_CERTS | string |  | Specify the semicolon-separated file names of certificates that should be enrolled into the UEFI firmware variables file using `virt-fw-vars`. |
+| UEFI_PFLASH_SECURE_BOOT | boolean |  | Enable or disable secure boot in the UEFI firmware variables file using `virt-fw-vars`. |
+| UEFI_PFLASH_RES | string |  | Specify the resolution to configure via UEFI firmware variables file using `virt-fw-vars`, e.g. `800x600` for a with of 800 pixels and a height of 600 pixels. |
```
Vogtinator: Nit: `RES` is a bit too short, could also stand for "result" or so. IMO `DISPLAY_RES` or `RESOLUTION` would be nicer.

Vogtinator (approved):

aylen384 (approved):

PR #2956, opened 2026-06-03
feat(autotest): run always_run modules after SIGTERM cancellation

Wabri:
On job cancellation the worker sends SIGTERM to the whole os-autoinst
process group. autotest's run _exit(1) immediately, so `always_run`
modules never ran properly on on cancellation, even if they run after a
fatal test failure. This makes cancellation act like a fatal failure:
stop scheduling normal modules but still run the always_run ones.

The issue is that keep autotest alive is not enough because the same
SIGTERM also put down isotovideo and the other services. This means that
`always_run` module died with "myjsonrpc: remote end terminated
connection".

This change add a graceful shutdown feature:
- autotest: handle_sigterm() sets $fatal_reason instead of _exit()
- isotovideo: _signal_handler() enters graceful shutdown on the first
  SIGTERM and keeps the main loop alive until tests_done peers, pending
  backend request, clear_serial_buffer).

All of this still keep `always_run` module related to the worker's
timeout.

Related: https://progress.opensuse.org/issues/197822

### DIFF
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -34,6 +34,9 @@ has [qw(backend command_handler)];
 # the loop status
 has loop => 1;
 
+# track if we're in graceful shutdown (waiting for always_run modules)
+has graceful_shutdown => 0;
+
 use constant {
     EXIT_STATUS_OK => 0,
     EXIT_STATUS_ERR_NO_TESTS => 100,
@@ -54,7 +57,7 @@ sub run ($self) {
         my ($ready_for_read, $ready_for_write, $exceptions) = IO::Select::select($io_select, undef, $io_select, $ch->timeout);
         for my $readable (@$ready_for_read) {
             my $rsp = myjsonrpc::read_json($readable);
-            $self->_read_response($rsp, $readable);
+            $self->_read_response($rsp, $readable, $io_select);
             last unless defined $rsp;
         }
         $ch->check_asserted_screen if defined($ch->tags);
@@ -63,8 +66,22 @@ sub run ($self) {
     return 0;
 }
 
-sub _read_response ($self, $rsp, $fd) {
+sub _read_response ($self, $rsp, $fd, $io_select = undef) {
     if (!defined $rsp) {
+        if ($self->graceful_shutdown && $io_select && defined $self->testfd && $fd != $self->testfd) {
+            $io_select->remove($fd);
+            my $ch = $self->command_handler;
+            if (defined $self->cmd_srv_fd && $fd == $self->cmd_srv_fd) {
+                $ch->cmd_srv_fd(undef);
+            }
+            elsif (defined $ch->backend_out_fd && $fd == $ch->backend_out_fd) {
+                $ch->send_to_backend_requester({ret => {}}) if $ch->backend_requester;
+                $ch->backend_fd(undef);
+                $ch->backend_out_fd(undef);
+            }
+            diag('isotovideo: a helper connection closed during graceful shutdown, continuing to run remaining always_run modules');
+            return;
+        }
         fctwarn sprintf 'THERE IS NOTHING TO READ %d %d %d', fileno($fd), fileno($self->testfd), fileno $self->cmd_srv_fd;
         $self->loop(0);
     } elsif ($fd == $self->command_handler->backend_out_fd) {
@@ -116,9 +133,10 @@ sub handle_commands ($self) {
     my $command_handler;
     # stop main loop as soon as one of the child processes terminates
     my $stop_loop = sub (@) { $self->loop(0) if $self->loop; };
+    my $stop_loop_unless_graceful = sub (@) { $self->loop(0) if $self->loop && !$self->graceful_shutdown; };
     $self->testprocess->once(collected => $stop_loop);
-    $self->backend->process->once(collected => $stop_loop);
-    $self->cmd_srv_process->once(collected => $stop_loop);
+    $self->backend->process->once(collected => $stop_loop_unless_graceful);
+    $self->cmd_srv_process->once(collected => $stop_loop_unless_graceful);
 
     $command_handler = OpenQA::Isotovideo::CommandHandler->new(
         cmd_srv_fd => $self->cmd_srv_fd,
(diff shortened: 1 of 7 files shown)


t/08-autotest.t
```diff
@@ -234,6 +234,47 @@ subtest 'test always_rollback flag' => sub {
...
+
+        $vm_stopped = 0;
+        my $output = combined_from { autotest::run_all };
+        like $output, qr/autotest received signal TERM.*saving results/, 'SIGTERM logged';
+        like $output, qr/scheduled stop of overall test execution after test cancellation/, 'cancellation scheduled';
+        like $output, qr{starting next tests/next\.pm}, 'executed always_run cleanup test';
+        ($died, $completed) = get_tests_done;
+        is $died, 0, 'tests not considered died';
+        is $completed, 0, 'tests did not complete successfully';
+        ok $vm_stopped, 'VM was stopped eventually';
```
okurz: You are introducing significant duplication. Please deduplicate

okurz (changes requested):

Martchus (changes requested):
You didn't state that you tested this with the openQA worker. It'll probably break when a job is cancelled by the openQA worker (e.g. because the user clicks on the cancel button on the web UI or because the job exceeds the `MAX_JOB_TIME`).

The problem I have in mind is that the openQA worker will send SIGTERM to the whole process group so it'll terminate the backend and other processes. If those processes don't delay their termination until autotest has finished this won't work.

By the way, an *interactive* Bash will also behave like the openQA worker when pressing Ctrl+C (except that it'll send SIGINT instead of SIGTERM). So you might also use that for testing during development/experimentation.

backend/baseclass.pm
```diff
@@ -95,9 +95,13 @@ sub die_handler ($msg) {
     $backend->close_pipes();
 }
 
+our $graceful_shutdown_requested = 0;    ## no critic (Variables::ProhibitPackageVars)
```
perlpunk: Why is this variable declared with `our`? Do you need this outside of this file?
If not, just use `my`.
We added `no critic` exceptions for existing violations because we have to stay backwards compatible, but for new code don't do this or explain why it is needed.
Otherwise we could just stop using perlcritic.

perlpunk (changes requested):

perlpunk (commented):
The PR talks about `always_run`, however the ticket has the following ACs:
> * AC1: Modules with test_flags `{ fatal => 1} `will be executed, even if the test run was aborted by a previously failing test module with `fatal => 1`
> * AC2: Modules with test_flags `{ fatal => 1}` will be executed, even if the test run was cancelled

I don't understand this.

Wabri:
> The PR talks about `always_run`, however the ticket has the following ACs:
> 
> > * AC1: Modules with test_flags `{ fatal => 1} `will be executed, even if the test run was aborted by a previously failing test module with `fatal => 1`
> > * AC2: Modules with test_flags `{ fatal => 1}` will be executed, even if the test run was cancelled
> 
> I don't understand this.

Since I first open this pr on 03/06, I don't recall the entire history, but if you look at https://progress.opensuse.org/issues/197822#note-5 oliver ask if they meant `always_run` instead of `fatal` and by looking at the phrase it make sense and it does not in the other way: with `fatal` true we expect fatal and not continue the execution.

Wabri:
> You didn't state that you tested this with the openQA worker. It'll probably break when a job is cancelled by the openQA worker (e.g. because the user clicks on the cancel button on the web UI or because the job exceeds the `MAX_JOB_TIME`).
> 
> The problem I have in mind is that the openQA worker will send SIGTERM to the whole process group so it'll terminate the backend and other processes. If those processes don't delay their termination until autotest has finished this won't work.
> 
> By the way, an _interactive_ Bash will also behave like the openQA worker when pressing Ctrl+C (except that it'll send SIGINT instead of SIGTERM). So you might also use that for testing during development/experimentation.

As I mention in the ticket https://progress.opensuse.org/issues/197822:
> I run the test and wait for the worker to handle it, than run the SIGTERM and look at the test that is still running

perlpunk:
> but if you look at https://progress.opensuse.org/issues/197822#note-5 oliver ask if they meant `always_run` instead of `fatal`

I read that, yeah. But I never saw a reply to that question, and IMHO this should be clarified and the ticket description fixed. Why should the ticket keep the wrong description forever? To confuse future readers?

Wabri:
> I read that, yeah. But I never saw a reply to that question, and IMHO this should be clarified and the ticket description fixed. Why should the ticket keep the wrong description forever? To confuse future readers?

I agree, but I still thought it might be helpful to move forward with this pr.

okurz (changes requested):
Please put more details in the commit message. Describe the motivation. Also, what happens if another SIGTERM is received? What happens if the always_run module gets stuck?

Wabri:
> Please put more details in the commit message. Describe the motivation. Also, what happens if another SIGTERM is received? What happens if the always_run module gets stuck?

You are righ, the first changes diverge a little bit after the latest changes. I've update the title, description on bot hcommit message and pr.

OpenQA/Isotovideo/Runner.pm
```diff
@@ -131,16 +149,20 @@ sub handle_commands ($self) {
...
+            if ($self->graceful_shutdown) {    # uncoverable statement
+                bmwqemu::diag('isotovideo: stopping backend');    # uncoverable statement
+                $self->backend->stop if defined $self->backend;    # uncoverable statement
+                $self->stop_commands('graceful shutdown');    # uncoverable statement
+            }
     });
     # uncoverable statement count:1
     # uncoverable statement count:2
     # uncoverable statement count:3
     # uncoverable statement count:4
```
Wabri: TODO: remove no need of those annotation anymore.

Martchus:
When I remember correctly, the worker will send SIGTERM to the whole process group (including the video encoder and ffmpeg and dewebsockify and whatever other processes the tests need). That's not nice to handle gracefully. So maybe we also need to adjust the worker here to send SIGTERM only to the main isotovideo process. (If it already does that then we're ok of course.)

> Also, what happens if another SIGTERM is received? What happens if the always_run module gets stuck?

The worker will send SIGKILL at some point, again to the whole process group. So we already have means to handle that. However, we might need to increase the time the worker waits until it sends SIGKILL considering we now might need to run further modules. Maybe the worker should be able to send SIGKILL immediately when a job is cancelled that has already been cancelled so we have a way of stopping those jobs after all (without having to wait a possibly long timeout).

Wabri:
Do we want to handle the worker change here or do we want to open a followup ticket for that?

okurz:
> Do we want to handle the worker change here or do we want to open a followup ticket for that?

I would say without further worker changes I don't even see this feature working

PR #2957, opened 2026-06-04
style(perlcritic): Add Community::WarningsSwitch

Wabri:
Community::WarningsSwitch forbids using -W switch to prevent enable warnings even in modules where should be disabled.

https://metacpan.org/pod/Perl::Critic::Policy::Community::WarningsSwitch

> The -w switch enables warnings globally in a perl program, including for any modules that did not explicitly enable or disable any warnings. The -W switch enables warnings even for modules that explicitly disabled them. The primary issue with this is enabling warnings for code that you did not write. Some of these modules may not be designed to run with warnings enabled, but still work fine. Instead, use [warnings](https://metacpan.org/pod/warnings) within your own code only.

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -1,6 +1,6 @@
 theme = community + openqa
 severity = 4
-include = strict ValuesAndExpressions::ProhibitInterpolationOfLiterals CodeLayout::ProhibitParensWithBuiltins BuiltinFunctions::RequireBlockMap BuiltinFunctions::ProhibitStringySplit ControlStructures::ProhibitNegativeExpressionsInUnlessAndUntilConditions ValuesAndExpressions::RequireQuotedHeredocTerminator Variables::ProhibitPackageVars CodeLayout::RequireTidyCode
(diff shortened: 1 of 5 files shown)


perlpunk:
> The -w switch enables warnings globally in a perl program, including for any modules that did not explicitly enable or disable any warnings

So strictly speaking, this is a breaking change, mostly for isotovideo :)
Breaking not in the sense that tests will now fail, but that warnings for modules that don't `use warnings` won't appear anymore. That includes test code, but also libraries from CPAN for example.

But I think we should do this, and test writers should be responsible.
@okurz WDYT and should we still increase the interface version for this change?

okurz:
> > The -w switch enables warnings globally in a perl program, including for any modules that did not explicitly enable or disable any warnings
> 
> So strictly speaking, this is a breaking change, mostly for isotovideo :) Breaking not in the sense that tests will now fail, but that warnings for modules that don't `use warnings` won't appear anymore. That includes test code, but also libraries from CPAN for example.
> 
> But I think we should do this, and test writers should be responsible. @okurz WDYT and should we still increase the interface version for this change?

Yes, change looks ok. No, don't need to bump interface version as the functionality of the tests will not change directly

Martchus (approved):

okurz (approved):

PR #2959, opened 2026-06-04
style(perlcritic): Add Community::EmptyReturn

Wabri:
Add return arguments to all returns without any

https://metacpan.org/pod/Perl::Critic::Policy::Community::EmptyReturn

> Context-sensitive functions, while one way to write functions that
    DWIM (Do What I Mean), tend to instead lead to unexpected behavior when
    the function is accidentally used in a different context, especially if
    the function's behavior changes significantly based on context. This
    also can lead to vulnerabilities when a function is intended to be used
    as a scalar, but is used in a list, such as a hash constructor or
    function parameter list. return with no arguments will return either
    undef or an empty list depending on context. Instead, return the
    appropriate value explicitly.

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -168,8 +168,8 @@ sub do_stop_vm ($self, @) {
 sub can_handle ($self, $args) {
     my $vars = \%bmwqemu::vars;
 
-    return unless $args->{function} eq 'snapshots';
-    return if $vars->{QEMU_DISABLE_SNAPSHOTS};
+    return undef unless $args->{function} eq 'snapshots';
+    return undef if $vars->{QEMU_DISABLE_SNAPSHOTS};
     my @models = ($vars->{HDDMODEL}, map { $vars->{"HDDMODEL_$_"} } (1 .. $vars->{NUMDISKS}));
     my $nvme = first { ($_ // '') eq 'nvme' } @models;
     return {ret => 1} unless $nvme;
@@ -341,7 +341,7 @@ sub save_memory_dump ($self, $args) {
 
     my $rsp = $self->handle_qmp_command({execute => 'query-status'}, fatal => 1);
     bmwqemu::diag("Migrating the machine (Current VM state is $rsp->{return}->{status})");
-    my $was_running = $rsp->{return}->{status} eq 'running';
+    my $was_running = $rsp->{return}->{status} eq 'running';    ## no critic (Community::EmptyReturn)
 
     mkpath('ulogs');
     $self->_migrate_to_file(compress_level => $compress_method eq 'internal' ? $compress_level : 0,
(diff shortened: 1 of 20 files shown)


Martchus:
Looks generally good but tests are failing due to some syntax mistakes.

backend/qemu.pm
```diff
@@ -210,9 +210,9 @@ sub _wait_while_status_is ($self, $status, $timeout, $fail_msg) {
 
     my $rsp = $self->handle_qmp_command({execute => 'query-status'}, fatal => 1);
     my $i = 0;
-    while (($rsp->{return}->{status} // '') =~ $status) {
+    while (($rsp->{return undef }->{status} // '') =~ $status) {
```
perlpunk: That's a hash key, not a return statement

backend/qemu.pm
```diff
@@ -426,13 +426,13 @@ sub save_storage ($self, $args) {
                 arguments => {device => $node, target => $my_node,
                     sync => 'full', 'job-id' => $id}};
             $self->handle_qmp_command($req, fatal => 1);
-            my $return;
+            my $return undef;
```
perlpunk: That's a variable, not a return statement

backend/qemu.pm
Wabri: Mh yes, I've miss configured the bash script I think!

backend/qemu.pm
Wabri: Same as above, let me check all the return I change...

perlpunk:
Have a look at the policy description, especially:
> Empty returns are permitted by this policy if the subroutine contains no explicit return values, indicating it is intended to be used in void context.

That means if a subroutine only has empty return statements, they are allowed.
So I suggest to only fix the actual violations and not replace every empty return with a `return undef`.

okurz (approved):

backend/qemu.pm
```diff
@@ -341,7 +341,7 @@ sub save_memory_dump ($self, $args) {
 
     my $rsp = $self->handle_qmp_command({execute => 'query-status'}, fatal => 1);
     bmwqemu::diag("Migrating the machine (Current VM state is $rsp->{return}->{status})");
-    my $was_running = $rsp->{return}->{status} eq 'running';
+    my $was_running = $rsp->{return}->{status} eq 'running';    ## no critic (Community::EmptyReturn)
```
perlpunk, line 344: That's a bug in the policy implementation, btw.

myjsonrpc.pm
```diff
@@ -106,7 +106,7 @@ sub read_json ($socket, $cmd_token = undef, $multi = undef) {
         handle_read_error($fd) until (my @res = $s->can_read);
 
         my $qbuffer;
-        if (!sysread $socket, $qbuffer, READ_BUFFER) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(); return }
+        if (!sysread $socket, $qbuffer, READ_BUFFER) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(); return undef }
```
perlpunk:
```suggestion
        if (!sysread $socket, $qbuffer, READ_BUFFER) { bmwqemu::fctwarn("sysread failed: $!") if is_debug(); return $multi ? () : undef }
```

perlpunk (commented):
LGTM, only https://github.com/os-autoinst/os-autoinst/pull/2959/changes#r3381911517

backend/qemu.pm
Wabri: Mh ok can we do it something to resolve this?

perlpunk: I will create a bug report

perlpunk (approved):

PR #2960, opened 2026-06-04
style(perlcritic): Add Variables::ProhibitPackageVars

perlpunk:
Variables::ProhibitPackageVars forbids using global package variables.

https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProhibitPackageVars

> Conway suggests avoiding package variables completely, because they expose
> your internals to other packages. Never use a package variable when a lexical
> variable will suffice. If your package needs to keep some dynamic state,
> consider using an object or closures to keep the state private.
>
> This policy assumes that you're using `strict vars` so that naked variable
> declarations are not package variables by default. Thus, it complains you
> declare a variable with our or use vars, or if you make reference to variable
> with a fully-qualified package name.

Issue: https://progress.opensuse.org/issues/155191

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -43,3 +43,6 @@ severity = 5
 # -- Superfluous use strict/warning.
 [OpenQA::RedundantStrictWarning]
 equivalent_modules = Test::Most
+
+[Variables::ProhibitPackageVars]
+add_packages = autotest bmwqemu log needle testapi Carp OpenQA::Isotovideo::Interface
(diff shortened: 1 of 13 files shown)


perlpunk:
I'm helping @Wabri here because this was a bit more involved than a simple style change.
I had to add lots of exceptions because we use a lot of global variables here (unfortunately) and many of them are even used from test modules. Instead we should have provided functions, because providing variables makes it impossible to change the internal implementation later, and it's hard to find out where a variable was is changed.

okurz (approved):

Wabri (approved):

PR #2965, opened 2026-06-09
style(perlcritic): Add Variables::ProhibitUnusedVariables

Wabri:
Unused variables should not be use

https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProhibitUnusedVariables

> Unused variables clutter code and require the reader to do mental
    bookkeeping to figure out if the variable is actually used or not.


t/10-test-image-conversion-benchmark.t
```diff
@@ -18,7 +18,7 @@ cv::init();
 require tinycv;
 
 
-my ($res, $needle, $image, $cand, $img_src);
+my ($res, $needle, $image, $cand, $img_src);    ## no critic (Variables::ProhibitUnusedVariables)
```
perlpunk: Why didn't it work to just remove the variables?

bmwqemu.pm
```diff
@@ -36,7 +36,7 @@ $| = 1;
 our $default_timeout = 30;    ## no critic (Variables::ProhibitPackageVars)
 our $openqa_default_share = '/var/lib/openqa/share';    ## no critic (Variables::ProhibitPackageVars)
 
-my @ocrrect;
+my @ocrrect;    ## no critic (Variables::ProhibitUnusedVariables)
```
perlpunk: Why didn't it work to just remove the variable?

t/10-test-image-conversion-benchmark.t
Wabri: I though that those are there for better readability of the entire tests, but if you say that I can remove them I will do it :100:

bmwqemu.pm
Wabri: I overthink about a possible deprecation and need it for some reason!

Martchus (approved):

okurz (approved):

t/10-test-image-conversion-benchmark.t
```diff
@@ -18,7 +18,7 @@ cv::init();
 require tinycv;
 
 
-my ($res, $needle, $image, $cand, $img_src);
+my ($image, $img_src);
```
perlpunk, line 21: Those can be removed as well, see #2986

PR #2972, opened 2026-06-10
style(perlcritic): Add BuiltinFunctions::RequireBlockGrep

Wabri:
Create block around grep string messages

https://metacpan.org/pod/Perl::Critic::Policy::BuiltinFunctions::RequireBlockGrep

> The expression forms of grep and map are awkward and hard to read. Use
    the block forms instead.

### DIFF
--- a/consoles/s3270.pm
+++ b/consoles/s3270.pm
@@ -9,6 +9,7 @@ use feature 'say';
 use Data::Dumper 'Dumper';
 use YAML::PP 'Dump';
 use Carp qw(confess cluck carp croak);
+use List::Util 'any';
 require IPC::Run;
 use IPC::Run::Debug;    # set IPCRUNDEBUG=data in shell environment for trace
 use Thread::Queue;
@@ -47,7 +48,7 @@ sub finish ($self) { IPC::Run::finish($self->{connection}) }
 sub send_3270 ($self, $command = '', %arg) {
     $arg{command_status} //= 'ok';
     confess "command_status must be 'ok' or 'error' or 'any', got $arg{command_status}."
-      unless (grep $arg{command_status}, ['ok', 'error', 'any']);
+      unless (any { $arg{command_status} eq $_ } qw(ok error any));
 
     $self->{in} .= $command . "\n";
     $self->{connection}->IPC::Run::pump until $self->{out} =~ /^(ok|error)/mg;
@@ -98,7 +99,7 @@ sub _handle_expect_3270_cycle ($self, $result, $start_time, %arg) {
         my $input_line = pop @$co;
         my @output_area = @$co;
 
-        @output_area = grep !/$arg{delete_lines}/, @output_area if defined $arg{delete_lines};
+        @output_area = grep { !/$arg{delete_lines}/ } @output_area if defined $arg{delete_lines};
 
         if (@output_area > 0) {
             $self->{raw_expect_queue}->enqueue(@output_area);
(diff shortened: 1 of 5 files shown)


okurz (approved):

Martchus:
Looks like there are still syntax errors.

consoles/s3270.pm
```diff
@@ -47,7 +47,7 @@ sub finish ($self) { IPC::Run::finish($self->{connection}) }
 sub send_3270 ($self, $command = '', %arg) {
     $arg{command_status} //= 'ok';
     confess "command_status must be 'ok' or 'error' or 'any', got $arg{command_status}."
-      unless (grep $arg{command_status}, ['ok', 'error', 'any']);
+      unless (grep { $arg{command_status} eq $_ } qw(ok error any));
```
perlpunk: Interesting. The previous code did not even work as intended, it would always return true.

Martchus: This should probably use `any`:
```suggestion
      unless (any { $arg{command_status} eq $_ } qw(ok error any));
```

Martchus (dismissed):

okurz (approved):

Martchus (approved):

t/39-dewebsockify.t
```diff
@@ -137,7 +137,7 @@ subtest 'WebSocket handshake fails (no HTTP code)' => sub {
     mock_build_ws_tx($mock_ua, 'MockTxFailNoCode');
     start_dewebsockify();
     accept_client('dummy_socket_1');
-    ok grep(/Unable to upgrade to WebSocket connection/, @log_messages),
+    ok grep { /Unable to upgrade to WebSocket connection/ } @log_messages,
```
perlpunk: I know that's not obvious, but `grep` considers everything after the block as arguments, so the test label will end up as an argument for grep, and the test will not have a label anymore.
It works because for the logic of that test it doesn't make a difference.
```suggestion
    ok +(grep { /Unable to upgrade to WebSocket connection/ } @log_messages),
```
Same for all changes below as well.

perlpunk (changes requested):

perlpunk (approved):

perlpunk:
Somehow https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2972/os-autoinst:os-autoinst-test doesn't proceed to build.
SLE_15_SP7_Backports os-autoinst:os-autoinst-devel-test / os-autoinst:os-autoinst-openvswitch-test is still reported as unresolvable.

Martchus:
> SLE_15_SP7_Backports os-autoinst:os-autoinst-devel-test / os-autoinst:os-autoinst-openvswitch-test is still reported as unresolvable.

That the 1st is unsolvable is expected because we don't build the devel package for SLE-15-SP7 anymore. It is a while like that and this didn't prevent automatic merging.

That the 2nd is unsolvable is strange. It is due to:

> have choice for openvswitch needed by os-autoinst-openvswitch: openvswitch openvswitch3, have choice for openvswitch-switch needed by os-autoinst-openvswitch: openvswitch openvswitch3

However, our project config is supposed to prevent that: https://build.opensuse.org/projects/devel:openQA:GitHub/prjconf

EDIT: I've adjusted the project config in the PR project, see https://build.opensuse.org/projects/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2972/prjconf. We already have that in the parent project so it probably won't help but it is worth trying.

Wabri:
> EDIT: I've adjusted the project config in the PR project, see https://build.opensuse.org/projects/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-2972/prjconf. We already have that in the parent project so it probably won't help but it is worth trying.

Thanks marius :pray:

PR #2974, opened 2026-06-10
style(perlcritic): Add Modules::ProhibitConditionalUseStatements

Wabri:
Replace `use <module>` with `require <module>` removing compile time imports

https://metacpan.org/pod/Perl::Critic::Policy::Modules::ProhibitConditionalUseStatements

> Modules included via "use" are loaded at compile-time. Placing
    conditional logic around the "use" statement has no effect on whether
    the module will be loaded. Doing so can also serve to confuse the reader
    as to the author's original intent.
>
> If you need to conditionally load a module you should be using "require"
    instead.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -34,6 +34,7 @@ use OpenQA::Exceptions;
 use Time::Seconds;
 use English -no_match_vars;
 use OpenQA::NamedIOSelect;
+use Data::Dumper;
 
 use constant FULL_SCREEN_SEARCH_FREQUENCY => $ENV{OS_AUTOINST_FULL_SCREEN_SEARCH_FREQUENCY} // 5;
 use constant FULL_UPDATE_REQUEST_FREQUENCY => $ENV{OS_AUTOINST_FULL_UPDATE_REQUEST_FREQUENCY} // 5;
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

okurz (approved):

backend/baseclass.pm
```diff
@@ -637,8 +637,8 @@ sub check_socket ($self, $fh, $write = undef) {
         }
     }
     else {
-        use Data::Dumper;
-        die 'no command in ' . Dumper($cmd);
+        require Data::Dumper;
```
Martchus: This should probably just go to the top.

backend/baseclass.pm
```diff
@@ -637,8 +637,8 @@ sub check_socket ($self, $fh, $write = undef) {
...
+        die 'no command in ' . Data::Dumper::Dumper($cmd);
```
Martchus: I don't think that change is needed.

Martchus (changes requested):

Martchus (approved):

backend/baseclass.pm
perlpunk: It would have been, if `use` had been replaced with `require`, because `require` doesn't import anything.

perlpunk (approved):

PR #2976, opened 2026-06-11
fix: Fix enrolling cert in UEFI firmeware vars for SecureBoot

Martchus:
* Use `virt-fw-vars … --enroll-cert … --add-db …` because just `--enroll-cert` is not enough
* Verified by cloning the job mention in https://progress.opensuse.org/issues/200156#note-23; without this change it fails due to SecureBoot; with it the test can execute (tested as far as the `start_install` module)

Related ticket: https://progress.opensuse.org/issues/200156

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -304,8 +304,8 @@ sub configure_pflash ($self, $vars) {
 
         my @secureboot_args = _make_secure_boot_configuration($vars->{UEFI_PFLASH_SECURE_BOOT});
         my @certs = split qr/;/, $vars->{UEFI_PFLASH_CERTS} // '';
-        my @cert_args = map { ('--enroll-cert' => $_) } shift @certs // ();
         my $uuid = '37adf63d-93fb-4af5-9901-12f8767d3841';    # fixed "well-known" UUID
+        my @cert_args = map { ('--enroll-cert' => $_, '--add-db', $uuid, $_) } shift @certs // ();
         push @cert_args, map { ('--add-db', $uuid, $_, '--add-kek', $uuid, $_) } @certs;
         my @res_args = _make_resolution_configuration($vars->{UEFI_PFLASH_RESOLUTION});
         if (@secureboot_args || @cert_args || @res_args) {
--- a/t/18-qemu.t
+++ b/t/18-qemu.t
@@ -544,21 +544,19 @@ subtest configure_pflash => sub {
     my @expected_base_args = ('virt-fw-vars', '-i', '/usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin', '-o', "$dir/ovmf-x86_64-ms-4m-vars-adjusted.bin");
     subtest 'UEFI_PFLASH_CERTS and UEFI_PFLASH_SECURE_BOOT' => sub {
         my @commands;
+        my @certs = map { "/certs/$_.crt" } qw(foo bar);
         $mock_proc->redefine(runcmd => sub { push @commands, [@_] });
         $vars{UEFI_PFLASH_VARS} = '/usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin';
-        $vars{UEFI_PFLASH_CERTS} = '/certs/foo.crt;/certs/bar.crt';
+        $vars{UEFI_PFLASH_CERTS} = join ';', @certs;
         $vars{UEFI_PFLASH_SECURE_BOOT} = 1;
         $proc->configure_pflash(\%vars);
-        my @expected_commands = ([@expected_base_args, '--set-true', 'SecureBootEnable', '--set-false', 'CustomMode', '--enroll-cert', '/certs/foo.crt']);
-        my @cert_args = splice @{$commands[0]}, scalar @{$expected_commands[0]};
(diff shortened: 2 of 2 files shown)


Wabri (approved):

okurz (approved):

foursixnine (approved):

Vogtinator:
I suspect it failed because the original -vars.bin had no PK enrolled?

okurz (changes requested):
s/firmeware/firmware/

Martchus:
> I suspect it failed because the original -vars.bin had no PK enrolled?

It failed because `--encroll-cert` only seems to add the key to the KEK but not the db. Probably that's also why your modified spec file uses `--add-key` in addition to `--enroll-cert`.

Vogtinator:
> > I suspect it failed because the original -vars.bin had no PK enrolled?
> 
> It failed because `--encroll-cert` only seems to add the key to the KEK but not the db. Probably that's also why your modified spec file uses `--add-key` in addition to `--enroll-cert`.

*`-add-db` I guess

Yeah, that's possible. IIRC the PK is also authorized to sign executables, but maybe not.

I wonder if there's a use-case for tests which have a specific PK or KEK or only enrolling DB is enough.

Martchus:
The code adds the keys to the KEK and the DB now which should be good enough for now. Extending this in the future as needed shouldn't be a problem.

OpenQA/Qemu/Proc.pm
```diff
@@ -304,8 +304,8 @@ sub configure_pflash ($self, $vars) {
 
         my @secureboot_args = _make_secure_boot_configuration($vars->{UEFI_PFLASH_SECURE_BOOT});
         my @certs = split qr/;/, $vars->{UEFI_PFLASH_CERTS} // '';
-        my @cert_args = map { ('--enroll-cert' => $_) } shift @certs // ();
         my $uuid = '37adf63d-93fb-4af5-9901-12f8767d3841';    # fixed "well-known" UUID
+        my @cert_args = map { ('--enroll-cert' => $_, '--add-db', $uuid, $_) } shift @certs // ();
```
Vogtinator, line 308: Actually, this may be missing `--no-microsoft --microsoft-kek none`. Without those, `--enroll-cert` may add MS keys.

Unfortunately using only `--add-db` for certs may break if an empty -vars is used as base, as that would be missing a PK which would disable SB IIRC.

Vogtinator (changes requested):

OpenQA/Qemu/Proc.pm
Martchus: It isn't *only* using `--add-db`. It uses `--enroll-cert` and `--add-db` together for the first cert (and `--add-kek` and `--add-db` for further certs as `--enroll-cert` can only be used once).

I saw that you are using `--no-microsoft --microsoft-kek none` in the spec file but haven't included them because removing certs from the base seems behavior we don't necessarily need and if so it should probably be enabled via yet another setting, e.g. `UEFI_PFLASH_NO_MICROSOFT_CERTS=1`.

okurz (approved):

OpenQA/Qemu/Proc.pm
Vogtinator: > `It isn't only using --add-db`

What I meant is that `--enroll-cert` may break some vars, but using only `--add-db` as alternative would not work either.

> I saw that you are using --no-microsoft --microsoft-kek none in the spec file but haven't included them because removing certs from the base seems behavior we don't necessarily need and if so it should probably be enabled via yet another setting, e.g. UEFI_PFLASH_NO_MICROSOFT_CERTS=1.

The base is empty, there are no certs. Without `--no-microsoft`, it adds them, which we don't want: It may silently fall back to booting with MS-signed shim, ignoring the SUSE or openSUSE certificate. So we do need this and IMO it should be the implicit default.

Vogtinator: > The base is empty, there are no certs. Without --no-microsoft, it adds them, which we don't want: It may silently fall back to booting with MS-signed shim.

Actually, it looks like the `uefi-opensuse` machine type is no longer used, so this is not tested at all anymore :-/

foursixnine: @Vogtinator just to understand what would be the use case for this particular scenario (where no ms certificates are present too)? We can always add them back

I guess that would be `openSUSE-UEFI-CA-Certificate-2048.crt` right?

Vogtinator: > @Vogtinator just to understand what would be the use case for this particular scenario (where no ms certificates are present too)? We can always add them back

Verifying that TW boots with just the openSUSE cert enrolled and Leap boots with just openSUSE and SLES certs enrolled.

> I guess that would be `openSUSE-UEFI-CA-Certificate-2048.crt` right?

Yup.

Martchus: Ok, then I'll also create another PR adding `--no-microsoft` by default when `UEFI_PFLASH_CERTS=…` is specified.

Martchus: https://github.com/os-autoinst/os-autoinst/pull/2977

PR #2978, opened 2026-06-12
refactor: avoid external testapi internally

okurz:
Motivation:
basetest.pm improperly used testapi.pm, which defines the external test API.
Internal modules should not rely on external-facing APIs to respect Design by
Contract principles.

Design Choices:
Move is_serial_terminal to the internal autotest.pm module. Update
basetest.pm to use autotest::is_serial_terminal and access $bmwqemu::vars
directly, making testapi delegate to it to preserve the public API.

Benefits:
Establishes a clean architectural boundary between internal code and public
test APIs, preventing circular dependencies and ensuring contract boundaries.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -10,7 +10,6 @@ use Feature::Compat::Try;
 
 use bmwqemu ();
 use ocr;
-use testapi ();
 use autotest ();
 use MIME::Base64 'decode_base64';
 use OpenQA::Exceptions;
@@ -435,9 +434,10 @@ sub record_resultfile ($self, $title, $output, %nargs) {
 sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $string //= '';
     # take screenshot for documentation (screenshot does not represent fail itself)
-    $self->take_screenshot() unless (testapi::is_serial_terminal);
+    $self->take_screenshot() unless (autotest::is_serial_terminal);
 
-    my $pretty = $testapi::distri ? $testapi::distri->get_pretty_serial_marker() : (testapi::get_var('PRETTY_SERIAL_MARKER') || testapi::get_var('HIDE_MARKER_EVALUATION'));
+    my $distri = bmwqemu::distribution();
+    my $pretty = $distri ? $distri->get_pretty_serial_marker() : ($bmwqemu::vars{PRETTY_SERIAL_MARKER} || $bmwqemu::vars{HIDE_MARKER_EVALUATION});
     my $internal = $args{internal_marker};
     my $output_string = $string;
     my $captured_val;
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -216,10 +216,19 @@ sub ensure_valid_vars () {
 # local vars
 
 our $backend;    ## no critic (Variables::ProhibitPackageVars)
+my $distribution_object;
 
 # local vars end
 
 # util and helper functions
+
+sub distribution () {
+    return $distribution_object;
+}
+
+sub _set_distribution ($distri) {
+    $distribution_object = $distri;
+}
 
 sub current_test () {
     require autotest;
--- a/testapi.pm
+++ b/testapi.pm
@@ -75,7 +75,14 @@ our @EXPORT_OK = qw(is_serial_terminal);
 
 our %cmd;    ## no critic (Variables::ProhibitPackageVars)
 
+package testapi::tieddistri {
+    sub TIESCALAR ($class) { bless {}, $class }
+    sub FETCH ($self) { bmwqemu::distribution() }
+    sub STORE ($self, $value) { bmwqemu::_set_distribution($value) }
+}
+
 our $distri;    ## no critic (Variables::ProhibitPackageVars)
+tie $distri, 'testapi::tieddistri';
 
 our $realname = 'Bernhard M. Wiedemann';    ## no critic (Variables::ProhibitPackageVars)
 our $username;    ## no critic (Variables::ProhibitPackageVars)
@@ -816,15 +823,7 @@ For more info see consoles/virtio_console.pm and consoles/serial_screen.pm.
 
 =cut
 
-sub is_serial_terminal () {
-    state $ret;
-    state $last_seen = '';
-    if (defined current_console() && current_console() ne $last_seen) {
-        $last_seen = current_console();
-        $ret = query_isotovideo('backend_is_serial_terminal', {});
-    }
-    return $ret->{yesorno};
-}
+sub is_serial_terminal () { autotest::is_serial_terminal }
 
 
 =head2 wait_serial
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -85,7 +85,7 @@ sub prepare ($self) {
 
 sub load_schedule ($self) {
     # set a default distribution if the tests don't have one
-    $testapi::distri = distribution->new;
+    bmwqemu::_set_distribution(distribution->new);
 
     load_test_schedule;
 }
--- a/autotest.pm
+++ b/autotest.pm
@@ -324,8 +324,9 @@ sub loadtest ($script, %args) {
     $test = $name->new($category);
     $test->{script} = $script;
     $test->{fullname} = $fullname;
-    $test->{serial_failures} = $testapi::distri->{serial_failures} // [];
-    $test->{autoinst_failures} = $testapi::distri->{autoinst_failures} // [];
+    my $distri = bmwqemu::distribution();
+    $test->{serial_failures} = $distri ? $distri->{serial_failures} // [] : [];
+    $test->{autoinst_failures} = $distri ? $distri->{autoinst_failures} // [] : [];
 
     if (defined $args{run_args}) {
         unless (blessed($args{run_args}) && $args{run_args}->isa('OpenQA::Test::RunArgs')) {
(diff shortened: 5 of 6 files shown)


testapi.pm
```diff
@@ -816,15 +816,7 @@ For more info see consoles/virtio_console.pm and consoles/serial_screen.pm.
...
-sub is_serial_terminal () {
-    state $ret;
-    state $last_seen = '';
-    if (defined current_console() && current_console() ne $last_seen) {
-        $last_seen = current_console();
-        $ret = query_isotovideo('backend_is_serial_terminal', {});
-    }
-    return $ret->{yesorno};
-}
+sub is_serial_terminal () { autotest::is_serial_terminal }
```
Martchus, line 826: I don't think we should introduce another level of indirection for the functions in `testapi.pm` so we shouldn't start doing this here.

basetest.pm
```diff
@@ -435,9 +434,9 @@ sub record_resultfile ($self, $title, $output, %nargs) {
 sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $string //= '';
     # take screenshot for documentation (screenshot does not represent fail itself)
-    $self->take_screenshot() unless (testapi::is_serial_terminal);
+    $self->take_screenshot() unless (autotest::is_serial_terminal);
```
Martchus, line 437: What is wrong with calling this function also internally? It would be wrong to call an internal function externally but calling an external function internally is perfectly fine.

For `testapi::get_var` I get it: This function is just a wrapper around hash access and using it internally is quite useless. So here not using it reduces complexity by avoiding a useless level of indirection.

For `is_serial_terminal` this doesn't apply. In fact, your version makes the code just slightly more complex for zero practical benefit.

Martchus (changes requested):

basetest.pm
Martchus: The commit message mentions "Design by contract" but this change does not introduce use of "contracts" (as in the contracts feature recently introduced in C++). Note that contracts can be useful for internal and external functions so I'm wondering why you mix up a change like this PR with contracts.

basetest.pm
```diff
@@ -435,9 +434,9 @@ sub record_resultfile ($self, $title, $output, %nargs) {
...
 
-    my $pretty = $testapi::distri ? $testapi::distri->get_pretty_serial_marker() : (testapi::get_var('PRETTY_SERIAL_MARKER') || testapi::get_var('HIDE_MARKER_EVALUATION'));
+    my $pretty = $testapi::distri ? $testapi::distri->get_pretty_serial_marker() : ($bmwqemu::vars{PRETTY_SERIAL_MARKER} || $bmwqemu::vars{HIDE_MARKER_EVALUATION});
```
perlpunk: This still relies on `$testapi::distri`, which you only get if you `use testapi` (or if it is used indirectly).
So removing `use testapi` above hides that.
Ideally we would find a way to also remove this kind of dependency.

perlpunk (changes requested):

basetest.pm
perlpunk: > Ideally we would find a way to also remove this kind of dependency.

We could do this via a tied scalar (same as we did with `bmwqemu::vars`)

okurz: ok, done

testapi.pm
okurz: so what do you suggest instead?

basetest.pm
okurz: https://en.wikipedia.org/wiki/Design_by_contract

bmwqemu.pm
```diff
@@ -216,6 +216,7 @@ sub ensure_valid_vars () {
 # local vars
 
 our $backend;    ## no critic (Variables::ProhibitPackageVars)
+our $distri;    ## no critic (Variables::ProhibitPackageVars)
```
perlpunk: I think it's a bad idea to turn one global variable into a tied scalar that just uses another global variable in another package.
If we do this, it should be a function call around a lexical my variable, not a global variable that can be changed from everywhere

perlpunk (changes requested):

bmwqemu.pm
okurz: alright, so turned into a "lexical my variable" with getter+setter

basetest.pm
Martchus: Have you read the article? It goes in the same direction as the mentioned C++ contracts (and even has a C++ example that looks exactly like what I had in mind). So again, your change does not help with that. Where in the article does it say that external/public functions should not be called from internal implementations? What would help (in the sense that I understand contracts and the article describes them): adding additional assertions regardless

And that can help with correctness regardless of whether a function is public or private and from where it is called.

The article also interestingly states:

> Contracts can be written by [code comments](https://en.wikipedia.org/wiki/Comment_(computer_programming)), enforced by a [test suite](https://en.wikipedia.org/wiki/Test_suite), or both, even if there is no special language support for contracts.

(Although I'm wondering how a testsuite can "enforce" contracts.)

testapi.pm
Martchus: I suggest we keep calling the external function from internal implementations. There's nothing wrong with that here.

bmwqemu.pm
```diff
@@ -216,11 +216,20 @@ sub ensure_valid_vars () {
...
 
 # local vars end
 
 # util and helper functions
 
+sub distribution () {
+    return $distribution_object;
+}
+
+sub _set_distribution ($distri) {
```
Martchus, line 229: Looks like you want to call this function other modules. Then don't make it private:


```suggestion
sub set_distribution ($distri) {
```

testapi.pm
```diff
@@ -75,7 +75,14 @@ our @EXPORT_OK = qw(is_serial_terminal);
...
 our %cmd;    ## no critic (Variables::ProhibitPackageVars)
 
+package testapi::tieddistri {
+    sub TIESCALAR ($class) { bless {}, $class }
+    sub FETCH ($self) { bmwqemu::distribution() }
+    sub STORE ($self, $value) { bmwqemu::_set_distribution($value) }
+}
+
 our $distri;    ## no critic (Variables::ProhibitPackageVars)
+tie $distri, 'testapi::tieddistri';
```
Martchus, line 85: I'm wondering whether that's really better design. We effectively still have a global distribution object with that change but now two ways of accessing it. That makes the code more complicated and in my opinion also harder to follow, especially if one is not that familiar with Perl and how `tie` works.

@perlpunk What do you think?

Martchus (changes requested):

basetest.pm
okurz: > Have you read the article?

No, of course not. I just liked the words of the title and made up what it could mean and implemented "something"

Martchus: If this is an ironic comment: It would be more constructive to point out the relevant section in that article that I must have overlooked that says one should avoid calling external functions from internal implementations. (Of course even if that is stated there it would be questionable whether that makes sense.)

Martchus: And my question was really serious because the article seems to support my point of view and not the changes of this PR. So I'm really confused.

testapi.pm
```diff
@@ -75,7 +75,14 @@ our @EXPORT_OK = qw(is_serial_terminal);
 
 our %cmd;    ## no critic (Variables::ProhibitPackageVars)
 
+package testapi::tieddistri {
```
perlpunk, line 78: We could provide a `testapi::distri()` function calling `bmwqemu::distribution()`.
Accessing `$testapi::distri` could then be deprecated.

testapi.pm
perlpunk: Yeah, we can't really remove the global `$testapi::distri` variable, it's legacy. But we would provide an access for tests (readonly) and an "inner" access for isotovideo.

testapi.pm
```diff
@@ -75,7 +75,14 @@ our @EXPORT_OK = qw(is_serial_terminal);
...
+    sub TIESCALAR ($class) { bless {}, $class }
+    sub FETCH ($self) { bmwqemu::distribution() }
+    sub STORE ($self, $value) { bmwqemu::_set_distribution($value) }
```
perlpunk, line 81: Do we actually need this? Do tests have to have the possibility to change the variable?

aylen384:
I share Marius' opinion here.
I think it is better to call the external api internally and even more so we should consistently do that and avoid e.g. accessing `$bmwqemu::vars` in more than a single place (which is the implementation of the external facing method). This would allow much easier refactoring and reduce duplication, indirection and spaghetti code. Also the code readability would benefit.

Martchus:
I guess it is fine to modify `$bmwqemu::vars` directly, though. Modifying it directly means we don't have additional function call overhead, copy overhead and we can do efficient in-place modifications if needed (like `$bmwqemu::vars{foo} //= …` or `$bmwqemu::vars{foo} =~ s/…/…/` or even `substr($bmwqemu::vars{foo}, …) = ….`).

So I think there's some middle ground. If we can benefit from direct access of internal data structures we should do that from internal code. Otherwise it makes sense to use public APIs also internally.

perlpunk:
> I guess it is fine to modify `$bmwqemu::vars` directly, though. Modifying it directly means we don't have additional function call overhead, copy overhead and we can do efficient in-place modifications if needed

That's an illusion though because it is a tied hash now and will call methods instead :)

Martchus:
Oh, right. I forgot that we changed that.

PR #2983, opened 2026-06-13
fix: handle existing pipes gracefully

okurz:
* fix: cleanly recreate virtio console pipe if already exists
* fix: suppress spurious virtio console unlink warnings

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -589,7 +589,7 @@ sub virtio_console_names () {
 sub virtio_console_fifo_names () { map { $_ . '.in', $_ . '.out' } virtio_console_names }
 
 sub console_fifo ($name) {
-    return bmwqemu::fctwarn("Fifo pipe '$name' already exists!") if -e $name;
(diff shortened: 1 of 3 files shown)


Wabri (approved):

backend/qemu.pm
```diff
@@ -589,13 +589,17 @@ sub virtio_console_names () {
 sub virtio_console_fifo_names () { map { $_ . '.in', $_ . '.out' } virtio_console_names }
 
 sub console_fifo ($name) {
-    return bmwqemu::fctwarn("Fifo pipe '$name' already exists!") if -e $name;
+    if (-e $name) {
+        unlink $name or $!{ENOENT} or bmwqemu::fctwarn("Could not unlink existing pipe $name: $!");
+    }
```
Martchus: With `or $!{ENOENT}` we can avoid  the additional `-e $name` check (which is racy anyway):
```suggestion
    unlink $name or $!{ENOENT} or bmwqemu::fctwarn("Could not unlink existing pipe $name: $!");
```

Martchus (approved):
The change looks generally good but some tests failed.

PR #2977, opened 2026-06-12
feat: Avoid silent fallback to MS certs with `UEFI_PFLASH_CERTS`

Martchus:
* Remove Microsoft certificates when certificates are specified explicitly via `UEFI_PFLASH_CERTS` to prevent tests from silently falling back to using them
* See https://github.com/os-autoinst/os-autoinst/pull/2976#discussion_r3401490825

Related ticket: https://progress.opensuse.org/issues/200156

PR #2985, opened 2026-06-15
fix: Fix comment about OVMF firmware locations

Martchus:
The comment was invalidated when `/usr/share/qemu/ovmf-x86_64-ms-4m-code.bin` was added. This change uses inline-comments to avoid this problem.

### DIFF
--- a/bmwqemu.pm
+++ b/bmwqemu.pm
@@ -40,16 +40,16 @@ our $screenshotpath = 'qemuscreenshot';    ## no critic (Variables::ProhibitPack
 
 # global vars
 
-# Known locations of OVMF (UEFI) firmware: first is openSUSE, second is
-# the kraxel.org nightly packages, third is Fedora's edk2-ovmf package,
-# fourth is Debian's ovmf package.
 our @ovmf_locations_no_secure_boot = (    ## no critic (Variables::ProhibitPackageVars)
-    '/usr/share/qemu/ovmf-x86_64-4m-code.bin', '/usr/share/qemu/ovmf-x86_64-ms-code.bin',
+    '/usr/share/qemu/ovmf-x86_64-4m-code.bin',    # openSUSE qemu-ovmf package
+    '/usr/share/qemu/ovmf-x86_64-ms-code.bin',    # openSUSE qemu-ovmf package
 );
 our @ovmf_locations = (    ## no critic (Variables::ProhibitPackageVars)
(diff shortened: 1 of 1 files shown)


okurz (approved):

Wabri (approved):

Martchus:
6 checks are missing according to mergify as only 22 are successful but >=28 successful checks are required. It looks the same on other open PRs. I'm not sure what's missing here, though. One of the last automatically merged PRs is https://github.com/os-autoinst/os-autoinst/pull/2977 and it was merged with only 23 checks passed.

This one from May has 37 checks: https://github.com/os-autoinst/os-autoinst/pull/2926

We have of course removed certain package builds since then but it is strange that auto-merging still worked a few days ago as we haven't changed anything since then.

I'll merge this PR manually for now.

perlpunk:
> 6 checks are missing according to mergify as only 22 are successful but >=28 successful checks are required. It looks the same on other open PRs. I'm not sure what's missing here, though.

I commented here:
https://github.com/os-autoinst/os-autoinst/pull/2972#issuecomment-4718973058

So packages in OBS are still unresolvable, and it was the same here as well, I assume. You can always click on the OBS links to check what's missing.

Martchus:
Really? I checked OBS and there was only one unresolvable package but that one is expected. Otherwise all builds were successful. The OBS checks are also present on GitHub (which you can still see via "View details").

And now after I merged manually it also says "23 checks passed" (and not just 22). So this is actually identical to https://github.com/os-autoinst/os-autoinst/pull/2977 which as still automatically merged.

EDIT: In that other PR there's actually one unresolvable package that's unexpected, see https://github.com/os-autoinst/os-autoinst/pull/2972#issuecomment-4720731592. Maybe I missed that here.

okurz:
> Really? I checked OBS and there was only one unresolvable package but that one is expected.

Why should any unresolvables be "expected"?

Martchus:
As stated in my comment on the other PR:

> That the 1st is unsolvable is expected because we don't build the devel package for SLE-15-SP7 anymore.

Maybe one can disable the specific `-test` package for that specific target to avoid this unresolvable build. When I remember correctly, this unresolvable build was factored into the count of expected successful checks so it is not the problem here. (And besides, we are missing 6 successful checks - although very weirdly PRs with just 23 checks have also been merged in the past - despite requiring 28 or more.)

PR #2989, opened 2026-06-17
test: reliably assert pipe size adjustments

okurz:
Motivation:
Unprivileged user limits on total pipe allocation (pipe-user-pages-soft) can
cause F_SETPIPE_SZ to fail with EPERM for large sizes like 640x480x3, leading
to sporadic test failures in multi-process or high-activity environments.

Design Choices:
Lowered the test resolution to 100x100 for the success subtest, reducing the
requested pipe size to 30,000 bytes. This fits within the default pipe
capacity and requires no new page allocations, avoiding any quota failures.

Benefits:
Ensures 100% deterministic test outcomes across all developer and CI
environments without requiring conditional branches, custom mock helpers, or
retries.


Wabri (approved):
I'm not sure about the change, but if you are sure and tested, I'm ok with it. We are still on time to revert it.

Martchus (approved):

PR #2990, opened 2026-06-17
fix: Avoid unresolvable os-autoinst-devel-test package for SLE-SP7

Martchus:
Related ticket: https://progress.opensuse.org/issues/202671

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -13,6 +13,8 @@ pr:
             paths:
               - target_project: openSUSE:Backports:SLE-15-SP7:Update
                 target_repository: standard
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

okurz (approved):

Martchus:
~~Looks like this didn't work.~~ I had the condition reversed. It now seems to work.

~~The 2nd unresolvable (due to the openvswitch choice) is ~~also~~ still present. So that's not temporarily but adding the project config of the PR sub project likely actually helped.~~ Fixed with the 2nd commit.

okurz (approved):

PR #2993, opened 2026-06-18
test: assert pipe size adjustment dynamically

okurz:
Motivation:
F_SETPIPE_SZ can fail with EPERM when the user's total pipe allocation
exceeds pipe-user-pages-soft. The prior attempt in 4cd87405 assumed the
default pipe capacity (65536 bytes) would accommodate 30KB without new
page allocations, but on systems where the default is only 8192 bytes
the resize still requires increasing the buffer count and is blocked.

Design Choices:
Probe F_SETPIPE_SZ support on a dummy pipe at subtest start. Adjusted
the assertion to expect unchanged default size when resizing is blocked.

Benefits:
Deterministic test outcomes regardless of system pipe defaults or
user-level pipe allocation limits.

### DIFF
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -730,6 +730,11 @@ subtest 'adjusting pipe size for external video encoder ' => sub {
         $bmwqemu::vars{XRES} = undef;
         $bmwqemu::vars{YRES} = undef;
     };
+    pipe my $dummy_r, my $dummy_w;
+    my $pipe_set_supported = fcntl $dummy_w, Fcntl::F_SETPIPE_SZ, 30000;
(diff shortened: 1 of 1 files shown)


Martchus (approved):

Martchus:
It weakens the test a little bit but I guess it is ok. The lines need to be marked as uncoverable or you somehow merge them into one statement.

okurz:
> It weakens the test a little bit but I guess it is ok. The lines need to be marked as uncoverable or you somehow merge them into one statement.

amended with a combined statement

aylen384 (approved):

PR #3001, opened 2026-06-26
Implement SPICE backend support

okurz:
* test: add parallel-safe full-stack SPICE wrapper
* fix: correct shifted zero mapping on SPICE console
* fixup! feat: fix key input and event loop on SPICE
* feat: fix key input and event loop on SPICE
* ci: configure Dependabot for Rust crate updates
* feat: integrate spice-client-glib via C FFI bindings
* build: integrate Cargo build and Rust lint in CMake
* fix: manage SPICE daemon lifecycle and build checks
* feat: implement consoles::SPICE client in Perl
* feat: implement SPICE inputs and input routing
* feat: implement SPICE handshake and frame capture
* feat: scaffold Rust spice-bridge and IPC
* feat: support spice-based VM interaction in QEMU

Additionally this needs the following addition to `_service` in the OBS project:
```
  <service name="cargo_vendor" mode="buildtime">
    <param name="srcdir">os-autoinst</param>
    <param name="compression">zst</param>
    <param name="update">yes</param>
  </service>
```

I also applied that manually temporarily to https://build.opensuse.org/projects/devel:openQA:GitHub:os-autoinst:os-autoinst:PR-3001/packages/os-autoinst/files/_service?expand=1

And for additional dependencies for SLE15: `osc linkpac openSUSE:Factory cargo-vendor-filterer devel:openQA:Leap:15.7`


cmake/test-targets.cmake
```diff
@@ -138,6 +138,22 @@ else ()
     message(STATUS "Set GITLINT_PATH to the path of the gitlint executable to enable git commit message checks.")
 endif ()
 
+# add test for Rust code style and clippy
+find_program(RUSTFMT_PATH rustfmt)
+find_program(CARGO_BIN cargo)
```
Martchus, line 143: For the sake of consistency:
```suggestion
find_program(CARGO_PATH cargo)
```

cmake/test-targets.cmake
```diff
@@ -138,6 +138,22 @@ else ()
...
+if (RUSTFMT_PATH AND CARGO_BIN)
```
Martchus, line 144: `RUSTFMT_PATH` is written but never read/used in a meaningful way. That's not ideal because the lookup via `find_program` might differ from what `cargo` uses (e.g. one can override it completely via `-DRUSTFMT_PATH=…`). This code should probably invoke `rustfmt` directly if possible.

cmake/test-targets.cmake
```diff
@@ -138,6 +138,22 @@ else ()
...
+    add_test(
+        NAME test-local-rust-style
+        COMMAND "${CARGO_BIN}" fmt --manifest-path "${CMAKE_CURRENT_SOURCE_DIR}/rust/spice-bridge/Cargo.toml" --check
+        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+    )
+    add_test(
+        NAME test-local-rust-clippy
+        COMMAND "${CARGO_BIN}" clippy --manifest-path "${CMAKE_CURRENT_SOURCE_DIR}/rust/spice-bridge/Cargo.toml" -- -D warnings
```
Martchus, line 152: So `clippy` is always available (in contrast to `fmt`)?

rust/spice-bridge/src/spice.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub width: i32,
+    pub height: i32,
+    pub stride: i32,
+    pub shmid: i32,
+    pub data: *mut u8,
+    pub marked: i32,
+}
+
+fn extract_frame(channel: *mut SpiceDisplayChannel, user_data: gpointer) {
+    unsafe {
```
Martchus, line 320: Considering these huge `unsafe` blocks I'm wondering whether it is worthwhile introducing Rust here. It is probably better to simply extend our existing tinycv library. Maybe there's a way to reduce the amount of unsafe code, though.

rust/spice-bridge/src/spice.rs
```diff
@@ -0,0 +1,426 @@
...
+        let mut primary: LocalSpiceDisplayPrimary = std::mem::zeroed();
+        let success = spice_display_get_primary(
+            channel as *mut SpiceChannel,
+            0,
+            &mut primary as *mut _ as *mut _,
+        );
+
+        if success != 0 && !primary.data.is_null() {
+            let ctx = &*(user_data as *const SpiceContext);
+            let mut fb = ctx.fb.lock().unwrap();
```
Martchus, line 330: If we go for Rust we should generally care about correctness and proper error handling. So use proper error handling instead of sprinkling `.unwrap()` calls all over the place.

rust/spice-bridge/src/spice.rs
```diff
@@ -0,0 +1,426 @@
...
+                        "Unsupported SPICE primary surface format: {}",
+                        primary.format
+                    );
+                }
+            }
+        }
+    }
+}
+
+#[allow(dead_code)]
```
Martchus, line 405: Annotations like this shouldn't be added without an explanatory comment.

rust/spice-bridge/tests/ipc_test.rs
```diff
@@ -0,0 +1,148 @@
...
+            let _ = fs::remove_file(path);
+        }
+    }
+}
+
+#[test]
+fn test_spice_bridge_ipc() {
+    let socket_path = "/tmp/test-rust-spice-bridge-full.sock";
+
+    let binary_path = "../../target/debug/spice-bridge";
```
Martchus, line 29: Don't hardcode the target name.

rust/spice-bridge/tests/ipc_test.rs
```diff
@@ -0,0 +1,148 @@
...
+        let path = Path::new(&self.socket_path);
+        if path.exists() {
+            let _ = fs::remove_file(path);
+        }
+    }
+}
+
+#[test]
+fn test_spice_bridge_ipc() {
+    let socket_path = "/tmp/test-rust-spice-bridge-full.sock";
```
Martchus, line 27: Don't hardcode the resolution of the temporary dir.

CMakeLists.txt
```diff
@@ -159,6 +161,19 @@ add_custom_target(
 )
 add_dependencies(symlinks videoencoder snd2png debugviewer tinycv)
 
+# Build the SPICE bridge daemon using Cargo
+find_program(CARGO_BIN cargo)
```
Martchus, line 165: That's the 2nd call of this kind so the code should probably be de-duplicated.

dependencies.yaml
```diff
@@ -74,6 +74,9 @@ build_base_requires:
   pkgconfig(sndfile):
   pkgconfig(theoraenc):
   '%opencv_require':
+  cargo:
+  cargo-packaging:
+  pkgconfig(spice-client-glib-2.0):
```
Martchus, line 79: I'm not sure whether it makes sense to add these dependencies unconditionally. Did you check whether all of our usual build targets have them in the required version?

backend/qemu.pm
```diff
@@ -1062,21 +1072,59 @@ sub start_qemu ($self) {
...
-            description => q{QEMU's VNC}});
+    my $console;
+    if ($vars->{QEMU_SPICE}) {
+        my $spiceport = $vars->{SPICE_PORT} // (5900 + ($vars->{VNC} // 0));
+        my $socket_path = "/tmp/spice-bridge-$spiceport.sock";
+
+        my $bridge_bin = find_bin("$FindBin::Bin/../rust/spice-bridge/target/debug", 'spice-bridge')
+          // find_bin("$FindBin::Bin/../rust/spice-bridge/target/release", 'spice-bridge')
+          // find_bin("$FindBin::Bin/rust/spice-bridge/target/debug", 'spice-bridge')
+          // find_bin("$FindBin::Bin/rust/spice-bridge/target/release", 'spice-bridge')
```
Martchus, line 1083: Don't hardcode or at least don't repeat the target names.

backend/qemu.pm
```diff
@@ -1062,21 +1072,59 @@ sub start_qemu ($self) {
...
-        'vnc-base',
-        {
-            hostname => 'localhost',
-            connect_timeout => 3,
-            port => $vncport,
-            description => q{QEMU's VNC}});
+    my $console;
+    if ($vars->{QEMU_SPICE}) {
+        my $spiceport = $vars->{SPICE_PORT} // (5900 + ($vars->{VNC} // 0));
+        my $socket_path = "/tmp/spice-bridge-$spiceport.sock";
```
Martchus, line 1078: Don't hardcode `/tmp`.

t/99-full-stack-spice.t
```diff
@@ -0,0 +1,25 @@
...
+$ENV{OS_AUTOINST_SPICE} = 1;
+
+# Evaluate and execute the original full stack test
+try {
+    local $SIG{__DIE__} = undef;
+    do "$Bin/99-full-stack.t";
+}
+catch ($e) {
+    die $e;
+}
```
Martchus, line 25: This test would pass without any of the other changes. Add checks to verify that the spice backend was actually used.

Martchus (changes requested):
It looks like this introduces an additional executable that communicates with the backend executable. It would make sense to extend the architecture documentation accordingly. It would also make sense to generally explain how the Rust-code ties in with the rest of the code base.

It looks like the commit history should be simplified. A lot of code is introduced in some commits that's just changed completely again anyway.

Martchus:
It is also not so nice that we'll have one more process with this. I'm wondering whether it was possible to simply handle the SPICE protocol directly within the backend process like we do with the VNC protocol. That would be simpler and hence probably better.

okurz (commented):
Thanks for the review. I addressed the following points:
1. **Commit History**: I have squashed and completely rebased the commit history into 3 clean, logical commits (Feature implementation, Test wrapper, CI config).
2. **Architecture Docs**: extended `doc/architecture.md` to explicitly mention the `spice-bridge` Rust daemon, its purpose, and how it ties into the rest of the codebase via Unix sockets.
3. **One More Process**: Unlike VNC, which is a relatively simple RFB protocol, SPICE seems to utilize multiple independent async channels (Main, Inputs, Display, Cursor, etc.) and relies on the GLib Main Event Loop. Implementing this natively in Perl (like VNC) would likely require massive effort and risk blocking the single-threaded backend process. By using an external Rust daemon (`spice-bridge`), we can safely link against `spice-client-glib-2.0` (using FFI) and run the GLib main loop in its own isolated process. The bridge then exposes a simple, lightweight JSON-over-Unix-socket IPC mechanism for the Perl backend to consume, keeping the core os-autoinst architecture lean and avoiding C/C++ thread-safety issues.
4. **Coverage**: Added coverage what codecov reported as missing
5. **OBS Build Configuration**: In the process to cover the OBS `cargo_vendor` workflow

other points still open. Will look later

AdamWill:
I feel like it might be nice to have a bit better explanation of the "why" here. The only thing I can find is one sentence in the first commit message: "SPICE offers superior video and input capabilities compared to VNC." That's a bit thin. What superior capabilities? What are the concrete benefits of this compared to VNC?

Martchus:
> One More Process: Unlike VNC, which is a relatively simple RFB protocol, SPICE seems to utilize multiple independent async channels (Main, Inputs, Display, Cursor, etc.) and relies on the GLib Main Event Loop. Implementing this natively in Perl (like VNC) would likely require massive effort and risk blocking the single-threaded backend process.

I haven't asked for a native implementation in Perl. We discussed this before and I always said we should *not* implement this Perl. We can still implement it in Rust and GLib using multiple threads - without the need for multiple processes. We already do the same with C++ and OpenCV.

However, I guess I'm just writing with an AI here and not with @okurz. Too bad. (That's also why I initially didn't reply at all. If you just give my comments some AI to let it generate a response then that's the end of a real discussion.)

okurz:
> However, I guess I'm just writing with an AI here and not with @okurz. Too bad. (That's also why I initially didn't reply at all. If you just give my comments some AI to let it generate a response then that's the end of a real discussion.)

Yes, same as in every daily meeting you are unfortunately not able to talk to my real self but only an AI assistant.

Martchus:
I'm not complaining about you not being approachable. I'm complaining about having to read clearly AI generated comments that don't consider context from prior "real" discussions and read like generic "bla bla" instead of addressing what I said in my previous comment. You *can* use AI but it would make sense to still consider what of the answer actually applies and cut the "bla bla".

consoles/SPICE.pm
```diff
@@ -0,0 +1,295 @@
...
+        my $ppm_header = 'P6' . "\n" . $self->width . ' ' . $self->height . "\n" . '255' . "\n";
+        my $image = tinycv::from_ppm($ppm_header . $raw_data);
+        $self->_framebuffer($image);
+    }
+}
+
+sub map_and_send_key ($self, $keys, $down, $delay) {
+    my $socket = $self->socket or die 'No socket available';
+
+    my %scancodes = (
```
mimi1vx, line 100: The 100+ line scancode + shift-map hashes are rebuilt on every map_and_send_key call.

rust/spice-bridge/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+                    eprintln!("Failed to write reply: {}", e);
+                    break;
+                }
+                if let Err(e) = writer.write_all(&data) {
+                    eprintln!("Failed to write raw frame bytes: {}", e);
+                    break;
+                }
+            }
+            Ok(Command::KeyEvent { key, down }) => {
+                let scancode = key.parse::<u32>().unwrap_or(0x1c); // default to Enter if not numeric
```
mimi1vx, line 105: injection Enter ? why ?

rust/spice-bridge/src/spice.rs
```diff
@@ -0,0 +1,426 @@
...
+            let h = primary.height as usize;
+            let stride = primary.stride as usize;
+
+            let required_len = w * h * 3;
+            if fb.data.len() != required_len {
+                fb.data.resize(required_len, 0);
+            }
+
+            match primary.format {
+                32 | 96 => {
```
mimi1vx, line 344: magic numbers ?

rust/spice-bridge/src/spice.rs
```diff
@@ -0,0 +1,426 @@
...
+        let info = Box::from_raw(raw);
+        spice_inputs_channel_position(
+            info.channel,
+            info.x as gint,
+            info.y as gint,
+            0,
+            info.button_mask as gint,
+        );
+        if info.button_mask > 0 {
+            spice_inputs_channel_button_press(info.channel, 1, 1);
```
mimi1vx, line 74: Right/middle clicks and button releases won't work.

consoles/SPICE.pm
```diff
@@ -0,0 +1,295 @@
...
+        for my $sc (@list) {
+            my $final_sc = $state ? $sc : ($sc | 0x80);
+            my $cmd = encode_json({
+                    cmd => 'key_event',
+                    key => "$final_sc",
+                    down => $state ? Mojo::JSON->true : Mojo::JSON->false,
+            }) . "\n";
+            $socket->print($cmd);
+            # Read status reply
+            my $resp = $socket->getline();
```
mimi1vx, line 261: Key up/down replies are read but discarded, and no timeout. my $resp = $socket->getline(); blocks indefinitely if the bridge stalls;

t/27-consoles-spice.t
```diff
@@ -0,0 +1,188 @@
...
+    $mock_socket->set_always('getline', "{\"status\":\"ok\"}\n");
+    $client->socket($mock_socket);
+
+    stderr_like { $client->map_and_send_key('unknown_key', 1, 0.1) } qr/Unknown key mapping for SPICE/, 'unknown key warns';
+    lives_ok { $client->map_and_send_key('A', 1, 0.1) } 'uppercase letter works';
+    lives_ok { $client->map_and_send_key('!', 1, 0.1) } 'shift symbol works';
+    lives_ok { $client->map_and_send_key('1-2', undef, 0.1) } 'number works';
+    lives_ok { $client->map_and_send_key('a-b', 0, 0.001) } 'multiple keys with small delay';
+};
+done_testing();
```
mimi1vx, line 159: isnt this call too early ? next test is never run

consoles/spice_base.pm
```diff
@@ -0,0 +1,92 @@
...
+use Mojo::Base 'consoles::video_base', -signatures;
+use consoles::SPICE;
+use Time::HiRes qw(usleep);
+
+use bmwqemu ();
+
+use constant SPICE_TYPING_LIMIT_DEFAULT => consoles::video_base::TYPING_LIMIT_DEFAULT;
+
+sub disable ($self) {
+    if ($self->{vnc}) {
```
mimi1vx, line 15: vnc ? in spice ? why ?

PR #3004, opened 2026-07-01
fix: Check also hidden files in checklist plugin

okurz:
include-hidden-files is false by default, so it would not complain about
e.g. .gitlint


perlpunk (changes requested):
Duplicate of #3003 ?

PR #3005, opened 2026-07-02
feat: optimize and shorten pretty serial hook

okurz:
Motivation:
Every character typed over VNC incurs latency and can lead to mistyping.
Optimizing hook names and removing whitespace speeds up test setup.

Design Choices:
Shortened variables and function names in bash hook (_oap, _OANM, _OAM).
Stripped redundant whitespaces around shell operators in the hook string.

Benefits:
Faster test setup phases and lower latency over VNC-based consoles.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -184,11 +184,13 @@ sub script_run ($self, $cmd, @args) {
         $str = testapi::hashed_string('SR' . $cmd . $args{timeout});
         $wait_pattern = qr/$str-(\d+)-/;
         if ($level == 2) {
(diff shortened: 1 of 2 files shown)


Martchus (commented):
It would maybe make sense to add a comment where these variables are introduced explaining what they mean. At least I find `_QAM` and `_QANM` otherwise a bit hard to distinguish.

okurz:
> It would maybe make sense to add a comment where these variables are introduced explaining what they mean. At least I find `_QAM` and `_QANM` otherwise a bit hard to distinguish.

good point. Added comments explaining that.

perlpunk (approved):

Martchus (approved):

PR #3009, opened 2026-07-07
style(perlcriticrc): migrate to blocklist approach

Wabri:
Switch .perlcriticrc from include to exclude approach to enable all core policy by default, similar to openqa configuration.

Related: https://progress.opensuse.org/issues/155191


okurz (changes requested):
+1 but s/blacklist/blocklist/

Wabri:
> +1 but s/blacklist/blocklist/

Ok, I'm also fixing the perlcritics warnings!

okurz:
Please bump the version number as noted in https://github.com/os-autoinst/os-autoinst/pull/3009#issuecomment-4906352295 . You are likely not changing functionally but maybe overriding functions behaves different after the introduction of signatures.

Wabri:
> Please bump the version number as noted in [#3009 (comment)](https://github.com/os-autoinst/os-autoinst/pull/3009#issuecomment-4906352295) . You are likely not changing functionally but maybe overriding functions behaves different after the introduction of signatures.

I've apply the bump in the commit https://github.com/os-autoinst/os-autoinst/pull/3009/changes/c278c6ab88b0c1572209c110a00651e451e88723

ppmclibs/tinycv.pm
```diff
@@ -13,7 +13,9 @@ require Exporter;
 require DynaLoader;
 
 our @ISA = qw(Exporter DynaLoader);
+## no critic (Modules::ProhibitAutomaticExportation)
 our @EXPORT = qw();
+## use critic
```
perlpunk: Just add the comment at the end of the line, most lines should be short enough for that:
```suggestion
our @EXPORT = qw();    ## no critic (Modules::ProhibitAutomaticExportation)
```

lockapi.pm
```diff
@@ -8,8 +8,10 @@ use Mojo::Base 'Exporter', -signatures;
 use Scalar::Util 'looks_like_number';
 use List::Util qw(min);
 use Time::Seconds;
+## no critic (Modules::ProhibitAutomaticExportation)
```
perlpunk, line 11: same as above, just add the comment on the line, same for the other occurrences

Wabri: It could not work because the automatic export will threat them as export and not comments, or am I just falling into the editor lsp trap?

Wabri: Just to be clear:
image

Wabri: The same apply on `testapi.pm`

Wabri: I've tested locally as you mention and the result are:
```
14: #   ::warning file=lockapi.pm,line=12,col=1,title=Symbols are exported by default - severity 4::[Modules::ProhibitAutomaticExportation] Use '@EXPORT_OK' or '%EXPORT_TAGS' instead
```

This is the change:

```
our @EXPORT = qw(mutex_create mutex_lock mutex_unlock mutex_try_lock mutex_wait ## no critic (Modules::ProhibitAutomaticExportation)
  barrier_create barrier_wait barrier_try_wait barrier_destroy);
```

Wabri: I've tested also with `testapi.pm` and same results

ppmclibs/tinycv.pm
Wabri: Updated locally, it will be also here in the next push

perlpunk: Actually in this case you can just remove the whole line, as the array is empty

script/os-autoinst-generate-needle-preview
```diff
@@ -46,10 +46,9 @@ sub process_file ($filename) {
     my $image_data = b64_encode($pngpath->slurp, '&#10;');
 
     open $SVGFILE, '>', path($pp->dirname, "${name}.svg");
-    select $SVGFILE;
 
     # The generated file should be identical to the svg file as saved with inkscape as 'Standard SVG'
-    printf <<~"EOM";
+    printf $SVGFILE <<~"EOM";
```
perlpunk: you would have to add that to all of the following `print` and `say` statements as well. maybe in this case it's better to keep the select

t/14-isotovideo.t
```diff
@@ -220,7 +220,7 @@ subtest 'isotovideo with wheels' => sub {
             is 0, checkout_wheels($case_dir, $wheels_dir), 'wheels checkout out';
             load_test_schedule;
             like $diags[0], qr/scheduling ink$i/, 'module from the wheel scheduled';
-            my $ret = eval "Copy::Writer::Content${i}::write()";
+            my $ret = eval { Copy::Writer::Content ${i} ::write() };
```
perlpunk: This will likely not work.
What could work (untested):

```suggestion
            my $ret = eval { "Copy::Writer::Content${i}"->can('write')->() };
```

autotest.pm
```diff
@@ -108,11 +108,10 @@ e.g. by making use of the openQA asset download feature.
 sub _debug_python_version () {
     state $python_loaded = 0;
     return if $python_loaded++;
-    my $code = <<~'EOM';
-    use Inline::Python;
-    return Inline::Python::py_eval('__import__("sys").version', 0);
-    EOM
-    my $debug = eval $code;
+    my $debug = eval {
+        use Inline::Python;
```
perlpunk: Every `use` statement is executed at compile time. (Unless in a string eval like before).

This should work:
```suggestion
        require Inline::Python;
```

bmwqemu.pm
```diff
@@ -371,13 +371,14 @@ sub STORE ($self, $key, $val) {
 
 sub FIRSTKEY ($self) {
     my $data = $self->{data};
-    my @k = keys %$data;    # reset
-    my $next = each %$data;
+    $self->{_keys} = [keys %$data];    # reset keys
+    $self->{_index} = 0;    # reset index
+    return $self->{_keys}[0];
```
perlpunk: Please revert (also below).
This is an implementation of a tied hash, so it needs to implement what `each` does, so it needs to use it.
Using our own implementation is just too much effort and fragile.

lockapi.pm
perlpunk: I see, right, this spans multiple lines.

perlpunk:
Maybe it would still be good to split this into multiple PRs?

Wabri:
> Maybe it would still be good to split this into multiple PRs?

Because it's better to review? I'm finding no problem on creating commit for every policy and fix them by interactively rebase.

okurz (approved):

autotest.pm
```diff
@@ -306,8 +305,8 @@ sub loadtest ($script, %args) {
     state %loaded;    # keep track of loaded packages
                       # note: Never load a test module that would result in the same package twice as this would only lead to warnings
                       #       like "Subroutine run redefined at …".
-    eval _make_test_code_to_eval($script_path, $script, $name, \$is_python) unless $loaded{$script_path}++;
-    if (my $err = $@) {
+    my $err = eval _make_test_code_to_eval($script_path, $script, $name, \$is_python) unless $loaded{$script_path}++; ## no critic (BuiltinFunctions::ProhibitStringyEval)
```
perlpunk: The return value of eval is just like the return value of normal code, it is not the same as `$@` and will not contain the eval error

Wabri: Fixed, thanks!

consoles/ssh_screen.pm
```diff
@@ -52,7 +52,7 @@ sub do_read {    # no:style:signatures
           unless $error_seen{$errcode}++;
 
         last if ($args{timeout} == 0);
-        select undef, undef, undef, 0.25;
+        sleep 0.25;
```
Martchus, line 56: Will this be really equivalent? Note that `select` allows fractional timeouts by default whereas `sleep` requires an additional import for that (see https://perldoc.perl.org/Time::HiRes).

consoles/video_stream.pm
```diff
@@ -340,7 +340,7 @@ sub _receive_frame_ustreamer ($self) {
 
         # tell ustreamer we are reading, otherwise it won't write new frames
         my $clock = clock_gettime(CLOCK_MONOTONIC);
-        substr($ustreamer_map, $client_clock_offset, 16) = pack 'D', $clock;
+        substr($ustreamer_map, $client_clock_offset, 16, pack 'D', $clock);
```
Martchus: To make it clear that `$clock` is passed to `pack`:
```suggestion
        substr($ustreamer_map, $client_clock_offset, 16, pack('D', $clock));
```

I must say that I like the current version of the code most as it makes it the most obvious that the substring is replaced.

Martchus (changes requested):
Considering the test API changes it would make sense to run a few random OSADO tests to see whether this breaks anything. (Although omitting the params that are now required was probably already nothing that actually worked.)

okurz (changes requested):
Now this became too big to be merged with confidence as part of one PR. I recommend to pull out the changes in testapi.pm into a dedicated PR first

Wabri:
> Now this became too big to be merged with confidence as part of one PR.

I can probably say the same for #3001 :rofl:

consoles/video_stream.pm
Wabri: Putting that parenthesis will generate:

```
14: #   ::warning file=consoles/video_stream.pm,line=343,col=9,title=Builtin function called with parentheses - severity 1::[CodeLayout::ProhibitParensWithBuiltins] See page 13 of PBP
14: #   ::warning file=consoles/video_stream.pm,line=343,col=58,title=Builtin function called with parentheses - severity 1::[CodeLayout::ProhibitParensWithBuiltins] See page 13 of PBP
```

I'll put an annotation and avoid those justifying it favor of readability :+1:

ppmclibs/tinycv.pm
```diff
@@ -12,8 +12,9 @@ use Math::Complex 'sqrt';
 require Exporter;
 require DynaLoader;
 
-our @ISA = qw(Exporter DynaLoader);
-our @EXPORT = qw();
+use parent qw(Exporter DynaLoader);
+
+our @EXPORT = qw();    ## no critic (Modules::ProhibitAutomaticExportation)
```
perlpunk, line 17: Just remove this line completely

testapi.pm
```diff
@@ -775,8 +773,8 @@ Return the given variable as array reference (split variable value by , | or ; )
 =cut
 
 sub get_var_array ($var, $default = undef) {
-    my @vars = split /,|;/, $bmwqemu::vars{$var} || '';
-    my @default = split /,|;/, $default || '';
+    my @vars = split /[,|;]/, $bmwqemu::vars{$var} || '';
```
perlpunk, line 776: So now this is also splitting on the character `|` which it wasn't before

perlpunk:
> > Maybe it would still be good to split this into multiple PRs?
> 
> Because it's better to review? I'm finding no problem on creating commit for every policy and fix them by interactively rebase.

Not necessarily one PR per policy, but this is too big. We can always revert if something is missed, but it will be easier to find out what was the cause.

okurz:
> > Now this became too big to be merged with confidence as part of one PR.
> 
> I can probably say the same for #3001 🤣

I am open for suggestions how to split it

Wabri:
> Not necessarily one PR per policy, but this is too big. We can always revert if something is missed, but it will be easier to find out what was the cause.

Make sense, all the policy is resolve by now so I can just cherry-pick them one by one or at least the one that could be connected once with the other and create prs. :+1:

consoles/ssh_screen.pm
```diff
@@ -7,6 +7,7 @@ use Mojo::Base 'consoles::serial_screen', -signatures;
 use Carp 'croak';
 use Net::SSH2 'LIBSSH2_ERROR_EAGAIN';
 use Time::Seconds;
+use Time::HiRes;
```
perlpunk, line 10: You need to explicitly import the `sleep` function.

.perlcriticrc
```diff
@@ -45,5 +48,41 @@ severity = 5
...
+max_mccabe = 160
+
+[Subroutines::ProhibitManyArgs]
+# Reduce me!
+max_arguments = 12
+# Ignore first arg if called $self or $class
+skip_object = 1
+
+[Community::PreferredAlternatives]
+allowed_modules = List::MoreUtils
```
perlpunk, line 88: The only place where we use List::MoreUtils is 
```
backend/baseclass.pm
29:use List::MoreUtils 'uniq';
```
We can get rid of List::MoreUtils and replace that with
```
use List::Util qw(uniq);
```
That should be available since perl 5.26, if we agree that this is our minimum perl version

consoles/video_stream.pm
Martchus: I find "Builtin function called with parentheses" generally questionable. I get it that this is about consistency but in general parentheses can make code much more readable because one sees at a glance what belongs together.

consoles/video_stream.pm
```diff
@@ -340,7 +340,7 @@ sub _receive_frame_ustreamer ($self) {
 
         # tell ustreamer we are reading, otherwise it won't write new frames
         my $clock = clock_gettime(CLOCK_MONOTONIC);
-        substr($ustreamer_map, $client_clock_offset, 16) = pack 'D', $clock;
+        substr($ustreamer_map, $client_clock_offset, 16, pack('D', $clock));
```
perlpunk, line 343: We can still make it clear what belongs together this way:
```suggestion
        substr($ustreamer_map, $client_clock_offset, 16, (pack 'D', $clock));
```

Martchus: Ok, that looks good, too. (Although probably a bit unfamiliar to anyone who usually uses other programming languages.)

testapi.pm
Wabri: Changed on https://github.com/os-autoinst/os-autoinst/pull/3019

consoles/video_stream.pm
Wabri: Made in https://github.com/os-autoinst/os-autoinst/pull/3021

perlpunk:
I assume this can be closed as it was split into smaller PRs

PR #3010, opened 2026-07-07
test: fix race condition in t/29-backend-driver.t

okurz:
Motivation:
The stop_backend call returns immediately before the process is fully collected.
Under some timing conditions, the expected "backend exited" message is not yet
logged at the subsequent assertion. Observed during my work on
* #2819

Design Choices:
Call $driver->stop (which blocks on the process exiting) before checking the
diagnostic log arrays, ensuring a deterministic sequence of events.

Benefits:
Fixes a flaky test failure in driver tests, ensuring robust and reliable build
runs.


Wabri (approved):

perlpunk (approved):
It's not so nice that the return value of `combined_from` is not checked at all, but that was already like that before

PR #3014, opened 2026-07-09
fix(backend): bring back support for legacy BIOS

okurz:
Motivation
Commit 4585515 erroneously removed support for custom legacy BIOS when cleaning
up the deprecated UEFI_PFLASH logic. We need a way to provide custom firmware
images for non-UEFI VMs.

Design Choices
The `BIOS` variable is re-added to the checked attribute list to resolve its
path correctly. In start_qemu, if UEFI is not requested, the BIOS file path
is passed via the `-bios` parameter.

Benefits
This restores the ability to supply custom BIOS firmware files for legacy VMs,
satisfying use-cases that depend on non-UEFI custom firmwares.

### DIFF
--- a/OpenQA/Qemu/Proc.pm
+++ b/OpenQA/Qemu/Proc.pm
@@ -293,6 +293,16 @@ sub configure_pflash ($self, $vars) {
     my $bdc = $self->blockdev_conf;
 
     return $self unless $vars->{UEFI};
+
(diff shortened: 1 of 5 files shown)


marmarek (approved):
It works, thanks :)

Martchus (approved):

perlpunk (approved):

PR #3015, opened 2026-07-09
Changes for running ustreamer on RPi5

marmarek:
The new format version is just about bumping supported ustreamer version. Other two are about RPi5. See commit messages for details.

Some questions:
- [ ] Is it okay to leave old ustreamer formats support? Theoretically they aren't needed, on a fully updated system, but I do still run older version on some nodes (although I believe all are at least at version 7 now).
- [x] The video URL parsing is getting a bit unwieldy. Maybe it's time to switch to a proper URL parsing function (which one?) instead of using regexes? Especially since now one parameter has a space in a value (with proper parsing, it could be encoded as `%20`)...
- [ ] The documentation for the `GENERAL_HW_VIDEO_STREAM_URL` parameters is getting long. Should it be moved somewhere (and also extended a bit - in its current form, I doubt it's enough...), or is it okay to have such long description in a table cell?


consoles/video_stream.pm
```diff
@@ -135,6 +135,8 @@ sub _get_ffmpeg_cmd ($self, $url) {
 sub _get_ustreamer_cmd ($self, $url, $sink_name) {
     my $fps = $url =~ s/[\?&]fps=([0-9]+)// ? $1 : 5;
     my $format = $url =~ s/[\?&]format=([A-Z0-9]+(swap)?)// ? $1 : 'UYVY';
+    my $mediadev = $url =~ s/[\?&]mediadev=([/a-zA-Z0-9-]+)// ? $1 : '';
+    my $mediaentity = $url =~ s/[\?&]mediaentity=([_ a-zA-Z0-9-]+)// ? $1 : '';
```
Martchus: You can use `Mojo::URL` for URL parsing; we depend on it anyway.

E.g.:
```
Mojo::URL->new("/dev/openqa-video?fps=2&format=BGR24&mediadev=/dev/media-csi&mediaentity=tc358743 11-000f")->query->param("mediaentity")'
```
returns:
```
tc358743 11-000f
```

Martchus (commented):
Looks good, although at this point it would make sense to do the URL parsing properly (see inline comment).

consoles/video_stream.pm
marmarek: Looks much nicer :)

okurz (approved):

Martchus (approved):

PR #3019, opened 2026-07-13
style(testapi): apply multiple policies

Wabri:
- style(perlcritic): RegularExpressions::ProhibitSingleCharAlternation
- style(testapi.pm): apply Subroutines::RequireArgUnpacking on testapi.pm
- style(testapi): apply Modules::ProhibitAutomaticExportation on testapi


okurz:
I suggest to bump the interface number as commented in https://github.com/os-autoinst/os-autoinst/pull/3019#issuecomment-4960379493 as there might be a subtle change, e.g. when one would want to overwrite/mock a test API function. other cases should not be impacted

Wabri:
> I suggest to bump the interface number as commented in [#3019 (comment)](https://github.com/os-autoinst/os-autoinst/pull/3019#issuecomment-4960379493) as there might be a subtle change, e.g. when one would want to overwrite/mock a test API function. other cases should not be impacted

Yas you are right, I forgot during the cherry-picking to include that from https://github.com/os-autoinst/os-autoinst/pull/3009/changes/f65efd4547543eacc0c3cbccfc304db08dfd8097 let me pick that

okurz (approved):

Martchus:
Have you tested this with some manual test runs (e.g. of some random osado tests)?

The line `sub diag (@args) { bmwqemu::diag(@args) }` is now weirdly no longer counted as covered. Maybe that's just a sporadic issue; otherwise we'll have to fix/workaround it.

Wabri:
> The line `sub diag (@args) { bmwqemu::diag(@args) }` is now weirdly no longer counted as covered. Maybe that's just a sporadic issue; otherwise we'll have to fix/workaround it.

I've try to put a test to cover the diag subroutine, see https://github.com/os-autoinst/os-autoinst/commit/a06812fa094d124100cc0e1b1881a2fe67a827b2, but it didn't change. I've now replace the signatures with the previous used no signatures style, but it didn't change. I thought I was behind master, but I'm not. I've no clue why is doing that.

perlpunk:
> I've try to put a test to cover the diag subroutine, see [a06812f](https://github.com/os-autoinst/os-autoinst/commit/a06812fa094d124100cc0e1b1881a2fe67a827b2), but it didn't change. I've now replace the signatures with the previous used no signatures style, but it didn't change. I thought I was behind master, but I'm not. I've no clue why is doing that.

I have an idea and can look into it on friday

perlpunk:
I don't really see it in the logs, but what I see that coverage data is completely missing for 08-autotest:
https://github.com/os-autoinst/os-autoinst/actions/runs/29410379509/job/87335778937?pr=3019#step:3:674
```
t/04-testapi-python.t                       100.0  100.0
t/05-distribution.t                          98.3   98.3
t/07-commands.t                             100.0  100.0
t/10-terminal.t                             100.0  100.0
t/10-test-image-conversion-benchmark.t      100.0  100.0
```
## Bad news

And I reproduced that locally on master with this change:
```diff
diff --git a/t/08-autotest.t b/t/08-autotest.t
index 38cc43c3..6f9f0636 100755
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -274,6 +274,8 @@ $mock_autotest->redefine(query_isotovideo => sub ($command, $arguments) {
         $command eq 'backend_can_handle' && $arguments->{function} eq 'snapshots' ? $enable_snapshots : 1;
 });
 
+done_testing; exit;
+
 my $record_resultfile_called;
 $mock_basetest->redefine(record_resultfile => sub { ++$record_resultfile_called });
 stderr_like { autotest::run_all } qr/oh noes/, 'run_all outputs status on stderr';
```
```
./tools/invoke-tests --coverage t/08-autotest.t
...
[13:19:06] t/08-autotest.t .. All 33 subtests passed 
[13:19:07]

Test Summary Report
-------------------
t/08-autotest.t (Wstat: 139 (Signal: SEGV, dumped core) Tests: 33 Failed: 0)
  Non-zero wait status: 139
Files=1, Tests=33,  1 wallclock secs ( 0.03 usr  0.01 sys +  0.57 cusr  0.15 csys =  0.76 CPU)
Result: FAIL
```
You can see a segfault there.
It's sporadic, but happens quite often. But only when running with coverage.
And no coverage is generated in that case.

A little experimenting shows that it seems to be related to the mocking/unmocking.
That's of course bad news as this shouldn't be happening, and its not easy to create a small reproducer, because the more code I remove, the less often it happens.

## Good news:
But I'm fixing it by avoiding any global Test::MockModule instances and moving them to the subtests, which is also a good refactoring in general, because it's much more clear what module/method is actually mocked in every subtest, instaed of having this wild chain of mock/unmock calls.

I'm preparing a PR now.

perlpunk:
I created #3031 .

I tested this PR on top and it works, so you can rebase after my PR is merged.

perlpunk (approved):

PR #3024, opened 2026-07-13
style(perlcritics): apply policies

Wabri:
- style(perlcritic): apply Community::Each policy 
- style(perlcritic): apply ProhibitQuotesAsQuotelikeOperatorDelimiters 
- style(perlcritic): apply BuiltinFunctions::ProhibitSleepViaSelect policy 
- style(perlcritic): ValuesAndExpressions::ProhibitInterpolationOfLiterals 
- style(perlcritic): apply Miscellanea::ProhibitUselessNoCritic policy 

reference: https://progress.opensuse.org/issues/155191

PR #3023, opened 2026-07-13
Style(perlcritics): apply Community::Wantarray, ClassHierarchies::ProhibitExplicitISA, Community::ConditionalImplicitReturn policies

Wabri:
- style(perlcritic): apply Community::Wantarray policy 
- style(perlcritic): apply ClassHierarchies::ProhibitExplicitISA policy 
- style(perlcritic): apply Community::ConditionalImplicitReturn policy 

reference: https://progress.opensuse.org/issues/155191

PR #3022, opened 2026-07-13
Style(perlcritics): include Documentation::RequirePackageMatchesPodName, BuiltinFunctions::ProhibitStringyEval, InputOutput::ProhibitOneArgSelect, ErrorHandling::RequireCheckingReturnValueOfEval policies

Wabri:
- style(perlcritic): add Documentation::RequirePackageMatchesPodName
- style(perlcritic): apply BuiltinFunctions::ProhibitStringyEval policy
- style(perlcritic): apply InputOutput::ProhibitOneArgSelect policy
- style(perlcritic): add ErrorHandling::RequireCheckingReturnValueOfEval

reference: https://progress.opensuse.org/issues/155191

PR #3025, opened 2026-07-13
Style/155191/export blocklist

Wabri:
This is the last commit to complete the migration from include to exclude, this should be merge after: #3019, #3021, #3022, #3023, #3024

- style(perlcriticrc): migrate to blocklist approach 
- style(perlcritic): apply TestingAndDebugging::ProhibitNoWarnings policy 
- style(perlcritic): apply Community::PreferredAlternatives 
- style(perlcritic): apply Subroutines::RequireArgUnpacking policy
- style(perlcritic): apply Modules::ProhibitAutomaticExportation policy

Related: https://progress.opensuse.org/issues/155191

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -58,7 +58,7 @@ add_packages = autotest bmwqemu log needle testapi Carp OpenQA::Isotovideo::Inte
 allowed_modules = FindBin
 
 [TestingAndDebugging::ProhibitNoWarnings]
-allow = redefine
+allow = redefine once
 
 # The following policies might be interesting:
 # InputOutput::ProhibitInteractiveTest
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -1,7 +1,11 @@
-theme = community + openqa
-severity = 4
+theme = community + openqa + core
+severity = 1
 
-include = strict ValuesAndExpressions::ProhibitInterpolationOfLiterals CodeLayout::ProhibitParensWithBuiltins BuiltinFunctions::RequireBlockMap BuiltinFunctions::ProhibitStringySplit ControlStructures::ProhibitNegativeExpressionsInUnlessAndUntilConditions ValuesAndExpressions::RequireQuotedHeredocTerminator Variables::ProhibitPackageVars CodeLayout::RequireTidyCode Community::WarningsSwitch Community::EmptyReturn Variables::ProhibitUnusedVariables RegularExpressions::ProhibitUselessTopic BuiltinFunctions::ProhibitUselessTopic CodeLayout::ProhibitQuotedWordLists BuiltinFunctions::RequireBlockGrep Modules::ProhibitConditionalUseStatements BuiltinFunctions::ProhibitLvalueSubstr Community::OpenArgs Plicease::ProhibitArrayAssignAref Community::Wantarray ClassHierarchies::ProhibitExplicitISA Community::ConditionalImplicitReturn Documentation::RequirePackageMatchesPodName BuiltinFunctions::ProhibitStringyEval InputOutput::ProhibitOneArgSelect ErrorHandling::RequireCheckingReturnValueOfEval Community::Each ProhibitQuotesAsQuotelikeOperatorDelimiter BuiltinFunctions::ProhibitSleepViaSelects ValuesAndExpressions::ProhibitInterpolationOfLiterals Miscellanea::ProhibitUselessNoCritic RegularExpressions::ProhibitSingleCharAlternation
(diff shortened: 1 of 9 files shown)


Wabri:
This is now fixed, the only missing policies are the one in testapi.pm that are handled in #3019

testapi.pm
```diff
@@ -644,7 +645,7 @@ sub assert_screen_change : prototype(&@) {    # no:style:signatures
     # wait_screen_change uses prototype which expects code block as an argument
     # This resolves compile time issues
     my ($coderef, @args) = @_;
-    wait_screen_change(\&{$coderef}, @_) or die 'assert_screen_change failed to detect a screen change';
+    wait_screen_change(\&{$coderef}, @args) or die 'assert_screen_change failed to detect a screen change';
```
perlpunk, line 648: Hm, this is now passing one element less to `wait_screen_change`.
But the current code looks like it has a bug. How did this ever work?
```
sub wait_screen_change : prototype(&@) {
    my ($callback, $timeout, %args) = @_;
    $timeout ||= 10;
```
This way `$timeout` was always getting the coderef assigned. Which would result in a very high number when used in numerical context.

Any ideas from others?

perlpunk: The commit where it "broke" is from 2019: a238c03fddbb85d514d92ffc8bb98006a592e6f4

https://github.com/os-autoinst/os-autoinst/pull/1142/changes#r3604948119

perlpunk: I assume it just worked because the effect was just that timeout is very high.
But with this change the timeout defaults to 10, which might actually cause failures.

Wabri: Also I miss why this changed has been removed from #3019 where it belongs to, probably I miss the change during the multiple rebase I made!

perlpunk: @okurz @Martchus WDYT about this?

okurz: I think fixing this is a good idea but it is a change of behaviour so I suggest to put this change into a separate pull request and bumping the interface number. I also checked osado and found that `assert_screen_change` is used 10 times so not much but still potential for problems especially as critical modules like "partition_setup" or "jeos/firstrun" use this

Martchus: We could nevertheless try with 10 seconds and announce the change in #eng-testing so reviewers are aware. We can also set the timeout to 30 or even 60 seconds if we want to avoid the trouble. (60 seconds is hopefully enough. At least I'd hope nobody uses that function to wait for a longer time without passing a high timeout explicitly.)

perlpunk: But splitting it out into an extra PR makes sense IMHO

okurz: I suggest to keep the original value. This is how it was originally designed and IMHO still makes sense

perlpunk (changes requested):
https://github.com/os-autoinst/os-autoinst/pull/3025#discussion_r3604857825
Can you split this into a separate PR?

Wabri:
> [#3025 (comment)](https://github.com/os-autoinst/os-autoinst/pull/3025#discussion_r3604857825) Can you split this into a separate PR?

Opened: https://github.com/os-autoinst/os-autoinst/pull/3050

testapi.pm
```diff
@@ -32,6 +32,7 @@ use constant DEFAULT_MAX_INTERVAL => 250;
 use constant DEFAULT_TIMEOUT => 90;
 use constant TIMEOUT_INCREASE => 10;
 
+## no critic (Modules::ProhibitAutomaticExportation)
```
perlpunk, line 35: Did you forget the `## use critic` after it?

.perlcriticrc
```diff
@@ -1,7 +1,11 @@
...
...s::RequireExtendedFormatting RegularExpressions::RequireLineBoundaryMatching Subroutines::ProhibitBuiltinHomonyms Subroutines::ProhibitExplicitReturnUndef Subroutines::ProhibitSubroutinePrototypes Subroutines::ProhibitUnusedPrivateSubroutines Subroutines::ProtectPrivateSubs Subroutines::RequireFinalReturn ValuesAndExpressions::ProhibitCommaSeparatedStatements ValuesAndExpressions::ProhibitConstantPragma ValuesAndExpressions::ProhibitEmptyQuotes ValuesAndExpressions::ProhibitEscapedCharacters ValuesAndExpressions::ProhibitImplicitNewlines ValuesAndExpressions::ProhibitLeadingZeros ValuesAndExpressions::ProhibitLongChainsOfMethodCalls ValuesAndExpressions::ProhibitMagicNumbers ValuesAndExpressions::ProhibitMismatchedOperators ValuesAndExpressions::ProhibitMixedBooleanOperators ValuesAndExpressions::ProhibitNoisyQuotes ValuesAndExpressions::RequireConstantVersion ValuesAndExpressions::RequireInterpolationOfMetachars Variables::ProhibitAugmentedAssignmentInDeclaration Variables::ProhibitLocalVars Variables::ProhibitPunctuationVars Variables::ProhibitReusedNames Variables::RequireInitializationForLocalVars Variables::RequireLocalizedPunctuationVars CodeLayout::ProhibitParensWithBuiltins
```
perlpunk: Why is `CodeLayout::ProhibitParensWithBuiltins` excluded?
Do you plan this in a followup PR?

Wabri: No, I dont remember why I've put that there. Updated!

testapi.pm
Wabri: No, because it's already there.

.perlcriticrc
```diff
@@ -83,3 +83,6 @@ max_mccabe = 160
 max_arguments = 12
 # Ignore first arg if called $self or $class
 skip_object = 1
+
+[Community::PreferredAlternatives]
+allowed_modules = List::MoreUtils
```
perlpunk, line 88: List::MoreUtils is used only once, like I wrote in an earlier PR.
We might check in a followup if we can replace this with List::Util::uniq which is in core since perl 5.26

perlpunk (approved):

Martchus (approved):

okurz (approved):

PR #3026, opened 2026-07-14
feat: prefer ffmpeg 7 on modern distributions

okurz:
Motivation
Leap 16.0 and Tumbleweed support highly efficient AV1 encoding via
libsvtav1 but require ffmpeg version 7. Currently, os-autoinst only
recommends ffmpeg >= 4, which can cause Zypper to revert/downgrade to
v4 during updates.

Design Choices
Update the Recommends line in dist/rpm/os-autoinst.spec to utilize a
boolean rich dependency '(ffmpeg >= 7 or ffmpeg >= 4)'. This makes
Zypper prefer version 7 on newer systems while safely falling back to
version 4 on older platforms.

Benefits
Ensures production workers automatically pull in and maintain ffmpeg-7
for high-performance AV1 video compression without manual
intervention.

Related issue: https://progress.opensuse.org/issues/203847

### DIFF
--- a/dist/rpm/os-autoinst.spec
+++ b/dist/rpm/os-autoinst.spec
@@ -157,7 +157,7 @@ Recommends:     python3-Pillow-tk
 # Optional dependency for QEMU's built-in samba service (enabled via QEMU_ENABLE_SMBD=1)
 Recommends:     samba
 # More efficient video encoding is done automatically if ffmpeg is present
-Recommends:     ffmpeg >= 4
+Recommends:     (ffmpeg >= 7 or ffmpeg >= 4)
 Requires(pre):  %{_bindir}/getent
 Requires(pre):  %{_sbindir}/useradd
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build


dist/rpm/os-autoinst.spec
```diff
@@ -157,7 +157,7 @@ Recommends:     python3-Pillow-tk
 # Optional dependency for QEMU's built-in samba service (enabled via QEMU_ENABLE_SMBD=1)
 Recommends:     samba
 # More efficient video encoding is done automatically if ffmpeg is present
-Recommends:     ffmpeg >= 4
+Recommends:     (ffmpeg >= 7 or ffmpeg >= 4)
```
Martchus, line 160: From a mathematical standpoint this changes does nothing. If this makes a difference, rpm/zypper work in mysterious ways. Based on my experience I doubt that so I made some tests.

I have `os-autoinst` installed on my TW system since I've set it up and cannot reproduce the issue you're trying to solve despite updating regularly.

* `zypper info --recommends os-autoinst` says just ` ffmpeg >= 4`
* `ffmpeg-4` is available, e.g. `zypper in ffmpeg-4` would install it (replacing `ffmpeg-8`)
* invoking `sudo zypper in 'ffmpeg >= 4'` just says `'ffmpeg-8', welches 'ffmpeg >= 4' bietet, ist bereits installiert.`

I also cross-checked the situation on Leap 16.0 on openqaworker23:

* `zypper in 'ffmpeg >= 7'` installs `ffmpeg-7` replacing `ffmpeg-4`
* then `zypper in 'ffmpeg >= 4'` gives `'ffmpeg-7', welches 'ffmpeg >= 4' bietet, ist bereits installiert.`
* even `zypper in --force --recommends 'os-autoinst'` will just reinstall `os-autoinst` without downgrading ffmpeg

So maybe some other script/constraint is trying to keep ffmpeg at version 4 on the machine you're observing the downgrades?

Martchus (changes requested):

okurz:
> So maybe some other script/constraint is trying to keep ffmpeg at version 4 on the machine you're observing the downgrades?

true. Verified with multiple clean container based test systems. It seems we can just manually upgrade to ffmpeg-7 on current production systems and should be good.

PR #3028, opened 2026-07-15
fix: Set timeout on copying a large iso image

nanzhg:
Timeout failure occurs while copying a large ISO image (14GB). Based on the time spent while copying the image, just setting the timeout parameter to fix it.
https://openqa.suse.de/tests/23358603/modules/bootloader_svirt/steps/16
`Time out waiting for data (-9 LIBSSH2_ERROR_TIMEOUT) at /usr/lib/perl5/vendor_perl/5.42.0/x86_64-linux-thread-multi/Net/SSH2.pm line 51, <$fh> line 25.`

Related ticket: https://progress.opensuse.org/issues/204117

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -383,7 +383,7 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     return $dest_image;
 }
 
-sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_openqa_datastore, $vmware_disk_path, $vmware_disk_path_thinfile) {
+sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_openqa_datastore, $vmware_disk_path, $vmware_disk_path_thinfile, $copy_timeout = 600) {
     # If the file exists, make sure someone else is not copying it there right now,
     # otherwise copy image from NFS datastore.
     my $nfs_dir = $backingfile ? 'hdd' : 'iso';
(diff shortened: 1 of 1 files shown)


consoles/sshVirtsh.pm
```diff
@@ -398,7 +398,7 @@ sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_open
       'else ' .
       "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
       'fi;';
-    my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer');
+    my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer', timeout => 600);
```
okurz: better prevent magic numbers deep in the code. What you can do at least is to put that into a function parameter

```perl
sub _copy_image_vmware ($self, …, $copy_timeout => 600) {
…
my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer', timeout => $copy_timeout);
```

okurz (changes requested):

consoles/sshVirtsh.pm
nanzhg: Already updated. Thank you!

Julie-CAO (approved):

okurz (approved):

alice-suse (approved):

alice-suse:
@okurz We need this PR  for current 16.0 QU VR which is due tomorrow. Could you help merge it if all ok?

okurz:
> @okurz We need this PR for current 16.0 QU VR which is due tomorrow. Could you help merge it if all ok?

All our PRs are merged automatically as soon as at least two maintainers of a project approve. This hasn't happened here unfortunately but as I also trust the reviews done by you I can make an exception and merge manually ;)
Note that this change will be deployed automatically in about 6h also on OSD *if* all checks in the automatic deployment pass.

PR #3031, opened 2026-07-17
Refactor t/08-autotest.t

perlpunk:
Under certain circumstances t/08-autotest.t ran into a segfault when running
with coverage enabled.

For example when adding

    done_testing; exit;

after line 274.

Experimenting showed that the mocking/unmocking seems to be responsible
for that somehow.

Moving all Test::MockModule instances to the subtests makes this go
away.

The refactoring is split into several commits to better see the changes.

### DIFF
--- a/t/08-autotest.t
+++ b/t/08-autotest.t
@@ -115,8 +115,6 @@ subtest 'load_test && run_all' => sub {
     is $completed, 1, 'start+next+start should complete';
 };
 
-my $died;
-my $completed;
 # Test loading snapshots with always_rollback flag. Have to put it here, before loading
 # runargs test module, as it fails.
 my ($reverts_done, $snapshots_made) = (0, 0);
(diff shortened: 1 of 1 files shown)


t/08-autotest.t
```diff
@@ -241,117 +245,132 @@ subtest 'test always_rollback flag' => sub {
         stderr_like { $w = warning { autotest::run_all } } qr/Snapshots are not supported/, 'run_all outputs on stderr';
         like $w, qr/always_rollback requested but snapshots are not supported by the backend/, 'fails with explicit error message';
     };
-    $mock_basetest->unmock($_) for qw(runtest test_flags);
-    $mock_autotest->unmock($_) for qw(load_snapshot make_snapshot query_isotovideo);
+    undef $mock_autotest;
+    undef $mock_basetest;
```
perlpunk, line 249: These two lines are somehow still necessary.
Maybe the garbage collection at the end of the subtest is not run early enough, because when the variables move out of scope the same thing should happen as when we undef them explicitly.

okurz (approved):

Wabri (approved):

PR #3032, opened 2026-07-18
perf(backend): reduce `check_screen` timeout cost

mimi1vx:
## Motivation

`check_screen` timeout responses serialized and transferred the same reduced
screenshot history as `assert_screen`, although `testapi.pm` discarded every
entry except the final mismatch after receiving it. With 1024x768 screenshots,
this caused avoidable JSON, IPC, and memory costs.

## Design choices

- Serialize only the post-reduction final mismatch for `check_screen`.
- Preserve the complete reduced evidence set for `assert_screen`.
- Preserve the existing timeout-level `image` for wire compatibility.
- Preserve the final mismatch image, candidates, frame, stall metadata, and
  pause-on-timeout behavior.
- Add a mode-aware benchmark using real tinycv PPM generation and myjsonrpc
  transfer.

openQA does not consume the internal `failed_screens` response. It receives the
result details and screenshot files produced by testapi, which already records
only the final mismatch for `check_screen`.

## Results

On the pinned Linux arm64 development container, 20 sequential iterations of
the controlled 23-image timeout fixture produced:

| Check timeout metric | Before | After |
|---|---:|---:|
| Failed screenshots | 22 | 1 |
| Payload bytes | 74,257,436 | 6,457,165 |
| Median construction | 26.2 ms | 3.1 ms |
| Median JSON encode | 139.5 ms | 11.6 ms |
| Median JSON decode | 53.7 ms | 4.4 ms |
| Median myjsonrpc transfer | 270.2 ms | 23.5 ms |
| Peak RSS | 590,480 KiB | 198,088 KiB |

The normalized check response hash remained unchanged. The complete assert
payload hash also remained unchanged.

## Testing

- Focused backend, testapi, command-processing, JSON-RPC, driver, and selector
  tests: 148 assertions passed.
- Style and Perl::Critic: 192 assertions passed.
- Full Perl suite: 73 files and 1,012 assertions passed, including
  `t/99-full-stack.t`.
- Heaptrack check workload: peak heap fell from 622.81 MiB to 172.87 MiB.
- Upstream gitlint commit-message check passed.

The aggregate `ninja check` functional target passed. Its author-test aggregate
reported only `xt/27-make-update-deps.t`, which intentionally requires an
otherwise clean tracked worktree and therefore sees the PR changes themselves;
no dependency file changed.

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -997,6 +997,7 @@ sub _failed_screens_to_json ($self) {
             push @$failed_screens, $final_mismatch if ($sim < 50);
         }
     }
+    $failed_screens = [$failed_screens->[-1]] if $self->assert_screen_check && @$failed_screens;
     my @json_fails = map {
         my ($img, $failed_candidates, $testtime, $similarity, $frame) = @$_;
         {candidates => $failed_candidates, image => encode_base64($img->ppm_data), frame => $frame}
--- a/t/03-testapi.t
+++ b/t/03-testapi.t
@@ -553,6 +553,34 @@ subtest 'check_assert_screen' => sub {
         is $report_timeout_called, 1, 'report_timeout called only once';
     };
 
+    subtest 'mode-specific failure evidence recording' => sub {
+        my @recorded;
+        $mock_basetest->redefine(record_screenfail => sub ($self, %args) {
+                push @recorded, \%args;
+        });
+        my @failed_screens = map {
+            {image => $dummy_image, candidates => ["candidate-$_"], frame => $_}
+        } 1 .. 3;
+        my %multi_response = (timeout => 1, tags => [qw(fake tags)], failed_screens => \@failed_screens);
+
+        testapi::_check_backend_response(\%multi_response, 1, 10, 'foo');
+        is_deeply [map { [$_->{frame}, $_->{needles}, $_->{result}, $_->{overall}] } @recorded],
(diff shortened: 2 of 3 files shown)


okurz (changes requested):
Overall, nice that you are looking into optimizing os-autoinst performance but I don't quite get it yet. Under which circumstances does the optimization show an actual benefit? Also I think the benchmark script isn't necessary to be committed after this PR is merged, is it?

mimi1vx:
**What the change does / where it helps.** It only touches the `check_screen` timeout path — `assert_screen` behavior is unchanged. Currently, whenever a `check_screen` times out, `_failed_screens_to_json` base64-encodes and ships the whole reduced mismatch set (up to ~21 screenshots) back to isotovideo. But `check_screen` is designed to fail routinely and its caller only branches on the boolean result — it never inspects the historical mismatch screenshots. So for the `check` case the change serializes and transfers only the final diagnostic mismatch. `assert_screen` still gets the complete reduced evidence set, since that's where the failure screenshots are actually consumed.

**Circumstances showing a benefit.** Test flows with many optional/polling `check_screen` calls that hit their timeout — each such timeout avoids ~20 base64 encodes plus the serialization and IPC transfer of that PPM data to the isotovideo process. It's negligible for a single call and won't change the happy path at all; the win is in aggregate over runs that lean heavily on `check_screen`

**benchmark script** dropped

okurz (approved):

Martchus (approved):

PR #3036, opened 2026-07-21
ci: Use latest Devel::Cover git

perlpunk:
to see if there are any breakages

### DIFF
--- a/t/00-compile-check-all.t
+++ b/t/00-compile-check-all.t
@@ -24,6 +24,7 @@ my $TEST_SKIP = [
 ];
 
 my $test = Test::Compile->new();
+$test->verbose(1);
 my @files;
 
 # Prevent any non-tracked files or files within .git (e.g. in.git/rr-cache) to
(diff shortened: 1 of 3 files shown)


perlpunk:
I reported the uncovered `elsif` lines here: https://github.com/pjcj/Devel--Cover/issues/633

perlpunk:
https://github.com/pjcj/Devel--Cover/issues/633 was fixed \o/

Now there are only two changes left:
```
consoles/serial_screen.pm:57                                 | sub hold_key ($self, $args) { croak $trying_to_use_keys }
consoles/serial_screen.pm:59                                 | sub release_key ($self, $args) { croak $trying_to_use_keys }
```
and I think it's likely that those are indeed not covered.

perlpunk:
> Now there are only two changes left:
> 
> ```
> consoles/serial_screen.pm:57                                 | sub hold_key ($self, $args) { croak $trying_to_use_keys }
> consoles/serial_screen.pm:59                                 | sub release_key ($self, $args) { croak $trying_to_use_keys }
> ```
> 
> and I think it's likely that those are indeed not covered.

That was actually because they were indeed uncovered, see #3041
The signature check failed, so strictly speaking the subroutine was called, but the body of the subroutine was not executed.

My guess is this was only revealed now in perl 5.44 because probably the subroutine signature implementation changed.

PR #3039, opened 2026-07-22
feat: gracefully handle temporarily unreachable hypervisors

okurz:
Motivation
Due to maintenance or periodic reboots, remote hypervisor hosts may be
temporarily unavailable when an openQA worker attempts to run a job.
Since openQA jobs can run for hours, a brief reboot window shouldn't
cause an immediate, unrecoverable job failure.

Design Choices
We add a pre-check for reachability using `wait_for_ssh_port` in
`sshVirtsh::activate` before initiating VNC or XML generation. If the
host is unreachable after `SVIRT_HYPERVISOR_SSH_TIMEOUT` (default 1800),
we die with a clear message.

Benefits
- Prevents starting Xvnc or parsing XML when the backend is unreachable.
- Sets up the ability for openQA to automatically clone/retry jobs that
  hit this error.

Related issue: https://progress.opensuse.org/issues/204510

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -41,6 +41,13 @@ sub activate ($self) {
     # initialize SSH console(s)
     $self->_init_ssh($args);
 
+    # Pre-check: Ensure hypervisor is reachable before starting Xvnc/activating console
+    my $hostname = $self->{ssh_credentials}->{default}->{hostname};
+    my $timeout = $bmwqemu::vars{SVIRT_HYPERVISOR_SSH_TIMEOUT} // 1800;
+    if (!$self->wait_for_ssh_port($hostname, timeout => $timeout)) {
+        die "backend died: hypervisor host $hostname is not reachable after $timeout seconds\n";
+    }
+
     # start Xvnc
     $self->SUPER::activate;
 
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

consoles/sshVirtsh.pm
```diff
@@ -41,6 +41,13 @@ sub activate ($self) {
     # initialize SSH console(s)
     $self->_init_ssh($args);
 
+    # Pre-check: Ensure hypervisor is reachable before starting Xvnc/activating console
+    my $hostname = $args->{hostname} or die 'we need a hostname to ssh to';
```
Martchus: This `die` is C&P from `_init_ssh`. We should probably use `$self->{ssh_credentials}->{default}->{hostname}` here without the `die` to make it clear that we rely on the initialization done via `_init_ssh`.

Martchus (changes requested):

consoles/sshVirtsh.pm
okurz: good idea. done

Martchus (approved):

kalikiana:
@okurz FYI still failing on **Author tests**

```
2/9 Test  #3: test-local-python-style .............***Failed    0.02 sec
noqa-comments: `ruff: noqa` comment used instead of `ruff:file-ignore`
 --> t/fake/tests/faulty.py:1:1
  |
1 | # ruff: noqa: F401, D100, CPY001, INP001
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2 | import thismoduleshouldnotexist
  |
help: Use `ruff:file-ignore` instead

Found 1 error.
```

PR #3040, opened 2026-07-22
perf(backend): track write offset instead of re-copying encoder buffers

mimi1vx:
## Why

Partial `syswrite`s to the video encoder pipe re-copied the entire unwritten
tail of the current frame buffer via `substr` on every short write. For
large frames written in small kernel-sized chunks this produced O(n²)
copying inside the capture loop's hot path.

## What changed

`_write_buffered_data_to_file_handle` now keeps the head buffer in place and
tracks a per-file-handle byte offset, writing via the 4-arg
`syswrite($fh, $data, $length, $offset)` instead of copying the remainder
back into the queue. The buffer is only shifted off the queue once fully
drained.

## Benchmark

Local measurement draining a ~2.36 MiB frame in 4 KiB chunks through a fake
file handle:

| Metric | Before | After |
|---|---:|---:|
| Bytes copied by tail-copy path | ~712 MiB (746,239,584 bytes) | 0 |
| Wall time to drain one frame | ~26.7 ms | ~1.0 ms |

Output byte stream reaching the encoder verified byte-identical before and
after the change.

## Testing

- `make test-perl-testsuite TESTS="t/23-baseclass.t"`
- `make test-perl-testsuite TESTS="xt/01-style.t xt/02-perlcritic.t"`
- `make tidy-perl` (no diff)

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -149,15 +149,16 @@ sub _write_buffered_data_to_file_handle ($self, $program_name, $array_of_buffers
     # buffer is kept in place and written from a tracked offset to avoid re-copying the
     # unwritten tail on every partial write
     my $data = $array_of_buffers->[0];
+    my $len = length $data;
     my $offset = $self->{_encoder_write_offset}{$fh} // 0;
-    my $data_written = $fh->syswrite($data, length($data) - $offset, $offset);
+    my $data_written = $fh->syswrite($data, $len - $offset, $offset);
     die "$program_name not accepting data: $!" unless defined $data_written;
 
     $offset += $data_written;
-    if ($offset == length $data) {
+    if ($offset == $len) {
         # buffer fully written, move on to the next one
         shift @$array_of_buffers;
-        delete $self->{_encoder_write_offset}{$fh};
+        $self->{_encoder_write_offset}{$fh} = 0;
     }
     else {
         $self->{_encoder_write_offset}{$fh} = $offset;
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -145,13 +145,23 @@ sub run ($self, $cmdpipe, $rsppipe) {
 }
 
 sub _write_buffered_data_to_file_handle ($self, $program_name, $array_of_buffers, $fh) {
-    # write as much data as possible (this is called when $fh is ready to write)
-    my $data = shift @$array_of_buffers;
-    my $data_written = $fh->syswrite($data);
+    # write as much data as possible (this is called when $fh is ready to write); the head
+    # buffer is kept in place and written from a tracked offset to avoid re-copying the
+    # unwritten tail on every partial write
+    my $data = $array_of_buffers->[0];
+    my $offset = $self->{_encoder_write_offset}{$fh} // 0;
+    my $data_written = $fh->syswrite($data, length($data) - $offset, $offset);
     die "$program_name not accepting data: $!" unless defined $data_written;
 
-    # put remaining data it back into the queue
-    unshift @$array_of_buffers, substr $data, $data_written unless $data_written == length $data;
+    $offset += $data_written;
+    if ($offset == length $data) {
+        # buffer fully written, move on to the next one
+        shift @$array_of_buffers;
+        delete $self->{_encoder_write_offset}{$fh};
+    }
+    else {
+        $self->{_encoder_write_offset}{$fh} = $offset;
+    }
 
     # remove file handle from selects if there's no more data to write
     if (!@$array_of_buffers) {
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -391,6 +391,53 @@ sub _prepare_video_encoder ($baseclass) {
     $baseclass->{external_video_encoder_image_data} = [55 .. 60];
 }
 
+# fake file handle that only accepts a few bytes per syswrite call, used to
+# exercise the offset-based partial-write handling in
+# _write_buffered_data_to_file_handle without relying on real pipe buffering
+package FakeShortWriteFh;
+sub new ($class, $chunk_size) { return bless {chunk_size => $chunk_size, written => ''}, $class }
+
+sub syswrite ($self, $data, $length = undef, $offset = 0) {
+    my $available = (defined $length ? $length : length($data) - $offset);
+    my $n = $available < $self->{chunk_size} ? $available : $self->{chunk_size};
+    $self->{written} .= substr $data, $offset, $n;
+    return $n;
+}
+
+# records calls instead of touching a real IO::Select, since the fake fh above is not a real filehandle
+package RecordingSelect;
+sub new ($class) { return bless {removed => []}, $class }
+sub remove ($self, $fh) { push @{$self->{removed}}, $fh }
+
+package main;
+
+subtest '_write_buffered_data_to_file_handle offset writes' => sub {
+    my $baseclass = backend::baseclass->new();
+    $baseclass->{select_read} = RecordingSelect->new;
+    $baseclass->{select_write} = RecordingSelect->new;
+    my $fh = FakeShortWriteFh->new(3);
+
+    my $first_buffer = 'ABCDEFGHIJ';    # 10 bytes, drained in 3-byte chunks
+    my $second_buffer = 'KLMNOPQ';    # 7 bytes
+    my $buffers = [$first_buffer, $second_buffer];
+
+    backend::baseclass::_write_buffered_data_to_file_handle($baseclass, 'Encoder', $buffers, $fh);
+    is scalar @$buffers, 2, 'head buffer not shifted off after a partial write';
+    is $buffers->[0], $first_buffer, 'head buffer left untouched (no tail copy) after a partial write';
+    is $fh->{written}, 'ABC', 'first chunk written';
+
+    backend::baseclass::_write_buffered_data_to_file_handle($baseclass, 'Encoder', $buffers, $fh) for 1 .. 3;
+    is scalar @$buffers, 1, 'head buffer shifted off only once fully written';
+    is $buffers->[0], $second_buffer, 'next buffer now at the head of the queue';
+    is $fh->{written}, $first_buffer, 'first buffer reassembled in order with no corruption at chunk boundaries';
+
+    backend::baseclass::_write_buffered_data_to_file_handle($baseclass, 'Encoder', $buffers, $fh) for 1 .. 3;
+    is scalar @$buffers, 0, 'queue empty once all buffers fully written';
+    is $fh->{written}, $first_buffer . $second_buffer, 'full byte stream reassembled in order, losslessly, across multiple calls';
+    is_deeply $baseclass->{select_read}->{removed}, [$fh], 'file handle removed from select_read once queue drained';
+    is_deeply $baseclass->{select_write}->{removed}, [$fh], 'file handle removed from select_write once queue drained';
+};
+
 subtest 'video-encoder' => sub {
     my $baseclass = backend::baseclass->new();
     _prepare_video_encoder($baseclass);


mimi1vx:
The `test` job failed on `check-line-coverage` reporting 2 uncovered lines in `consoles/serial_screen.pm:57,59` (`hold_key`/`release_key`). This is a pre-existing flaky coverage check unrelated to this PR:

- This change only touches `backend/baseclass.pm` and `t/23-baseclass.t`; it does not touch `consoles/serial_screen.pm` or `t/26-serial_screen.t`.
- Those two lines are exercised by `t/26-serial_screen.t:21-22` (`dies_ok { $screen->hold_key }` / `release_key`) and are covered in isolated local coverage runs.
- The identical 2-line failure (`serial_screen.pm:57,59`) also occurred recently on an unrelated branch (e.g. `feature/s390x_graceful_2`, run 29950948809), confirming it is an intermittent coverage-merge flake.

All other checks (Author tests, fullstack, style, perlcritic, OBS builds) pass. Could a maintainer please re-run the `test` job?

okurz:
> The `test` job failed on `check-line-coverage` reporting 2 uncovered lines in `consoles/serial_screen.pm:57,59` (`hold_key`/`release_key`). This is a pre-existing flaky coverage check unrelated to this PR:

This is likely fixed by a new Devel::Cover version. Likely fixed soon by @perlpunk in https://progress.opensuse.org/issues/203757

perlpunk:
> This is likely fixed by a new Devel::Cover version. Likely fixed soon by @perlpunk in https://progress.opensuse.org/issues/203757

What makes you think that?

okurz:
> > This is likely fixed by a new Devel::Cover version. Likely fixed soon by @perlpunk in https://progress.opensuse.org/issues/203757
> 
> What makes you think that?

I thought you mentioned that. Then what else would fix that?

perlpunk:
> > > This is likely fixed by a new Devel::Cover version. Likely fixed soon by @perlpunk in https://progress.opensuse.org/issues/203757
> > 
> > 
> > What makes you think that?
> 
> I thought you mentioned that. Then what else would fix that?

1. Checking if this is actually covered or not and if not file a Devel::Cover issue (1)
3. I'm not assigned to the ticket you mentioned (although I have been doing almost all work for it so far)


(1)
I did mention those lines in https://github.com/os-autoinst/os-autoinst/pull/3036#issuecomment-5046540970 and explicitly wrote "and I think it's likely that those are indeed not covered."

perlpunk:
Regarding the uncovered lines: This is now appearing out of the blue without changes on our side nor a change in the Devel::Cover version, so I assume it's because we test on Tumbleweed and we have perl 5.44 there now, so it might be a subtle change in perl itself.

perlpunk:
#3041 should fix it and cover those methods.

backend/baseclass.pm
```diff
@@ -145,13 +145,23 @@ sub run ($self, $cmdpipe, $rsppipe) {
...
+    my $data_written = $fh->syswrite($data, length($data) - $offset, $offset);
     die "$program_name not accepting data: $!" unless defined $data_written;
 
-    # put remaining data it back into the queue
-    unshift @$array_of_buffers, substr $data, $data_written unless $data_written == length $data;
+    $offset += $data_written;
+    if ($offset == length $data) {
+        # buffer fully written, move on to the next one
+        shift @$array_of_buffers;
+        delete $self->{_encoder_write_offset}{$fh};
```
Martchus: Since you're already optimizing: I guess it is cheaper to keep the element in the hash map and just change its value.

```suggestion
        $self->{_encoder_write_offset}{$fh} = 0;
```

backend/baseclass.pm
```diff
@@ -145,13 +145,23 @@ sub run ($self, $cmdpipe, $rsppipe) {
...
+    my $offset = $self->{_encoder_write_offset}{$fh} // 0;
+    my $data_written = $fh->syswrite($data, length($data) - $offset, $offset);
     die "$program_name not accepting data: $!" unless defined $data_written;
 
-    # put remaining data it back into the queue
-    unshift @$array_of_buffers, substr $data, $data_written unless $data_written == length $data;
+    $offset += $data_written;
+    if ($offset == length $data) {
+        # buffer fully written, move on to the next one
+        shift @$array_of_buffers;
```
Martchus, line 160: This might also not be very efficient but I'm not sure how Perl handles `shift` internally. If we're lucky it just changes the pointer to the first element and discards the first buffer. If we're unlucky it does swap or even copy operations. (Optimizing this further is of course not required to merge this PR, I'm just thinking what else is not optimal here.)

backend/baseclass.pm
```diff
@@ -145,13 +145,23 @@ sub run ($self, $cmdpipe, $rsppipe) {
...
 sub _write_buffered_data_to_file_handle ($self, $program_name, $array_of_buffers, $fh) {
-    # write as much data as possible (this is called when $fh is ready to write)
-    my $data = shift @$array_of_buffers;
-    my $data_written = $fh->syswrite($data);
+    # write as much data as possible (this is called when $fh is ready to write); the head
+    # buffer is kept in place and written from a tracked offset to avoid re-copying the
+    # unwritten tail on every partial write
+    my $data = $array_of_buffers->[0];
+    my $offset = $self->{_encoder_write_offset}{$fh} // 0;
+    my $data_written = $fh->syswrite($data, length($data) - $offset, $offset);
```
Martchus: Since you're already optimizing: I wonder whether we can avoid computing the length here manually.
```suggestion
    my $data_written = $fh->syswrite($data, undef, $offset);
```

Martchus (approved):
Looks good. I only have a few remarks. I'm also wondering how you came up with the figures you mention under "Benchmark" in the PR description.

backend/baseclass.pm
```diff
@@ -145,13 +145,23 @@ sub run ($self, $cmdpipe, $rsppipe) {
...
     die "$program_name not accepting data: $!" unless defined $data_written;
 
-    # put remaining data it back into the queue
-    unshift @$array_of_buffers, substr $data, $data_written unless $data_written == length $data;
+    $offset += $data_written;
+    if ($offset == length $data) {
```
Martchus: We could probably avoid the repeated `length $data`.

backend/baseclass.pm
mimi1vx: Done in 275d0f6a — the entry is now reset to `0` in place instead of `delete`d, so the two long-lived encoder pipes keep a stable hash slot and avoid delete/insert churn.

backend/baseclass.pm
mimi1vx: Done in 275d0f6a — `length $data` is now computed once into `my $len` and reused for both the `syswrite` length and the completion check.

backend/baseclass.pm
mimi1vx: Good idea, but I checked and it's unsafe on our supported perl: passing `undef` as the LENGTH argument to `syswrite` does **not** mean "rest of the string from OFFSET" — it writes 0 bytes. Verified on perl 5.34:

```
$ perl -e 'open(my $w,">","/tmp/t"); my $d="ABCDEFGHIJ"; my $n=syswrite($w,$d,undef,3); print "wrote=$n\n"'
wrote=0
```

whereas the explicit length works as expected:

```
$ perl -e 'open(my $w,">","/tmp/t"); my $d="ABCDEFGHIJ"; my $n=syswrite($w,$d,length($d)-3,3); print "wrote=$n\n"'
wrote=7
```

So I kept the explicit `$len - $offset` (now using the hoisted `$len`). A zero-byte write here would also stall the encoder drain loop, so this one is a correctness issue rather than just style.

backend/baseclass.pm
mimi1vx: Good point. In perl, `shift @array` on an array of scalars is cheap: it doesn't copy or move the remaining elements — it just advances the array's internal start pointer (AvARRAY/AvFILL bookkeeping) and drops the refcount on the shifted SV. The frame buffer string bytes themselves are never copied by the shift; only the SV holding the (already fully-drained) head buffer is released. So the queue drain is O(1) per completed buffer, and combined with the offset change the per-frame path is now free of the O(n²) tail copies. I left it as-is.

mimi1vx:
> I'm also wondering how you came up with the figures you mention under "Benchmark".

Fair question — here's the reproducible standalone harness. It isolates the drain hot path with a fake filehandle that accepts at most a pipe-sized chunk per `syswrite` (mimicking a nearly-full encoder pipe), then drains one ~2.36 MiB frame in 4 KiB chunks through both the old `substr`/`unshift` tail-copy and the new offset-based drain, counting the bytes physically copied by the tail-copy path and the wall time.

```perl
use strict; use warnings; use v5.20;
use feature 'signatures'; no warnings 'experimental::signatures';
use Time::HiRes qw(time);

package ChunkFh;    # accepts at most CHUNK bytes per syswrite, discards output
sub new ($c, $chunk) { bless { chunk => $chunk, n => 0 }, $c }
sub syswrite ($s, $data, $len = undef, $off = 0) {
    my $avail = defined $len ? $len : length($data) - $off;
    my $w = $avail < $s->{chunk} ? $avail : $s->{chunk};
    $s->{n} += $w; return $w;
}
package main;

my ($FRAME, $CHUNK) = (2_360_000, 4096);
my $frame = 'x' x $FRAME;

our $old_bytes_copied = 0;
sub old_drain ($buffers, $fh) {                 # pre-PR logic
    my $data = shift @$buffers;
    my $w = $fh->syswrite($data);
    unless ($w == length $data) {
        my $tail = substr $data, $w;            # O(n) tail copy per short write
        $old_bytes_copied += length $tail;
        unshift @$buffers, $tail;
    }
}
sub new_drain ($state, $buffers, $fh) {         # this PR
    my $data = $buffers->[0]; my $len = length $data;
    my $off = $state->{off} // 0;
    my $w = $fh->syswrite($data, $len - $off, $off);
    $off += $w;
    if ($off == $len) { shift @$buffers; $state->{off} = 0 } else { $state->{off} = $off }
}

my $fh1 = ChunkFh->new($CHUNK); my $b1 = [$frame];
my $t0 = time; old_drain($b1, $fh1) while @$b1; my $old_ms = (time-$t0)*1000;

my $fh2 = ChunkFh->new($CHUNK); my $b2 = [$frame]; my $st = {};
my $t1 = time; new_drain($st, $b2, $fh2) while @$b2; my $new_ms = (time-$t1)*1000;

printf "OLD: tail-copy=%.0f MiB, drain=%.1f ms\n", $old_bytes_copied/1048576, $old_ms;
printf "NEW: tail-copy=0,       drain=%.1f ms  (%.0fx)\n", $new_ms, $old_ms/$new_ms;
```

Latest run on my machine (perl 5.34, 576 short writes of a 2.36 MiB frame):

```
frame=2360000 bytes, chunk=4096 bytes, short-writes=576
OLD: tail-copy bytes=678703104 (647 MiB), drain=33.1 ms, all 2360000 bytes delivered=yes
NEW: tail-copy bytes=0, drain=0.3 ms, all 2360000 bytes delivered=yes
speedup=128.8x
```

The absolute ms figures vary a bit per run/machine and differ slightly from the original PR-description table (that run measured ~712 MiB / ~26.7 ms → ~1.0 ms), but the shape is the consistent point: the old path copies **O(frame_size × short_writes)** bytes — quadratic in the number of chunks a frame is split into — while the new path copies **zero**. The tail-copy byte count is the deterministic, machine-independent metric; it's ~647–712 MiB depending on exactly where the chunk boundaries fall.

The lossless/ordered reassembly across chunk boundaries is what `t/23-baseclass.t`'s new `_write_buffered_data_to_file_handle offset writes` subtest pins down.

Martchus (approved):

okurz:
@mimi1vx can you clarify what is the actual net benefit of this change on the top level?

Martchus:
**Less copying, so less CPU and memory usage.** Whether this is actually notable/measurable in real test runs is of course the question. (But also not something we'll have to answer to get this merged. And improvement is an improvement, even if it is small.)

okurz (approved):
> **Less copying, so less CPU and memory usage.** Whether this is actually notable/measurable in real test runs is of course the question. (But also not something we'll have to answer to get this merged. And improvement is an improvement, even if it is small.)

yes, but needing to understand and maintain the code which is at least an increase of cost :)

backend/baseclass.pm
perlpunk: Documented here btw: https://perldoc.perl.org/perlguts#Offsets

mimi1vx:
> but needing to understand and maintain the code which is at least an increase of cost :)

I would say that changing from an O(n^2) algorithm to an O(1) algorithm outweighs the burden of adding 5 lines of code.

PR #3042, opened 2026-07-24
perf(backend): read serial output incrementally in wait_serial

mimi1vx:
## Why

`wait_serial`'s poll loop called `serial_text()`, which re-opens `serialfile`, seeks back to the fixed start offset, and slurps everything to EOF on **every** poll. If `N` bytes accumulate over `P` polls before a match (or timeout), this re-reads and re-scans the same growing unmatched region on every iteration, producing `O(N*P)` disk reads and allocations in a hot wait path used throughout test execution.

## What changed

The loop now tracks an advancing read position (`$read_pos`) and reads only the newly appended bytes each poll via `read_serial($read_pos)`, appending them to an accumulating buffer instead of re-slurping from the fixed `serial_offset`. The buffer contents, regex/`no_regex` matching, offset bookkeeping, and return value are unchanged — only the redundant disk re-reads are eliminated.

- `backend/baseclass.pm`: incremental `read_serial($read_pos)` loop in `wait_serial` instead of a full `serial_text()` re-slurp per poll.
- `t/23-baseclass.t`: regression tests for a match straddling a poll boundary, and for the no-data/timeout case (confirms the result is an empty string, not `undef`, matching `read_serial`'s actual EOF semantics on a freshly opened handle).

## Benchmark

2 MiB accumulated over 100 polls, non-matching regex until the final poll (local measurement, throwaway script, not included in this PR):

| | before | after |
|---|---|---|
| cumulative bytes read | ~105.9 MiB | ~2.08 MiB (~50x fewer) |
| wall time for poll loop | ~18 ms | ~4 ms |

`t/23-baseclass.t` output verified identical before/after (equivalence oracle: same assertions pass against both the modified and the original `wait_serial`).

## Testing

- `prove t/23-baseclass.t` — 33/33 pass
- `prove xt/01-style.t xt/02-perlcritic.t` — clean

### DIFF
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -859,16 +859,6 @@ sub clear_serial_buffer ($self, @) {
 }
 
 
-=head2 serial_text
-
-Returns the output on the serial device since the last call to clear_serial_buffer
-
-=cut
-
-sub serial_text ($self) {
-    return ($self->read_serial($self->{serial_offset}))[0];
-}
-
 =head2 read_serial
 
 Returns the output and the offset after reading on the serial device from position
@@ -899,8 +889,10 @@ sub wait_serial ($self, $args) {
     $regexp = [$regexp] if ref $regexp ne 'ARRAY';
     my $initial_time = time;
     my $current_offset = $self->{serial_offset};
+    my $read_pos = $self->{serial_offset};
     while (time < $initial_time + $timeout) {
-        $str = $self->serial_text();
+        (my $chunk, $read_pos) = $self->read_serial($read_pos);
+        $str = ($str // '') . $chunk;
         for my $r (@$regexp) {
             if (!$args->{no_regex} && $str =~ m/$r/) {
                 $current_offset += $LAST_MATCH_END[0];
--- a/t/23-baseclass.t
+++ b/t/23-baseclass.t
@@ -500,6 +500,31 @@ BdsDxe: starting Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x8,0x0)'}, '
     is_deeply $baseclass->wait_serial({%dargs, regexp => qr/welcome$/, timeout => 1}), {matched => 0, string => "\nWelcome to GRUB!\n"}, 'Test regex mismatch';
     is_deeply $baseclass->wait_serial({%dargs, regexp => 'something wrong', timeout => 1, no_regex => 1}), {matched => 0, string => "\nWelcome to GRUB!\n"}, 'Test string literal mismatch';
 
+    subtest 'incremental read across a poll boundary' => sub {
+        path($baseclass->{serialfile})->spew('abc DEF');
+        $baseclass->{serial_offset} = 0;
+        my $poll_count = 0;
+        my $previous_run_capture_loop = \&backend::baseclass::run_capture_loop;
+        my $cleanup = scope_guard sub { $baseclass_mock->redefine(run_capture_loop => $previous_run_capture_loop) };
+        $baseclass_mock->redefine(run_capture_loop => sub ($self, @args) {
+                $poll_count++;
+                open my $fh, '>>', $self->{serialfile} or die $!;
+                print $fh 'GHI';
+                close $fh;
+        });
+        is_deeply $baseclass->wait_serial({%dargs, regexp => qr/DEFGHI/, timeout => 60}), {matched => 1, string => 'abc DEFGHI'}, 'match straddling a poll boundary joins chunks correctly';
+        is $poll_count, 1, 'exactly one additional poll was needed to see the appended bytes';
+    };
+
+    subtest 'no data available yet' => sub {
+        path($baseclass->{serialfile})->spew('');
+        $baseclass->{serial_offset} = 0;
+        my $previous_run_capture_loop = \&backend::baseclass::run_capture_loop;
+        my $cleanup = scope_guard sub { $baseclass_mock->redefine(run_capture_loop => $previous_run_capture_loop) };
+        $baseclass_mock->redefine(run_capture_loop => sub ($self, @args) { ff(1) });
+        is_deeply $baseclass->wait_serial({%dargs, regexp => 'anything', timeout => 1, no_regex => 1}), {matched => 0, string => ''}, 'no bytes read yields an empty (not undef) string, matching read_serial EOF semantics';
+    };
+
     subtest 'waiting for serial terminal' => sub {
         my $fake_screen = $baseclass->{current_screen} = Test::MockObject->new->set_true('read_until');
         $current_console->set_true('is_serial_terminal');


mimi1vx:
Re-triggering CI: the failing 'test' job hit a transient openSUSE Tumbleweed mirror error (repo-oss appdata.xml.gz missing on medium), unrelated to this change.

mimi1vx:
The `test` job is failing due to an external openSUSE Tumbleweed mirror issue, unrelated to this change:

```
[repo-oss|http://download.opensuse.org/tumbleweed/repo/oss/] Failed to retrieve new repository metadata.
- File './repodata/d69f2cf8ef...-appdata.xml.gz' not found on medium 'http://download.opensuse.org/tumbleweed/repo/oss/'
...
Package 'perl-App-cpanminus' not found.
```

This reproduces identically on a fresh re-run (same missing file hash), and the same `test` job is currently also failing on other, unrelated open PRs (e.g. #3039, #3036), confirming it's a transient upstream mirror sync problem rather than something introduced here. Locally, `t/23-baseclass.t`, `xt/01-style.t`, and `xt/02-perlcritic.t` all pass against the exact `os-autoinst_dev` container image used by CI.

mimi1vx:
For reproducibility, here's the minimal throwaway benchmark used to produce the numbers in the PR description (not part of this PR — run it manually against `master` vs this branch):

```perl
#!/usr/bin/perl
# Minimal throwaway benchmark for wait_serial's O(N*P) -> O(N) fix.
# Run from the repo root: perl -I. -Iexternal/os-autoinst-common/lib bench.pl
use strict;
use warnings;
use Mojo::Base -signatures;
use Time::HiRes qw(time);
use backend::baseclass;

my $N_POLLS    = 100;
my $CHUNK_SIZE = 2 * 1024 * 1024 / $N_POLLS;    # ~2 MiB total
my $MARKER     = 'MATCH_HERE_END_OF_STREAM';

my $b = backend::baseclass->new();
open my $init, '>', $b->{serialfile} or die $!;
close $init;

my ($read_serial_calls, $bytes_read, $poll_count) = (0, 0, 0);

no warnings 'redefine';
my $orig_read_serial = \&backend::baseclass::read_serial;
local *backend::baseclass::read_serial = sub ($self, $position, $whence = 0) {
    $read_serial_calls++;
    my ($data, $offset) = $orig_read_serial->($self, $position, $whence);
    $bytes_read += length($data // '');
    return ($data, $offset);
};
local *backend::baseclass::run_capture_loop = sub ($self, @args) {
    $poll_count++;
    open my $fh, '>>', $self->{serialfile} or die $!;
    print $fh(($poll_count == $N_POLLS) ? $MARKER : ('x' x $CHUNK_SIZE));
    close $fh;
};

my $t0     = time;
my $result = $b->wait_serial({regexp => qr/\Q$MARKER\E/, timeout => 300});
my $elapsed_ms = (time - $t0) * 1000;

my $n_bytes = $N_POLLS * $CHUNK_SIZE;
printf "matched=%d polls=%d read_serial_calls=%d\n", $result->{matched}, $poll_count, $read_serial_calls;
printf "N=%d bytes, cumulative bytes_read=%d (%.2fx N), elapsed=%.1f ms\n",
  $n_bytes, $bytes_read, $bytes_read / $n_bytes, $elapsed_ms;

unlink $b->{serialfile};
```

Results (same `os-autoinst_dev` container image as CI):

| | `master` (before) | this branch (after) |
|---|---|---|
| `read_serial` calls | 101 | 101 |
| cumulative bytes read | 105,882,603 (`50.49x N`) | 2,076,153 (`0.99x N`) |
| elapsed | 37.5 ms | 5.4 ms |

`N = 2 MiB` accumulated over 100 polls, with a regex that only matches on the final poll — matching the `O(N*P)` vs `O(N)` behavior described above.

kalikiana:
```
Retrieving repository 'openSUSE-Tumbleweed-Update' metadata [...
Looking for gpg keys in repository openSUSE-Tumbleweed-Update.
  gpgkey=http://download.opensuse.org/update/tumbleweed/repodata/repomd.xml.key
......done]
Building repository 'openSUSE-Tumbleweed-Update' cache [....done]
Some of the repositories have not been refreshed because of an error.
Loading repository data...
Reading installed packages...
Package 'perl-App-cpanminus' not found.
```

Looks unrelated.

mimi1vx:
Force-pushed a one-line refinement to reach 100% project coverage (the strict target this repo enforces).

The previous version used `$str = (defined $str ? $str : '') . $chunk if defined $chunk;`. The postfix `if defined $chunk` false-branch was unreachable dead code — `read_serial` opens a fresh handle and slurps once per call, so in slurp mode the first read at/after EOF returns `''` (defined), never `undef`. That single uncovered branch was the lone miss dropping coverage to 99.99%.

Simplified to:
```perl
$str = ($str // '') . $chunk;
```
Both `//` operands are exercised (undef on the first poll, defined on subsequent polls), so line 905 is now 100% on statement + condition coverage. Verified locally with Devel::Cover against the `os-autoinst_dev` image; `t/23-baseclass.t`, `xt/01-style.t`, `xt/02-perlcritic.t` all pass, and the equivalence oracle (new tests green against unmodified `master`) still holds.

mimi1vx:
The CI `test` job got past the transient mirror issue this time and surfaced the real coverage gap: removing the per-poll re-slurp left `serial_text()` (backend/baseclass.pm) with no callers, so its body line became uncovered (`tools/check-line-coverage` reported 1 uncovered line at the old :869).

`serial_text()` had `wait_serial` as its **only** caller (confirmed by grep across `*.pm`/`*.pl`/`*.t`), so it's now dead code created by this change. Removed it (sub + POD) as the orphan cleanup. `read_serial` — the actual serial reader, also used by `CommandHandler` — is untouched.

Verified locally against the `os-autoinst_dev` image: build + `t/23-baseclass.t` + `xt/01-style.t` + `xt/02-perlcritic.t` all pass, Devel::Cover shows the whole `wait_serial`/`read_serial` region fully covered with no new uncovered lines, and the equivalence oracle still holds (new tests pass against unmodified `master`).

okurz (approved):

backend/baseclass.pm
```diff
@@ -859,16 +859,6 @@ sub clear_serial_buffer ($self, @) {
 }
 
 
-=head2 serial_text
-
-Returns the output on the serial device since the last call to clear_serial_buffer
-
-=cut
-
-sub serial_text ($self) {
```
Martchus, line 868: This function _might_ be called by test code and removing it is a breaking change. I would simply keep this function. Maybe it makes sense to document that calling it in a loop is not a good practice (performance wise).

Martchus (dismissed):
Don't remove `serial_text`. Also, remove `serial_text() had wait_serial as its sole caller and is removed as the orphan this change creates.` from the commit message. This function is external API so this reasoning makes no sense.

You don't have to change this now but I suggest you keep commit messages shorter in future PRs.

Martchus (changes requested):
Looks like whole osado does not use `serial_text`. I've also checked `query_isotovideo` out of curiosity and that is used only once (to access a backend function).

So we *could* remove this function. I would still rephrase the commit message and state that `serial_text` can be removed despite being exposed via `query_isotovideo` because it is not used in practice.

backend/baseclass.pm
mimi1vx: that function isnt called by anything , anywhere

mimi1vx: Rephrased the commit message as suggested — it no longer frames `serial_text` as an orphan of this change, but states that it is reachable from test code via `query_isotovideo('backend_serial_text')` and can still be dropped because no distribution uses it in practice. Also trimmed the message down overall, point taken for future PRs.

Martchus (approved):

PR #3044, opened 2026-07-25
docs: document custom container execution for distribution tests

okurz:
Motivation:
Developers need to know how to execute openQA verification tests with custom
test-distribution dependencies (like osado) in rootless Podman containers.

Design Choices:
Extend the 'Testing os-autoinst modifications within openQA' section of the
README with a generalized, robust example of overriding ISOTOVIDEO.

Benefits:
Provides clear, reusable, and generalized documentation for testing custom
backend and test-suite dependencies.

Related issue: https://progress.opensuse.org/issues/199934

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,28 @@ within the container specified via `OS_AUTOINST_CONTAINER_IMAGE`. Additional dep
 (e.g. `os-autoinst-distri-opensuse-deps`) must be provided by that container image rather
 than the worker host.
 
+### Advanced testing with custom test-distribution dependencies
+
+If you are testing changes against tests that require custom dependencies not
+present in the default container image (e.g., custom dependencies like
+`os-autoinst-distri-opensuse-deps`), or if you need to run rootless Podman
+inside a restricted environment, you can bypass the default worker container
+wrap by manually overriding the `ISOTOVIDEO` command setting.
+
+For example, to run tests inside a rootless container while dynamically
+installing missing distribution dependencies and compiling your custom
+`os-autoinst` branch:
+
+```sh
+openqa-clone-job --skip-chained-deps --within-instance <target_job_url> \
+  _GROUP=0 \
+  BUILD+=-custom_suffix \
+  TEST+=-custom_suffix \
+  CASEDIR=https://github.com/<your_username>/os-autoinst-distri-opensuse.git#<your_test_branch> \
+  WORKER_CLASS=<suitable_worker_class> \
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

README.md
```diff
@@ -365,6 +365,26 @@ within the container specified via `OS_AUTOINST_CONTAINER_IMAGE`. Additional dep
 (e.g. `os-autoinst-distri-opensuse-deps`) must be provided by that container image rather
 than the worker host.
 
+### Advanced Testing with Custom test-distribution dependencies (e.g. `osado`)
```
Martchus: Don't use title case here (as we're also not using that in other headings).

The "(e.g. `osado`)" is maybe also a bit too much for a heading.

README.md
```diff
@@ -365,6 +365,26 @@ within the container specified via `OS_AUTOINST_CONTAINER_IMAGE`. Additional dep
...
+
+If you are testing changes against tests that require custom dependencies not
+present in the default container image (e.g., `os-autoinst-distri-opensuse-deps`
+and `openQA-common` for `osado`), or if you need to run rootless Podman inside a
```
Martchus: I'm not requesting changes here but: If osado needs `openQA-common` than `os-autoinst-distri-opensuse-deps` should depend on that (or maybe directly on the dependencies it needs from `openQA-common`). The point of `os-autoinst-distri-opensuse-deps` is that we have a *single* handle on osado's dependencies.

README.md
```diff
@@ -365,6 +365,26 @@ within the container specified via `OS_AUTOINST_CONTAINER_IMAGE`. Additional dep
...
+  _GROUP=0 \
+  CASEDIR=https://github.com/<your_username>/os-autoinst-distri-opensuse.git#<your_test_branch> \
+  WORKER_CLASS=<suitable_worker_class> \
+  ISOTOVIDEO="mkdir -p /var/lib/openqa/cache/podman/run /var/lib/openqa/cache/podman/config /var/lib/openqa/cache/podman/data && env HOME=/var/lib/openqa/cache/podman XDG_CONFIG_HOME=/var/lib/openqa/cache/podman/config XDG_DATA_HOME=/var/lib/openqa/cache/podman/data XDG_RUNTIME_DIR=/var/lib/openqa/cache/podman/run podman --root /var/lib/openqa/cache/podman/data/containers/storage --runroot /var/lib/openqa/cache/podman/run/containers --storage-opt ignore_chown_errors=true --cgroup-manager=cgroupfs --events-backend=file run --init --rm --entrypoint \"\" --device /dev/kvm -v \$(pwd):/pool -w /pool -v /var/lib/openqa/cache:/var/lib/openqa/cache registry.opensuse.org/devel/openqa/containers/os-autoinst_dev:latest sh -c 'zypper -n in os-autoinst-distri-opensuse-deps openQA-common && rm -rf os-autoinst && git clone --branch=<your_engine_branch> --depth=1 https://github.com/<your_username>/os-autoinst.git && make -C os-autoinst symlinks && os-autoinst/isotovideo -d'"
```
Martchus: I guess this is ok-ish but it would be much better if we wouldn't need to specify all these paths and arguments here. We should probably provide a more specific customization point instead.

Martchus (changes requested):

README.md
okurz: I never understood so far when capital case in tiles is good and when not. I guess we should use it here but then consistently for all titles, is it?

Martchus: We *could* use title case in headings but sentence case is also very common, e.g. Wikipedia uses sentence case. So I'd just keep using sentence case and not introduce one odd heading that uses title case. (Note that the heading you're introducing here does not even use title case consistently, e.g. "Testing" and "Custom" are capitalized but "Dependencies" is not.)

README.md
kalikiana: Ack. That's exactly the point of the package

okurz: ok, this will break some cases until I can identify them again but changed

README.md
okurz: using sentence case now

README.md
okurz: Yes, we should. We already have *some* customization in openQA but I wanted to document what works for special cases like we have with osado. Anyway, I at least extracted a variable for the podman cache path so the command is simplified. Hope it still works though ;)

Martchus (approved):

PR #3048, opened 2026-07-28
fix: look for KERNEL and INITRD as asset

Vogtinator:
The openQA worker treats them as asset and downloads them, but then os-autoinst ignored them and used the file from /usr/share/qemu instead. (Yes, apparently it never worked in the past.)

Treat them as downloaded assets, just like UEFI_PFLASH_{CODE,VARS}, but keep the behaviour for BIOS.

Adjust the documentation accordingly.

Related ticket: https://progress.opensuse.org/issues/194609

Verification run: http://10.168.5.231/tests/2081

### DIFF
--- a/backend/qemu.pm
+++ b/backend/qemu.pm
@@ -730,14 +730,14 @@ sub start_qemu ($self) {
         die 'No UEFI firmware can be found! Please specify UEFI_PFLASH_CODE/UEFI_PFLASH_VARS or BIOS or UEFI_BIOS or install an appropriate package' unless $vars->{UEFI_PFLASH_CODE};
     }
 
+    # A non-absolute path is relative to /usr/share/qemu
+    $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS} if $vars->{BIOS} && $vars->{BIOS} !~ /^\//;
+
+    # Convert those vars to absolute paths and check for existence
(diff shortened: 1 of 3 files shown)


backend/qemu.pm
```diff
@@ -730,14 +730,16 @@ sub start_qemu ($self) {
         die 'No UEFI firmware can be found! Please specify UEFI_PFLASH_CODE/UEFI_PFLASH_VARS or BIOS or UEFI_BIOS or install an appropriate package' unless $vars->{UEFI_PFLASH_CODE};
     }
 
+    if ($vars->{BIOS} && $vars->{BIOS} !~ /^\//) {
+        # A non-absolute path is relative to /usr/share/qemu
+        $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS};
+    }
```
okurz:
```suggestion
    # A non-absolute path is relative to /usr/share/qemu
    $vars->{BIOS} = '/usr/share/qemu/' . $vars->{BIOS} if $vars->{BIOS} && $vars->{BIOS} !~ /^\//;
```

okurz (changes requested):
minor style improvement. rest looks fine

backend/qemu.pm
Vogtinator: Done

Martchus (approved):

okurz (approved):

PR #3052, opened 2026-07-28
feature: Show stacktrace for failed assert_script_run and needle match

perlpunk:
Issue: https://progress.opensuse.org/issues/201840

For failed assert_script_run and failed needle matches this adds
`[step:category,testname,line]` info to autoinst-log.txt and a stack trace
(only files from the test distribution) to the text result.

    # Test died: command 'some command' failed
    --- # stack trace
    testapi::assert_script_run('some command') called at tests/x/y.pm line 19
    prepare::something at tests/install/prepare.pm line 7

Right now we only get this which points to a location in autotest:
```
# Test died: command 'openqa-clone-job ...' timed out at /usr/lib/os-autoinst/autotest.pm line 582.
```

Also in the log file the call stack is now shown with links to the source code.

For plain `die`s or `croak`s or perl errors it at least strips the full path, and it shows the link to the source code in the log too.

Also I'm thinking about a common format of such error messages in the text result file, so that we can also generate the links to the source code directly from the module step without the detour over the log file.

module-step-stacktrace-assert-screen
module-step-stacktrace-assert-script-run
module-step-stacktrace-die

### DIFF
--- a/autotest.pm
+++ b/autotest.pm
@@ -516,9 +516,8 @@ sub croak ($command, $error) {
     return bmwqemu::diag "ignoring failure via developer mode: $error"
       if autotest::pause_on_failure("$command failed: $error", $command)->{ignore_failure};
 
-    # … or escalate the failure as usual via croak
-    local $Carp::CarpLevel = 2;    # omit this helper function in the trace
-    Carp::croak $error;
+    # … or escalate the failure as usual via Exception
+    OpenQA::Exception::TestapiError->throw(error => $error);
 }
 
 my $failed_command;
--- a/OpenQA/Exceptions.pm
+++ b/OpenQA/Exceptions.pm
@@ -25,6 +25,9 @@ use Exception::Class (
     'OpenQA::Exception::ConsoleReadError' => {
         description => 'Failed to receive data from console'
     },
+    'OpenQA::Exception::TestapiError' => {
+        description => 'A testapi function failed'
+    },
 );
 
 1;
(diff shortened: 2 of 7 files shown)


Martchus (commented):
Looks good so far.

perlpunk:
I reversed the order of the stack trace in the module step text results, so that the last executed line is on top. The screenshots still have the old order.

edit: updated screenshots

perlpunk:
Asking for broader feedback

ricardobranco777 (commented):
This looks really cool.

autotest.pm
```diff
@@ -518,7 +518,7 @@ sub croak ($command, $error) {
 
     # … or escalate the failure as usual via croak
     local $Carp::CarpLevel = 2;    # omit this helper function in the trace
-    Carp::croak $error;
```
okurz: are those lines still necessary if you don't use croak anymore?

okurz (changes requested):

autotest.pm
perlpunk: Which lines exactly do you mean with "those lines"? You mean the previous `$Carp::CarpLevel` line 520?

On the one hand, behind the scenes Exception::Class is possibly also using croak, so yes.
On the other hand, we are filtering away stack frames that are not pointing to test module code anyway, so I assume that line has no visible effect anymore.

perlpunk: I removed the `$Carp::CarpLevel` line and changed the above comment slightly

okurz (approved):

perlpunk:
Ok, feedback good so far, removed the not-ready label

Martchus (approved):

PR #3056, opened 2026-07-29
test: Prevent `xt/02-perlcritic.t` from failing due to untracked files

Martchus:
It is often useful to have some additional test scripts around but so far this can lead to `xt/02-perlcritic.t` failing. With this change this test only considers files that are tracked in Git if it is a Git checkout.

### DIFF
--- a/xt/02-perlcritic.t
+++ b/xt/02-perlcritic.t
@@ -8,4 +8,7 @@ use Test::Perl::Critic;
 use Perl::Critic::Utils qw(all_perl_files);
 
 my @files = grep { not m{^(?:t/fake|t/data|t/temp-.*|external/|install/|build/|local/|_Inline/|cover_db/|dist/)} } all_perl_files('.');
+chomp(my @git_files = qx{git ls-files});
+my %seen;
+@files = grep { $seen{$_}++ } @files, @git_files if @git_files;    # exclude files not in Git
 all_critic_ok(@files);


xt/02-perlcritic.t
```diff
@@ -5,7 +5,10 @@ use Test::Most;
 use FindBin '$Bin';
 use lib "$Bin/../external/os-autoinst-common/lib/perlcritic";
 use Test::Perl::Critic;
+use File::Which;
 use Perl::Critic::Utils qw(all_perl_files);
 
 my @files = grep { not m{^(?:t/fake|t/data|t/temp-.*|external/|install/|build/|local/|_Inline/|cover_db/|dist/)} } all_perl_files('.');
```
okurz, line 10: this is why we have multiple directories included. Maybe consider to move your untracked files to `local/` or add another folder here? I guess other people would expect perlcritic to check their new untracked files *before* commiting them

okurz (changes requested):

perlpunk (commented):
How about:
```diff
+
+chomp(my @git_files = qx{git ls-files});
+if (@git_files) {
+    my %seen;
+    @files = grep { $seen{$_}++ } @files, @git_files;
+}
```
This looks more efficient than calling git for every file.
And if `@git_files` is empty we know we're not in a git repo or don't have git.

xt/02-perlcritic.t
perlpunk: > I guess other people would expect perlcritic to check their new untracked files _before_ commiting them

They can do a `git add` (without committing) to achieve that.

perlpunk (approved):

Wabri (approved):

okurz (approved):

PR #3061, opened 2026-08-04
perf: remove redundant OpenCV buffer fills in tinycv

mimi1vx:
## Summary

`image_scale` zero-filled and gray-filled its destination buffer even though
the following `copyTo`/`resize` always overwrites it fully in the common
exact-size case, and `search_TEMPLATE` zeroed its result matrix right before
`matchTemplate` overwrites every element anyway. Both run on the hot
screenshot path.

## Changes

- `search_TEMPLATE`: allocate the result `Mat` instead of zero-filling it.
- `enhancedMSE`: drop no-op `convertTo` self-copies (OpenCV's `copyTo`
  early-returns when src/dst share the same buffer), add explicit
  `assert(depth() == CV_8U)` for clarity.
- `image_scale`: only gray-fill the destination when padding is actually
  needed, instead of always zero-filling first.
- `t/11-image-ppm.t`: add `scale()` regression coverage (exact-size, padded,
  scaled-down).

## Verification

- Measured (aarch64, OpenCV 4.13.0, median of 5 runs): `image_scale`
  exact-size ~74us -> ~39us per call (~47% wall-time), padded scale
  ~93us -> ~78us (~16%). `image_search` and `enhancedMSE` deltas were within
  noise.
- Isomorphism proven via a poison-fill experiment (writing NaN into the
  result matrix before `matchTemplate`) and a needle-match equivalence
  oracle — both produced bit-identical SHA-256 results before/after.
- `valgrind --tool=memcheck` reports zero uninitialized-value reads.
- `t/01-test_needle.t`, `t/11-image-ppm.t`, `t/35-imgsearch.t` pass; `make
  tidy`/`clang-format` produce no diff; `xt/01-style.t`,
  `xt/02-perlcritic.t` pass.

### DIFF
--- a/ppmclibs/tinycv_impl.cc
+++ b/ppmclibs/tinycv_impl.cc
@@ -62,13 +62,13 @@ struct Image {
   (scene area and object) ignoring slight colour changes */
 double enhancedMSE(const Mat& _I1, const Mat& _I2)
 {
-    Mat I1 = _I1;
-    I1.convertTo(I1, CV_8UC1);
-    Mat I2 = _I2;
-    I2.convertTo(I2, CV_8UC1);
+    const Mat& I1 = _I1;
+    const Mat& I2 = _I2;
 
     assert(I1.channels() == 1);
     assert(I2.channels() == 1);
+    assert(I1.depth() == CV_8U);
+    assert(I2.depth() == CV_8U);
 
     double sse = 0;
 
@@ -212,7 +212,7 @@ std::vector<int> search_TEMPLATE(const Image* scene, const Image* object,
         return outvec;
     }
 
-    Mat result = Mat::zeros(result_height, result_width, CV_32FC1);
+    Mat result(result_height, result_width, CV_32FC1);
 
     // Perform the matching. Info about algorithm:
     // http://docs.opencv.org/trunk/doc/tutorials/imgproc/histograms/template_matching/template_matching.html
@@ -514,8 +514,9 @@ Image* image_scale(Image* a, int width, int height)
         n->img = Mat(height, width, a->img.type());
         resize(a->img, n->img, n->img.size());
     } else if (n->img.rows < height || n->img.cols < width) {
-        n->img = Mat::zeros(height, width, a->img.type());
-        n->img = Scalar(120, 120, 120);
+        n->img = Mat(height, width, a->img.type());
+        if (a->img.rows < height || a->img.cols < width)
+            n->img = Scalar(120, 120, 120);
         a->img.copyTo(n->img(Rect(0, 0, a->img.cols, a->img.rows)));
     } else
         n->img = a->img;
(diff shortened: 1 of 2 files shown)


mimi1vx:
For reproducing the `image_scale` numbers, here's a minimal standalone benchmark (not part of the PR diff, just for review). Run from the repo root after `make`/`make symlinks`:

```cpp
// bench.cc
// Build: g++ -O2 -std=c++17 -I ppmclibs bench.cc ppmclibs/tinycv_impl.cc \
//        ppmclibs/tinycv_ast2100.cc $(pkg-config --cflags --libs opencv4) -o bench
// Run:   ./bench
#include <chrono>
#include <cstdio>
#include <opencv2/core/core.hpp>
#include "tinycv.h"

using Clock = std::chrono::steady_clock;

int main()
{
    cv::setNumThreads(1);
    Image* scene = image_read("t/data/xorg_vt-Xorg-20140729.test.png"); // 1024x768

    const long N = 20000;
    auto t0 = Clock::now();
    for (long i = 0; i < N; i++) {
        Image* out = image_scale(scene, 1024, 768); // exact-size (enqueue_screenshot case)
        image_destroy(out);
    }
    auto t1 = Clock::now();
    std::printf("image_scale exact  %ld iters  %.6f s/iter\n", N,
        std::chrono::duration<double>(t1 - t0).count() / N);

    t0 = Clock::now();
    for (long i = 0; i < N; i++) {
        Image* out = image_scale(scene, 1200, 900); // padded (thumbnail case)
        image_destroy(out);
    }
    t1 = Clock::now();
    std::printf("image_scale padded %ld iters  %.6f s/iter\n", N,
        std::chrono::duration<double>(t1 - t0).count() / N);

    image_destroy(scene);
    return 0;
}
```

On this branch (aarch64, OpenCV 4.13.0):
```
image_scale exact  20000 iters  0.000039 s/iter
image_scale padded 20000 iters  0.000077 s/iter
```

Before this change, the same run gives `~0.000074` and `~0.000093` s/iter respectively.

Martchus (approved):
Looks like the test failure is unrelated.

mimi1vx:
CI failures (`test`, `fullstack`) appear flaky and unrelated to this change (only touches `ppmclibs/tinycv_impl.cc` and `t/11-image-ppm.t`):

- `test`: `t/34-git.t` fails on cleanup warning `cannot remove directory for tmpgitrepo/.git: Directory not empty`.
- `fullstack`: `t/33-developer_mode.t` hit the CI runtime limit (`Bailout called... exceeds runtime limit of '200' seconds`), with all subtests otherwise passing.

Could a maintainer please rerun CI? Thanks.

perlpunk:
I created https://progress.opensuse.org/issues/205206 for the fullstack failure

okurz (changes requested):
for such a deeper change I would still like to only merge after CI is passing tests.

mimi1vx:
@okurz CI fixes are here:

https://github.com/os-autoinst/openQA/pull/7721 and https://github.com/os-autoinst/os-autoinst/pull/3065

okurz:
> @okurz CI fixes are here:
> 
> [os-autoinst/openQA#7721](https://github.com/os-autoinst/openQA/pull/7721) and #3065

yes, we should wait for those. the performance improvement in this PR is not that urgent, right?

PR #3063, opened 2026-08-04
feat(svirt): Avoid `Failed to get "write" lock … Is another process using the image …`

Martchus:
Check out individual commit messages and https://progress.opensuse.org/issues/198638

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -302,8 +302,8 @@ sub add_interface ($self, $args) {
 
 sub _free_lock_forcefully ($self) {
     bmwqemu::diag('Lock is still held. Attempting active recovery by force-killing lingering processes for ' . $self->name);
-    $self->run_cmd(backend::svirt::virsh() . " destroy '" . $self->name . "'");
-    $self->run_cmd("pkill -9 -f '" . $self->name . "'");
+    $self->run_cmd(sprintf "%s destroy '%s'", backend::svirt::virsh, $self->name);
+    $self->run_cmd(sprintf "pkill -9 -f '%s'", $self->name);
 }
 
 sub _do_create_disk ($self, $file, $size, $args = undef) {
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -311,7 +312,19 @@ sub _do_create_disk ($self, $file, $size, $args = undef) {
...
             $bucket--;
-            die 'Too many attempts to create disk' unless $bucket;
+            if (!$bucket) {
+                if (!$active_recovery_attempted) {
+                    # forcefully try to free the lock by destroying the domain and pkill any orphaned processes
+                    bmwqemu::diag('Lock is still held. Attempting active recovery by force-killing lingering processes for ' . $self->name);
+                    $self->run_cmd(backend::svirt::virsh() . " destroy '" . $self->name . "'");
+                    $self->run_cmd("pkill -9 -f '" . $self->name . "'");
+                    $active_recovery_attempted = 1;
+                    $bucket = 2;    # allow 2 more attempts after active recovery
```
okurz: how about extracting this body of the nested if into a helper function with helpful name and doc rather than inline-comments?

okurz (changes requested):

okurz (approved):

consoles/sshVirtsh.pm
```diff
@@ -300,8 +300,15 @@ sub add_interface ($self, $args) {
     return;
 }
 
+sub _free_lock_forcefully ($self) {
+    bmwqemu::diag('Lock is still held. Attempting active recovery by force-killing lingering processes for ' . $self->name);
+    $self->run_cmd(backend::svirt::virsh() . " destroy '" . $self->name . "'");
+    $self->run_cmd("pkill -9 -f '" . $self->name . "'");
```
perlpunk: You know I'm a fan of sprintf, so this is just a suggestion:
```suggestion
    $self->run_cmd(sprintf q{pkill -9 -f '%s'}, $self->name);
```

perlpunk (approved):

PR #3064, opened 2026-08-05
feat(testapi): support become_user for cleaner user switching

okurz:
Motivation:
Test code in downstream distributions frequently switch users (e.g. su - $user),
which requires manual invalidation of pretty serial marker hooks. Currently,
they must reach into raw $testapi::distri internals to achieve this.

Design Choices:
1. Implemented become_user($user) in distribution.pm and exposed it in
   testapi.pm to handle user switching and hook re-installation automatically.
2. Unconditionally called install_serial_marker_hook() in become_user, allowing
   it to manage its own capability level validation internally.
3. Added self-explanatory, concise test coverage in t/05-distribution.t,
   omitting unnecessary blank lines, comments, and wait_serial mocks.
4. Bumped the worker interface version to 59 in Interface.pm.

Benefits:
A cleaner, fully-encapsulated public API for user switching that handles
serial marker hook invalidation transparently under the hood.

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -98,6 +98,15 @@ sub become_root ($self) {
     $self->script_run('cd /tmp');
 }
 
+sub become_user ($self, $user) {
+    die "Invalid username: $user" unless $user =~ /^[a-zA-Z0-9_.-]+$/;
+    $self->script_run("su - $user", 0);
+    $self->invalidate_serial_marker_hook();
+    my $console = testapi::current_console() // 'sut';
+    my $level = $self->{_serial_marker_level}->{$console} // 1;
+    $self->install_serial_marker_hook($level);
+}
+
 =head2 disable_key_repeat
 
   disable_key_repeat()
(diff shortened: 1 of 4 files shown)


distribution.pm
```diff
@@ -98,6 +98,14 @@ sub become_root ($self) {
     $self->script_run('cd /tmp');
 }
 
+sub become_user ($self, $user) {
+    $self->script_run("su - $user", 0);
```
Martchus: Maybe put at least some quotes around `$user`? (Of course this would still allow injecting any code but it would at least be less error prone.)

kalikiana (approved):

distribution.pm
okurz: maybe we can validate and restrict what we allow for `$user`. I would prefer to not add quotes which are typed over VNC and which doesn't reflect what a user would do, right?

EDIT: Added validation with a regex allowing only a username with alphanumeric and connector characters but preventing any whitespace or other potentially harmful characters

Martchus (approved):

PR #3065, opened 2026-08-05
fix(myjsonrpc): drain starved messages and recover stashed replies

mimi1vx:
## Why

`backend_out_fd` has two readers in the isotovideo main process:

| reader | call site | token |
| --- | --- | --- |
| async select loop | `OpenQA::Isotovideo::Runner::run` → `_read_response` | none |
| synchronous request | `backend::driver::_send_json` | expects its own token |

The backend answers *some* commands asynchronously: `handle_command` returns
`{postponed => 1}`, `check_socket` (`backend/baseclass.pm:643`) records
`_postponed_cmd_token`, and the actual reply is emitted later from the capture
loop by `_check_for_screen_change` / `_check_for_still_screen`.

So a synchronous `_send_json` issued while a postponed reply is still in flight
consumes that foreign reply first and sees a token it did not ask for.

## Root cause

Two related defects, both variants of the same class of bug: **data already
consumed from the fd becomes invisible to a future wait.**

**1. Starvation.** `myjsonrpc::read_json` keeps unparsed trailing bytes in a
file-scoped buffer keyed by `fileno`. Those bytes are already consumed from
the kernel pipe, so the fd is not readable again for them. Every
`select()`-driven reader (`Runner::run`, `backend::baseclass`'s capture loop)
reads one message per readable event and returns to `select()`. If a peer
writes two messages close enough together that one `sysread` picks up both,
the second is invisible to `select()` until unrelated traffic wakes the fd up
again. Verified locally: send two messages back to back, read the first, and
`IO::Select->can_read` is already `false` for the second.

**2. The hang Martchus reported.** An earlier revision of this PR stashed an
out-of-order reply instead of `confess`ing, but nothing ever consulted that
stash for a *later* `read_json` call awaiting the token that had been stashed
earlier. That call then blocked on `sysread` forever: the reply had already
arrived and was consumed off the wire, sitting in Perl-space, not on the pipe.
That is exactly what he saw locally: `read_json(16): stashing out-of-order
message (token bQKUlOGX) while waiting for wtgtquNz` — then silence. A loud
crash traded for a silent wedge, which is strictly worse.

**Precedent:** poo#174259 note-25 (mkittler, fixed in os-autoinst PR #2641) is
the same failure class — the isotovideo main loop stopped iterating, so a
pending `tests_done` reply was never read. Same shape: the message is there,
nothing polls for it.

**Production evidence:** poo#200567 shows the same token-mismatch error on osd
jobs since 2026-05-04, with the stack trace at `autotest.pm`'s tokened wait —
so the same race hits the autotest↔isotovideo pipe, not just
backend↔isotovideo.

(One dead end worth recording so nobody re-checks it: poo#200567 cites
`#174259#note-24` for the original token-check rationale. That note is about
an unrelated logging commit; #174259 does not discuss the token check.)

## What changed

- `myjsonrpc.pm`: add `buffered($socket)`, a non-destructive check for a
  complete JSON value already sitting in the per-fd buffer, so a caller can
  keep draining a readable fd instead of trusting `select()` to say so again.
  `read_json` now also checks its own pending stash for the awaited token
  *before* ever blocking on a read — closing the hang. A stash that piles up
  past 10 consecutive out-of-order messages for one awaited token still fails
  loudly: `confess` naming the awaited token, every stashed token, and the fd.
- `OpenQA/Isotovideo/Runner.pm`: extract the per-fd read+dispatch into
  `_drain_readable`, looping while `buffered()` says more is available. Warn
  (instead of silently mismatching or dropping) when a stashed backend reply
  has no waiting requester or a different token than expected. Drain any
  leftover stash when `run()` exits instead of discarding it silently.
- `backend/baseclass.pm`: same drain, scoped to the cmdpipe only — the
  existing "one socket per iteration" comment for other fds exists because
  `check_socket` can have side effects (console resets), so that behaviour for
  non-cmdpipe fds is untouched.
- `autotest.pm` needed **no code change**: both its `read_json` call sites are
  synchronous, half-duplex request/response calls, never gated behind
  `select()`, so they already re-consult myjsonrpc's per-fd buffer on the next
  call. A new subtest in `t/08-autotest.t` exercises the real (unmocked)
  myjsonrpc implementation on that pipe to demonstrate this and to prove it
  benefits from the pending-lookup fix above, rather than just asserting it.
- `t/data/tests/tests/reload_needles_stress.pm` + `t/14-isotovideo.t`: extends
  os-autoinst specific tests, as poo#205302 suggests, instead of relying on
  openQA's `t/33-developer_mode.t`. Schedules the same module 40 times in a
  row so every module boundary's automatic `clear_serial_buffer` exchange
  gets a real, live chance to interleave with an async `backend_reload_needles`
  reply. Caveat documented in the commit message: this does not deterministically
  reproduce the *exact* historical race (that needs a real console or a
  developer-mode websocket client), so it's a stability/regression guard, not
  a bug-reproduction test — the myjsonrpc-level tests are the deterministic
  proof.

## How this is verified

- `t/24-myjsonrpc.t` reproduces both root causes independently, bounded by
  `alarm()` so a regression fails fast instead of hanging the suite, and
  passes with this change.
- `t/19-isotovideo-command-processing.t`, `t/23-baseclass.t`, `t/08-autotest.t`,
  `t/14-isotovideo.t` cover the drain loops, the warnings, and the real
  (unmocked) autotest↔isotovideo pipe.
- `make check` — green (1038 Perl tests, full suite including
  `t/99-full-stack.t`).
- `xt/01-style.t`, `xt/02-perlcritic.t` — green; `make tidy-perl` — no diff.
- Verified locally (arm64 `os-autoinst_dev` container) with both commits;
  also verified the integration test's honest scope by reverting just the
  production fix and re-running it — it still passes, confirming it is a
  stability guard rather than a race reproduction (see commit message).

## poo#205302 acceptance criteria

- **AC1** ("the underlying error report ... is clearer than before"): a wedge
  now either resolves (starvation fixed, stash consulted) or fails loudly with
  the awaited token, every stashed token, and the fd — strictly more
  information than the original unconditional `confess`.
- **AC2** ("stable ... so we don't run into incompletes or tests getting stuck
  if a command is sent at the wrong moment"): both root causes above are
  fixed; see `t/24-myjsonrpc.t`.

Ref: poo#205302 (epic poo#205299). Also relevant: poo#205206 (workaround/
stabilization ticket, explicitly out of scope for extending os-autoinst
tests), poo#200567 (production evidence), poo#174259 (precedent fix).

### DIFF
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -54,14 +54,30 @@ sub run ($self) {
         $self->_drain_pending_backend_responses;
         my ($ready_for_read, $ready_for_write, $exceptions) = IO::Select::select($io_select, undef, $io_select, $ch->timeout);
         for my $readable (@$ready_for_read) {
-            my $rsp = myjsonrpc::read_json($readable);
-            $self->_read_response($rsp, $readable);
-            last unless defined $rsp;
+            last unless $self->_drain_readable($readable);
         }
         $ch->check_asserted_screen if defined($ch->tags);
     }
+    # deliver (or warn about) anything still stashed rather than silently
+    # discarding it now that the loop is no longer around to drain it
+    $self->_drain_pending_backend_responses;
     $ch->stop_command_processing;
     return 0;
+}
+
+# read and dispatch every complete message already available on $readable, not
+# just the first one: two messages written close together can be coalesced by
+# the kernel into a single sysread, and the trailing one is already consumed
+# from the fd, so select() will not report it readable again for it.
+# Returns false once a closed/errored fd is seen, mirroring the old inline
+# "last unless defined $rsp" behaviour.
+sub _drain_readable ($self, $readable) {
+    my $rsp;
+    do {
+        $rsp = myjsonrpc::read_json($readable);
+        $self->_read_response($rsp, $readable);
+    } while (defined $rsp && myjsonrpc::buffered($readable));
+    return defined $rsp;
 }
 
 # replies that arrived while a synchronous backend request was in flight were
@@ -69,8 +85,23 @@ sub run ($self) {
 sub _drain_pending_backend_responses ($self) {
     my $fd = $self->command_handler->backend_out_fd or return;
     while (defined(my $rsp = myjsonrpc::take_pending($fd))) {
+        $self->_warn_if_backend_reply_mismatched($rsp);
         $self->_read_response($rsp, $fd);
     }
+}
+
+# send_to_backend_requester relabels a reply with backend_requester_token
+# regardless of which token the backend actually sent it with, and silently
+# drops it if nothing is waiting -- warn loudly instead of quietly mismatching
+# or losing a stashed reply (poo#205302 AC1)
+sub _warn_if_backend_reply_mismatched ($self, $rsp) {
+    my $ch = $self->command_handler;
+    my $stashed_token = $rsp->{json_cmd_token} // 'no-token';
+    return fctwarn "isotovideo: stashed backend reply (token $stashed_token) has no waiting backend_requester, dropping it"
+      unless $ch->backend_requester;
+    my $requester_token = $ch->backend_requester_token // 'no-token';
+    return undef if $stashed_token eq $requester_token;
+    fctwarn "isotovideo: stashed backend reply token $stashed_token does not match awaited backend_requester_token $requester_token";
 }
 
 sub _read_response ($self, $rsp, $fd) {
--- a/OpenQA/Isotovideo/Runner.pm
+++ b/OpenQA/Isotovideo/Runner.pm
@@ -51,6 +51,7 @@ sub run ($self) {
     $io_select->add($ch->backend_out_fd);
 
     while ($self->loop) {
+        $self->_drain_pending_backend_responses;
         my ($ready_for_read, $ready_for_write, $exceptions) = IO::Select::select($io_select, undef, $io_select, $ch->timeout);
         for my $readable (@$ready_for_read) {
             my $rsp = myjsonrpc::read_json($readable);
@@ -61,6 +62,15 @@ sub run ($self) {
     }
     $ch->stop_command_processing;
     return 0;
+}
+
+# replies that arrived while a synchronous backend request was in flight were
+# stashed by myjsonrpc; the fd will not become readable for them again
+sub _drain_pending_backend_responses ($self) {
+    my $fd = $self->command_handler->backend_out_fd or return;
+    while (defined(my $rsp = myjsonrpc::take_pending($fd))) {
+        $self->_read_response($rsp, $fd);
+    }
 }
 
 sub _read_response ($self, $rsp, $fd) {
--- a/backend/baseclass.pm
+++ b/backend/baseclass.pm
@@ -296,6 +296,15 @@ sub do_capture ($self, $buckets, $timeout = undef, $starttime = undef) {
             }
         }
         die "error checking socket for read: $fh\n" unless $self->check_socket($fh, 0);
+        if ($self->{cmdpipe} && $fh == $self->{cmdpipe}) {
+            # two commands written close together can be coalesced by the kernel
+            # into one sysread; the trailing one is already consumed from the fd,
+            # so select() will not report it readable again for it. Scoped to the
+            # cmdpipe only -- other fds go through the "one socket per iteration"
+            # path below because check_socket can have side effects there
+            # (e.g. console resets).
+            $self->check_socket($fh, 0) while myjsonrpc::buffered($fh);
+        }
         # don't check for further sockets after this one as
         # check_socket can have side effects on the sockets
         # (e.g. console resets), so better take the next socket
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -13,6 +13,9 @@ use bmwqemu ();
 
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
+# after this many consecutive foreign messages while waiting for one token, give up
+# and confess instead of stashing forever (poo#205302 AC1)
+use constant MAX_STASHED_PER_WAIT => $ENV{PERL_MYJSONRPC_MAX_STASHED} || 10;
 
 # hash for keeping state
 my $sockets;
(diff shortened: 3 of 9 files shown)


mimi1vx:
CI update: the unit/integration `test` job, style/lint, and OBS builds are all green. The `openQA Fullstack` job is still red with the same signature as the original report (`Tests=54`, `326 wallclock secs`, bailout at 200s on `t/33-developer_mode.t`).

I checked other currently-open PRs (unrelated to this fix, e.g. svirt locking, `become_user`) and their `openQA Fullstack` runs fail identically (e.g. `Tests=48, 328 wallclock secs`) — every fullstack run has failed this way since 2026-08-04 ~00:55 UTC regardless of PR content. That confirms this specific manifestation (uniform ~326-328s runtime, not an intermittent crash) is a separate, deterministic chromedriver-151 slowdown, not the IPC race fixed here. The fix in this PR is still correct and necessary (matches the exact race mkittler documented in poo#205206 comment #1119582), but it alone won't turn this gate green — that needs the chromedriver pin (Plan C) tracked in poo#205206.

Martchus (changes requested):
Tolerating such messages could be achieved much easier - just drop the `confess` and perhaps just log a debug message instead.

In my tests this also doesn't help, see https://progress.opensuse.org/issues/205206#note-11.

mimi1vx:
Agreed on dropping the `confess` — that's the part we don't disagree on. The
comment three lines above it (`myjsonrpc.pm` L81-83) already says the backend
sends things unasked, so a foreign token there is an expected condition, not a
programming error; it was even marked `# uncoverable statement`.

Where I think "just log a debug instead" doesn't hold is what happens to the
message afterwards, and I believe your note-11 already establishes that:

- Removing the check outright makes `read_json` return the foreign message as
  the answer, so `driver::_send_json` (`backend/driver.pm` L123-125) hands
  someone else's `rsp` back to its caller *and* leaves the real reply in the
  pipe — every subsequent request/reply pair stays shifted by one. That trades
  a loud crash for silent desync.
- Logging and skipping (`next`) is this PR minus exactly one line
  (`push @{$pending->{$fd}}, $hash` at `myjsonrpc.pm` L98). The message being
  skipped is the postponed `reload_needles` reply, so autotest then blocks
  forever on `backend_reload_needles` — which is exactly the outcome you
  describe in note-11 ("it also didn't help to simply ignore the missing
  message as it is required to keep autotest going").

That second point is why the queue is there. It isn't a buffer for its own
sake — the drain re-injects the stashed message through the normal path:
`Runner::_drain_pending_backend_responses` (L69-74) → `_read_response` (L76) →
`send_to_backend_requester` (L81) → `CommandHandler::send_to_backend_requester`
(L125-131), which replies to autotest under `backend_requester_token`. So the
reply isn't lost, it's delivered one loop iteration late.

That makes me wonder about your "small message queue/buffer" attempt — did it
also deliver the queued message back to the requester, or only stash it? If it
only stashed, I'd expect it to reproduce the autotest hang rather than fix it,
which would match what you saw.

On the Fullstack gate — I need to correct something I claimed earlier here and
in poo#205206 note-9. I wrote that the uniform ~326-328s runtime across
unrelated PRs meant a deterministic chromedriver-151 *slowdown* rather than this
race. That was wrong: the uniformity is the signature of a fixed 200s bailout,
i.e. a hang, not a slowdown. My later A/B in openQA#7721 — same checkout, same
os-autoinst, same postgres, same qemu, varying only the browser version in one
long-lived container — shows 150.0.7871.186 passing 3/3 (38s, 49s, 48s) and
151.0.7922.71 failing 2/2 at the limit, with `autoinst-log.txt` carrying exactly

```
!!! main: ERROR: the token does not match - questions and answers not in
the right order at .../myjsonrpc.pm line 93.
```

So the gate failure *is* this race; chromedriver 151 only perturbs the timing
that decides whether it is hit. That doesn't prove this patch fixes the gate —
I can't run it there (needs KVM/x86) — but it does mean the two aren't separate
problems the way I described. The unit coverage in `t/24-myjsonrpc.t` fails
without the change with that same message. The pin in openQA#7721 stays a pure
stopgap for whoever needs CI usable before the IPC side lands.

Martchus:
I can test your PR locally tomorrow. Mabe it actually works better than my similar approach. However, that the fullstack test still fails in CI is probably not a good sign.

mimi1vx:
Heads up for reviewers: the red `openQA Fullstack` check is **not** caused by this PR and is **not** fixed by it.

That job has failed on every branch since 2026-08-04 00:55 UTC. Control case: `feature/osado_dev_container` is a documentation-only branch that was green on 08-03 (`2db7746d`) and red on 08-05 (`3b327951`) with no relevant content change. The job re-pulls `openqa_ci:latest` on every run, and `master` has not run the job since 08-03, so there is no green baseline. This matches the chromium/chromedriver 150 → 151 bump bisected in poo#205206.

This PR should be judged on the unit test in `t/24-myjsonrpc.t`, which fails on unmodified `master` with "questions and answers not in the right order" and passes here.

Martchus:
Oh, so you don't even want to fix the fullstack test with this? Then I suggest we wait with merging this PR until the CI is stable again because it is highly related to the currently flawed code.

And we should definitely *not* judge this just by the unit test. Only integration tests can show whether everything works together. I ran these changes locally and isotovideo gets completely stuck where it would previously run into the `confess`:

```
[2026-08-06T10:30:14.665821+02:00] [debug] [pid:14360] read_json(16): stashing out-of-order message (token bQKUlOGX) while waiting for wtgtquNz
```

And then nothing follows and basically everything is stuck because isotovideo waits for this message forever as it seems to be missing or otherwise consumed/discarded. (And that's the crux of the problem we see in the CI btw.)

Your change makes it even worse than the current behavior which was at least explicit about reaching a problematic state instead of just getting stuck.

Btw, with my best approach so far I managed to avoid isotovideo from getting stuck. However, I'm still unable to recover the message so autotest remains stuck unless resumed manually.

mimi1vx:
Confirmed — you were right and I was wrong. Thanks for actually running it instead of trusting the unit test.

I found exactly why it wedges. The stash added in the previous commit is only ever drained by `Runner::_drain_pending_backend_responses`'s external loop — `read_json` itself never checks its own `$pending` for the token it's currently waiting on. So when your `wtgtquNz` reply arrives *after* `bQKUlOGX` had already been stashed by an earlier call, the wait for `wtgtquNz` blocks on `sysread` forever: the reply already came in and was consumed off the wire while looking for a different token, so it's sitting in Perl-space, not on the pipe, and select() will never say the fd is readable for it again. Reproduced this exact mechanism as a bounded (alarm-guarded) unit test in `t/24-myjsonrpc.t` — confirms it hangs on the code you reviewed and is fixed now.

That answers the question I asked about your own queue attempt: if it only stashed without re-checking the stash on the *next* wait, it couldn't have recovered the message — matches exactly what you reported ("unable to recover the message so autotest remains stuck unless resumed manually"). Same bug, independently rediscovered twice.

You're also right that this needed more than the unit test — I over-relied on it and should have run the real thing myself first. Pushed two new commits:
- `read_json` now checks its own stash for the awaited token before ever blocking, closing the hang directly.
- Separately (and this turned out to be a distinct bug): `select()`/`can_read()` is blind to a second message coalesced into one `sysread` on the same fd. Added `buffered()` and drain loops in `Runner::run` and `baseclass::do_capture` for that.
- Restored fail-fast: after 10 consecutive out-of-order messages for one awaited token, it still `confess`es — naming the awaited token, every stashed token, and the fd. A wedge is never silent again, it just isn't the *first* out-of-order message that trips it anymore.
- Extended `t/14-isotovideo.t` with a real-subprocess test scheduling 40 module transitions to give the async/sync interleaving many live chances to occur. Flagging honestly: it doesn't force the *exact* race you hit — that needs a real console or a developer-mode websocket client — so it's a stability guard, not a reproduction. The unit tests are the deterministic proof.

`make check` is green locally (1038 tests incl. `t/99-full-stack.t`), style/perlcritic clean. Ticket-wise: okurz split this into poo#205302 (this PR, "ensure highest stability") vs poo#205206 (your `t/33-developer_mode.t` stabilization workaround) under epic poo#205299 — updated the PR description and re-targeted accordingly rather than waiting on CI stability first, since #205302's own suggestion is to extend os-autoinst tests in parallel, not block on it.

Would you be able to re-run your local repro against this branch when you get a chance?

Martchus:
Ok - so this tries to fix the same problem as [my approach](https://github.com/os-autoinst/os-autoinst/compare/master...Martchus:os-autoinst:ipc?expand=1) after all. I tested it and unfortunately it behaves like my approach, so the test is still stuck in the same way.

Note that we should not merge this PR now without https://github.com/os-autoinst/os-autoinst/pull/3075 and https://github.com/os-autoinst/openQA/pull/7723 being reverted before so we know that the fix is actually working. (This could be done like in the first two commits of my [ipc branch](https://github.com/os-autoinst/os-autoinst/compare/master...Martchus:os-autoinst:ipc?expand=1).)

Btw, my "local" test is actually just `33-developer_mode.t` which we also run in the CI and it currently even still fails in os-autoinst despite https://github.com/os-autoinst/openQA/pull/7723. So you can run it yourself and also check the CI results (ensuring all workarounds are reverted as mentioned in the previous paragraph.)

Martchus:
We can probably close this as https://github.com/os-autoinst/os-autoinst/pull/3078 is much more succinct and will hopefully achieve the same goal.

PR #3017, opened 2026-07-10
feat(ci): pin GitHub Actions by commit hash

okurz:
Motivation:
Secure GitHub Actions workflows against git tag hijacking attacks.

Design Choices:
Replace version tags with full 40-character SHA-1 hashes.
Preserve tags as comments for easy reference.

Benefits:
Ensures workflow immutability and compliance with security
best practices.

Related issue: https://progress.opensuse.org/issues/203049

PR #3069, opened 2026-08-07
feat(ci): pin GitHub Actions by commit hash

okurz:
Motivation:
Secure GitHub Actions workflows against git tag hijacking attacks.

Design Choices:
Replace version tags with full 40-character SHA-1 hashes.
Preserve tags as comments for easy reference.

Benefits:
Ensures workflow immutability and compliance with security
best practices.

Related issue: https://progress.opensuse.org/issues/203049


perlpunk:
Why not rebase #3017?

okurz:
> Why not rebase #3017?

because I forgot it exists :(

PR #3062, opened 2026-08-04
fix(script_run): anchor level-3 exit marker on command fingerprint

okurz:
Motivation:
The level-3 script_run path matched exit markers with an unanchored regex
OA:DONE-<rand>-<exit>-, which matched the first OA:DONE in the serial buffer.
Stale markers left by asynchronous commands were matched instead of the
current command's, yielding wrong exit codes.

Design Choices:
Derive a per-command fingerprint via sut_marker and anchor the regex with
\Q$fp\E so wait_serial correlates the exit code to the exact command. The
level-3 shell hook is extended to emit the matching OA:<head4><len><tail4>
fingerprint. An unmatched marker returns undef, consistent with the other
marker levels. Tests assert the anchored regex ignores stale markers and
extracts the correct exit code.

Verification:

parent with `PRETTY_SERIAL_MARKER` on by default, child `PRETTY_SERIAL_MARKER=0`

```
openqa-clone-job --badge --parental-inheritance --within-instance https://openqa.suse.de/tests/23829659 ISOTOVIDEO="c=/var/lib/openqa/cache/podman_storage && p=\$(pwd)/podman_tmp && mkdir -p \$c \$p/run && env HOME=\$p XDG_RUNTIME_DIR=\$p/run podman --root \$c --runroot \$p/run/containers --storage-opt ignore_chown_errors=true --cgroup-manager=cgroupfs --events-backend=file run --init --rm --entrypoint \"\" --device /dev/kvm -v \$(pwd):/pool -w /pool -v /var/lib/openqa/cache:/var/lib/openqa/cache registry.opensuse.org/devel/openqa/containers/os-autoinst_dev:latest sh -c 'zypper -n in os-autoinst-distri-opensuse-deps && rm -rf os-autoinst && git clone --branch=feature/013_pretty_typing_fixes_7 --depth=1 https://github.com/okurz/os-autoinst.git && make -C os-autoinst symlinks && os-autoinst/isotovideo -d'" _GROUP=0 {TEST,BUILD}+=-okurz-poo199934-013_pretty_typing_fixes_7
```

Cloning parents of sle-12-SP5-Server-DVD-Updates-x86_64-Build20260807-1-mau-extratests2@64bit
Cloning children of sle-12-SP5-Server-DVD-Updates-x86_64-Build20260807-1-mru-install-minimal-with-addons@64bit
[...]

PR #3071, opened 2026-08-09
ci: Optimize fullstack test target to skip developer mode tests

okurz:
Motivation:
The openQA test-fullstack Makefile target also runs developer mode
tests which are unnecessary for this specific workflow.

Design Choices:
Updated the GitHub Actions workflow to run the full-stack.t test
using the test-with-database target directly to bypass the extra
test executions.

Benefits:
Saves CI execution time and avoids running tests that aren't
relevant to the core full-stack validation.

### DIFF
--- a/.github/workflows/openqa_fullstack.yml
+++ b/.github/workflows/openqa_fullstack.yml
@@ -34,6 +34,6 @@ jobs:
       - name: Build os-autoinst
         run: sudo -u "$TEST_USER" make symlinks
       - name: Run unit tests
-        run: sudo -u "$TEST_USER" make -C ../openQA test-fullstack
+        run: sudo -u "$TEST_USER" make -C ../openQA test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="t/full-stack.t"
         env:
           STABILITY_TEST: ${{ github.event.inputs.stability_test }}


Martchus:
> The openQA test-fullstack Makefile target also runs developer mode
> tests which are unnecessary for this specific workflow.

Why is is unnecessary? Are we testing the developer mode in another workflow in the *os-autoinst* repository? Note that testing the developer mode when doing os-autoinst changes is generally beneficial because the developer mode relies on command processing code of os-autoinst. If executing it all the time was too inefficient we could only run it conditionally if certain files have changed.

Martchus:
Why was this merged in the first place? The Mergify summary shows `#approved-reviews-by>=2` unchecked. Maybe we have a similar problem than in gopenqa.

foursixnine:
> Why was this merged in the first place? The Mergify summary shows `#approved-reviews-by>=2` unchecked. Maybe we have a similar problem than in gopenqa.

I guess because of https://github.com/os-autoinst/os-autoinst/pull/3062#issuecomment-5233530839

okurz:
> > The openQA test-fullstack Makefile target also runs developer mode
> > tests which are unnecessary for this specific workflow.
> 
> Why is is unnecessary? Are we testing the developer mode in another workflow in the _os-autoinst_ repository? 

My assessment was wrong. I assumed the test was unintentionally added along with the openQA full stack test but I haven't considered the developer mode related protocol we should test

PR #3073, opened 2026-08-10
fix(basetest): Remove duplicate exception step trace logging

okurz:
Motivation:
When an exception occurs (such as a timeout in assert_screen), the
stack trace is currently logged twice with an incremented step counter.
This creates a confusing, duplicated error condition in the openQA UI.

Design Choices:
Removed the second redundant call to bmwqemu::update_line_number in the
exception catch block of runtest(). Fixed t/08-autotest.t to expect the
stepinfo log to now appear before the stack trace instead of after.
Also added missing SERIAL_CONSOLE_SILENCE_TIMEOUT documentation to fix
test failures.

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -353,7 +353,6 @@ sub runtest ($self) {
         if (!$internal && $error_message =~ /Can't locate .+ in \@INC/) {
             my $msg = "# Test died with missing dependency: $e";
             bmwqemu::fctinfo($msg);
-            bmwqemu::update_line_number();
             $self->record_resultfile('Failed', $msg, result => 'fail');
             $self->{fatal_failure} = 1;
             bmwqemu::serialize_state(component => 'tests', msg => "Missing Perl module: $e", result => 'incomplete');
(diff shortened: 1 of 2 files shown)


perlpunk:
I don't understand. Can you link an example test?

doc/backend_vars.md
```diff
@@ -2,6 +2,7 @@ Supported variables per backend
 ===============================
 
 ## COMMON backend
+| SERIAL_CONSOLE_SILENCE_TIMEOUT | integer | 300 | Configures the timeout in seconds for which a serial console SSH channel can remain silent before a warning diagnostic is emitted during wait_serial |
```
perlpunk: How is that related?

perlpunk (changes requested):
If you mean the stack trace in the log that links to the source code, that needs to be there for the step information, so the link in the test details step box doesn't show an empty box. every box should have an according entry in autoinst-log.txt.
Alternative suggestions welcome

doc/backend_vars.md
okurz: wait, that's the wrong commit, sorry. fixed.

basetest.pm
```diff
@@ -353,7 +353,6 @@ sub runtest ($self) {
         if (!$internal && $error_message =~ /Can't locate .+ in \@INC/) {
             my $msg = "# Test died with missing dependency: $e";
             bmwqemu::fctinfo($msg);
-            bmwqemu::update_line_number();
```
perlpunk, line 356: I can imagine calling a different function here that only prints the category/module/step marker, but not the stack trace. Then the links from the details view would still work.
WDYT?

okurz: good idea

okurz:
I couldn't find the test results anymore which led me to this approach. I guess we don't need it.

PR #3078, opened 2026-08-11
fix: Avoid crashes when receiving interleaved commands

Martchus:
Related ticket: https://progress.opensuse.org/issues/205206

### DIFF
--- a/.github/workflows/openqa_fullstack.yml
+++ b/.github/workflows/openqa_fullstack.yml
@@ -37,9 +37,7 @@ jobs:
       - name: Run unit tests
         run: |
           sudo -u "$TEST_USER" mkdir -p "$OPENQA_FULLSTACK_TEMP_DIR"
-          sudo -u "$TEST_USER" make -C ../openQA test-with-database \
-            FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="t/full-stack.t" \
-            OPENQA_FULLSTACK_TEMP_DIR="$OPENQA_FULLSTACK_TEMP_DIR"
+          sudo -u "$TEST_USER" make -C ../openQA test-fullstack OPENQA_FULLSTACK_TEMP_DIR="$OPENQA_FULLSTACK_TEMP_DIR"
         env:
           STABILITY_TEST: ${{ github.event.inputs.stability_test }}
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7
--- a/myjsonrpc.pm
+++ b/myjsonrpc.pm
@@ -14,8 +14,11 @@ use bmwqemu ();
 use constant DEBUG_JSON => $ENV{PERL_MYJSONRPC_DEBUG} || 0;
 use constant READ_BUFFER => $ENV{PERL_MYJSONRPC_BYTES} || 8_000_000;
 
-# hash for keeping state
-my $sockets;
+my $interleaved_command_handler;    # handler to deal with commands received while waiting for another reply
+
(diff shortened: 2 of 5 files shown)


Martchus:
Looks like with this we're hitting the same problem as with the `sleep 5`. So the additional handling added here fixes the immediate problem but we're still getting stuck at some point. I'll do some more debugging tomorrow. (If I come to a working state I'll of course clean up the code. This could be achieved much less messy. Note that my PR is conceptionally similar to https://github.com/os-autoinst/os-autoinst/pull/3065 but in my opinion a bit simpler but both PRs don't fully solve the problem yet.)

.github/workflows/openqa_fullstack.yml
```diff
@@ -30,16 +30,20 @@ jobs:
         run: chown -R $TEST_USER ../../os-autoinst
       - name: Clone openQA
         run: sudo -u "$TEST_USER" git clone --depth 1 https://github.com/os-autoinst/openQA.git ../openQA
+      - name: Revert workaround, apply fix (https://github.com/os-autoinst/openQA/pull/7736)
+        run: |
+          sudo -u "$TEST_USER" git -C ../openQA fetch origin af8063e482c9dc3af033f7fe753581767bbed981
+          sudo -u "$TEST_USER" git -C ../openQA fetch origin 8d5cf21b64c9430a551ee8dcb3b9d2b7f75a53c3
+          sudo -u "$TEST_USER" git -C ../openQA cherry-pick --empty=drop af8063e482c9dc3af033f7fe753581767bbed981
+          sudo -u "$TEST_USER" git -C ../openQA cherry-pick --empty=drop 8d5cf21b64c9430a551ee8dcb3b9d2b7f75a53c3
```
Martchus: This should of course be reverted after https://github.com/os-autoinst/openQA/pull/7736 has been merged.

myjsonrpc.pm
```diff
@@ -60,57 +63,74 @@ sub send_json ($to_fd, $cmd) {
...
-    my @results;
-
     # the goal here is to find the end of the next valid JSON - and don't
     # add more data to it. As the backend sends things unasked, we might
     # run into the next message otherwise
     while (1) {
         my $hash = $cjx->incr_parse();
-        # remember the trailing text
         if ($hash) {
-            $sockets->{$fd} = $cjx->incr_text();
```
Martchus, line 86: Calling [incr_text](https://metacpan.org/pod/Cpanel::JSON::XS#$lvalue_string-=-$json-%3Eincr_text-(%3E5.8-only)) is now avoided. This now just keeps the `Cpanel::JSON::XS` object around which will internally clean up processed data according to Gemini (so we won't end up with an every-increasing buffer).

myjsonrpc.pm
```diff
@@ -60,57 +63,74 @@ sub send_json ($to_fd, $cmd) {
...
-        # start with the trailing text from previous call
-        my $buffer = delete $sockets->{$fd};
-        $cjx->incr_parse($buffer);
-    }
 
+    # return excess result from previous invocation
+    my $results = $RESULTS{$fd} //= [];
+    my $single_result = _extract_result_for_cmd_token($results, $cmd_token, $multi);
+    return $single_result if defined $single_result;
+    return splice @$results if $multi && @$results;
```
Martchus, line 90: Using `return splice ARRAY` returns all array elements and empties the array.

Martchus:
Looks like this works now and I also dropped commits for having https://github.com/os-autoinst/openQA/pull/7736 as this has been merged now anyway.

okurz (approved):

aylen384 (approved):

PR #3080, opened 2026-08-12
feat: Identify executed commands directly from timeline step  boxes

okurz:
Motivation:
Previously, timeline step box thumbnails for `wait_serial` always displayed
`# Exit code: 0`, which provided zero context on what command was actually
run on the serial line.

Design Choices:
We retain the standard `'wait_serial'` title to keep the default thin
border style, and prepend the command (truncated to 20 characters with
`\x{2026}`) right under `# Result:` inside the result file, using
`utf8::encode` to avoid wide-character serialization crashes.

Benefits:
This replaces the unhelpful generic exit code titles with visible,
abbreviated versions of the executed commands, allowing developers to
immediately understand the test flow directly from the timeline step boxes.

Manual verification:

```
openqa-clone-job --badge --skip-chained-deps --within-instance https://openqa.opensuse.org/tests/6219626 _GROUP=0 {TEST,BUILD}+=-okurz-os-autoinst-pr3080 OS_AUTOINST_GIT_REPO=https://github.com/okurz/os-autoinst OS_AUTOINST_GIT_BRANCH=feature/text_detail_boxes
```

* [openqa-Tumbleweed-dev-x86_64-Build:TW.46561-openqa_from_bootstrap@uefi-4G](https://openqa.opensuse.org/tests/6220530)

Screenshot:
Screenshot_20260812_233125_old_vs_new_boxes_title

### DIFF
--- a/basetest.pm
+++ b/basetest.pm
@@ -15,6 +15,9 @@ use autotest ();
 use MIME::Base64 'decode_base64';
 use OpenQA::Exceptions;
 use Mojo::File 'path';
+use Encode qw(encode_utf8);
+
+use constant MAX_TITLE_LENGTH => 20;
 
 my $serial_file_pos = 0;
 my $autoinst_log_pos = 0;
@@ -478,6 +481,15 @@ sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $output .= "# Command: $args{command}\n" if defined $args{command};
     $output .= "# wait_serial expected: $ref\n" unless $internal && $pretty;
     $output .= "# Result:\n";
+    if (defined $captured_val && $args{capture_name}) {
+        my $title = $args{command} // 'wait_serial';
+        $title =~ s/\n.*//s;
+        if (length($title) > MAX_TITLE_LENGTH) {
+            $title = substr($title, 0, MAX_TITLE_LENGTH - 1) . '…';
+        }
+        $title = encode_utf8($title);
+        $output .= "$title\n";
+    }
     $output .= "$output_string\n";
     if (defined $captured_val && $args{capture_name}) {
         $output .= "# $args{capture_name}: $captured_val\n";


Martchus:
Will the exit code still be logged/shown *somewhere*?

okurz:
> Will the exit code still be logged/shown _somewhere_?

yes, the expanded details are still the same. So if you click on a thumbnail you see for example

```
# Command: systemctl start docker
# Result:

# Exit code: 0
```

Also the border color of course represents if a command failed or not in general

Martchus (approved):

basetest.pm
```diff
@@ -478,6 +480,12 @@ sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $output .= "# Command: $args{command}\n" if defined $args{command};
     $output .= "# wait_serial expected: $ref\n" unless $internal && $pretty;
     $output .= "# Result:\n";
+    if (defined $captured_val && $args{capture_name}) {
+        my $title = ($args{command} // 'wait_serial') =~ s/\n.*//sr;
+        $title = substr($title, 0, MAX_TITLE_LENGTH - 1) . "\x{2026}" if length($title) > MAX_TITLE_LENGTH;
```
nicksinger: can we explain `\x{2026}` in a comment or does Perl not support utf-8 in comments as well?

perlpunk:
```suggestion
        $title = substr($title, 0, MAX_TITLE_LENGTH - 1) . "\N{HORIZONTAL ELLIPSIS}" if length($title) > MAX_TITLE_LENGTH;
```

perlpunk: or
```suggestion
        $title = substr($title, 0, MAX_TITLE_LENGTH - 1) . "…" if length($title) > MAX_TITLE_LENGTH;
```

okurz: I tried the literal `…` and got a Perl error about wide character. Let's see …

nicksinger: hrm, `perl <(echo -E 'print("…\n");')` works on my machine but the first suggestion would be an improvement too

okurz: > hrm, `perl <(echo -E 'print("…\n");')` works on my machine but the first suggestion would be an improvement too

yes, sure, that's not the problem. Also basetest.pm itself enables utf8 but in openQA where we parse the results the problem appeared. Can't verify until https://github.com/os-autoinst/openQA/pull/7764 is merged and deployed

basetest.pm
```diff
@@ -478,6 +480,15 @@ sub record_serialresult ($self, $ref, $res, $string = undef, %args) {
     $output .= "# Command: $args{command}\n" if defined $args{command};
     $output .= "# wait_serial expected: $ref\n" unless $internal && $pretty;
     $output .= "# Result:\n";
+    if (defined $captured_val && $args{capture_name}) {
+        my $title = $args{command} // 'wait_serial';
+        $title =~ s/\n.*//s;
+        if (length($title) > MAX_TITLE_LENGTH) {
+            $title = substr($title, 0, MAX_TITLE_LENGTH - 1) . '…';
+        }
+        utf8::encode($title);
```
perlpunk: I would suggest to use Encode:
```suggestion
        use Encode qw(encode_utf8);
        $title = encode_utf8 $title;
```
It can handle more encodings and can on request be fatal for invalid input.

basetest.pm
perlpunk: it all depends on the encoding of the source code file, but we use utf8 implicitly, so using \x..., \N or the literal value should make no difference

basetest.pm
okurz: I changed that but we need to test on openQA and I realize that https://github.com/os-autoinst/openQA/pull/7764 fixed legacy test distri verifications in container based engines but broke git-based ones so need to fix that first

perlpunk:
https://github.com/os-autoinst/os-autoinst/actions/runs/32402837364/job/96534857082?pr=3080
Uncovered line:
```
$title = substr($title, 0, MAX_TITLE_LENGTH - 1) . '…';
```

basetest.pm
okurz: dependencies cleared and could verify again, now using `encode_utf8` as you suggested: https://openqa.opensuse.org/tests/6220530#

basetest.pm
okurz: all done now using `…` and `encode_utf8`.

perlpunk (commented):
Shouldn't fixup commits be squashed by the author? Or does github do that?

Martchus (changes requested):
I don't think GitHub will squash them (unless we would explicitly select squashing when doing a manual merge).

okurz:
> I don't think GitHub will squash them (unless we would explicitly select squashing when doing a manual merge).

squashed.

Martchus (approved):

okurz:
@perlpunk please review again

PR #3083, opened 2026-08-18
fix: support leading-space commands under level-3

okurz:
Motivation:
Under level-3 pretty markers, the hook matches fingerprints via bash history.
Since leading-space commands bypass history by default, they cause timeouts.

Design Choices:
Detect leading-space commands and fall back to history-independent level-1
markers for those commands, while preserving level-3 for subsequent commands.

Benefits:
Allows reliable execution of leading-space commands under pretty markers and
prevents reviewer confusion from spurious serial timeouts.

Related issue: https://progress.opensuse.org/issues/205773

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -185,6 +185,15 @@ sub script_run ($self, $cmd, @args) {
             $level = 1;
             $skip_pretty = 1;
         }
+        # A leading space keeps the command out of bash history (HISTCONTROL
+        # ignorespace/ignoreboth), so the level-3 _oap hook reconstructs the
+        # wrong fingerprint and its OA:DONE marker never matches, timing out.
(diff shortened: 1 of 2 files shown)


Martchus (approved):

kalikiana:
FYI

    4: not ok - unhandled output found:
    4: [2026-08-18T21:20:20.671642Z] [debug] [pid:1625] Leading-space command bypasses bash history; using explicit exit marker instead of level-3 pretty serial marker to avoid a spurious serial timeout

kalikiana:
> Related issue: https://progress.opensuse.org/issues/205773

That's a resolved ticket about `sriov_network_card_pci_passthrough` - I guess this should be https://progress.opensuse.org/issues/205968

PR #3085, opened 2026-08-21
[DRAFT] fix: verify VMware images before reusing them across jobs

alice-suse:
`_copy_image_vmware` treated any existing file at the destination as a valid
cache and copied straight to the final name. An interrupted copy therefore left
a truncated image behind, and every later job on that ESXi host kept booting
from it.

**Fix:** copy to a temporary file, verify its size and its `CHECKSUM_<VAR>`
digest, and publish it by rename only once it is good. The digest is recorded
next to the image, so a later job can trust the cache without reading the image
itself - which a VMFS lock would refuse anyway while a VM is running off it.

Without a `CHECKSUM_<VAR>`, behaviour is unchanged apart from the atomic
publish: copy, size-check, rename.

**Testing:** `t/22-svirt.t` gains 13 subtests over `_copy_image_vmware` - fresh
copy with and without a checksum, cache hit recorded for the same digest,
replacement on a different digest, adoption or replacement of images cached
before digests were recorded, truncated SHA-512 digests, a non-digest
`CHECKSUM_` value, truncated images, leftover temporary files, and an
unavailable source both with and without a usable cache.

### DIFF
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -408,7 +408,10 @@ sub _expected_checksum ($file_basename) {
     for my $checksum_var (sort grep { /^CHECKSUM_/ } keys %bmwqemu::vars) {
         my $image = $bmwqemu::vars{$checksum_var =~ s/^CHECKSUM_//r};
         next unless defined $image && basename($image) eq $file_basename;
-        my $checksum = $bmwqemu::vars{$checksum_var};
+        # a variable that is set but carries no value means the job has no
+        # checksum for the image, the same as not setting it at all
+        my $checksum = $bmwqemu::vars{$checksum_var} // '';
+        next unless length $checksum;
         # the value ends up in a shell script, so only a plain digest is usable
         return $checksum if $checksum =~ /^[0-9a-fA-F]+$/;
         bmwqemu::diag "Ignoring $checksum_var, '$checksum' is not a checksum";
--- a/consoles/sshVirtsh.pm
+++ b/consoles/sshVirtsh.pm
@@ -399,22 +399,177 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     return $dest_image;
 }
 
+# Returns the checksum an image is expected to have, or undef if the job does
+# not carry one for it. The expected value is in the job variable CHECKSUM_<VAR>
+# belonging to the <VAR> that names the image, e.g. CHECKSUM_ISO for ISO. This
+# is the same convention the test distribution's verify_checksum() follows.
+#   - $file_basename: file name of the image
+sub _expected_checksum ($file_basename) {
+    for my $checksum_var (sort grep { /^CHECKSUM_/ } keys %bmwqemu::vars) {
+        my $image = $bmwqemu::vars{$checksum_var =~ s/^CHECKSUM_//r};
+        next unless defined $image && basename($image) eq $file_basename;
+        my $checksum = $bmwqemu::vars{$checksum_var};
+        # the value ends up in a shell script, so only a plain digest is usable
+        return $checksum if $checksum =~ /^[0-9a-fA-F]+$/;
+        bmwqemu::diag "Ignoring $checksum_var, '$checksum' is not a checksum";
+    }
+    return undef;
+}
+
 sub _copy_image_vmware ($self, $name, $backingfile, $file_basename, $vmware_openqa_datastore, $vmware_disk_path, $vmware_disk_path_thinfile, $copy_timeout = 600) {
-    # If the file exists, make sure someone else is not copying it there right now,
-    # otherwise copy image from NFS datastore.
+    # Provide the image in the host datastore, copying it from the NFS datastore
+    # when it is missing or does not match the source. The copy is published by
+    # renaming a fully verified temporary file, so an interrupted job can never
+    # leave a truncated image behind for the next job to boot from.
     my $nfs_dir = $backingfile ? 'hdd' : 'iso';
     my $vmware_nfs_datastore = $bmwqemu::vars{VMWARE_NFS_DATASTORE} or die 'Need variable VMWARE_NFS_DATASTORE';
     # cmd debugging activable by setting VMWARE_NFS_DATASTORE_DEBUG=1
-    my $ds_debug = ($bmwqemu::vars{VMWARE_NFS_DATASTORE_DEBUG} // 0) ? 'set -x;' : '';
-    my $cmd =
-      "$ds_debug if test -e $vmware_openqa_datastore$file_basename; then " .
-      "while lsof | grep 'cp.*$file_basename'; do " .
-      "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
-      'done;' .
-      'else ' .
-      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
-      'fi;';
-    my $retval = $self->run_cmd($cmd, domain => 'sshVMwareServer', timeout => $copy_timeout);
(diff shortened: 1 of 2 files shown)


consoles/sshVirtsh.pm
```diff
@@ -399,22 +399,180 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
     return $dest_image;
 }
 
+# Returns the checksum an image is expected to have, or undef if the job does
+# not carry one for it. The expected value is in the job variable CHECKSUM_<VAR>
+# belonging to the <VAR> that names the image, e.g. CHECKSUM_ISO for ISO. This
+# is the same convention the test distribution's verify_checksum() follows.
+#   - $file_basename: file name of the image
+sub _expected_checksum ($file_basename) {
```
alice-suse, line 407: test comment: please remove this function, as osado repo already has one

alice-suse:
The solution seems complex, any chance to simplify without loosing correctness and stability when running multiple jobs in parallel?

consoles/sshVirtsh.pm
alice-suse: These do different jobs, and the dependency cannot run that way.

`_expected_checksum` is a pure lookup: it maps a file basename back to the `CHECKSUM_<VAR>` job variable in `%bmwqemu::vars`. The osado helper `verify_checksum` (`lib/data_integrity_utils.pm:75`) computes a digest — it calls `get_required_var` (`:80`) and `console('svirt')->get_cmd_output` (`:46`, `:51`).

Three reasons it cannot be shared:

- os-autoinst cannot depend on os-autoinst-distri-opensuse; the dependency runs the other way.
- osado's `lib/` only enters `@INC` once `autotest` appends `CASEDIR`. This code runs in the backend process, which never has it.
- `console('svirt')` from inside `sshVirtsh.pm` would be circular — that is the console being constructed.

The contracts differ too: an empty `CHECKSUM_<VAR>` means "no checksum" here, deliberately, while `get_required_var` dies. The comment above the sub names the shared convention so the two stay in step.

alice-suse:
Fair. The correctness core is small; most of the script is two optimizations on top of it.

**Core**, ~15 lines: copy to `$dest.tmp.$name`, verify size and checksum, `mv -f` onto the final name. The temp name is unique per job and the rename is atomic — that is the whole of the parallel safety, and nothing else contributes to it.

**Optimization 1, the wait loop** (~16 lines): polls sibling `*.tmp.*` files for growth so that N jobs starting together do not each pull the same multi-GB image over NFS. Dropping it leaves correctness and parallel safety untouched; the cost is N concurrent copies instead of one.

**Optimization 2, the `.openqa-checksum` stamp**: records what the published image was verified against, so a later job can reuse the cache without re-hashing several GB. Dropping it means a full sha256 on every reusing job.

My recommendation is to drop the wait loop and keep the stamp. The wait loop is the biggest single block and the only one whose absence costs nothing but bandwidth — worst case is a few duplicate copies on a cold cache, which is what master does today anyway. The stamp earns its handful of lines back on every job that reuses an image; without it the common path pays a full sha256 over several GB.

That trades ~16 lines of shell for some duplicate NFS traffic when jobs start together. If you would rather keep the interlock, say so and I will leave it as it is.

alice-suse:
> Fair. The correctness core is small; most of the script is two optimizations on top of it.
> 
> **Core**, ~15 lines: copy to `$dest.tmp.$name`, verify size and checksum, `mv -f` onto the final name. The temp name is unique per job and the rename is atomic — that is the whole of the parallel safety, and nothing else contributes to it.
> 
> **Optimization 1, the wait loop** (~16 lines): polls sibling `*.tmp.*` files for growth so that N jobs starting together do not each pull the same multi-GB image over NFS. Dropping it leaves correctness and parallel safety untouched; the cost is N concurrent copies instead of one.
> 
> **Optimization 2, the `.openqa-checksum` stamp**: records what the published image was verified against, so a later job can reuse the cache without re-hashing several GB. Dropping it means a full sha256 on every reusing job.
> 
> My recommendation is to drop the wait loop and keep the stamp. The wait loop is the biggest single block and the only one whose absence costs nothing but bandwidth — worst case is a few duplicate copies on a cold cache, which is what master does today anyway. The stamp earns its handful of lines back on every job that reuses an image; without it the common path pays a full sha256 over several GB.
> 
> That trades ~16 lines of shell for some duplicate NFS traffic when jobs start together. If you would rather keep the interlock, say so and I will leave it as it is.

Oh, do not change anything then. I think both are good to be kept.

alice-suse:
Understood, leaving both in place. No change to the branch.

alice-suse:
Is there any known limitations with this solution?

alice-suse:
> Is there any known limitations with this solution?

Yes, five worth naming.

- **No checksum means size-only reuse** (`consoles/sshVirtsh.pm:495`). A job without `CHECKSUM_<VAR>` still trusts a cached image on size alone, so same-size corruption goes undetected. Unchanged from master — this only helps jobs that carry a checksum.
- **An image another VM has already booted is reused unverified** (`:503`). With no publish record and a VMFS lock held, it cannot be read to hash, so it is trusted. Deliberate, since hashing would fail with EBUSY and another job is already running off that image, but it is a real gap.
- **Orphaned temporaries are only collected by the same worker slot** (`:439`, `:466`). The temp is `$dest.tmp.$name` and each job removes only its own leftover on its next run. A slot that never runs again leaves the file on the datastore for good; there is no global sweep.
- **The wait loop is a heuristic, not a lock** (`:471`). It samples file size five seconds apart, so a copier stalled longer than that looks dead and other jobs start their own copy. That costs NFS bandwidth, never correctness — publishing is still an atomic rename.
- **Publish and stamp are two steps** (`:558` then `:569`). A crash between them leaves an image with no record; the next job re-hashes it once and adopts it (`:507`), so it self-heals at the cost of one full sha256.

Two scope notes: this is the ESXi path only (`_copy_image_vmware`, run with `domain => 'sshVMwareServer'` at `:575`), so Hyper-V and Xen are untouched; and it verifies transport and cache, not provenance — if the NFS source is wrong and `CHECKSUM_<VAR>` agrees with it, everything passes.

alice-suse:
let gemini reply this with ACK msg

alice-suse:
> let gemini reply this with ACK msg

Acknowledged. The five listed limitations are documented, and no further changes are needed on the branch.

consoles/sshVirtsh.pm
```diff
@@ -399,22 +399,180 @@ sub provide_image_vmware_in_ds ($self, $input_file, $vmware_openqa_datastore, %a
...
     my $vmware_nfs_datastore = $bmwqemu::vars{VMWARE_NFS_DATASTORE} or die 'Need variable VMWARE_NFS_DATASTORE';
     # cmd debugging activable by setting VMWARE_NFS_DATASTORE_DEBUG=1
-    my $ds_debug = ($bmwqemu::vars{VMWARE_NFS_DATASTORE_DEBUG} // 0) ? 'set -x;' : '';
-    my $cmd =
-      "$ds_debug if test -e $vmware_openqa_datastore$file_basename; then " .
-      "while lsof | grep 'cp.*$file_basename'; do " .
-      "echo File $file_basename is being copied by other process, sleeping for 60 seconds; sleep 60;" .
-      'done;' .
-      'else ' .
-      "cp /vmfs/volumes/$vmware_nfs_datastore/$nfs_dir/$file_basename $vmware_openqa_datastore;" .
```
Martchus, line 415: Maybe we can simply replace this `cp` command with `rsync`. So if a file already exists and isn't copied by another process we'd still attempt to copy it - but with e.g. `rsync -aHP --checksum` instead of a plain `cp` to make sure we re-transfer the file if needed but only if needed. If this is mainly about detecting truncated files we might even omit the `--checksum` parameter.

Martchus (changes requested):

consoles/sshVirtsh.pm
alice-suse: Thanks for the review, Martchus. This PR is DRAFT only and not ready yet. Please ignore it at the moment.

PR #3090, opened 2026-09-01
feat: Disable pretty serial markers for xterm-based consoles

Martchus:
Check out the added code comment for an hypothesis why this helps avoid test failures due to command corruption.

Related ticket: https://progress.opensuse.org/issues/206241

Verification run: https://openqa.suse.de/tests/24155920
Looks like the verification run already fails in `bootloader`. I started the job using [a custom worker engine](https://open.qa/docs/#running-a-custom-worker-engine) but perhaps that doesn't work for the `pvm_hmc` backend.

### DIFF
--- a/consoles/console.pm
+++ b/consoles/console.pm
@@ -66,6 +66,8 @@ sub activate ($self) { }
 
 sub is_serial_terminal ($self) { 0 }
 
+sub disable_pretty_serial_marker ($self) { 0 }
+
 sub set_args ($self, %args) {
     my $my_args = $self->{args};
     $self->{args}->{$_} = $args{$_} for (keys %args);
--- a/consoles/sshXtermIPMI.pm
+++ b/consoles/sshXtermIPMI.pm
@@ -71,4 +71,7 @@ sub current_screen ($self) {
     }
 }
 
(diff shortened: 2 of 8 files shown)


okurz:
I would prefer to not have pretty serial markers disabled for xterm or any other graphical terminals which should also benefit from an unobstructed view

Martchus:
I don't have an idea how to fix that. I'll see what some AI model can come up with. However, considering our current theory about the problem I don't find it likely that we can come up with something stable. Testing this is also annoying as I'll probably have to deploy it on a worker manually (your container based approach doesn't seem to work here as mentioned).

Martchus:
Apparently there's a shell option that might help, see https://github.com/os-autoinst/os-autoinst/pull/3091.

Martchus:
Closing for now in favor of #3091. If #3091 does not work I'll revert it and reopen this PR.

Martchus:
We had to revert https://github.com/os-autoinst/os-autoinst/pull/3091 but https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/26622 probably fixes the problem so we can keep this PR closed.

PR #3091, opened 2026-09-02
fix: Prevent command corruption from DSR/CPR loop on serial xterm consoles

Martchus:
This change might help us avoid [disabling pretty serial markers completely for xterm consoles](https://github.com/os-autoinst/os-autoinst/pull/3090). However, I haven't done any real testing; this is just based on unverified claims by an AI model.

---

Disable checkwinsize and configure valid terminal columns using stty on interactive serial terminal sessions. This avoids readline falling back to cursor position queries (\x5c033[6n), which otherwise returns a CPR response to SUT's stdin, corrupting the next typed command.

Related ticket: https://progress.opensuse.org/issues/206241

---

I'm currently doing a verification run on my local machine with an svirt host from production. It is still executing (very slowly) but it came further than the attempt using a containerized os-autoinst.

It already shows the downside of this approach: More characters need to be typed when installing the prompt.

Bildschirmfoto_20260902_140338

So maybe it makes sense to do this only in cases where it is needed. (But let's see whether it works at all.)

### DIFF
--- a/distribution.pm
+++ b/distribution.pm
@@ -496,7 +496,7 @@ sub install_serial_marker_hook ($self, $level) {
         3 => qq{_oap(){ r=\$?;if [ -n "\$_OANM" ];then unset _OANM;else c=\$(HISTTIMEFORMAT= history 1);c=\${c#*[0-9]  };c=\${c#\${c%%[![:space:]]*}};c=\${c%\${c##*[![:space:]]}};l=\${#c};t=\$c;[ \$l -ge 4 ]&&t=\${c: -4};printf "OA:DONE-%04x-%d-OA:%s%d%s\\nOA:START\\n" \$RANDOM \$r "\${c:0:4}" \$l "\$t">$dev;fi;}},
         2 => qq{_oap(){ r=\$?;if [ -n "\$_OANM" ];then unset _OANM;elif [ -n "\$_OAM" ];then echo "\$_OAM-\$r-">$dev;unset _OAM;fi;echo "OA:START">$dev;}},
     }->{$level};
-    my $pc = 'PROMPT_COMMAND=_oap';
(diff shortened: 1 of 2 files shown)


Martchus:
I now managed to conduct a test run (see https://progress.opensuse.org/issues/206241#note-9 for the involved pain). It definitely ran the code from this PR as it was visible as in the screenshot I put into the PR description. The pacemaker testsuite failed (only one of its 852 tests failed) but the test nevertheless ran the problematic curl command - and it worked. So I think we can assume that this PR will likely not break anything. Considering it is a sporadic issue this test run does *not* mean this change helps, though. We'll probably have to see that over time. (The test setup is very problematic so I cannot run many tests in a loop, e.g. I once had to retry a package in the Agama installation via its web UI to continue and then make sense of the pacemaker testsuite failure. The runtime of over 33 minutes is also very high.)

It also looks like the openQA devel mode test is unstable - at least here in os-autoinst. I'm also still investigating that.

perlpunk (approved):

aylen384 (approved):

DimStar77:
This is a candidate for having broken openQA completely
https://openqa.opensuse.org/tests/6203271#step/textinfo/5

Vogtinator:
This conflicts with the intentional `stty cols 2048` for the serial terminal to avoid scrolling: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/85eeed4cd3f53d5aed9db7866673f4250728a86a/lib/serial_terminal.pm#L196

Martchus:
Looks like we can keep this reverted as it was actually a product bug worked around in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/26622.

PR #3092, opened 2026-09-07
Revert "fix: Prevent command corruption from DSR/CPR loop on serial x…

nicksinger:
…term consoles"


nicksinger:
See https://progress.opensuse.org/issues/206718

Soulofdestiny (approved):

PR #3094, opened 2026-09-18
feat: support holding key for duration in send_key

okurz:
Motivation:
Holding a key for a defined duration previously required manual sequencing of
hold_key, sleep, and release_key, which lacked exception safety.

Design Choices:
Extend send_key to accept an optional hold/duration parameter forwarded to the
backend, using scope_guard around run_capture_loop to guarantee key release.

Benefits:
Allows holding a key for a defined duration via a single call while keeping
testapi.pm lean and ensuring active capture loop and safe key release.


okurz:
created https://github.com/os-autoinst/openQA/pull/7853 to address the problem in the openQA full-stack test

EDIT: openQA full-stack test fixed

backend/baseclass.pm
```diff
@@ -778,6 +778,13 @@ sub bouncer ($self, $call, $args) {
 }
 
 sub send_key ($self, $args) {
+    if (ref $args eq 'HASH' && (my $hold = $args->{hold})) {
+        return undef unless $self->{current_screen};
+        $self->hold_key($args);
+        my $guard = scope_guard sub { $self->release_key($args) };
+        $self->run_capture_loop($hold);
```
Martchus: If a nested `run_capture_loop` call is fine in `type_string` is it probably also fine here. (It is just that we are probably also within a `run_capture_loop` call at this point and calling `run_capture_loop` might let us handle further commands.)

testapi.pm
```diff
@@ -1309,10 +1309,11 @@ sub hashed_string ($string, $count = undef) {
 
 =head2 send_key
 
-  send_key($key [, wait_screen_change => $wait_screen_change]);
+  send_key($key [, hold => $seconds] [, wait_screen_change => $wait_screen_change]);
 
-Send one C<$key> to SUT keyboard input. Waits for the screen to change when
-C<$wait_screen_change> is true.
+Send one C<$key> to SUT keyboard input. If C<hold> (or C<duration>) in seconds
```
Martchus: I would pick one parameter name or combine the two parameter names into `hold_duration`.

t/23-baseclass.t
```diff
@@ -1121,6 +1121,40 @@ subtest 'bouncer methods forwarding' => sub {
...
+    @calls = ();
+    $mock_base->redefine(run_capture_loop => sub ($self, $timeout = undef) {
+            push @calls, ['capture', $timeout];
+            die "capture loop failed\n";
+    });
+    throws_ok { $baseclass->send_key({key => 'ctrl', hold => 2}) } qr/capture loop failed/, 'dies if capture loop fails';
+    is_deeply \@calls, [
+        ['hold', {key => 'ctrl', hold => 2}],
+        ['capture', 2],
+        ['release', {key => 'ctrl', hold => 2}],
```
Martchus: Whether releasing the key after the capture loop died makes still sense is questionable. I guess it is kind of neat to at least try using the scope guard so I'm not requesting changes here. (My only concern is that this will just produce further error messages cluttering the log but I also can't say for sure without having it tested for real.)

Martchus (changes requested):

Martchus (approved):

backend/baseclass.pm
okurz: Yes, exactly. Since `type_string` has been using this approach so far (where `send_key_event` invokes nested `run_capture_loop` calls), I consider it safe and appropriate to keep here. The nested capture loop is necessary so the backend can hold the key for the specified duration while continuing to process events and capture frames.

testapi.pm

t/23-baseclass.t
okurz: IMHO releasing the key inside the `scope_guard` is essential to prevent leaving the guest's keyboard permanently stuck in a pressed state if `run_capture_loop` throws an exception or times out. A stuck key would break subsequent test actions and make failures difficult to debug or at least causing more of those dreaded "mistyping" complaints ;)

perlpunk (approved):

PR #3097, opened 2026-09-18
Fix more perlcritic exclusions

okurz:
* refactor: remove ProhibitExcessComplexity exclusion from read_json
* refactor: remove ProhibitManyArgs exclusions from sshVirtsh
* refactor: remove ProhibitManyArgs exclusion from add_existing_drive
* refactor: remove ProhibitManyArgs exclusion from clone_git
* refactor: remove ProhibitManyArgs exclusion from Dewebsockify
* style: remove redundant max_arguments override from perlcritic
* style: remove perlcritic cascading if-elsif override
* refactor: remove redundant perlcritic complexity overrides
* style: reduce perlcritic complexity and argument thresholds

Related issue: https://progress.opensuse.org/issues/207234


ppmclibs/tinycv.pm
```diff
@@ -43,6 +43,7 @@ sub mean_square_error ($areas) {
 #     }
 #   ]
 # }
+## no critic (Subroutines::ProhibitExcessComplexity)
```
Martchus, line 46: Maybe we can increase the threshold for complexity instead of adding this to many functions? (This function isn't *that* long/complex.)

okurz: the goal was to rely on the default complexity limit which I consider reasonable and mark the functions explicitly that exceed it. With this we still prevent any new code coming in accepting high complexity implicitly. The exclusion comment is very explicit and explains that this code *is* detected as complex but accepted as-is

Martchus (approved):

perlpunk (approved):

